PR #474, opened 2013-06-14
Box2dLights (jni version) integration into libgdx

Petititi:
Hi!
I just finished some JNI optimisations for box2dLight. The performance are very good!
I also integrate the box2dLight into libgdx in order to have access to box2d's native address.
I tried to not impact libgdx too much but had to add 2 function in src\com\badlogic\gdx\physics\box2d\World.java...

Please tell me if I can do something to make easier the integration into libgdx!

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [0.9.9]
+- added box2dLight project
 - removed FileHandle#file(), has no business in there.
 - removed box2deditor
 - removed custom typedarrays in gwt backend
--- a/gdx/src/com/box2dLight/box2dLight/Box2dLight.java
+++ b/gdx/src/com/box2dLight/box2dLight/Box2dLight.java
@@ -1,20 +1,20 @@
 /*
-* Copyright (c) 2013 Thibault Lelore (based on http://box2dlights.googlecode.com)
-*
-* This part of software is provided 'as-is', without any express or implied
-* warranty.  In no event will the authors be held liable for any damages
-* arising from the use of this software.
-* Permission is granted to anyone to use this software for any purpose,
-* including commercial applications, and to alter it and redistribute it
-* freely, subject to the following restrictions:
-* 1. The origin of this software must not be misrepresented; you must not
-* claim that you wrote the original software. If you use this software
-* in a product, an acknowledgment in the product documentation would be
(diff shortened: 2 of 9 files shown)


haimat:
It has been a while since this issue has been discussed, and I fully understand the reasons why Mario and Nate don't want to include box2dlights directly into libGDX core. However, such a JNI version of box2dlights would be really awesome, in particular if you use libGDX mainly on the Desktop, as I do.

Thus, has anything happended in this regard?
Any chance to have a JNI version of box2dlights?

NathanSweet:
Having box2dlights separate from libgdx core doesn't stop it from using JNI. I believe Mario is even willing to build non-core projects on the libgdx build server.

haimat:
That sounds really great, thanks for the fast reply Nate! Unfortunately I have no experience with JNI, so I don't know what would be necessary to have the JNI version of box2dlights. I hope Petititi still reads this thread, or someone else with more experience than me jumps in ;-)

Petititi:
HI!
I'm not a JNI expert (so correct me if i'm wrong) but I see a major issue of not include box2dlights into box2d build : they will not share the same memory space (?That's the thing I'm not sure...). So box2dlights will not have direct access to bodies list.
But if Nate is OK to include box2dlights into box2D, I can take some of my spare time to look how we can do this...

MobiDevelop:
I'd prefer not to integrate third-party stuff into the box2d extension as
it will complicate updating box2d.

xoppa:
> they will not share the same memory space (?That's the thing I'm not sure...). So box2dlights will not have direct access to bodies list.

That won't be a problem.

haimat:
That would be so awesome.
Thanks to all of you guys!

Petititi:
Right now, I have no free time, but by mid February, I will try to extract box2dlights from box2d project!
I will probably have some problem to add external dependencies to box2d (the libgdx compilation process is quite complicated, isn't it), but I will try something!

haimat:
Thank you very much, I really appreciate your efforts!

nhydock:
Seeing that there's already the box2DLights extension available under the libgdx umbrella, why not have the JNI pull request and integration directly in that instead of libgdx main?

https://github.com/libgdx/box2dlights

PR #1014, opened 2013-11-28
added valueOf(String keyname) and toString(int keycode) to Input.Keys

jrenner:
Allows getting a string representation of the hardcoded int keycodes without using an enum.

one concern: BUTTON_CIRCLE and F12 both map to 255, I chose to handle it by ignoring BUTTON_CIRCLE, since F12 is much more likely to be pressed (what is BUTTON_CIRCLE anyways?)

~~The lookup in valueOf could be made faster (it iterates through all keycodes), but using an IntMap means minimal memory footprint, and it should be fast enough for vast majority of cases.~~

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -240,7 +240,8 @@ public interface Input {
 		public static final int F11 = 254;
 		public static final int F12 = 255;
 
-		/** @return a human readable representation of the keycode */
+		/** @return a human readable representation of the keycode.
+		 *  The returned value can be used in {@link Input.Keys#valueOf(String)} */
 		public static String toString(int keycode) {
 			if (keycode < 0) {
 				throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


JFixby:
Sorry, but why ``` BUTTON_CIRCLE == F12```?

Darkyenus:
Because:
> one concern: BUTTON_CIRCLE and F12 both map to 255, I chose to handle it by ignoring
> BUTTON_CIRCLE, since F12 is much more likely to be pressed (what is BUTTON_CIRCLE anyways?)

JFixby:
Why BUTTON_CIRCLE and F12 are mapped to the same keycode?

Darkyenus:
Ask whoever mapped them originally, nobody (here) even knows what BUTTON_CIRCLE is.

JFixby:
Looks like game-pad buttons



```
/** Keys.
	 * 
	 * @author mzechner */
	static public class Keys {
        ...

		public static final int BUTTON_CIRCLE = 255;
		public static final int BUTTON_A = 96;
		public static final int BUTTON_B = 97;
		public static final int BUTTON_C = 98;
		public static final int BUTTON_X = 99;
		public static final int BUTTON_Y = 100;
		public static final int BUTTON_Z = 101;
		public static final int BUTTON_L1 = 102;
		public static final int BUTTON_R1 = 103;
		public static final int BUTTON_L2 = 104;
		public static final int BUTTON_R2 = 105;
		public static final int BUTTON_THUMBL = 106;
		public static final int BUTTON_THUMBR = 107;
		public static final int BUTTON_START = 108;
		public static final int BUTTON_SELECT = 109;
		public static final int BUTTON_MODE = 110;
```

I guess someone should fix it and change the mapping. Should I open an issue?

PR #1054, opened 2013-12-04
Added UDP support to LibGDX

johnnyapol:
Note: Missing support for GWT, but not sure if possible. Looking into
alternatives.

This adds the ability to use UDP datagrams in LibGDX. Complete with documentation.
Note: This creates it's own interface UDPSocket due to the inability to (easily) use Socket.java without some breakage


kastork:
Have you guys considered Vertx for the networking stuff?  It has a sockjs server (and client) for java and for javascript.  IMO Sockjs is superior to a raw websocket impelementation because of the fallback modes, and the whole event bus model of Vertx seems like a good fit for gaming systems.

The server side java is all based on Netty, so you can lose all the servlet container cruft too.

http://vert-x3.github.io

PR #1646, opened 2014-04-06
Created generic KeyFrames class based on g2d Animation class.

davidminor:
This extends the functionality of the Animation class to objects
other than TextureRegion by introducing a new class KeyFrames<T>. The
Animation class remains as an implementation of KeyFrames<TextureRegion>.

I've often found myself reimplementing the functionality of the Animation class to keep track of other things over a period of frames, e.g. collision during an animation, so I thought maybe this generic class would be helpful to other people as well. It should be backwards compatible except for the PlayMode enum moving to KeyFrames.

Let me know if this is acceptable and I'll fill out the CLA.

Thanks!

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -16,7 +16,6 @@
 
 package com.badlogic.gdx.graphics.g2d;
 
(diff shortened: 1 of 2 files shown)


mustii82:
so we have to parse getKeyFrames() like this?

((TextureRegion)walkright.getKeyFrames()[0]).getRegionWidth()

on LibGDX 1.9.5 this generic version is merged.

Issue #1800, opened 2014-05-04
Add option to scale NinePatch


obigu:
The current NinePatch implementation is not scalable, making it dependant on the Viewport dimensions.

Code example:

```
package com.test.ninepatch;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.Texture;
import com.badlogic.gdx.graphics.g2d.NinePatch;
import com.badlogic.gdx.scenes.scene2d.utils.NinePatchDrawable;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;
import com.badlogic.gdx.scenes.scene2d.Stage;
import com.badlogic.gdx.scenes.scene2d.ui.Button;
import com.badlogic.gdx.scenes.scene2d.ui.Button.ButtonStyle;
import com.badlogic.gdx.utils.viewport.ExtendViewport;

public class NinePatchTest extends ApplicationAdapter {

    private static final float VIEWPORT_HEIGHT = 120f;
    private static final float VIEWPORT_WIDTH = 180f;
//  private static final float VIEWPORT_HEIGHT = 1200f;
//  private static final float VIEWPORT_WIDTH = 1800f;
    Stage stage;
    SpriteBatch batch;
    NinePatchDrawable ninepatchDrawable;
    Button button;

    @Override
    public void create() {
        batch = new SpriteBatch();
        Texture texture = new Texture("ninepatch.png");
        NinePatch ninePatch = new NinePatch(texture, 24, 24, 24, 24);
        // ninePatch.setScale(0.3f, 0.3f);
        NinePatchDrawable ninepatchDrawable = new NinePatchDrawable(
                ninePatch);
        ButtonStyle style = new ButtonStyle(ninepatchDrawable,
                ninepatchDrawable, ninepatchDrawable);
        Button button = new Button(style);
        button.setSize(VIEWPORT_WIDTH / 2f, VIEWPORT_HEIGHT / 2f);
        button.setColor(Color.BLUE);
        button.setPosition(VIEWPORT_WIDTH / 4f, VIEWPORT_HEIGHT / 4f);
        stage = new Stage(new ExtendViewport(VIEWPORT_WIDTH, VIEWPORT_HEIGHT) {
        }, batch);
        stage.addActor(button);

    }

    @Override
    public void render() {
[...]

PR #1801, opened 2014-05-04
- Add option to scale Ninepatch

obigu:
See https://github.com/libgdx/libgdx/issues/1800

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/NinePatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/NinePatch.java
@@ -57,6 +57,8 @@ public class NinePatch {
 	private int idx;
 	private final Color color = new Color(Color.WHITE);
 	private int padLeft = -1, padRight = -1, padTop = -1, padBottom = -1;
+	private float scaleX = 1f;
+	private float scaleY = 1f;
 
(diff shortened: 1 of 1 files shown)


dylanritter:
Hi,  is there a convenient way to set a scale without having to manually construct and then pass in the drawable to an actor ?  For example, when creating a TextButton it's very convenient to specify a style name and then leave it at that, however with the 'setScale' method on the ninepatch itself it seems that it would require instantiating each of the possible drawables (up, down, etc.) and scaling those then setting them up in the button, thus destroying the elegance of using the style.

I guess what i'm looking for is that the setScale method on the button instance would apply itself to the backgrounds in the style being used.  

While we are here, i haven't found any effect of using the setScale method on an actor, everything seems to draw at the same size regardless and have always had to resort to manually scaling actors in setSize or .width().height() when using a table.

Thank you,

Dylan

NathanSweet:
This is an issue tracker, IRC or the forums are more appropriate for questions.
https://github.com/libgdx/libgdx/wiki/Getting-help
https://github.com/libgdx/libgdx/wiki/Scene2d.ui#rotation-and-scale
It sounds like you want to create a scaled up style, then use that for your buttons. Most people use multiple ninepatch regions each of the desired size rather than scaling their images.

dylanritter:
Thank you for your reply.  I have further questions on this so will go to the forums as you suggest.
Dylan

PR #1936, opened 2014-06-05
SpotLight per fragment lightning

VinceAngel:
Ok let's explain in a few words : here is my implementation for the gdx lights : I added the spotlight, and make a per fragment rendering shader BUT this is not working completely and I don't know what is the issue so any help would be very appreciated.
The Libgdx tests are useful to reveal what's wrong : see LightTest, and SpotLightTest.

Also notice there are warnings in the shader : 
```
Shader compilation : 
Shader compilation : 
Shader compilation : Vertex info
-----------
0(166) : warning C7514: OpenGL does not allow varying of type struct DirectionalLight[2]
0(183) : warning C7514: OpenGL does not allow varying of type struct PointLightInterpolated[5]
0(208) : warning C7514: OpenGL does not allow varying of type struct SpotLightInterpolated[2]

Fragment info
-------------
0(82) : warning C7514: OpenGL does not allow varying of type struct DirectionalLight[2]
0(99) : warning C7514: OpenGL does not allow varying of type struct PointLightInterpolated[5]
0(125) : warning C7514: OpenGL does not allow varying of type struct SpotLightInterpolated[2]

```
Not sure this is the issue, maybe yes... hoping there is an expert among you.

I also would like to add the normal map rendering, but one thing at a time, unless you can submit it if you want, with pleasure !

Thanks !

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/Environment.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/Environment.java
@@ -18,10 +18,7 @@ package com.badlogic.gdx.graphics.g3d;
 
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Cubemap;
-import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
-import com.badlogic.gdx.graphics.g3d.environment.DirectionalLight;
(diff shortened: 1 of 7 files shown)


badlogic:
@xoppa please close this if you think we won't merge it.

PR #1939, opened 2014-06-06
libGDX game crashes on ART, Google's successor for dalvik.

schleinzer:
Reason is the allocation of a ByteBuffer of size "0" in "IndexBufferObject". This breaks all libGDX Meshes which don't make use of indexes and set the "maxIndexes" to 0 (which in turn breaks quite a lot of libGDX apps on ART I assume.)

While allocating a 0-sized ByteBuffer is perfectly fine according to the javadoc, ART has a bug that makes programs crash if they try to do so. Only sizes >0 do work.

This is a bug in ART admitted and even fixed by Google with https://android-review.googlesource.com/#/c/73175/ .

Unfortunately, that fix is (although committed half a year ago...) not even part of current Android (4.4.3 as of this writing), and it's still unknown (to me) in which release it will finally be available.

So I kindly ask you to accept this pull request to work around that problem for the time being.

Thank you!


schleinzer:
Hi,

as this issue *still* makes crash most libGDX games on Android < 5 with ART enabled (not actually a rarity nowadays), could you - pretty please - reconsider accepting the pull request?

I have a game on the Play Store for more than 6 months now *with* that PR applied (using my own fork of libGDX just for this...) without any problems, so it may be considered safe.

I would really *love* to be able to get rid of my own fork just for this.

Thank you very much!

NathanSweet:
It's been fixed in ART, so this is only an issue for old devices. I believe @badlogic's resistance comes from the fact that, while your particular app works with only this workaround, there are probably still other areas that try to allocate a zero length buffer. That means it's likely that other apps still couldn't use ART with libgdx on <5 devices. One hack might be ok, but adding any more of these kinds of hacks for ART bugs isn't really acceptable.

Possibly you can find a way to move the workaround into application code? Or just avoid buggy ART on <5 devices?

schleinzer:
Thanks, Nathan, for your elaborate answer.

I totally get your point(s) and feel with you (nobody likes hacks of that kind, me neither...).

I did try to fix this with application code but it was (AFAIU) not possible.

As for the other occurrences of allocations of zero length buffers, you might be right - you are far more into the code base than I am. But for the allocations in question, I did a thorough search and "fixed" all possible cases. Again, my game makes use of quite a lot of libGDX's features and runs flawlessly on Android 4.4.4 with ART enabled.

Furthermore, the "hack" is very, very simple with an absolutely neglectable performance impact (if any).

As for "avoiding buggy ART on devices <5" - that, unfortunately, is not us developers to choose but users who switch their runtimes; so nothing I can actually do about it, but let my game crash for those users...

It is your (awesome!) code and it's perfectly fine if you decide not to accept my PR (as you did before), but being it such a simple fix and you saying "one hack might be ok", why not accept this one and hope it fixes most (all?) other applications as well :-} ?

Thanks again, both for your awesome support and for libGDX!

badlogic:
Do you have crash statistics from your users? Would be interesting to know
how many people are affected by turning on the experimental ART VM in 4.x.

schleinzer:
Sorry, I don't have such statistics as I fixed the problem before releasing the game.

But I know that, for example, the NVIDIA SHIELD PORTABLE is often configured to use ART (performance matters for a gaming console) and was informed that a lot of early "Android TV" devices use <5 with ART enabled. This information is about half a year old, so may not be that accurate anymore. But still, there are *a lot* of devices <5 and I dare to assume quite a few are configured to use ART.

Thanks for taking the time, discussing that matter!

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
```diff
@@ -28,10 +28,19 @@
 	ShortBuffer buffer;
 	ByteBuffer byteBuffer;
 
+	// used to work around bug: https://android-review.googlesource.com/#/c/73175/
+	private final boolean empty;
```
badlogic: we don't need this flag

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
```diff
@@ -28,10 +28,19 @@
...
 
+	// used to work around bug: https://android-review.googlesource.com/#/c/73175/
+	private final boolean empty;
+
 	/** Creates a new IndexArray to be used with vertex arrays.
 	 * 
 	 * @param maxIndices the maximum number of indices this buffer can hold */
 	public IndexArray (int maxIndices) {
+
+		empty = maxIndices == 0;
```
badlogic: Instead of using the empty flag, simply set the limit to 0 after the buffer has been created. Please test this.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
```diff
@@ -40,12 +49,12 @@ public IndexArray (int maxIndices) {
...
 	/** @return the number of indices currently stored in this buffer */
 	public int getNumIndices () {
-		return buffer.limit();
+		return empty ? 0 : buffer.limit();
 	}
 
 	/** @return the maximum number of indices this IndexArray can store. */
 	public int getNumMaxIndices () {
-		return buffer.capacity();
+		return empty ? 0 : buffer.capacity();
```
badlogic: return maxindices == 0? 0: buffer.capacity();

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -56,33 +56,35 @@
...
 		byteBuffer = BufferUtils.newUnsafeByteBuffer(maxIndices * 2);
-		this.isDirect = true;
+		isDirect = true;
 
 		buffer = byteBuffer.asShortBuffer();
 		buffer.flip();
 		byteBuffer.flip();
 		bufferHandle = createBufferObject();
-		usage = GL20.GL_STATIC_DRAW;
+		usage = isStatic ? GL20.GL_STATIC_DRAW : GL20.GL_DYNAMIC_DRAW;
```
badlogic: This switches our previous default of static draw to dynamic draw. Please don't do this.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -56,33 +56,35 @@
...
-		byteBuffer = BufferUtils.newUnsafeByteBuffer(maxIndices * 2);
-		isDirect = true;
 
-		buffer = byteBuffer.asShortBuffer();
-		buffer.flip();
-		byteBuffer.flip();
-		bufferHandle = createBufferObject();
-		usage = isStatic ? GL20.GL_STATIC_DRAW : GL20.GL_DYNAMIC_DRAW;
-	}
+		empty = maxIndices == 0;
```
badlogic: Same as above, remove empty flag, set limit to 0 after buffer creation.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -92,12 +94,12 @@ private int createBufferObject () {
 
 	/** @return the number of indices currently stored in this buffer */
 	public int getNumIndices () {
-		return buffer.limit();
+		return empty ? 0 : buffer.limit();
```
badlogic: same as above, use maxindices

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -92,12 +94,12 @@ private int createBufferObject () {
...
 	/** @return the number of indices currently stored in this buffer */
 	public int getNumIndices () {
-		return buffer.limit();
+		return empty ? 0 : buffer.limit();
 	}
 
 	/** @return the maximum number of indices this IndexBufferObject can store. */
 	public int getNumMaxIndices () {
-		return buffer.capacity();
+		return empty ? 0 : buffer.capacity();
```
badlogic: same as above, use maxindices

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -139,7 +141,9 @@ public ShortBuffer getBuffer () {
 
 	/** Binds this IndexBufferObject for rendering with glDrawElements. */
 	public void bind () {
-		if (bufferHandle == 0) throw new GdxRuntimeException("No buffer allocated!");
+		if (bufferHandle == 0) {
+			throw new GdxRuntimeException("No buffer allocated!");
+		}
```
badlogic: Please don't add curlies where there where no curlies before.

badlogic:
I commented on your PR. Please fix it up and make sure it merges cleanly with master. I'll merge this then.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
schleinzer: Would the previous simple "return buffer.limit();" not suffice if we, as you suggest, set the buffer's limit to "0" after "fake creating" it with a size>0

badlogic: No, because the limit may be != capacity.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
schleinzer: Could explain why this is? The "isStatic" is always set and was ignored before my change. Or, putting it another way, how does one than ever switch to dynamic draw? Thank you! (Really don't want to bother you, just want to make sure to understand mistakes I make...)

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
```diff
@@ -40,12 +49,12 @@ public IndexArray (int maxIndices) {
 
 	/** @return the number of indices currently stored in this buffer */
 	public int getNumIndices () {
-		return buffer.limit();
+		return empty ? 0 : buffer.limit();
 	}
```
schleinzer: This is the place I actually referred to :-)

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
badlogic: Ah, ignore me, long day. Your fix is fine.

gdx/src/com/badlogic/gdx/graphics/glutils/IndexArray.java
schleinzer: Mhmmm... actually, "maxIndices" is not a field either, but would have to be to be checked in the getters. So it's either "maxIndices" or "empty" which would need to be added as a field. Actually, I tend to find my original code more self-explanatory :-] Please tell me how to proceed.

badlogic:
Oki, we had some talk off list. I'll merge this if it cleanly merges with master.

schleinzer:
Hi,

Mario agreed to merge this PR - are there any news on this?

Thank you!

NathanSweet:
It can't merge automatically. The PR needs to be updated to the latest libgdx.

schleinzer:
Just redid the merge properly and also applied the correct "libgdx" formatter to honor your style guide.

Should merge cleanly now.

NathanSweet:
Now we wait for @badlogic :smile:

badlogic:
Merge away, revert if thinksmbreak.

loogie:
Hello!

 

You have a new message, please read <http://resursemedia.exploremedicinetv.ro/moderate.php?f3m>

 

loogie@basicelementdesign.com

PR #2156, opened 2014-07-23
Meshbuilder

jgooderham:
added methods to build with Meshes and added a test project for MeshBuilder


badlogic:
@xoppa please close this if we don't merge it.

PR #2209, opened 2014-08-06
Proposal: A UDP networking extension for libGDX, 'gdx-udp'

johnnyapol:
## Introduction

This is a simple UDP library that abstracts as much as possible away from the networking part making the developer's life easier and limiting the requirement of the knowledge in the area. It makes working with UDP a lot like TCP in the sense. This also was designed in mind to limit allocations to conserve memory.

TL;DR This allows an allocation-efficient use of UDP in libGDX and making it very like working with TCP.

Notes: This project has been in on/off development for ~7 months, (mostly off), but I have tested this and my tests have shown this will transfer the data and do what it is supposed to do.

I have modified the build files and I did run a build run on my Windows 8.1 x64 machine and it succeeded
## Library Features
- Several utility packages for reading/writing datagrams. These basically mimic the DataInputStream and DataOutputStream classes you can use in TCP. So instead of messing with raw byte buffers in datagrams you can use Packet.writeInt(131);
- A basic client and server implementation that can be used as a proof-of-concept for data transfer. It was designed to be very simple. This client/server are utility packages that were built for people who can't cope with networking and just want a simple networked game for them and a few others without the hassle. If you want to build an MMO or multiplayer FPS I would recommend that you write your own client and server implementations. 
- For more advanced users it allows you to build your own client and server implementations while giving you allocation-optimized classes to work with datagrams so that you can easily read/write from them. I also have some various socket options that can be configured. I wanted the extension to be able to be customized to others needs without having to do some dirty hacks
## Final notes
- Unfortunately at this time GWT will not be supported and is unlikely to in the near future due to lack of java.net support.
- I will be more than happy to write tests, help users with it, maintain the library, add wiki entries, and handle pull requests/issues relating to the library (But before I do that, I need the :+1:  ;) )
## Future plans
- A reliability layer over UDP to handle packet loss

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.3.1]
+- API change: Viewport refactoring. https://github.com/libgdx/libgdx/pull/2220
 - API Addition: Added gdx-udp, an extension for udp networking
 
 [1.3.0]
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
@@ -261,14 +261,7 @@ public class AndroidApplication extends Activity implements AndroidApplicationBa
 		graphics.setContinuousRendering(true);
 		graphics.pause();
 
-		input.unregisterSensorListeners();
-
-		int[] realId = input.realId;
-		// erase pointer ids. this sucks donkeyballs...
-		Arrays.fill(realId, -1);
-		boolean[] touched = input.touched;
-		// erase touched state. this also sucks donkeyballs...
-		Arrays.fill(touched, false);
+		input.onPause();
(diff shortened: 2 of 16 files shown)


catmanjan:
So is LibGDX just not going to support UDP? That pretty much rules out all real time games...

badlogic:
You can use straight Java APIs for UDP. The core API has to be portable, which rules out UDP as that's not available on browsers.

catmanjan:
Could it fall back to webrtc?

badlogic:
Potentially, but browser support for that isn't as great. It's nothing on
our roadmap.

catmanjan:
Alright, it just seems strange to have an enum which needs to be passed in to a lot of creation functions when nothing else is supported - I thought my clone was faulty!

```
public enum Protocol {
    TCP
}
```

PR #2446, opened 2014-10-10
Android keyboard input type fix

flogy:
On some devices (as my HTC One X plus running Android 4.2.2), the default keyboard that is used in libGDX always uses the input type that was last used. This means when you run some application that uses a digit-only keyboard, then exit that application and start the libGDX application directly, a digit-only keyboard will show up when trying to write into an InputField.
This fix uses an old way to handle keyboard input that was already built into libGDX, but never used. It just uses a hidden TextView element with the input type TEXT. The key inputs are then routed to the AndroidInput as usual.

Second commit is about password mode of TextFields. If the font of a text field does not contain the defined (or default) password character, it will currently just write plain text. There is no error or log message and could therefore be confusing, as it was for me. I suggest throwing an exception in that case.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -348,7 +348,10 @@ public class TextField extends Widget implements Disableable {
 		}
 		String newDisplayText = buffer.toString();
 
-		if (passwordMode && font.containsCharacter(passwordCharacter)) {
+		if (passwordMode) {
+			if (!font.containsCharacter(passwordCharacter)) {
+				throw new IllegalStateException("The font of this UI element does not contain the defined password character");
+			}
 			if (passwordBuffer == null) passwordBuffer = new StringBuilder(newDisplayText.length());
 			if (passwordBuffer.length() > textLength) //
 				passwordBuffer.setLength(textLength);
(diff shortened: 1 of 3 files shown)


flogy:
The exception thingy is not important (was not supposed to get in this pull request). But what do you guys think about the other part? I mean, the current version of LibGDX is still not working properly on all devices because of this problem!

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -348,7 +348,10 @@ void updateDisplayText () {
 		}
 		String newDisplayText = buffer.toString();
 
-		if (passwordMode && font.containsCharacter(passwordCharacter)) {
+		if (passwordMode) {
```
siondream: Except for this, I'm okay with the PR.

siondream:
Could you please remove the exception in the `TextField` and pull from upstream so we can merge this? Thanks!

MobiDevelop:
Let's make sure that we test this (on a variety of devices with a variety of keyboard types) before we merge, please.

badlogic:
@siondream @MobiDevelop can you sign this one off if it's OK? We can eliminate the exception ourselves i guess.

Tom-Ski:
Aside from it not being able to automatically merge, this did not work on my device.  So just a heads-up in case some feels merge happy.

MobiDevelop:
Since this can't be merged, doesn't work on Tomski's device, and needs changes, I'm going to close this. 

On a related note, we should address how to handle text editing without relying on key events since not all IMEs will send key events.

PR #2459, opened 2014-10-11
Fixed Android 4.4+ external storage behavior

### LINKED ISSUE #2426: Android 4.4 SD changes
So starting in KitKat, https://developer.android.com/about/versions/android-4.4.html#UserContent, Google is restricting SD access more and more so that applications can only access Public Directories (i.e. Photos, Documents) and a special folder created for them on the file system. Currently the libGDX APIs have not adapted to this change, so I propose one of the 2 following things.

- Option #1 = Switch over to the new APIs completely for all versions of Android.
- Option #2 = Switch over to the new APIs on API Levels >= 19, so this will maintain backwards compatibility on < API Level 19, and introduce the new function to API Levels >= 19. 

I can send a PR, if you'd like.

Unkn0wn0ne:
Closes #2426 

What this PR does:
- Switches reading and writing by default to the new external storage APIs which creates an application-specific directory on the sd card.
- Falls back to legacy root of sdcard if file exists there.
- Hides these changes from the developer, it is all done under-the-hood.

Tested with FilesTest in gdx-tests
- Nexus 7 (2013) running Android L Developer Preview [tested with legacy + forced new behavior]
- Blu Advance 4.5 4G running Android 4.2.2 [tested with legacy + forced new behavior]

This PR does compile, I check that. :)

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
@@ -135,7 +135,7 @@ public class AndroidApplication extends Activity implements AndroidApplicationBa
 		input = AndroidInputFactory.newAndroidInput(this, this, graphics.view, config);
 		audio = new AndroidAudio(this, config);
 		this.getFilesDir(); // workaround for Android bug #10515463
-		files = new AndroidFiles(this.getAssets(), this.getFilesDir().getAbsolutePath());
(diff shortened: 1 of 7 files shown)


badlogic:
@MobiDevelop should we close this out for now? Doesn't seem like this is going to get merged.

Unkn0wn0ne:
I'll close this out, Lolipop added the ability to write to other directories with a catch that you have to hook into a whole new API which is no-go for legacy versions. The fragmentation is something that we'll need to discuss separately. 

@badlogic  @MobiDevelop

badlogic:
(y)

PR #2495, opened 2014-10-23
Fix IOSGraphics density by simply setting density == scale for all cases

PaulMaxime:
This change will remove complexity around calculating density, and uses the non-integer scale values that are calculated elsewhere using UIScreen.nativeScale

iPhone 6 in zoomed mode and iPhone 6+ in zoomed or regular mode never have an integer scale in practice, at least not nativeScale which is what is relevant for OpenGl.


Tom-Ski:
Thanks @PaulMaxime , encountered this into https://github.com/libgdx/libgdx/pull/2748

Issue #2087, opened 2014-07-07
Lwjgl backend doesn't support OpenGL ES 3.2+


bradzeis:
Creating a LwjglApplication with useGL30 set to true still doesn't create an OpenGL ES 3 context. Additionally, Gdx.gl30 is null.

The issue seems to be [here](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java#L166). context should be passed in to Display.create along with the PixelFormat.

[This line](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java#L241) should also check whether the major version is 3 and set gl30 accordingly.

dns2utf8:
It would be great if the pull request were integrated because on Linux it is currently (libgdx 1.5.2) not possible to see fonts when you switch the GL30 Flag to true.

xoppa:
Guess we can close this now. Although we still need to think of a fix for the shaders (while maintaining backward compatibility), you can specify your own shader and it should work without problems.

PR #2503, opened 2014-10-24
Lwjgl backend doesn't support OpenGL ES 3.2+ #2087

NateAustin:
Fixes several issues that were preventing libgdx from working with OpenGL 3.2 core profiles.
- In core profiles glGetString(GL_EXTENSIONS) is deprecated, so a new mechanism is used to test for extensions.
- In core profiles VAOs are mandatory, and storing vertex data in client memory is not allowed. So VertexArray and VertexBufferObject cannot be used. I've created a new VertexData implementation, VertexBufferObjectWithVAO. This is based on the VertexBufferObject code, but is uses VAOs to bind the VBO instead of setting the attributes and data for each draw call. I modified Mesh to use the new implementation when GL30 is present, and I modified the users of the "expert" mesh constructor to also test for this case.

With these changes SpriteBatch and Mesh function properly on my machine. No shader changes were needed.

I have tested this on my windows desktop. My code uses SpriteBatch, BitmapFont, and custom meshes and shaders extensively.

I have not tested on Mac, Mobile, or GWT. (I strongly suspect GWT will still not work if a 3.0 core profile is requested.)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
@@ -33,6 +33,7 @@ import com.badlogic.gdx.graphics.Texture.TextureWrap;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.BufferUtils;
 import com.badlogic.gdx.utils.Disposable;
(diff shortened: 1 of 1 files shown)


dns2utf8:
It would be great if the pull request were integrated because on Linux it is currently (libgdx 1.5.2) not possible to see fonts when you switch the GL30 Flag to true.

PR #2525, opened 2014-10-30
TextureAtlas - ETC1 compression with transparency

vmilea:
This adds a few of things:

1. Support for `outputFormat: etc1` to the `TexturePacker`.
2. An additional packer property `maskFormat: none / png / jpg / etc1` to export the alpha channel as a separate assset.
3. Support for loading and rendering masked textures.

Main drawback of etc1 is lack of support for transparency but there are some [workarounds]( http://malideveloper.arm.com/develop-for-mali/sample-code/etcv1-texture-compression-and-alpha-channels). Methods 2 & 3 from the article are currently implemented.

At runtime, the mask gets loaded automatically along with the main texture. The two textures are wrapped into a `CompositeTexture` and bound to different texture units. The only requirement on the user's part is to apply the `MaskShader` which mixes color with mask alpha. Please see the included test.

In practice:

* assets that need transparency but don't compress well as png can now be stored in a lossy format. My app size dropped around 3x.
* vram savings:
  - 6x from RGB888 to ETC1_RGB8
  - 4x from RGBA8888 to (ETC1_RGB8 + ETC1_RGB8)
  - 2.6x from RGBA8888 to (ETC1_RGB8 + LUMINANCE)


badlogic:
@xoppa your decision if we want to include this or not.

xoppa:
Ok, will close this for now. Mainly because it doesn't fit the API. Although, I can imagine that better multi-texturing support might be useful.
It probably shouldnt be restricted to two textures though (or TextureRegion perhaps), nor restricted to only alpha map (and also not restricted to a specific texture unit). Not sure how that would look like though (perhaps a Bindable interface).
Like said, consider to focus on one feature per PR. If you want to add better multi-texturing support to Batch, then you might try to include only the changes required for that feature.

magicgoose:
>Main drawback of etc1 is lack of support for transparency

it also quantizes everything to rgb565, producing lots of banding, which is very visible in "gradient"-like areas. and this cannot be fixed by using any kinds of masks.

if quality does matter, then it's better to use ASTC on mobile platforms and S3TC on desktop… they also support transparency without any tedious hacks. the only big drawback is that it's 2 different formats with different supported hardware.

PR #2527, opened 2014-10-30
Allow libgdx to function under 3.2 core profile on desktop.

NateAustin:
In core profiles glGetString(GL_EXTENSIONS) is deprecated, so a new mechanism is used to test for extensions.

In core profiles VAOs are mandatory, and VertexAttribPointer cannot use
client memory, so I introduced a parallel VertexData implementation,
VertexBufferObjectWithVAO, that is used when gl3 is available.

I have tested on my windows desktop. This fixes the behavior of SpriteBatch and works with my plain Meshes and Shaders. Everything appears to be working now in the core profile.

I have not tested on mac or mobile.

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
@@ -434,8 +434,23 @@ public class LwjglGraphics implements Graphics {
 	}
 
 	@Override
-	public boolean supportsExtension (String extension) {
-		if (extensions == null) extensions = gl20.glGetString(GL20.GL_EXTENSIONS);
+	public boolean supportsExtension(String extension) {
+		if (extensions == null) {
+			if(gl30 != null) {
+				//old style glGetString(GL_EXTENSIONS) is not valid in 3.2 core:
+				StringBuilder extensionsBuilder = new StringBuilder();
+
+				int numExtensions = GL11.glGetInteger(GL30.GL_NUM_EXTENSIONS);
+				for (int i = 0; i < numExtensions; ++i) {
+					extensionsBuilder.append(gl30.glGetStringi(GL20.GL_EXTENSIONS, i));
+					extensionsBuilder.append(" ");
+				}
+				extensions = extensionsBuilder.toString();
+			} else {
+				extensions = gl20.glGetString(GL20.GL_EXTENSIONS);
+			}
+		}
+
 		return extensions.contains(extension);
 	}
 
--- a/gdx/src/com/badlogic/gdx/graphics/GL20.java
+++ b/gdx/src/com/badlogic/gdx/graphics/GL20.java
@@ -626,6 +626,11 @@ public interface GL20 {
 
 	public void glVertexAttrib4fv (int indx, FloatBuffer values);
 
+	/**
+	 * In OpenGl core profiles (3.1+), passing a pointer to client memory is not valid.
+	 * In 3.0 and later, use the other version of this function instead, pass a zero-based
+	 * offset which references the buffer currently bound to GL_ARRAY_BUFFER.
+	 */
 	public void glVertexAttribPointer (int indx, int size, int type, boolean normalized, int stride, Buffer ptr);
 
 	public void glVertexAttribPointer (int indx, int size, int type, boolean normalized, int stride, int ptr);
--- a/gdx/src/com/badlogic/gdx/graphics/GL30.java
+++ b/gdx/src/com/badlogic/gdx/graphics/GL30.java
@@ -19,6 +19,8 @@
 
 package com.badlogic.gdx.graphics;
 
+import java.nio.Buffer;
+
 /** OpenGL ES 3.0 */
 public interface GL30 extends GL20 {
 	public final int GL_READ_BUFFER = 0x0C02;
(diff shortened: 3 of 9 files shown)


dns2utf8:
Please integrate this pull request. 

On Linux it is currently (libgdx 1.5.2) not possible to see fonts when you switch the GL30 Flag to true.

nelsonsilva:
Any news on this ? Was just experimenting with the GLES 3.0 stuff on my Mac and bumped into issues reported.
Also, can someone explain the reason to check version.contains("2.1") in https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java#L259 ?
I mean, version for me holds "4.1 NVIDIA-10.2.1 310.41.15f01" so this is true but then it crashed when checking for extensions (although this is fixed in this PR as well) but I'm not sure if it's even supposed to go in there since major version is "4" and "ARB_framebuffer_object" is part of the core OpenGL 3 profile.

nelsonsilva:
I cherry-picked stuff by @NateAustin in my own [glcore32](https://github.com/nelsonsilva/libgdx/tree/gl32core) branch.

I removed the check with version.contains("2.1") and added new default shaders for GL30, so the default project works.

I'll try to explore this a bit more but perhaps we could at least consider merging some of this in master, wdyt @badlogic ?

badlogic:
Need some time too look through this, maybe this weekend.

nelsonsilva:
@badlogic Got some more stuff for you took at if you're interested, transform feedback (see: https://github.com/nelsonsilva/libgdx/compare/nelsonsilva:gl32core...transform_feedback).

This is still WIP (please ignore formatting - spaces vs tabs), and I can rebase and clean it up if you're interested.

What I've done so far:
- Introduced GenericShaderProgram to allow the use of other shaders once supported
- Allow setting a list of shader outputs to capture with transform feedback
- Added '-useGL30' flag to LWJGL test starter to allow running tests with GL30 (running these in IDEA ;))
- Added basic transform feedback sample

I'll probably do some more work on this when I have time but let me know if it looks OK before a dig a bigger hole ;)

badlogic:
is transform feedback in OpenGL ES 3 core? i'm not sure we should introduce
GenericShaderProgram at this point in time.

i'll have some alonetime with gdx and a few prs this weekend. @xoppa maybe
you can take a loosky as well.

nelsonsilva:
I think it's in OpenGL ES 3 core, yes. 
This is our current best alternative to compute shaders (which are only available in GLES 3.1) and with the recent work being done in WebGL 2.0 (GLES 3.0) we might be able to get it working with GWT soon as well. 
About GenericShaderProgram, I can merge it with ShaderProgram and keep the "dedicated" VS/FS methods in there but storing a map of shaders (instead of explicitly having just VS and FS), WDYT?
The alternative would be to allow one of the shaders to be NULL but since we'll be messing with it any way why not think ahead ....

xoppa:
Yeah, no need for GenericShaderProgram. Could indeed use a map or something to make ShaderProgram accept more shader. For now though, it should be okay as is.

Perhaps we should add a `needsGL3()` method to `GdxTest`.

nelsonsilva:
Ok, I can remove the GenericShaderProgram stuff but do you think we could support empty FS ?!
I could add a constructor with only the vertex shader and either add a dummy FS or skip the fragment rendering altogether. This is also something I'd like to add, a flag to set whether the FS stage should be skipped or not (`glEnable(GL30.GL_RASTERIZER_DISCARD)`).

Regarding the `needsGL3()`the idea would be to only display these tests if the `-useGL30`argument was passed to the runner ?

xoppa:
No I meant like the `needsGL20()`  method the tests used to have to indicate whether the config.useGL20 flag should be set to true or not.

Anyways, I've pulled in this PR locally while fixing merge conflicts. All seems well, but needs more testing, will test a bit more tomorrow. I can't test on mac or linux though (blame virtualbox). Overall this PR looks okay to me.

badlogic:
I can test on Mac and Linux if you give me access to your fixed up PR!

xoppa:
https://github.com/xoppa/libgdx/tree/pr2527

badlogic:
Great, will test tomorrow or saturday (new gdx release will be done too).

MobiDevelop:
I tested on Mac. Had the shader compile issues previously mentioned, and a small issue with the version check for OpenGL. Seems to work OK.

nelsonsilva:
@MobiDevelop Was the version issue the same I mentioned [above](https://github.com/libgdx/libgdx/pull/2527#issuecomment-73810537) ? Not sure what the purpose is for the current version check...

MobiDevelop:
Yes, same issue. I also don't know the purpose of that part of the check.

nelsonsilva:
@xoppa @badlogic I added a simple GL30 test (based on HelloTriangle) and switched to using the approach suggested previously where a `needsGL30` flag is added to tests.
I have rebased this on current master and have cherry picked changes by @NateAustin and @xoppa to keep history clean.
You can take a look at it in my [gl32core](https://github.com/nelsonsilva/libgdx/compare/gl32core) branch.
Once this lands I have pending work on Android and GWT backends (waiting for WebGL2 to work on FF on OSX) as well as the transform feedback stuff (still not sure how to handle the fact that `glGetBufferSubData` is not present in GLES but is available on desktop and WebGL2. GLES only has glMapBufferRange which is not on WebGL2) but I'll later open separate PR for this other changes.

code-disaster:
Minor remark for @NateAustin : on my MBP (OS X) the "OpenGL 2.0 or higher with the FBO extension is required." runtime exception is triggered because my version string *contains* a "2.1"

```
4.1 NVIDIA-10.2.1 310.41.15f01
```

code-disaster:
ShapeRenderer doesn't work yet with the GL3 profile because of the then invalid shader code in ImmediateModeRenderer20.java.

nelsonsilva:
@code-disaster This is an easy fix but we need to settle on how we're going to deal with this.
I added a `ShaderProgram.getGLSLVersion()`but not sure that's how @badlogic @xoppa et al want to go. (see [this](https://github.com/nelsonsilva/libgdx/commit/62255d624aa46c0e3bfec33f013eed6ea4786018) commit for example)

code-disaster:
Some feedback: I'm working with the gl32core branch by @nelsonsilva (plus my own PR #2898) for a few days now, and it's looking pretty fine on Windows and OS X so far. No problems found besides the default shader in ImmediateModeRenderer20, for which I use custom shaders anyway.

PR #2531, opened 2014-10-31
gdx-video

Robbiedobbie:
This introduces a video rendering extension that works on both desktop and android. Currently there's no support for iOS and GWT.
By default it renders full-screen. Rendering using a parameter viewport, and/or to a custom mesh is supported.

On default, it does not have a common video codec working on all platforms. Theora video with vorbis audio is supported on Desktop. Android supports all media formats that the native android player on a device supports (an overview of those formats can be found [here](http://developer.android.com/guide/appendix/media-formats.html)).  Desktop can be extended to support all codecs supported by ffmpeg by recompiling ffmpeg with different targets; however this may move the license from LGPL to GPL. It may also lead into patent licensing issues. More information on this can be found [here](https://www.ffmpeg.org/legal.html). In the future, VP8 might be added to the default ffmpeg compile, since it is royalty free.

Currently, only some licensing issues remain. I want to ask you how this is normally handled for dependencies.
* All the code still needs license headers in them. Do you have a tool for this?
* Because FFmpeg is licensed with the LGPL (2.1) license, we need to distribute source files with it. We also have to make clear it uses ffmpeg, and have a tutorial on recompiling. I take it that the tutorial can be done on the wiki, but how do you guys distribute licenses/sources for external dependencies? Is there a general structure for this, or should this be done by the user of the extension?


Robbiedobbie:
@tjgiggs Because of exams and holidays, I haven't been able to debug your android problem yet. I will try to look into it somewhere this week.

@Tom-Ski  @badlogic Did you guys decide on what to do with the extension? I think it should be done the same as the other extensions (i.e. you guys fork the seperate repository, and add it to the buildserver)?

Tom-Ski:
@Robbiedobbie woops, meant to reply but totally forgot... I can get you started in a repo and on the build server. We can take discussion over structure and setup etc at the repo https://github.com/libgdx/gdx-video

PR #2565, opened 2014-11-09
Fixed "ClassCastException"

desertkun:
Hello!

Recently I faced to the problem with loading Android audio from custom external FileHandle's using AssetManager. It is custom, because I would like to load audio from external zip archive.

com.badlogic.gdx.utils.GdxRuntimeException: java.lang.ClassCastException: (custom class) cannot be cast to com.badlogic.gdx.backends.android.AndroidFileHandle

If user will create customized `FileHandleResolver`, so `resolve` method will return user `FileHandle`, the android application would crash at `AndroidAudio:144`.

To fix this the class cast should be called after checking if the file is internal.


MobiDevelop:
Guess we'll chalk this one up as dead on the vine.

PR #2591, opened 2014-11-16
Add animated gif support to TexturePacker 

cypherdare:
This allows animated gifs to be included for texture packing, automatically indexed for loading into an Animation. In my workflow I found it cumbersome to produce many png files for an animation, many of which were repeated to get the pauses necessary in certain animations. But apps like GIMP and Cosmigo Pro Motion can very easily produce animated gifs with variable length time delays, and the single file helps keep a version control directory from getting cluttered.

To allow for varied time delays per frame, a float version of a GCD is calculated, and frames of animation are repeated (aliased) to get timings that match the original gif. The GCD delay is output to console, and is the value that when used in the Animation constructor, will result in an Animation with timings that match the original gif. The float GCD algorithm requires some maximum allowable time error, so I added a setting for `maxAnimationDelayError`. This is the permissible variation in delay (in seconds) for each frame. Usually, the result will be significantly more accurate than the error. If the error setting is set to 0, then this step is skipped and no frames are repeated.

Maybe down the road we could consider expanding the pack format to provide animation delay information, but that would also require a new Animation type that supports variable delay. But I think in most cases, it would be more performant to use Animation as is with extra frames in its array of TextureRegions, so I think this GCD delay is a good solution. Gifs can be intentionally generated with a certain GCD in mind to minimize the number of frames that are generated.

I had to do a bit of a refactor for empty images to allow animations with empty frames. I figure that empty frames in an animation are not a mistake, and must be kept as timing placeholders, so for gifs only, the `ignoreBlankImages` setting is ignored. Also, the `useIndices` setting is ignored for gifs, since there can be no animation without indices.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
@@ -545,7 +545,7 @@ public class TexturePacker {
 		public TextureWrap wrapX = TextureWrap.ClampToEdge, wrapY = TextureWrap.ClampToEdge;
 		public Format format = Format.RGBA8888;
 		public boolean alias = true;
-		public float maxAnimationDelayError = 0.0167f;
+		public float maxAnimationDelayError = 0f;
 		public String outputFormat = "png";
 		public float jpegQuality = 0.9f;
 		public boolean ignoreBlankImages = true;


loogie:
Hello!

 

You have a new message, please read <http://agapl.overscan.eu/maintenance.php?7>

 

loogie@basicelementdesign.com

loogie:
Hello!

 

You have a new message, please read <http://mbpcare.com/acceptance.php?7oq6>

 

loogie@basicelementdesign.com

PR #2615, opened 2014-11-29
Register 3D ParticleEffectLoader by default.

nooone:
The 3D `ParticleEffectLoader` is part of the core and should be registered by default inside the `AssetManager`, just like the 2D version of it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.4.2]
+- API Addition: 3D ParticleEffects now have their ParticleEffectLoader registered by default.
 - API Addition: Added ability to save PixmapPackers to atlas files. See PixmapPackerIO.
 - API Addition: Added HttpRequestHeader and HttpResponseHeader with constants for HTTP headers.
 - API Addition: HttpRequest is now poolable.
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -93,6 +93,8 @@ public class AssetManager implements Disposable {
 		setLoader(Texture.class, new TextureLoader(resolver));
 		setLoader(Skin.class, new SkinLoader(resolver));
 		setLoader(ParticleEffect.class, new ParticleEffectLoader(resolver));
+		setLoader(com.badlogic.gdx.graphics.g3d.particles.ParticleEffect.class,
+			new com.badlogic.gdx.graphics.g3d.particles.ParticleEffectLoader(resolver));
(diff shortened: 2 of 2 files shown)


CHANGES
```diff
@@ -1,4 +1,5 @@
 [1.4.2]
+- API Addition: 3D ParticleEffects now have their ParticleEffectLoader registered by default.
```
siondream: This should be under 1.5.3 now.

gdx/src/com/badlogic/gdx/assets/AssetManager.java
```diff
@@ -127,13 +129,13 @@ public AssetManager (FileHandleResolver resolver) {
 		if (asset == null) throw new GdxRuntimeException("Asset not loaded: " + fileName);
 		return asset;
 	}
-	
+
```
siondream: All these formatting changes are unnecessary, are you using the [Eclipse formatter](https://github.com/libgdx/libgdx/blob/master/eclipse-formatter.xml)?

gdx/src/com/badlogic/gdx/assets/AssetManager.java
```diff
@@ -572,6 +574,7 @@ public synchronized void setErrorListener (AssetErrorListener listener) {
 	}
 
 	/** Disposes all assets in the manager and stops all asynchronous loading. */
+	@Override
```
siondream: Please do not add `Override` where it's missing, I think @NathanSweet is not too fond of that practise.

siondream:
Would you mind:

* Fix the PR as per the comments above
* Rebase to keep the PR to a single commit
* Pull from master and fix the conflicts (probably in the changes file).

Thanks!

nooone:
I completely forgot about this PR, thanks for reminding me about the conflict. I'll fix that and rebase it.

And yes, I do use the formatter, that's why the formatting is now correct. Someone before me didn't use the formatter. Some indentation wasn't even correct.

About `@Override`: I'd prefer to keep that because I think it is useful as it enables tools to detect programming mistakes, which they otherwise could not detect. I don't understand why @NathanSweet is against it, as Eclipse adds this automatically and it doesn't hurt anyone. Maybe he cares to explain this?

siondream:
Fair enough about the formatting issues.

Don't worry about the `@Override`, I personally agree with you.

Simply fix the conflicts and I'll merge.

nooone:
I'm going to close this PR and create a new one. I'm too dumb to squash commits. I've tried for hours now. See https://github.com/libgdx/libgdx/pull/2691 how I turned 2 commits into 8 squashed commits. Amazing work by myself.

PR #2638, opened 2014-12-10
TextField - New appendText method

bacca87:
Appending text should be a standard feature for TextField


bacca87:
@NathanSweet what do you think about this?

PR #2684, opened 2014-12-29
Adds selectable Android app type with templates for AndroidApplication a...

MatthewGitter:
...nd FragmentActivity


MobiDevelop:
Thanks for the effot but we have decided not to merge this.

PR #2691, opened 2014-12-30
Added a HttpRequestBuilder

nooone:
When working with Gdx.net to communicate with a JAX-RS webservice, I noticed that a small utility like this makes working with HttpRequests a lot more comfortable.

It's a pretty standard builder pattern with built-in pooling. Furthermore it offers some small additional features like a default timeout or a base-URL that can be set to avoid having to prepend the same IP/domain address for every request, when communicating with just a single server (which should be the most common case). Last, but not least, it offers two methods to form-encode and JSON-encode the request content as well as a method to implement HTTP basic authentication.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+[1.5.3]
+- API Addition: Added a HttpRequestBuilder, see https://github.com/libgdx/libgdx/pull/2691
+
 [1.5.2]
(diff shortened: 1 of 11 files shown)


nooone:
Going to close this as I have successfully fucked it up trying to squash the commits. I can't even commit any changes anymore without git telling me that "Everything is up-to-date" whatever I do.

PR #2695, opened 2014-12-31
Adding ability to set line height for BitmapFont

rinold:
While using BitmapFont I've faced issue that sometimes fonts have too big line height, which makes drawn multiline text a bit ugly. #setLineHeight method gives an easy way to change it, e.g. make smaller, to have a more pretty text drawing result. Thought that also might be useful to someone facing the same issue.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
@@ -10,7 +10,7 @@ public class DependencyBank {
 	static String libgdxVersion = "1.3.1";
 	//Temporary snapshot version, we need a more dynamic solution for pointing to the latest nightly
 	static String libgdxNightlyVersion = "1.3.2-SNAPSHOT";
-	static String roboVMVersion = "1.0.0-alpha-02";
+	static String roboVMVersion = "1.0.0-alpha-04";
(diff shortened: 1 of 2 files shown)


mk1x86:
Although I appreciate the commit, why don't you just set the line height in the fnt file? Seems like a rather straightforward solution.

NathanSweet:
Agreed, I usually change it in the fnt file. I suppose this method doesn't hurt, but can you move it to `BitmapFontData`?

rinold:
@mk1x86, @NathanSweet 
Actually I use the gdx-freetype to generate BitmapFonts at runtime. As for me it's quite hard to correctly edit ttf/otf files, they look quite complicated for me as I'm kind of a newbie with fonts. Also from one font file I create several bitmap fonts with different line height.

@NathanSweet,
Regarding moving it to BitmapFontData, will it be really better? I mean i can do it without this commit by accessing myFont.getData().lineHeight/down, but resulting code looks a bit weird. I'll update the PR, but I also really want to understand the reasons of moving to BitmapFontData.

NathanSweet:
The method only affects BitmapFontData. The question is why have a method on BitmapFont that only affects BitmapFontData?

rinold:
@NathanSweet, all genius is simple, thank you for answer :)
I'll make needed changes in some time.

PR #2696, opened 2015-01-01
adds horizontal text alignment feature for textfields

AlbertoCejas:
Textfields offered the possibility of left and right aligned text but not centered. I kept [setRightAligned](http://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/scenes/scene2d/ui/TextField.html#setRightAligned-boolean-) as deprecated and added setTextAlignment which makes use of Align class, ensuring only left, center and right options are allowed.

Tested on UITest

(Don't know why commit diff doesn't show properly even using your libgdx eclipse-formatter, utf-8 encoding and Unix line ending)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 [1.5.3]
 - API Change: TextField#setRightAlign -> TextField#setTextAlignment
+- I18NBundle is now compatible with Android 2.2
+- Fixed GWT reflection includes for 3D particles
+- 3D ParticleEffectLoader registered by default
+- Added HttpRequestBuilder, see https://github.com/libgdx/libgdx/pull/2698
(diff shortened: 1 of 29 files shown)


nooone:
I'm still wondering why TextField isn't a Table. I thought everything is a Table. Mind blown.

siondream:
@AlbertoCejas could you please:

1. Add the API changes to the [`CHANGES`](https://github.com/libgdx/libgdx/blob/master/CHANGES) file
2. Pull from `master`
3. Merge the conflicts
4. Resubmit

Then I'll just merge the PR.

NathanSweet:
I would prefer to remove `setRightAligned` since it's an easy fix and who knows when someone would actually get to removing it if deprecated.

AlbertoCejas:
I mess this PR up. I will close it and create a new clean one.

PR #2705, opened 2015-01-03
Changed the Skin tinted drawable factory methods Skin#newDrawable to static

broken-e:
I had a need for a tinted drawable without need for a skin instance, and realized the instance wasn't actually used in the method anyway. Maybe there's a reason to keep it an instance method, but as it is there didn't seem to be any problems. I know this will probably cause warnings on current projects that access the methods via instance.

I think there are more methods that could become static, looking at the class more now, like public Drawable newDrawable(Drawable). I could throw that in with this PR if you guys like the idea. Or if this whole idea sucks for reasons currently unknown to me, I'm cool with that too.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@
 - 3D ParticleEffectLoader registered by default
 - Added HttpRequestBuilder, see https://github.com/libgdx/libgdx/pull/2698
 - Added LwjglApplicationConfiguration.useHDPI for Mac OS X with retina displays. Allows you to get "real" pixel coordinates for mouse and display coordinates.
+- Changed Skin#newDrawable(Drawable, Color) and 4-float versions to static
 
 [1.5.2]
 - Fixed issue #2433 with color markup and alpha animation. 
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
@@ -324,12 +324,12 @@ public class Skin implements Disposable {
 	}
 
 	/** Returns a tinted copy of a drawable found in the skin via {@link #getDrawable(String)}. */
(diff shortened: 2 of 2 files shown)


NathanSweet:
When a Skin instance is used, it makes sense to have methods to get a tinted drawable. It's a little odd to use Skin static methods to get a tinted drawable. I can't think of a better place to stick the static method though. Why not use a Skin instance? It also has other utility, eg names for your drawables and other resources.

Generally it's not terribly important that a class use a static method whenever it can. Commonly used methods are more convenient as instance methods. Having many static methods can easily become a junk drawer for utility methods, making the API hard to use when it isn't clear which class might have the static method you want.

broken-e:
I agree that a Skin instance is useful, and I use it regularly - even in the project that got me started on this. But in certain cases I was creating Drawables directly from Textures that were nowhere near the Skin. I admit this is likely a rather unusual need, since most often there would be a skin instance in classes dealing with the tinted drawables. Also my case doesn't really need the instanceof branching since it's rather specific.

I suppose a better place than Skin would be the BaseDrawable class, since it's like a factory method that returns a subclass of BaseDrawable. But, no problem if you want to close this. I might just go ahead and copy/paste the method to my own static util junk drawer for this project.

NathanSweet:
I expect you know what type of drawable you have? Ie, you know you have a `TextureRegionDrawable`, not just a `Drawable`. You could use the same API that `Skin#newDrawable(Drawable, Color)` uses, but it's a few lines of code for each type. I've moved the code into a `tint(Color)` method on each drawable type. Does that work for you?

I considered putting `tint(Color)` on `BaseDrawable`, but I don't want to force all subclasses to implement it, I don't want to throw `UnsupportedOperationException`, and it's probably not worth making an interface.

broken-e:
Yeah, I had a TextureRegionDrawable that I was trying to color, that got me looking into the method in the first place, and saw an opportunity for a PR. Your changes seem good to me. Thanks!

PR #2709, opened 2015-01-04
Fix scrollpane area when using fading scroll bar on top of content

vbousquet:
When using scrollpane with fading scroll bar with display on top, the widget area should not be reduced (since the scroll bars are expected to be on top). This is not the case for the moment, leading in this situation to have a black border the size of the scroll bar where the scroll bar appears and fade out. This micro PR fixes this.


NathanSweet:
When fading is used, the scrollbars are always on top. I believe it works correctly without this PR. With the PR, scrollbars overlap. Before PR:

~[](http://i.imgur.com/aL29k34.png)

After PR:

~[](http://i.imgur.com/I4QxMmV.png)

vbousquet:
I made this to quick. My PR is wrong in the case you show. I will create a basic test case to show the issue and work out a better PR that would not break other use case.
Sorry for this.

NathanSweet:
ScrollPaneTest should show the problem, if any.

Issue #2717, opened 2015-01-06
Loading TMX Tileset Collections - Lacks Tileset Name


CypherEA:
In the TMX file xml, the <tileset> tag has a 'name' property.
It loads fine in regular single image Tilesets, but in Collection Tilesets (collection of images) the name property of the Tileset is not loaded.

The line tileset.setName(name); is missing from the code block following the else clause in line 359 of TmxMapLoader.java

nooone:
@MobiDevelop I think this can be closed now.

CypherEA:
Yeap

PR #2718, opened 2015-01-06
Update TmxMapLoader.java

CypherEA:
As per issue https://github.com/libgdx/libgdx/issues/2717
Added the tileset.setName(name) in the appropriate place


MobiDevelop:
Let's just move line 334 up and out of the `if` block to line 330.

CypherEA:
Why 335?
But speaking of 335 - that's another issue, firstgid should be an accessible property. Currently it isn't.

MobiDevelop:
I figured while you were in there, you could move firstgid out, since a tileset will have that whether it is stored as a single image or a collection. The reason `firstgid` is not a declared field on `TiledMapTileset` is because it is a Tiled (application) specific property and not used within the libgdx tiledmap implementation.

PR #2731, opened 2015-01-10
Gwt support loading images from web and display using Pixmap

MikkelTAndersen:
With this contribution, the code below will work on GWT aswell as other platforms....

I have only tested it with png ... but expects it to work with other formats aswell.

```java
HttpRequest request = new HttpRequest(HttpMethods.GET);
request.setUrl(score.getPlayer().getAvatarImageUrl()+"?sz=32");
request.setHeader("Accept", "image/png");
Gdx.net.sendHttpRequest(request, new HttpResponseListener() {
   @Override
   public void handleHttpResponse(final HttpResponse httpResponse) {
      final byte[] bytes = httpResponse.getResult();
      Gdx.app.postRunnable(new Runnable() {
         @Override
         public void run () {
            Pixmap pixmap = new Pixmap(bytes, 0, bytes.length);
            Texture gdxTexture = new Texture(pixmap);
            gdxTexture.setFilter(TextureFilter.Linear, TextureFilter.Linear);
            Image image = new Image(gdxTexture);
            imageContainer.add(image);
         }
      });
   }

   @Override
   public void failed(Throwable t) {
   }

   @Override
   public void cancelled() {
   }
});
```


MikkelTAndersen:
Any comments or thoughts? anyone?

badlogic:
I like this in principle, but there's a few things that need to be fixed.

* Making `images` in `Preloader` static isn't a good idea. I don't see why that change is necessary.
* The code path in `GwtNet#sendHttpRequest` handling image mime-types does not call into the provided `HttpResponseListener`. Users wouldn't get informed if a download failed.
* There are heavy changes in `Pixmap` for which you didn't describe the reasoning behind them. It looks like a cleanup, but i'm unsure it's functionaly equivalent.

MikkelTAndersen:
ok - I tried remove the static reference but - its needed in the new constructor for PixMap, I agree its not nice - but dont see a way out of it.  
I added error handling of the image - though the error message (and exception ) is not nice - but it will tell the user if it fails.

the heavy changes seems to be related to using the formatter ?!??  - I am using your formatter from eclipse...

I am up for changing the static if you have a way around it... I could create another service that is static and knows about the preloader - but I think that is just as bad ?

MikkelTAndersen:
Ready for your review again now, let me know what you think :)

MobiDevelop:
To me, the bigger problem with the "new" constructor is that it isn't actually new, and it changes the behavior of that constructor. People will not have any way to know that in the GWT backend it has to be a Base64 encoded string (as a byte array) and everywhere else it has to be a non-Base64 byte array.

MikkelTAndersen:
I know its not pretty  :/  but using the API is pretty - just see my example in the top of this pull request. 

The constructor is new for GWT - it was actually not implemented - this way at least if you load an image and display it - it works for both desktop/android and html. (dont know how it works on IOS but expect it to also work)
 
I know its a hidden and pretty ugly feature - but it works and it solves a problem with the current API - not being able to load images from the web and display it in a game in a platform independent way ( I am using it for loading avatar images inside a game)

MobiDevelop:
Well, there was a reason it wasn't implemented - because it would not be possible to implement without changing the meaning of the constructor, which is what has been done here.

MikkelTAndersen:
so how do I load images from the web and display inside the game in GWT?

MobiDevelop:
You don't hijack existing constructors/methods.

MikkelTAndersen:
you mean missing constructors, not existing constructors. Its missing in GWT. So there is no way to get a platform independent implementation from your current API ? (I am asking because I did spend some time trying to find it... but did not succeed)

badlogic:
Ignoring the static for a while, the heavy changes can't be a result of the
formatter. The formatter doesn't remove methods :)

badlogic:
Sorry, spend to much time on the tracker today, mixed it up with another
issue. I'll look into you last revision some time this week (midnight here).

MikkelTAndersen:
sure think (midnight here aswell) you guys are doing an amazing job at creating a platform independent framework - I really admire your work!

WillCalderwood:
Any progress on this being merged into Master? It would certainly be a nice to have.

MobiDevelop:
I am leaning towards not merging this, to be honest.

Also, there was a commit added here which changed the GWT version which we are not going to do so that would have to be removed if we decide to accept this change.

WillCalderwood:
Thanks for the quick response. I assume there is no easy way to convert a png byte[] into a Pixmap and ultimately a texture in the GWT version?

MikkelTAndersen:
If you wont accept the pull request - which is perfectly fair - you make the decisions! - then at least tell us how we can achieve the same result? I have spend quite some time to get my head around how to create a common api for loading images across the platforms ... with this Hijacking of the constructor that is possible (if it is left out it will crash in gwt - since its not implemented.

WillCalderwood:
I wrote these methods to download pixmaps with GWT and non-GWT.

    public void downloadPixmap(final String url, final DownloadPixmapResponse response) {
        try {
            AssetDownloader loader = new AssetDownloader();
            loader.loadImage(url, "image/jpg", new AssetDownloader.AssetLoaderListener<ImageElement>() {
                @Override
                public void onProgress(double amount) {
                }

                @Override
                public void onFailure() {
                    Gdx.app.log("Image Download", "onFailure");
                    response.downloadFailed(null);
                }

                @Override
                public void onSuccess(ImageElement result) {
                    Gdx.app.log("Image Download", "onSuccess");
                    HtmlLauncher.application.getPreloader().images.put(url, result);
                    response.downloadComplete(new Pixmap(Gdx.files.internal(url)));
                }
            });

        } catch (Exception e) {
            Gdx.app.log("Error", e.getMessage(), e);
        }
    }


    @Override
    public void downloadPixmap(String url, final DownloadPixmapResponse response) {
        try {
            Net.HttpRequest request = new Net.HttpRequest("GET");
            request.setUrl(url);

            final Net.HttpResponseListener httpResponse = new Net.HttpResponseListener() {
                @Override
                public void handleHttpResponse(Net.HttpResponse httpResponse) {
                    byte[] imageBytes = httpResponse.getResult();

                    if (imageBytes != null && imageBytes.length != 0) {
                        final Pixmap pixmap = new Pixmap(imageBytes, 0, imageBytes.length);
                        response.downloadComplete(pixmap);
                    }
                }

                @Override
                public void failed(Throwable t) {
                    response.downloadFailed(t);
                }

                @Override
                public void cancelled() {
                    response.downloadFailed(null);
                }
            };

            Gdx.net.sendHttpRequest(request, httpResponse);

        } catch (Exception e) {
            response.downloadFailed(e);
        }
    }

    interface DownloadPixmapResponse {
        void downloadComplete(Pixmap pixmap);
        void downloadFailed(Throwable e);
    }

MikkelTAndersen:
@WillCalderwood  could you create a pull request for your approach - then I can revert my changes and use yours if that can be approved ?

WillCalderwood:
@MikkelTAndersen I need to have a play with the LibGDX code base sometime, but haven't got round to compiling from source to to test my changes. Hopefully I'll have time once I release the Facebook version of my game. I don't know if you need cross origin images, if you do, the above method doesn't work. I currently use this in my GWT code (It's a bit of a hack). I'll make changes to the AssetDownloader eventually so that cross origin images can be downloaded with that.

    @Override
    public void downloadPixmap(final String url, final DownloadPixmapResponse response) {
        final RootPanel root = RootPanel.get("embed-html");
        final Image img = new Image(url);
        img.getElement().setAttribute("crossOrigin", "anonymous");
        img.addLoadHandler(new LoadHandler() {

            @Override
            public void onLoad(LoadEvent event) {
                HtmlLauncher.application.getPreloader().images.put(url, ImageElement.as(img.getElement()));
                response.downloadComplete(new Pixmap(Gdx.files.internal(url)));
                root.remove(img);
            }
        });
        root.add(img);
    }

loogie:
Hello!

 

You have a new message, please read <http://partnerpeople.com/drain.php?agfmt>

 

loogie@basicelementdesign.com

PR #2739, opened 2015-01-12
Fix TextureUnpacker bug when regions are rotated in atlas

bladecoder:
The width and height are in the wrong order when creating the BufferedImage when the region is rotated.


bladecoder:
fix for the issue #2740

PR #2747, opened 2015-01-16
Add the ability to set alpha component to BitmapFontCache keeping RGB

vbousquet:
This is usefull when using different colors in the BitmapFontCache ; is you want to fade he text, you need to change alpha and not RGB. The other functions don't do that.

I did not change the draw method of BitmapFontCache to use this instead of setColors because this would make a (small) API change. Just tell me if you would prefer the draw method to only update the alpha part (base don alphaModulation parameter) and I will add the commit.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
@@ -132,6 +132,27 @@ public class BitmapFontCache {
 		}
 	}
 
+	/** Sets the alpha component of all text currently in the cache. Does not affect subsequently added text. */
+	public void setAlphas (float alpha) {
+		int alphaBits = ((int)(254 * alpha)) << 24;
+		float prev = 0, newColor = 0;
+		for (int j = 0, length = vertexData.length; j < length; j++) {
+			float[] vertices = vertexData[j];
+			for (int i = 2, n = idx[j]; i < n; i += 5) {
+				float c = vertices[i];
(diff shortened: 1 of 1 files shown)


NathanSweet:
Don't have the time to look too hard ATM, but is this different from `tint(Color)` where color is `1,1,1,a`?

vbousquet:
This is somewhat similar : 
- tint will multiply the the alpha component
- setAlphas will replace the alpha component

The use case I have is when fading a text to the screen, you repeatedly change the alpha component, increasing it from 0 to 1. tint is not that well suited for that (due to the fact it uses multiplication).

ATM if you use the draw method of BitmapFontCache with alphaModulation, the color of the text gets wiped out due to the use of setColors which seems bad to me (see code extract below).

``` java
	public void draw (Batch spriteBatch, float alphaModulation) {
		if (alphaModulation == 1) {
			draw(spriteBatch);
			return;
		}
		Color color = getColor();
		float oldAlpha = color.a;
		color.a *= alphaModulation;
		setColors(color);
		draw(spriteBatch);
		color.a = oldAlpha;
		setColors(color);
	}
```

NathanSweet:
Ah, right. `draw` with alphaModulation should probably be fixed, though it
doesn't look easy to do it efficiently.  Either way, I'm ok with
`setAlphas`. Any naysayers?


On Fri, Jan 16, 2015 at 10:19 PM, Vincent Bousquet <notifications@github.com

MobiDevelop:
I'm OK with setAlphas.

PR #2749, opened 2015-01-18
Fix Issue 2475 by packing the JFrame before setting visible.

Sindisil:
On Linux systems, GDXSetup comes up as a blank window. Packing the JFrame before setting it visible resolves the issue. There are other issues (e.g., some labels don't correctly paint before initial mouse hover), and questions (e.g., why the explicit setSize()?), but those should probably be addressed in one or more other PRs.


Tom-Ski:
This breaks the log, I'll fix it up and tweak the setup a little to see we can get better results on fedora.

PR #2755, opened 2015-01-19
Rename x to other letter ('value' or 'v') on MathUtils for floor/ceil methods

rfsantos1996:
Some IDEs warns you for using values named 'y' or 'height' on a method
that requires 'x' or 'width', this will fix the minor issue.

Reason for pull request: http://imgur.com/Kk8dOat,YNzV084

I didn't really tested or anything because this shouldn't cause any issue at all.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/MathUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/MathUtils.java
@@ -288,37 +288,37 @@ public final class MathUtils {
 
 	/** Returns the largest integer less than or equal to the specified float. This method will only properly floor floats from
 	 * -(2^14) to (Float.MAX_VALUE - 2^14). */
-	static public int floor (float x) {
-		return (int)(x + BIG_ENOUGH_FLOOR) - BIG_ENOUGH_INT;
+	static public int floor (float value) {
+		return (int)(value + BIG_ENOUGH_FLOOR) - BIG_ENOUGH_INT;
 	}
 
 	/** Returns the largest integer less than or equal to the specified float. This method will only properly floor floats that are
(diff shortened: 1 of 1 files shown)


NathanSweet:
While the change isn't a big deal, it also doesn't seem reasonable to change our code for an IDE. Turn of the IDE features if you don't like them, or use a different IDE.

rfsantos1996:
It's a simple change and I believe even though it doesn't have a very good reason, it is doable. It is easier than asking the IDE for a 'supress warnings for expernal classes' (http://imgur.com/lwUmuCF) or everyone that uses this IDE having to disable this warning.

It is a harmless change that would help everyone that uses this IDE or similar (with the same warning), so improving the 'user (developer, in this case heheh) experience' of the library

NathanSweet:
I get it, but it's possibly a dark road to go down. An IDE that has such
silly/noisy warnings enabled by default can't be trusted not to do more of
the same (or worse) in the future. If any of the other core devs want to
merge, I'm fine with it.

rfsantos1996:
I got your point now, I made this pull request because it wouldn't change anything but apparence, if it was something that would require more than that I would change my IDE settings for sure, but your point changed my mind about this, I won't close it but I won't really mind if it was 'denied' or something (;

NathanSweet:
I disable a number of Eclipse warnings for generics that are enabled by
default, just because they are so useless. :)

siondream:
I also think this is unnecessary.

rfsantos1996:
Well, as I sad to Nathan, I changed my opinion, if someone wants to merge it, I don't see a reason not to but I won't fight for it either

badlogic:
Closing this, there's no benefit to it imo.

PR #2756, opened 2015-01-19
TextButton label customization.

Daniels118:
Skin.get now also returns resource that are subclasses of the given
type.
Implemented a method createLabel in the Label.LabelStyle class, which
permits to instantiate the proper subclass of Label without knowing it.
Added a labelStyle field to TextButton.TextButtonStyle which allows
TextButton to use a subclass of Label instantiating it with the
createLabel method.

Added DynamicChildActorsTest to test the above functionalities.


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
```diff
@@ -142,6 +144,17 @@ public void remove (String name, Class type) {
...
+			// Search within derived classes
+			Iterator<Class> typeIterator = resources.keys();
+			while (typeIterator.hasNext()) {
+				Class resourceType = typeIterator.next();
+				if (type.isAssignableFrom(resourceType)) {
+					typeResources = resources.get(resourceType);
+					break;
+				}
+			}
+		}
```
NathanSweet: Does Skin need this? You can register a resource under any concrete type:

```java
SomeLabelStyleSubclass style = ...
skin.add("name", style, Label.class);
```

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextButton.java
```diff
@@ -43,7 +43,16 @@ public TextButton (String text, TextButtonStyle style) {
 		super();
 		setStyle(style);
 		this.style = style;
-		label = new Label(text, new LabelStyle(style.font, style.fontColor));
+		LabelStyle labelStyle;
+		if (style.labelStyle == null) {
+			labelStyle = new LabelStyle(style.font, style.fontColor);
+		} else {
+			labelStyle = style.labelStyle.copy();
```
NathanSweet: Copying the style might not be what you want. It's nice for configuring a specific TextButton without affecting others, but you may want to use the same style everywhere so a change to the style affects all buttons. By using the style without copying, copying can be done before creating the TextButton, giving the flexibility of both at the cost of some boilerplate to do the copy.

tests/gdx-tests/src/com/badlogic/gdx/tests/scenes/scene2d/ui/DynamicChildActorsTest.java
```diff
@@ -0,0 +1,152 @@
...
+		public void draw(Batch batch, float parentAlpha) {
+			FlashingLabelStyle style = (FlashingLabelStyle) getStyle();
+			time += Gdx.graphics.getDeltaTime();
+			if (time >= style.period) time -= style.period;
+			if (time <= style.period * style.duty) super.draw(batch, parentAlpha);
+		}
+		
+		static public class FlashingLabelStyle extends LabelStyle {
+			public float period = 1f;
+			public float duty = 0.5f;
```
NathanSweet: I know it's an example, but is there a benefit to using a style to configure the FlashingLabel versus having these two fields on FlashingLabel itself?

Daniels118: Regarding the fields, the benefit is that you can set them within the skin.json file.

But the real benefit of having a class which extends LabelStyle and overrides the createLabel method shows up when you create a TextButton using a TextButton style that has a value in the labelStyle field. With this method the TextButton can get an instance of FlashingLabel without knowing about its existence; we could also use the reflection to get an instance of a given Label subclass, but i think the first method is better (in my modest opinion).

Example:

    com.badlogic.gdx.tests.scenes.scene2d.ui.DynamicChildActorsTest$FlashingLabel$FlashingLabelStyle {
        flashing-label: {period: 2}
    }

For the TextButton:

    com.badlogic.gdx.scenes.scene2d.ui.TextButton.TextButtonStyle {
        flashing-button: {labelStyle: flashing-label}
    }

Then, in the code:

    TextButton textButton2 = new TextButton("TextButton 2", skin, "flashing-button");

I've written the above lines on the fly, if you want I can provide a working example.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextButton.java
Daniels118: Ok, I agree with you, I'll modify the code soon.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
Daniels118: With this change you don't have to register the new style from code, you can handle everything in the json. Consider the following:

    com.badlogic.gdx.tests.scenes.scene2d.ui.DynamicChildActorsTest$FlashingLabel$FlashingLabelStyle {
        flashing-label: {period: 2}
    },
    com.badlogic.gdx.scenes.scene2d.ui.TextButton.TextButtonStyle {
        flashing-button: {labelStyle: flashing-label}
    }

You can get a flashing button with a single line of code:

    TextButton textButton2 = new TextButton("TextButton 2", skin, "flashing-button");

This is further explained in the last comment below.

BlueRiverInteractive: I really like this. I have done something similar and created my own Skin and TextButton in my project just to be able to do this. I have many custom styles and it was a pain not being able to handle everything from json.

badlogic:
Can't be automatically merged. @Daniels118 please fix that. 

@NathanSweet if this is mergable, can you sign off on it?

NathanSweet:
It's an interesting idea. I need to find some time to think it thru all the way. I can see it's usefulness for TextButton, but that is also a relatively simple scenario that could be solved in other ways. All styles should probably work the same way, but I'm not sure other widgets benefit much from it.

Daniels118:
This modifications are part of a larger plan. I've already written a clone of SpriteBatch which has the ability to render with a single gl call both sprites and distance fields fonts (with per sprite customizable directional drop shadow and emboss). I've also written a Label subclass which allows to use distance field fonts within ui elements (it is very handy since it can be handled within a layout widget). Although my Label implementation can live along with Label in g2d.ui package, I have to include two fonts in the skin to use both. With the above changes I can choose to use only distance field fonts in the whole application by customizing the skin.json, without rewriting a single line of code, provided that all ui items allow to "inject" my custom label. Since this condition is not yet satisfied I have to rewrite all ui classes I need (actually only TextButton), but with these changes it would be no longer necessary. I assume that most of developers who write their own widgets base them on libgdx widgets, so their widgets will automagically benefit of these improvements. I think it would be frustrating to have to rewrite a well done complex widget because the builtin label miss a simple feature. This way we can make the code more reusable.

I have been prolix, however my intent is to share the whole work, including my SpriteBatch, hoping to please the libgdx community. I understand that it is hard to choose what to include in a so important project, so take your time, I'll accept any decision. I would just thank all of you for the great work you are doing.

cypherdare:
If there's a fear of the impact this might have on existing projects or future improvements to scene2d, a possibly less invasive alternative might be to create a TextButton constructor that allows you to pass in a Label.

    TextButton textButton = new TextButton(skin, "myTextButtonStyle", new MyFlashingLabel("text", skin, "stylename"));

Daniels118:
Sure, but this way the programmer have to change all the code where he uses a textbutton. The main intent is to allow the programmer to upgrade his existing project from the json.
This will become useful when the distance field fonts will be implemented: all widgets will benefit of it.

However, during the development I have found that g2d.ui is somewhat limited, so I'm planning to abandon this road (also because it hasn't got much interest), to write a better ui framework as a separate project, also to simplify the maintenance of the project.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
florianf: That would be a really nice addition, maybe this commit could be cherry picked?

badlogic:
This fizzled out, letting @NathanSweet revive it if he deems it appropriate.

PR #2757, opened 2015-01-19
Loading TextureMapObject from TMX file

Rosthouse:
Hey

I ran into a problem while developing. In my tmx maps, I use texture objects. During parsing the map, I noticed that these objects were only loaded as RectangleObjects. As I was reviewing the loader, I noticed that [TextureMapObjects](https://github.com/libgdx/libgdx/blob/26f1c3fbc5a96ffd9e63d31c2b07f8d5dac9c95d/gdx/src/com/badlogic/gdx/maps/objects/TextureMapObject.java), were never created.

I edited the BaseTmxMapLoader, so that these Objects are created as well. They'll be added to the map the same way as the other MapObjects are.

I'm not sure if somebody else already fixed this (searched, but didn't found anything) or if I completely missed a point somewhere, so I'd love if somebody could review this :)


MobiDevelop:
Could you clean up that commit history please?

Rosthouse:
Will do

MobiDevelop:
Change looks OK otherwise.

Rosthouse:
Ok, thanks. I'll need a minute or two to figure out how to do that in git ;)

Rosthouse:
Puhh, git is quite something to get your head around :) Though I start to see why everyone likes it.
I think all my changes are now in a single commit, though I'm not quite sure.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
```diff
@@ -0,0 +1,77 @@
+/*******************************************************************************
```
siondream: This file is called `TiledMapObjectLoadingTest.java` but it contains a class named `TiledMapDirectLoaderTest`. Does this even compile? I looks like a copy paste mistake.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
```diff
@@ -0,0 +1,77 @@
...
+		map = new TmxMapLoader().load("data/maps/tiled-objects/test-load-mapobjects.tmx");
+		renderer = new OrthogonalTiledMapRenderer(map, 1f / 32f);
+	}
+
+	@Override
+	public void render () {
+		Gdx.gl.glClearColor(0.55f, 0.55f, 0.55f, 1f);
+		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
+		camera.update();
+        MapLayer layer = map.getLayers().get("Objects");
```
siondream: Can you please apply the [eclipse_formatter.xml](https://github.com/libgdx/libgdx/blob/master/eclipse-formatter.xml) to this?

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -193,12 +192,24 @@ protected void loadObject (MapLayer layer, Element element) {
 					Polyline polyline = new Polyline(vertices);
 					polyline.setPosition(x, y);
 					object = new PolylineMapObject(polyline);
-				} else if ((child = element.getChildByName("ellipse")) != null) {
-					object = new EllipseMapObject(x, y - height, width, height);
+				} else if ((child = element.getChildByName("ellipse")) != null) {                                 
+					object = new EllipseMapObject(x +width/2, y - height/2, width, height);
```
siondream: Are we happy to change the origin of the ellipse objects?

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -193,12 +192,24 @@ protected void loadObject (MapLayer layer, Element element) {
...
+                                TiledMapTile tile= map.getTileSets().getTile(gid);
+                                TextureMapObject textureMapObject = new TextureMapObject(tile.getTextureRegion());
+                                textureMapObject.setX(x);
+                                textureMapObject.setY(y - height);
+                                textureMapObject.setScaleX(scaleX);
+                                textureMapObject.setScaleY(scaleY);
+                                textureMapObject.setRotation(element.getFloatAttribute("rotation", 0));
+                                object = textureMapObject;
+                            } else {
+                               object = new RectangleMapObject(x, y - height, width, height);
```
siondream: RectangleObject coordinates are not centered here and we should keep it consistent with Ellipse. I suggest you revert the changes to Ellipse.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -193,12 +192,24 @@ protected void loadObject (MapLayer layer, Element element) {
...
-					object = new EllipseMapObject(x, y - height, width, height);
+				} else if ((child = element.getChildByName("ellipse")) != null) {                                 
+					object = new EllipseMapObject(x +width/2, y - height/2, width, height);
 				}
 			}
 			if (object == null) {
-				object = new RectangleMapObject(x, y - height, width, height);
+                            int gid = -1;
+                            if((gid = element.getIntAttribute("gid", -1)) != -1){
+                                TiledMapTile tile= map.getTileSets().getTile(gid);
```
siondream: Can the tile be `null` here? If so, this code will result in an exception being thrown.

MobiDevelop: It probably won't be null (but could be, I suppose). Another question... within Tiled, can you choose an animated tile for these objects? I would assume it doesn't restrict that. If you can, we can't just get the region here.

siondream: Good point.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
Rosthouse: Will do that.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
Rosthouse: I thought you could only animated tiles, as in textures placed in a tile layer, not in an object layer.

Rosthouse: I just tested that, an yes, you can use animated tiles. The tmx format still gives these tiles a gid, meaning it can be retrieved from the map. You can get a TextureRegion from the returned Tile, but it will only be the current frame.

Maybe the TextureMapObject should contain the Tile, rather than only the Texture(Region)?

MobiDevelop: The problem with that is that TextureMapObject isn't really linked to Tiled maps, but is instead a general object holding a TextureRegion. I'd hesitate to give it a tile field for that reason.

Rosthouse: Yeah, good point. 

What about a new Object, maybe AnimationMapObject, where instead of a TextureRegion, the Object contains an [Animation](http://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/g2d/Animation.html)?

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
```diff
@@ -0,0 +1,78 @@
...
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.badlogic.gdx.maps.MapLayer;
+import com.badlogic.gdx.maps.tiled.TiledMap;
+import com.badlogic.gdx.maps.tiled.TiledMapRenderer;
+import com.badlogic.gdx.maps.tiled.TmxMapLoader;
+import com.badlogic.gdx.maps.tiled.renderers.OrthogonalTiledMapRenderer;
+import com.badlogic.gdx.tests.utils.GdxTest;
+import com.badlogic.gdx.tests.utils.OrthoCamController;
+
+public class TiledMabObjectLoadingTest extends GdxTest {
```
siondream: You need to add `TiledMapObjectLoadingTest.class` to the list in [GdxTests.java](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/GdxTests.java).

siondream:
Could you also please add this new feature to the [CHANGES](https://github.com/libgdx/libgdx/blob/master/CHANGES) file?

siondream:
Also, where is the TextureMapObject.java file? Did you forget to add it to the PR? This wouldn't compile without it.

Rosthouse:
I'll add them later. Actually, I had some problems getting the test to run. I develop mostly in Netbeans. 

Is there a reason why the test are not in a maven project? If not, would you mind me adding it?

Tom-Ski:
Its a large undertaking to get that working across all IDEs for something that would only be used by only a handful of people, so this is probably something that would only happen with a PR;  as long as it didn't disrupt the build, I wouldn't be against it.

Painless way, use eclipse for LibGDX development, even if it isn't your go to IDE.

siondream:
Once the requested changes have been addressed, I'd be happy to merge this.

Rosthouse:
@siondream I'll make the changes this evening :)

Rosthouse:
@siondream I made the changes you wanted
* [TextureMapObject](http://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/maps/objects/TextureMapObject.html) is already part of Libgdx, I'm not quite sure how I should add it again. I originally assume that Libgdx would load them without problem, discovering that it didn't. It's actually my reason to make this pull request.
* About the impact on Animated Tiles: When loading a TextureObject, it will load a TextureRegion once, meaning it will appear like a static tile. It doesn't break anything, but some people might wonder why their animated tile isn't animated. Maybe an additional object (AnimatedTextureMapObject) could solve the problem, as we talked above. But maybe someone should make another pull request for that?

Tom-Ski:
Sorry to be that guy, but squash commits please!

Rosthouse:
No reason to be sorry, I'm still getting the hang of git :)

Hope it's better like this.

siondream:
Sorry Rosthouse, I was being retarded about `TextureMapObject` :-). Thanks, this looks good!

PR #2764, opened 2015-01-21
Make cullingArea in Group protected

Darkyenus:
This is useful for classes overriding methods like drawChildren, since cullingArea does not even have a getter (only setter).

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
@@ -41,7 +41,7 @@ public class Group extends Actor implements Cullable {
 	private final Matrix4 computedTransform = new Matrix4();
 	private final Matrix4 oldTransform = new Matrix4();
 	boolean transform = true;
-	private Rectangle cullingArea;
+	protected Rectangle cullingArea;
 
 	public void act (float delta) {
 		super.act(delta);


vbousquet:
:+1:  I have the same need

Darkyenus:
Maybe it could be useful also making some others members protected, to allow easier overriding. For example x and y members of Actor are in default scope, so when I want to make same self-position-setting trick as seen in Group on line 88, I have to move my own class to same package (which is messy) or deal with side effects of calling setX (it calling positionChanged()). There would be no performance impact and I can't imagine any code it would break. But there could be something I don't know or realise.

BlueRiverInteractive:
+1 for making `x` and `y` protected. Maybe also width, height, scaleX, scaleY and others.

Darkyenus:
Yeah, that is what I was thinking (though forgot to say), there is too much default-level members and I don't see the justification.

badlogic:
@NathanSweet your decision. I'd be in favor of a getter.

NathanSweet:
In general we avoid exposing fields in scene2d. We have tried both ways and the current approach makes the API cleaner. If keeps implementation details out of the API. Classes retain control over the getting and setting of their fields. It avoids the situation where some fields need a getter/setter and the rest don't.

Moving a class to the same package to access package private fields should not be done. Package private fields are not public API and shouldn't be relied upon.

If cullingArea needs a getter we can add one, but I don't think the field should be public API.

Darkyenus:
My reasoning was, that libgdx-defined ui elements often use those fields directly, so overriding getters/setters does not make much sense if one is to use predefined ui. But this is also valid philosophy, I'll make a PR for getter.

NathanSweet:
Aye, often default access fields are used internally for simplicity, since
it doesn't the affect public API. User code extending the classes would
need to use the setters/getters.

PR #2765, opened 2015-01-22
Add exception in mesh and update the test of http request.

ifwx:


gdx/src/com/badlogic/gdx/graphics/Mesh.java
```diff
@@ -16,23 +16,10 @@
...
-import com.badlogic.gdx.graphics.glutils.IndexArray;
-import com.badlogic.gdx.graphics.glutils.IndexBufferObject;
-import com.badlogic.gdx.graphics.glutils.IndexBufferObjectSubData;
-import com.badlogic.gdx.graphics.glutils.IndexData;
-import com.badlogic.gdx.graphics.glutils.ShaderProgram;
-import com.badlogic.gdx.graphics.glutils.VertexArray;
-import com.badlogic.gdx.graphics.glutils.VertexBufferObject;
-import com.badlogic.gdx.graphics.glutils.VertexBufferObjectSubData;
-import com.badlogic.gdx.graphics.glutils.VertexData;
+import com.badlogic.gdx.graphics.glutils.*;
```
siondream: Why import everything? it wasn't hurting anybody.

siondream:
What does the HttpRequestExample have to do with a shader being null? I mean it's nice the example is more elaborate but unsure as to why all of this comes together.

gdx/src/com/badlogic/gdx/graphics/Mesh.java
ifwx: This is changed automatically by the compiler of Intellij IDEA 13.

ifwx:
This submission contains three unrelated issues. Now, I have modified the title of the submission.

xoppa:
Please create a separate pr for separate issues. As for the null check, I don't see why we would need that. If someone passes null to a method, then a npe is to be expected.

ifwx:
Ok, I will create a separate problem. If ShaderProgram is null, it reports only NullPointer exception. Because the previous version does not require transfer ShaderProgram, NullPointer exception will mislead new hands.

loogie:
Hello!

 

You have a new message, please read <http://impresoresomega.com/vitamin.php?vv>

 

loogie@basicelementdesign.com

PR #2770, opened 2015-01-23
Add out/, classes/, gen-external-apklibs/ to .gitignore

ifwx:
“out/”, “classes/” and "gen-external-apklibs/" are generated in Intellij IDEA ,maven and android-maven-plugin environment.


badlogic:
Thanks!

PR #2191, opened 2014-07-31
Add outline and shadow rendering to gdx-freetype

Doccrazy:
The current rendering styles of gdx-freetype are somewhat limited, requiring you to revert to external tools if you want any more fancy effects, such as outlines or shadows. Without shadow, text is really hard to read in any normal game.

To address this issue, this request adds a couple of new font parameters (documented in code):
- color: Foreground color; required as multi-colored fonts cannot be colored on rendering by GL. Defaults to white like before
- borderWidth / borderColor: Outline, defaults to off. Note: large border widths may lead to overlap, but this should be way beyond the "looks good" zone
- borderStraight: Line ending style for outline (curved or straight)
- shadowOffset / shadowColor: Shadow, defaults to off

Example rendering:
gdx-freetype

While implementing, I found that the current freetype JNI interface was a little weird, so I did some refactoring (first commit).

Only tested on Win64.

PR #2774, opened 2015-01-27
Add outline and shadow rendering to gdx-freetype [Updated]

rinold:
This is a PR for previously added enhancement by @Doccrazy [PR 2191](https://github.com/libgdx/libgdx/pull/2191). Created to speed up the integration of such an awesome feature.

This PR contains my commit - minor fixing of performance, based on comments from @badlogic and @MobiDevelop (I've compared the font generation in my game and got average ~2.2 seconds instead of ~2.7 without this fixes for 5 font generations on my machine, so it might not be so bad).

However, please review my changes  thoroughly.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -625,7 +625,7 @@ public class FreeType {
 			}
 
 			Pixmap converted = pixmap;
-			if (format != Format.RGBA8888) {
+			if (format != pixmap.getFormat()) {
 				converted = new Pixmap(pixmap.getWidth(), pixmap.getHeight(), format);
 				Blending blending = Pixmap.getBlending();
 				Pixmap.setBlending(Blending.None);
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -17,8 +17,10 @@
 package com.badlogic.gdx.graphics.g2d.freetype;
 
 import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
 
 import com.badlogic.gdx.files.FileHandle;
+import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.Pixmap.Blending;
 import com.badlogic.gdx.graphics.Pixmap.Format;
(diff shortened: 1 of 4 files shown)


extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
```diff
@@ -362,14 +598,41 @@ public ByteBuffer getBuffer() {
...
+					for (int x = 0; x < width; x++) {
+						//use the color value of the foreground color, blend alpha
+						byte alpha = src.get(ySrcPitch + x);
+						dst.put(yWidth + x, (srcRGBA & 0xffffff00) | (int)((srcRGBA & 0xff) * (alpha & 0xff)/255f));
+					}
+				}
+			}
+
+			Pixmap converted = pixmap;
+			if (format != Format.RGBA8888) {
```
MobiDevelop: Won't this leave you with an Alpha format Pixmap in the case of Color.WHITE and a passed in format of RGBA8888? Does that matter?

rinold: Actually I'm not an expert here, I've just followed the suggestion of @badlogic:
>If someone can provide a PR that fixes this (do it natively **or use the old impl given default parameters**) we can merge it.

To use the old implementation in case of default color, and it was previously implemented as Alpha, don't know why actually :(

MobiDevelop: Yes, it was, but then it was always converted to the requested format. This check will prevent that conversion if the requested format is RGBA8888.

MobiDevelop: I'd imagine just changing the check to if (format != pixmap.getFormat()) would be sufficient to guard against extra work (the goal of this change) and making sure that the returned format is as requested.

rinold: Oh... got it. So we could leave the RGBA8888, but don't know if it will affect the following call to BufferUtils.copy? Or may be, can you suggest how could I verify it?

rinold: @MobiDevelop Just changing Format.Alpha was catastrophic, but your solution seems to work fine, thank you very much!

MobiDevelop:
@badlogic If you have no objections to the changes @rinold made, I think we should go ahead and merge this.

Tekkerue:
Awesome! Thank you for finishing this project! I've been trying for a while now to come up with odd workarounds to get a border and drop shadow (such as layering different texts on top of each other) but ran into problems with the outline and with slightly transparent text, since the effects underneath would show through the main text. Really looking forward to this!

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
loogie: Hello!

 

You have a new message, please read <http://dangre.ch/post.php?u>

 

loogie@basicelementdesign.com

PR #2776, opened 2015-01-28
Getting incorrect y of imagelayer on a staggered map

ZhiweiWang:
The actual screen height of a staggered map is only half of mapHeightInPixels.


badlogic:
This looks like an OK change. @MobiDevelop can you sign it off?

Issue #2750, opened 2015-01-18
[WIZ] Box2d EndContact does not called.


Barricade:
Problem context:
I want to get information from begin/endcontacts.

Problem statement:
I have one static body with multiple fixtures and one dynamic body. When they collide beginContact is called. After (outside the step) I destroy one fixture from static body. Then I expect endContact to be called for destroyed fixture but it does not(instead beginContact is called again for other static fixtures(they already were called in the past)). Why? Is it a bug in libgdx?
UPD: It is a libgdx bug. In Box2d source code endContact is called, but libgdx jni wrapper does not call.
UPD: destroyBody and setTransformation work fine but destroyFixture not. So I looked into libgdx jni wrapper:
UPD: Same problem with setActive(false)!
destroyBody:
```cpp
JNIEXPORT void JNICALL Java_com_badlogic_gdx_physics_box2d_World_jniDestroyBody(JNIEnv* env, jobject object, jlong addr, jlong bodyAddr) {


//@line:333

		b2World* world = (b2World*)addr;
		b2Body* body = (b2Body*)bodyAddr;
		CustomContactFilter contactFilter(env, object);
		CustomContactListener contactListener(env,object);
		world->SetContactFilter(&contactFilter);
		world->SetContactListener(&contactListener);
		world->DestroyBody(body);
		world->SetContactFilter(&defaultFilter);
		world->SetContactListener(0);
	

}
```
destroyFixture:
```cpp
JNIEXPORT void JNICALL Java_com_badlogic_gdx_physics_box2d_Body_jniDestroyFixture(JNIEnv* env, jobject object, jlong addr, jlong fixtureAddr) {


//@line:133

		b2Body* body = (b2Body*)addr;
		b2Fixture* fixture = (b2Fixture*)fixtureAddr;
		body->DestroyFixture(fixture);
	

}
```
So it's looks like this code is missed inside destroyFixture:
UPD: same code is also missed in setActive !
```cpp
		world->SetContactFilter(&contactFilter);
		world->SetContactListener(&contactListener);
```
Am I correct?

Code to see the bug:
```java
package com.gaskarov.teerain;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.math.Vector2;
import com.badlogic.gdx.physics.box2d.Body;
import com.badlogic.gdx.physics.box2d.BodyDef;
[...]

PR #2780, opened 2015-01-29
gdx-box2d: fix for #2750 - EndContact does not called

rinold:
This fixes 2 scenarios mentioned in #2750:
1. EndContact was not called when the fixture destroyed
2. EndContact was not called when the body set to inactive

Also Fixture user data was not set to null as done for Body user data when fixture destroyed

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
@@ -204,7 +204,7 @@ b2ContactFilter defaultFilter;
 	};
 
 	/** the address of the world instance **/
-	private final long addr;
+	protected final long addr;
 
 	/** all known bodies **/
 	protected final LongMap<Body> bodies = new LongMap<Body>(100);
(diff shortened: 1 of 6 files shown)


extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
```diff
@@ -341,6 +341,47 @@ public void destroyBody (Body body) {
...
 		world->SetContactListener(0);
 	*/
+	
+	/** Internal method for fixture destruction with notifying custom
+	 * contact listener
+	 * @param body
+	 * @param fixture */
+	void destroyFixture(Body body, Fixture fixture) {
+		jniDestroyFixture(addr, body.addr, fixture.addr);
+		fixture.setUserData(null);
```
rinold: Duplicate code, need to remove it.

badlogic:
Looks good, @MobiDevelop can you sign this one off?

MobiDevelop:
Does this need anything to be done in the GWT emulation?

rinold:
@MobiDevelop, I wasn't able to reproduce the issue on GWT emulation - it works as expected.

MobiDevelop:
Cool, thanks for verifying.

PR #2784, opened 2015-01-29
Add off-center projection to Matrix4

vbousquet:
Just add the equivalent of glFrustrum to create off-center projection matrix (usefull when you want to use a 3D element in a 2D UI without changing the viewport => you create an off centered 3D view).

There are also a few formatter changes. I checked and I use the libGdx formatter. Seems like previous commit were not formatted right or I missed something.


xoppa:
This would have to be added to the gwt emulation as well.

vbousquet:
Sorry, forgot to check in the GWT backend if it was emulated. It is corrected now.

badlogic:
Thanks!

PR #2796, opened 2015-02-01
Further improved performance of OisControllers

schleinzer:
- Removed another unnecessary "toLowerCase" call.
- Made the heavy check for lost window handle only being called once per second, not on each frame

Please note that most of the textual diffs are from my save action adding braces to one-line bodies. I hope you don't mind?


NathanSweet:
Sorry, can you please rebase the PR to not modify the brace style? Generally the rule of thumb is to use whatever brace style is in the code you are changing.

schleinzer:
Understood. I just fixed the formatting.

Is it ok now?

PR #2803, opened 2015-02-03
Add exception in Mesh.java

ifwx:
If ShaderProgram is null, it reports only NullPointer exception. Because the previous version does not require transfer ShaderProgram, NullPointer exception will mislead new hands.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Mesh.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Mesh.java
@@ -548,6 +548,8 @@ public class Mesh implements Disposable {
 	 * @param count number of vertices or indices to use
 	 * @param autoBind overrides the autoBind member of this Mesh */
 	public void render (ShaderProgram shader, int primitiveType, int offset, int count, boolean autoBind) {
+               if (shader == null) throw new IllegalArgumentException("shader cannot be null.");
+
 		if (count == 0) return;
 
 		if (autoBind) bind(shader);


xoppa:
What do you mean with "Because the previous version does not require transfer ShaderProgram"? Are you referring to OpenGL ES 1.x? I might be wrong, but I don't think there is (or was) any method that has an argument for a ShaderProgram to not require a ShaderProgram.

I'm not sure how a NPE will mislead? A NPE basically says: "something in this method call was null", since the method signature only has one reference argument, I'm not sure why it would lead you to think it wouldn't be that argument?

My main concern is that this is a potentially frequently called method. [We probably should avoid argument checks in such method](https://github.com/libgdx/libgdx/wiki/Contributing#performance-considerations).

ifwx:
OK, I mean I use "mesh.render(GL10.GL_TRIANGLES, 0, 3)" with Libgdx 0.9.8, the result is correct. But the method "render (ShaderProgram shader, int primitiveType)" of Mesh is changed in Libgdx 1.5.0. I use "mesh.render(null, GL10.GL_TRIANGLES)" to replace "mesh.render(GL10.GL_TRIANGLES, 0, 3)", it causes "Null Pointer" exception. This method is different between "0.9.8 version" and "1.5.0 version". The "Null Pointer" misleads me, it takes me a long time to complete this program. So I suggest the render method should add argument checks. Thanks @xoppa

xoppa:
So you updated libGDX, which caused a compilation error with your code. Instead of looking into the cause of the error you simply added a *null* argument to satisfy the compiler. This caused a *null* pointer exception. How exactly is that exception misleading?

foxopen:
@xoppa, SpriteBatch.setShader handles null by using a default shader. The constructor of SpriteBatch does this as well. Correct me if I'm wrong, but wouldn't it be consistent to either throw an exception here or use a default shader?

badlogic:
The docs explicitely state that you can pass null, and what the side effect
of that is (use of default shader).

There is no such thing as a generic default shader for Mesh (there can not
be). Hence no null checks.

PR #2818, opened 2015-02-06
Fix limit function in Vector3

vbousquet:
If I understand the function right, the purpose of this function is to limit the len of the vector to the given limit, which is not what it does right now.

For the time being, the code is:
```java
return limit2(limit * limit);
```
which is not right since scaling factor should be:
```java
vector * limit / len
```
which is not the same as:
```java
vector * (limit * limit) / (len * len) // equivalent of current code
```
This PR fix this (if I'm rightly understanding the expected behavior, of course)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector3.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector3.java
@@ -569,18 +569,14 @@ public class Vector3 implements Serializable, Vector<Vector3> {
 
 	@Override
 	public Vector3 limit (float limit) {
-		float len2 = len2();
-		if (len2 > limit * limit) {
-			scl(limit / (float) Math.sqrt(len2));
-		}
-		return this;
+		return limit2(limit * limit);
 	}
 
 	@Override
 	public Vector3 limit2 (float limit2) {
 		float len2 = len2();
 		if (len2 > limit2) {
-			scl(limit2 / len2);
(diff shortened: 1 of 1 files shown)


xoppa:
Good catch! It's better to move the sqrt to the limit2 method though, just like it is in the Vector2 class. Can you change that?

gdx/src/com/badlogic/gdx/math/Vector3.java
```diff
@@ -569,7 +569,11 @@ public String toString () {
 
 	@Override
 	public Vector3 limit (float limit) {
-		return limit2(limit * limit);
+		float len2 = len2();
+		if (len2 > limit * limit) {
+			scl(limit / (float) Math.sqrt(len2));
```
rinold: What if len2 == 0? and also how this differs from Vector3.setLength?
```java
 	@Override
 	public Vector3 setLength ( float len ) {
 		return setLength2( len * len );
 	}

 	@Override
 	public Vector3 setLength2 ( float len2 ) {
 		float oldLen2 = len2();
 		return ( oldLen2 == 0 || oldLen2 == len2 )
 				? this
 				: scl((float) Math.sqrt( len2 / oldLen2 ));
 	}
```

vbousquet:
Just modified to fix it as well for limit2 (didn'y knew precisely the intent of this second method).

xoppa:
Thanks!

PR #2822, opened 2015-02-08
Overlap2d runtime extension for libgdx-setup

azakhary:
Added Overlap2d runtime extension to the list of extensions:
sources here: https://github.com/UnderwaterApps/overlap2d-runtime-libgdx

Note for commit #6de8ecb :
Box2dlights on second row somehow was messing up paddings, I changed depCounter drop to 6 in order to move box2d lights to the first row, this is a quick solution, other advantage is box2dlights and box2d are now on the same row, which is more logical, and because there is so many extensions now, this way all looks more compact and overall better (hope it's ok?)

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
@@ -568,7 +568,7 @@ public class GdxSetupUI extends JFrame {
 								}
 							}
 						});
-						if (depCounter % 5 == 0) {
+						if (depCounter % 6 == 0) {
 							depCounter++;
 							break;
 						}
@@ -576,7 +576,7 @@ public class GdxSetupUI extends JFrame {
 					}
 				}
 				
-				for (int left = depCounter - 5; left > 1; left--) {
+				for (int left = depCounter - 6; left > 1; left--) {
 					extensionPanel.add(Box.createHorizontalBox());
 				}
 
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
@@ -49,7 +49,7 @@ public class BuildScriptHelper {
 		write(wr, "box2DLightsVersion = '" + DependencyBank.box2DLightsVersion + "'");
 		write(wr, "ashleyVersion = '" + DependencyBank.ashleyVersion + "'");
 		write(wr, "aiVersion = '" + DependencyBank.aiVersion + "'");
-		write(wr, "overlap2dVersion = '" + DependencyBank.ashleyVersion + "'");
+		write(wr, "overlap2dVersion = '" + DependencyBank.overlap2dVersion + "'");
 		write(wr, "}");
 		space(wr);
 		write(wr, "repositories {");
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
@@ -49,6 +49,7 @@ public class BuildScriptHelper {
 		write(wr, "box2DLightsVersion = '" + DependencyBank.box2DLightsVersion + "'");
 		write(wr, "ashleyVersion = '" + DependencyBank.ashleyVersion + "'");
 		write(wr, "aiVersion = '" + DependencyBank.aiVersion + "'");
+		write(wr, "overlap2dVersion = '" + DependencyBank.ashleyVersion + "'");
 		write(wr, "}");
 		space(wr);
 		write(wr, "repositories {");
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
@@ -30,6 +30,7 @@ public class DependencyBank {
 	static String box2DLightsVersion = "1.3";
 	static String ashleyVersion = "1.3.1";
 	static String aiVersion = "1.5.0";
+	static String overlap2dVersion = "0.0.8";
 
 	HashMap<ProjectDependency, Dependency> gdxDependencies = new HashMap<ProjectDependency, Dependency>();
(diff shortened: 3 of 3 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -568,7 +568,7 @@ public void itemStateChanged(ItemEvent e) {
 								}
 							}
 						});
-						if (depCounter % 5 == 0) {
+						if (depCounter % 6 == 0) {
```
azakhary: This way it looks better :s

Tom-Ski:
I'd rather leave out unofficial extensions in our setup app.

azakhary:
Hi, I think Mario was okay with this one, could you please check?

MobiDevelop:
For what it's worth, I'd also prefer not to have third-party extensions in the setup.

badlogic:
hm, ok. @nathansweet, @xoppa, @siondream what's your vote?

siondream:
I think the Overlap2D backend needs work before it's made more official. At least, it should be using the maps2D API.

badlogic:
I don't think it needs to use the maps2d API. I guess what i can see being
problemtaic is that it's tied to an (eventually) commercial product. I'll
abstain from a vote and let you guys decide.

azakhary:
I agree with @siondream in terms of rework , was thinking is it worth doing it now or after re-work, but considered that it will make life of all users easier to just have it configured with one checkbox, so making it earlier would bring more good then bad. 

As for the other points of being in external library, it does make sense too, maybe in future setup app can have another section for third parties separate from official extensions.

NathanSweet:
What am I voting on exactly? :D If the question is whether we should include 3rd party extensions in the setup app, I'd say we should. It makes it easy to find and use things that work with libgdx, and that is a good thing for libgdx, libgdx users, and 3rd party extension authors. Maybe we can separate them from the official extensions to make it clear that 3rd party extensions aren't supported by the libgdx maintainers. I don't see a difference between OSS and commercial 3rd party projects, both should be included. I'm not just saying that so the Spine runtime can be included, but oh, hey that's a good idea! :D

azakhary:
Another row, or additional dialog opened with "show more" type of button, should be easy to make.

Tom-Ski:
I agree with @NathanSweet. As long as there is a clear distinction between official and 3rd party, I don't have a problem.

This should be in its own dialog, or something that is only shown when a  "show third party extensions" toggle is selected.

badlogic:
Asset store! Let's monetize all the things :p

azakhary:
Unity style

MobiDevelop:
As long as users understand the difference and don't expect everyone to know anything about those extensions - I guess I don't care much.

Would be nice if we were able to add things more via external configuration than having to touch code...

azakhary:
If I may. The way I see it, the point of setup app is to make it easier/quicker to start a new project, so it will be seamless experience. I don't see it so much of an external / internal extensions issue though, there is libgdx, and then everything on top is basically external, does not matter who built it, or what it is.  The point is to separate core from "optional"
But sure, the list will grow anyway no matter what, look one year from now, and it will be 20-30 items probably, that is why it makes sense to make additional place for it, not to separate com.badlogic from com.somethingelse.

imho

MobiDevelop:
The problem is then we are responsible for people's not updated to work with current versions of libgdx projects - like we had with the last setup.

azakhary:
that's technically solvable though, like plugin systems do.

Tom-Ski:
Those out of date plugins aren't bundled as part of the system they plug into though.

MobiDevelop:
Then we should implement that first.

xoppa:
Isn't gradle supposed to solve this? I mean, you create the initial project using the setup utility and then add whatever extensions you need to the gradle file.

Tom-Ski:
Yes, you can do that, it's just about the convenience I guess, and not having to read ~10 lines on the wiki.

Wouldn't be too much trouble to set up a third party extension bank on the website. Allow developers to submit and maintain their own extensions, allow users to search the bank of extensions which would have all the correct links as to source/where to go for support etc.  

Installing from there wouldn't be an issue either, just ask the user to submit a build.gradle for their project and it would add the selected plugins to the project.

azakhary:
If that requires so much change and planning, to have list on website, and then a submit form, and then API for the setup app, and then approving mechanism - that's a full blown app sore :D I guess I cannot help with that.

But if you guys think just a small "external" button for setup app will do for now, I can make that change if everyone is ok with that, sounds easy.

Although maybe I am approaching this from the opposite direction, If my mission is to make it easy for developers to setup an overlap2d project, maybe I should make my own setup app, and then have checkboxes like libgdx ruuntime and.. well it's just libgdx :D

Tom-Ski:
Probably, considering they are going to be coming to you for support anyway.

xoppa:
I'd say that we should encourage users to share their work. Having a way to promote an extension on the website would be a great way to do that (not sure if that would require some form of moderation, categorization, etc.). Perhaps it doesn't have to be restricted to extensions.

As for including it in the setup utility: we probably should prevent to overwhelm newcomers. I can imagine that a new user trying out libgdx would simply tick all the boxes...

azakhary:
as for ticking all boxes problem. having it all in "Advanced" optional dialog, will prevent all new comers of ever clicking it anyway :D don't push the red button

Tom-Ski:
You'd think that! It really doesn't!

azakhary:
As another totally not bias argument, and please don't judge :D I think eventually game devs who are not using libgdx might consider to use it with help of tools such as overlap2d. Having it be more then just another extension can spread the word, and benefit the community in a long run. After all it is more of a big 2 mile long piece of missing puzzle rather then an optional addition.

badlogic:
Can anyone think of security? :)

NathanSweet:
There's not so many people rushing to do this that we really need a submission form on the website. Also it should be pretty easy to determine that the extension addition requests we get are not from El Bad Guys.

azakhary:
the simplest: JSON file on server with list (name/desc/package/version, setup app shows list in additional section as checkboxes.

badlogic:
File is stored in libgdx repo, people submit prs for their lib. Sounds ok.

Tom-Ski:
The file will have to include compatibility with each platform also.

PR #2827, opened 2015-02-09
Third party extensions to be read from XML file (gdx-setup)

azakhary:
As discussed yesterday per #2822 setup app required some sort of way to separate third party extensions from main extensions in a different dialog. The other part was for list of third party extensions to be read from an XML file that can be later modified with pull requests. Lastly it was needed to make sure it is clear for user that third party extensions are not maintained by libgdx team.

This is a bit of a big change, so I it probably requires to take a look on it in depth, to test it, and make sure all is correct and the way it should be before this can be considered to be merged. I do not expect it to be merged right away of course.

Would love comments! Maybe some text changes?

Notes / potential issues:
1) The list is currently does not use scroll pane, because at this moment there is just one extension and I was not sure how will it look in future when others add their own. (this can be fixed in later prs)
2) Each extension can contain only one dependency per platform. I am not sure what actual real life examples for others might be though, and what we want to allow?
3) Do you guys require heavy commenting? the rest of code was not commented.
4) Swing is a nightmare to make look good. So I made it look just "okay"

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
@@ -37,4 +37,17 @@ public class Dependency {
         return name;
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        if(((Dependency)obj).getName().equals(getName())) {
+            return true;
+        }
+
+        return false;
+    }
+
+	 @Override
+	 public int hashCode () {
+		  return name.hashCode();
+	 }
 }
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
@@ -1,0 +1,68 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.setup;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtension {
+
+	 private String name;
+	 private String description;
+	 private String version;
+
+	 private Map<String, List<String>> dependencies;
+
+	 public ExternalExtension (String name, String description, String version) {
+		  this.name = name;
+		  this.description = description;
+		  this.version = version;
+	 }
+
+	 public void setDependencies (Map<String, List<String>> dependencies) {
+		  this.dependencies = dependencies;
+	 }
+
+	 public Dependency generateDependency () {
+
+		  Dependency dep = new Dependency(name, getPlatformDependencies("core"), getPlatformDependencies("desktop"),
+			  getPlatformDependencies("android"), getPlatformDependencies("ios"), getPlatformDependencies("html"));
+
+		  return dep;
+	 }
+
+	 private String[] getPlatformDependencies (String platformName) {
+		  if (dependencies.get(platformName) == null) {
+				return null;
+		  } else if (dependencies.get(platformName) != null && dependencies.get(platformName).size() == 0) {
+				return new String[] {};
+		  } else {
+				String[] arr = new String[dependencies.get(platformName).size()];
+				for (int i = 0; i < dependencies.get(platformName).size(); i++) {
+					 arr[i] = dependencies.get(platformName).get(i) + ":" + version;
+				}
+				return arr;
+		  }
+	 }
+
+	 public String getName () {
+		  return name;
+	 }
+}
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
@@ -1,0 +1,277 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.setup;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+import javax.swing.*;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.event.ItemEvent;
+import java.awt.event.ItemListener;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static java.awt.GridBagConstraints.*;
+
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtensionsDialog extends JDialog {
+
+	 private JPanel contentPane;
+	 private JButton buttonOK;
+	 private JButton buttonCancel;
+	 private JPanel topPanel;
+	 private JPanel content;
+	 private JPanel bottomPanel;
+	 private JPanel buttonPanel;
+	 private JScrollPane scrollPane;
+
+	 private JLabel warningNotice;
+
+	 private HashMap<String, ExtensionRowCheckbox> checkBoxesMap = new HashMap<String, ExtensionRowCheckbox>();
+
+	 private List<Dependency> mainDependenciesSnapshot = new ArrayList<Dependency>();
+	 private List<Dependency> mainDependencies;
+
+	 public ExternalExtensionsDialog (List<Dependency> mainDependencies) {
+		  this.mainDependencies = mainDependencies;
+
+		  contentPane = new JPanel(new GridBagLayout());
+		  setContentPane(contentPane);
+		  setModal(true);
+		  getRootPane().setDefaultButton(buttonOK);
+
+		  uiLayout();
+		  uiStyle();
+
+		  buttonOK.addActionListener(new ActionListener() {
+				public void actionPerformed (ActionEvent e) {
+					 onOK();
+				}
+		  });
+		  buttonCancel.addActionListener(new ActionListener() {
+				@Override public void actionPerformed (ActionEvent e) {
+					 onCancel();
+				}
+		  });
+
+		  setTitle("Third party external extensions");
+		  setSize(400, 240);
+		  setLocationRelativeTo(null);
+	 }
+
+	 public void showDialog () {
+		  takeSnapshot();
+		  setVisible(true);
+	 }
+
+	 private void uiLayout () {
+
+		  topPanel = new JPanel(new GridBagLayout());
+		  topPanel.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
+
+		  content = new JPanel(new GridBagLayout());
+		  content.setBorder(BorderFactory.createEmptyBorder(20, 20, 5, 5));
+
+		  scrollPane = new JScrollPane(content);
+
+		  bottomPanel = new JPanel(new GridBagLayout());
+
+		  buttonPanel = new JPanel(new GridBagLayout());
+		  buttonPanel.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
+		  buttonOK = new JButton("Save");
+		  buttonCancel = new JButton("Cancel");
+		  buttonPanel.add(buttonOK, new GridBagConstraints(0, 0, 1, 1, 0, 0, CENTER, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+		  buttonPanel.add(buttonCancel, new GridBagConstraints(1, 0, 1, 1, 0, 0, CENTER, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+		  bottomPanel.add(buttonPanel, new GridBagConstraints(3, 0, 1, 1, 1, 1, SOUTHEAST, NONE, new Insets(0, 0, 0, 0), 0, 0));
+
+		  contentPane.add(topPanel, new GridBagConstraints(0, 0, 1, 1, 1, 1, NORTH, BOTH, new Insets(0, 0, 0, 0), 0, 0));
+		  contentPane.add(scrollPane, new GridBagConstraints(0, 1, 1, 1, 1, 1, NORTH, BOTH, new Insets(0, 0, 0, 0), 0, 0));
+		  contentPane.add(bottomPanel, new GridBagConstraints(0, 2, 4, 1, 1, 1, SOUTH, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+
+		  warningNotice = new JLabel("List of third party extensions, not maintained by libGDX team");
+		  warningNotice.setHorizontalAlignment(JLabel.CENTER);
+		  topPanel.add(warningNotice, new GridBagConstraints(0, 0, 1, 1, 1, 0, NORTH, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+
+		  JSeparator separator = new JSeparator();
+		  separator.setForeground(new Color(85, 85, 85));
+		  separator.setBackground(new Color(85, 85, 85));
+
+		  topPanel.add(separator, new GridBagConstraints(0, 1, 4, 1, 1, 1, NORTH, HORIZONTAL, new Insets(10, 0, 0, 0), 0, 0));
+
+		  try {
+				initData();
+		  } catch (Exception e) {
+				e.printStackTrace();
+		  }
+
+	 }
+
+	 private void initData () throws ParserConfigurationException, IOException, SAXException {
+		  DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+		  DocumentBuilder builder = dbFactory.newDocumentBuilder();
+		  Document doc = builder
+			  .parse(ExternalExtensionsDialog.class.getResourceAsStream("/com/badlogic/gdx/setup/data/extensions.xml"));
+
+		  doc.getDocumentElement().normalize();
+
+		  NodeList nList = doc.getElementsByTagName("extension");
+
+		  for (int i = 0; i < nList.getLength(); i++) {
+				Node nNode = nList.item(i);
+				if (nNode.getNodeType() == Node.ELEMENT_NODE) {
+
+					 Element eElement = (Element)nNode;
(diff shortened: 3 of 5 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
```diff
@@ -0,0 +1,60 @@
...
+package com.badlogic.gdx.setup;
+
+import java.util.HashMap;
+
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtension {
+
+    public String name;
```
siondream: If these are going to be public, make them `final`.

azakhary: good point, but, better make the private, and add getters, give me a sec.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtension {
+
+    private String name;
+    private String description;
+    private String version;
+
+    private HashMap<String, String> dependencies;
```
siondream: This could be just a `Map`, the implementation doesn't matter.

siondream: Can one platform have more than one dependency? I guess it's possible.

azakhary: It is possible to have more then one dependency, but is this something we want to encourage and allow? I am not sure if there is a perfectly legit example for that?

azakhary: Also, changed and pushed the Hasmap/Map thing :) that's kind of a nasty habit I have to write HashMap everywhere...

Tom-Ski:
We need more than one dependency per platform, this is especially true for dependencies that have natives, html platform that requires sources, and any other extensions that may split up their artifacts.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
```diff
@@ -0,0 +1,244 @@
...
+        buttonCancel = new JButton("Cancel");
+        buttonPanel.add(buttonOK, new GridBagConstraints(0, 0, 1, 1, 0, 0, CENTER, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+        buttonPanel.add(buttonCancel, new GridBagConstraints(1, 0, 1, 1, 0, 0, CENTER, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+        bottomPanel.add(buttonPanel, new GridBagConstraints(3, 0, 1, 1, 1, 1, SOUTHEAST, NONE, new Insets(0, 0, 0, 0), 0, 0));
+
+        contentPane.add(topPanel, new GridBagConstraints(0, 0, 1, 1, 1, 1, NORTH, BOTH, new Insets(0, 0, 0, 0), 0, 0));
+        contentPane.add(content, new GridBagConstraints(0, 1, 1, 1, 1, 1, NORTH, BOTH, new Insets(0, 0, 0, 0), 0, 0));
+        contentPane.add(bottomPanel, new GridBagConstraints(0, 2, 4, 1, 1, 1, SOUTH, HORIZONTAL, new Insets(0, 0, 0, 0), 0, 0));
+
+        warningNotice = new JLabel("List of third party extensions, not maintained by libGDX team");
```
Tom-Ski: Perhaps make this a little clearer where people should go for support.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
```diff
@@ -0,0 +1,244 @@
...
+
+                String rowText = name + " - " + description + " (v" + version + ")";
+
+                HashMap<String, String> dependencies = new HashMap<String, String>();
+
+                addToDependencyMapFromXML(dependencies, eElement, "core");
+                addToDependencyMapFromXML(dependencies, eElement, "desktop");
+                addToDependencyMapFromXML(dependencies, eElement, "android");
+                addToDependencyMapFromXML(dependencies, eElement, "ios");
+                addToDependencyMapFromXML(dependencies, eElement, "html5");
```
Tom-Ski: Can we call this html5 for continuity.

azakhary:
screenshot 2015-02-09 21 12 46
And here is how it looks, so you guys can see it without running

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -87,6 +87,8 @@
 	List<ProjectType> modules = new ArrayList<ProjectType>();
 	List<Dependency> dependencies = new ArrayList<Dependency>();
 
+    ArrayList<ExternalExtension> externalExtensions = new ArrayList<ExternalExtension>();
```
Tom-Ski: Please run this whole PR through the eclipse formatter provided.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -432,6 +443,8 @@ public void actionPerformed (ActionEvent e) {
 		JLabel extensionsLabel = new JLabel("Extensions");
 		List<JPanel> extensionsPanels = new ArrayList<JPanel>();
 
+        SetupButton showMoreExtensionsButton = new SetupButton("Show More");
```
Tom-Ski: Change to "Show Third Party Extensions"

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
```diff
@@ -0,0 +1,244 @@
...
+        }
+        for(int i = 0; i < extensionsSnapshot.size(); i++) {
+            extensions.add(extensionsSnapshot.get(i));
+            checkBoxeMap.get(extensionsSnapshot.get(i).getName()).setSelected(true);
+        }
+    }
+
+    private void addToDependencyMapFromXML(HashMap<String, String> dependencies, Element eElement, String platform) {
+        if(eElement.getElementsByTagName(platform).item(0) != null) {
+            dependencies.put(platform, eElement.getElementsByTagName(platform).item(0).getTextContent());
```
Tom-Ski: Project dependencies can also be null if they are not supported by that platform.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -575,20 +588,23 @@ public void itemStateChanged(ItemEvent e) {
...
+
 				for (int left = depCounter - 5; left > 1; left--) {
 					extensionPanel.add(Box.createHorizontalBox());
 				}
 
 				extensionsPanels.add(extensionPanel);
-			}
 
-			add(extensionsLabel, new GridBagConstraints(0, 8, 1, 1, 0, 0, WEST, WEST, new Insets(20, 0, 0, 0), 0, 0));
+                extensionPanel.add(showMoreExtensionsButton);
```
Tom-Ski: Lets put this on its own line outside of the extension panel, as that will get in the way of the official extensions.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
azakhary: Can you provide an exact text ?

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
azakhary: could you please elaborate more on this one, it is called html5, no? Not sure where should the change be made.

azakhary:
For having multiple dependencies per platform, I will try to do it today, but might happen tomorrow :)

extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
```diff
@@ -0,0 +1,12 @@
...
+    <extension>
+       <name>Overlap2D</name>
+       <description>level and ui editor runtime</description>
+       <package>com.underwaterapps.overlap2druntime</package>
+       <version>0.0.8</version>
+       <dependencies>
+           <core>com.underwaterapps.overlap2druntime:overlap2d-runtime-libgdx</core>
+       </dependencies>
+    </extension>
+</extensions>
```
Tom-Ski: An adapted xml to support multiple dependencies per platform and unsupported platforms would look something like: 

```xml
<extensions>
    <extension>
       <name>Overlap2D</name>
       <description>level and ui editor runtime</description>
       <package>com.underwaterapps.overlap2druntime</package>
       <version>0.0.8</version>
       <projects>
           <core>
               <dependency>com.underwaterapps.overlap2druntime:overlap2d-runtime-libgdx</dependency>
           </core>
           <desktop></desktop> //No additional dependencies 
           <android></android>  //No additional dependencies
           <ios>null</ios>  // Not supported
           <html>null</html>  //Not supported
       </projects>
    </extension>
</extensions>
```

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
azakhary: That would be a huge button I image you guys want it to keep less noticeable. Maybe "Third Party" ?

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
azakhary: makes sense.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
Tom-Ski: It wouldn't be that large, and on a new line with reasonable padding away from the generate buttons it won't be a problem.  I don't want to hide it away, I want clear distinction :)

extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
azakhary: All good, except if "no additional dependencies" they just remove that line entirely to keep it short. What you think?

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
azakhary: got it.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
azakhary: html it is! :)

Tom-Ski: In our Gradle scripts that are generated and in the rest of the setup, as well as LibGDX codebase we call the html project "html" rather than "html5".

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
Tom-Ski: "If you come to LibGDX for help with these extensions you will be shot".

I'm not entirely sure, perhaps we can just make the dialog a little more complex and have hyperlinks to each of the extension homes.  So we can add a website link in the xml that will be displayed for each dependency?

extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
Tom-Ski: I think we should keep them in just so it's a little clearer to other people wanting to add their extension.

Omitting the line completely if there are no additional dependencies and keeping the line if its not supported is a little weird.  I'd rather keep it verbose, and perhaps even commented, to avoid confusion.

kotcrab: I think that each dependency should also have version of LibGDX that they are compatible with. For example my extension that relies heavily on scene2d.ui may not work with different version of LibGDX that I made and tested it with. 

Please look [here](https://github.com/kotcrab/VisEditor/wiki/VisUI#libgdx-compatibility).  I think that adding similar thing to xml file would be pretty useful, maybe it also would be good to mark old versions as outdated in gdx-setup.

Tom-Ski: Good point @kotcrab ! Let's definitely include this.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
azakhary: While we are at it, LEFT, CENTER or RIGHT? :D

azakhary: center looks cool though

azakhary:
Okay,

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
azakhary: Would be great to have the dialog have links for each extension, but seems a bit complicated for now, maybe just adjust the text a little bit so it will make sense for them, and then some day later, do that other thing. 
Something like: please do not contact libgdx for support with this extensions. (under the main label)

azakhary:
Hey guys, so to sum up. Any action required on my side at this moment? Has it been tested?

Tom-Ski:
You need to test this, use the LibGDX formatter, and rebase into one commit.

azakhary:
Thanks, we'll update when done.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtension {
+
+    private String name;
+    private String description;
+    private String version;
+
+    private Map<String, ArrayList<String>> dependencies;
```
siondream: Equally, this should be `List` instead of `ArrayList`

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
```diff
@@ -0,0 +1,266 @@
...
+import java.io.IOException;
+import java.util.*;
+
+
+import static java.awt.GridBagConstraints.*;
+
+/**
+ * Created by azakhary on 2/9/2015.
+ */
+public class ExternalExtensionsDialog extends JDialog {
```
siondream: Don't know if this is supported, but I think the Dialog should have a maxHeight and add vertical scrollbars if there are many extensions. Otherwise it'd grow too much and be hard to handle.

azakhary: I added the vertical scroll that will appear in case it is too big. But I think when there will be around 5-10 extensions, it will look good if dialog height get's a little bit bigger, and then when even mroe extensions will be added, then they can enable scrolls. Or maybe when there will be that many extensions, it will make sense to make a more complicated dialog with more advanced options. But this is all for the future, for now there is just max size and scroll pane. which is ok I assume.

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtension.java
azakhary: done

azakhary:
Well... all was good but now GdxSetupUI.java DIFF looks weird after applying eclipse formatted.
Can you guys please help? :s

azakhary:
Also, all squashed into one commit.

azakhary:
Okay All seem to be good now.

extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
```diff
@@ -37,4 +37,13 @@ public String getName() {
         return name;
     }
 
+    @Override
+    public boolean equals(Object obj) {
```
siondream: Always override `hashCode()` when you override `equals()`, here is [why](http://www.xyzws.com/javafaq/why-always-override-hashcode-if-overriding-equals/20)

Tom-Ski:
Awesome! I'll give it a test tonight, but it looks good!

azakhary:
Thanks a lot! please do test, I make mistakes :)) I will try to go through some scenarios today as well, to be on a safe side.

extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
azakhary: good point, and totally true, I just thought that because in case of hash code I have to write more code to generate the integer and this is the only use case possible in entire app it would be a quick solution. Because if I would go for every detail, I would have to use other unique identifiers for extension instead of extension name. And I didn't want to over complicate  it.

azakhary: in a  nutshell I was just lazy though :D you are right.

davebaol: Just generate hashCode() and equals() with your IDE :wink:

azakhary: hmm... i didn't knew you could do that. Thanks!

azakhary:
@Tom-Ski  how is it going? I tested some stuff too and it looks good.

azakhary:
So there was one bug with unchecking the external extension after generating and then generating again, that is now fixed.

azakhary:
@Tom-Ski how things are looking on your side?

extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
```diff
@@ -37,4 +37,17 @@ public String getName() {
...
+        if(((Dependency)obj).getName().equals(getName())) {
+            return true;
```
siondream: If `obj` is `null` or is not an instance of `Dependency` this will crash. Could you please fix it?

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -86,6 +87,7 @@
 	ProjectBuilder builder;
 	List<ProjectType> modules = new ArrayList<ProjectType>();
 	List<Dependency> dependencies = new ArrayList<Dependency>();
+	ArrayList<ExternalExtension> externalExtensions = new ArrayList<ExternalExtension>();
```
siondream: `List<ExternalExtension> externalExtensions = new ArrayList<ExternalExtension>();`

siondream:
As far as I concerned, those are my last comments.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
azakhary: funny story, I made it List first, then got carried on with other thing, copy pasted a line back, and it became ArrayList again. So close! :D

azakhary:
Let me get to this tomorrow, should take several minutes. But I also want to know how testing went.

Tom-Ski:
This is good, I'll merge and fix up some things. Thanks @azakhary !

Tom-Ski:
Just altered the layout so it's a bit more friendly and expandable. Also added a hyperlink for support page and compatibility version.  See https://github.com/libgdx/libgdx/commit/3ec8483f89c584e86ceec5425dd86a2fb95bc9ba

Updated UI
Updated UI

azakhary:
This loos so much awesomer! you rock! :)

extensions/gdx-setup/src/com/badlogic/gdx/setup/Dependency.java
loogie: Hello!

 

You have a new message, please read <http://kff.com.my/unfold.php?w6x>

 

loogie@basicelementdesign.com

loogie:
Hello!

 

You have a new message, please read <http://prodeci.com/one-third.php?gmygh>

 

loogie@basicelementdesign.com

extensions/gdx-setup/src/com/badlogic/gdx/setup/ExternalExtensionsDialog.java
loogie: Hello!

 

You have a new message, please read <http://apartamenty-pokojeas.pl/confuse.php?0f>

 

loogie@basicelementdesign.com

PR #2830, opened 2015-02-09
Added missing GL30.GL_LOGIC_OP for glBlendEquation

rinold:
Just found that I wasn't able to use the GL_LOGIC_OP as it is missing, however GL_MIN and GL_MAX are defined. If this constant was not omitted specially we could add it's definition to GL30 also.


xoppa:
Please include the reference as to why you think this is valid constant. Afaik glLogicOp is fixed render pipe and removed in gles2. Except for gles1, none of the [gles headers](https://www.khronos.org/registry/gles/#headers3) contain a GL_LOGIC_OP constant. And the [glBlendEquation](https://www.khronos.org/opengles/sdk/docs/man3/html/glBlendEquation.xhtml) specification doesn't specify `GL_LOGIC_OP` as valid mode.

rinold:
@xoppa, this is my bad, sorry, I've understand now that it was omitted specially. Was reading the wrong specification (gl3, but not es). Will check it twice next time.

PR #2831, opened 2015-02-10
Fixed copyJni() srcOffset parameters.

code-disaster:
See http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=18322 .

I've extended BufferUtilsTest for verification.


xoppa:
Thanks! Although it's probably better to keep consistency among the srcOffset and dstOffset arguments and use bytes for them. Would it be possible for you to fix and test this in the native code? If not, I could do that probably tomorrow.

Btw, did you run the test on gwt as well?

code-disaster:
Normally I would agree to keep parameters consistent, but in this case only the private native copyJni() calls are affected. The code is less verbose this way (no pointer casts required), and minimal faster (no arithmetic shift). I did not set up my libGDX clone for compiling native code yet, so that would have been more work for me, too.

Sorry, I didn't run the test on GWT, for some reason Chrome doesn't recognise the developer plugin on my current system (OS X 10.10).

xoppa:
Thanks!

PR #2835, opened 2015-02-11
Made Cell#defaults() public 

gjroelofs:
So that when Table#defaults() is changed, it can be reverted without having to call Table#reset().

Table.defaults().defaults() is obviously not the best code, but Cell is pooleable so reset() is already taken.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Cell.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Cell.java
@@ -889,7 +889,7 @@ public class Cell<T extends Actor> implements Poolable {
 	}
 
 	/** Set all constraints to cell default values. */
-	void defaults () {
+	public void defaults () {
 		minWidth = Value.minWidth;
 		minHeight = Value.minHeight;
 		prefWidth = Value.prefWidth;


NathanSweet:
`.defaults().defaults()` wouldn't be an acceptable public API since they do
different things. I think `reset()` can do all the work just fine.

gjroelofs:
'reset()' is currently used to implement Pooleable.
Using it on a defaults() cell will give NPE during layout.
(A mistake I made myself as well as I thought reset() would return the defaults.)

It cannot be used for both resetting the defaults, and the pooleable behavior implementation.
What do you suggest alternatively then?

NathanSweet:
What NPE?

gjroelofs:
Can't reproduce it on 1.5.4, so I'll send in a new PR with defaults() in
reset().
Reading the code the reset() values are always set in add().

I'm was pretty sure an NPE was caused by calling reset() on defaults, but
that's in our inhouse project that uses <1.3.
Probably a goof on my part. Apologies for the mix up.

NathanSweet:
Note I shuffled around a little code in addition to merging Cell#defaults with #reset.

Do we need a new PR? If you can identify a problem with the current code, I'm happy to work on a fix. I tried resetting cell, row, column defaults and it seems to work fine. Spine and my other apps are working too.

gjroelofs:
I'll just open a PR when, or better said, if the bug pops up.
I can't reproduce it either, so I'll just chalk this up to PICNIC for now.

Thanks for the quick fix!

PR #2836, opened 2015-02-11
TiledMap rendering failing to respect firstgid value for multiple external tilesets

psiegeldb:
I started looking into this when I noticed an odd difference between how the Tiled editor works vs. the libgdx renderer when an external tileset's image is changed out from under it.  To expose this issue you must have multiple external tilesets listed and then expand the image of an existing tileset early in the list.  The tiled editor continues to load the existing map just fine, though it will modify all gids after the expansion point on the next save.  The libgdx renderer, however, will incorrectly render the newly added tiles instead of the pre-existing tiles in the later tileset.

Consider the included test data in this pull request.  Two copies of the same map are included.  In the first, there are two external tilesets:
  <tileset firstgid="1" source="grass.tsx"/>
  <tileset firstgid="2" source="water.tsx"/>

Each tileset has exactly 1 tile, thus the firstgid of the first tileset is 1, and the second is 2.  All tiles painted in the included tile layer are therefor values 1 or 2.

The second version references a modfied version of the first tileset, though is otherwise idenctical to the first:
  <tileset firstgid="1" source="grass_extended.tsx"/>
  <tileset firstgid="2" source="water.tsx"/>

In grass_extended.tsx, the image has been expanded to include 4 tiles, thus has tile IDs 1, 2, 3, and 4.  However, water.tsx still indicates that it's firstgid is 2, thus we have two tilesets that both claim to have tile gid 2.  The layer data is unchanged, and therefore expects tile gid 2 to refer to the first tile in the water.tsx tileset, not the second grass tile.

The fix changes how tiles are searched for by gid, reverse-iterating the tilesets under the assumption that they are in order of ascending firstgid.  It then searches each tileset for specified tile only if the gid requested is greater than the firstgid value.

As an additional efficiency, I have recorded the firstgid value in the TiledMapTileSet object.  I know this value is already included in the MapProperties, but I thought for a value referenced so frequently during a render loop it would be preferably to extract it into a member variable.


MobiDevelop:
The reason that firstgid was not part of the TileMapTileSet object is because that is a Tiled specific construct.

psiegel:
Apologies, made this pull request while logged in as the wrong user.  Not sure if that matters, but I'm closing it and recreating it under the correct user.

PR #2837, opened 2015-02-11
TiledMap rendering failing to respect firstgid value for multiple external tilesets

psiegel:
I started looking into this when I noticed an odd difference between how the Tiled editor works vs. the libgdx renderer when an external tileset's image is changed out from under it. To expose this issue you must have multiple external tilesets listed and then expand the image of an existing tileset early in the list. The tiled editor continues to load the existing map just fine, though it will modify all gids after the expansion point on the next save. The libgdx renderer, however, will incorrectly render the newly added tiles instead of the pre-existing tiles in the later tileset.

Consider the included test data in this pull request. Two copies of the same map are included. In the first, there are two external tilesets:

Each tileset has exactly 1 tile, thus the firstgid of the first tileset is 1, and the second is 2. All tiles painted in the included tile layer are therefor values 1 or 2.

The second version references a modfied version of the first tileset, though is otherwise idenctical to the first:

In grass_extended.tsx, the image has been expanded to include 4 tiles, thus has tile IDs 1, 2, 3, and 4. However, water.tsx still indicates that it's firstgid is 2, thus we have two tilesets that both claim to have tile gid 2. The layer data is unchanged, and therefore expects tile gid 2 to refer to the first tile in the water.tsx tileset, not the second grass tile.

The fix changes how tiles are searched for by gid, reverse-iterating the tilesets under the assumption that they are in order of ascending firstgid. It then searches each tileset for specified tile only if the gid requested is greater than the firstgid value.

As an additional efficiency, I have recorded the firstgid value in the TiledMapTileSet object. I know this value is already included in the MapProperties, but I thought for a value referenced so frequently during a render loop it would be preferably to extract it into a member variable.


psiegel:
MobiDevelop commented 15 minutes ago
The reason that firstgid was not part of the TileMapTileSet object is because that is a Tiled specific construct. It is never referenced during rendering.

Right, my proposal though is to use it during rendering to maintain backwards compatibility with tile maps that reference external tilesets which are subsequently modified to include more tiles.

MobiDevelop:
I think we need to determine if this is actually something we want to do (I'm not sure it is) and if so, fix it the correct way by adjusting the gids in the layer data like Tiled does.

psiegel:
That's certainly a reasonable approach, but would require heavy lifting by the loader.  Also, it may be confusing to users if the gids in the runtime do not match the data.

Actually, now that you mention it, it seems to me that firstgid does not belong in the TiledMapTileSet object at all, but really in the TiledMap object (or perhaps the TiledMapTileSets object).  Especially if you have more than one map that refers to the same external tileset, some of which have out-dated firstgids and some of which have newer correct values.  Now I wonder though if one was to load multiple tile maps that refer to the same external tile set, do they each create their own copy of the tileset or do they refer to the same object?

MobiDevelop:
They'd keep their own copies of tilesets.

psiegel:
OK, well, this little change vs. modifying all gids is certainly something we could debate, but I'm curious if you still have objections to addressing this issue at all.

MobiDevelop:
To be honest, I'd be quite surprised if this was something that happened frequently in real use... Where the underlying tsx tilesets have been updated to include more (or less) tiles without having opened and saved with tmx files using them.

psiegel:
Here's my use case: 

Let's say I release a game with a standard set of 4 tilesets which we use to create say two dozen levels with.  After release we decide to release downloadable content of 6 more levels.  While developing those new levels, our artist decides he wants to add a couple new tiles to the tileset to use with the new levels.  With this fix, we can distribute the updated tileset and the 6 new levels to the end user and everything works fine.  Without it, distributing just those files would result in the original 24 levels appearing totally broken.  Our options are then:

1.  Open up every one of the original 24 levels in Tiled, resave to have the correct updated gid values, then distribute to the client all 30 levels with the new tilesets.  This has the disadvantage of requiring extra download by the end user, as well as opens us up to errors if we miss a level in the process.

2.  Create duplicate copies of the tilesets for use in all new levels, so on the client we have now have two different tilesets, those used by the original 24 levels and those used by the new 6 levels.  Obviously, this scales poorly as we release further updates and have to start juggling many different revisions of nearly identical tilesets.

I'd argue this is a pretty small change for a very helpful method of allowing us to update our tilesets down the road while maintaing backward compatibility with old maps.

MobiDevelop:
Well, you aren't going to get around including multiple copies of the tilesets in your scenario because you'll always have the bundled copy.

As for looking up the tiles at render time, that is not going to happen as it would be incredibly slow.  The loader assigns tile objects to cells at load time... That is the only time this getTile method is called (unless the user calls it herself within her code).

psiegel:
True, the original shipped with client will have the bundled copy, though this only solves the problem for one iteration.  As I mentioned it does not scale -- the more maps and expansions you create, the more copies of the tileset you end up having to maintain.  Wouldn't it be easier to maintain a single copy of each tileset that all maps could use?

And yes, you are right, the extra step of duplicating the firstgid for the sake of efficiency is probably unnecessary.  I missed that getTile is only called once during load, so I suppose retrieving it out of the MapProperties is fine.

I'd be happy to ditch that part of it but I maintain that the reverse iteration of the TileSets in getTile is still desirable.  Heck, thinking about it, you could even ditch the check against firstgid entirely.  Reverse iteration by itself is probably enough to have the desired result of obtaining the correct tile when there's a conflict.  I really don't see any downside to that change.

MobiDevelop:
Yeah, I don't really care if the order is reversed. If it helps in this
case and doesn't hurt the normal case - let's go ahead with it.

psiegel:
Cool, I'll clean up the pull request tomorrow morning to limit it to just that change.

psiegel:
Sorry, I have not much experience with git.  I reverted my earlier commits and created a new commit with just the changes discussed.  Git seems to have rolled all these changes into this pull request.  Is that correct?  The only relevant commit for inclusion in this pull request is:

https://github.com/psiegel/libgdx/commit/a19376e837c7c5def79f854e6026a192c294b164

Should I cancel this pull request and create a new one with just that commit?  Or is there some way to trim the extraneous commits from this pull request?  Or am I over thinking this and we're good to go?

MobiDevelop:
It would be cleaner to just pull that one commit into a new PR.

psiegel:
OK, think I finally figured out how to do it.  New pull request here:
https://github.com/libgdx/libgdx/pull/2843

loogie:
Hello!

 

You have a new message, please read <http://bunstede.nl/second.php?z>

 

loogie@basicelementdesign.com

PR #2843, opened 2015-02-14
Changed TiledMapTileSets.getTileSet to backwards iterate

psiegel:
This allows better backwards compatibility with older maps created with a shared external tileset should the image of that tileset be later expanded.  Backwards iteration prefers earlier versions of conflicting gids in this case.  Includes test to illustrate the above issue.

Clean-up of closed pull request: https://github.com/libgdx/libgdx/pull/2837


gdx/src/com/badlogic/gdx/maps/tiled/TiledMapTileSets.java
```diff
@@ -67,7 +67,14 @@ public void removeTileSet (TiledMapTileSet tileset) {
 	/** @param id id of the {@link TiledMapTile} to get.
 	 * @return tile with matching id, null if it doesn't exist */
 	public TiledMapTile getTile (int id) {
-		for (TiledMapTileSet tileset : tilesets) {
+		// The purpose of backward iteration here is to maintain backwards compatibility
+		// with maps created with earlier versions of a shared tileset.  The assumption
+		// is that the tilesets are in order of ascending firstgid, and by backward
+		// iterating precedence for conflicts is given to later tilesets in the list, 
+		// which are likely to be the earlier version of any given gid.  
+		// See TiledMapModifiedExternalTilesetTest for example of this issue.
```
MobiDevelop: Probable better not to reference tests here but it's ok for now.

PR #2846, opened 2015-02-14
Fixed a NPE in HttpResponse.getResultAsString()

nooone:
In case a HttpResponse does not contain any content, using `HttpResponse.getResultAsString()` will result in a NullPointerException:

    java.lang.NullPointerException
        at java.io.Reader.<init>(Reader.java:78)
        at java.io.InputStreamReader.<init>(InputStreamReader.java:72)
        at com.badlogic.gdx.utils.StreamUtils.copyStreamToString(StreamUtils.java:74)
        at com.badlogic.gdx.net.NetJavaImpl$HttpClientResponse.getResultAsString(NetJavaImpl.java:75)

This PR adds a nullcheck to `HttpResponse.getResultAsString()` and returns `""` instead.


badlogic:
Thanks!

Issue #2847, opened 2015-02-14
Locking conflict when sending multiple HttpRequests


nooone:
Using LibGDX 1.5.4, sending multiple `HttpRequests` quickly after one another *may* result in the following Exception. It happens on Android and Desktop.

    java.lang.IllegalMonitorStateException
        at java.util.concurrent.locks.ReentrantLock$Sync.tryRelease(ReentrantLock.java:122)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1238)
        at java.util.concurrent.locks.ReentrantLock.unlock(ReentrantLock.java:428)
        at com.badlogic.gdx.net.NetJavaImpl$1.run(NetJavaImpl.java:208)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
        at java.lang.Thread.run(Thread.java:818)

[This PR](https://github.com/libgdx/libgdx/pull/2771) fixing [this issue](https://github.com/libgdx/libgdx/issues/2656) seems to be responsible for this new behaviour. The problem seems to be that each sent request results in a new Thread. These threads use the same lock in a conflicting way.

PR #2849, opened 2015-02-14
Fix for Issue #2847

nooone:
I have replaced the lock with synchronized methods. Fixes https://github.com/libgdx/libgdx/issues/2847


nooone:
A little update on this:

Currently both 1.5.4 and 1.5.5-SNAPSHOT have problems with HTTP. 1.5.4 ends in a NPE when a response did not have any content (fixed via https://github.com/libgdx/libgdx/pull/2846). Unfortunately with 1.5.5-SNAPSHOT came https://github.com/libgdx/libgdx/pull/2771 which leads to an exception in case of different responses arriving at the same time. *This* PR should fix that and the other issue as well.

I have tried it in my own game, as well as via the NetAPITest on both desktop and android and it works.

Since I do not own an iPhone or Mac, I cannot test it on these platforms.

MobiDevelop:
Looks OK to me... If not we'll tar and feather you.

PR #2852, opened 2015-02-15
Cache colour float for ShapeRenderer

Tom-Ski:
* Adds `color (float)` to ImmediateModeRenderer inteface and implements
* Optimizes ShapeRenderer methods with cached colour float.


MobiDevelop:
If the user modifies the color via `getColor()`, the cached value won't have been updated. We either need to make it clear they'd have to then call `setColor()`, or get the float bits at the method calls which would still be an improvement over calculating for every vertex in the ImmediateModeRenderer.

MobiDevelop:
Looks good to me, ship it.

PR #2856, opened 2015-02-18
Expose clearListeners method on ControllerManager

nfantone:
Currently, the only way of removing a previously added `ControllerListener` is by **holding a reference** to that instance. More often than not that is either impractical or, having it accessible, implies a violation of the architecture or forces to bend classes' responsibilities.

I have noticed that the `ControllerManager` interface defines a `clearListeners` method, which is the only one not accessible through a static method on the `Controllers` class. Was there a reason why this one was left behind?

Having this publicly accessible would be pretty handy.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 [1.5.5]
 - API Change: 3D Animation, NodeAnimation keyframes are separated into translation, rotation and scaling
 - Added capability to enable color markup from inside skin json file.
+- API Change: Exposed method ControllerManager#clearListeners on Controllers class
 
 [1.5.4]
 - Added support for image layers in Tiled maps (TiledMapImageLayer)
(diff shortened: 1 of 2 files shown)


siondream:
I am okay with this, can you add this API change to the [CHANGES](https://github.com/libgdx/libgdx/blob/master/CHANGES) file? Thanks!

nfantone:
Yes, of course. I won't be able to do it today, but you'll have it by tomorrow.

MobiDevelop:
I'd think a minor addition like this could be omitted  from CHANGES.

nfantone:
Done! Found the time to do it today. Squashed commits.

siondream:
Thanks a lot @nfantone

nfantone:
You are quite welcome.

PR #2857, opened 2015-02-18
Tween - Interpolation Helper

davcri:
I created the ```Tween``` class in the _gdx.utils_ package. This class it's a sort of helper classes for libGDX's [Interpolation](https://github.com/libgdx/libgdx/wiki/Interpolation) ([code](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Interpolation.java)) class. It uses it's functions, adding a simple API for programmers.

I also created a simple test for this class called ```TweenTest```, I don't know if it is necessary to include it in this pull request, but it can be useful to understand what Tween is and how it works.

__License__: Apache v2.0
__Code formatting__: I used the eclipse formatter
__CLA__: I haven't signed the CLA yet, I'll do that if necessary.

---

I posted also [on the forum](http://www.badlogicgames.com/forum/viewtopic.php?f=17&t=18400&sid=e8f6a8c7829cc5f4f7ae90e2ef10d757) where there is some information that I'll report here: 

Speaking with the guys on the IRC I ended up with few doubts: 

__Q__: This code should be in the Interpolation class.
I prefer to mantain the things separate, but I don't know much about the libGDX conventions. So tell me what you think.

__Q__: Universal Tween Engine has much more features. 
Yes but I prefer to have a simple and Object Oriented approach, that is also integrated with libGDX. Other features may be added in future.

__Q__: Is "Tween" the correct name for this?
I don't really know :D There are many terms that are used for these things: easing, tween, interpolation. For me it's just equations and an API to handle them, moreover I'm Italian and I don't know about the use of these terms in the common language.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Tween.java
+++ b/gdx/src/com/badlogic/gdx/utils/Tween.java
@@ -1,0 +1,100 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.utils;
+
+import com.badlogic.gdx.math.Interpolation;
+
+/** Interpolates values using differents easing equations.
+ * 
+ * Helper class for gdx.utils.Interpolation. */
+public class Tween {
+	float initialValue;
+	float finalValue;
+	float duration;
+	float percentage = 0;
+	float elapsed = 0;
+	float value = 0;
+	private boolean isTweenActive = true;
+	Interpolation interpolation;
+
+	/** Creates a new tween.
+	 * 
+	 * @param interpolation Equation used for the interpolation.
+	 * @param initialValue Initial value of the tween.
+	 * @param finalValue Final value of the tween.
+	 * @param duration Duration in seconds. */
+	public Tween (Interpolation interpolation, float initialValue, float finalValue, float duration) {
+		this.initialValue = initialValue;
+		this.finalValue = finalValue;
+		this.duration = duration;
+		this.interpolation = interpolation;
+	}
+
+	/** Updates the tween.
+	 * 
+	 * @param delta Frame time. */
+	public void update (float delta) {
+		if (isTweenActive && percentage <= 1) {
+			elapsed += delta;
+			percentage = elapsed / duration;
+
+			// don't let the percentage go over 1
+			if (percentage > 1) percentage = 1;
+
+			value = initialValue + (finalValue - initialValue) * interpolation.apply(percentage);
+		}
+	}
+
+	/** Gets the current tween value.
+	 * 
+	 * @return Value of the tween. */
+	public float getValue () {
+		return value;
+	}
+
+	/** Pauses the tween.
+	 * 
+	 * After pausing it, it can be resumed with the resume() method. */
+	public void pause () {
+		isTweenActive = false;
+	}
+
+	/** Resumes a paused or stopped tween. */
+	public void resume () {
+		isTweenActive = true;
+	}
+
+	/** Stops the tween reverting the interpolation to the initial state. */
+	public void stop () {
+		isTweenActive = false;
+		elapsed = 0;
+		value = initialValue;
+	}
+
+	/** Restars the tween. */
+	public void restart () {
+		stop();
+		resume();
+	}
+
+	/** Tells if the tween is active.
+	 * 
+	 * @return true if the tween is active, false otherwise. */
+	public boolean isTweenActive () {
+		return isTweenActive;
+	}
+}
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 [1.5.5]
 - API Change: 3D Animation, NodeAnimation keyframes are separated into translation, rotation and scaling
+- API Addition: Added new class Tween to gdx.utils package
 
 [1.5.4]
 - Added support for image layers in Tiled maps (TiledMapImageLayer)
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/GdxTests.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/GdxTests.java
@@ -219,6 +219,7 @@ public class GdxTests {
 		TimeUtilsTest.class,
 		TouchpadTest.class,
 		TreeTest.class,
+		TweenTest.class,
 		UISimpleTest.class,
 		UITest.class,
 		Vector2dTest.class,
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/TweenTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/TweenTest.java
@@ -1,0 +1,90 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.utils;
+
+import com.badlogic.gdx.Gdx;
+import com.badlogic.gdx.Input.Keys;
+import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.Texture;
+import com.badlogic.gdx.graphics.g2d.Sprite;
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.badlogic.gdx.math.Interpolation;
+import com.badlogic.gdx.utils.Tween;
+
+/** Creates a sprite and manages its properties using Tween.
+ * 
+ * Interpolates the sprite position from left to right and fades the sprite in, changing the alpha properties.
+ * 
+ * Keys: 
+ * - P: pauses/resumes the tweens animation 
+ * - S: stops the tweens animation  
+ * - click: restarts the tweens */
+public class TweenTest extends GdxTest {
+	SpriteBatch batch;
+	Sprite img;
+	Tween positionTween;
+	Tween alphaTween;
+
+	@Override
+	public void create () {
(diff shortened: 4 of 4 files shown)


gdx/src/com/badlogic/gdx/utils/Tween.java
```diff
@@ -0,0 +1,100 @@
...
+	 * @param delta Frame time. */
+	public void update (float delta) {
+		if (isTweenActive && percentage <= 1) {
+			elapsed += delta;
+			percentage = elapsed / duration;
+
+			// don't let the percentage go over 1
+			if (percentage > 1) percentage = 1;
+
+			value = initialValue + (finalValue - initialValue) * interpolation.apply(percentage);
```
NathanSweet: Interpolation has an `apply` method that takes start and end values.

gdx/src/com/badlogic/gdx/utils/Tween.java
```diff
@@ -0,0 +1,100 @@
...
+		this.interpolation = interpolation;
+	}
+
+	/** Updates the tween.
+	 * 
+	 * @param delta Frame time. */
+	public void update (float delta) {
+		if (isTweenActive && percentage <= 1) {
+			elapsed += delta;
+			percentage = elapsed / duration;
```
NathanSweet: Percentage doesn't really need a field, it could be computed. Also could have a getter.

gdx/src/com/badlogic/gdx/utils/Tween.java
```diff
@@ -0,0 +1,100 @@
...
+		this.finalValue = finalValue;
+		this.duration = duration;
+		this.interpolation = interpolation;
+	}
+
+	/** Updates the tween.
+	 * 
+	 * @param delta Frame time. */
+	public void update (float delta) {
+		if (isTweenActive && percentage <= 1) {
```
NathanSweet: Is it useful to call update on a paused tween? Seems more likely an app has a list of active tweens and it updates them all.

NathanSweet:
Interpolation is stateless, so something like this doesn't belong there.

A "tween" is simply a float field that you increment and use a duration and Interpolation to get a value. Having a class to encapsulate that is fine. I'm not sure it should have pause/resume/stop methods though. Instead it should always be playing, just don't update it if you want it paused. This discourages keeping game state (beyond the elapsed time) inside the tween as well as updating a tween that does nothing. It could still have restart (or reset?) to put it back at the beginning, as well as setPercent/getPercent and setTime/getTime. It could also have repeat modes: normal, reverse, loop, loop reverse, loop pingpong.

As for the name, I'm not sure. I guess tween isn't terrible. Any other ideas?

siondream:
Maybe I'm saying something stupid but... Could the Actor Actions make use of this? They basically have stateful interpolations, am I wrong? If they could and we don't make use of this, we could be duplicating logic unnecessarily.

NathanSweet:
See FloatAction. I've never used it though, not sure anyone ever has.

xoppa:
Which problem is this actually solving?

gdx/src/com/badlogic/gdx/utils/Tween.java
davcri: Thanks, I missed it ! I'll fix this.

gdx/src/com/badlogic/gdx/utils/Tween.java
davcri: Totally right. 
Removing percentage field and adding a getter.

gdx/src/com/badlogic/gdx/utils/Tween.java
davcri: No it isn't. I do that because I wanted to delegate all of the responsabilities to the Tween class. I don't know if this is helpful or not. 
I was thinking also about a sort of TweenContainer class that holds a list of Tweens and manages them. This class could pauses, stops, resume or also concatenate them. But this is just an idea for the future.

davcri:
Ok I'm changing the Tween class according to the suggestions of Nathan. I'm adding also some repeat modes. 
Tomorrow I'll make and push the commits to this repository. 

I never used Scene2D, so I don't know if this could be merged.

PS:siodream I read your blog post about surviving with GWT and libGDX, I found it really useful thanks :smile:

NathanSweet:
Note that Xoppa brings up a serious question.

davcri:
I thought he was referring toFloatAction. 
This PR doesn't fix nothing. It is an helper class for the Interpolation class that is present in libGDX. In some cases where someone wants a simple tween class but doesn't want to import other tweening engines, or write it from scratch.

xoppa:
Is it about naming? I mean if you want to have a simple tween, then you could do something like:
`position.set(from).lerp(to, BounceOut.apply(a = MathUtils.clamp(a + delta * speed, 0f, 1f)));`
You don't need a tweening engine to create a simple tween. Note that in this example I'm actually tweening something (the position), while this Tween class basically only interpolates the alpha value (might be my lack of english though). 

I'm not saying that we shouldn't include a helper class to encapsulate this, just trying to understand why anyone would use it. Perhaps it would be more useful to take a more generic approach (e.g. add a `Tweenable` (`Lerpable` :D) interface or alike). Or include other animating functionality like looping, chaining, etc (remember #1646).

You must have had a reason to go for this specific implementation, what was it?

xoppa:
Sorry, I should read better. You're obviously tweening a single floating point value (e.g. the alpha value of a color, or the x component of a coordinate). So, by simple you meant restricted to a single float.

davcri:
> I'm not saying that we shouldn't include a helper class to encapsulate this, just trying to understand why anyone would use it. Perhaps it would be more useful to take a more generic approach (e.g. add a Tweenable (Lerpable :D) interface or alike). Or include other animating functionality like looping, chaining, etc (remember #1646).

Yes I could add more features, yesterday I added baisc looping. I could work on repeat, chain, delay and callbacks. what do you think ? 
For the Tweenable interface I can't think about a good implementation that is useful for both code mantainers and users.

> It could still have restart (or reset?) to put it back at the beginning, as well as setPercent/getPercent and setTime/getTime

* setPercent(float alpha) should throw an error if alpha is out of the [0,1] range ? Or just a print with Gdx.app.error() ?
* setTime() is really useful ? I found the name a bit misleading.

xoppa:
I'm not saying that you should add more features. I'm just asking why anyone would use this. Adding classes or features which have no or very limited use-cases doesn't sound like a good plan. It might even confuse users and/or encourage bad practice.

davcri:
I think that a class that takes the responsibilities for all the interpolations is very useful in game programming: splash screens, moving objects, images fading, transitions, camera movements. Yes you could use the``` Vector2.Interpolate()``` method, but you should add other code to handle more complicated effects like loop, repeat, chaining.
I got the idea to write this class when I saw that libGDX has this Interpolation class that does a great job, so I thought: why we shouldn't take advantage of it, creating a simple API for programmers? 
The overhead is acceptable and programmers will not need to reinvent the wheel every time or import an external engine.

Let's take for example [this video made by _dermetfan_](https://www.youtube.com/watch?v=2PqwxYVlivA). He creates an image that fades in, wait 2 seconds and then fades out, using the _Universal Tween Engine_. Source code here 
[Splash.java](https://bitbucket.org/dermetfan/blackpoint2/src/21c010c6da76113ff4f7a374743626483b81c23a/Blackpoint2/src/net/dermetfan/blackpoint2/screens/Splash.java?at=default)
[SpriteAccessor.java](https://bitbucket.org/dermetfan/blackpoint2/src/21c010c6da76113ff4f7a374743626483b81c23a/Blackpoint2/src/net/dermetfan/blackpoint2/tween/SpriteAccessor.java?at=default)

Using a Tween class will bring some advantages:  
- you don't need to create an accessor class
- it's more object oriented: object properties has to be setted manually using something like ```myObj.setValue(tween.getValue())```. 
- more intuitive and simpler because users needs to write less code
- you don't need to import an external engine

rinold:
I'm very sorry for interfering the conversation, and please, consider that everything below is my humble opinion...

But, how do we really benefit from it in the real life? I mean it's great for a simple example you created in a test, but in a real life I don't think anybody is using such a big and great thing as libGDX to create one image screen application. And if we are using it for something more, then:

1. What about memory and garbage? - Universal Tween Engine is using pooling and I can sleep calm and forget about it, in your case - created Tween seems to be immutable regarding it's values and duration and  the game will be flooded with allocs of new Tweens in the result.

2. As I see - you will need to write much-much more code, to pause/resume and controll ALL the Tweens for ALL your objects in your game.

3. It lacks, sometimes essential things, like callbacks, relative value tweens and so on. Some strange guys might use your Tween and then in the middle find out that it lacks something, and as they already wrote tons of code, they will ask for or add new code to it. And in the end we will come to something like UTE clone inside of libGDX...

As I've said previously - it's only my IMHO and I don't mind of adding it in general, but please think about the points I've stated above...

MobiDevelop:
It seems to me that having a Tween object that only handles single float values isn't particularly useful.

xoppa:
@davcri The examples you give are only very limited possible with this class. I don't see how interpolating a single float and requiring the user to apply this himself is more object orientated (if anything, it's less object orientated). While you might find this approach intuitive, I doubt it is for others.

What about having a slightly more generic approach, for example (untested code):
```java
public abstract static class Tween<T> {
	T what;
	T from;
	T to;
	Interpolation interpolation;
	float current;
	float speed;

	public void reset(T what, T from, T to, Interpolation interpolation, float duration) {
		this.what = what;
		this.from = from;
		this.to = to;
		this.interpolation = interpolation;
		this.speed = 1f / duration;
		current = 0f;
	}

	public boolean update(float delta) {
		current = MathUtils.clamp(current + speed * delta, 0f, 1f);
		tween(interpolation.apply(current));
		return current < 1f;
	}

	protected abstract void tween(float a);

	// the other methods...
}
```

davcri:
@rinold: You are not interfering at all, I accept every critics :smile: This code is very limited, and I'm happy to discuss to improve code.
I agree with all of your points: Memory management needs to be improved and it lacks features. 
This class is quite useless at the moment, but I surely want to improve it !

@MobiDevelop: Actually, yes. 

@xoppa: I like this, thank you for the suggestion ! I wouldn't put the ```T what``` field, just because I like to manually get a value from a Tween and I don't like the fact that the Tween touches other objects; but that's just me, there isn't really a big difference.

I would add another generic type : 

``` java
public class Tween<T1, T2>
```
in this way we could do something like:

``` java
public class alphaTween<Sprite, float> extends Tween {
    public void apply(Sprite what, float from, float to, Interpolation interpolation, float duration) {
        // implementation...
    }
}
```
But... as said by rinold,  in this way there's a risk to clone the UTE.

xoppa:
Well in that case it would be (untested code):
```java
public class VectorTween<T extends Vector> extends Tween<T> {
    public void apply(T what, T from, T to, float alpha) {
        what.set(from).lerp(to, alpha);
    }
}
```
which you could use as (untested code):
```java
VectorTween<Vector2> tween = new VectorTween<Vector2>();
tween.reset(new Vector2(), new Vector2(0,0), new Vector2(10, 10), interpolation, 10f);
...
tween.update(Gdx.graphics.getDeltaTime());
sprite.setPosition(tween.getWhat());
```
Anyways, I think we agree that this PR (as it currently is) is not ready to be merged. Perhaps work out a new PR from scratch with these comments in mind? Could first try some different approaches and compare them e.g. for different scenario's.

I think there are two things you're trying to do (which might not necessarily be related to each other):
- encapsulate everything required to perform an interpolation/tween
- "playback" functionality like looping, reverse/speed etc.

davcri:
Sure this is not ready, closed.
Summarizing, these things need to be done:

- Add repeat modes
- Use a generic interface approach
- Add Callbacks, chaining, maybe other stuff
- More testing, to see if it can be really useful

Tell me if I'm forgetting something. 

It will take some time, from the next week I have to came back to college :smile:
Thank you all guys, I'm sorry for creating a bit of a mess with this PR.

Issue #2714, opened 2015-01-05
AssetManager: better progress indication for TextureAtlasLoader and probably others as well


klaus7:
Wouldn't it be better to add to the count of `com.badlogic.gdx.assets.AssetManager#toLoad` all dependencies from the assets and change this line at https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/assets/AssetManager.java#L470 from `if (tasks.size() == 1) loaded++;` to simply `loaded++;`?

I'm only loading one big TextureAtlas, hence the progress indication is of little use right now for me.

I changed the classes as described and it works as expected. 

What do you think?

klaus7:
Sorry, I guess I didn't describe my concern well enough.

When using the `AssetManager` and loading a big `TextureAtlas` (mine has 22 pages) the whole atlas counts as one asset. The individual pages are seen as dependencies and do not increase the `toLoad` count. My proposal is to change this behaviour, so dependencies (in case of the `TextureAtlas` this would be its pages) also increase the `toLoad` count, while at the same time the `loaded` must also be increased for every asset it loads. 

This would lead to a more granular progress.

As it seems other developers also stumbled upon this issue. (http://badlogicgames.com/forum/viewtopic.php?f=11&t=11298 and http://badlogicgames.com/forum/viewtopic.php?t=11122)

NathanSweet:
`​​
​loaded
​`​
/
​`​
toLoad
​`​
are used to determine progress percentage. Doing as you suggest would mean
the percentage would go backward. Eg 10%, 20%, then dependencies are added,
15%, etc.
​The dependencies are added by `AssetLoadingTask` only when the asset needs
to load, because even knowing what the dependencies are can require some
loading. Eg, for `TextureAtlas`, to parse the atlas file. I'm all for
improving it, but I'm not sure how.

M
​y​
​​
​ suggestion t​
​o queue y​
​our dependencies is what the guy in the second thread you linked is doing.
Call load on your 22 images so they are reflected by `loaded`/`toLoad`
rather than relying on the dependency mechanism where they won't.

badlogic:
An atlas with 22 pages is kinda exactly oposite of what you want to achieve
with the atlas. Unless you sorted all images so that when drawing a scene
you only reference one of those pages.

klaus7:
@NathanSweet Ok I see, that's the pitfall. Maybe I can think of another solution, otherwise I will do it like the guy in that thread.

[...]

PR #2875, opened 2015-02-22
Include dependencies count when reporting asset manager progress.

lnr0626:
This makes it so that when an asset has dependencies, the getProgress() call reflects the actual process for loading the assets. 

I'm not sure if it was created this way intentionally, but in my mind this gives a more accurate indication of the load progress than how it currently is. This can however make the percent completion go backwards at times (which I guess could be why it is the way that it currently is...). If this is an issue for including this change, I'll look into fixing it.

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetLoadingTask.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetLoadingTask.java
@@ -39,6 +39,7 @@ class AssetLoadingTask implements AsyncTask<Void> {
 	final long startTime;
 
 	volatile boolean asyncDone = false;
+	volatile boolean dependenciesInjected = false;
 	volatile boolean dependenciesLoaded = false;
 	volatile Array<AssetDescriptor> dependencies;
 	volatile AsyncResult<Void> depsFuture = null;
@@ -64,6 +65,7 @@ class AssetLoadingTask implements AsyncTask<Void> {
 			dependencies = asyncLoader.getDependencies(assetDesc.fileName, resolve(loader, assetDesc), assetDesc.params);
 			if (dependencies != null) {
 				manager.injectDependencies(assetDesc.fileName, dependencies);
+				dependenciesInjected = true;
 			} else {
 				// if we have no dependencies, we load the async part of the task immediately.
 				asyncLoader.loadAsync(manager, assetDesc.fileName, resolve(loader, assetDesc), assetDesc.params);
@@ -101,6 +103,7 @@ class AssetLoadingTask implements AsyncTask<Void> {
 				return;
 			}
 			manager.injectDependencies(assetDesc.fileName, dependencies);
+			dependenciesInjected = true;
 		} else {
 			asset = syncLoader.load(manager, assetDesc.fileName, resolve(loader, assetDesc), assetDesc.params);
 		}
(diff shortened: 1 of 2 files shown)


NathanSweet:
See here:
https://github.com/libgdx/libgdx/issues/2714#issuecomment-68799514

lnr0626:
Thanks for the quick response. I'll see if I can come up with anything for getting around that.

NathanSweet:
Unless you know all the assets up front, I don't think there is much we can do. Some assets need to do loading just to know what other assets they depend on (eg, a texture atlas has to parse the atlas file to know what images it needs). As the developer, likely you know what assets you need, so you can just add them all to the asset manager.

Probably the easiest solution is to run you app and note the number from `getLoadedAssets`. If that value is `x`, you can use `getLoadedAssets() / x` to get a percentage that considers all assets that will be loaded.

lnr0626:
There is not a way to accurately give a progress percentage without knowing the number of assets up front, however there should be a way to take into account the number of dependencies loaded using the tasks stack. I've been playing around and have gotten something that works doing this - just pushed it.

NathanSweet:
What is this doing exactly?

lnr0626:
```
int tasksLength = tasks.size();
float completedProgress = loaded / (float) toLoad;
```
`completedProgress` is the percent of requested assets which have been fully loaded - it's what is currently being returned.

The next bit is the estimation for the progress of the asset which is currently being loaded. If the length of the stack named `tasks` is greater than 1, it means that the dependencies for this task are currently being loaded (or at least that was what I was able to determine). If that's the case, then the progress is estimated as .9 / the number of tasks (which is roughly the number of dependencies to load + 1). 

If the tasksLength is equal to 1, then that could that either the asset task has not yet attempted to inject the dependencies for that asset, or that all of the dependencies have been loaded. To determine which one is the case, I had to add a dependenciesInjected variable. This defaults to false and is set to true once the dependencies for that asset have been injected into the AssetManager.

If the dependencies have already been injected, and the size of `tasks` == 1, that means that all of the dependencies have been loaded, so we use .9 as the progress estimation for the current asset. if the dependencies have not yet been injected, then the progress is estimated as 0.

The progress for the current asset is divided by the number of assets to load to calculate how much to add to `completedProgress` to estimate the actual current progress.
```
float progressForCurrentAsset = 0;
if(tasksLength > 1) {
	progressForCurrentAsset = 0.9f / (tasksLength * toLoad);
} else if(tasksLength == 1) {
	AssetLoadingTask task = tasks.peek();
	if(task.dependenciesLoaded || task.dependenciesInjected) {
		progressForCurrentAsset = 0.9f / toLoad;
	}
}
return Math.min(1, completedProgress + progressForCurrentAsset);
```

NathanSweet:
It's not clear to me how this affects the progress bar and I don't have a setup to test it. I expect it will still go backwards?

If we want to loading dependencies to be reflected in the progress, we could have the "tasks" increase `toLoad` and `loaded`. However, doing that means the progress goes backward:

```
2 assets queued to load
loading 1 0%
loaded 1 50%
loading 2 50%
adds 1 dependency 33% <- backward!
loading dependency 33%
loaded dependency 66%
loaded 2 100%
```

Going backward is suboptimal for a progress bar, in that case you might as well display an indeterminate progress animation. I still contend the best solution is to know how many assets will be loaded beforehand.

lnr0626:
It doesn't go backwards. Given:
 assets (A, B, and C), A has 0 dependences, B has 2

It would look like:
```
2 assets queued to load
loading A 0%
loaded A 50%
loading B 50%
adds 2 dependencies
loading dependency 65%
loaded dependency 80%
loading dependency 80%
loaded dependency 95%
loaded B 100%
```

I'm not sure if the numbers are exactly correct, but that is the general trend that you would see. I can get some actual numbers for you later today if you would like that.

badlogic:
Closing this as it fizzled out and is an application level issue really.

Issue #2880, opened 2015-02-26
Interface Shape2D does not provide contains() method


haimat:
All classes implementing the `Shape2D` interace (like `Circle` or `Rectangle`) provide a method `contains()`, but the interface itself does not. However that would be really useful.

Would it be possible to add this method to the interface as well?

misterdustinface:
Dependency inversion.  Contains would have a subclass of Shape2D as a
parameter, so Shape2D would then be coupled to it's subclasses.

misterdustinface:
*its subclasses.  It is also possible to make the parameter a Shape2D, but
then within the contains methods you have to branch over the instanceof
keyword (if the implementation is in java).  This is generally considered
to be bad style, but it could work for this situation.

xoppa:
`Shape2D` is only a tagging interface, see #2178. There are many methods that it could have, but it doesn't. Everything is possible though. What exactly is the issue you're reporting? Please provide enough information to reproduce the issue. For example include the practical use-case for a feature you're requesting or preferable submit a [pull request](https://github.com/libgdx/libgdx/wiki/Contributing).

misterdustinface:
Oh yeah definitely a tagging interface.  This issue should be closed.

nooone:
> Dependency inversion. Contains would have a subclass of Shape2D as a parameter, so Shape2D would then be coupled to it's subclasses.

@misterdustinface The `contains(...)` methods actually take either `float x, float y` or `Vector2`. "Dependency inversion" is no problem here.

If all classes implementing `Shape2D` have this method, it would probably not be a bad idea to add it to the interface. IMO it makes sense that a 2D shape can test for a point.

However, there is one problem: `Polyline`. Polyline is the only one which does not have this method. IMO it is however questionable, whether a "polyline" should be a `Shape2D` at all. In my understanding a shape is "closed", has an area and can contain a point. An arbitrary line however does not satisfy this definition.

nooone:
I made a PR with the proposed changes. I think the additional methods fit.

misterdustinface:
But is there a reason that these subclasses of Shape2D are using Ad-Hoc polymorphism?  The parameters associated with the contains method are more variable than you have expressed (some include additional vectors OR shapes)

[...]

PR #2885, opened 2015-02-28
Added contains to Shape2D.

nooone:
This is the PR for https://github.com/libgdx/libgdx/issues/2880.

Polyline is no Shape2D anymore and `Shape2D` now has two methods `contains(float x, float y)` and `contains(Vector2 point)`.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Polygon.java
+++ b/gdx/src/com/badlogic/gdx/math/Polygon.java
@@ -194,6 +194,7 @@ public class Polygon implements Shape2D {
 	}
 
 	/** Returns whether an x, y pair is contained within the polygon. */
+	@Override
 	public boolean contains (float x, float y) {
 		final float[] vertices = getTransformedVertices();
 		final int numFloats = vertices.length;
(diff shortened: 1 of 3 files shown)


MobiDevelop:
I personally consider Polyline a shape, just not an enclosed shape.

nooone:
That means the next logical OOP step is to introduce `EnclosedShape`?

MobiDevelop:
Or, more simply, let Polyline return false from the contains method if it
can't contain a point.

nooone:
To keep it simple, I did it like that.

cypherdare:
Any reason Polygon can't extend Polyline? Looks to me like it simply enforces one more vertex, and there's a lot of copy-pasted code between the two.

xoppa:
Thanks @nooone! I think the shape discussion goes beyond the `contains` method. But until someone is going to pick that up, I guess it doesn't harm to move the `contains` method to an interface. Although it might be better when it's moved to its own interface, I don't think that's worth it. So, unless someone is against it, I'd say let's merge it and move on.

nooone:
I don't think that the guy who wanted this change still needs it, since it's 7 months ago by now.

Merge or close?

Issue #2585, opened 2014-11-14
glGetError() logging would help catch bugs


sehugg:
Some bugs like #2396 have come up where it's been difficult to pin down the cause, and it has ended up being improper issue of GL commands (e.g. outside of the render() loop).

It'd be helpful to log non-zero results from glGetError() to catch these conditions early. Maybe just log once, or each time the value changes? Maybe just one class shared among the backends, or maybe we can put it in the core code without touching the backends?

According to Apple, glGetError() may degrade performance so we would probably want a developer-mode switch.

PR #2889, opened 2015-03-01
GLDebugger

Darkyenus:
Adds GLDebugger, that works similarly to GLProfiler - on enable it wraps active Gdx.gl instances and checks for GL errors after each GL call. When error occurs, it will notify customisable listener that will, by default, log the error and where it happened. Inspired by #2585. Did test on desktop, should work everywhere, however I am not sure about gwt. Since it is opt-in, it won't break any existing code.


nooone:
Looks good. I went through it and checked if all methods are calling the correct delegate methods and that each one has a `check()` and I couldn't find any mistakes.

My only concern is how it will work together with `GLProfiler`. The following would end in `GLDebugger` still being enabled if I'm right:

    GLDebugger.enable();
    GLProfiler.enable();
    GLDebugger.disable();
    GLProfiler.disable();

gdx/src/com/badlogic/gdx/graphics/debugging/GLDebugger.java
```diff
@@ -0,0 +1,111 @@
...
+	/** This listener will be called when GLDebugger is enabled and any GL call sets error number (retrievable by glGetError call).
+	 *
+	 * Default is {@link GLDebugger#LOGGING_LISTENER}. */
+	public static GLDebuggerErrorListener listener = LOGGING_LISTENER;
+
+	/** Enables error checking by replacing the {@code GL20} and {@code GL30} instances with debugging ones. */
+	public static void enable () {
+		if (!isEnabled()) {
+			Gdx.gl30 = Gdx.gl30 == null ? null : new GL30Debugger(Gdx.gl30);
+			Gdx.gl20 = Gdx.gl30 != null ? Gdx.gl30 : new GL20Debugger(Gdx.gl20);
```
MobiDevelop: Should this null check be against Gdx.gl20?

MobiDevelop: Oh, maybe not. Just a bit weird.

Darkyenus:
@nooone One has to be careful when dealing with both GLProfiler and GLDebugger, but I assume that people will use them only sometimes during development, so it shouldn't be that hard. I don't think that making this part foolproof is worth the trouble and added code complexity. But maybe I could document it.

gdx/src/com/badlogic/gdx/graphics/debugging/GLDebugger.java
Darkyenus: enable() and disable() were carefully copied from GLProfiler, because it does basically the same thing and that code is already proven to work. There is one concern however, line [100 in file GLDebugger](https://github.com/libgdx/libgdx/pull/2889/files#diff-07cfe7d207b530cf767547cbbd911849R100) will probably fail when GL30 is enabled. That is because when GL30 is enabled, Gdx.gl30,gl20 and gl will all point to GL30Debugger. That class is not a subclass of GL20Debugger that is checked against here, so it will stay wrapped. This exact problem is present in GLProfiler.

Now that I'm looking at it even more carefully, it will fail even on the line before, when gl30 is enabled. Enabling is correct though. So unless I am missing something and the code is indeed correct, I'll patch it for both profiler and debugger. Should I do it in this PR or make a new one?

Darkyenus:
Can be automerged again

gdx/src/com/badlogic/gdx/graphics/debugging/GLDebugger.java
```diff
@@ -0,0 +1,116 @@
...
+		public void onError (int error) {
+			String place = null;
+			try {
+				final StackTraceElement[] stack = Thread.currentThread().getStackTrace();
+				for (int i = 0; i < stack.length; i++) {
+					if ("check".equals(stack[i].getMethodName())) {
+						if (i + 1 < stack.length) {
+							final StackTraceElement glMethod = stack[i + 1];
+							place = glMethod.getMethodName();
+						}
```
nooone: Have you tried this as well with GWT? I'm currently unable to check it myself, but is `Thread.currentThread().getStackTrace()` available on GWT? And if it is, won't the stacktrace probably be obfuscated ("check".equals won't work)?

Darkyenus: I did not test this on GWT. (But I did test it on other platforms.) The idea is, that this ErrorListener is intended for non-GWT platforms, where the capabilities are high, so the reporting can be powerful. I have never worked with GWT, so I don't know if/how would this be possible there. But there are always other or custom ErrorListeners for that. So unless this breaks GWT build, it should be sufficient (and due to fallback couple lines below, should still work).

Also, this will never work flawlessly for obfuscated code, but that is the price of obfuscation. And if you can deobfuscate exception stack traces, fallback has got you covered.

MobiDevelop: This would break the GWT build, but this class could be emulated on GWT to remove offending parts.

Darkyenus: I am not sure what is the best way to go about this. I have three ideas:
- Emulate whole GLDebugger for GWT, without that one offending line.
  - Most API friendly, but creates a lot of maintenance overhead
- Move LOGGING_LISTENER to a separate class and exclude that from GWT
  - Low maintenance overhead, but not that pretty, I don't like this much
- Remove the "offending gl-call finding" functionality altogether and use only stack trace from `new Exception`, as fallback does now

I am not sure which to choose. Third one is simplest, but I quite like that feature, so I don't want to remove it. (But I will remove it if people think that it is useless) There is also one more alternative, combination of all of them, and that is moving all listeners to a separate class (perhaps as static fields of GLDebuggerErrorListener) and emulating that, such that in GWT version, only what-is-now-fallback is used.

Any thoughts?

nooone: Not option 1 imho. Emulating a whole class just for a small extra functionality on GWT is too much overhead.

I like moving out the part that won't work on GWT (option 2), but I'd also be fine with a more simpler version of GLDebugger, which won't result in any GWT problems at all (option 3).

Darkyenus:
I have moved GLDebuggerErrorListener to a separate file which is emulated for GWT. I have even managed to get the original functionality somewhat working in GWT (when running with readable stack traces, like in super dev mode). Everything should be fine now.

nooone:
I think it should now be decided whether this will be merged, or not. And if it will, @Darkyenus has to fix the merge conflicts.

Darkyenus:
Rebased

nooone:
@badlogic "I'd merge any PR in that regard." (https://github.com/libgdx/libgdx/issues/2585)

xoppa:
Looks good. But why not merge it with GLProfiler? Do we really need these two separated? That would also solve the issue with combining them.

Darkyenus:
I'll look into doing that while keeping the API sane and preferably backwards compatible.

Darkyenus:
I have squashed all previous commits and then moved GLDebugger functionality to GLProfiler. Should be ready.

gdx/src/com/badlogic/gdx/graphics/profiling/GL20Profiler.java
```diff
@@ -156,41 +192,55 @@ public void glDrawElements (int mode, int count, int type, Buffer indices) {
...
 		gl20.glGenTextures(n, textures);
+		check();
+	}
+
+	@Override
+	public int glGenTexture () {
+		int result = gl20.glGenTexture();
+		check();
+		return result;
 	}
```
nooone: Where does this new method come from? It's missing `calls++`.

gdx/src/com/badlogic/gdx/graphics/profiling/GL20Profiler.java
```diff
@@ -278,144 +342,190 @@ public void glTexSubImage2D (int target, int level, int xoffset, int yoffset, in
...
 		calls++;
-		return gl20.glCreateShader(type);
+		int result = gl20.glCreateShader(type);
+		check();
+		return result;
+	}
+
+	@Override
+	public void glDeleteBuffer (int buffer) {
+		gl20.glDeleteBuffer(buffer);
```
nooone: calls++ is also missing here.

gdx/src/com/badlogic/gdx/graphics/profiling/GL20Profiler.java
```diff
@@ -278,144 +342,190 @@ public void glTexSubImage2D (int target, int level, int xoffset, int yoffset, in
...
 	@Override
 	public void glDeleteBuffers (int n, IntBuffer buffers) {
 		calls++;
 		gl20.glDeleteBuffers(n, buffers);
+		check();
+	}
+
+	@Override
+	public void glDeleteFramebuffer (int framebuffer) {
+		gl20.glDeleteFramebuffer(framebuffer);
```
nooone: No calls++

nooone:
Besides those few missing calls, this looks good to me.

Darkyenus:
Oops, sorry, fixed

xoppa:
Thanks! Perhaps you can expand the [wiki page](https://github.com/libgdx/libgdx/wiki/Profiling) a bit?

nooone:
5.5 months later :D

Darkyenus:
Thanks! Wiki updated.

Issue #846, opened 2013-10-19
Quaternion.transform(Vector3) is not thread safe


theagentd:
Quaternion.transform(Vector3) is not thread safe since it uses two static variables, tmp1 and tmp2:

```
private static Quaternion tmp1 = new Quaternion(0, 0, 0, 0);
private static Quaternion tmp2 = new Quaternion(0, 0, 0, 0);

public Vector3 transform (Vector3 v) {
    tmp2.set(this);
    tmp2.conjugate();
    tmp2.mulLeft(tmp1.set(v.x, v.y, v.z, 0)).mulLeft(this);

    v.x = tmp2.x;
    v.y = tmp2.y;
    v.z = tmp2.z;
    return v;
}
```

This is very bad practice and I discovered this issue the hard way ("Why are my skeleton animated objects twitching like hell?!"). A fix would be to "hard code" the procedure, which I also believe is slightly faster. The solution I'm using right now can be found here:

http://pastie.org/8414910

It may need some very light reworking to fit the style of LibGDX.

JBraig:
There doesn't seem to have been any progress regarding this issue in the last seven years, but I think the topic deserves another look after all this time. I recently stumbled over this issue by accident, and while I know that libgdx isn't thread safe unless explicitly specified, this is a level above what I'd usually think of when thinking of thread safety - I only discovered this problem when independent Matrix4-instances in different threads started interfering with each other. Some functions like Matrix4.mulLeft() should be fairly straightforward to unroll, for others I'm not so sure. According to the [old, now archived forum post,](https://web.archive.org/web/20150920150715/http://www.badlogicgames.com/forum/viewtopic.php?f=23&t=7814) this was done to relieve GC pressure, and there's some mention of modifying the API so users can pass their own vectors, matrices or whatever instead of using the temp ones, yet this seems to never have been implemented. Any specific reason for that?

theagentd:
I use JOML now, which has better performance, a more consistent API design and no threading issues. For the use case of skeleton animation, switching to JOML resulted in over triple the performance.

PR #2892, opened 2015-03-02
Remove of static temporaries in math classes

code-disaster:
With this PR I want to revive an issue previously discussed in e.g. #846. It's one step towards thread safety in the math classes.

There are still some static temporaries left: Vector3.tmpMat and Matrix4.quat/quat2. The long-lived deprecated Matrix4.tmp could now be removed.

I've run a couple of tests to verify my changes, but as far as I've seen not all modified functions are covered by them. Also, I've only run the tests on desktop systems. That said, a code review and **some tests on Android/iOS** would be wise. ;) Also, feel free to try the changes with your project if you are heavily using Matrix4/Quaternion.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Matrix3.java
+++ b/gdx/src/com/badlogic/gdx/math/Matrix3.java
@@ -509,24 +509,56 @@ public class Matrix3 implements Serializable {
 	}
 
 	/** Multiplies matrix a with matrix b in the following manner:
+	 *
+	 * <pre>
+	 * mul(A, B) => A := AB
+	 * </pre>
+	 * @param mata The float array representing the first matrix. Must have at least 9 elements.
+	 * @param matb The float array representing the second matrix. Must have at least 9 elements. */
+	private static void mul (float[] mata, float[] matb) {
+		float v00 = mata[M00] * matb[M00] + mata[M01] * matb[M10] + mata[M02] * matb[M20];
+		float v01 = mata[M00] * matb[M01] + mata[M01] * matb[M11] + mata[M02] * matb[M21];
+		float v02 = mata[M00] * matb[M02] + mata[M01] * matb[M12] + mata[M02] * matb[M22];
+
+		float v10 = mata[M10] * matb[M00] + mata[M11] * matb[M10] + mata[M12] * matb[M20];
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/math/Matrix3.java
```diff
@@ -566,19 +514,19 @@ public Matrix3 transpose () {
...
+		float v01 = mata[M00] * matb[3] + mata[M01] * matb[4] + mata[M02] * matb[5];
+		float v02 = mata[M00] * matb[6] + mata[M01] * matb[7] + mata[M02] * matb[8];
+
+		float v10 = mata[M10] * matb[0] + mata[M11] * matb[1] + mata[M12] * matb[2];
+		float v11 = mata[M10] * matb[3] + mata[M11] * matb[4] + mata[M12] * matb[5];
+		float v12 = mata[M10] * matb[6] + mata[M11] * matb[7] + mata[M12] * matb[8];
+
+		float v20 = mata[M20] * matb[0] + mata[M21] * matb[1] + mata[M22] * matb[2];
+		float v21 = mata[M20] * matb[3] + mata[M21] * matb[4] + mata[M22] * matb[5];
+		float v22 = mata[M20] * matb[6] + mata[M21] * matb[7] + mata[M22] * matb[8];
```
nooone: Why were the constants partly be replaced?

xoppa:
Using varargs instead of array might cause an additional allocation, which we really want to avoid in these frequently called methods. Better move the allocation to the caller (e.g. `mul(val, new float[] { a, b, etc.})`) or add an additional signature which takes all the float as argument (e.g. `mul(float[], float, float, float, float, etc.)`)

code-disaster:
@xoppa you are right about the allocation, I wasn't aware of that. Indeed the byte code shows newarray instructions. This of course limits the usefulness of my changes. It can be easily fixed for Matrix3, but for Matrix4 it probably needs a second native mul() function to avoid the allocation.

xoppa:
Yeah, I'm not sure this actually solves an issue now. I guess this PR is about whether we want to complicate code for thread safety.

MobiDevelop:
I don't want to complicate code for thread safety.

code-disaster:
I'm ok with that. Let me close this PR then.

Issue #2877, opened 2015-02-24
TimSort verification from run length


cdomville:
See https://bugs.openjdk.java.net/browse/JDK-8072909 and http://envisage-project.eu/proving-android-java-and-python-sorting-algorithm-is-broken-and-how-to-fix-it/ for more details, but the OpenJDK implementation has changed TimSort, as noted at http://hg.openjdk.java.net/jdk9/jdk9/jdk/rev/e276aa5b8a4b to fix an ArrayIndexOutOfBoundsException with worst case long arrays that relates to the number of invariants. I know this is minor, but so we're consistent with the JVM, this should probably be updated for consistency reasons as well.

NathanSweet:
Very interesting! Any chance you can send a PR? Keep in mind that OpenJDK is GPL, so it must not be used as a reference to copy the fix or any other code from. Code can be written yourself of course, come from Android (Apache 2, same as libgdx), or other compatible sources.

cdomville:
The Java TimSort bug repo linked in the blog post has Apache licensing, so I'm assuming the quoted fix in the blog post (to mergeCollapse()) is licensed in the same manner because of this change, so that's what I'll implement.

PR #2895, opened 2015-03-03
Fix TimSort runLength

cdomville:
This pull request adds an additional check to search the invariant for the last 4 runs in the
runLength array to shore up an invariant that was not true for sufficiently large array sizes.

Refer to #2877 to get the gist of the reasoning behind the change. The source used to derive the implementation uses an Apache 2 licence for the test suite (see https://github.com/abstools/java-timsort-bug/blob/master/LICENSE), so, from what I can see, the proposed fix is also licensed in the same manner by the same party (due to the different places this invariant would need to be fixed to be universal, such as AOSP).


NathanSweet:
Thanks for figuring this out!

cdomville:
No bother, been meaning to contribute more to the project, just been really busy. :)

PR #2897, opened 2015-03-03
Lwjgl30 override EXTFramebufferObject usage

code-disaster:
I'm doing some experiments with multi-target rendering (G-Buffers), for now using the gl32core branch by @nelsonsilva. I'm doing a separate PR because the changes are independent of this pending work.

- Overrides all uses of EXTFramebufferObject with the GL3 core functions.
- The second, small change is a new Texture() constructor which takes an existing GL handle. I'm using this to easily pass G-Buffer textures to Batch.draw() for debugging.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Texture.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Texture.java
@@ -124,6 +124,14 @@ public class Texture extends GLTexture {
 		load(data);
 		if (data.isManaged()) addManagedTexture(Gdx.app, this);
 	}
+
+	public Texture (TextureData data, int glHandle) {
+		super(GL20.GL_TEXTURE_2D, glHandle);
+		load(data);
+		if (data.isManaged()) {
+			addManagedTexture(Gdx.app, this);
+		}
+	}
 	
 	public void load (TextureData data) {
 		if (this.data != null && data.isManaged() != this.data.isManaged())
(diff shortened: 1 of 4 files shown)


code-disaster:
Added another small change to allow using a custom shader with ShapeRenderer.

MobiDevelop:
It's dangerous territory putting multiple unrelated changes into a single pull request.

That said, this looks OK to me.

gdx/src/com/badlogic/gdx/graphics/Texture.java
```diff
@@ -124,6 +124,14 @@ public Texture (TextureData data) {
 		load(data);
 		if (data.isManaged()) addManagedTexture(Gdx.app, this);
 	}
+
+	public Texture (TextureData data, int glHandle) {
+		super(GL20.GL_TEXTURE_2D, glHandle);
+		load(data);
+		if (data.isManaged()) {
+			addManagedTexture(Gdx.app, this);
```
xoppa: Adding a manged texture will create a new handle on context loss, which is inconsistent if you let the user specify its handle.

code-disaster: What'd be a good way to handle this case? Always consider the texture unmanaged and throw an error if data.isManaged() returns true?

xoppa: Depends on the use-case. For LwjglGL30 (probably any device that supports gles3) it can be safely ignored afaik, but this api is not specifc to lwjgl or gles3. Perhaps make this constructor protected and add a javadoc warning to override the reload method as well.

code-disaster: Ok, in this case it's probably better to not change the API at all, because the user would have to sub-class Texture anyway. Let's close this one and I'll open a new PR without those changes.

PR #2898, opened 2015-03-03
Lwjgl30 override EXTFramebufferObject usage

code-disaster:
Separated from #2897. Overrides all uses of EXTFramebufferObject with the GL3 core functions.


xoppa:
Thanks!

PR #2899, opened 2015-03-03
ShapeRenderer: allow use of custom shader

code-disaster:
Already implemented by ImmediateModeRenderer20, just not exposed through the ShapeRenderer API.


CHANGES
```diff
@@ -4,6 +4,7 @@
 - API Change: Exposed method ControllerManager#clearListeners on Controllers class
 - API Change: Net#openURI now returns a boolean to indicate whether the uri was actually opened.
 - API Change: DefaultShader now always combines material and environment attributes
+- API Change: Added ShapeRenderer constructor to pass a custom shader program to ImmediateModeRenderer20.
```
NathanSweet: Everybody uses "API change", but probably would be better to use that (or something else) for changes that affect existing code.

PR #2902, opened 2015-03-04
Fix stuff in SuperKoalio

UgoRaffaele:
* Added multiple touches detection on mobile;
* Expanded "jump" area (starting from 0.5 of X axis, previously 0.75);
* Fixed MapTiles Layers references (now "walls", previously 1) and ClassCastException of TiledMapTileLayer;


tests/gdx-tests-android/assets/data/maps/tiled/super-koalio/level1.tmx
```diff
@@ -16,7 +16,7 @@
...
   </data>
  </layer>
- <imagelayer name="koalio-color" x="352" y="136" opacity="0.5">
+ <imagelayer name="koalio-color" x="352" y="136" opacity="0.5" visible="0">
```
siondream: These are needed for the ImageLayer test, maybe we can make that test use a copy of this map with the visible flags set to true.

UgoRaffaele: Sorry, I didn't investigate properly.
Are the other changes fine? I could submit the revised commit in this case.

PR #2907, opened 2015-03-05
Add spotlight support in Environment

realitix:
Hello,
This is my first Pull Request.
This Pull Request add support of spotlight in environment. It doesn't modify the shader so spotlight can't be used with default shader. But it allows to use spotlight in custom shader.


realitix:
I had no answer until now so if something is wrong don't hesitate to tell me. This is the first commit of a big list.

xoppa:
Thanks! See also #1936. I guess it doesn't harm to have this class for the time being. Will fix it up a bit after merging.

PR #2913, opened 2015-03-06
changing the loopduration

JoEschner:
Whoops, this is very important to correctly determine the frame.


MobiDevelop:
Also didn't compile.

PR #2916, opened 2015-03-06
Added FilehandleResolver for classpaths

ClaudiuBele:
Case: The resolver will be useful for libraries that use Libgdx as a backend and have default assets packed in the library in case a desired asset can't be found to draw/play/etc. 

The InternalFileHandleResolver does not suffice, as it prioritizes looking in the internal storage first. From the documentation: "If a file can't be found on the internal storage, the file module falls back to searching the file on the classpath". In my case ClasspathFileHandleResolver will be used because we do not want to potentially use assets with the same name that the user has placed in the assets folder. 

Example: Have a "Missing texture" texture drawn when developer using my library want to draw a texture from a filehandle that does not exist. The "Missing texture" could in this case be located at MissingTexture.png in the classpath. If the developer using the library already has a MissingTexture.png file in internal storage, that one would be used instead if we resolve the file using InternalFileHandleResolver. We don't want the developer using my library have to know by heart file names that have to be avoided.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@
 - Added ShapeRenderer constructor to pass a custom shader program to ImmediateModeRenderer20.
 - API Change: Group#toString now returns actor hierarchy. Group#print is gone.
 - Added SpotLight class, see https://github.com/libgdx/libgdx/pull/2907
+- Added support for resolving file handles using classpaths (ClasspathFileHandleResolver)
 
 [1.5.4]
 - Added support for image layers in Tiled maps (TiledMapImageLayer)
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -172,3 +172,4 @@ CodePoKE https://github.com/codepoke
 Konijnendijk https://github.com/Konijnendijk
 nooone https://github.com/nooone
 kotcrab https://github.com/kotcrab
+ClaudiuBele https://github.com/ClaudiuBele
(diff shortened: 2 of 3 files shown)


CHANGES
```diff
@@ -5,6 +5,7 @@
 - API Change: Net#openURI now returns a boolean to indicate whether the uri was actually opened.
 - API Change: DefaultShader now always combines material and environment attributes
 - API Change: Added ShapeRenderer constructor to pass a custom shader program to ImmediateModeRenderer20.
+- API Change: Added ClasspathFileHandleResolver for resolving assets via Gdx.files.classpath(string).
```
MobiDevelop: Probably better to call the kinds of things API Addition instead of API Change. That said, I'm not sure  it is even necessary to specifically call out this in CHANGES - I guess it doesn't hurt though.

ClaudiuBele: You might be right, there's already a line close to the start of the1.5.5 version notes that starts with "Added". Might as well drop the "API Change: ". What do you think?

xoppa: What about using "API Change" only for changes to the api (something that might break existing code)?

MobiDevelop: Yes, @NathanSweet also suggested that recently. Seems like a reasonable approach that will make it easier to grep those particular changes at a glance.

ClaudiuBele: Would this fit the current updates? 
"- Added support for resolving file handles using classpaths (ClasspathFileHandleResolver)"

Tom-Ski:
Yup! Could you squash these commits into one, then we can merge.

ClaudiuBele:
Done, there should be only one commit now.

siondream:
Thanks!

PR #2917, opened 2015-03-07
Added id as a property to MapObject for Issue #2873

### LINKED ISSUE #2873: Match Object Ids between Tiled and libgd
In the latest Tiled Version(11.0) unique Ids got added to objects.
These Ids however are not the ones you get when you use
"map.getObjects().get(id)"
For example
map.getObjects().get(1)
returns the object with the unique tiled Id of 35 in my case.

It would be nice if these ids would be consitent between libgdx and tiled

RenrokB:
Added id as a property since Tiled recently added unique IDs to objects. Closes #2873.


MobiDevelop:
This will crash on tmx files created with an older version of Tiled. Please use the variant of `getIntAttribute` which takes a default value. I'd also prefer not adding it to the properties if the attribute does not exist.

RenrokB:
Should be better now.

PR #2925, opened 2015-03-09
Fixed bug in EarClippingTriangulator where vertex offset is ignored when computing triangles

afomins:
Bug: When vertices array describes multiple polygons then only first polygon is triangulated correctly; offset of all following polygons is ignored and function returns random triangles.

Fix:
1) Added offset when initializing indices
2) Added offset when computing winding


NathanSweet:
Thanks! Working on this thing is always brutal. :)

PR #2933, opened 2015-03-13
Add FrameBufferCubemap class.

realitix:
This class allows to render in a cubemap framebuffer easily.

It's very useful for point light shadowing.
It works very well and it's documented.


realitix:
I have to rewrite this by update Framebuffer.

realitix:
Ok I updated it. It extends FrameBuffer now.

xoppa:
Thanks! I don't think that a framebuffer class should be controlling (or be aware of) the camera. Should be more flexible and just a few loc to leave this up to the caller. Having a set/get/nextSide should be enough. Besides that it looks good to me.

Btw, it is very unlikely that the caller actually ever needs a full blown Cubemap with 6 Pixmaps, thats just wasting memory. The same goes for the FrameBuffer class, of which many users probably never actually will need the Pixmap.

realitix:
I understand what you said but this part of the code:

```
case NegativeX:
+ camera.up.set(0, -1, 0);
+ camera.direction.set(-1, 0, 0);
+ break;
+ case NegativeY:
+ camera.up.set(0, 0, -1);
+ camera.direction.set(0, -1, 0);
+ break;
+ case NegativeZ:
+ camera.up.set(0, -1, 0);
+ camera.direction.set(0, 0, -1);
+ break;
+ case PositiveX:
+ camera.up.set(0, -1, 0);
+ camera.direction.set(1, 0, 0);
+ break;
+ case PositiveY:
+ camera.up.set(0, 0, 1);
+ camera.direction.set(0, 1, 0);
+ break;
+ case PositiveZ:
+ camera.up.set(0, -1, 0);
+ camera.direction.set(0, 0, 1);
+ break;
+ default:
+ break;
+ }

```

has to be in the framebuffer or somewhere because it's required to get a good cubemap. Maybe a FramebufferCubemapHelper. I don't know where you would see this part of code but it has to be somewhere.

xoppa:
Although not *required to get a good cubemap* (it's perfectly fine to use whatever projection you need), I agree that mapping a cubemap side to an orientation is useful. However, that isn't the responsibility of (nor should it be restricted to) the framebuffer. The framebuffer is a render target, it should not care about projection.

The orientation is a property of the side (hence the naming). So I guess that would be a good place to put such value. So the caller could do for example:

    cam.up.set(side.up);
    cam.direction.set(side.direction);
    cam.update();

realitix:
Ok, smart.

realitix:
Xoppa, I updated the framebufferCubemap. I hope this one is the good ;-)

gdx/src/com/badlogic/gdx/graphics/Cubemap.java
```diff
@@ -43,32 +44,49 @@
...
+			this.direction = new Vector3(directionX, directionY, directionZ);
 		}
 
 		/** @return The OpenGL target (used for glTexImage2D) of the side. */
 		public int getGLEnum () {
 			return glEnum;
 		}
+
+		/** @return The up vector of the side. */
+		public Vector3 getUp () {
```
xoppa: Could you change this and the direction getter to:

    public Vector3 getUp (Vector3 out) { return out.set(up); }

xoppa:
Can you explain (preferable with a link to the spec) why you're using negative Y for the up vector? I have a feeling that you be compensating for the flipping v issue.

I'm not quite sure about casting the GLTexture to the Texture. This kinda exposes an API issue, e.g. one could call `FrameBufferCubemap#getColorBufferTexture()` (which is perfectly valid api-wise) and get a cast exception. Perhaps we should use generics instead. On the other hand that would complicate the API and might break existing code.

I guess it makes sense to add a generic GLFramebuffer class (just like there's a GLTexture) which represents an OpenGL framebuffer without specific implementation. And then let FrameBuffer (`GLFrameBuffer<Texture>`) and CubemapFrameBuffer (`GLFrameBuffer<Cubemap>`) be specific implementations of that class.

realitix:
OK it's a good idea, I will commit a first version with this concept.
Just to be sure, I create the GLFrameBuffer<T> class and then I extend it:
FrameBuffer extends GLFrameBuffer<Texture>
FrameBufferCubemap extends GLFrameBuffer<Cubemap>

Is this what you mean ?

realitix:
FrameBuffer extends GLFrameBuffer \<Texture\>
FrameBufferCubemap extends GLFrameBuffer \<Cubemap\>

Markdown escape....

xoppa:
Sorry for the delay. Yes, that's indeed what I meant.

realitix:
Ok Xoppa, it's done.
I wait for your comment.
I tested it it works well.
I us Negative Y because Positive Y does not work but I can't explain the reason.

xoppa:
Looks good. The negative Y needs some substantiation though. Can you add your test? I will then pull in locally to test and merge.

realitix:
Ok Xoppa, I added tests.

xoppa:
Looks like the test doesn't run (the Cubemap is never constructed), can you update it?

xoppa:
The test throws at GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT, any clue what might be wrong? https://github.com/xoppa/libgdx/tree/pr2933

realitix:
I experienced this before. Try adding the depth buffer by passing true in the framebuffercubemap constructor.

realitix:
I updated it because there were new conflict with Cubemap

realitix:
I updated tests to create depth buffer, it may be the source of your problem. Can you try again ?

xoppa:
Yes using a depth buffer it works.

realitix:
It's a hardware bug. On my computer it works without depth buffer

xoppa:
Sorry for the delay. Need some more time to check some things, but overall I think its good. No need to update the PR after changes, I keep it updated in a separate branch.

realitix:
Ok Xoppa.
New conflicts but I let you manage it.

realitix:
Hello Xoppa,
Did you have time to watch at this FramebufferCubemap ?

xoppa:
Sorry for the delay. It's merged now.

realitix:
Great thanks !

PR #2940, opened 2015-03-15
[gdx-jnigen] ARM compatibility

Robbiedobbie:
I've been working on a project which uses gdx-jnigen for generating and loading native libraries. The project runs on any linux, and is most beneficial on a Raspberry Pi, which has the ARM architecture. 
Currently it is not possible to build a version for 32bit, 64bit and arm architectures at the same time. Therefore, I added some options to the buildscript generator, which gives more freedom, also for future additions.
I also added the ability to detect an ARM platform, and load the appropriate native library.

This will be usefull in all sorts of projects, and will also be helpfull if one would decide to port libgdx to another ARM platform (Iike the Rpi) later.


deathbeam:
+1

xoppa:
Thanks!

PR #2941, opened 2015-03-15
Fix lwjgl opengl - opengl es version check

xoppa:
Currently the lwjgl version check might cause some issues (e.g. when the version vendor string contains the string "2.1"). For GLES3 the OpenGL 3.2 core profile [was used](https://github.com/libgdx/libgdx/commit/08711a2152eda79f20c2db6064b4f1b444b6ea0c), while afaik GLES3 requires at least 4.3. Also, currently GLES3 is enabled even if not requested. This PR tries to fix these issues along with some minor changes. E.g. I've moved all the version checking into separate methods for readability and added links to the resources on which the implementation is based. Also moved the extensions to an array instead of long string. Since I was at it, I also added a `compatibleWithGLES2` method. While currently not used, this method could be used to implement the currently missing methods for true GLES 2 compatibility. I added an argument to the `createDisplayPixelFormat` method so it can fallback without modifying the config in case 4.3+ is not available and useGL30 is requested.

Problem is that I can't test it, other than on my PC which supports OpenGL 4.5.


badlogic:
IIRC the 3.2 core profile is required on Mac OS X

xoppa:
I guess it doesn't harm to merge this, even with only limited testing, since the version check is currently broken anyway. If anyone agrees then feel free to hit the merge button.

code-disaster:
Tested on OS X 10.10, MPB late 2013. Version 4.1 is the maximum supported ("4.1 NVIDIA-10.2.1 310.41.15f01").

MobiDevelop:
So, where did we land on this? Are we good to go here?

code-disaster:
I have this GL30 version in use since a few weeks now on OS X and Windows on multiple machines, but for OS X I had to adjust the requested profile down to 4.1 on my MBP.

@xoppa added a comment linking to http://en.wikipedia.org/wiki/OpenGL_ES#OpenGL_ES_3.0 which states "OpenGL 4.3 provides full compatibility with OpenGL ES 3.0". I think the question is what "full compatibility" really means, and if we need that.

Maybe it's an option to add a minimum GL profile version to LwjglApplicationConfiguration?

xoppa:
I've made the context version configurable. Because the current LwjglGL30 implementation does use methods of the GL43 class (for fbo invalidate), I made it default to 4.3. Users can now override it to support lower versions as well (obviously at the cost of those methods not being supported).

code-disaster:
Looking good!

Fun fact: initialisation now passes on my MacBookPro with the default 4.3 context, though GL_VERSION string says "4.1".

xoppa:
> This extension is not supported on MacOS X. However, in order to enable the GL 3.2 context on MacOS X 10.7 or newer, an instance of this class must be passed to LWJGL. The only valid configuration is ContextAttribs(3, 2, CONTEXT_CORE_PROFILE_BIT_ARB), anything else will be ignored.
http://legacy.lwjgl.org/javadoc/org/lwjgl/opengl/ContextAttribs.html
Hmmm, perhaps we should manually check the version.

code-disaster:
We are using this branch for a few months now, on different OS X and Windows machines, with gles30ContextMajorVersion=3 and gles30ContextMinorVersion=3. Should be safe to merge from my point of view.

xoppa:
@code-disaster thanks for testing. I've changed it to be 3.2 default to be safe for os-x and pulled in latest changes. Would you mind to verify this? If it still works for you then it should be good to pull.

code-disaster:
Verified, still working. Only a minor, cosmetic issue: two System.out.println() in createDisplayPixelFormat() which write "4.3+" to the console.

xoppa:
Thanks for testing, I'll fix the println after merging as well as add a line to the CHANGES.

PR #2945, opened 2015-03-16
libgdx-utils as 3rd party extension

dermetfan:
I'd like to add my little support library as a 3rd party extension to the gdx-setup app. Users already asked for this in the past before the XML configuration existed.

> Does your project aim to extend LibGDX with a specific goal in mind?

The goal is to offer additional features (like pre-made scene2d.ui widgets, the `TmxMapWriter`, `AnnotationAssetManager` or the [`PolygonRegionLoader`](https://github.com/libgdx/libgdx/pull/1602) which was merged into libGDX) and utility/helper classes to make game dev even more convenient. I'd let that count as specific enough even though it's rather broad.

> Is your project useful to LibGDX users?

I think so, it's catered towards them.

> Is your project well established?

Depends on what is well established enough. The repository currently has 9 forks and 46 watchers on Bitbucket. The only statistics I have are those from sonatype:
chart

> Your project is well maintained

There was not a single month without commits since I started this project in September 2013 and it's usually compatible with the latest libGDX release.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
@@ -24,6 +24,9 @@
         <version>0.10.0</version>
         <compatibility>1.5.5</compatibility>
         <website>http://dermetfan.net/libgdx-utils.php</website>
+        <gwtInherits>
+		<inherit>libgdx-utils</inherit>
+	</gwtInherits>
         <projects>
 	        <core>
 		        <dependency>net.dermetfan.libgdx-utils:libgdx-utils</dependency>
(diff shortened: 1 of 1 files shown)


MobiDevelop:
I'm okay with adding your library to the setup, will let @Tom-Ski verify it all looks good since he's the setup-keeper.

Tom-Ski:
Yup will do. Just need to find time to test the html extension stuff. Ill try to do that tomorrow. This PR will need adapting after that is merged though.

Tom-Ski:
Ok @dermetfan if you could fix up your PR, and include your GWT includes also, then rebase to one commit and I'll merge.

dermetfan:
Sorry @Tom-Ski, what does including my GWT includes mean? AFAIK _extensions.xml_ can't declare modules to inherit in _GdxDefinition(SuperDev).gwt.xml_.

kotcrab:
After merging #2934 gwt inherits are now supported, see [extensions.xml](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml#L28-L30).

dermetfan:
Cool, thanks @kotcrab. see #2961

Issue #2886, opened 2015-03-01
BitmapFont markup crashes if String is just "["


AzmiTerato:
Im using latest nightlies,

BitmapFont.drawWrappedText crashes if I just pass "[" as the string to draw.

Exception in thread "LWJGL Application" java.lang.StringIndexOutOfBoundsException: String index out of range: 2
	at java.lang.String.charAt(Unknown Source)
	at com.badlogic.gdx.graphics.g2d.TextMarkup.parseColorTag(TextMarkup.java:75)
	at com.badlogic.gdx.graphics.g2d.BitmapFontCache.addToCache(BitmapFontCache.java:421)
	at com.badlogic.gdx.graphics.g2d.BitmapFontCache.addWrappedText(BitmapFontCache.java:737)
	at com.badlogic.gdx.graphics.g2d.BitmapFont.drawWrapped(BitmapFont.java:307)
.....

davebaol:
It works for me. Please, provide a [SSCCE](http://sscce.org/) reproducing your issue.

MobiDevelop:
It does crash for me... Here's a modified BitmapFontTest.

```java
package com.badlogic.gdx.tests;

import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.g2d.BitmapFont;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;
import com.badlogic.gdx.tests.utils.GdxTest;

public class BitmapFontTest extends GdxTest {
	private SpriteBatch spriteBatch;
	private BitmapFont font;

	@Override
	public void create () {
		spriteBatch = new SpriteBatch();
		font = new BitmapFont(Gdx.files.internal("data/verdana39.fnt"), false);
		font.setMarkupEnabled(true);
		font.setColor(Color.RED);
	}

	@Override
	public void render () {
		Gdx.gl.glClearColor(1, 1, 1, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
		
		spriteBatch.begin();
		font.drawWrapped(spriteBatch, "[", 100, 100, 100);
		spriteBatch.end();
		
	}

	@Override
	public void dispose () {
		font.dispose();
		spriteBatch.dispose();
	}
}
```

[...]

Issue #1589, opened 2014-03-30
Support for complex font rendering (Chinese, Arabic, ...)


adrianmay:
For languages like Chinese, it's not practical for an app to deliver its own font. Therefore there's a desire for libgdx to be able to use fonts delivered by the platform. 

Some discussion already occurred here:
[Roll-your-own Chinese font?](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=13203)
and here:
[Development Forum: Supporting Asian fonts in BitmapFont](http://www.badlogicgames.com/forum/viewtopic.php?f=23&t=13031)

Earlier attempts at the same challenge have run into difficulties but there doesn't seem to be an exhaustive list of those difficulties. Perhaps that should be compiled here before attempting to develop a new plan.

bill-lin:
+1 to using system font
I am developing an app in Chinese and it is very painful to handle chinese input and cross platform.

tdlrobin:
+1 to using system font

abdlquadri:
@NathanSweet I edited Hiero to handle RTL. Details here. Tested Arabic though. http://www.badlogicgames.com/forum/viewtopic.php?f=15&t=18592

abdlquadri:
@sridharsundaram I will be interested in how you implemented that. Also does if work for the other platforms?

abdlquadri:
Hi All,

Just wondering what progress has been made on using System Font and how I can contribute to this. Any pointer to where to start would be great.

Regards.

sridharsundaram:
I had to modify BitmapFontCache in addToCache

....
    if (isComplexScriptLayoutEnabled()) {
      List<Glyph> glyphs = getGlyphsAfterShaping(str, start, end, data);
      requireGlyphSequence(glyphs);
      for (Glyph g: glyphs) {
        addGlyph(g, x + g.xoffset * scaleX,
            y + g.yoffset * scaleY,
            g.width * scaleX,
            g.height * scaleY);
        x += g.xadvance * scaleX;
      }
      return x - startX;
    }
....

  // Here we make a JNI call to get the complex script shaping done
  private List<Glyph> getGlyphsAfterShaping(CharSequence str, int start,
      int end, BitmapFontData data) {
    List<Glyph> glyphs = new ArrayList<Glyph>();
    int[] glyphIndices = complexScriptLayout.getGlyphsAfterShaping(str,
start, end);
    for (int glyphIndex: glyphIndices) {
[...]

PR #2955, opened 2015-03-19
Separation of glyphs and characters for BitmapFontCache

NathanSweet:
This PR refactors how BitmapFontCache works. Previously it was assumed each character in a string mapped to a glyph for display. As discussed in #1589, this is not the case for some types of font rendering. This PR has two main goals: 1) to separate characters and glyphs, 2) to make parsing the string and preparing for display cleaner. The new approach works like this:

1. GlyphLayout, a new class, is given a string of characters. It first breaks the strings into "runs". A run is a string of characters all the same color and without newlines. The reason to break the text into runs is that each run can be "shaped" individually, which means to convert the characters into glyphs for display. When breaking the string into runs, GlyphLayout parses color markup tags and stores the color for each run.
1. Next, the runs are shaped using a new BitmapFontData method, `getGlyphs`. This takes the characters for the run and generates both a list of glyphs and a list of positions for each glyph. The default implementation just looks up the glyph for the character as has always been done. A more complex implementation could use [HarfBuzz](http://www.freedesktop.org/wiki/Software/HarfBuzz/) or another lib to do the shaping.
1. Next, the shaped runs are laid out. If wrapping is enabled and a run's glyphs exceed the wrap width, then glyphs are split for line wrapping using another new BitmapFontData method, `getWrapIndex`. The run is ended, the remaining glyphs are moved to a new run on the next line, and the layout process repeats. I'm not sure if line wrapping glyphs is the best solution, but at least where the wrap happens can be customized. Once all runs are laid out, if center or right alignment is requested then run positions are shifted appropriately.
1. At this point GlyphLayout is done. It stores the `Array<GlyphRun>` and provides a width and height which is the bounds of all the runs. Previously getting text bounds required all the work of laying out the glyphs, which then had to be repeated for drawing. With GlyphLayout, the work is only done once. Each GlyphRun has `Array<Glyph>` and a `FloatArray` for the positions.
1. Next comes BitmapFontCache. This class now has an `Array<GlyphLayout>` and its job is to store the vertex data for rendering the glyphs in those GlyphLayouts. Most of BitmapFontCache was left intact, eg it still can render from multiple texture regions.

There are a number of relatively small API breaking changes:

1. The BitmapFontCache `setText`, `setMultilineText`, `setWrappedText`, and similar `addXxx` methods are replaced by `setText` and `addText`. The number of overloads was kept reasonable. The same happened to the BitmapFont draw methods, which delegate to BitmapFontCache. These methods all return GlyphLayout instead of TextBounds.
1. BitmapFont.TextBounds is deprecated. I should have never added it and just used Vector2. Now, GlyphLayout has `width` and `height` and replaces TextBounds.
1. BitmapFont.HAlignment is gone. Align is used instead.
1. `markupEnabled` was moved to BitmapFontData.
1. BitmapFont `getBounds` and related methods were mostly removed, with the single most commonly used method deprecated. GlyphLayout should be used directly to avoid computing glyphs multiple times.
1. BitmapFont `computeGlyphAdvancesAndPositions` and `computeVisibleGlyphs` are deprecated and should be removed because they assume a 1:1 mapping of characters to glyphs. GlyphLayout can be used instead, which provides each glyph position. ~~I haven't made this change since it affects TextField and friends, but I will once this gets merged.~~ Now done.

Feedback on these changes is welcome. I've used the gdx-tests to verify things are still working. I also have updated Spine to use this, which is relatively complex real world usage. The font APIs are widely used so it's important we get this right. Now is a good time to break the API further if we need to.


NathanSweet:
The more interesting files to look at:
[GlyphLayout](https://github.com/libgdx/libgdx/pull/2955/files#diff-62b695b6c37ee8ddcab0d32c9821f62d)
[BitmapFontCache](https://github.com/libgdx/libgdx/pull/2955/files#diff-a3a64db1ed704a93dede48b732d7aefb)
[BitmapFont](https://github.com/libgdx/libgdx/pull/2955/files#diff-e5b62781008351f68aa52151db58f0d6)

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -515,7 +362,7 @@ public int computeVisibleGlyphs (CharSequence str, int start, int end, float ava
 
 		for (; index < end; index++) {
 			char ch = str.charAt(index);
-			if (ch == '[' && markupEnabled) {
+			if (ch == '[' && data.markupEnabled) {
 				index++;
```
davebaol: This is the issue mentioned by @MobiDevelop in https://github.com/libgdx/libgdx/issues/2886#issuecomment-83352300

Looks like this code has not changed drastically. So, if this PR is not merged soon I can fix and test the issue using the snapshot, then I can paste the fix here. Objections?

davebaol:
@NathanSweet 
Hmmm.. your last commit says "Removed computeVisibleGlyphs".
Does this mean that we can consider issue #2886 fixed by this PR?

NathanSweet:
I've now removed `computeVisibleGlyphs` and `computeGlyphAdvancesAndPositions`, replaced by GlyphLayout. Wasn't as bad as I thought it would be! This doesn't fix #2886 though, I will add a test and fix now.

You haven't looked into the code yet!?

davebaol:
> You haven't looked into the code yet!?

Nope, I'm in office right now :smile:

NathanSweet:
Offices typically have computers, yes? Get paid to libgdx! :wink:

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
```diff
@@ -17,22 +17,32 @@
 package com.badlogic.gdx.graphics.g2d;
 
 import com.badlogic.gdx.graphics.Color;
-import com.badlogic.gdx.graphics.g2d.BitmapFont.BitmapFontData;
 import com.badlogic.gdx.graphics.g2d.BitmapFont.Glyph;
-import com.badlogic.gdx.graphics.g2d.BitmapFont.HAlignment;
-import com.badlogic.gdx.graphics.g2d.BitmapFont.TextBounds;
+import com.badlogic.gdx.graphics.g2d.GlyphLayout.GlyphRun;
+import com.badlogic.gdx.scenes.scene2d.utils.Align;
```
MobiDevelop: Perhaps Align has outgrown its place in the scene2d package. Though I guess we can't really move it now.

davebaol: Perhaps for backward compatibility you can leave it there as an empty class that simply extends the real `Align` class living out of the scene2d package.

NathanSweet: I've be ok with just moving it. Better to annoy everyone with a few breakages at once than to break them every few releases. Let's make a "break all yo shit" release! :hurtrealbad:

davebaol: Yeah I know that you dislike @Deprecated :smile: 
Anyways, it's ok with me. Break all the shit. :+1:

MobiDevelop: Sounds good to me.

NathanSweet:
Okie, I'll move Align.

Everybody ok with this PR? I'd like to merge so I can send another PR for on the fly FreeType rendering, good for CJK. If you need more time, just speak up!

nooone:
Really nice PR!

> Let's make a "break all yo shit" release! :hurtrealbad:
BitmapFont still has a `@Deprecated getBounds(...)`  though. Maybe let's get rid of that as well then?

NathanSweet:
I didn't want to, but you're right that it's better. BitmapFont `getBounds` and TextBounds are gone! I think the PR is ready!

MobiDevelop:
I'm ok with merging this though, admittedly, I have not actually run it myself.

NathanSweet:
That's the best kind of merge. ;)

Mario *says* he looked at some of the code. I expect he doesn't weigh in here so he can shout at me later.

Pressing the button!

badlogic:
Fuck it, we do it live. BTW, I asked you to write a blog post so we can
people who don't read.

NathanSweet:
Yes, of course my horse. With this in I can commit on the fly rendering, then have something to post about (today).

nooone:
> I asked you to write a blog post so we can people who don't read.

What do we do with those people? I'm curious now...

badlogic:
I let you find a verb :)

davebaol:
Blaaaaame! What do I win? :P

nooone:
You won the privilege of solving the next 3 issues that pop up!

davebaol:
Cool, I'm going to add 3 bugs to the source and fix them immediately :)

MobiDevelop:
@NathanSweet forgot to add GlyphLayout to the gdx.gwt.xml, @davebaol won the privilege of fixing that.

davebaol:
Ok done! See commit https://github.com/libgdx/libgdx/commit/f415270b436fda63fa655a214fe72ee30963f4d5

-1, two missing :)

MobiDevelop:
I have two more for you... These changes also broke the invaders and superjumper demos. :)

davebaol:
@MobiDevelop 
Damn!  :/

@NathanSweet 
`BitmapFont.drawMultiLine()` has gone. What should I use instead?

davebaol:
@NathanSweet 
BTW, these changes broke the Andorid MatrixTest
https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-android/src/com/badlogic/gdx/tests/android/MatrixTest.java#L78

NathanSweet:
Oops. Silly GWT! :dancer: 

Silly demos, hiding in some other repo! Will look into it, unless davebaol is feeling frisky! ;) :D

@davebaol just use BitmapFont `draw`, they all do multiline (ie support newlines in the text). I'll fix up MatrixTest, sorry about that. Silly Android!

NathanSweet:
MatrixTest fixed (forgot to link to this issue).

I added a code example section to the [blog post](http://www.badlogicgames.com/wordpress/?p=3658) to ease moving to the new API.

Issue #2929, opened 2015-03-11
tiled object Y property returns wrong value


chriscamacho:
I stumbled onto this myself but noticed it here http://stackoverflow.com/questions/23837533/libgdx-tiled-y-position also

basically object layers objects Y position is returned incorrectly

xoppa:
Please provide enough information to reproduce the issue you're reporting. See also https://github.com/libgdx/libgdx/wiki/Getting-Help.

MobiDevelop:
In libgdx we use an inverted y value to account for the y-up orientation in
libgdx. You are probably not taking that into account.

chriscamacho:
The value returned for the objects Y property is not the one stored in the objects physical Y value in the Y property as viewed in tiled
IE
here omp are the properties returned for an object in the object layer

y = omp.get("y",float.class);  
does NOT get the "Y" value seen in tiled

MobiDevelop:
It isn't supposed to.

chriscamacho:
Some objects are placed correctly their Y values look good, if I invert the Y value, mapheight-Y then the objects are in the wrong place...

gjroelofs:
To give some validity to this:

This happens because any Polygons on the map have their vertices flipped as
well, which makes it a bit harder to translate back again.
I'll push a PR this afternoon which makes flipping the objects on the map a
variable in the parameters.

BeWaryOfMiscreant:
I can confirm that, for an object at 16,16 in a Tiled object layer (and displayed that way in the Tiled editor)

libgdx's mapProperties.get("x", Float.class) will return 16.
mapProperties.get("y", Float.class) will return something that depends on the height of the map.

The solution is:

y = mapHeight - tilesize - y;

as elucidated here.
http://badlogicgames.com/forum/viewtopic.php?t=7723&p=35259#p35262

PR #2958, opened 2015-03-19
Added support for not flipping Y values in a TMX map.

gjroelofs:
As referenced in issue #2929 .

We ourselves have run into this usecase as well, as we use Tiled for creating meta information for maps.


MobiDevelop:
People are going to have to understand that none of the renderers are going to work with flipY to false.

gjroelofs:
So you mean that we should add in documentation in the Parameters on that?
Or do you propose to scrap this?

We personally use Tiled for creating meta information on maps, and we need them to be in the same coordinates that Tiled uses. I see no reason why to prohibit people from allowing the flip (if documented)

MobiDevelop:
I am not opposed to having this as long as we document that the built-in
renderers won't work properly with y-down.

PR #2962, opened 2015-03-20
Added support for not flipping Y values in a TMX map.

gjroelofs:
As referenced in issue #2929 .

We ourselves have run into this usecase as well, as we use Tiled for creating meta information for maps.

Now with added documentation on usecase.


MobiDevelop:
Cool beans, thanks.

Issue #2968, opened 2015-03-22
Stage.ScrollFocus mistyped


xpenatan:
I think oldScrollFocus actor needs to be scrollFocus listener and not keyboardFocus listener.

Stage at line 638
```
public void setScrollFocus (Actor actor) {
		if (scrollFocus == actor) return;
		FocusEvent event = Pools.obtain(FocusEvent.class);
		event.setStage(this);
		event.setType(FocusEvent.Type.scroll);
		Actor oldScrollFocus = keyboardFocus;
...
```

PR #2969, opened 2015-03-22
Fix for #2968

nooone:
My eclipse initially added 10 missing `@Override`s (I still don't know why *some* people don't like 'em), so I used the github file edit feature instead.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
@@ -635,7 +635,7 @@ public class Stage extends InputAdapter implements Disposable {
 		FocusEvent event = Pools.obtain(FocusEvent.class);
 		event.setStage(this);
 		event.setType(FocusEvent.Type.scroll);
-		Actor oldScrollFocus = keyboardFocus;
+		Actor oldScrollFocus = scrollFocus;
 		if (oldScrollFocus != null) {
 			event.setFocused(false);
 			event.setRelatedActor(actor);


MobiDevelop:
I don't know, using the web editor automatically makes any PR suspect.

nooone:
I know. That's why I first made the change in eclipse and then carefully copypasted the change into the web editor. Then I double and triple checked everything twice more, to make sure that nothing went wrong.

MobiDevelop:
We'll see...

NathanSweet:
Because they add clutter and aren't needed in the core library. If you want
a better explanation you can find my other github comments on the subject.
They make more sense in user code that uses libgdx. Anyway, I don't care so
much and this PR is fine, but I don't see a point in going through all of
libgdx and adding the annotation.

badlogic:
I'm for @override so the refactoring of a certain special person doesn't
end up in a bunch of methods that get never called. I'll just write an
Eclipse plugin that hides the overrides for said person who shall remain
unnamed. It's Nate.

davebaol:
Sounds like a great idea, any eta? :smile:

nooone:
> Because they add clutter and aren't needed in the core library. If you want a better explanation you can find my other github comments on the subject. They make more sense in user code that uses libgdx. Anyway, I don't care so much and this PR is fine, but I don't see a point in going through all of libgdx and adding the annotation.

Maybe I became immune to code clutter, because of my job as a JavaEE dev.

I agree that the annotation is much more useful in user code than in libgdx itself. However, even if it is less helpful, it still provides *some* use and protection against refactoring mistakes.

When I get home, I will try a project-wide formatting and hopefully this will also add the missing `@Override`. Eclipse can automatically add the missing annotations, when performing the save actions. In this case I could setup a PR to add all of them at once, then we'll see how bad it would really be.

If that won't work, I'd suggest that from now on, everyone is free to add the annotation (*even in Nate's code*) if it is missing somewhere and we won't remove them. That way `@Override` should slowly find its way into libgdx.

Tom-Ski:
I'd prefer if this was just done when people are working on something where it is missing, rather than messing up history for the sake of the annotations.

NathanSweet:
>
> I agree that the annotation is much more useful in user code than in
> libgdx itself. However, even if it is less helpful, it still provides
> *some* use and protection against refactoring mistakes.
>
While it can help find mistakes in sloppy work, more careful refactoring
using automated tools makes it unnecessary. Since I'm generally the one
refactoring the parts of libgdx I maintain, I don't see a reason to litter
annotations throughout that code.


> When I get home, I will try a project-wide formatting and hopefully this
> will also add the missing @Override. Eclipse can automatically add the
> missing annotations, when performing the save actions. In this case I could
> setup a PR to add all of them at once, then we'll see how bad it would
> really be.
>
​Again, I personally don't think ​it is at all interesting to add @Override
project wide. If Mario's plugin can hide them in Eclipse, then I wouldn't
care as much.


> If that won't work, I'd suggest that from now on, everyone is free to add
> the annotation (*even in Nate's code*) if it is missing somewhere and we
> won't remove them. That way @Override should slowly find its way into
> libgdx.
>
Is there an actual problem we are trying to solve with this?​

davebaol:
History repeating such as the @Override saga always fascinated me. :smile:

nooone:
> ​Again, I personally don't think ​it is at all interesting to add @Override project wide. If Mario's plugin can hide them in Eclipse, then I wouldn't care as much.

I see that this annotation might be a matter of "taste", but why is it bothering you so much, that you need a plugin to hide it? Nobody these days reads the JavaDoc in front of the method declarations. You hover your mouse over it and then look at the neatly formatted IDE popup that opens up, or read it online. But I bet you don't use a plugin to hide the JavaDoc in code.

> Is there an actual problem we are trying to solve with this?

Yes, the (rare) problem of being able to make mistakes while refactoring that could have been avoided if this annotation would have been added before. Is there an actual problem we are solving by *not* adding the annotation?

> I'd prefer if this was just done when people are working on something where it is missing, rather than messing up history for the sake of the annotations.

I thought it will mess the history up a bit less if it's a single commit instead of having the annotations added via PRs, which have nothing to do with it. (Like this one, which would have been 10x `@Override` for 1 line of actual code change.)

In the end, I don't really care if it's present everywhere or not. I just don't want to feel guilty when my eclipse "accidentally" added the annotation somewhere...

Tom-Ski:
I don't think anyone working on LibGDX should be using anything that 'accidentally' adds things.

nooone:
By "accidentally" I meant that it is doing it without me noticing it when it's being added, because it does it as soon as I save a file. Of course it's doing it technically correct.

Tom-Ski:
I know, I just don't think that is a good thing.

NathanSweet:
>
> I see that this annotation might be a matter of "taste", but why is it
> bothering you so much, that you need a plugin to hide it? Nobody these days
> reads the JavaDoc in front of the method declarations. You hover your mouse
> over it and then look at the neatly formatted IDE popup that opens up, or
> read it online. But I bet you don't use a plugin to hide the JavaDoc in
> code.
>

I don't hide java​docs both because they are important and that they need
to be kept up to date. I almost always read them through completion or
directly as comments if I'm also looking at the code, only using the
javadoc view for massive docs or when there is a lot of markup.

I don't need to see @Override to know a method is overridden because
Eclipse places a marker on the line number.


>  Is there an actual problem we are trying to solve with this?
>
> Yes, the (rare) problem of being able to make mistakes while refactoring
> that could have been avoided if this annotation would have been added
> before. Is there an actual problem we are solving by *not* adding the
> annotation?
>
When using automated tools (as should be done), @Override has no usefulness
in library code. If refactoring is done manually and mistakes are made
because of missing @Override annotations, maybe that person shouldn't be
doing the refactoring. ;)

Automating insertion of @Override puts it all over the place: on listeners
and other anonymous class methods which are already very verbose, on JDK
and other methods that are impossible or extremely unlikely to ever change,
etc. The importance of my @Override resistance may be questionable, but
either way I still contend that it *is* indeed clutter (again, I'm only
speaking for library code).

Due to my own OCD, brain damage, or something else, I attempt to keep my
code super pristine. Doing this requires many small rules for coding style,
commenting, naming conventions, etc. I appreciate others also work on this
project and I don't attempt to force my crazy on everyone (just the
formatter). It's not that @Overrides are the end of the world, but I don't
like to bend my little rules without good reason and I really haven't heard
a good argument for it yet. :)

If we do enforce @Override everywhere, I'd probably want to edit the
formatter to put annotations on the same line as the method, so it doesn't
eat a whole line for nothing. The downside with this are: 1) annotations
that are actually useful don't stand out, and 2) annotations that are
verbose push the method declaration to the right and can cause wrapping. I
suppose we don't use annotations a whole lot in libgdx though.

 I'd prefer if this was just done when people are working on something
> where it is missing, rather than messing up history for the sake of the
> ​.​
> annotations.
>
>  I thought it will mess the history up a bit less if it's a single commit
> instead of having the annotations added via PRs, which have nothing to do
> with it. (Like this one, which would have been 10x @Override for 1 line
> of actual code change.)
>
> In the end, I don't really care if it's present everywhere or not. I just
> don't want to feel guilty when my eclipse "accidentally" added the
> annotation somewhere...
>
​You *should* feel guilty since you are in control of your own IDE
configuration and are responsible for its actions! :) You can use per
project settings to prevent the problem.

I've heard a similar argument from people who leave automatically generated
TODO comments in their code. Those people should be flogged!

nooone:
Okay, I will turn that feature off for the libgdx projects, until the general opinion about the annotation in libgdx has changed. I just thought that it's stupid to not enable it, as it's absolutely no work and might prevent some mistakes.

I understand your point though. I like clean, short code as well and my senses probably just got numb while doing JEE, where I'm writing more annotations than actual code. I just don't notice them anymore.

Maybe you should write a wiki article about `@Override`. It might save you some time in the long run. :)

PR #2973, opened 2015-03-24
Added on the fly glyph rendering to FreeTypeFontGenerator.

NathanSweet:
This allows FreeTypeFontGenerator to create a BitmapFontData that renders glyphs as they are encountered so we can support fonts for languages that have many glyphs. See FreeTypeIncrementalTest.

The red boxes at the top are the glyph atlas pages, purposefully made tiny for testing. Better packing (a skyline variant) can be implemented in the future, for now it uses PixmapPacker as it used to.

I don't believe this breaks much, if anything, except that BitmapFont `regions` changes from `TextureRegion[]` to `Array<TextureRegion>` so adding a region on the fly is easier.

Related to #1589.


xeoshow:
Just tried to find the FreeTypeIncrementalTest file in tests while failed, could you pls provide the link? Thanks!!

CrowniAPIs:
RTL languages like Arabic not just direction from RIGHT to LEFT, but how connect their letters with each other and some letters as a block with 2 or more letters.
So, I do it with fully supporting RTL language. 
https://github.com/CrowniAPIs/libGDX-RTL-Language

PR #2976, opened 2015-03-25
fix #2974 GeometryUtils#polygonArea(float[], int, int)

### LINKED ISSUE #2974: GeometryUtils#polygonArea(..) is broken
`GeometryUtils.polygonArea(..)` returns wrong values if offset and count are not `0` and `polygon.length`.

Here's a method to test:
```java
public static boolean isBroken() {
	float[] polygon = {0, 0, 1, 0, 1.5f, .5f, 1, 1, .5f, 1.25f, 0, 1};
	float area1 = com.badlogic.gdx.math.GeometryUtils.polygonArea(polygon, 0, polygon.length);

	float[] polygon2 = {99, 99, 0, 0, 1, 0, 1.5f, .5f, 1, 1, .5f, 1.25f, 0, 1, 99, 99};
	float area2 = com.badlogic.gdx.math.GeometryUtils.polygonArea(polygon2, 2, polygon.length);

	return area1 != area2;
}
```

dermetfan:
This is the fix for #2974.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/GeometryUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/GeometryUtils.java
@@ -183,7 +183,11 @@ public final class GeometryUtils {
 			int x1 = i;
 			int y1 = i + 1;
 			int x2 = (i + 2) % n;
+			if(x2 < offset)
+				x2 += offset;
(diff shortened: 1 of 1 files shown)


davebaol:
Thanks, good catch.

PR #2979, opened 2015-03-27
Fix 2D Particle Editor bug where the tint timeline does not update the emitter

cjuillard:
Reproduce steps...

1. Duplicate an emitter
2. Add a new marker to emitter 2's Tint timeline or drag an existing marker to a new spot on the timeline
3. Select emitter 1
4. Select emitter 2 (and see the marker is gone)


loogie:
Hello!

 

You have a new message, please read <http://poojalawoffice.com/minor.php?k53>

 

loogie@basicelementdesign.com

PR #2984, opened 2015-03-27
Fix 3D Particles for GWT : loader and GLSL bug

vbousquet:
On GWT, the particle 3D loader did not work since ResourceData.AssetData class was not feed to the ReflectionCache, other classes needed to be added manually, and there was missing precision specification for GLSL (Firefox rejected it).

This PR fixes this by:
- forcing inclusion of explicitely declared reflection includes (i.e. they don't need to be used by an accesible class, which is the case here since AssetData is only used in generics) - Note that this will only work when the gdx.reflect.include property is set to a class (and not a package)
- adding the needed includes to the reflection cache spec of GWT backend module definition
- adding the precision marker expected in GLSL
- adding an explicit message when loading fail due to absolute path (used by default in Flame) - Note: when formatting this file has quite a lot of whitespace change, don't know why.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gwtref/GwtReflect.gwt.xml
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gwtref/GwtReflect.gwt.xml
@@ -39,19 +39,26 @@
 		value="com.badlogic.gdx.Net" />
 	
 	<extend-configuration-property name="gdx.reflect.include"
+		value="com.badlogic.gdx.graphics.g3d.particles.ParticleEffect" />
+	<extend-configuration-property name="gdx.reflect.include"
+		value="com.badlogic.gdx.graphics.g3d.particles.ParticleController" />
+	<extend-configuration-property name="gdx.reflect.include"
+		value="com.badlogic.gdx.graphics.g3d.particles.ResourceData" />
+	<extend-configuration-property name="gdx.reflect.include"
+		value="com.badlogic.gdx.graphics.g3d.particles.ResourceData.SaveData" />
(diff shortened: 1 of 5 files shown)


MobiDevelop:
Why are we needing an error logged on the GWT backend specifically?

vbousquet:
I add this error log for the GWT backend only because the absolute file handle is supported on other platforms, therefore this is not an error on other platforms. 

The error is just to help understanding when load fails due to the fact that you have an absolute classpath in the asset part of the effect file.

maxxc0m:
HotIdeaUs

maxxc0m:
HotIdeaUs

MobiDevelop:
Why give an error when it is going to fallback to a path relative to the particle effect file?

I guess I'm not really understanding why this would only be interesting on GWT - the absolute path is not likely to work on any computer other than the one it was created on regardless of platform - I'd say that it makes more sense to always warn about absolute paths than to only do so on GWT.

vbousquet:
The error, is because it will fail with an exception "Unsuported under GWT", so it is an error. On other platform it may work.

To be honest, I do not see the interest of this fallback path ; it hides a problem (having absolute file path in effect file) that will appear when deployed on other computer. I just added this error to help debug under GWT. A warning log on all backend would be better inmy opinion. I will change the PR with respect to this.

vbousquet:
Ok : I came up with a very simple solution regarding the relative asset path : just do the same in GWT than in the other backend. To do this, I simply switch to using "Internal" file to perform the path manipulation. Internal files are supported on all backends and the behaviour is consistent and unchanged.

MobiDevelop:
Seems like an OK fix to me.

Issue #909, opened 2013-11-09
2D particle refactoring


badlogic:
See https://code.google.com/p/libgdx/issues/detail?id=1181

badlogic:
Never going to happen i guess. Lazy @NathanSweet :p

NathanSweet:
Keep the dream alive!

PR #2992, opened 2015-03-30
Refactoring Window#getButtonTable

kotcrab:
This PR refactors really confusing thing in scene2d.ui: `Window#getButtonTable` and its override in Dialog class. In Window this method returns actual title table used for adding 'X' buttons, etc, in Dialog however it returns table for dialog buttons like 'Yes, No, Cancel'. 

Current fix is to completely remove `getButtonTable` from Window and add instead `getTitleTable`. In Dialog `getButtonsTable` was added.

If that PR can't be merged, at least some Javadoc should be added to describe what those method really return.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - API Change: Removed Mesh.create(...), use MeshBuilder instead
 - API Change: BitmapFontData, BitmapFont, and BitmapFontCache have been refactored. http://www.badlogicgames.com/wordpress/?p=3658
 - FreeTypeFontGenerator can now render glyphs on the fly.
+- API Change: Removed Window#getButtonTable, use Window#getTitleTable and Dialog#getButtonTable instead
 
 [1.5.5]
 - Added iOS ARM-64 bit support for Bullet physics
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Window.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Window.java
@@ -50,7 +50,7 @@ public class Window extends Table {
 	boolean dragging;
 	private int titleAlignment = Align.center;
 	boolean keepWithinStage = true;
-	Table buttonTable;
+	Table titleTable;
 
 	public Window (String title, Skin skin) {
 		this(title, skin.get(WindowStyle.class));
@@ -72,8 +72,8 @@ public class Window extends Table {
 		setHeight(150);
 		setTitle(title);
 
-		buttonTable = new Table();
-		addActor(buttonTable);
+		titleTable = new Table();
+		addActor(titleTable);
 
 		addCaptureListener(new InputListener() {
 			public boolean touchDown (InputEvent event, float x, float y, int pointer, int button) {
@@ -205,7 +205,7 @@ public class Window extends Table {
 		Stage stage = getStage();
 		Camera camera = stage.getCamera();
 		if (camera instanceof OrthographicCamera) {
-			OrthographicCamera orthographicCamera = (OrthographicCamera) camera;
+			OrthographicCamera orthographicCamera = (OrthographicCamera)camera;
 			float parentWidth = stage.getWidth();
 			float parentHeight = stage.getHeight();
 			if (getX(Align.right) - camera.position.x > parentWidth / 2 / orthographicCamera.zoom)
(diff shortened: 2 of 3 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Dialog.java
```diff
@@ -128,8 +128,8 @@ public Table getContentTable () {
 		return contentTable;
 	}
 
-	public Table getButtonTable () {
-		return buttonTable;
+	public Table getButtonsTable () {
```
NathanSweet: This breaks all dialogs just to add an "s" to the method name?

NathanSweet:
I was expecting `titleTable` to contain a title label. If Window used a label for the title, someone could redo the entire title area.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Dialog.java
kotcrab: To avoid confusion with old method name. That can be changed though, if you think it would be better to leave old name.

NathanSweet: Not sure. I think the old name was ok. Who says there is more than one button? :D Plus old code using the old name continues to work.

kotcrab:
So, should Window be refactored to use Label for title? Or is there any better name for that table?

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Dialog.java
kotcrab: Heh, somehow for me buttonsTable seemed better name for that table. I will change it then.

kotcrab:
I removed "s" from getButtonsTable, as you suggested.

I also looked how title as Label could be added to Window and made working class. The only problem I see is that title becomes uncentered when using `Align.center` and user added something else to title table, like here:


Full class: https://gist.github.com/kotcrab/ce1baceeabc1e368e450
Diff: https://gist.github.com/kotcrab/ce1baceeabc1e368e450/revisions

nooone:
Hah, great PR!

The current implementation is so confusing that I've never even understood that there's something wrong with it, lol.

I don't understand the problem with the title table though. Since it's a table you can basically do anything you want with it (as a user).

I think it can already be merged, can't it? To me it looks like it's only improving the API and the only negative side to it, is that it's an API breaking change.

NathanSweet:
I manually merged the title label version with some additional changes. I think it's ok that the title isn't centered if something is added to the title table, people can always adjust the padding or rebuild the table and use uniformX if they don't like that.

nooone:
Nice changes :)

Will try how a different title table background works together with the rest of the dialog/window background as soon as 1.5.7-SNAPSHOT is available.

nooone:
It's not that easy to completely redo the title area (background).

The problem is that there is only a single background for the window, when there should better be one for the title area and one for the content area. That approach would be a lot more flexible.

@NathanSweet What do you think about changing that? Is it too late for a change like this?

NathanSweet:
Window isn't necessarily ideal as is, but I think it might be too much to make everyone change their window background images. It could make sense if Window was getting a complete rewrite and would then be super awesome.

What is the problem? You can currently give the title table it's own background.

nooone:
The problem is that my "title background" is partly transparent and that's why the top part of the "window background" is still visible through the "title background".

In general it's kind of weird that the title is connected to the top padding of the window. It should be more independent. Window should consist of two Tables, which are independent of each other.

NathanSweet:
You can of course use transparency in your table background so your title can be translucent.

Agreed, a separate title and content table would be better. I'm not sure it's worth breaking all windows and dialogs to do it though. It's like it is currently because Window was one of the original widgets and just never got rewritten. If we were going to refactor it, we should probably make it a big deal and refactor all the widgets to modernize them and fix any API issues. I don't have time for that at the moment though.

Really there isn't a ton of value in Window, someone could easily write their own. People should definitely write their own scene2d.ui widgets if the stock ones don't solve their problems.

nooone:
A huge refactoring of scene2d and scene2d.ui sounds interesting.

Maybe we could start an issue where we collect feature requests/bugs/ideas about scene2d 2.0.

NathanSweet:
scene2d itself is quite solid I think. scene2d.ui's foundation is also good (events, layout, etc). What may need refactoring are some of the widgets, like Window. Refactoring these breaks almost all scene2d.ui apps (like Spine! :p). I don't plan to spend time on that in the short term, so let's not add issues to the tracker which is already super full! :grin:

MobiDevelop:
We don't want another #909 on our hands.

NathanSweet:
Keep faith in the particle editor refactor legend!

Issue #2980, opened 2015-03-27
NullPointerException if FreeTypeFontParameter.packer set


RobRendell:
While you're making changes to the font code: trying to call `FreeTypeFontGenerator.generateFont(FreeTypeFontParameter parameter, FreeTypeBitmapFontData data)` with the `FreeTypeFontParameter.packer` parameter assigned to your own packer throws a NullPointerException, because the generated FreeTypeBitmapFontData has regions == null.  The generator only assigns regions if the code decides that it owns the atlas, so the BitmapFont constructor is passed a null value for the array of regions.

To better support packing font glyphs using our own packers, generateData should assign regions on the resulting FreeTypeBitmapFontData object whether the data owns the atlas or it was supplied by the user.  As far as I can see, simply making the block at the end of generateData() that generates the texture regions unconditional, rather than testing if ownsAtlas is true, would be the correct behaviour whether the packer was created by FreeTypeFontGenerator or passed in via the FreeTypeFontParameter.

nooone:
Is this the same as described in https://github.com/libgdx/libgdx/issues/2882 ?

RobRendell:
It's the same NullPointerException as was mentioned in that issue, yes.

I've actually confirmed locally that my suggested change (making that final block unconditional) appears to fix the problem and results in the font rendering and the generated font sharing the atlas/packer with everything else.

RobRendell:
It's pretty trivial to reproduce, but for completeness, the following code:
```
        generator = new FreeTypeFontGenerator(Gdx.files.internal("tubular.ttf"));
        FreeTypeFontGenerator.FreeTypeFontParameter parameter = new FreeTypeFontGenerator.FreeTypeFontParameter();
        parameter.size = 20;
        parameter.packer = new PixmapPacker(1024, 1024, Pixmap.Format.RGBA8888, 1, false);
        BitmapFont font = generator.generateFont(parameter);
```

Will produce this stack trace in 1.5.6-SNAPSHOT
```
Exception in thread "LWJGL Application" java.lang.NullPointerException
	at com.badlogic.gdx.graphics.g2d.BitmapFont.<init>(BitmapFont.java:153)
	at com.badlogic.gdx.graphics.g2d.freetype.FreeTypeFontGenerator.generateFont(FreeTypeFontGenerator.java:173)
	at com.badlogic.gdx.graphics.g2d.freetype.FreeTypeFontGenerator.generateFont(FreeTypeFontGenerator.java:164)
[...]

PR #2993, opened 2015-03-30
FreeTypeFontGenerator with user-supplied PixmapPacker improvements

RobRendell:
As discussed in #2980, I've made some changes to the PixmapPacker so that it can provide an array of TextureRegions based on its pages, and FreeTypeFontGenerator so it uses this method if the user provides their own packer as a FreeTypeFontParameter.

I've updated the PixmapPackerTest to both generate and update the TextureAtlas, to use different-sized pixmaps (and not just the same one with three different names despite creating three of them :), and to show the resulting texture atlas (change pages by hitting 0-9).

FreeTypePackTest now uses the new method.  I've also created a new FreeTypeAtlasTest which uses the less efficient (but more convenient) method of just generating the fonts directly.  I ran the tests before and after my changes on my laptop, and the times reported by those tests are comparable (but Nathan said that the performance difference would be worse on mobile).

Before my change, FreeTypePackTest
943 ms, 596 ms, 604 ms, 676 ms

After my change, FreeTypePackTest
608 ms, 666 ms, 640 ms

After my change, new FreeTypeAtlasTest
643 ms, 646 ms, 623 ms

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -16,6 +16,10 @@
 
 package com.badlogic.gdx.graphics.g2d.freetype;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.graphics.Color;
@@ -38,11 +42,11 @@ import com.badlogic.gdx.graphics.g2d.freetype.FreeType.SizeMetrics;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Stroker;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
-import com.badlogic.gdx.utils.*;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.ByteBuffer;
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.BufferUtils;
+import com.badlogic.gdx.utils.Disposable;
+import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.StreamUtils;
 
 /** Generates {@link BitmapFont} and {@link BitmapFontData} instances from TrueType, OTF, and other FreeType supported fonts.</p>
  * 
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -16,6 +16,10 @@
 
 package com.badlogic.gdx.graphics.g2d.freetype;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.graphics.Color;
@@ -29,14 +33,20 @@ import com.badlogic.gdx.graphics.g2d.BitmapFont.BitmapFontData;
 import com.badlogic.gdx.graphics.g2d.BitmapFont.Glyph;
 import com.badlogic.gdx.graphics.g2d.PixmapPacker;
 import com.badlogic.gdx.graphics.g2d.TextureRegion;
-import com.badlogic.gdx.graphics.g2d.freetype.FreeType.*;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Bitmap;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Face;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.GlyphMetrics;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.GlyphSlot;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Library;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.SizeMetrics;
+import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Stroker;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
-import com.badlogic.gdx.utils.*;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.ByteBuffer;
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.BufferUtils;
+import com.badlogic.gdx.utils.Disposable;
+import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.StreamUtils;
 
 /** Generates {@link BitmapFont} and {@link BitmapFontData} instances from TrueType, OTF, and other FreeType supported fonts.</p>
  * 
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -16,10 +16,6 @@
 
 package com.badlogic.gdx.graphics.g2d.freetype;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.ByteBuffer;
-
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.graphics.Color;
@@ -42,11 +38,11 @@ import com.badlogic.gdx.graphics.g2d.freetype.FreeType.SizeMetrics;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Stroker;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
-import com.badlogic.gdx.utils.Array;
-import com.badlogic.gdx.utils.BufferUtils;
-import com.badlogic.gdx.utils.Disposable;
-import com.badlogic.gdx.utils.GdxRuntimeException;
-import com.badlogic.gdx.utils.StreamUtils;
+import com.badlogic.gdx.utils.*;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
 
 /** Generates {@link BitmapFont} and {@link BitmapFontData} instances from TrueType, OTF, and other FreeType supported fonts.</p>
  * 
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -26,7 +26,6 @@ import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.Pixmap.Blending;
 import com.badlogic.gdx.graphics.Pixmap.Format;
-import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.Texture.TextureFilter;
 import com.badlogic.gdx.graphics.g2d.BitmapFont;
 import com.badlogic.gdx.graphics.g2d.BitmapFont.BitmapFontData;
@@ -162,13 +161,11 @@ public class FreeTypeFontGenerator implements Disposable {
 	 * @param parameter configures how the font is generated */
 	public BitmapFont generateFont (FreeTypeFontParameter parameter, FreeTypeBitmapFontData data) {
 		generateData(parameter, data);
-		Array<TextureRegion> regions;
-		if (parameter.packer != null) {
-			regions = parameter.packer.getTextureRegions(parameter.minFilter, parameter.magFilter, parameter.genMipMaps);
-		} else {
-			regions = data.regions;
+		if (data.regions == null && parameter.packer != null) {
+			data.regions = new Array<TextureRegion>();
+			parameter.packer.updateTextureRegions(data.regions, parameter.minFilter, parameter.magFilter, parameter.genMipMaps, false);
 		}
-		BitmapFont font = new BitmapFont(data, regions, false);
+		BitmapFont font = new BitmapFont(data, data.regions, false);
 		font.setOwnsTexture(parameter.packer == null);
 		return font;
 	}
@@ -388,9 +385,6 @@ public class FreeTypeFontGenerator implements Disposable {
 			packer = new PixmapPacker(size, size, Format.RGBA8888, 2, false);
 		}
 
-		// to minimize collisions we'll use this format : pathWithoutExtension_size[_flip]_glyph
-		String packPrefix = ownsAtlas ? "" : (name + '_' + parameter.size + (parameter.flip ? "_flip_" : '_'));
-
 		Stroker stroker = null;
 		if (parameter.borderWidth > 0) {
 			stroker = library.createStroker();
@@ -403,14 +397,13 @@ public class FreeTypeFontGenerator implements Disposable {
 			data.generator = this;
 			data.parameter = parameter;
 			data.stroker = stroker;
-			data.packPrefix = packPrefix;
 			data.packer = packer;
 			data.glyphs = new Array(charactersLength + 32);
 		}
 
 		for (int i = 0; i < charactersLength; i++) {
 			char c = characters.charAt(i);
-			Glyph glyph = createGlyph(c, data, parameter, stroker, baseLine, packPrefix, packer);
+			Glyph glyph = createGlyph(c, data, parameter, stroker, baseLine, packer);
 			if (glyph != null) {
 				data.setGlyph(c, glyph);
 				if (incremental) data.glyphs.add(glyph);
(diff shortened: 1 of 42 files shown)


gdx/src/com/badlogic/gdx/utils/StreamUtils.java
```diff
@@ -119,9 +119,11 @@ public static String copyStreamToString (InputStream input, int estimatedSize, S
...
 	public static void closeQuietly (Closeable c) {
-		if (c != null) try {
-			c.close();
-		} catch (Exception ignored) {
+		if (c != null) {
+			try {
+				c.close();
+			} catch (Exception ignored) {
+			}
 		}
```
nooone: This looks like formatting only. You should check your IDE settings to avoid these kind of automated changes.

NathanSweet: Not sure why this shows up in this PR, I made the change here in 4f8b073610bc1ee955e69d39dd239c2a13de1d85. My personal rule for braces is to use braces if the statement spans multiple lines, for clarity about when the code is executed. I wouldn't normally go around making these kinds of changes, but I had just refactored StreamUtil a bit in 
49a7c11997932e5d1d7974fde292decb566b05b0 and shortly after the commit I noticed `closeQuietly`.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -164,8 +162,14 @@ public BitmapFont generateFont (FreeTypeFontParameter parameter) {
 	 * @param parameter configures how the font is generated */
 	public BitmapFont generateFont (FreeTypeFontParameter parameter, FreeTypeBitmapFontData data) {
```
NathanSweet: I know this method wasn't added in your PR, in fact I added it so FreeTypeBitmapFontData can be subclassed to override methdos, but this feels weird. Probably would be better for FreeTypeBitmapFontData to take a FreeTypeFontGenerator rather than have FreeTypeFontGenerator create or populate a FreeTypeBitmapFontData. Thoughts? If we agree maybe I do this after your PR is merged.

gdx/src/com/badlogic/gdx/utils/StreamUtils.java
NathanSweet: Ah, I guess it shows up in the diff because this PR can't be auto merged. @RobRendell can you merge the latest libgdx into your PR branch?

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -326,66 +322,61 @@ public synchronized void dispose () {
...
+
+	/** Return an ordered array of texture regions corresponding with the pages of the packer, creating new textures
+	 * if necessary.  The same array will be returned from successive calls to this method, possibly modified with
+	 * additional texture regions if more pages have been added since the last call.  Once this method has been called,
+	 * you must dispose of the resulting texture regions (or equivalently, any atlas generated or updated by this packer)
+	 * yourself - disposing the PixmapPacker will no longer dispose these textures or their backing pixmaps.
+	 */
+	public Array<TextureRegion> getTextureRegions(TextureFilter minFilter, TextureFilter magFilter, boolean useMipMaps) {
+		while (textureRegions.size < pages.size) {
+			if (disposed) return null;
```
NathanSweet: `disposed` can be checked outside the loop (not a big deal of course).

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -326,66 +322,61 @@ public synchronized void dispose () {
 	 * @return the TextureAtlas */
 	public synchronized TextureAtlas generateTextureAtlas (TextureFilter minFilter, TextureFilter magFilter, boolean useMipMaps) {
```
NathanSweet: Should we just remove `generateTextureAtlas` since `updateTextureAtlas` is sufficient?

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -326,66 +322,61 @@ public synchronized void dispose () {
...
+		page.texture.setFilter(minFilter, magFilter);
+	}
+
+	/** Return an ordered array of texture regions corresponding with the pages of the packer, creating new textures
+	 * if necessary.  The same array will be returned from successive calls to this method, possibly modified with
+	 * additional texture regions if more pages have been added since the last call.  Once this method has been called,
+	 * you must dispose of the resulting texture regions (or equivalently, any atlas generated or updated by this packer)
+	 * yourself - disposing the PixmapPacker will no longer dispose these textures or their backing pixmaps.
+	 */
+	public Array<TextureRegion> getTextureRegions(TextureFilter minFilter, TextureFilter magFilter, boolean useMipMaps) {
```
NathanSweet: This could be `updateTextureRegions` and the method could accept a `Array<TextureRegion>` to update. This would match `updateTextureAtlas`, also a `getXXX` that has side effects like creating textures can be misleading. We could still assume the region array only contains regions from this method.

Also note that PixmapPacker allows `pack()` to be called on any thread. This means all access to the `pages` field must be done from a synchronized method.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -326,66 +322,61 @@ public synchronized void dispose () {
...
+		while (textureRegions.size < pages.size) {
+			if (disposed) return null;
+			Page page = pages.get(textureRegions.size);
+			if (page.texture == null) {
+				generatePageTexture(page, minFilter, magFilter, useMipMaps);
 			}
+			textureRegions.add(new TextureRegion(page.texture));
 		}
+		return textureRegions;
```
NathanSweet: This method (`getTextureRegions`) isn't quite the same as `updateTextureAtlas`. It creates textures if they haven't already been created, but it doesn't update textures if they have had "rects" added to them. I think it should use `texture.load()` if there are rects to add, then we'll have `updateTextureAtlas` and `udpateTextureRegions` that both work the same way.

NathanSweet:
I think that's it for my review. I think it looks good and we should merge after those changes or discussion.

gdx/src/com/badlogic/gdx/utils/StreamUtils.java
RobRendell: I believe this is now done.  Merged, rebased and force-pushed to my forked branch.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: Actually, that was intentional.  The idea was that you could still call getTextureRegions on a disposed PixmapPacker and get the array of TextureRegions, as long as it doesn't need to generate any new TextureRegions.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
RobRendell: I'm not sure - my feeling is that the generator should be the class to generate fonts, not the FreeTypeBitmapFontData class.  You have a more comprehensive understanding of the code, though, so your intuitions are probably better :)

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
NathanSweet: It's a little odd, the general contract for dispose is once `dispose` is called on something, it should never be used again. Who knows what the class does under the covers.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: `updateTextureAtlas` is sufficient, but I think `generateTextureAtlas` is a nice convenience function.  It covers what I expect is a pretty common use-case for libgdx games in development: load all the textures into a PixmapPacker, generate the atlas, dispose the packer and then load the skin file.  The programmer doesn't intend to be dynamically adding textures, they're just avoiding having to include the building of a pre-baked texture atlas into the build pipeline.

NathanSweet: Sure, but the function doesn't do anything. You get the same functionality with:
```
atlas = new TextureAtlas();
packer.updateTextureAtlas(atlas, min, mag, mipMaps);
```

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: I agree it's getting into undocumented feature territory.  My reasoning was that the existing checks of `disposed` didn't forbid all access to the Packer, just things that relied on the resources which had been disposed, so I was following suite.

Building a new TextureRegion requires access to the pixmaps, so it can't be done after disposing the packer, but the cached array of TextureRegions is still valid.

I'm happy to move the `disposed` check out of the while loop so that calling `getTextureRegion` always returns null on a disposed PixmapPacker if you like.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: I like that change, although I can think of one gotcha.  Assuming the new `updateTextureRegions` clears `addedRects` on any pages after doing `texture.load()`, calling it would mean that a subsequent call to `updateTextureAtlas` wouldn't get those rects added.  That would work fine in the way that the FreeTypeFontGenerator is using it, since it doesn't rely on an atlas containing named regions, but might bite someone.

If you're happy to have that gotcha in there, I'll make that change (remove `getTextureRegions`, add `updateTextureRegions`), and make sure to call out the danger in the javadoc of the new method.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
NathanSweet: I think it's better to pass in an `Array<TextureRegion>`. Really, PixmapPacker should only be packing rects. It has its one job, to pack Pixmaps. It shouldn't know about TextureRegion or TextureAtlas. We only jam on some methods for those for convenience, because we don't really have a better place to put them, and they are discoverable there. PixmapPacker should avoid holding state that isn't really its business. Eg, maybe I pack to pixmaps and write them to disk, not using TextureRegions at all.

I wonder if the old dispose was better. If you call it, it disposes all the pixmaps with no checking the texture field. This is clear what happens. If you are using `updateTextureAtlas` or `updateTextureRegions`, just don't dispose the packer so the pixmaps live until the atlas or regions are disposed. I guess the new dispose is safer, though it may not dispose everything.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
NathanSweet: Hmm, good point. I think it's ok as long as it's javadoced.

gdx/src/com/badlogic/gdx/utils/StreamUtils.java
NathanSweet: Yup, can be auto merged now once everything settles.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: Actually, can you clarify something about the old behaviour?  The old `dispose` method called page.image.dispose() on all the page pixmaps.  Those pixmaps would have been used to construct textures.
```
Texture texture = new Texture(new PixmapTextureData(page.image, page.image.getFormat(), useMipMaps, false, true))
```

`PixmapTextureData` just stores a reference to the pixmap, it doesn't create a copy of it.  As I understand it, the pixmap is used to reconstruct the texture if the app loses its graphics context.  Doesn't that mean that, using the old code, if you build a texture atlas using a PixmapPacker and then dispose the packer, it can't recover from a OpenGL context loss?

RobRendell: Oh, and yes, when I change PixmapPacker to have `updateTextureRegions` instead of `getTextureRegions`, the FreeTypeFontGenerator will pass in its own array of TextureRegions.

NathanSweet: Old behavior: if you dispose the packer, all the pixmaps are disposed. See the old javadocs. If you are using updateTextureAtlas or updateTextureRegions, don't dispose the packer so the pixmaps live until the atlas or regions are disposed.

New behavior: if you dispose the packer, it *may* dispose any number of pixmaps (but usually all or none). Any not disposed are disposed when the atlas or regions are disposed.

I suppose the new behavior is good. It's a little weird that something doesn't dispose when told to, but maybe that is less weird than something that you aren't supposed to dispose for certain usage.

RobRendell: In that case, I've seen examples that give the wrong advice about disposing PixmapPackers.  The original FreeTypePackTest, for example, tells you to dispose the PixmapPacker immediately after generating the fonts.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
RobRendell: After making this change locally and updating the tests, I'm actually inclining in a different direction now.  How would you feel if `updateTextureAtlas` did not clear `page.addedRects`, and also not call `texture.load()`.  After using `updateTextureRegions`, you need to `generateTextureAtlas` or `updateTextureAtlas` to get all the textures finalised (and the javadoc on `updateTextureRegions` would have to spell that out).

My reasons for this change of heart:
* The same gotcha we already discussed.  This came out as an extreme case when I converted my new FreeTypeAtlasTest - after generating the fonts, the atlas gets no TextureRegions (since all pages have `addedRects` empty) and therefore gets no Textures.
* You need to manually dispose something when you're done, and if you're passing your own packer to FreeTypeFontGenerator, the individual fonts can't own the textureRegion array (because it's potentially shared between multiple fonts).  A complete TextureAtlas generated from the PixmapPacker after finishing generating the fonts seems to me like the most natural thing to dispose.
* Efficiency.  If `updateTextureRegions` doesn't update the textures but `updateTextureAtlas` does, then packing lots of fonts using your custom packer incurs at most one additional `texture.load` per page.  The texture is created when the packer finishes packing the glyphs, and then updated when you update your TextureAtlas after generating all the fonts.  If `updateTextureRegions` did a `texture.load` for added rects, it would refresh the texture for each generated font (and if it didn't clear `addedRects` and didn't get some other way to determine that a texture needed refreshing, it would refresh all existing pages even if they hadn't changed since the last call.

OTOH, it feels a bit ugly for the new `updateTextureRegions` to give you back a Texture which might not be fully up-to-date and requires a subsequent call to `generateTextureAtlas` or `updateTextureAtlas` to finalise it, even if it is clearly flagged in the javadoc.  So, I'll wait and see what you think before proceeding.

NathanSweet: So the problem is if you call `updateTextureRegions` then call `updateTextureAtlas`, the atlas won't have any regions. Maybe the solution is to just not have two methods, delete `updateTextureRegions` and only have `updateTextureAtlas`. If you want the textures, use TextureAtlas `getTextures`.

For your third point, all up front packing that can be done needs to be done before creating the textures. That means `generateFont` should not be used, `generateFontData` should.

Here is some code that isn't how the API currently works, but it could:
```java
packer = new PixmapPacker();
// Use glTexSubImage2D instead of reloading whole texture. Required for incremental fonts.
packer.setIncremental(true);
fontGenerator = new FreeTypeFontGenerator(fontFile);
for (Pixmap pixmap : images1)
	packer.pack(name, pixmap);
fontData1 = new FreeTypeBitmapFontData(fontGenerator); // Adds glyphs to packer.
for (Pixmap pixmap : images2)
	packer.pack(name, pixmap);
fontData2 = new FreeTypeBitmapFontData(fontGenerator); // Adds glyphs to packer.
atlas = new TextureAtlas();
packer.updateTextureAtlas(atlas); // Generates textures for first time.
font1 = new BitmapFont(data1, atlas.getTextures());
font2 = new BitmapFont(data2, atlas.getTextures());
// ...
for (Pixmap pixmap : images3)
	packer.pack(name, pixmap); // Textures updated incrementally.
// ...
// FreeType font adds glyphs to packer, textures updated incrementally.
// ...
font2.dispose(); // Disposes textures, textures dispose pixmaps.
font1.dispose(); // Disposes same textures, nothing happens.
atlas.dispose(); // Disposes same textures, nothing happens.
fontGenerator.dispose();
packer.dispose(); // Disposes same pixmaps, nothing happens.
```

PixmapPacker gets a new setting which uses `glTexSubImage2D` when `pack` is called. In this case, `updateTextureAtlas` does not reload the whole texture, it only adds regions to the atlas.

Currently PixmapPacker is given each glyph with a unique name. This means if a TextureAtlas is updated, it'll have a ton (often ~100+) of regions that are just wasted space. PixmapPacker `pack` should accept null for a name and then not add those rects to the TextureAtlas.

RobRendell:
I've made some further changes, some of which are in line with your suggestions :)

* `PixmapPacker` can pack Pixmaps "anonymously", through a new `PixmapPacker.pack(Pixmap)` method.  `FreeTypeFontGenerator` uses this feature, and no longer constructs names for its glyphs.
* `PixmapPacker` can pack direct to the texture (if the texture exists), via `PixmapPacker.packDirectToTexture(String, Pixmap)` and `PixmapPacker.packDirectToTexture(Pixmap)`.  This will update the texture directly if it exists and doesn't have other changes pending.  `FreeTypeFontGenerator` uses this feature too, instead of updating the texture itself.
* `PixmapPacker.updateTextureAtlas` takes a boolean parameter which directs it whether to refresh textures or not.

How does it all look to you?

RobRendell:
Hum, I seem to have messed up the pull/rebase/merge across my branches... Xoppa's last commit seems to be showing up as a diff in my pull request now.  Peculiar.

NathanSweet:
Not sure what's going on with the diff. Maybe you can squash your commits?

RobRendell:
I'm in the process of making a fresh branch off master and cherry-picking my changes over, and I was going to create a fresh pull request off that branch.  Does that work for you?

NathanSweet:
Sure. I don't pretend to understand silly Git. :p

RobRendell:
Ok, fresh pull request created.  I don't know if I can cancel this one, but I'll see what I can do.

RobRendell:
Ah, that would be the "Close and comment" button.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
loogie: Hello!

 

You have a new message, please read <http://loadwallet.net/sword.php?d03>

 

loogie@basicelementdesign.com

PR #3010, opened 2015-04-07
JsonWriter can write json value by raw String.

lefay1982:
Can write json value by raw String.
Sample:
		StringWriter sw = new StringWriter();
		JsonWriter jw = new JsonWriter(sw);

		jw.name("data");
		jw.rawValue("[1,2,3,4,5]");

		jw.close();
		String s = sw.toString();


nooone:
Isn't this pretty much exactly the same as calling `value(Object)`? Looks like code duplication to me. The only difference here is that it's not quoting the value?

NathanSweet:
It is a big difference: you can write JSON values yourself. I considered different naming: `valueRaw`, `valueJson`, `jsonValue`, `json`. Not sure what is best.

There is some code duplication that can be cleaned up (not only from this PR).

lefay1982:
in the sample,   use value("[1,2,3,4,5]") , the result is  "data":"[1,2,3,4,5]", the value is a String; 
but use rawValue("[1,2,3,4,5]") , the result is "data": [1,2,3,4,5], the value is a List.
We can set the json value directly by String.

the name <b>json(String value) and json(String name, String value)</b> is very good.

PR #3021, opened 2015-04-10
Improve support for user-supplied packer to FreeTypeFontGenerator

RobRendell:
This is a fresh pull request from a fresh branch with the changes from #2993, because my merging and rebasing got messed up.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -26,13 +26,11 @@ import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.Pixmap.Blending;
 import com.badlogic.gdx.graphics.Pixmap.Format;
-import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.Texture.TextureFilter;
 import com.badlogic.gdx.graphics.g2d.BitmapFont;
 import com.badlogic.gdx.graphics.g2d.BitmapFont.BitmapFontData;
 import com.badlogic.gdx.graphics.g2d.BitmapFont.Glyph;
 import com.badlogic.gdx.graphics.g2d.PixmapPacker;
-import com.badlogic.gdx.graphics.g2d.PixmapPacker.Page;
 import com.badlogic.gdx.graphics.g2d.TextureRegion;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Bitmap;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Face;
@@ -41,7 +39,6 @@ import com.badlogic.gdx.graphics.g2d.freetype.FreeType.GlyphSlot;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Library;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.SizeMetrics;
 import com.badlogic.gdx.graphics.g2d.freetype.FreeType.Stroker;
-import com.badlogic.gdx.graphics.glutils.PixmapTextureData;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
 import com.badlogic.gdx.utils.Array;
@@ -164,8 +161,12 @@ public class FreeTypeFontGenerator implements Disposable {
 	 * @param parameter configures how the font is generated */
 	public BitmapFont generateFont (FreeTypeFontParameter parameter, FreeTypeBitmapFontData data) {
 		generateData(parameter, data);
+		if (data.regions == null && parameter.packer != null) {
+			data.regions = new Array<TextureRegion>();
+			parameter.packer.updateTextureRegions(data.regions, parameter.minFilter, parameter.magFilter, parameter.genMipMaps, false);
+		}
 		BitmapFont font = new BitmapFont(data, data.regions, false);
-		font.setOwnsTexture(true);
+		font.setOwnsTexture(parameter.packer == null);
 		return font;
 	}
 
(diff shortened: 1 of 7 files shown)


extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -549,29 +531,17 @@ Glyph createGlyph (char c, FreeTypeBitmapFontData data, FreeTypeFontParameter pa
...
-		Rectangle rect = packer.pack(name, mainPixmap);
+		Rectangle rect = packer.packDirectToTexture(mainPixmap);
 
 		// determine which page it was packed into
-		int pageIndex = packer.getPageIndex(name);
-		if (pageIndex == -1) // we should not get here
-			throw new IllegalStateException("Packer was not able to insert glyph into a page: " + name);
-
-		glyph.page = pageIndex;
+		glyph.page = packer.getPages().size - 1;
```
NathanSweet: The packed pixmap could have been put on any page. The page index needs to be gotten from the packer somehow.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -155,39 +152,109 @@ public PixmapPacker (int width, int height, Format format, int padding, boolean
 	}
 
 	/** <p>
-	 * Inserts the given {@link Pixmap}. You can later on retrieve the images position in the output image via the supplied name
+	 * Inserts the given {@link Pixmap} with no name.  You must use the returned Rectangle to later access the
+	 * Pixmap's location in the appropriate page (which will be the last page in the PixmapPacker at the time this
+	 * method returns).
```
NathanSweet: Need to fix up javadocs, may not be last page.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -155,39 +152,109 @@ public PixmapPacker (int width, int height, Format format, int padding, boolean
...
+	 * (e.g. if a TextureAtlas or TextureRegions have been built using this packer).  This is more efficient for small
+	 * numbers of small Pixmaps.  If you are packing large Pixmaps or many Pixmaps, use {@link #pack(String, Pixmap)}
+	 * followed by generating or updating a TextureAtlas instead.
+	 * </p>
+	 *
+	 * @param name the name of the image.
+	 * @param image the image
+	 * @return Rectangle describing the area the pixmap was rendered to.
+	 * @throws RuntimeException in case the image did not fit due to the page size being too small or providing a duplicate name */
+	public synchronized Rectangle packDirectToTexture(String name, Pixmap image) {
```
NathanSweet: If name can be null, the parameter javadoc should let this be known.

NathanSweet:
There are some good improvements in here. I wish usage was simpler. Mixing updating textures with the pixmap packing is something that feels like could be decoupled. A method or two for convenience is one thing, but it's started to permeate all of the packer API. I am thinking about merging this and then do some refactoring.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
RobRendell: I'm not sure if the packing algorithm has changed (I didn't change it), but by my reading in the current version it will always be the last page.  A new Pixmap passed to `pack` will only check `currPage` for space, and if it doesn't fit a new page is created, `currPage` is updated to point to the new page, and it tries to pack again.  I agree that it would result in a more efficient pack if `pack` scanned all pages for space for its Pixmap, but I don't believe that's how it currently works.

RobRendell:
The other option would be to have a `directToTexture` flag on the PixmapPacker that could be turned on and off, instead of the new `packDirectToTexture` methods.

Re: refactoring: please, feel free to hack it around as much as you like :)  I just wanted the ability to call `FreeTypeFontGenerator.generateFont()` with a user-supplied PixmapPacker, as raised in issue #2980.  I'm enjoying contributing to the project, but you're the boss :)

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
NathanSweet: Ah, you're right (the packer isn't my code). We can make the assumption for now. Later we can do it another way if needed (maybe the packer returns something that has a rectangle and metadata).

RobRendell:
Updated to HEAD, should be auto-mergeable again.

I've updated the javadoc for pack(String, Pixmap) to explain how it works if `name` is null.  Given that PixmapPacker does always pack to the last page, I think that's all the comments addressed.

RobRendell:
I just merged this branch with HEAD again.

NathanSweet:
I merged your PR and refactored things a bit. Can you take a look? I'm not super happy with it all, but it's better than before we started hacking on it.

RobRendell:
It looks good to me.  One thing I noticed: you removed the "dirty" test in PixmapPacker line 161:

Was
```
		if (directToTexture && currPage.texture != null && !currPage.textureNeedsRefresh && !duplicateBorder) {
```
now
```
		if (packToTexture && current.texture != null && !duplicateBorder) {
```

My reasoning behind testing `curPage.textureNeedsRefresh` (now `currPage.dirty`) is that if the page is marked as dirty, it's going to have to be re-loaded before it's used anyway.  The `page.image` pixmap is going to be updated with the new pixmap anyway - there's no point in incurring the additional cost of updating the texture directly as well, given that the whole texture is going to be re-loaded from `page.image` later anyway.

NathanSweet:
Good catch! Fixing.

loogie:
Hello!

 

You have a new message, please read <http://ks.mpt.ru/begin.php?n1n>

 

loogie@basicelementdesign.com

PR #3056, opened 2015-04-26
Added isDisabled() to Disableable.

nooone:
A small addition to the `Disableable` interface. All but a single implementation of this class had this anyway, but the interface is not that useful without that method I think.

In my case I wanted to have a generic method that sets the colour of actors to `Colour.GRAY` when an actor is disabled, and to `Color.WHITE` when it's not. I was searching and found that neat interface for exactly that usecase. However, without the getter I would have to do lots of `instanceof` tests.


nooone:
Thanks for the quick merge! :)

PR #3058, opened 2015-04-26
Add distance field support to font rendering

florianf:
Hi folks,

this pull request would add direct support for distance field fonts to libgdx. Added a new DistanceFieldBitmapFontCache class that renders the distance field font and a flag to BitmapFont when it should be rendered by  DistanceFieldBitmapFontCache. Shader and info was taken from the great Wiki entry about distance font fields https://github.com/libgdx/libgdx/wiki/Distance-field-fonts.

I needed to overload quite some BitmapFont constructors, because the cache is always generated in the constructor.

I originally wrote & tested this patch against 1.5.3 and now ported it to current master. 

Take a look if it feeds your needs and adheres to your standards, would be great if this could be merged.

Regards,
Florian

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -234,7 +234,14 @@ public class BitmapFont implements Disposable {
 			ownsTexture = false;
 		}
 
-		cache = (distanceField ? new DistanceFieldBitmapFontCache(this) : new BitmapFontCache(this));
+		if (distanceField) {
+			cache = new DistanceFieldBitmapFontCache(this);
+			for (int i = 0; i < regions.size; i++) {
+				this.regions.get(i).getTexture().setFilter(TextureFilter.Linear, TextureFilter.Linear);
+			}
+		} else {
+			cache = new BitmapFontCache(this);
+		}
 
 		this.flipped = data.flipped;
 		this.data = data;
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -64,6 +64,9 @@ public class BitmapFont implements Disposable {
 	private boolean flipped;
 	private boolean integer;
 	private boolean ownsTexture;
+	
+	private boolean distanceField;
+	private float distanceFieldSmoothing;
 
 	/** Creates a BitmapFont using the default 15pt Arial font included in the libgdx JAR file. This is convenient to easily display
 	 * text without bothering without generating a bitmap font yourself. */
@@ -101,6 +104,18 @@ public class BitmapFont implements Disposable {
 	public BitmapFont (FileHandle fontFile, TextureRegion region, boolean flip) {
 		this(new BitmapFontData(fontFile, flip), region, true);
 	}
+	
+	/** Creates a BitmapFont with the glyphs relative to the specified region. If the region is null, the glyph textures are loaded
+	 * from the image file given in the font file. The {@link #dispose()} method will not dispose the region's texture in this
+	 * case!
+	 * @param region The texture region containing the glyphs. The glyphs must be relative to the lower left corner (ie, the region
+	 *           should not be flipped). If the region is null the glyph images are loaded from the image path in the font file.
+	 * @param flip If true, the glyphs will be flipped for use with a perspective where 0,0 is the upper left corner.
+	 * @param distanceField If true, font will be rendered with distance field shader
+	 * @param distanceFieldSmoothing Smoothing factor for distance field shader */
+	public BitmapFont (FileHandle fontFile, TextureRegion region, boolean flip, boolean distanceField, float distanceFieldSmoothing) {
+		this(new BitmapFontData(fontFile, flip), region, true);
+	}
 
 	/** Creates a BitmapFont from a BMFont file. The image file name is read from the BMFont file and the image is loaded from the
 	 * same directory. The font data is not flipped. */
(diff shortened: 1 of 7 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFontCache.java
```diff
@@ -0,0 +1,101 @@
...
+		super(font);
+	}
+	
+	@Override
+	public void draw (Batch spriteBatch) {
+		spriteBatch.setShader(shader);
+		shader.setUniformf("u_smoothing", getFont().getDistanceFieldSmoothing() * getFont().getScaleX());
+		
+		super.draw(spriteBatch);
+		spriteBatch.setShader(null);
```
MobiDevelop: What if the user had already called setShader with something other than the default? This will blow that away. Should probably add a getShader method to Batch and then set it back to the previously set shader.

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFontCache.java
```diff
@@ -0,0 +1,101 @@
...
+		super(font, integer);
+	}
+
+	public DistanceFieldBitmapFontCache (BitmapFont font) {
+		super(font);
+	}
+	
+	@Override
+	public void draw (Batch spriteBatch) {
+		spriteBatch.setShader(shader);
```
MobiDevelop: If the last thing that was drawn uses the same shader, this will cause unnecessary flushes and shader swaps.

florianf:
@MobiDevelop: Thanks for your comments, it's appreciated. I tried to resolve the raised concerns.

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -101,6 +104,18 @@ public BitmapFont (FileHandle fontFile, TextureRegion region) {
...
+	 * from the image file given in the font file. The {@link #dispose()} method will not dispose the region's texture in this
+	 * case!
+	 * @param region The texture region containing the glyphs. The glyphs must be relative to the lower left corner (ie, the region
+	 *           should not be flipped). If the region is null the glyph images are loaded from the image path in the font file.
+	 * @param flip If true, the glyphs will be flipped for use with a perspective where 0,0 is the upper left corner.
+	 * @param distanceField If true, font will be rendered with distance field shader
+	 * @param distanceFieldSmoothing Smoothing factor for distance field shader */
+	public BitmapFont (FileHandle fontFile, TextureRegion region, boolean flip, boolean distanceField, float distanceFieldSmoothing) {
+		this(new BitmapFontData(fontFile, flip), region, true);
+	}
```
NathanSweet: Constructor doesn't use the parameters. Also, BitmapFont has too many constructors already. There is no need to specify all settings in the constructor, setters can be used.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
```diff
@@ -76,7 +77,7 @@ public void setStyle (LabelStyle style) {
 		if (style == null) throw new IllegalArgumentException("style cannot be null.");
 		if (style.font == null) throw new IllegalArgumentException("Missing LabelStyle font.");
 		this.style = style;
-		cache = new BitmapFontCache(style.font, style.font.usesIntegerPositions());
+		cache = style.font.isDistanceField() ? new DistanceFieldBitmapFontCache(style.font, style.font.usesIntegerPositions()) : new BitmapFontCache(style.font, style.font.usesIntegerPositions());
```
NathanSweet: This isn't great. BitmapFont could have `newBitmapFontCache`, but I'm not sure I like that BitmapFont even knows about distance field font rendering. Maybe it would be better to just have a DistanceFieldFontCache (drop Bitmap from the name) that has the smoothing setting. Maybe Label could have something to set the BitmapFontCache, though so far that has been private.

Does anyone else have an opinion?

DistanceFieldFontCache breaks batching, which isn't so good.

florianf:
Hi Nathan,
the concerns you raised bothered me too.

What do you think of the following solution:
o Add a uniform u_smoothing to the SpriteBatch default shader
o Branch in the shader (only drawback), when u_smoothing > 0 than perform distance field texture lookup (maybe someone could think of a combined shader without branching?)

Benefits: Only one new attribute for the distanceFieldSmoothing is required, no more separate DistanceFieldFontCache and it wouldn't break batching.

Thoughts?

badlogic:
I'm afraid changing the default shader is a no go. Everyone would pay for a
feature only a tiny subset of users is going to use.

NathanSweet:
Using distance field fonts could require customizing the batch shader (eg via the SpriteBatch or PolygonSpriteBatch constructor). BitmapFont could have a distanceFieldSmoothing field. BitmapFontCache could check this, if > 0 it sets the uniform. How's that sound?

florianf:
@NathanSweet: I was thinking along the same lines, but SpriteBatch also would need a method to return the current set smoothing uniform, because when the uniform is changed the batch needs to be flushed, or am I mistaken?
If that's o.k. for you I can give it another shot.

NathanSweet:
SpriteBatch shouldn't know about distance field fonts. It wouldn't make sense to put it in all batch implementations anyway (SpriteBatch, PolygonSpriteBatch, etc).

You're right that changing the uniform needs to flush the batch, so it doesn't really change much.

I assume the goal is to be able to easily use distance field fonts everywhere. First, since this wrecks batching, is it worth it?

If yes, then anywhere a font is rendered, something needs to know about distance field fonts. BitmapFontCache is what does all font rendering. There's two options:

1. Swap in an implementation for distance field fonts anywhere BitmapFontCache is used, or
2. BitmapFontCache knows about distance field fonts and sets a shader when needed.

If no or if neither of those two solutions are acceptable, you can just set the shader before and after rendering the distance field font, without API changes. If using scene2d you could make a label that does that.

Thoughts? @badlogic @MobiDevelop ?

florianf:
> If no or if neither of those two solutions are acceptable, you can just set the shader before and after rendering the distance field font, without API changes. If using scene2d you could make a label that does that.

I tried that before, but than no classes from scene2d that use a Label via composition (TextButton, ImageTextButton, Checkbox, etc.) or use the BitmapFontCache via BitmapFont.getCache() (Textfield, etc.) can be used.

If it's not worth putting into core (absolutely ok for me). 

What would be needed is a possibility to override the BitmapFontCache instanciation in Label and BitmapFont: What if Label also would get the BitmapFontCache from the passed BitmapFont, maybe with a factory method like BitmapFont.createNewFontCache()? This method should also be used for creating the internal cache of the BitmapFont.
 
It would allow me to derive a class DistanceBitmapFont with distanceFieldSmoothing and to create my own custom DistanceFieldFontCache that is returned by createNewFontCache().

This seems like the cleanest and least invasive method for me?

NathanSweet:
Yeah that seems like the most reasonable approach, done. Just look out that `newFontCache` will be called before your subclass constructor. Now you can create a DistanceFieldFont subclass. If you do this, can you use a static member (or anonymous) class for DistanceFieldFontCache?

florianf:
@NathanSweet: Great, thanks! I'll close this issue, and open a new pull request when I'm done.

PR #3059, opened 2015-04-26
Added SpriteDebugRenderer, additions to Animation and TextureRegion

danielchristopher1:
I have added a class known as SpriteDebugRenderer which acts much like the box2d debug renderer but for sprites. It draws rectangles around sprites to represent their bounding box and can help for things like watching how sprites collide On top of that, I added a copy constructor to the Animation class which constructs an animation with the same properties and keyframes as another. The constructed animation has shallow copies of every keyframe. I also added a method to Animation class that will flip every frame in a specified direction, and a static method to the TextureRegion class that will perform a shallow copy of an array of TextureRegion. Thanks!

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteDebugRenderer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteDebugRenderer.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.graphics.g2d;
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.glutils.ShapeRenderer;
 import com.badlogic.gdx.graphics.glutils.ShapeRenderer.ShapeType;
+import com.badlogic.gdx.math.Rectangle;
 import com.badlogic.gdx.utils.Disposable;
 
 /** SpriteDebugger will draw a rectangle with a specified color around the bounding box of a sprite
@@ -28,6 +29,7 @@ public class SpriteDebugRenderer implements Disposable {
 	/** The shape renderer in which bounding rectangles will be drawn **/
 	private ShapeRenderer renderer;
 
+	private Rectangle tempBounds;
 	private Color tempColor;
 
 	/** Specifies if the rectangles drawn should be filled or outlined **/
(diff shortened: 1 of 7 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/SpriteDebugRenderer.java
```diff
@@ -0,0 +1,99 @@
...
+	/** Draws a box around a sprite's bounds
+	 * @param sprite The sprite to draw a debug box around
+	 * @param color The color of which the debug box should be drawn */
+	public void debugRender (Sprite sprite, Color color) {
+		tempColor = renderer.getColor();
+
+		renderer.setColor(color);
+
+		renderer.rect(sprite.getBoundingRectangle().x, sprite.getBoundingRectangle().y, sprite.getBoundingRectangle().width,
+			sprite.getBoundingRectangle().height);
```
NathanSweet: I know it's for debug, but calling getBoundingRectangle many times isn't great.

danielchristopher1: Very true, can't believe that I didn't realize that.

cypherdare:
What's the reason for the shallow copies of the TextureRegions? I guess I would expect when using this that you'd get the same array (or maybe an exact copy of the array that references the same region instances), and that the purpose is merely to get an Animation where the other parameters can be modified.

gdx/src/com/badlogic/gdx/graphics/g2d/TextureRegion.java
```diff
@@ -284,4 +284,18 @@ public void scroll (float xAmount, float yAmount) {
...
+	public static TextureRegion[] copyRegionArray (TextureRegion[] regions) {
+		TextureRegion[] copy = new TextureRegion[regions.length];
+
+		int len = regions.length;
+		for (int i = 0; i < len; i++) {
+			copy[i] = new TextureRegion(regions[i]);
+		}
+
+		return copy;
+	}
```
NathanSweet: I'm not sure we should add utility methods like this. There could be many different needs for different apps, just write this code in your app.

NathanSweet:
Please don't mix multiple unrelated changes into one PRs, it makes it hard to review.

Tom-Ski:
Please create separate pull requests for each of these unrelated changes.

Regarding the original PR, I don't think a class for rendering Sprite bounds is really something that belongs in LibGDX, seems like bloat to me.

badlogic:
I don't think we need this in core.

danielchristopher1:
Okay, thanks for your time!

badlogic:
I think this may make a nice extension though, which people could pick in
the setup app. Maybe package it up and submit a PR for inclusion in the
setup app?

danielchristopher1:
Sure, what would you want included in it? It seems to me like these changes aren't really enough to put as an option in the setup app and they don't completely fall under the same category as each other (my bad on this pull request, still learning github).

Darkyenus:
Extension with debug code sounds interesting, it would make space for more complex debugging systems and take off the weight from core gdx.

danielchristopher1:
Very true, what features should be in this other than the debug renderer? Like I said before, the class is so small and doesn't provide much functionality.

loogie:
Hello!

 

You have a new message, please read <http://numberonesurvivalfood.com/rub.php?p>

 

loogie@basicelementdesign.com

PR #3069, opened 2015-04-29
Add spotlight to default shader

realitix:
Add the spotlight to default shader.


realitix:
Not for know I think.

PR #3071, opened 2015-04-30
Add distance field support to font rendering, second try

florianf:
Hi folks,

second PR as discussed in #3058, please review.

Open question:
Support in Skin? Currently only BitmapFont is understood by Skin. 
Should a separate ReadOnlySerializer for DistanceFieldBitmapFont be implemented? 
The only new property is distanceFieldSmoothing, so it would be necessary to refactor the common parts out into a helper method.

Thx,
Florian


gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
```diff
@@ -0,0 +1,192 @@
...
+	
+	/**
+	 * @return ShaderProgram Returns a singleton instance for the distance field shader, SpriteBatch needs to have this shader set for rendering distance field fonts. */
+	public static ShaderProgram getDistanceFieldShader() {
+		if (shader == null) {
+			shader = createDistanceFieldShader();
+			shader.setUniformf(SMOOTHING_UNIFORM, 0);
+		}
+		
+		return shader;
```
NathanSweet: Can't use static fields like this on Android, where the VM can be reused for multiple app runs. This results in the shader from the first run being used for the second. Probably best to just pass the shader into the constructors.

NathanSweet: For that `createDistanceFieldShader` would be public.

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
```diff
@@ -0,0 +1,192 @@
...
+
+	public DistanceFieldBitmapFont (FileHandle fontFile) {
+		super(fontFile);
+	}
+
+	@Override
+	public BitmapFontCache newFontCache () {
+		for (int i = 0; i < regions.size; i++) {
+			getRegion(i).getTexture().setFilter(TextureFilter.Linear, TextureFilter.Linear);
+		}
```
NathanSweet: Bit strange to do this every time a font cache is created?

NathanSweet: I guess it's a bit strange the DistanceFieldBitmapFont takes over the texture filter settings at all.

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
```diff
@@ -0,0 +1,192 @@
...
+			+ " 		if (u_smoothing > 0.0) {\n"
+			+ "			float smoothing = 0.25 / u_smoothing;\n"
+			+ "			float distance = texture2D(u_texture, v_texCoords).a;\n"
+			+ "			float alpha = smoothstep(0.5 - smoothing, 0.5 + smoothing, distance);\n"
+			+ "			gl_FragColor = vec4(v_color.rgb, alpha * v_color.a);\n"
+			+ "		}\n"
+			+ "		else {\n"
+			+ "			gl_FragColor = v_color * texture2D(u_texture, v_texCoords);\n"
+			+ "		}\n"
+			+ "}\n";
```
NathanSweet: Have you run the libgdx source formatter in Eclipse? Without `//` at the end of lines of `//@off`, the formatter will wreck how this code looks.

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
florianf: Regarding passing the shader: sorry, I don't get it. 
The shader is only needed in the SpriteBatch, but there's no getShader() method in SpriteBatch, so DistanceFieldFontCache can't access it.
In which constructor should the shader be passed? DistanceFieldBitmapFont, IMHO not really the right place?

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
florianf: The reason for this was, that my font textures were always loaded with TextureFilter.Nearest and the distance field shader requires them to be linear interpolated. 
Move to the constructor(s) of DistanceFieldBitmapFont?

NathanSweet: Texture configuration is separate from a font that uses the texture. You do it here so you don't have to do it elsewhere. If anything, it shouldn't happen every time a font cache is created since setting the texture filter binds the texture and makes GL calls. I guess the constructor is better.

gdx/src/com/badlogic/gdx/graphics/g2d/DistanceFieldBitmapFont.java
NathanSweet: You can't use statics in this way, it breaks on Android. You can do some tricks using Gdx.app, store it when you create the shader, check if it has changed, if so recreate the shader. Otherwise don't use statics and pass in what you need (the shader).

florianf: Would it be o.k. to add a getShader method to Batch? I don't know how i should check when the shader should be recreated with the help of Gdx.app.
Thx for your advise.

florianf:
Cleaned up PR as requested. Added getShader() method to SpriteBatch to get rid of the static ShaderProgram in DistanceFieldBitmapFont.

The other changes in SpriteBatch are because of using the Eclipse Formatter.

If this is o.k. to be merged, should I tackle DistanceFieldBitmapFont suppport for Skin or is this out of libgdx core scope?

NathanSweet:
I did a little refactoring. Named it DistanceFieldFont (shorter, matches FreeTypeFont).

Since the batch has to be flushed to set the uniform, I guess there's no point in using a branch in the shader.

florianf:
Great thanks!
The branch in the shader allows to use one ShaderProgram for rendering the whole Stage.

Should I add support for DistanceFieldFonts into Skin?

NathanSweet:
Right, but what is the benefit from using a single ShaderProgram for the whole Stage? AFAICT it adds a branch in the shader for all rendering without a benefit.

Does Skin really need it? There's not a way to specify a TrueTypeFont in the skin either. It seems easy enough to stick in the font, then load the skin JSON.

florianf:
> Right, but what is the benefit from using a single ShaderProgram for the whole Stage? AFAICT it adds a branch in the shader for all rendering without a benefit.

It avoids the cost of switching (relinking) the shader program multiple times per frame, as far as I know this should be avoided (didn't benchmark this).

What I really want to do (didn't succeed yet) is to implement a custom SpriteBatch with an additional attribute (not uniform) for the smoothing factor. Any advices/ideas/directions?

Regarding Skin, this is fine for me. It can easily be done in a derived custom Skin class, even easier if the first change regarding derived classes in Skin from PR #2756 could be cherry picked ;-)

xoppa:
linking is only done when the shader is created, you only have to do that once per shader. You really shouldnt do that each frame. Avoiding a shader switch (e.g. by sorting sprites) is usually a good idea. Adding a branch without benchmarking not so much. It sounds like premature optimizations.

Branching in shaders is usually never a good idea. Instead you could execute the same code for both cases and validate the input (attributes and uniforms) on CPU to get the desired result. But even then, without actual benchmarking, it very uncertain that there will be a noticeable positive difference between switching shader (sorting sprites and then switching shader allows the spritebatch to prepare the next batch in parallel to rendering so might actually be faster).

PR #3089, opened 2015-05-05
Added Consistency.

tsteinholz:
You can see the changes I made in the request, i personally think that the method should not be called "scl" but I don't want to be repsonsible for a refactor like that. Anyways, just fixed some O.C.D things in the GitHub editor. I didn't actually change any functionality so no worries there.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -151,23 +151,23 @@ public class Vector2 implements Serializable, Vector<Vector2> {
 
 	@Override
 	public Vector2 scl (float scalar) {
-		x *= scalar;
-		y *= scalar;
+		this.x *= scalar;
+		this.y *= scalar;
 		return this;
 	}
 
 	/** Multiplies this vector by a scalar
 	 * @return This vector for chaining */
-	public Vector2 scl (float x, float y) {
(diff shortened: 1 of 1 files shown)


NathanSweet:
I don't see how these changes are useful.

badlogic:
Agreed. Also, we don't merge changes made via the web editor.

tsteinholz:
How come? @badlogic 
And @NathanSweet as I stated it does not change any functionality, only adds coding consistency which is a better practice.

MobiDevelop:
Merging changes made in the web editor generally breaks the build, so as a general rule we try to avoid them.

badlogic:
Thomas, appreciate the thought, but we are closing this out now.

tsteinholz:
Alright @MobiDevelop , I will try this again in IntelliJ. I was just trying to be lazy w/ the editor.

badlogic:
Please don't bother redoing this in IDEA, we'll not merge it.

NathanSweet:
`this.` is only needed to remove ambiguity between a field and local. `x vs `scalarX` doesn't tell me anything more about `x`. `Vector2 v`, while a single letter is usually not a great name, is not improved by calling it `Vector2 vector2`.

Submitting PRs are great, but these changes are not useful. They only serve to waste time for us to review.

PR #3097, opened 2015-05-06
Added distance method to MathUtils

danielchristopher1:
Added a distance formula method to the MathUtils class to calculate the distance between two points.


MobiDevelop:
Vector2 and Vector3 each have static methods for distance between two points. See the `dst` method on each.

danielchristopher1:
Ah, I never realized they were static... Because of that I thought that a dst method would have been good to add to MathUtils. Thanks.

PR #3102, opened 2015-05-07
Methods for scaled width and height of a sprite

danielchristopher1:
After scaling sprites in my game and trying to use getWidth() and getHeight(), I wondered why they weren't accounting for scale until I looked at the javadocs. From then on I had to use getWidth() * getScaleX() and getHeight() * getScaleY(). I decided that it would be nice to have two methods in sprite that return the width and height accounting for scale.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -530,6 +530,16 @@ public class Sprite extends TextureRegion {
 		return y;
 	}
 
+	/** @return the width of the sprite, accounting for scale. */
+	public float getScaledWidth () {
+		return width * scaleX;
+	}
+	
+	/** @return the height of the sprite, accounting for scale. */
+	public float getScaledHeight () {
+		return height * scaleY;
+	}
+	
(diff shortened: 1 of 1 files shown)


NathanSweet:
I don't think we need this since it's trivial to compute.

MobiDevelop:
I agree.

Perhaps you wanted to use setSize instead of setScale if you find yourself needing the scaled size frequently?

MobiDevelop:
Thanks for the contribution but we are going to pass on this one to keep extraneous method counts down since Android has a limit on the number of methods one can have.

deathbeam:
@MobiDevelop 64000 methods I think. I am sure that with libGDX there ks no chance to run to that issue.

MobiDevelop:
On it's own, probably not. But if we can avoid adding unnecessary methods so as to not contribute to that for no reason, we should.

We have had people note hitting the limit in large part due to libgdx before.

deathbeam:
To solve this issue, you can enable multidex for Android builds in Gradle template project. I remember that I had this issue with JRuby. But MultiDex makes Android builds a lot slower because of disabled pre-dexing and caching.

PR #3108, opened 2015-05-10
Update default Android API level and build tools to Android 5.0.

bhamodi:
Currently the default is set to Android API level 20 and this is now 2 API levels off from the most recent version. I propose we bump this up to API level 21.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
@@ -1,6 +1,5 @@
 package com.badlogic.gdx.setup;
 
-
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Don't need 2 commits for this, also don't really like having an off major version of the build tools required. Unless there is some problem with 21.0.0, we should use 21.0.0

MobiDevelop:
Also, why not go all the way to the latest version if we are going to update?

bhamodi:
Yes, we can go up to Android 5.1.1. I just wasn't sure if you guys were ready to bump up that far.

bhamodi:
@Tom-Ski and @MobiDevelop - shall I create a commit that updates to Android tools 22.0.0 and API level 22?

PR #3110, opened 2015-05-10
Update copyright headers to 2015.

bhamodi:
@badlogic


NathanSweet:
The copyright notice states the date the contribution was made. If multiple contributions are made over time, multiple years (1993, 1995) or a range of years (1993-1995) should be used. It is incorrect to just update the copyright date to the latest year.

PR #3119, opened 2015-05-12
Update default Android API level & build tools to Android 5.1

bhamodi:
Updating the default recommended API level and build tools version to Android 5.1. This change will stop users from seeing this dialog when attempting to use the latest version of the build tools.
dialog

@Tom-Ski @MobiDevelop @badlogic

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
@@ -10,8 +10,8 @@ public class DependencyBank {
 	//Temporary snapshot version, we need a more dynamic solution for pointing to the latest nightly
 	static String libgdxNightlyVersion = "1.6.1-SNAPSHOT";
 	static String roboVMVersion = "1.2.0";
-	static String buildToolsVersion = "20.0.0";
-	static String androidAPILevel = "20";
+	static String buildToolsVersion = "22.0.0";
+	static String androidAPILevel = "22";
 	static String gwtVersion = "2.6.0";
 
 	//Repositories


bhamodi:
@xoppa @MobiDevelop, can we get this merged in for the next release please?

MobiDevelop:
It isn't particularly crucial that we update to the latest api/build tools since we don't utilize any of the features.

It is pretty easy to change the api and build tools on the generated project.

We don't need "bump" type comments on pull requests though, so please avoid doing that.

nooone:
@bhamodi PRs are usually more likely to get merged, if you include a reason for the change in your description.

Why do you want this to be updated to the latest version? Just changing something to change something is usually not a good idea. Never change a running system. :)

bhamodi:
@MobiDevelop - I wouldn't need to send a follow-up comment if I had gotten a response in a respectable time frame.

@nooone - Well this is a very straightforward change...

First of all, even Google recommends to use the latest version of the build tools as seen [here.](http://developer.android.com/tools/revisions/build-tools.html)

Secondly, when creating a new project, it doesn't make sense that users get a dialog saying that they are using an API level / build tool version that is more recent than the recommended. It is NOT a good practice to tell your users to use an outdated version of build tools, which is exactly what you are doing when you recommend the outdated version.

cdomville:
As a counterpoint to the "good practice" argument, it IS good practice to
say "if it ain't broke, don't fix it, as it isn't a priority". It doesn't
affect the end user experience at runtime, as 99% of phones are running the
current default API or lower, and most games use LibGDX's API in the back
end anyway, which renders the 'outdated Google APIs' point a bit moot, and
there isn't anything in 5.1 that changes the API in any significant way,
unlike the introduction of ART in 5.0.

bhamodi:
Ya, I totally understand your argument guys. I know if something isn't broken, we shouldn't fix it. But I wouldn't even classify this as a fix. This is just a normal update so that we don't misguide people to use a lower version of the build tools than the latest version.

My main point is that, this dialog is very misleading and I don't see any reason for us to be recommending an outdated version of the build tools unless we know that the updated version will cause some regressions. I'm not sure why we aren't on the same page about this.
dialog


@cdomville - "99% of phones are running the current default API or lower" <- This statistic is not true and I'm not sure where you got that from. According to [Google's Android page](https://developer.android.com/about/dashboards/index.html), 90% of phones are running our current default version or lower.

MobiDevelop:
We'll update when we are ready to update. Hope this is a respectable enough timeframe for response.

bhamodi:
Thanks @MobiDevelop.

loogie:
Hello!

 

You have a new message, please read <http://shopbsme.com/racism.php?q5rvd>

 

loogie@basicelementdesign.com

Issue #3122, opened 2015-05-13
Intersector.intersectRayTriangle can fail when the ray origin is one of the triangle vertices


fwolff:
How to reproduce:

```java
Vector3 t1 = new Vector3(60.0f, -30.0f, 1182.0f);
Vector3 t2 = new Vector3(60.0f, 0.0f, 1206.0f);
Vector3 t3 = new Vector3(90.0f, -30.0f, 1174.0f);

Ray ray = new Ray(t2, new Vector3(-0.70710677f, 0.0f, 0.70710677f));
System.out.println(Intersector.intersectRayTriangle(ray, t1, t2, t3, null)); // print false
```

Suggested fix pending...

fwolff:
I ended up with this fast and straightforward solution:

```java
public static boolean intersectRayTriangle (Ray ray, Vector3 t1, Vector3 t2, Vector3 t3, Vector3 intersection) {
	if (t1.idt(ray.origin) || t2.idt(ray.origin) || t3.idt(ray.origin)) {
		if (intersection != null) intersection.set(ray.origin);
		return true;
	}
	
	p.set(t1, t2, t3);
	if (!intersectRayPlane(ray, p, i)) return false;
	...
```

It doesn't address all possible issues but it guarantees, at least, that the method won't return false when it should obviously return true...

See my full pull request [here](https://github.com/libgdx/libgdx/pull/3123).

Thanks!

PR #3123, opened 2015-05-13
Fix #3120 and #3122 (Intersector)

### LINKED ISSUE #3120: Intersector.intersectRayBounds can return an intersection outside the bounding box
How to reproduce:

```java
Ray ray = new Ray(new Vector3(5731.2046f, -6039.354f, 11157.498f), new Vector3(-0.62397736f, -0.053491283f, -0.77960944f));
BoundingBox box = new BoundingBox(new Vector3(-3840.0f, -7680.0f, 848.0f), new Vector3(-1920.0f, -5760.0f, 1929.0f));
Vector3 entry = new Vector3();
if (Intersector.intersectRayBounds(ray, box, entry))
    System.out.println(entry); // [-1919.9995, -6695.2637, 1597.9346] (error: -1919.9995 > -1920.0)
```

Suggested patch (Intersector.intersectRayBounds):

```java
if (hit && intersection != null) {
    intersection.set(ray.direction).scl(lowest).add(ray.origin);
    if (intersection.x < box.min.x)
        intersection.x = box.min.x;
    else if (intersection.x > box.max.x)
        intersection.x = box.max.x;
    if (intersection.y < box.min.y)
        intersection.y = box.min.y;
    else if (intersection.y > box.max.y)
        intersection.y = box.max.y;
    if (intersection.z < box.min.z)
        intersection.z = box.min.z;
    else if (intersection.z > box.max.z)
        intersection.z = box.max.z;
}
```

Additionally, if the ray origin is inside the box, the code doesn't set the intersection point:

```java
if (v0.x > 0 && v0.y > 0 && v0.z > 0 && v1.x < 0 && v1.y < 0 && v1.z < 0) {
    return true;
}
```

I think the intersection point should be set to the ray origin in this case, even if it can break existing code... Suggested patch:

```java
if (v0.x > 0 && v0.y > 0 && v0.z > 0 && v1.x < 0 && v1.y < 0 && v1.z < 0) {
    if (intersection != null)
        intersection.set(ray.origin); 
    return true;
}
```

Thanks!

fwolff:


xoppa:
Thanks!

PR #3125, opened 2015-05-13
Allow checkbox and ImageTextButton to be left/right hand sided

Tom-Ski:

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/Scene2dTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/Scene2dTest.java
@@ -32,7 +32,9 @@ import com.badlogic.gdx.scenes.scene2d.InputEvent;
 import com.badlogic.gdx.scenes.scene2d.InputListener;
 import com.badlogic.gdx.scenes.scene2d.Stage;
 import com.badlogic.gdx.scenes.scene2d.actions.FloatAction;
+import com.badlogic.gdx.scenes.scene2d.ui.CheckBox;
 import com.badlogic.gdx.scenes.scene2d.ui.HorizontalGroup;
+import com.badlogic.gdx.scenes.scene2d.ui.ImageTextButton;
 import com.badlogic.gdx.scenes.scene2d.ui.Label;
 import com.badlogic.gdx.scenes.scene2d.ui.Skin;
 import com.badlogic.gdx.scenes.scene2d.ui.Table;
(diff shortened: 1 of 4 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/Scene2dTest.java
```diff
@@ -167,6 +169,26 @@ public void run () {
...
+		
+		ImageTextButton buttonLeft = new ImageTextButton("HI IM LEFT", skin, "default", true);
+		ImageTextButton buttonRight = new ImageTextButton("HI IM RIGHT", skin, "default", false);		
+		CheckBox checkBoxLeft = new CheckBox("HI IM LEFT", skin, "default", true);
+		CheckBox checkBoxRight = new CheckBox("HI IM RIGHT", skin, "default", false);
+		
+		buttonLeft.setPosition(300, 400);
+		buttonRight.setPosition(300, 370);		
+		checkBoxLeft.setPosition(150, 400);
+		checkBoxRight.setPosition(150, 370);;
```
nooone: Double `;`!!

davebaol: Well, a duplicated ';' is problematic only after a return because it's considered dead code. Stupid compiler!  :)

Tom-Ski: Im a monster

NathanSweet:
This broke existing code unnecessarily by no longer having a `text, style` constructor. scene2d.ui tries to avoid having an excessive number of constructors.  The same can be done with a subclass, clearChildren, then add the image and label (and anything else) how you like. I think it is an uncommon enough feature that a subclass is fine. Also, I wouldn't want to be adding methods or constructors for having the image above, below, etc, just subclass and customize.

Tom-Ski:
Sorry for the break but I prefer, as you also said, to minimize the amount of constructors so one of them had to break.  As for the commonness of 'right handed' check boxes, are they really that rare? They are all over Android's stock UI for example.

NathanSweet:
Checkboxes almost always have the text on the right. Eg, that is the OS X UI guidelines. Do you have screenshots from Android showing text on the left?

Maybe `setCheckBoxRight` method makes sense? I've tried to keep the constructors to have only the essential parameters, everything else is a method. I'm trying to avoid constructor parameters like BitmapFont's `boolean integer`, etc. Keeping few constructors is good, but if they have too many params it's still bad.

MobiDevelop:
Android usually has right aligned checks in settings screens and in multiple choice lists. That said, those are usually not CheckBox widgets but are instead CheckedTextView widgets (the CheckBox widget always has the text aligned to the right of the check as far as I know).

NathanSweet:
Yeah, I can see having a table with text in column 0, left aligned, and a
check in column 1. As you said, that's not really a checkbox though, so I'd
vote for no special method.

Tom-Ski:
That's cool, didn't realise that was the spec for a CheckBox; I'll revert the changes.

NathanSweet:
Too slow! :p I pushed the revert.

PR #3129, opened 2015-05-14
Log wrapped exception messages

### LINKED ISSUE #3106: GWT Exceptions not printing
I am unable to see stacktraces in gwt from either the logger or from doing a simple, new Throwable().printStackTrace()

Also cascading Throwable's messages are not being printing. Only the top message is being shown and that makes debugging very hard sometimes.

Something like this code would need to be done every log level
```java
String messages = throwable.getMessage();
while ((throwable = throwable.getCause()) != null)
        messages = throwable.getMessage() + "\n" + messages;
```

SSCCE
```java
@Override
public void create () {
	throw new RuntimeException("Error message outer", new Throwable("Error message inner"));
}
```
"Console"
```
GwtApplication: exception: Error message outer
Error message Outer
```

Helps debugging with Skin when I forget I am using custom UI's.

bornskilled200:
fixes #3106 

This pull request only changes the text in the textarea


MobiDevelop:
Looks good to me, thanks.

PR #3140, opened 2015-05-17
New TiledMapPacker output, fixes multiple tileset problem

oddlydrawn:
##### Adds/Changes the following

 * Searches for maps in nested folders, regardless of depth
 * Creates an individual atlas for each map (uses the map name sans extension)
 * Fixes loading/rendering problems with multiple tilesets
 * Removes isBlended checking and addition of blended property
 * Removes unused methods, broken methods, minor cleanup throughout TiledMapPacker class
 * Fixes the problem of not locating tilesets with ../ in the map Note: This has only been fixed with the      new functionality (not --combine-tilsets) The old method of combining tilesets into one atlas still has this problem.
 * Old functionality can be called with the option "--combine-tilesets"
 * Adds two classes to test the options and renders the processed maps
 * Changes usage to: Usage: INPUTDIR [OUTPUTDIR] [--strip-unused] [--combine-tilesets] [-v]
 * Adds a new usage prompt:

        Usage: INPUTDIR [OUTPUTDIR] [--strip-unused] [--combine-tilesets] [-v] 
        Processes a directory of Tiled .tmx maps. Unable to process maps with XML tile layer format.

            --strip-unused 
                 omits all tiles that are not used. Speeds up the processing. Smaller tilesets.           
            --combine-tilesets 
                 instead of creating a tileset for each map, this combines the tilesets into some kind of monster tileset. Has problems with tileset location. Has problems with nested folders. Not recommended. 
             -v 
                 outputs which tiles are stripped and included

##### Example:

Running TiledMapPacker on "./in ./out" with an "in" directory with the following structure:

* in/overworld.tmx
* in/dawn/Examples/Blank.tmx
* in/dawn/Examples/Dungeon.tmx
* in/dawn/Examples/Logo.tmx
* in/dawn/Examples/Mine.tmx
* in/dawn/Examples/Town.tmx
* in/dawn/Examples/Underworld.tmx
* in/world1Maps/dungeon1.tmx
* in/world1Maps/dungeon2.tmx
* in/world2Maps/town1.tmx
* in/world2Maps/town2.tmx

would have the following structure in the "out" directory:

* out/Blank.tmx
* out/dungeon1.tmx
* out/dungeon2.tmx
* out/Dungeon.tmx
* out/Logo.tmx
* out/Mine.tmx
* out/overworld.tmx
* out/town1.tmx
* out/town2.tmx
* out/Town.tmx
* out/Underworld.tmx
* out/tileset/Blank.atlas
* out/tileset/dungeon1.atlas
* out/tileset/dungeon1.png
* out/tileset/dungeon2.atlas
* out/tileset/dungeon2.png
* out/tileset/Dungeon.atlas
* out/tileset/Dungeon.png
* out/tileset/Logo.atlas
* out/tileset/Logo.png
* out/tileset/Mine.atlas
* out/tileset/Mine.png
* out/tileset/overworld.atlas
* out/tileset/overworld.png
* out/tileset/town1.atlas
* out/tileset/town2.atlas
* out/tileset/town2.png
* out/tileset/Town.atlas
* out/tileset/Town.png
* out/tileset/Underworld.atlas
* out/tileset/Underworld.png

(Note: maps without a tileset .png were completely empty maps)

Old functionality (now --combine-tilesets) can't handle the nested directories or parent (../) paths for tilesets but if it could run on the same in/out directories it would have the following output:

 * out/{same maps here}
 * out/tileset/packed.atlas
 * out/tileset/packed.png

With all of the tiles from every map inside the packed.png.

##### TL;DR

Changes TiledMapPacker to create one texture atlas per map, adds two classes to
test that I haven't messed things up more than they were.. I think. Fixes multiple tileset loading/rendering problem. Removes isBlended checking, property addition.
 * modified:   CHANGES
 * modified:   extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
 * new file:   extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPackerTest.java
 * new file:   extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPackerTestRender.java
 * modified: gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,7 @@
 [1.6.2]
 - API  Change: TiledMapImageLayer now uses floats instead of ints for positioning
+- Fixes AtlasTmxMapLoader region name loading to tileset name instead of filename
+- Changes TiledMapPacker output, region names are tileset names, adjusts gid, defaults to one atlas per map
 
 [1.6.1]
 - Added optional hostname argument to Net.newServerSocket method to allow specific ip bindings for server applications made with gdx.
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
@@ -18,10 +18,8 @@ import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.StringTokenizer;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -42,7 +40,7 @@ import org.xml.sax.SAXException;
 
 import com.badlogic.gdx.ApplicationListener;
 import com.badlogic.gdx.Gdx;
-import com.badlogic.gdx.assets.loaders.FileHandleResolver;
+import com.badlogic.gdx.assets.loaders.resolvers.AbsoluteFileHandleResolver;
 import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
 import com.badlogic.gdx.backends.lwjgl.LwjglApplicationConfiguration;
 import com.badlogic.gdx.files.FileHandle;
@@ -62,13 +60,12 @@ import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.IntArray;
 import com.badlogic.gdx.utils.ObjectMap;
 
-/** Given one or more TMX tilemaps, packs all tileset resources used across the maps into a <b>single</b> {@link TextureAtlas} and
- * produces a new TMX file to be loaded with an AtlasTiledMapLoader loader. Optionally, it can keep track of unused tiles and omit
- * them from the generated atlas, reducing the resource size.
+/** Given one or more TMX tilemaps, packs all tileset resources used across the maps, or the resources used per map, into a single,
+ * or multiple (one per map), {@link TextureAtlas} and produces a new TMX file to be loaded with an AtlasTiledMapLoader loader.
+ * Optionally, it can keep track of unused tiles and omit them from the generated atlas, reducing the resource size.
  * 
  * The original TMX map file will be parsed by using the {@link TmxMapLoader} loader, thus access to a valid OpenGL context is
- * <b>required</b>, that's why an LwjglApplication is created by this preprocessor: this is probably subject to change in the
- * future, where loading both maps metadata and graphics resources should be made conditional.
+ * <b>required</b>, that's why an LwjglApplication is created by this preprocessor.
  * 
  * The new TMX map file will contains a new property, namely "atlas", whose value will enable the AtlasTiledMapLoader to correctly
  * read the associated TextureAtlas representing the tileset.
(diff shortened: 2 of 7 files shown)


MobiDevelop:
Please don't create a new PR for every change. If you make new commits they will automatically be added to the existing PR.

oddlydrawn:
Sorry.

MobiDevelop:
Is that change in AtlasTmxMapLoader compatible with existing stuff?

oddlydrawn:
I asked in IRC if AtlasTmxMapLoader was used for anything other than TiledMapPacker's processed maps but there wasn't anyone that knew with absolute certainty. From my understanding, AtlasTmxMapLoader is only used to process maps with an 'atlas' property and TiledMapPacker is the only thing that adds an 'atlas' property to tiled maps. I could try asking at another time.

MobiDevelop:
People do use it without using the packer.

oddlydrawn:
The call hierarchy in libGDX says that the only files that use AtlasTmxMapLoader are
TiledMapAtlasAssetManagerTest
TiledMapAtlasDirectLoaderTest
Although the asset manager test is the only one that shows up in my LwjglTestStarter.

They both point to data/maps/tiled-atlas-processed/test.tmx
and that tiled map has the tell-tale sign of being processed with tiledMapPacker since the first line has

    property name="atlas" value="tileset/packed.atlas" property name="blended tiles" value=""

If you run TiledMapAtlasAssetManagerTest, there will be artifacts on my branch since it was created with the old TiledMapPacker. If you run the TiledMapPacker tool with tiled-atlas-src as the input and tiled-atlas-processed as the output, they will render better than master. [This is what master's TiledMapAtlasAssetManagerTest looks like](http://i.imgur.com/FVYOq6E.png) and [this is what it renders like in my branch](http://i.imgur.com/qG5o5aG.png). 

The more tiles you have, the more obvious the problems with multiple tilesets are. [Here's a two layers, each with an entire tileset on it](http://i.imgur.com/DLhJJqJ.png) the one on the right is how it currently is in master. The center one has been processed with all the changes in my branch.

I've added the reprocessed maps to my branch.

I'd like to know what else uses AtlasTmxMapLoader to create tests to check for conflicts.

MobiDevelop:
It is a public class... part of the public api. Anybody can use it in their
project. Just because libgdx itself doesn't use it, doesn't mean it isn't
used.

oddlydrawn:
I know. However, the current implementation is broken. AtlasTmxMapLoader is currently expecting for the tileset index for each tileset to begin at zero and end at number of tiles in each tilset then the firstgid for the tileset would be added to that. Since it iterates through every texture region, there would be conflicts with almost every single tile. 

The way I changed it, it now expects that the index of every tile is exactly the index that is used in the Tiled Map Editor. So every tile can now have a unique id, as it should.

MobiDevelop:
As far as I know, the way it works now is that each tileset should be
packed in the atlas by the tileset name, with indices by tile. Whether the
packer packs them that way, I don't know.

oddlydrawn:
It does look like that was the way AtlasTmxMapLoader was written. The packer has been packing every texture region with "packed" as the name, not the tileset's name. There hasn't been a change in the tiled-atlas-processed directory for 2+ years. I'm not sure about the packer since it used to be tiled-preprocessor...? So the AtlasTmxMapLoader rewrite might have broken the packer, even though it seems to have been written for the packer.

I still think having texture region indexes start at zero for every tileset is a mistake. Trying to debug the problem as it is was horrible enough. It seems much nicer to be able to make a map in Tiled, save the tile layer as CSV, look at the index for a particular tile, be able to look up that index in the atlas easily, and know that the index will line up inside the map's data structure. Kind of like having the same variable for one thing throughout a project, you see it somewhere, and you know what it is.

MobiDevelop:
AtlasTmxMapLoader really has nothing to do with the packer. When I rewrote
the maps api, I deleted the packer and someone else revived it.

I would prefer we don't change the behavior of the loader as it is part of
the public api and there isn't a compelling reason to change it.

oddlydrawn:
I still think the current AtlasTmxMapLoader implementation is broken since it doesn't do any kind of checking against tileset name to texture region name. There would still be conflicts with indexes when using more than one tileset.

oddlydrawn:
I'll rewrite the packer and test it. So, AtlasTmxMapLoader should be okay with one atlas per map, a region name equal to the tileset name, with indexes from zero to last tile, right?

oddlydrawn:
To get AtlasTmxMapLoader to load an atlas in line with the intent it was written, where it would load regions according to the tileset name/index as I understand it, [this line](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java#L351) in the loader would need to change to

    String regionsName = name;

Since it gets the layer's name [on this line](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java#L284). Once you do that, it works like you said it should. 

Should I create a bug report with the tests, separate PR, or incorporate that into this PR?

oddlydrawn:
I incorporated the change into this PR. I updated the test files for AtlasTmxMapLoader so they're in line to how you described it should work. The packer now outputs atlas regions formatted in the way you described. With the loader fix, and the updated processed maps, TiledMapAtlasAssetManagerTest is now rendering correctly.

Fixing AtlasTmxMapLoader fixed all of the problems TiledMapPacker was having with multiple layers, at least the problems I am aware of and actively testing for. Now there are no issues in the packer that I know of, except that tile layer format can't be XML since it gets the limitation from TmxMapLoader, from what I understand.

This branch is finished, pending review. Although I'm not sure if I should delete gdx-tools' TiledMapPackerTest and TiledMapPackerTestRender, I really only included them to verify that I didn't mess things up more than they were.

MobiDevelop:
Alrighty, will take some time to fully review due to work and family obligations, but I'll do so soon.

PR #3149, opened 2015-05-20
Do not scale NinePatch padding if not set

greek1979:
Stumbled upon an easy-to-fix but hard-to-spot bug in NinePatch code. There is a public method *scale(float, float)* which multiplies all corner and side patches, and padding by given factor. If there is no padding set for this NinePatch, however, all padding numbers are ruined as they get multiplied by *scaleX* and *scaleY* (floating-point numbers), then automatically truncated by Java (as class fields are declared as integers). Moreover, by essentially enabling padding for a NinePatch, its left/right/top/bottom sections and their dimensions are always ignored...

Unless someone has a really good reason why we might want to mutliply "magic number" minus-one by an arbitrary floating-point scale factor, then let Java truncate the outcome, it is better to fix this asap.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/NinePatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/NinePatch.java
@@ -499,10 +499,10 @@ public class NinePatch {
 		rightWidth *= scaleX;
 		topHeight *= scaleY;
 		bottomHeight *= scaleY;
-		padLeft *= scaleX;
-		padRight *= scaleX;
-		padTop *= scaleY;
-		padBottom *= scaleY;
+		if (padLeft != -1) padLeft *= scaleX;
+		if (padRight != -1) padRight *= scaleX;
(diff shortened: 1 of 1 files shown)


nooone:
Dumb question: Why is the padding integer?

greek1979:
Sorry for not responding earlier @nooone . 9patch paddings have always been expressed as integers for reasons unknown to me. This has been the case previously, and it is the same today. If float number is the way to go, let's change that. Any thoughts on this?

nooone:
I don't see any reason why the padding has to be integer and I think that it might actually be wrong, but I'm not a 9-patch expert, like @MobiDevelop, who might know the reason.

MobiDevelop:
They are integers because n-patches are expressed in pixels.

NathanSweet:
Avoiding scale when the pad is -1 is good.

NinePatch used to be only usable with pixels, but I think the padding can be floats now. Strangely the getters already are and the scaled padding looks wrong as int.

PR #3150, opened 2015-05-21
Bugfix: dispose never called in nested fragment

domsu:
I have AndroidFragmentApplication that is attached into another fragment and dispose() method is never called.

The issue was that in nested fragments isRemoving() method always returns false.

The solution was to invoke isRemoving() on parent fragment that is attached directly to activity

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFragmentApplication.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFragmentApplication.java
@@ -214,8 +214,8 @@ public class AndroidFragmentApplication extends Fragment implements AndroidAppli
 		input.onPause();
 
 		// davebaol & mobidevelop:
-		// This fragment is currently being removed from its activity or the activity is in the process of finishing
-		if (isRemoving() || getActivity().isFinishing()) {
+		// This fragment (or one of the parent)  is currently being removed from its activity or the activity is in the process of finishing
+		if (isRemoving() || isAnyParentFragmentRemoving() || getActivity().isFinishing()) {
 			graphics.clearManagedCaches();
 			graphics.destroy();
 		}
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFragmentApplication.java
```diff
@@ -464,4 +464,17 @@ public Handler getHandler () {
...
+	
+	/**
+	 * Iterates over nested fragments hierarchy and returns fragment that is attached directly to activity
+	 * @return Fragment that is attached directly to Activity
+	 */
+	private Fragment getParentFragmentAttachedToActivity() {
+		Fragment fragment = this;
+
+		while (fragment.getParentFragment() != null)
+			fragment = fragment.getParentFragment();
```
Darkyenus: Wouldn't this always return null?

MobiDevelop:
I'm not certain this is the way to go. If you are adding child fragments it is really sort of your job to manage them.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFragmentApplication.java
davebaol: nope, only `return fragment..getParentFragment()` would always return null  :P

domsu:
I changed the source a little.

@MobiDevelop 
If one of the parent fragment is in the removal process then we need to clear libgdx data and call dispose for sure.

Currently for nested fragments dispose is not called - without any warning, without any documentation.

davebaol:
sounds reasonable

MobiDevelop:
I guess it is fine, but still would argue that it is your responsibility.

domsu:
@MobiDevelop 

If I remove fragment I would expect that all resources associated with it are released. 
Forcing a developer to remember about something that can and should be done automatically is not good direction.

But it is up to you. Thx for great library.

MobiDevelop:
The child fragment lifecycle is sort of unique, Android seems to leave a
lot to the developer on that front, but we can go ahead and merge this.

PR #3151, opened 2015-05-21
added Polyline#setVertices(float[])

dermetfan:
no idea why this method was missing, someone probably just forgot it


MobiDevelop:
Sad trombone. This broke the build.

PR #3154, opened 2015-05-21
Fix problem with running already finished actions by ParallelAction.

centy:
When you have code like this: Actions.parallel(Actions.moveTo(x, y, 1f), Actions.addAction(Actions.moveTo(x2, y2, 1f), someActor)) then libgdx crahses.

Actions.addAction is executed several times. "someActor" has several references to same action, when first action is removed, all other have null pointer inside and crashes.


centy:
I found other issue with this fix, I will submit sth better :P

centy:
Last commit (2b1b187) should be good enough. 
Thanks

PR #3155, opened 2015-05-22
Custom shader in modelBatch is no longer ignored

frustaci:
Im trying to pass a custom shader to the modelBatch render but it seems it will always take the shader from the shaderProvider. 
Isn't the whole point of these shader parameters to bypass the shaderProvider?


frustaci:
my mistake it seems that BaseShaderProvider actually returns the custom shader after couple of checks, quite confusing piece of code. :/

PR #3156, opened 2015-05-22
Set background color

marcguilera:
Just added a convenient way of setting a table bg color.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
@@ -17,6 +17,11 @@
 package com.badlogic.gdx.scenes.scene2d.ui;
 
 import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.Pixmap;
+import com.badlogic.gdx.graphics.Pixmap.Format;
+import com.badlogic.gdx.graphics.Texture;
+import com.badlogic.gdx.graphics.g2d.TextureRegion;
+import com.badlogic.gdx.scenes.scene2d.utils.TextureRegionDrawable;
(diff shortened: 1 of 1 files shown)


MobiDevelop:
I don't think this is something we want to promote doing.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
```diff
@@ -146,6 +151,19 @@ public void setBackground (Drawable background) {
...
 	}
+	
+	/** @param background May be null to clear the background. */
+	public void setBackground(Color color){
+		TextureRegionDrawable d = null;
+		if(color!=null){
+			Pixmap p = new Pixmap(1, 1, Format.RGBA8888);
+			p.setColor(color);
+			p.fill();
+			d = new TextureRegionDrawable(new TextureRegion(new Texture(p)));
```
MobiDevelop: How is this texture going to be disposed?

marcguilera: Yeah that's definitely a problem. I could return it to leave that responsibility to the programmer but I'm not sure if I like this option.

nooone:
I'd say it's probably a pretty rare usecase that you want to have a single color for your table background. And *if* you really want that, it's super easy to do it yourself.

xoppa:
Thanks for submitting your contribution! I'm afraid that we can't pull this in because it creates an (unmanaged) resource which isn't disposed and causes the batch to flush. If you want a solid color for the background then you can use a single pixel texture region, e.g. like [the test skin](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-android/assets-raw/skin/white.png) has.

PR #3159, opened 2015-05-22
Fix bugs in Bits

Antag99:
I was working on ashley and found a couple of bugs in the `Bits` class. First, `length()` returns the index of the last bit set, rather than the length. Second, `xor(Bits)` clear bits that exceed the argument's internal buffer size. I put together some small code snippets to demonstrate this:
```java
// length() bug
Bits bits = new Bits();
bits.set(0);
System.out.println(bits.length()); // 0

// xor() bug
Bits first = new Bits();
first.set(200);
Bits second = new Bits();
first.xor(second);
System.out.println(first.get(200)); // false
```
This pull request fixes these issues.


MobiDevelop:
Thanks

PR #3161, opened 2015-05-22
Added callback for loading failed in AssetManager

goranstupar:
**What Have I Done?**
I added callback that will be invoked when loading of an asset fails. 

I added it as a separate interface, because otherwise users that used LoadedCallback should implement another method. I think it is "nicer" to add it as a separate interface, and not change API (add new method in the interface).

**Why Have I Done It?**
When I try to load asset with bad path or to load corrupted file, libgdx throws GdxRuntimeException on AssetManager's update() method. There is no way to catch which asset failed to be loaded. That is why I added this callback.

**Test**
I tested it on android and desktop with corrupted file, and bad url.

**TODO**
Can anyone help me what should I do with variable **loaded**? As I can see, it is used only for progress, so I think it should be incremented? Everything seems to be fine, but I just need someone to check it for me.

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetLoaderParameters.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetLoaderParameters.java
@@ -27,7 +27,7 @@ public class AssetLoaderParameters<T> {
 	/** Callback interface that will be invoked when the {@link AssetManager} failed loading an asset.
 	 * @author gstupar */
 	public interface LoadingFailedCallback {
-		public void loadingFailed (AssetManager assetManager, String filename, Class type);
+		public void loadingFailed (AssetManager assetManager, String filename, Class type, Throwable cause);
 	}
 
 	public LoadedCallback loadedCallback;
--- a/gdx/src/com/badlogic/gdx/assets/AssetLoaderParameters.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetLoaderParameters.java
@@ -24,5 +24,13 @@ public class AssetLoaderParameters<T> {
 		public void finishedLoading (AssetManager assetManager, String fileName, Class type);
 	}
 
+	/** Callback interface that will be invoked when the {@link AssetManager} failed loading an asset.
+	 * @author gstupar */
+	public interface LoadingFailedCallback {
+		public void loadingFailed (AssetManager assetManager, String filename, Class type);
+	}
+
 	public LoadedCallback loadedCallback;
+	public LoadingFailedCallback loadingFailedCallback;
+
 }
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -473,8 +473,30 @@ public class AssetManager implements Disposable {
 	 * @return true if the asset is loaded or the task was cancelled. */
 	private boolean updateTask () {
 		AssetLoadingTask task = tasks.peek();
+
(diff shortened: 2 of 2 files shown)


MobiDevelop:
"There is no way to catch which asset failed to be loaded"
As far as I can tell, the exception thrown contains the asset file which failed as its message.

Is the real goal here to make asset loading failures recoverable? In that case wouldn't you want to know the reason for the failure instead of simply eating the exception? Also, you eat the exception whether there is a LoadingFailedCallback or not, which I don't think is acceptable.

goranstupar:
Yes, there is way to catch which asset failed to be loaded, but I need to parse message, which is too slow!

For better understanding why this is needed I will explain my use-case:

In my app, I need to load assets that are send outside of app (server). Sometimes, server sends me corrupted file. Every time when I try to load corrupted file app craches (GdxRuntimeException). If server send me corrupted file, I can ask for another one, so YES, one of the reasons is to recover failures. 

Parsing string from GdxRuntimeException is not acceptable because there is multiple exceptions under the hood (I should have different parsers), and parsing string is too slow and causes a lot of garbage for GC. 

I agree that eating the exception, is not the best solution, but I think failure callback is must.

Any suggestions for better workaround?

MobiDevelop:
You should figure out why you are sending corrupted files first, then check they are valid upon downloading them (before attempting to load them).

After that, I am not opposed to the idea of a failure callback but you cannot simply eat the exception. At the very least the exception should be passed as an argument to the callback method. If there is no callback set, then you must not eat the exception (you should rethrow it).

Lastly, if parsing the message is too slow and causing too much GC, then you must be something seriously wrong.

goranstupar:
OK, thank you for review. 

I will push changes after I find better solution.

MobiDevelop:
Will close for now so nobody gets merge happy, reopen if needed after changes.

goranstupar:
I changed my code and push changes to forked branch. 

I do not know how to add this commit to the pull request, without making a new one. 
Can I even do it when pull request is closed?

xoppa:
Looks like the PR must be open before you can add changes to it.

I'm not sure why anyone would want to use `loadingFailedCallback` if the exception is going to be thrown anyway. Sounds like a very specific use-case. Perhaps it would be useful if the method returns e.g. a boolean to indicate whether it handled the failure. But even then, there still might be some issues with dependencies etc.

goranstupar:
Use case is not very specific, but it is not game oriented (when most of/all assets are prepared in advance and are stored in asset folder). 

When assets are not prepared in advance, and are coming from 3th part and may be corrupted (or even do not exist), then it would be nice to know which asset failed to be loaded, with nice callback like this.

scientistnm:
I have the similar use case, and it also isn't game oriented.
It is application oriented, where I have framework above the libgdx.

I let user to load and specify path to it's assets.
It can be stored somewhere on the disk, it's not necessary placed in assets folder.
And assets can be loaded in runtime.

I think that in this case some failed callback would be nice, since now I crash user's application if he tries to load some asset that doesn't exists, and it just a bad practice.

Failed callback would be great, because it would let user to implement it, and use some default image instead requested one (e.g. gallery of images).

goranstupar:
Anyone?

goranstupar:
xoppa, any news on this one?

NathanSweet:
Please try the solution I've committed.

goranstupar:
This is great! Thanks!

Hvala puno

NathanSweet:
Molim!

PR #3167, opened 2015-05-25
Added support for custom decal material in Decal class.

### LINKED ISSUE #2986: Decal setter for DecalMaterial
Hello,

I was trying to change blending of decals in my app.
But I needed to use glBlendFuncSeparate instead of glBlendFunc.

So I needed to extended Decal class and DecalMaterial, and override DecalMaterial set() method.
To change the instance od DecalMaterial in Decal class I needed to to set new instance of my custom DecalMaterial in constructor of custom Decal class. Like:

material = new CustomDecalMaterial();

It's working and it's fine, but I think that would be much cleaner solution if Decal has setter for custom material.

setMaterial(DecalMaterial material){
     this.material = material;
}

This would be much cleaner solution for customizing blending of decals.

scientistnm:
fixes #2986 

Added support for setting custom decal material to Decal class.
Custom material can be applied via constructor or via setter.

Note: This is my first pull request ever and first day of using GIT :)
So any comments are welcome, if I messed up something.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/decals/Decal.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/decals/Decal.java
@@ -55,10 +55,15 @@ public class Decal {
 	public Vector2 transformationOffset = null;
 	protected Vector2 dimensions = new Vector2();
 
-	protected DecalMaterial material = new DecalMaterial();
+	protected DecalMaterial material;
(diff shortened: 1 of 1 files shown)


MobiDevelop:
Looks ok to me.

scientistnm:
If this is ok.
When I can expect to be merged?

PR #3169, opened 2015-05-26
update TiledMapImageLayer.java

Disk1of5:
Hello, this proposed update allows you to move the TiledImageLayer after the map has loaded.


MobiDevelop:
Thanks, added this and some other changes (not sure why we had ints for the x and y).

PR #3175, opened 2015-05-28
fix the OrthograhicCamera unprojection problem

yangweigbh:
the OrthographicCamera use Gdx.graphic width and height to do the unprojection,
but actually should use its own viewportWidth and viewportHeight to do the unprojection

Change-Id: I12e48ee4bb8360d1100760338f273e4d0bb80bc8


badlogic:
I'm not sure what this is supposed to fix? `OrthographicCamera` inherits the unproject methods from `Camera`. They work as intended.

Also, your PR contains unrelated code changes to `FrameBufferObject`.

yangweigbh:
sorry, i bring two commits in one PR, i will seperately PR them, and i will give a more detailed commit message for this commit

PR #3176, opened 2015-05-28
fix nested inner framebuffer end will not restore the outter framebuffer problem

yangweigbh:
fix nested inner framebuffer end will not restore the outter framebuffer problem

when using the framebuffer nested, when inner framebuffer call end, the default framebuffer
will be bind instead of the outer framebuffer. actually the outer framebuffer should be
rebind. also the outer framebuffer's viewport should be restored.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
@@ -21,6 +21,7 @@ import java.nio.ByteOrder;
 import java.nio.IntBuffer;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Stack;
 
 import com.badlogic.gdx.Application;
 import com.badlogic.gdx.Application.ApplicationType;
@@ -86,6 +87,24 @@ public class FrameBuffer implements Disposable {
 	/** format **/
 	protected final Pixmap.Format format;
 
+	static private Stack<Integer> frameBufferHandlerStack = new Stack<Integer>();
+
+	static private Stack<ViewPort> viewPortStack = new Stack<ViewPort>();
+
+	class ViewPort {
+		int x;
+		int y;
+		int width;
+		int height;
+
+		public ViewPort (int x, int y, int width, int height) {
+			super();
+			this.x = x;
+			this.y = y;
+			this.width = width;
+			this.height = height;
+		}
+	}
 	
 	/** Creates a new FrameBuffer having the given dimensions and potentially a depth buffer attached.
 	 * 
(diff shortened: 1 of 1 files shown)


Darkyenus:
I am not sure you are supposed to nest framebuffers, what is the use of it anyway? They don't inherit anything, so I don't see how would they benefit from nesting.

yangweigbh:
say if i want to render the scene to offscreen, i have to begin a framebuffer(outer buffer), then in the rendering, i want to draw use the scene2d widget in the 3d space, so i will begin a new framebuffer(inner buffer), and draw the scene2d widget in the frame buffer and attach the color texture of the framebuffer on a decal. So there form a nested framebuffer. i know you can end the previous one before the the begin of the next one, but it's not convenient. you can also refer to the question on stackoverflowhttp://stackoverflow.com/questions/25471727/libgdx-nested-framebuffer

发自我的 iPhone

MobiDevelop:
From your SO post, it looks like you have an application-level solution to this already... That is probably where this sort of stuff belongs.

yangweigbh:
i think application solution is not convenient and other developer may encounter the same problem and cause them to debug and find the solution again, better fix it from the framework side

发自我的 iPhone

badlogic:
I'm not in favor of merging this. Especially not by using statics And
Stack<Integer>.

Nesting buffers, texture binds, mesh binds is making state management
terrible. This should be solved on the app level.

xoppa:
Thanks for the contribution! I can see how this can be useful. However I'm afraid that we won't merge this. This should be done by the user. Let me explain:

Switching render target forces the GPU to flush which is a very costly operation. Even more, switching render target without clearing it, requires the GPU to copy (reconstruct) the original buffer. "Nesting" FBO's like this looks like a really a bad design (btw, I don't think that it's actually nesting, because they are not related to each other). I really don't think that we should encourage such bad practice.

Also, this change forces the user to use the stack, it is not optional. It is perfectly valid to bind one fbo and then bind another and then bind the default, there is no need to "unbind" (`end()`) a previously bound fbo.

From a design point of view it isn't the responsibility of Framebuffer to keep track of which are/were bound (the statics you have are a good indication for that). That would be better suited for a helper class (like in your SO post).

Libgdx doesnt keep track of the context because of the overhead that comes with. Users are allowed to create fbo's without using the Framebuffer class (think cubemap, mrt, etc.). This are context switches the Framebauffer cannot and should not be aware of. If you would really like to keep track of those in a generic way then it should be done by encapsulating the gl calls.

yangweigbh:
thanks for the reply

发自我的 iPhone

PR #3177, opened 2015-05-29
Add exponent to SpotLight.

realitix:
Exponent is the rapidity of the falloff of light from its target direction.


xoppa:
Thanks! Would it be possible for you to provide a small test to show the usage? See https://github.com/libgdx/libgdx/tree/master/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d, testshaders can be placed e.g. here: https://github.com/libgdx/libgdx/tree/master/tests/gdx-tests-android/assets/data/g3d/shaders.

realitix:
I'm impementing the shadow system we spoke, so when it will be ready, I will show you. Moreover, because I use a phong shader, we will see the exponent in action.

PR #3183, opened 2015-05-31
Label layout color glitch fix

josephknight:
If a tinted label (having its color set via setColor instead of style font) was not wrapped or had ellipses set then the label’s layout’s font color would not be set to white but would instead be set to it’s last used color. Therefore when draw() tinted the layout you ended up with compounded and incorrect tinting.


josephknight:
Sorry, git mistake. Will repost without other issue's commits included.

PR #3185, opened 2015-06-01
Fixes incorrect label colors under certain circumstances.

josephknight:
Fixes incorrect tint color occurring if label has a non-white setColor tint applied and label’s computePrefSize method called while label is in non-wrapping state or has ellipses set.

Problem with improper label colors occurring on non-white labels in a scrollpane when scrollbar appears has been traced to here. We shouldn’t ignore resetting color to white upon compute size, otherwise the cache font’s current color is used which, if not white, will not provide a proper base for tinting when drawn.


NathanSweet:
How can I reproduce the problem that this PR fixes?

josephknight:
Thank you for your interest. I am attempting to re-replicate the issue (my project has left that state behind) to determine if this is still an issue since recent changes in label and bitfonts may have solved it elsewhere. I'll respond here as soon as I have info.

josephknight:
This appears to have been fixed by you when you changed computePrefSize() to use Label.prefSizeLayout as its working reference instead of the main layout field. Nice move, and thanks.

PR #3209, opened 2015-06-14
Add screen transitions

iXeption:
Adds frame buffer based screen transitions like, alpha-fading, color-fading, sliding and slicing.
The Game class has been extended to FadingGame, which applies transitions, when setScreen is called. Allows callbacks on start and end of the transitions.

ScreenTransition Test included:
left Mouse - toggle Screen
right Mouse - change transition

Be free to contact me for wiki tutorial or something if you want to accept the pull request.

### DIFF
--- a/gdx/src/com/badlogic/gdx/FadingGame.java
+++ b/gdx/src/com/badlogic/gdx/FadingGame.java
@@ -16,12 +16,10 @@
 
 package com.badlogic.gdx;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.g2d.Batch;
 import com.badlogic.gdx.graphics.glutils.FrameBuffer;
+import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.transition.ScreenTransition;
 import com.badlogic.gdx.utils.transition.TransitionListener;
 
@@ -34,10 +32,10 @@ public class FadingGame extends Game {
 
 	protected FrameBuffer currentScreenFBO;
 	protected FrameBuffer nextScreenFBO;
+	final protected Batch batch;
 	protected Screen nextScreen;
-	protected Batch batch;
 
-	private List<TransitionListener> listerns;
+	private final Array<TransitionListener> listeners;
 
 	private float transitionDuration;
 	private float currentTransitionTime;
@@ -46,19 +44,22 @@ public class FadingGame extends Game {
 
 	public FadingGame (Batch batch) {
 		this.batch = batch;
+		this.listeners = new Array<TransitionListener>();
 	}
 
 	@Override
 	public void create () {
-		currentScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
-		nextScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
-		listerns = new ArrayList<TransitionListener>();
+		this.currentScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
+		this.nextScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
 	}
 
 	@Override
 	public void dispose () {
 		if (screen != null) screen.hide();
 		if (nextScreen != null) nextScreen.hide();
+
+		this.currentScreenFBO.dispose();
+		this.nextScreenFBO.dispose();
 
 	}
 
@@ -118,7 +119,7 @@ public class FadingGame extends Game {
 	}
 
 	@Override
-	public void resize (int width, int height) {				
+	public void resize (int width, int height) {
 		if (screen != null) screen.resize(width, height);
 		if (nextScreen != null) nextScreen.resize(width, height);
 	}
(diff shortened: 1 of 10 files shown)


gdx/src/com/badlogic/gdx/FadingGame.java
```diff
@@ -0,0 +1,210 @@
...
+		}
+
+	}
+
+	@Override
+	public void resize (int width, int height) {
+		boolean resize = currentScreenFBO.getHeight() != height || currentScreenFBO.getWidth() != width
+			|| nextScreenFBO.getHeight() != height || nextScreenFBO.getWidth() != width;
+
+		if (false) {
```
MobiDevelop: Probably not an intentional `if (false)`, I assume.

iXeption: You a re right, sorry forgot to delete it.

gdx/src/com/badlogic/gdx/FadingGame.java
```diff
@@ -0,0 +1,199 @@
...
+ * screen is set.
+ * @author iXeption */
+public class FadingGame extends Game {
+
+	protected FrameBuffer currentScreenFBO;
+	protected FrameBuffer nextScreenFBO;
+	protected Screen nextScreen;
+	protected Batch batch;
+
+	private List<TransitionListener> listerns;
```
Darkyenus: Typo. Also this could be final and initialized right here or in the constructor, I see no reason for it to be initialized in create(). And a minor detail, something like clearTransitionListeners() could be useful.

gdx/src/com/badlogic/gdx/FadingGame.java
```diff
@@ -0,0 +1,199 @@
...
+			}
+
+		}
+
+	}
+
+	@Override
+	public void resize (int width, int height) {				
+		if (screen != null) screen.resize(width, height);
+		if (nextScreen != null) nextScreen.resize(width, height);
```
Darkyenus: Unless I am missing something, currentScreenFBO and nextScreenFBO should be resized too. I suggest making that a separate protected method to allow customized resizing, since creating them can already be customized and FBOs are heavy objects.

gdx/src/com/badlogic/gdx/FadingGame.java
```diff
@@ -0,0 +1,199 @@
...
+		currentScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
+		nextScreenFBO = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
+		listerns = new ArrayList<TransitionListener>();
+	}
+
+	@Override
+	public void dispose () {
+		if (screen != null) screen.hide();
+		if (nextScreen != null) nextScreen.hide();
+
```
Darkyenus: Shouldn't the FBOs be disposed here as well, when they are created here?

gdx/src/com/badlogic/gdx/utils/transition/ColorFadeTransition.java
```diff
@@ -0,0 +1,97 @@
...
+			Color fadeOut = new Color(Color.WHITE);
+			fadeOut.a = 1.0f - fade;
+			batch.setColor(fadeOut);
+
+			batch.draw(currentScreenTexture, 0, 0, width / 2, height / 2, width, height, 1, 1, 0, 0, 0, (int)width, (int)height,
+				false, true);
+
+		}
+
+		Color c = new Color(Color.WHITE);
```
Tom-Ski: Should use a temp variable rather than new instances. Same goes for the rest of this PR.

gdx/src/com/badlogic/gdx/FadingGame.java
Tom-Ski: Should also be using LibGDX collections

gdx/src/com/badlogic/gdx/FadingGame.java
iXeption: Yeah had that before, but it doesn't work properly.

gdx/src/com/badlogic/gdx/FadingGame.java
iXeption: Ill fix it, thanks

gdx/src/com/badlogic/gdx/FadingGame.java
iXeption: You are right.

gdx/src/com/badlogic/gdx/FadingGame.java
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+	}
+
+	@Override
+	public void resize (int width, int height) {
+		if (screen != null) screen.resize(width, height);
+		if (nextScreen != null) nextScreen.resize(width, height);
+	}
```
nooone: I think you will have to create new FBOs if the screen size changes.

nooone:
I'd rename `FadingGame` to something more general (`ScreenTransitionGame`?), because the name is too specific, imho.

Furthermore, I'm not sure how well this works with `Viewport`s, namely `glViewport()`. If we assume a `FitViewport` with black bars, your FBOs will have the size of the whole screen, including the black bars. That's okay, but after you convert it to a texture and draw it via the given batch, you will draw the whole screen texture (including the black bars), in the area of the glViewport, which means *in between* the black bars, actually doubling them.

I'd suggest to add a viewport to your test and see how it works, also while resizing during a transition (see my previous code-comment).

iXeption:
I will figure it out, if i have the time.

gdx/src/com/badlogic/gdx/utils/transition/ColorFadeTransition.java
```diff
@@ -0,0 +1,92 @@
...
+	@Override
+	public void render (Batch batch, Texture currentScreenTexture, Texture nextScreenTexture, float percent) {
+		float width = currentScreenTexture.getWidth();
+		float height = currentScreenTexture.getHeight();
+		float x = 0;
+		float y = 0;
+
+		if (interpolation != null) percent = interpolation.apply(percent);
+
+		Gdx.gl.glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
```
saltares: There's a bunch of logic in these transitions that gets duplicated every time. Can you please put them in a base class as protected methods and have the transitions extend it?

iXeption: Yes i can do that when i take a look at the viewport stuff.

MobiDevelop:
I'm not sure this is something that belongs in libgdx core. It would probably be better published as a separate library.

iXeption:
Sorry i am really busy at the moment, so i could not work on the code.
For my unterstanding screen transitions are a core thing for an engine, but because i can not guarantee to fix this in the next weeks, i will publish it as a lib.

saltares:
I actually agree with @MobiDevelop. Will close this PR now, this should be an external library.

PR #3215, opened 2015-06-19
sdkLocationText uses correct directory structure based on OS name.

jallen720:
Made sample location for Android SDK consistent with OS directory structure.


badlogic:
Thanks!

PR #3220, opened 2015-06-20
Added methods to create random vector

felipefcm:


Gtoknu:
This is actually very wrong.
This way, the points picked will tend to be near the center, instead of distributed. Take a look here on why, and how to fix it: http://mathworld.wolfram.com/DiskPointPicking.html

Gtoknu:
Don't take me wrong, the idea is not bad at all, but there is some stuff to be sorted out. I bet it won't take 10 minutes to fix it and send it over again.

florianbaethge:
I think we could even advance this to a whole Sampler class... With methods for uniformly sampling normalized vectors (or points) ON a unit sphere surface, uniformly distributed inside a sphere, 2D inside circle etc... there are some more... within a certain segment of a sphere etc etc etc... I use this a lot and for many games it could come in handy...

felipefcm:
Thanks for the comments, sorry if I seemed rude or something like it.
I wrote that code in my project and pasted it in the libgdx fork without even thinking much about it or testing it properly. I was so anxious to contribute that I made a terribly bad move. Sorry about that!

felipefcm:
I irrecoverably broke this pull request, still learning Git here, so I created another:
https://github.com/libgdx/libgdx/pull/3222

PR #3222, opened 2015-06-23
Adds methods to create random vector

felipefcm:
Also added a method to set the components of the Vector3 based on a spherical coordinate.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector.java
@@ -120,6 +120,11 @@ public interface Vector<T extends Vector<T>> {
 	 * @return This vector for chaining. */
 	T interpolate (T target, float alpha, Interpolation interpolator);
 
+	/**
+	 * Sets this vector to a random direction
+	 * @return This vector for chaining */
+	T setToRandomDirection ();
+
 	/** @return Whether this vector is a unit length vector */
 	boolean isUnit ();
 
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -409,6 +409,15 @@ public class Vector2 implements Serializable, Vector<Vector2> {
 		return lerp(target, interpolation.apply(alpha));
 	}
 
+	/**
+	 * Sets this vector to a random direction
+	 * @return This vector for chaining */
+	@Override
+	public Vector2 setToRandomDirection () {
+		float theta = MathUtils.random(0f, MathUtils.PI2);
+		return this.set(MathUtils.cos(theta), MathUtils.sin(theta));
+	}
+
 	@Override
 	public int hashCode () {
 		final int prime = 31;
--- a/gdx/src/com/badlogic/gdx/math/Vector3.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector3.java
@@ -108,6 +108,35 @@ public class Vector3 implements Serializable, Vector<Vector3> {
 		return this.set(vector.x, vector.y, z);
 	}
 
+	/**
+	 * Sets the components from the given spherical coordinate
+	 * @param azimuthalAngle The angle between x-axis in radians [0, 2pi]
+	 * @param polarAngle The angle between z-axis in radians [0, pi]
+	 * @return This vector for chaining */
+	public Vector3 setFromSpherical (float azimuthalAngle, float polarAngle) {
+		float cosPolar = MathUtils.cos(polarAngle);
+		float sinPolar = MathUtils.sin(polarAngle);
+
+		float cosAzim = MathUtils.cos(azimuthalAngle);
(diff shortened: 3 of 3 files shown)


nooone:
`Vector3.setFromSpherical(...)` might be useful in some cases, but I think the random methods are too specific to be added to libgdx.

Darkyenus:
And if they make it in (random methods), they should IMO take the resulting vector as a parameter, instead of allocating a new one.

felipefcm:
Thanks for the feedback. I created the random vector methods because they are largely used in game projects and they seemed generic enough, IMO.

gdx/src/com/badlogic/gdx/math/MathUtils.java
```diff
@@ -182,6 +182,24 @@ static public int randomSign () {
 		return 1 | (random.nextInt() >> 31);
 	}
 
+	/** Returns a random point inside the unit circle */
+	static public Vector2 randomUnitCirclePoint (Vector2 vector) {
```
Gtoknu: This is actually a random point in a disk, not a random point in a circle.

https://en.wikipedia.org/wiki/Disk_(mathematics) Disk is the area bounded by a circle. (circle is only the boundaries, disk is everything inside the circle)

gdx/src/com/badlogic/gdx/math/MathUtils.java
```diff
@@ -182,6 +182,24 @@ static public int randomSign () {
...
 
+	/** Returns a random point inside the unit circle */
+	static public Vector2 randomUnitCirclePoint (Vector2 vector) {
+		float theta = random(0f, PI2);
+		float sqrt = (float) Math.sqrt(random());
+		return vector.set(sqrt * cos(theta), sqrt * sin(theta));
+	}
+
+	/** Returns a random point on the unit sphere's surface */
+	static public Vector3 randomUnitSpherePoint (Vector3 vector) {
```
Gtoknu: Same way, this is ball picking, not sphere picking.

gdx/src/com/badlogic/gdx/math/MathUtils.java
felipefcm: Thanks!

gdx/src/com/badlogic/gdx/math/MathUtils.java
felipefcm: All the points belong to the same sphere.

Gtoknu: The sphere bounds the ball. True, all points it picks are INSIDE the sphere, but they lay on the ball, not the sphere.

felipefcm: I'm not sure if I understand what you mean. The method should output uniformly distributed points on a unit sphere surface. All the points have spherical radius 1, so how they could be inside the sphere (ball), and not on the surface?

Gtoknu: Re-read the code, and you're right, this one is picking on the sphere, not the ball. Pardon me!

gdx/src/com/badlogic/gdx/math/MathUtils.java
GreenLightning: Wouldn't it make sense to change this method so that it actually returns only normalized vectors (length = 1). I think this method is mostly going to be used to create a random direction and it would make things easier if the vectors had all the same length. Otherwise at least rename the method so that it is clear that it does return points in the unit circle and not only those on the unit circle.

GreenLightning:
@Darkyenus Yeah, that is fine, but I would add some parameterless versions, which create a new vector and pass it to the other methods in case a new vector is needed.

xoppa:
Thanks! Looks good! I guess that this would be used to set a vector to a random direction. It seems to be a bit inconsistent between Vector2 and Vector3. Could you please modify it a bit:
- Move the methods to the Vector2 and Vector3 classes and rename the method (something like `setToRandomDirection` or so, but I suck at naming so that's probably not the best name), make sure to add it to the Vector interface as well
- Remove the scaling, like the `sqrt` variable and `radius` argument.
- Add some javadocs that it will result in a random unit vector.
- Squash commits

xoppa:
@felipefcm what's the status on this?

felipefcm:
@xoppa Thanks for the improvements you suggested, I will work on them ASAP.

felipefcm:
Ok, I hope I didn't messed too much with the branch.

felipefcm:
@Darkyenus You're right!

xoppa:
Could you please use the eclipse formatter, remove the duplicate methods in MathUtils and squash your commits?

felipefcm:
Even when using Eclipse Formatter settings, for some reason 'toString()' and 'fromString()' methods line endings chars seems to get modified everytime. Using IntelliJ here. Do I have to use specific EOL symbols?

xoppa:
thanks!

PR #3224, opened 2015-06-23
Add Spotlight in DefaultShader.

realitix:
The glsl files are not updated so SpotLight won't work as is with the
DefaultShader. It can be useful to bring his own shader with phong
lightning to see the beauty of spot lights.

I tested it with the light tests and it creates no error.


xoppa:
Thanks! Looking forward to see a simple test.

PR #3231, opened 2015-06-25
Add builder to lights.

realitix:
Allow to create lights more easily.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/environment/BaseLight.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/environment/BaseLight.java
@@ -21,23 +21,13 @@ import com.badlogic.gdx.graphics.Color;
 public abstract class BaseLight {
 	public final Color color = new Color(0, 0, 0, 1);
 
-	public static class Builder {
-		float r = 1, g = 1, b = 1, a = 1;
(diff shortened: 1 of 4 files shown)


xoppa:
Ow, that's not what I expected. What's the use-case? Why would you create another class just to pass some values? In that case you might as well use the copy constructor.

xoppa:
To expand a bit; I thought that you wanted to add the builder methods to the lights them self. E.g. so you could do: `new SpotLight().setColor(Color.BLUE).setPosition(x, y, z).setDirection(Vector3.Z);` etc.

warmwaffles:
:+1: @xoppa that would be nice

realitix:
You are right ! I do it tomorrow

realitix:
I was following strictly the builder pattern but in that case it's not necessary because none of the arguments are mandatory.

xoppa:
Looks like the `setColor` method of PointLight and SpotLight isnt needed. Instead change the class to `BaseLight<T extends BaseLight<T>>`, `PointLight extends BaseLight<PointLight>`, etc. And make the `setColor` method return `T`

realitix:
I followed your advice. You can take a look.

xoppa:
Thanks!

PR #3234, opened 2015-06-26
This allows user that can change wrap logic in custom glyph layout.

scientistnm:
This allows user that can change wrap logic in custom glyph layout.
(e.g. remove multiplewhite spaces omitting when glyph run is wrapped if needed)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -267,7 +267,7 @@ public class GlyphLayout implements Poolable {
 		glyphRunPool.free(truncateRun);
 	}
 
-	private GlyphRun wrap (BitmapFontData fontData, GlyphRun first, Pool<GlyphRun> glyphRunPool, int wrapIndex, int widthIndex) {
+	protected GlyphRun wrap (BitmapFontData fontData, GlyphRun first, Pool<GlyphRun> glyphRunPool, int wrapIndex, int widthIndex) {
 		GlyphRun second = glyphRunPool.obtain();
 		second.color.set(first.color);
 		int glyphCount = first.glyphs.size;
@@ -311,7 +311,7 @@ public class GlyphLayout implements Poolable {
 	}
 
 	/** Adjusts the xadvance of the last glyph to use its width instead of xadvance. */
-	private void adjustLastGlyph (BitmapFontData fontData, GlyphRun run) {
+	protected void adjustLastGlyph (BitmapFontData fontData, GlyphRun run) {
 		Glyph last = run.glyphs.peek();
 		if (fontData.isWhitespace((char)last.id)) return; // Can happen when doing truncate.
 		float width = (last.xoffset + last.width) * fontData.scaleX - fontData.padRight;


davebaol:
Sounds reasonable

scientistnm:
If this is ok.
Will it be merged?

davebaol:
I think so, but better wait for @NathanSweet or some other team member :)

NathanSweet:
These methods contain quite complex logic that is internal to GlyphLayout. What exactly are you trying to customize? The wrap position is customized via BitmapFontData `getWrapIndex`.

scientistnm:
I'm having trouble with multiple whitespace omitting.
I would like to disable that feature, and to let the number of glyphs in glyph layout, and number of chars in rendered text be in sync.

Maybe there are some other possibilities for customization, but currently this is what I'm looking for.

What do you think?

scientistnm:
Those methods aren't using any private fileds, so I guess it would be ok, if user wants to override their behaivour.

NathanSweet:
I don't understand the problem you are having or what change you would make if the methods were public. Please show how you will change the methods.

scientistnm:
In my usecase I would remove those lines from these two methods:

for (; endIndex > 0; endIndex--)
if (!fontData.isWhitespace((char)first.glyphs.get(endIndex - 1).id)) break;

Here I will try to explain the problems I have again:

1. I try to create custom 3D input field widget using decals and drawing to FBO.
I have a string field which I update in keyTyped method, simply appending new chars to it.
Input field has some width and height. Everything renders well if I type non-white chars.
But when I type multiple white spaces together, and text gets wrapped with the new line starting with the white spaces, here comes the problem.

|012345          |
|'X'                  |

non-white char will start at the left side of widget, no matter how much whitespaces I typed before,
Because glyph layout has omitted them earlier.

I draw text like this:

glyphLayout.setText (bitmapFont, textInWidget, someColor, widgetWidth, widgetAlign, true).

I have seen your TextArea class, and there you can type as many whitespaces as you want, and there is no this kind of problem, but I can't figure out why and how it's working there.
There are complex calculus that are not documented and I'm not sure if I need all of that in my usecase anyway.

2. I try to calculate correct position of cursor in my custom 3D edit field that uses decal.
I update index cursor on keyDown event (left,right), cursor index is in sync with the char in current string. So when I try to calculate position of glyph (using xAdvances and y position of runs) in edit field based on "selected char" it gives me incorrect values, because sometimes index of char in string and the index of correct glyph in layout are not the same when whitespaces get omitted.

So this is my story here.

If you have problem with this pull request, please be kind and help me out to solve this problem on my own. Try explaining a little logic behind that code in wrap procedure, and cursor positioning in TextArea and TextField libgdx classes.

Thanks.

NathanSweet:
I didn't write TextArea, but it seems to render each line separately rather than rely on GlyphLayout's wrapping (also TextArea was written before GlyphLayout). That means no whitespace is removed when TextArea draws its text. TextArea works by storing the glyph positions of the text without wrapping (see TextField:377). It calculates the text on each line (see TextArea:277). To find the glyph under the cursor, it uses the glyph positions, subtracting the position of the first character for the line the cursor is on (see TextArea `letterUnderCursor`).

Removing the two lines you specified, plus adding spaces added to BitmapFontTest:


Same with smaller width:


Whitespace is still gone, so it doesn't appear to fix your problem. After fixing a bug with BitmapFont's padding and calculation of the space glyph when missing from the font, removing the first 2 for loops from `wrap` sort of does what you want:



Here only the whitespace that didn't fit is wrapped. That makes for odd behavior:



The space after "test" is wrapped to the next line, not sure why. It may be fixable, but this still isn't really the write kind of wrapping for a text area. "one two" should wrap to "one\ntwo", the newline represents the space. For "one two    " when only the last space doesn't fit it should wrap to "one\ntwo    ". Try it out in Notepad++.

I've changed wrapping to work more like Notepad++, where if whitespace needs to wrap, it also takes the previous word with it. This means we no longer need to strip whitespace.

scientistnm:
Hi,

It works better, but also I have a problem determing the current layout width.
I droped TextArea usecase for a moment, and focusing on TextField usecase.

In this approach I always have "singlelined" text, becase I render to texture, I need to use some stencil masking to simulate TextField scrolling when text lenght exceeds the widget width.

When you have a glyphLayout that has the whitespace on the end of the text, even when I don't use wrapping, it's not calculated in global layout width.

I use this method:
glyphLayout.setText(BitmapFont font, singlelinedText);

This way I'm unable to detect that I should scroll my text, because bounds of the glyphlayout doesn't change when I just type whitespaces constantly after some already rendered text.

NathanSweet:
I don't see that being the case. I edited BitmapFontTest line 113 to "test      " and changed to setText(font, str). Works fine:



Instead of clipping your text, you could simply just draw the characters fully within the bounds. It isn't quite as good, but this is what TextField does.

scientistnm:
I can comfirm that this now works much better, when I updated to nighly.
Whitespaces are included in layout bounds also.

I have a progress in implementing TextField...I will drop usecase TextArea for now.

Cheers!

PR #3236, opened 2015-06-26
Implement equals/hashCode for collection classes

### LINKED ISSUE #3204: Implement equals() and hashCode() for collection classes
The collection classes in libgdx should implement `equals()` and `hashCode()`. This includes the various `Entry` classes. Here is a list of classes missing `equals()` or `hashCode()` implementations:
- `com.badlogic.gdx.utils.Array` - only `equals()` is implemented
- `com.badlogic.gdx.utils.ArrayMap` - uses `ObjectMap$Entry`
- `com.badlogic.gdx.utils.BinaryHeap`
- `com.badlogic.gdx.utils.BinaryHeap$Node`
- `com.badlogic.gdx.utils.BooleanArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.ByteArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.CharArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.FloatArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.IntArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.IdentityMap`
- `com.badlogic.gdx.utils.IdentityMap$Entry` - could be replaced with `ObjectMap$Entry`
- `com.badlogic.gdx.utils.IntFloatMap`
- `com.badlogic.gdx.utils.IntFloatMap$Entry`
- `com.badlogic.gdx.utils.IntIntMap`
- `com.badlogic.gdx.utils.IntIntMap$Entry`
- `com.badlogic.gdx.utils.IntMap`
- `com.badlogic.gdx.utils.IntMap$Entry`
- `com.badlogic.gdx.utils.IntSet`
- `com.badlogic.gdx.utils.LongArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.LongMap`
- `com.badlogic.gdx.utils.LongMap$Entry`
- `com.badlogic.gdx.utils.ObjectFloatMap`
- `com.badlogic.gdx.utils.ObjectFloatMap$Entry`
- `com.badlogic.gdx.utils.ObjectIntMap`
- `com.badlogic.gdx.utils.ObjectIntMap$Entry`
- `com.badlogic.gdx.utils.ObjectMap`
- `com.badlogic.gdx.utils.ObjectMap$Entry`
- `com.badlogic.gdx.utils.ObjectSet`
- `com.badlogic.gdx.utils.OrderedMap` - extends `ObjectMap`, but semantics may be different
- `com.badlogic.gdx.utils.OrderedSet` - extends `ObjectSet`, but semantics may be different
- `com.badlogic.gdx.utils.PooledLinkedList`
- `com.badlogic.gdx.utils.ShortArray` - only `equals()` is implemented
- `com.badlogic.gdx.utils.SortedIntList`

Seriously, shouldn't all these primitive type variations be auto-generated?

Antag99:
This pull request implements `equals()` and `hashCode()` for all array and map classes in libgdx. It omits the `com.badlogic.gdx.utils.PooledLinkedList` and `com.badlogic.gdx.utils.SortedIntList` classes. It does, however, not omit `BinaryHeap`, which is patched in commit 83d6777fc71311e5b437d6234198b55d3982920d along with the map classes. Fixes #3204.


gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -459,6 +470,7 @@ public boolean equals (Object object) {
 		return true;
 	}
 
+	@Override
```
nooone: Oh noes, automatic `@Override` :hand:

Antag99: Meh. It was intentional.

nooone:
May I ask how those methods are implemented/generated? They all look pretty different and it's probably going to be difficult to review this PR.

Maybe a little test might help to verify that the methods work and follow the contract.

Antag99:
Mostly copy-pasted, but there are lots of minor differences that need to be handled. Some maps need to track the key that's `0`, and others need to handle if an object is null before calling `hashCode()`. I omitted the lists because they looked pretty legacy, and the linked list was kind of a list and iterator combined, which I found it hard to define equality for.

Antag99:
But yes, it might be difficult to review a changeset of 20 copy-pasted files. We are, however, kind of stuck in the mud right now.

gdx/src/com/badlogic/gdx/utils/IdentityMap.java
```diff
@@ -500,6 +500,38 @@ private int hash3 (int h) {
...
+	@Override
+	public boolean equals (Object obj) {
+		if (obj == this) return true;
+		if (!(obj instanceof IdentityMap)) return false;
+		IdentityMap<K, V> other = (IdentityMap) obj;
+		if (other.size != size) return false;
+		for (Entry<K, V> entry : entries()) {
+			V otherValue = other.get(entry.key);
+			if (entry.value == null) {
+				if (!other.containsKey(entry.key) || other.get(entry.key) != null) {
```
NathanSweet: This is doing a total of 3 map lookups in the case where entry.value is null. It could be done in one by passing a dummy object to `get (K key, V defaultValue)`, otherwise we should at least do it in two by omitting the third `get`.

gdx/src/com/badlogic/gdx/utils/Array.java
NathanSweet: You may have noticed from the source I'm a bit anal about the classes I maintain. I'd rather remove the `@Override`. Their use in library code is questionable, but their use on JDK classes that can never change is unnecessary.

gdx/src/com/badlogic/gdx/utils/IdentityMap.java
```diff
@@ -500,6 +500,38 @@ private int hash3 (int h) {
...
+					return false;
+				}
+			} else {
+				if (!entry.value.equals(other.get(entry.key))) {
```
NathanSweet: `otherValue` should also be used here.

gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
```diff
@@ -539,6 +539,31 @@ private int hash3 (int h) {
 		return (h ^ h >>> hashShift) & mask;
 	}
 
+	@Override
+	public int hashCode () {
+		int h = 0;
+		for (Entry entry : entries()) {
+			h += Float.floatToIntBits(entry.key) * 31;
```
NathanSweet: The key is an int.

gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -443,6 +443,17 @@ public T random () {
 		return result;
 	}
 
+	@Override
+	public int hashCode () {
+		int h = 1;
+		for (int i = 0, n = size; i < n; i++) {
+			h *= 31;
+			h += items[i] != null ? items[i].hashCode() : 0;
```
NathanSweet: Could be done in one array lookup. Maybe it would be optimized away, maybe not. We can get a bit more optimizy than usual in collection code.

gdx/src/com/badlogic/gdx/utils/ArrayMap.java
```diff
@@ -398,6 +398,38 @@ public void truncate (int newSize) {
 		size = newSize;
 	}
 
+	@Override
+	public int hashCode () {
+		int h = 0;
+		for (Entry<K, V> entry : entries()) {
+			h += (entry.key != null ? entry.key.hashCode() : 0) * 31;
+			h += (entry.value != null ? entry.value.hashCode() : 0);
```
NathanSweet: Should be multiplied by a prime. Both prime multiplications can be avoided when null.

gdx/src/com/badlogic/gdx/utils/ArrayMap.java
```diff
@@ -398,6 +398,38 @@ public void truncate (int newSize) {
...
+		ArrayMap<K, V> other = (ArrayMap) obj;
+		if (other.size != size) return false;
+		for (Entry<K, V> entry : entries()) {
+			Object otherValue = other.get(entry.key);
+			if (entry.value == null) {
+				if (!other.containsKey(entry.key) || other.get(entry.key) != null) {
+					return false;
+				}
+			} else {
+				if (!entry.value.equals(other.get(entry.key))) {
```
NathanSweet: Excessive gets. I assume all maps have this so I won't comment about it anymore.

gdx/src/com/badlogic/gdx/utils/IntIntMap.java
```diff
@@ -533,6 +533,30 @@ private int hash3 (int h) {
 		return (h ^ h >>> hashShift) & mask;
 	}
 
+	@Override
+	public int hashCode () {
+		int h = 0;
+		for (Entry entry : entries()) {
+			h += entry.key * 31 + entry.value;
```
NathanSweet: I think the value needs prime multiplication.

NathanSweet:
Looks pretty good, would like just some minor changes before merging, commented above. Did you test the methods? Any chance for a CollectionsTest, similar to JsonTest?

gdx/src/com/badlogic/gdx/utils/IntIntMap.java
Antag99: > I think the value needs prime multiplication.

Doesn't that result in a worse hash function? This will result in the same hash code for a reversed map.

gdx/src/com/badlogic/gdx/utils/IdentityMap.java
Antag99: I got rid of `otherValue`, since it was unused. It seems like the "unused" warnings have been suppressed. The case where `entry.value == null` is pretty rare, so I guess there is no point in optimizing it further. If we want this to run as fast as possible, then the backing arrays should probably be used instead of `Entries`.

Antag99:
I added `CollectionsTest`, that tests for correctness of `equals()` and `hashCode()`. It uses reflection to avoid massive code duplication.

gdx/src/com/badlogic/gdx/utils/IntIntMap.java
NathanSweet: Could use a different prime for the value, eg 37.

gdx/src/com/badlogic/gdx/utils/IdentityMap.java
NathanSweet: Probably just as easy to use `otherValue` instead of doing extra gets. Good point about using the backing arrays, that would be better.

xoppa:
What is the reason to use the same prime for each class (so that the hashcode of e.g. CharArray and ShortArray are the same if they contain the same values)? Also, the hashcode of every empty map/set (regardless the type) is 0 and of every empty array (regardsless the type) is 1. Is that intentional and why?

gdx/src/com/badlogic/gdx/utils/ArrayMap.java
```diff
@@ -398,6 +398,42 @@ public void truncate (int newSize) {
...
 
+	public int hashCode () {
+		K[] keys = this.keys;
+		V[] values = this.values;
+		int h = 0;
+		for (int i = 0, n = size; i < n; i++) {
+			K key = keys[i];
+			V value = values[i];
+			if (key != null) h += key.hashCode() * 31;
+			if (value != null) h += value.hashCode();
```
xoppa: Why is the multiplication with the prime at the key and not at `h`?

Antag99:
`31` is the convention for hash functions. The implementations are different for arrays and maps because map entries may appear in arbitrary order.

xoppa:
Could you provide a reference to that? Why is whether they may appear in arbitrary order relevant for empty collections?

Antag99:
See [this question] (http://stackoverflow.com/questions/299304/why-does-javas-hashcode-in-string-use-31-as-a-multiplier) on StackOverflow. I used `1` as a starting number for array hashcodes since that's what's being used in the other implementations (such as in the JDK). If it's really necessary, I can make it start at `0`.

xoppa:
Thanks for the reference, so the reason you've used 31 for the prime of every class is that it performs good on some hardware. That doesn't explain though why you've chosen the *same* prime for every class (and thus the hashcode of e.g. CharArray and ShortArray are the same if they contain the same value).

I understand that using 0 for empty sets/maps and 1 for empty arrays is convenient for your implementation.

To expand a bit: iirc the contract is that when `equals()` returns `true` then the `hashCode()` should be the same, but when the `hashCode()` is the same there's no guarantee that `equals()` returns `true`. However, algorithms might still use the `hashCode()` for optimizations and only use the `equals()` for verification. So two different things returning the same hashcode while not being equal might meet the contract. But if we can simply "fix" that e.g. by using a different prime, then I think we should at least consider that.

Antag99:
Well, `31` is used in literally *every* `hashCode()` implementation I have seen. (I don't mind changing it, however.) Hashcode collisions between different types shouldn't matter at all, as they shouldn't be used in the same map as keys. And yes, you're right about the contract of `equals()` and `hashCode()`.

One thing that I realized is that arrays may be *unordered*. In that case, `equals()` and `hashCode()` should not depend on the ordering of elements, and an ordered array should never be equal to an unordered one. I'll look into implementing this.

Antag99:
I've come up with two solutions to this comparing unordered arrays; the first uses a bitmask to speed up iteration, and the other simply scans for earlier duplicates and ensure they match.
##### Using bitmask:
```java
Bits bits = Pools.obtain(Bits.class);
scan: for (int i = 0; i < n; i++) {
	Object item = items1[i];
	for (int ii = bits.nextClearBit(0); ii != -1; ii = bits.nextClearBit(ii + 1)) {
		if ((item == null ? items2[ii] == null : item.equals(items2[ii]))) {
			bits.set(ii);
			continue scan;
		}
	}
	return false;
}
bits.clear();
Pools.free(bits);
```
##### Using nested iteration:
```java
for (int i = 0; i < n; i++) {
	Object item = items1[i];

	int cursor1 = 0;
	int cursor2 = 0;

	// Find earlier duplicates (note that this includes item)
	scan: for (; cursor1 <= i; cursor1++) {
		if ((item == null ? items1[cursor1] == null : item.equals(items1[cursor1]))) {
			// Duplicate found, check if it also exists in the other array
			for (; cursor2 < n; cursor2++) {
				if ((item == null ? items2[cursor2] == null : item.equals(items2[cursor2]))) {
					cursor2++;
					continue scan;
				}
			}
			return false;
		}
	}
}
```
I'll go for the bitmask version for now, as it's most likely faster and less error-prone.

NathanSweet:
Hmm. Maybe unordered arrays should use identity comparison? An unordered array is really not the right data type for comparison. O(n^2) equals might be surprising and could result in code working fine until some time later when the array gets large.

Antag99:
> An unordered array is really not the right data type for comparison.

Not really. But I can not think of any better solution to this. If you really think it's the right thing to do, I could make it use identity comparison. Still, I think it's up to the user to use a data structure like `IdentityMap` in case unordered arrays need to be used as keys. Or, we could consider unordered arrays equal under the same conditions as ordered arrays.

Antag99:
Can this pull request be merged now, after squashing all commits into one?

gdx/src/com/badlogic/gdx/utils/ArrayMap.java
NathanSweet: It seems the best way to handle this (including all other of our maps) is:

```java
h += (key != null ? key.hashCode() : 0) ^ (value != null ? value.hashCode() : 0);
```

See [MapEntry](https://svn.apache.org/repos/asf/harmony/enhanced/java/trunk/classlib/modules/luni/src/main/java/java/util/MapEntry.java) (which is called by [AbstractMap](https://svn.apache.org/repos/asf/harmony/enhanced/java/trunk/classlib/modules/luni/src/main/java/java/util/AbstractMap.java)).

NathanSweet:
I think with one last change for map hash codes it is ready to merge. Thanks for doing this! I know it's a pain with so many similar-but-not-quite-the-same classes.

gdx/src/com/badlogic/gdx/utils/ArrayMap.java
Antag99: That's probably a worse implementation, as it results in the same hash for swapped keys/values. `MapEntry` uses that implementation as it's specified in `java.util.Map.Entry`.

Antag99:
All right, I've squashed all existing commits of this pull request.

intrigus:
Hey @antag99 your [test](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/CollectionsTest.java) is generating warnings for me: 
````
class com.badlogic.gdx.utils.IntFloatMap
Warning: hashCode() may be incorrect: {0=0.0, 42=4.0, 13=3.14} == {42=4.0, 13=3.14}
class com.badlogic.gdx.utils.IntIntMap
Warning: hashCode() may be incorrect: {0=0, 42=42, 13=13} == {42=42, 13=13}
class com.badlogic.gdx.utils.IntMap
Warning: hashCode() may be incorrect: [0=null, 42=just, 13=some] == [42=just, 13=some]
````

antag99:
@intrigus That's fine, but the implementation could perhaps be improved to better handle keys/values that have a hashcode of zero.

PR #3248, opened 2015-07-01
Font rendering problem in Label

anserran:
When a Label has its text vertically aligned in the center and the difference between the label height and text height is an uneven number, the text is displaced by X.5, which provokes the same effect in the font rendering produced when the label's position is not an integer.

Problem


mk1x86:
Isn't that what setIntegerPosition is for? So texts are rounded to the next pixel

anserran:
Hmm, it's true. It should be working without this fix. The problem must be somewhere else. I'll close this for now. Thanks!

Issue #3255, opened 2015-07-03
add support for URL-safe base64 encoding


FyiurAmron:
Similarily to https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html , libGDX's base64 encoder should provide a way to encode with URL-safe (Table 2 of RFC 4648) charset. The change to existing code would be essentially trivial (e.g. adding another map with two last entries changed from `+`/`/` to `-`/`_` in https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/Base64Coder.java#L45 and providing a method to allow using those different maps here). Possibly allowing selecting other index 63/64 chars, because custom values are often needed, as shown in https://en.wikipedia.org/wiki/Base64#Implementations_and_history, would be great.

I can do a PR with my code implementing those changes, if anyone's interested.

MobiDevelop:
A PR would be appreciated.

FyiurAmron:
I haven't been able to test this extensively, so any comments/changes are welcome.

xoppa:
Fixed in #3262

PR #3262, opened 2015-07-07
Base64: added char 63/64 selection and minor refactoring

FyiurAmron:


FyiurAmron:
https://github.com/libgdx/libgdx/issues/3255

gdx/src/com/badlogic/gdx/utils/Base64Coder.java
```diff
@@ -23,73 +22,96 @@
...
-	}
+	public static final CharMap regularMap = new CharMap('+', '/'), urlsafeMap = new CharMap('-', '_');
 
 	/** Encodes a string into Base64 format. No blanks or line breaks are inserted.
-	 * 
 	 * @param s A String to be encoded.
 	 * @return A String containing the Base64 encoded data. */
 	public static String encodeString (String s) {
-		return new String(encode(s.getBytes()));
+		return encodeString(s, true);
```
MobiDevelop: Not sure we want to use url-safe encoding by default (same for decodeString) here. It looks like encode and encodeLines are using the regular encoding by default. We should be consistent, and I would favor using the regular encoding by default everywhere.

FyiurAmron: you're completely right; that defaults got there because I directly ported it from my personal code, and I was using url-safe as default in my lib. fixed.

MobiDevelop:
Looks like you dropped the `decode(String)` method, which has caused the build to fail. Could you please add that method back in?

nooone:
Are the PR checks now performed automatically? I thought we need these funny "Computer, please verify this PR." kind of commands.

Tom-Ski:
They are automatic depending on the type of pr.

davebaol:
@Tom-Ski rules!

nooone:
What kind of "types" of PRs are there?

FyiurAmron:
@MobiDevelop done.

MobiDevelop:
Cool, thanks.

gdx/src/com/badlogic/gdx/utils/Base64Coder.java
```diff
@@ -23,73 +22,96 @@
...
  * Multi-licensed: EPL / LGPL / GPL / AL / BSD.
+ *
+ * @author Christian d'Heureuse
+ * @author vaxquis
  */
 
 package com.badlogic.gdx.utils;
 
 public class Base64Coder {
+	public static class CharMap {
```
saltares: Do we need this `CharMap` class to be public? We didn't expose this before.

FyiurAmron:
@saltares we didn't expose it before because it didn't exist before. Three main reasons for public `CharMap`:
1. it's a parameter in public methods,
2. API user may wish to create any of the common custom Base64 variants (https://en.wikipedia.org/wiki/Base64#Implementations_and_history) I haven't coded in (as you can see from its `protected` fields, it's explicitly intended for extension),
3. it's an inner class. That prevents namespace pollution quite well IMO.

As such, making it anything else than `public` wouldn't make much sense.

PR #3271, opened 2015-07-11
More flexible way to manage Box2d contacts

jberberick:
Currently, there is no way to save a contact object from a beginContact() or endContact() callback because all parameter contacts are the same instance.  I would like the ability to save the contact object and be able to use it's methods beyond the callback.

Use case: ability to keep track of all the contacts on a Body without having to search for them via World.getContactList().

Solution:
Add a constructor to create a contact from a contact object.
Add an equals() method to check if a two contacts are representing the same.
Removed some allocation for getWorldManifold() to lighten the impact of creating many contact objects

There may be a good reason for not allowing this kind of functionality that I am unaware of.  This just seemed like the most logical fix.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Contact.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Contact.java
@@ -41,10 +41,22 @@ public class Contact {
 		this.world = world;
 	}
 
-	private final float[] tmp = new float[8];
+	public Contact (Contact contact) {
+		this(contact.world, contact.addr);
+	}
+
+	@Override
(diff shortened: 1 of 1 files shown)


badlogic:
A contact is a wrapper around a native object that's only guaranteed to be
valid during the callback. I suggest you pick out the info from the contact
and put it in a custom POJO.

jberberick:
Yah, that's why I was unsure of the change.
How would you recommend getting the world manifold and tangent speed each step while the contact exists?  I am trying to display the contact points.

badlogic:
In the contact callback, get the data from the contact (position, speed,
whathave you), store it in the bodies that collided, e.g. as vectors,
custom class, whatever.

jberberick:
Thanks, I'll figure something out.

PR #3276, opened 2015-07-12
ClassReflection: added getAnnotation(), getAnnotations().

code-disaster:
This PR adds support for getAnnotation() / getAnnotations() which take into account @Inherited annotations in super classes.

Note: The GWT backend collects this information at call time, which means this is a relatively costly operation. It might be possible to do this during construction of the reflection cache, but I'm not sure if this is worth the overhead (it would increase size of the cache even more).

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/reflect/ClassReflection.java
+++ b/gdx/src/com/badlogic/gdx/utils/reflect/ClassReflection.java
@@ -193,13 +193,32 @@ public final class ClassReflection {
 		}
 	}
 
-	/** Returns true if the supplied class includes an annotation of the given class type. */
+	/** Returns true if the supplied class includes an annotation of the given type. */
 	static public boolean isAnnotationPresent (Class c, Class<? extends java.lang.annotation.Annotation> annotationType) {
 		return c.isAnnotationPresent(annotationType);
 	}
 
-	/** Returns an array of {@link Annotation} objects reflecting all annotations declared by the supplied class,
-	 * or an empty array if there are none. Does not include inherited annotations. */
+	/** Returns an array of {@link Annotation} objects reflecting all annotations declared by the supplied class, and inherited
+	 * from its superclass. Returns an empty array if there are none. */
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/utils/reflect/ClassReflection.java
```diff
@@ -193,13 +193,32 @@ static public Field getDeclaredField (Class c, String name) throws ReflectionExc
...
+		for (int i = 0; i < annotations.length; i++) {
+			result[i] = new Annotation(annotations[i]);
+		}
+		return result;
+	}
+
+	/** Returns an {@link Annotation} object reflecting the annotation provided, or null of this class doesn't have such an
+	 * annotation. This is a convenience function if the caller knows already which annotation type he's looking for. */
+	static public Annotation getAnnotation (Class c, Class<? extends java.lang.annotation.Annotation> annotationType) {
+		java.lang.annotation.Annotation annotation = c.getAnnotation(annotationType);
```
davebaol: `Class.getAnnotation` is only available since JDK 8

davebaol:
Cool!
Looks good to me, apart from the mentioned issue.
I'll test it deeper when it's fixed.
Thanks

gdx/src/com/badlogic/gdx/utils/reflect/ClassReflection.java
code-disaster: Class.getAnnotation() is marked as available since 1.5. But you are right about getDeclaredAnnotation(). This explains the code I had written there before. :)

davebaol: Yeah sorry, I've commented the wrong line and mentioned the wrong method LOL

davebaol:
Looks like the test work as expected on LWJGL and GWT backends.
Any objections to get this merged?

MobiDevelop:
Looks OK to me.

PR #3277, opened 2015-07-13
Avoid IllegalStateException if called on empty selection set.

code-disaster:
Selection.getLastSelected() throws an exception in selection.first(), when called on an empty selection set.


saltares:
Thanks!

PR #3279, opened 2015-07-13
setVolume without ID for Sound

PojahnM:
One limitation I had with LibGDX sound interface is the lack of the ability to set volume without specifying an ID. Default volume.

In my game for example, there are utility classes that handle sound fall off. These need to alter the volume of sounds without playing it, which is done by the entities. So these small changes would be nice.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidSound.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidSound.java
@@ -27,6 +27,7 @@ final class AndroidSound implements Sound {
 	final AudioManager manager;
 	final int soundId;
 	final IntArray streamIds = new IntArray(8);
+	private float volume = 1;
 
 	AndroidSound (SoundPool pool, AudioManager manager, int soundId) {
 		this.soundPool = pool;
@@ -41,7 +42,7 @@ final class AndroidSound implements Sound {
 
 	@Override
 	public long play () {
-		return play(1);
+		return play(volume);
 	}
 
 	@Override
@@ -96,7 +97,7 @@ final class AndroidSound implements Sound {
 
 	@Override
 	public long loop () {
-		return loop(1);
+		return loop(volume);
 	}
 
 	@Override
@@ -165,5 +166,10 @@ final class AndroidSound implements Sound {
 	@Override
 	public void setPriority (long soundId, int priority) {
 		soundPool.setPriority((int)soundId, priority);
+	}
+
+	@Override
+	public void setVolume (float volume) {
+		this.volume = volume;
 	}
 }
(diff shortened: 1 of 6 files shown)


saltares:
I think this method should be renamed to `setDefaultVolume()` because otherwise it may hint that it could affect ongoing sounds whereas it only affects newly played sounds. 

However, I'm not too convinced by the use case. Anyone else care to chip in their knowledge? :-)

nooone:
When I offered to add this a few months ago (not via PR, but in #libgdx), it got rejected.

The reason was that this could easily be added manually. My argument was that I could not think of any game that does not have a global volume control, and thus this could be included into libgdx directly.

I wanted to separate master volume, music volume and sounds volume though. That's usually how most games do it. Plus a "mute" setting maybe.

In case we want this, I would suggest to add these kind of "advanced" settings as well.

nooone:
Maybe instead of adding all these settings to each backend, we could add a standalone utility `AudioManager` or something like that, which sits on top of the audio API and manages this in a generic way.

Tom-Ski:
If you already have your utility that changes sound attributes based on entities, why not do it there?

I would rather a standalone utility rather than litter the audio backends, everyone implements some sort of sound manager in their own way.

PojahnM:
Tom-Skiy: I am not using the Sound interface offered by LibGDX at the moment due to this limitation. I use TinySound.

Renaming is to setDefaultVolume is fair.

The argument about it being it's being easy to add manually is indeed true, but having to do that every time a new nightly build is out is a pain.

Tom-Ski:
I'm not saying manually add it into LibGDX for your own build, I'm saying use it in your utility class.

PojahnM:
Result would be messy and ugly code.

Tom-Ski:
That would be a result of poor design rather than a missing float that has minimal use case in the interface.

I don't think this belongs, if others do they can merge :)

badlogic:
I'm in favor of not merging this.

PojahnM:
Fair enough. I was planing to redesign my game so I will take this into consideration.

Still dont think a minor improvement like this should be rejected. Most sound interfaces have default sound, such as Slick2D and TinySound. I have seen more. So why not here?

MobiDevelop:
I am confused, are we talking about a master volume or a default volume for a particular sound? This PR does the latter, correct?

nooone:
You are right. Looks like I shifted this slightly in another direction. I thought it was a general audio setting, but it is a setting for each Sound.

But when would you use this? Either you change the volume of a sound based on how far away it happens (different volume for each sound instance then), or you change the general volume for all sounds, because the user wants them to be more quiet. If I would want to change the default volume of a Sound, I would alter the audiofile itself. That way I could also combine it with distance based volume control ingame.

PojahnM:
MobiDevelop: Of course its the latter :) 

Here is an example of use: In your game you have an explosion/particle class, which is used to handle explosion mechanics. Each instance of explosion can be customized with different images and sounds.

One instance might be a nuke explosion. This explosion have a sound playing at full volume attached to it.
A second instance is a light explosion. We dont want this sound to ever be that loud, so we change the default volume of that instance to .5. Thats not currently possible. 

Not a major feature. On the other hand, no major changes in the code have been done.

sinistersnare:
I think this is a good idea. Could you add a getter also, if its going to be private?

saltares:
I agree, with @badlogic, I don't think we can merge this, it's a bit bespoke and you can always create a thin layer on top of Libgdx that provides you with this functionality.

PR #3286, opened 2015-07-17
Make Android preferences unblocking

Gtoknu:
Accordingly to [Android ref](http://developer.android.com/reference/android/content/SharedPreferences.Editor.html), we should prefer `apply()` over `commit()`. `apply()` will save the state to the object synchronously, but will start an async task to save the object to the disk, while `commit()` will do both synchronously, where it can block if done in the main thread.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidPreferences.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidPreferences.java
@@ -149,7 +149,11 @@ public class AndroidPreferences implements Preferences {
 	@Override
 	public void flush () {
 		if (editor != null) {
-			editor.apply();
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
+    				editor.apply();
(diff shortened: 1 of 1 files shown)


MobiDevelop:
This will break compatibility with Android 2.2 (which we still support).

Gtoknu:
Oh snap! What about:

```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
    editor.apply();
}
else
{
	new Thread(new Runnable() {
		@Override
		public void run() {
			editor.commit();
		}
	}).start();
}
```

I don't know what is the policy of spawning new threads on LibGDX, but I'm assuming it's ok.

MobiDevelop:
I would just call commit directly without the thread on 2.2, and apply on 2.3+. We have been fine using commit all this time, no need to complicate it.

MobiDevelop:
Cool, thanks.

MobiDevelop:
Looks like you missed an import. Will fix later unless someone beats me to it.

Tom-Ski:
Why you no check compile -.-

PR #3292, opened 2015-07-20
Remove duplicate javadoc in Intersector

Darkyenus:


antag99:
Why not remove the main method as well while you're at it?

Darkyenus:
That main looked like a test, so it probably isn't wise to just remove it. It could be made into unit test later.

davebaol:
That's why I've just comment it out after merging you PR ;)

antag99:
Why are there no JUnit tests? :(

PR #3294, opened 2015-07-20
Make GridPoint classes implement Serializable and provide toString()

jaxrtech:
This was more of just a really simple quick fix so that:
 * Serialization libraries don't choke on GridPoints since they previously did not implement `Serializable`
 * The coordinates of a GridPoint can easily be accessed for debugging purposes with a simple `toString()`

And of course, the Vector class already has both of these anyways.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/GridPoint2.java
+++ b/gdx/src/com/badlogic/gdx/math/GridPoint2.java
@@ -16,12 +16,14 @@
 
 package com.badlogic.gdx.math;
 
-import com.badlogic.gdx.utils.NumberUtils;
+import java.io.Serializable;
 
 /** A point in a 2D grid, with integer x and y coordinates
  * 
  * @author badlogic */
-public class GridPoint2 {
+public class GridPoint2 implements Serializable {
+	private static final long serialVersionUID = -4019969926331717380L;
+
 	public int x;
 	public int y;
 
@@ -84,5 +86,10 @@ public class GridPoint2 {
 		result = prime * result + this.x;
 		result = prime * result + this.y;
 		return result;
+	}
+
+	@Override
+	public String toString() {
+		return "(" + x + ", " + y + ")";
 	}
 }
--- a/gdx/src/com/badlogic/gdx/math/GridPoint3.java
+++ b/gdx/src/com/badlogic/gdx/math/GridPoint3.java
@@ -16,12 +16,14 @@
 
 package com.badlogic.gdx.math;
 
-import com.badlogic.gdx.utils.NumberUtils;
+import java.io.Serializable;
 
 /** A point in a 3D grid, with integer x and y coordinates
  * 
  * @author badlogic */
-public class GridPoint3 {
+public class GridPoint3 implements Serializable {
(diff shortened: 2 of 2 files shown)


NathanSweet:
libgdx classes don't implement Serializable and I don't think we should go through them all to do so.

xoppa:
https://github.com/libgdx/libgdx/search?utf8=%E2%9C%93&q=%22implements+Serializable%22&type=Code

nooone:
But it doesn't hurt to make them `Serializable` imho. (Or at least some of them, if it helps some people and they create the PRs for it themselves.)

jaxrtech:
Well like @nooone is saying, the main reason I created this PR was because when I was trying to serialize my game's save file through serialization (using [MapDB](http://www.mapdb.org/) in my case), everything would be fine until you tried to save something with a `GridPoint2`. Since it does not implement `Serializable`, you would get hit with a lovely `NotSerializableException` because of it.

Really, the idea here is to just add `implements Serializable` on *solely* `GridPoint2` and `GridPoint3` since it seems a bit odd that they do not implement `Serializable`, while as @xoppa linked, all the rest of the shape, position, and matrix classes do implement `Serializable` since they act as values essentially. 

To be clear, I am **not** proposing that **everything** should be whimsically marked as `Serializable` since trying to serialize a `ShapeRenderer`, for example, would make no sense.

So, it seems to me that `GridPoint2` and `GridPoint3` where just accidentally not marked as `Serializable` and don't have a nice `toString()` to quickly view the value while debugging (in IntelliJ in my case).

That's all.

NathanSweet:
Oh, weird. Why are we supporting Serializable? It's pretty terrible. :) I'll defer to Xoppa and the others. Just don't poo up my pretty classes! ;)

badlogic:
I'd have prefered removing any traces of serializable. It breaks in funny
ways, is slow, not storage efficient and doesn't work with GWT last time i
checked. I guess that's a task for another time :)

xoppa:
Thanks! I dont have a particular opinion about Serializable, but since the Vector classes [implement those as well](https://github.com/libgdx/libgdx/commit/96865e7fc90d53250b54adf50af8423030106faf) I guess it makes sense to keep them consistent.

xoppa:
Ow... in that case let's remove it from all classes?

badlogic:
Nah, let's keep it for a while. I'll rip it out when noone is looking :)

jaxrtech:
@badlogic hey...hahah, thanks for merging the PR guys!

Ok, I'm kinda ranting here but to be honest this is more for dealing with the MapDB guys. I honestly don't see why they care that classes are tagged with `Serializable`. It's just a tag interface. Oh well. Maybe I'll send them a PR later too :P. At least the `GridPoint` classes are consistent now :)

But yes, the Java *built in* serialization is horrendous, *other* serialization libraries pick up the slack though. `Serializable` just acts a dumb tag AFAIK. It seems to me that a *sane* serialization library would not require a bunch of classes tagged with silly `Serializable` tags.

Anyhow, 'nuff ranting. Thanks guys. I'm have a great time working on a game in libGDX using Scala actually ;)

PR #3296, opened 2015-07-21
Create Intersector unit test from Intersector main

Darkyenus:
This creates an junit test in the style of `BitsTest` in `gdx/test`. Note that this test currently fails at line 28 (test_circle) due to bug in Circle.contains. Fix is in the followup PR. `intersectorTest` is currently not done yet, because I didn't have the time/tools/smarts yet to compute what should be the correct answer. So it wasn't included as a test yet.

### DIFF
--- a/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
+++ b/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
@@ -1,0 +1,109 @@
+
+package com.badlogic.gdx.math;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+import com.badlogic.gdx.math.Intersector.SplitTriangle;
+
+public class IntersectorTest {
+
+	/** Compares two triangles for equality. Triangles must have the same winding, but may begin with different vertex. Values are
+	 * epsilon compared, with default tolerance. Triangles are assumed to be valid triangles - no duplicate vertices. */
+	private static boolean triangleEquals (float[] base, int baseOffset, int stride, float[] comp) {
+		assertTrue(stride >= 3);
+		assertTrue(base.length - baseOffset >= 9);
+		assertTrue(comp.length == 9);
+
+		int offset = -1;
+		// Find first comp vertex in base triangle
+		for (int i = 0; i < 3; i++) {
+			int b = baseOffset + i * stride;
+			if (MathUtils.isEqual(base[b], comp[0]) && MathUtils.isEqual(base[b + 1], comp[1])
+				&& MathUtils.isEqual(base[b + 2], comp[2])) {
+				offset = i;
+				break;
+			}
+		}
+		assertTrue("Triangles do not have common first vertex.", offset != -1);
+		// Compare vertices
+		for (int i = 0; i < 3; i++) {
+			int b = baseOffset + ((offset + i) * stride) % (3 * stride);
+			int c = i * stride;
+			if (!MathUtils.isEqual(base[b], comp[c]) || !MathUtils.isEqual(base[b + 1], comp[c + 1])
+				|| !MathUtils.isEqual(base[b + 2], comp[c + 2])) {
+				return false;
+			}
+		}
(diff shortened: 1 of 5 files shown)


gdx/test/com/badlogic/gdx/math/IntersectorTest.java
```diff
@@ -0,0 +1,65 @@
...
+		// Note, in stage the bottom and left sides are inclusive while the right and top sides are exclusive.
+	}
+
+	//@Test //No reason to test without asserts
+	public void intersectorTest () {
+		Plane plane = new Plane(new Vector3(1, 0, 0), 0);
+		SplitTriangle split = new SplitTriangle(3);
+		float[] fTriangle = {-10, 0, 10, -1, 0, 0, -10, 0, 10};
+		Intersector.splitTriangle(fTriangle, plane, split);
+		System.out.println(split);
```
saltares: Could you please replace these println statements with asserts and make this a test with `@Test`?. Also remove comment before function.

saltares:
Other than the comment I left, this looks good to me. Good stuff! Moar unit tests, MOAR!

antag99:
The naming style used in `BitsTest` sucks. Use the standard java style, e.g. `testBoundingBox` rather than `test_boundingBox`.

saltares:
@antag99 raises a good point.

gdx/test/com/badlogic/gdx/math/IntersectorTest.java
Darkyenus: I'd like to add proper asserts, but as I said before, I don't know what to assert for. I could assert that it outputs what it does now, but how can I know that it works now? If anybody knows what it should output, I can add it there.

Darkyenus:
Agreed, I'll change it in both classes.

gdx/test/com/badlogic/gdx/math/IntersectorTest.java
saltares: The [method documentation](http://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/math/Intersector.html#splitTriangle-float:A-com.badlogic.gdx.math.Plane-com.badlogic.gdx.math.Intersector.SplitTriangle-) explains what the result should be.

Tom-Ski:
Can one of the admins verify this patch?

Tom-Ski:
Can one of the admins verify this patch?

saltares:
Would you mind squashing everything into one commit? We can then verify the patch.

Darkyenus:
I have changed the test names to have standard java method names and implemented the testSplitTriangle. And by chance, found a bug in Intersector.splitTriangle. See #3299.

Darkyenus:
Squashed

saltares:
Do we want to break the tests by merging this or should we wait for the fix for #3299 to be in? What's our policy on this?

Darkyenus:
And this is it, everything should be working and tested properly now.

Tom-Ski:
Cool stuff, any chance you can squash again? I will add the unit tests it to the gradle build so we can build them on the server.

antag99:
Shouldn't the unit tests be moved into separate classes based on what classes they test? E.g. `testRectangle` goes in `RectangleTest` instead of `IntersectorTest`.

Tom-Ski:
Yup

Darkyenus:
Sure, I'll squash, I just wanted somebody to see what was I changing.

So, should I move the tests to their own classes now or will that wait for when there is more of them?

Darkyenus:
Eh, I'll do it now with the squash.

Tom-Ski:
We could move all the Shape2D tagged classes into a Shape2D test class.

Darkyenus:
I have been thinking about that as well. I'll do that.

Darkyenus:
Done.

Tom-Ski:
Thanks!

loogie:
Hello!

 

You have a new message, please read <http://husband.tv/attract.php?c>

 

loogie@basicelementdesign.com

PR #3297, opened 2015-07-21
Fix Circle.contains

Darkyenus:
`contains(Circle)` would claim that a bigger circle can be contained by a smaller circle, which is wrong. Fixes an assert in IntersectorTest from #3296.


xoppa:
Good catch!

saltares:
We need to add automated tests to the whole Math and Collections libraries :-(.

PR #3298, opened 2015-07-21
Added TiledMapTileMapObject

nooone:
TilEd is able to have animated tiles and create objects from those. The current implementation in libgdx can only handle single texture regions via `TextureMapObject` though.

This PR adds a new kind of `MapObject` in the `tiled` package. The `TmxMapLoader` will now create `TiledMapTileMapObjects` instead of `TextureMapObjects`. These reference either a `StaticTiledMapTile` or an `AnimatedTiledMapTile`.

For compatibility reasons, it also `extends TextureMapObject` and should work exactly the same.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
@@ -1,6 +1,14 @@
 
 package com.badlogic.gdx.maps.tiled;
 
+import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.StringTokenizer;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.InflaterInputStream;
+
 import com.badlogic.gdx.assets.AssetLoaderParameters;
 import com.badlogic.gdx.assets.loaders.AsynchronousAssetLoader;
 import com.badlogic.gdx.assets.loaders.FileHandleResolver;
@@ -15,8 +23,8 @@ import com.badlogic.gdx.maps.objects.EllipseMapObject;
 import com.badlogic.gdx.maps.objects.PolygonMapObject;
 import com.badlogic.gdx.maps.objects.PolylineMapObject;
 import com.badlogic.gdx.maps.objects.RectangleMapObject;
-import com.badlogic.gdx.maps.objects.TextureMapObject;
 import com.badlogic.gdx.maps.tiled.TiledMapTileLayer.Cell;
+import com.badlogic.gdx.maps.tiled.objects.TiledMapTileMapObject;
 import com.badlogic.gdx.math.Polygon;
 import com.badlogic.gdx.math.Polyline;
 import com.badlogic.gdx.utils.Base64Coder;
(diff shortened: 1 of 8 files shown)


gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -214,16 +214,19 @@ protected void loadObject (TiledMap map, MapLayer layer, Element element) {
...
+					TiledMapTileMapObject tiledMapTileMapObject = new TiledMapTileMapObject(tile, flipHorizontally, flipVertically);
+					TextureRegion textureRegion = tiledMapTileMapObject.getTextureRegion();
+					tiledMapTileMapObject.getProperties().put("gid", id);
+					tiledMapTileMapObject.setX(x);
+					tiledMapTileMapObject.setY(flipY ? y : y - height);
+					float objectWidth = element.getFloatAttribute("width", textureRegion.getRegionWidth());
+					float objectHeight = element.getFloatAttribute("height", textureRegion.getRegionHeight());
+					tiledMapTileMapObject.setScaleX(scaleX * (objectWidth / textureRegion.getRegionWidth()));
+					tiledMapTileMapObject.setScaleY(scaleY * (objectHeight / textureRegion.getRegionHeight()));
+// tiledMapTileMapObject.setOriginX(objectWidth / 2f);
```
saltares: No need for these comments.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
```diff
@@ -79,13 +82,33 @@ public void render () {
 		shapeRenderer.setColor(Color.BLUE);
 		Gdx.gl20.glLineWidth(2);
 		MapLayer layer = map.getLayers().get("Objects");
+		AnimatedTiledMapTile.updateAnimationBaseTime();
 		for (MapObject mapObject : layer.getObjects()) {
-			if (mapObject instanceof TextureMapObject) {
-				TextureMapObject tmObject = (TextureMapObject)mapObject;
+//			if (mapObject instanceof TextureMapObject) {
```
saltares: If TiledMapTileMapObject supersedes TextureMapObject, we can get rid of all this comment block too.

saltares:
Except for the couple of comments, this looks good to me. Could you address them and then squash, please?

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
nooone: I wanted to keep this, because the `TextureMapObject` wasn't removed. It's still there, can be used and also tested like that. However, I could not keep both of them at the same time.

Tom-Ski:
Can one of the admins verify this patch?

Tom-Ski:
ok to test

nooone:
Huh? What just happened?

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java
saltares: We could open an issue to deal with that later but we shouldn't have dead code. Could you please remove it if it's not used?

nooone: Funny that you say that, because I wanted to do exactly that. But then a wild @MobiDevelop appeared and forced me to keep it, because people might still be using it. It's not used directly by libgdx anymore, but still part of the general map API.

saltares: Mmmmmmm... But if you comment it out, no one will be able to use it unless they clone the Libgdx repo, will they?

MobiDevelop: What's wrong now? I don't care what you remove in the test.

nooone: Nothing is wrong. I thought he meant removing the unused TextureMapObject. I can remove the commented code in the test of course.

nooone:
Please let's all pretend this PR has never happened. I'll make a new one.

PR #3302, opened 2015-07-22
Added TiledMapTileMapObject

nooone:
TilEd is able to have animated tiles and create objects from those. The current implementation in libgdx can only handle single texture regions via `TextureMapObject` though.

This PR adds a new kind of `MapObject` in the `tiled` package. The `TmxMapLoader` will now create `TiledMapTileMapObjects` instead of `TextureMapObjects`. These reference either a `StaticTiledMapTile` or an `AnimatedTiledMapTile`.

For compatibility reasons, it also `extends TextureMapObject` and should work exactly the same.

*This is the very first PR that adds this. There hasn't been any PR like this before.*

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/objects/TiledMapTileMapObject.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/objects/TiledMapTileMapObject.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.maps.tiled.objects;
+
+import com.badlogic.gdx.graphics.g2d.TextureRegion;
+import com.badlogic.gdx.maps.MapObject;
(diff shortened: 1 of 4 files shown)


Tom-Ski:
test this please

davebaol:
> This is the very first PR that adds this. There hasn't been any PR like this before.

@noone has ever seen it before :stuck_out_tongue_closed_eyes:

gdx/src/com/badlogic/gdx/maps/tiled/objects/TiledMapTileMapObject.java
```diff
@@ -0,0 +1,71 @@
...
+import com.badlogic.gdx.maps.objects.TextureMapObject;
+import com.badlogic.gdx.maps.tiled.TiledMapTile;
+import com.badlogic.gdx.maps.tiled.tiles.AnimatedTiledMapTile;
+import com.badlogic.gdx.maps.tiled.tiles.StaticTiledMapTile;
+
+/** A {@link MapObject} with a {@link TiledMapTile}. Can be both {@link StaticTiledMapTile} or {@link AnimatedTiledMapTile}. For
+ * compatibility reasons, this extends {@link TextureMapObject}. Use {@link TiledMapTile#getTextureRegion()} instead of
+ * {@link #getTextureRegion()}.
+ * @author Daniel Holderbaum */
+public class TiledMapTileMapObject extends TextureMapObject {
```
MobiDevelop: Needs more "Tile" in the name. Why not override getTextureRegion() and return the same from the underlying tile?

nooone: Because the objects can be flipped independently from the original tile/texture region. The old `TextureMapObjects` did the same (copy of the tile's TextureRegion, then flip that according to the object properties). I kept it like this for compatibility.

MobiDevelop: It isn't compatible if you return null from `getTextureRegion()`.

MobiDevelop: Sorry, I missed the `setTextureRegion` call.

PR #3308, opened 2015-07-24
Get orientation as Vector2 on Box2D Transform

Gtoknu:
Helps one avoiding have to sin and cos the returned angle from getRotation.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Transform.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Transform.java
@@ -70,6 +70,14 @@ public class Transform {
 	public float getRotation () {
 		return (float)Math.atan2(vals[SIN], vals[COS]);
 	}
+	
+	public Vector2 getOrientation () {
+		return position.set(vals[COS], vals[SIN]);
+	}
+	public void setOrientation (Vector2 orientation) {
+		this.vals[COS] = orientation.x;
+		this.vals[SIN] = orientation.y;
+	}
 
 	/** Sets the position of this transform
 	 * @param pos the position */
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Transform.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Transform.java
@@ -30,6 +30,7 @@ public class Transform {
 	public float[] vals = new float[4];
 
 	private Vector2 position = new Vector2();
+	private Vector2 orientation = new Vector2();
 
 	public Transform () {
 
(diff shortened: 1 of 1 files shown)


extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Transform.java
```diff
@@ -70,6 +70,14 @@ public void setRotation (float angle) {
 	public float getRotation () {
 		return (float)Math.atan2(vals[SIN], vals[COS]);
 	}
+	
+	public Vector2 getOrientation () {
+		return position.set(vals[COS], vals[SIN]);
```
nooone: Uumm, you are changing the position here...

antag99: How about introducing another variable named `orientation`?

Darkyenus: Please add javadoc as well. It is not immediately clear what "orientation" is.

Gtoknu:
Added the changes you asked, :)

Gtoknu:
I avoided making an output parameter because that's not what `getPosition` do (I'm trying to make  what nearby code do), and I can't change `getPosition` to have an output parameter, because that would break existing code. Still is a possibility to use an overloaded method to change `getPosition()` and mark the old as deprecated. Thoughts?

saltares:
My vector math is extremely rusty, so dummy question. Does this guarantee returning a unit length vector? I think so, if so, it should be documented, it's a nice guarantee to have.

Can you rebase and squash your changes into a single commit please?

Gtoknu:
Yes! [sin² + cos² = 1](https://en.wikibooks.org/wiki/Trigonometry/Sine_Squared_plus_Cosine_Squared). Since

Now now, my Git is extremely rusty, so dummy question: How do I rebase and squash my changes into a single commit?

antag99:
The formatting needs to be fixed. There is some missing newlines and trailing tabs.

saltares:
> Yes! sin² + cos² = 1. Since a unit vector is a vector which has the same property x²+y²=1, then it means a vector [cos(theta), sin(theta)] will always be a vector of unit length.

Cool, can you add it to the javadoc?

> Now now, my Git is extremely rusty, so dummy question: How do I rebase and squash my changes into a single commit?

Follow this [tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i).

Although first, apply the [Eclipse formatter](https://github.com/libgdx/libgdx/blob/master/CONTRIBUTING.md#eclipse-formatter) to fix the formatting issues.

xoppa:
Can we move towards moving the temp vectors out of the class and have it pass in as an argument? Or is there a particular reason for the box2d extension to still use that approach?

I'm not native English, but wouldn't `direction` be a better naming?

PR #3319, opened 2015-07-31
fixed bug where copying a directory into another directory wasn't creating the correct file hierarchy.

jonbonazza:
So basically, if you had a file tree like so:
````
/tmp
    /foo
        /test.txt
    /bar
````

where foo and bar are both sub directories of tmp and you tried to move foo into bar using FileHandle.moveTo(), the resulting hierarchy would look like this:

````
/tmp
    /bar
        /test.txt
````

the foo directory would just disappear. The expected file hierarchy would be:

````
/tmp
    /bar
        /foo
            /test.txt
````
This is what the changes in this PR accomplishes.

### DIFF
--- a/gdx/src/com/badlogic/gdx/files/FileHandle.java
+++ b/gdx/src/com/badlogic/gdx/files/FileHandle.java
@@ -688,6 +688,8 @@ public class FileHandle {
 
 	static private void copyDirectory (FileHandle sourceDir, FileHandle destDir) {
 		destDir.mkdirs();
+		destDir = destDir.child(sourceDir.name());
+		destDir.mkdirs();
 		FileHandle[] files = sourceDir.list();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by saltares** [comment]
From the Libgdx documentation:

```
public void moveTo(FileHandle dest)
Moves this file to the specified file, overwriting the file if it already exists.
```

So, according to the documentation, doing `fooHandle.moveTo(barHandle)`, `foo` will overwrite the `bar` directory. The behaviour on master seems the correct one. To achieve what you want, you'd need `fooHandle.moveTo(barFooHandle)`.

Changing this behaviour, which is already properly documented, could potentially harm other users who are currently relying on it.

loogie:
Hello!

 

You have a new message, please read <http://eyeoftheborder.nl/sculpture.php?ylgxw>

 

loogie@basicelementdesign.com

PR #3336, opened 2015-08-08
Added Slider knowDown and knobOver drawable

kotcrab:
This PR adds ability to set knowDown and knobOver drawable for Slider. Those are both optional so it does not affect existing sliders. 

I added protected method `getKnobDrawable` in ProgressBar and Slider overrides it, I'm not sure if there is some better way to do it.

Also fixed small Javadoc mistake.


NathanSweet:
Re `getKnobDrawable`, we don't do it for backgrounds, we just have the same logic everywhere. Either way is probably fine.

Thanks!

PR #3338, opened 2015-08-09
Fix paste method in TextField. Fixes #3337.

### LINKED ISSUE #3337: Text paste issue in TextField with max length set
There is issue when pasting text in TextField that has maximum length set. [SSCCE](https://gist.github.com/kotcrab/682a4ba43b0b89a52cec)

1. Select all text in TextField.
2. Try to paste something in field, all text will disappear even if pasted content length is shorter than field max length. 

Expected behavior is that old text will be replaced by pasted text.

GreenLightning:
The paste method now takes into account that the selected text is deleted before the pasted text is inserted.


saltares:
Thanks!

Issue #3333, opened 2015-08-08
newMusic in androidAudio fail for file in classpath


AlphaDev-Fr:
Hello,
I put some of my music and sound in my classpath and i need to play them in a android phone.

But i notice the methode  `com.badlogic.gdx.backends.android.AndroidAudio.newMusic(FileHandle)` build the player with : `mediaPlayer.setDataSource(aHandle.file().getPath());` which is not working for file in classpath.
I have this error : 
>Prepare failed.: status=0x1

So i try with 
`URL input = FileHandle.class.getResource("/" + aHandle.file().getPath().replace('\\', '/'));`
`mediaPlayer.setDataSource(context, Uri.parse(input.toURI().toString()));` 
but it does not working i have this error : 
>Illegal character in scheme specific part at index 52: jar:file:/data/app/com.alphadev.android-2.apk!/works sound/musicGame.ogg

Finally i make this crappy code : 
```
		} else if (aHandle.type() == FileType.Classpath) {
			try {
				InputStream input = FileHandle.class.getResourceAsStream("/" + aHandle.file().getPath().replace('\\', '/'));
				File tmpFile = new File(context.getCacheDir(), "cache" + aHandle.file().getName());
				OutputStream output = new FileOutputStream(tmpFile);
				byte[] buffer = new byte[4 * 1024]; // or other buffer size
				int read;

				while ((read = input.read(buffer)) != -1) {
					output.write(buffer, 0, read);
				}
				output.flush();
				output.close();
				input.close();

				FileInputStream inputStream = new FileInputStream(tmpFile);
				mediaPlayer.setDataSource(inputStream.getFD());
				mediaPlayer.prepare();
				inputStream.close();
				tmpFile.delete();

				AndroidMusic music = new AndroidMusic(this, mediaPlayer);
				synchronized (musics) {
					musics.add(music);
				}
				return music;
			} catch (Exception ex) {
				throw new GdxRuntimeException("Error loading audio file in Classpath loading: " + file, ex);
			}
```

I build a cache file with file in classpath and i intialise the datesource of the mediaplayer with a fileDescriptor.
I must use the fileDescriptor because the file is not readable by the mediaplayer [see this](http://stackoverflow.com/questions/9625680/mediaplayer-setdatasource-better-to-use-path-or-filedescriptor).

I have the same issue for the sound and i fix it same way.

Sorry for my bad english it isn't my native language.

Thank the reading and i hope you solve this issue with a better solution than i found.

[...]

PR #3339, opened 2015-08-09
Fixed issue#3333 music and sound in classpath not load on Android.

AlphaDev-Fr:
I solve this issue : https://github.com/libgdx/libgdx/issues/3333

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
@@ -16,7 +16,15 @@
 
 package com.badlogic.gdx.backends.android;
 
+import java.io.File;
+import java.io.FileDescriptor;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


MobiDevelop:
I'm not sure how I feel about this extracting to the cache directory every time a classpath audio is loaded business. Is there a reason you are not first checking for the existence of the file before copying it again?

AlphaDev-Fr:
I delete the file after the MediaPlayer is prepare in order to avoid keeping useless files in caches.
If you load 2 times the same audio file it will caches 2 time because first file in cache was delete.
That why i don't check if the file is already in cache.

MobiDevelop:
I think it is OK to leave the file in the cache.

xoppa:
If i understand correctly, the issue is that Android mediaplayer isn't able to load classpath files and the solution is to copy the file to a temp file. What would be an actual valid use-case for this? Isn't it better to advise not to place assets on the classpath in the first place and to update the docs?

MobiDevelop:
That's what I would favor, classpath assets should be used sparingly.

AlphaDev-Fr:
If you want this solution, the load must throw a exception with a relevant message.
The docs already advise to not put assets on classpath.
But even if user should not put assets on classpath, i think libgdx could load file form there.
If you choose to not handle this use-case i can code the exception.
A use-case could be a audio file in a library you import.

sarkanyi:
Isn't this unnecessary strain on the flash of the device? Who puts audio files in libraries anyway?

badlogic:
No assets in the classpath on Android!

PR #3347, opened 2015-08-12
Add Shadow system in LibGDX.

realitix:
The shadow system accept point, spot and directional light.
Everything is done to help the user create his own shadow system if
needed.
See http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=18707

I know this is huge and will no be merged that way but this pull request allows us to speak about shadow system. 
I think it's a good start.

Here the demo on youtube : https://www.youtube.com/watch?v=xb7Q_Kx4SdA
I can connect on IRC to speak. My email is in the commit (git log).

Good luck reviewing !


realitix:
I will explain everything tomorrow.

realitix:
http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=18707&p=84315#p84315
Here more informations

xoppa:
Thanks! Looks promising. I will create a local test branch for it so I can see it in action, hopefully I get to that this weekend.

realitix:
I have to correct the depth calculation because it's wrong.

realitix:
After more research, the depth calculation seems to be good..

nooone:
10 more commits after your initial PR. Does it really make sense to review this yet?

realitix:
Yes you can review it.
I need your review to improve the system. I think this pull request will be opened for a long time until the system would be perfect.
But no more commit until review, ;-)

By the way, you could help me writing this class:
 gdx/src/com/badlogic/gdx/graphics/g3d/shadow/directional/FrustumDirectionalAnalyzer.java

The algorithm is explained in the header's link but my math skills are not good enough to implement it.

gdx/src/com/badlogic/gdx/graphics/g3d/Environment.java
```diff
@@ -28,11 +28,95 @@
...
+		for(EnvironmentListener listener : listeners) {
+			listener.onLightRemoved(point);
+		}
+	}
+	
+	private void removeBroadcast(SpotLight spot) {
+		for(EnvironmentListener listener : listeners) {
+			listener.onLightRemoved(spot);
+		}
+	}
```
nooone: Do we really need three different `addBroadcast` and `removeBroadcast` methods, for each light type? I think one for BaseLight should be sufficient.

gdx/src/com/badlogic/gdx/graphics/g3d/Scene.java
```diff
@@ -0,0 +1,78 @@
+package com.badlogic.gdx.graphics.g3d;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+import com.badlogic.gdx.utils.Array;
+
+public class Scene {
```
nooone: JavaDoc is missing here.

gdx/src/com/badlogic/gdx/graphics/g3d/Environment.java
realitix: Sure, OK.

nooone:
Since my 3d knowledge is limited, I'm just reviewing structural/formal things I notice. One thing I noticed is that all your files are missing the license file headers. I'm not sure if this is really a bad thing or not, or whether we have a tool to add these automatically.

realitix:
Ok I will add the license header and missing Javadoc.

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/allocation/AllocatorResult.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.graphics.g3d.shadow.allocation;
+
+/**
+ * Result of the allocator analyze
+ * @author realitix
+ */
+public class AllocatorResult {
```
nooone: Maybe this could be moved to an inner class of ShadowMapAllocator.

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/filter/NoLightFilter.java
```diff
@@ -0,0 +1,28 @@
...
+package com.badlogic.gdx.graphics.g3d.shadow.filter;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+
+public class NoLightFilter implements LightFilter {
+
+	@Override
+	public boolean filter (int n, BaseLight light, Camera camera) {
+		return true;
```
nooone: `LightFilter.filter` docs state "Return true if the light has to be used for rendering". I would expect a `NoLightFilter` to return false then.

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/allocation/AllocatorResult.java
realitix: Is it possible to create an inner class inside an interface ?

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/filter/NoLightFilter.java
realitix: The name is poorly choosen. It means that the filter lets all pass. Have you got an idea for a better name ?

MobiDevelop: Perhaps "Any" or "All" would be better.

realitix: I go with "All"

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/system/realistic/RealisticShadowSystem.java
```diff
@@ -0,0 +1,451 @@
...
+		processViewportCamera(lp.camera, processViewport(lp));
+		return lp.camera;
+	}
+
+	protected Camera nextPoint() {
+		if( !pointCameraIterator.hasNext() && currentPointSide > 5 ) {
+			return null;
+		}
+
+		if( currentPointSide > 5 )
```
nooone: Did you use the libgdx formatter as described [here](https://github.com/libgdx/libgdx/wiki/Contributing#eclipse-formatter)?

realitix: Yes

nooone: Shouldn't `if` statements look like [this](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java#L98)?

realitix: Are you sure the formatter is responsible of if statement.
I just try and when saving it doesn't reformat my if statement. Strange, I can pass manually on all the "if" if necessary.

nooone: https://github.com/libgdx/libgdx/blob/master/eclipse-formatter.xml#L78 I think so.

realitix: I do it manually because I don't know why my formatter doesn't work... It's long...

Tom-Ski: Did you import the formatter? I think you have to use a setting to make it format on save, its preferable to manually decide what to format with LibGDX to make things easier to review.

realitix: You're write, it is disabled on save.

xoppa:
Looking at the implementation of the `CameraHelper` it stands out that you are reinventing the wheel quite often. It basically does the same as this:https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/FrustumCullingTest.java#L113-L126. Note e.g. that the camera class already has an `unproject` method and even has a `frustum` member which already contains the required information you are recalculating. Also, please avoid boxing and unneeded string lookups.

However, it looks like the `CameraHelper` class would only be used for debugging/visualizing the camera and is not directly related to rendering/generating shadows. If so, then let's move that out of this PR (and to its own PR).

xoppa:
It looks like you added a `Scene` class but are never actually using it other than in your test and by getting a single member (in which case you pass in the member directly). If so, then let's remove that from this PR.

Btw, I'm not sure how you intended the `Scene` to be used, but it is very unlikely for a scene to only have one `Environment` (anything bigger than a test scene).

xoppa:
The `EnvironmentListener` is only used when the lights are added/removed using the old helper methods. It will not be triggered when the user adds or removes a light directly to/from the attribute.

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/allocation/ShadowMapAllocator.java
```diff
@@ -0,0 +1,44 @@
...
+package com.badlogic.gdx.graphics.g3d.shadow.allocation;
+
+import com.badlogic.gdx.graphics.Camera;
+
+/** Shadow map allocator get the good texture region for each light
+ * @author realitix */
+public interface ShadowMapAllocator {
+
+	/** Result of the allocator analyze */
+	public class AllocatorResult {
```
xoppa: `ShadowMapRegion` is probably a better naming than `AllocatorResult`

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/directional/FrustumDirectionalAnalyzer.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+import com.badlogic.gdx.math.Frustum;
+import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.math.Vector3;
+
+/** @FIXME NOT WORKING, DO NOT USE
+ * @author realitix
+ * @see "http://gamedev.stackexchange.com/questions/81734/how-to-calculate-directional-light-frustum-from-camera-frustum" */
+public class FrustumDirectionalAnalyzer implements DirectionalAnalyzer {
```
xoppa: This class is never used it seems, can this be removed?

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/directional/DirectionalResult.java
```diff
@@ -0,0 +1,42 @@
...
+ ******************************************************************************/
+
+package com.badlogic.gdx.graphics.g3d.shadow.directional;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.math.Vector3;
+
+/** Contains the result of the directional analyzer
+ * @author result */
+public class DirectionalResult {
```
xoppa: This class looks unnecessary. Adding a camera `out` argument to the `analyze` method looks like it would be better suited.

realitix:
Hi Xoppa! Lot of things in one shot! I will do everything tomorrow.

xoppa:
Thank you! Take your time, it's a tough API you sink your teeth in. Perhaps we should (not as part of this PR btw) add some sort of basic `ShaderBuilder` class which, for now, could be as simple as regex replacing `#include` directives. So we can move common glsl code to separate files. That might also be useful for fixing the [gles3 shader issue](https://github.com/libgdx/libgdx/issues/2087).

nooone:
Why did you add some of these unused classes in the first place? Did you not plan to use them at all?

realitix:
In the order:
- Camerahelper: Indeed I can use the frustum, it would be simpler and faster. I will add it in another PR.

- I think a Scene class is useful and like you can see I use sometimes only one member but another custom implementation (of the interfaces) can use others members, we can't know what the developer will do with it. Moreover, it contains an array of model instances which is a generic way to own model instances instead of having 4 overloads functions (like in modelbatch). How would you see the scene ?

- For the EnvironmentListener, I am not confident about the usefulness, what do you think about ?

- The FrustumDirectionalAnalyzer.java is never used because I didn't manage to implement the algorithme (the link in the file header). It's the best method, can you try ?

- I added some class I don't use (like AllLightFilter.java) but a developer can needs them. Maybe it's not a good idea.

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/directional/DirectionalResult.java
realitix: Ok

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/directional/FrustumDirectionalAnalyzer.java
realitix: See the answer below

gdx/src/com/badlogic/gdx/graphics/g3d/shadow/allocation/ShadowMapAllocator.java
realitix: Ok

nooone:
> I added some class I don't use (like AllLightFilter.java) but a developer can needs them. Maybe it's not a good idea.

I think it's a good idea to have some default implementations that just work out of the box in any case, but might not be optimized. It helps to quickly get started, see some results and it can be optimized with a specialized implementation later on.

realitix:
I removed the scene. You can continue reviewing.

realitix:
I'm going to remove point light shadowing with realistic shadow system because the glsl varying drawback is huge.

nstolpe:
I've been using realitix's shadow system branch with my project after working on my own shadow setup for a while. I got it to work w/ the RealisticShaderSystem pretty easily and am going to start building some systems of my own. This will be a pretty great addition to the framework.

realitix:
I've talked with Xoppa and this pull request is too big to be reviewed. Moreover, I should put it in the tests before integrate it directly in the core. I will let open this pull request to keep a view on the whole system.
Here the splitting I will do:
1 - Add interfaces
2 - Add implementations for nearFar, allocation, directional, filter
3 - Add BaseShadowSystem
4 - Add RealisticShadowSystem
5 - Add ClassicalShadowSystem
6 - Add tests

badlogic:
Closing this out, as you seem to have multiple other PRs going regarding the same thing. Will re-open if you feel the need for it, but you can still access it in a closed state and make comments.

PR #3352, opened 2015-08-14
Implement parsing methods for vectors and rectangles

antag99:
Implementation of parseXXX methods for `Vector2`, `Vector3` and `Rectangle`, as well as enhancements of `toString()` format. The format of `Vector2` is changed from `[x:y]` to `(x,y)`, `Vector3` is changed from `[x, y, z]` to `(x,y,z)`, and the format of `Rectangle` is changed from `x,y,w,h` to `[x,y,w,h]`. This was done for consistency and simplified parsing, and shouldn't break too much stuff.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -365,10 +365,10 @@ public class Rectangle implements Serializable, Shape2D {
 		return "[" + x + "," + y + "," + width + "," + height + "]";
 	}
 
-	/** Parses the {@code Rectangle} represented by the given string according to the format of {@link #toString()}.
+	/** Sets this {@code Rectangle} to the value represented by the specified string according to the format of {@link #toString()}.
 	 * @param v the string.
-	 * @return the {@code Rectangle} represented by the given string. */
-	public static Rectangle fromString (String v) {
+	 * @return this rectangle for chaining */
+	public Rectangle fromString (String v) {
 		int s0 = v.indexOf(',', 1);
 		int s1 = v.indexOf(',', s0 + 1);
 		int s2 = v.indexOf(',', s1 + 1);
@@ -378,7 +378,7 @@ public class Rectangle implements Serializable, Shape2D {
 				float y = Float.parseFloat(v.substring(s0 + 1, s1));
 				float width = Float.parseFloat(v.substring(s1 + 1, s2));
 				float height = Float.parseFloat(v.substring(s2 + 1, v.length() - 1));
-				return new Rectangle(x, y, width, height);
+				return this.set(x, y, width, height);
 			} catch (NumberFormatException ex) {
 				// Throw a GdxRuntimeException
 			}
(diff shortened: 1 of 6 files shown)


xoppa:
Thanks!

Why are the methods static? I think I'd expect a `T#fromString()` instead of `T.parseT()` as "opposite" to `T#toString()` (I guess providing an opposite of toString is what this is about).

What is the purpose of e.g. using '(' instead of '['? Which consistency and parsing is gaining from that?

What is the actual purpose of this parseT (or fromString) method? Are you trying to implement some kind of serialization here?

antag99:
> Why are the methods static? I think I'd expect a T#fromString() instead of T.parseT() as "opposite" to T#toString() (I guess providing an opposite of toString is what this is about).

Yeah, `T.fromString(String)` is a better idea.

> What is the purpose of e.g. using '(' instead of '['? Which consistency and parsing is gaining from that?

Not really related to parsing; I just thought it was a prettier format.

> What is the actual purpose of this parseT (or fromString) method? Are you trying to implement some kind of serialization here?

My use case is for reading positions from tiled map properties.

antag99:
Fixed the merge conflicts, any chance of getting this merged soon? Btw, who is responsible for updating `CHANGES`?

nooone:
Usually the person who changes something is responsible. That's you in this case.

MobiDevelop:
I'm not sure adding some `fromString` methods for a few classes is really noteworthy enough for the changes file, but I guess it doesn't hurt.

I'd like to see the docs for `fromString` specify the actual valid format, rather than saying "according to the format of toString".You shouldn't have to look at the source of `toString` to find out the expected format.

MobiDevelop:
@Xoppa, out of curiosity... what was wrong with a static `fromString`? It is sort of weird to do `new Vector2().fromString(s)`, no?

antag99:
@MobiDevelop I think it made more sense considering pooled/temporary instances; might not be a good idea to create a new Vector2 for every call to `fromString()`.

MobiDevelop:
I don't know, if you are setting things like Vector2, Vector3 from strings you are already doing it wrong (if performance matters to you).

antag99:
True, made the methods static again.

xoppa:
I don't have a strong opinion against the method being static, I just expected it to be in line with the opposite method. But when thinking about it, a few reasons come to mind:

Leave creating the object to the caller, we could also do this by adding an (optional) `out` parameter as we do in some other methods. Although in this case I think that using member (not static) method is cleaner than an `out` argument. This would also remove the need for temp objects.

I don't have a specific use-case, but I can imagine that you don't necessarily want to parse a vector, but a property of a bigger object which happens to be a vector. E.g. `camera.direction`, `node.translation`, `gameObject.position`, etc. These vectors are typically final, so it would be creating a needless object that is only used as argument to the `set` method. `cam.position.set(Vector3.fromString(s));` vs `cam.position.fromString(s);`

It allows for a more generic approach. E.g. it could be put in the [`Vector`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Vector.java) interface. The caller then doesn't have to know which parse implementation it needs.

antag99:
OK, let's leave the methods non-static... it's pretty redundant though. Users can define static methods in a helper class if necessary.

antag99:
Removed the changes to CHANGES, as it's causing merge conflicts.

xoppa:
Thanks!

PR #3353, opened 2015-08-14
Added VisRuntime extension

kotcrab:
Added VisRuntime library to 3rd party extensions. This library is required to load projects exported from [VisEditor](https://github.com/kotcrab/VisEditor).


saltares:
Thanks!

PR #3354, opened 2015-08-15
Added 2 methods and a line of class doc

nrallakis:
Added addItem and removeItem methods and wrote a line on the List class documentation

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
@@ -235,7 +235,7 @@ public class List<T> extends Widget implements Cullable {
 		if (oldPrefWidth != getPrefWidth() || oldPrefHeight != getPrefHeight()) invalidateHierarchy();
 	}
 	
-	public void addItem (T... newItem) {
+	public void addItem (T newItem) {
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
```diff
@@ -231,6 +234,28 @@ public void setItems (Array newItems) {
...
+		float oldPrefWidth = getPrefWidth(), oldPrefHeight = getPrefHeight();
+
+		items.add(newItem);
+		selection.validate();
+
+		invalidate();
+		if (oldPrefWidth != getPrefWidth() || oldPrefHeight != getPrefHeight()) invalidateHierarchy();
+	}
+	
+	public void removeItem (T... item) {
```
nooone: I'd call it `removeItems` (plural) instead. Same with `addItems`

nooone:
Both `items.add(newItem)` as well as `items.remove(item, false)` don't compile. Did you actually even try this yourself?

deathbeam:
Probably change it to items.addAll and removeAll should compile?

nrallakis:
I had tried it myself... i just forgot to remove the dots, thats why it doesn't compile my bad!

nrallakis:
Now it compiles

PR #3361, opened 2015-08-19
Gyroscope Implementation for android

snovak:
This works for me on android.  It’s a copy over of the accelerometer implementation… just wiring up another sensor for use in libGDX!

I haven’t written any test, nor implemented for iOS.  My test is an app I’m working on.. Will go to production in Sept to check out.


.gitignore
```diff
@@ -57,3 +57,5 @@ libgdx-*.zip.MD5
 
 #ensure gdx-setup.jar works properly
 !/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gwt/war
+
+libgdx-1.5.7-SNAPSHOT
```
MobiDevelop: I don't think we'd want to call out specific version numbers.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
```diff
@@ -612,6 +630,8 @@ public boolean isButtonPressed (int button) {
 	final float[] R = new float[9];
 	final float[] orientation = new float[3];
 
+	
+//TODO: ?
```
Tom-Ski: Whats todo?

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
```diff
@@ -661,20 +681,35 @@ void registerSensorListeners () {
...
 				Sensor accelerometer = manager.getSensorList(Sensor.TYPE_ACCELEROMETER).get(0);
-				accelerometerListener = new SensorListener(this.nativeOrientation, this.accelerometerValues, this.magneticFieldValues);
+				accelerometerListener = new SensorListener(this.nativeOrientation, this.accelerometerValues, this.magneticFieldValues, this.gyroscopeValues);
 				accelerometerAvailable = manager.registerListener(accelerometerListener, accelerometer,
 					SensorManager.SENSOR_DELAY_GAME);
 			}
 		} else
 			accelerometerAvailable = false;
+		
+//TODO: implement Gyroscope
```
Tom-Ski: Same as above

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
snovak: Sorry about that, the pull request may have been a bit hastey.  I'm finding a few things myself.

MobiDevelop: This doesn't seem like a useful TODO.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
```diff
@@ -37,6 +37,9 @@
 
 	/** whether to use the accelerometer. default: true **/
 	public boolean useAccelerometer = true;
+	
+	/** whether to use the gyroscope. default: true **/
+	public boolean useGyroscope = true;
```
MobiDevelop: I don't think we need both Accelerometer and Gyroscope on by default. Gyroscope uses considerably more battery than accelerometer, so we probably want it off by default.

snovak:
Appologies,, my first pull request ever.  My day job lives in SVN.  So, I'm learning.  I like the review layout.  Easier to see my mistakes ;-).  I'll get these fixed pronto.. I'm also noticing something that I'll need to recompile and test on the device before committing.  I'll change the comments and have the rest done today.

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -570,6 +570,15 @@ private static void initializeKeyNames () {
 	/** @return The value of the accelerometer on its y-axis. ranges between [-10,10]. */
 	public float getAccelerometerZ ();
 
+	/** @return The value of the gyroscope on its x-axis. ranges between [-10,10]??? */
```
MobiDevelop: If you don't know the range, it is probably best to leave it out.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
snovak: I'm finding issue with this.  Unfortunately, I think it's a deeper issue.  I set useGyroscope = false in GDX source, exported GDX to my test environment, and my test stopped working.  Regardless if I enabled it in my AndroidLauncer.java in onCreate()
```java
		AndroidApplicationConfiguration config;
		
		config = new AndroidApplicationConfiguration();
		config.useGyroscope = true; 
		config.useAccelerometer = true;
		config.useCompass = true;
		initialize(app, config);
		
		super.onCreate(savedInstanceState);
```
with this configuration, Gdx.input.isPeripheralAvailable(Peripheral.Gyroscope) always returns false, if it's set to false in GDX source.

to determine if there is something I was missing in my implementation, I set useAccelerometer and useCompass GDX defaults to false.  With the same (above) config, now Gdx.input.isPeripheralAvailable(Peripheral.Accelerometer) also never comes back true.  

In addition, I did the reverse..... set GDX defaults to true across the board, and my app config all to false.  Both
Gdx.input.isPeripheralAvailable(Peripheral.Accelerometer) and Gdx.input.isPeripheralAvailable(Peripheral.Gyroscope) return true, disregarding the config parameters at initialize().  

Looking.....

snovak: Confirmed,, I'm an idiot. :-)

snovak:
Okay.. some changes were made.  Looking much cleaner.  Ready for review.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/input/MockInput.java
```diff
@@ -41,6 +41,24 @@ public float getAccelerometerY() {
...
+
+	@Override
+	public float getGyroscopeY() {
+		// TODO Auto-generated method stub
+		return 0;
+	}
+
+	@Override
+	public float getGyroscopeZ() {
+		// TODO Auto-generated method stub
```
nooone: I think these TODOs could be removed. It's a mock and it doesn't have to be implemented.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglAWTInput.java
```diff
@@ -828,4 +828,22 @@ public void getRotationMatrix (float[] matrix) {
...
+
+	@Override
+	public float getGyroscopeY () {
+		// TODO Auto-generated method stub
+		return 0;
+	}
+
+	@Override
+	public float getGyroscopeZ () {
+		// TODO Auto-generated method stub
```
nooone: IMHO those TODOs are also not necessary. `isGyroscopeAvailable` returns false, so these won't need to be implemented as well.

nooone:
Dumb question: Why are `isAccelerometerAvailable` etc not part of the `Input` interface? Only `isPeripheralAvailable` is there. Some implementations have these methods, but I'd have to downcast to actually use them. Shouldn't there be a generic, platform-independent way to check this?

Tom-Ski:
Whats wrong with checking isPeripheralAvailable?

nooone:
Nothing, it's a good thing to do. And so is checking isAccelerometerAvailable.

nooone:
Oops, I didn't get what this method actually does. All good, gonna delete my comments :)

Tom-Ski:
I don't follow, isPeripheralAvailable(Peripheral.Accelerometer) will actually return that.

nooone:
Sorry @Tom-Ski for my confusion. It actually still leaves me in doubt about the `isAccelerometerAvailable` (and similar) methods in some of the implementations. Why do we have them, if `isPeripheralAvailable` actually checks that?

Tom-Ski:
Looks like its just LwjglInput that does this? If so its probably just old code that needs to removed.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/input/MockInput.java
snovak: I'm happy to take them out.  While this implementation concentrates on android, I'm not familiar enough with other features/platforms to properly implement them..   Like MockInput.  I just don't know what it does.

snovak:
Those additional comments have been removed.  Thanks for the guidance!

@nooone, are you saying that I remove `isGyroscopeAvailable()` and `isAccelerometerAvailable()`?  Would that be out of scope for this?

Tom-Ski:
You can leave it for now.

snovak:
I can add a `TODO` ;-)

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -570,6 +570,12 @@ private static void initializeKeyNames () {
 	/** @return The value of the accelerometer on its y-axis. ranges between [-10,10]. */
 	public float getAccelerometerZ ();
 
+	public float getGyroscopeX ();
```
Tom-Ski: Should have some javadoc as to what is expected to return from these functions, namely the un-normalized gyro sample data.

"return The rate of rotation around the x axis. (rad/s)" should be good enough.

Tom-Ski:
Looks good, if you could just add javadoc to the Input interface methods, and squash these commits to one, that would be great!

cypherdare:
Might want to check the iOS API and the frame of reference it uses to make sure this can be future proofed for the eventual iOS implementation. I suppose the iOSInput class could reorient to the Android frame of reference so it doesn't have to be done in the core of a project.

Also, it might be nice to also add support for the Rotation Vector Sensor, which probably provides better results than pure gyroscope. It melds gyroscope with the accelerometer, tilt sensor and compass and smooths the results all behind the scenes. iOS has something similar in Core Motion CMAttitude. But that could be done in a separate pull.

snovak:
@cypherdare I've created a new pull request for this here https://github.com/libgdx/libgdx/pull/3594 I will certainly take a look at the Rotation Vector Sensor for a future implementation.

PR #3363, opened 2015-08-20
Add PrefixFileHandleResolver

xoppa:
Add a simple `FileHandleResolver` that allows you to specify a sub-folder to use within the base.


gdx/src/com/badlogic/gdx/assets/loaders/resolvers/PrefixFileHandleResolver.java
```diff
@@ -0,0 +1,39 @@
+
```
intrigus: You forgot to include the license header.

PR #3365, opened 2015-08-22
World::destroyBody returns fixtures to pool

### LINKED ISSUE #3364: box2d World: Failure to use Pool generates Fixture garbage
While a fixturePool exists in World.java, it isn't being used.  Namely, you obtain() objects from the fixturePool forever, and never free any of them after the parent body is destroyed.  While Body has a reset() method that purports to return the freed fixtures to the pool, it doesn't work correctly because 1) Body.reset isn't called until someone tries to reobtain the freed Body and 2) all the fixtures have been removed from the Body at that point, so an empty list of Fixtures is freed.

I have a fix for this.

advayDev1:
Without this the freeFixtures pool is never actually used and new fixtures are always allocated.
When you have many bullets/projectiles/particles getting created/destroyed this causes a lot of GC churn.

Fixes #3364

NOTE: I am placing my code modification in this PR into the Public Domain.  It is small and I don't want to sign the CLA.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
@@ -325,8 +325,11 @@ b2ContactFilter defaultFilter;
 		this.bodies.remove(body.addr);
 		Array<Fixture> fixtureList = body.getFixtureList();
 		while(fixtureList.size > 0) {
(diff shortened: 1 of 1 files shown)


Tom-Ski:
It's probably better to use the `Body#destroyFixture` method? @MobiDevelop any idea why `Body#destroyFixture` isn't used currently?

MobiDevelop:
I assume that `destroyFixture` isn't used because the native fixtures don't need to be destroyed explicitly as destroying the body would have already done so.

PR #3371, opened 2015-08-24
Scene2d TextField Added an optional constructor

nrallakis:
Constructs a textfield with an empty string


NathanSweet:
We try to reduce the number of constructors for scene2d classes. If we wanted constructors which don't take a string we'd need 3 (doubling the number). Just pass "".

PR #3381, opened 2015-08-28
Added APK Expansions support

sarkanyi:
Adds APK expansion support. This is especially important for Audio, because otherwise lots of hacks have to be made to play audio.
The ZipFileHandleResolver can be used with the AssetManager to automatically load assets from the OBB instead of the local storage (AssetManager.setLoader() ha to be used).
The changes wouldn't break anything and add some extra convenience for those who want to release on the Play Store and have lots of assets (both audio and graphics).

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -27,7 +27,7 @@ public class Color {
 	public static final Color BLACK = new Color(0, 0, 0, 1);
 
 	public static final Color WHITE = new Color(0xffffffff);
-	public static final Color LIGHT_GRAY = new Color(0x696969ff);
+	public static final Color LIGHT_GRAY = new Color(0xbfbfbfff);
 	public static final Color GRAY = new Color(0x7f7f7fff);
 	public static final Color DARK_GRAY = new Color(0x3f3f3fff);
 	public static final Color SLATE = new Color(0x708090ff);


Tom-Ski:
You're missing a few imports in your second commit.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
```diff
@@ -106,7 +106,11 @@ public Music newMusic (FileHandle file) {
 
 		if (aHandle.type() == FileType.Internal) {
 			try {
-				AssetFileDescriptor descriptor = aHandle.assets.openFd(aHandle.path());
+				AssetFileDescriptor descriptor;
+				if(aHandle.assets != null)
```
Tom-Ski: Please use the LibGDX formatter provided.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
```diff
@@ -0,0 +1,109 @@
...
+				return getInstance().extensions.contains(ext.toLowerCase(Locale.ENGLISH));
+			}
+			return false;
+		}
+	}
+
+	private AssetFileDescriptor assetFd;
+
+	public AndroidZipFileHandle(String fileName) {
+		super(null, fileName, AudioExtensions.isAudio(fileName) ? FileType.Internal : FileType.Absolute);
```
Tom-Ski: Why Internal if audio?

sarkanyi: Because it works that way :). I don't know exactly why, but it's tested and works. I missed the imports because I was adapting the stuff in my game and testing the changes, but then I moved only the new class/function.
If I think about it long and hard, I think it's actually a "hack", because this way the AndroidFileHandle.assets will be null for audio if it's not internal, enabling it to enter the second codepath which returns the actual AssetFileDescriptor. Android's AssetManager can't open it with openFd() to get the AssetFileDescriptor. That's certain.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInfo.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.backends.android;
+
+/** @author sarkanyi */
+public class AndroidInfo {
+	public static ZipResourceFile expansionFile;
```
MobiDevelop: I'd prefer AndroidFiles to hold the reference to the expansion file, and pass it to the AndroidZipFileHandle constructor. If you need to add a configuration argument or method to set the expansion file on AndroidFiles, that's probably better than holding it in a public static.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.badlogic.gdx.Files.FileType;
+import com.badlogic.gdx.files.FileHandle;
+
+/** @author sarkanyi */
+public class AndroidZipFileHandle extends AndroidFileHandle {
+	private AssetFileDescriptor assetFd;
+
+	public AndroidZipFileHandle(String fileName) {
+		super(null, fileName, FileType.Absolute);
```
MobiDevelop: I'd think FileType.Classpath is probably more appropriate than Absolute.

MobiDevelop: Or, possibly Internal.

MobiDevelop:
Generally, speaking I am OK with the idea of expansion file support, but I feel like the backend should handle the implementation transparently. Functionally, I'd expect the expansion files to simply extend the "internal" type to look in the expansion file if the file isn't found in the APK.

sarkanyi:
Ok, listened to the advices and rewrote the thing. Now APK expansion handling is transparent, only AndroidFiles.setAPKExpansion(mainVersion, patchVersion) has to be called by the user, the rest is automatic. If an asset isn't found in the Internal assets, then the expansion file is tried and if it's there, it'll be used.

sarkanyi:
This is replaced by https://github.com/libgdx/libgdx/pull/3408

Issue #3345, opened 2015-08-12
Make G3dModelLoader customizable


VinceAngel:
Hi,

Here is my issue : I'm customizing the json content for models, but the G3dModelLoader cannot be customized without copy/pasting the whole class.
I'm thinking about a PR to set all the parseMesh/Material/etc... methods public or protected to let them being overriden if necessary.

What do you think about this ?
Regards,
    Vincent.

FyiurAmron:
@VinceAngel personally, I'd say "give it a go", since that's the exact issue (non-extensibility) I've already voiced a couple of times regarding various classes. I have a feeling some devs gone with Josh Bloch's saying "either prepare class & document it for extensibility or explicitly prevent it" - still, IMO that just means a minor refactoring and/or doc preparation should be in place, not that there shouldn't be a way to extend library classes.

nooone:
Yeah, it would be nice if all package-private/default scoped fields in libgdx would be protected... not being able to extend some classes without making a PR beforehand was something I've encountered a few times before.

xoppa:
Is this a duplicate of #3186

FyiurAmron:
@xoppa with the bits OP said about this I wouldn't call it a dupe. He explicitly mentioned the JSON content, so I guess he's referring to e.g. https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/loader/G3dModelLoader.java#L81 , https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/loader/G3dModelLoader.java#L143 etc.

VinceAngel:
Not really a duplicate, but both can be done in the same time to make loader and model "builder" able to be customized.

badlogic:
Since we got no PR for this, and since the JSON format is what gets output by fbxconv which nobody except Xoppa is working on, i'm closing this.

Issue #3186, opened 2015-06-02
(OOP) modify Model class to allow subclasses with custom loading sequence calling existing loading methods


FyiurAmron:
To implement a different `load()` behaviour, one currently has to Ctrl+C/Ctrl+V the entire related code from Model, even if it is necessary to only change a slightest bit.

The proposed solution would be to change all related methods () from `private` to `final protected`, allowing easy modular changes and/or replacements via subclasses with constructors calling no-arg `Model()` c-tor with custom `load()` variant.

As a side note, `nodePartBones` can be made either final and accessible for subclasses via `final protected`, to allow using it in custom subclass loader, or (possibly better, since AFAIK it's only used during loading and just hogging the memory for the Model later on) converted from field to `loadNodes()` method variable, passed to `loadNode()` as a parameter.

If this's somewhat too big for an instant dev patch, I'd be more than happy to do a PR for it.

xoppa:
Yeah, need to refactor that part a bit. If I find the time, I will have a look at it later this week. Thanks for reporting.

VinceAngel:
Maybe https://github.com/libgdx/libgdx/pull/3705 have to be referenced too

PR #3390, opened 2015-09-03
Exposes SpriteBatch members as protected

realitix:
Allows the sprite batch to be override by converting all "private" to "protected".
I wanted to add RenderContext to the sprite batch but I couldn't due to all theses private properties.
So I had to copy/paste all the sprite batch. This is not a good solution.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -1085,6 +1085,7 @@ public class SpriteBatch implements Batch {
 		return !blendingDisabled;
 	}
 
+	@Override
 	public boolean isDrawing () {
 		return drawing;
 	}
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
@@ -34,29 +34,29 @@ import com.badlogic.gdx.utils.NumberUtils;
  * @author mzechner
  * @author Nathan Sweet */
 public class SpriteBatch implements Batch {
-	private Mesh mesh;
+	protected Mesh mesh;
 
-	final float[] vertices;
-	int idx = 0;
-	Texture lastTexture = null;
-	float invTexWidth = 0, invTexHeight = 0;
+	protected final float[] vertices;
+	protected int idx = 0;
+	protected Texture lastTexture = null;
+	protected float invTexWidth = 0, invTexHeight = 0;
 
-	boolean drawing = false;
+	protected boolean drawing = false;
 
-	private final Matrix4 transformMatrix = new Matrix4();
-	private final Matrix4 projectionMatrix = new Matrix4();
-	private final Matrix4 combinedMatrix = new Matrix4();
+	protected final Matrix4 transformMatrix = new Matrix4();
+	protected final Matrix4 projectionMatrix = new Matrix4();
+	protected final Matrix4 combinedMatrix = new Matrix4();
 
-	private boolean blendingDisabled = false;
-	private int blendSrcFunc = GL20.GL_SRC_ALPHA;
-	private int blendDstFunc = GL20.GL_ONE_MINUS_SRC_ALPHA;
+	protected boolean blendingDisabled = false;
+	protected int blendSrcFunc = GL20.GL_SRC_ALPHA;
+	protected int blendDstFunc = GL20.GL_ONE_MINUS_SRC_ALPHA;
 
-	private final ShaderProgram shader;
-	private ShaderProgram customShader = null;
-	private boolean ownsShader;
+	protected final ShaderProgram shader;
+	protected ShaderProgram customShader = null;
+	protected boolean ownsShader;
 
-	float color = Color.WHITE.toFloatBits();
-	private Color tempColor = new Color(1, 1, 1, 1);
+	protected float color = Color.WHITE.toFloatBits();
+	protected Color tempColor = new Color(1, 1, 1, 1);
 
 	/** Number of render calls since the last {@link #begin()}. **/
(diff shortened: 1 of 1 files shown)


saltares:
Making everything `protected` is not a good solution either. I don't think this can be merged as it exposes the entire internal representation of `SpriteBatch`.

Will let other people decide as to if there is a real need for making `SpriteBatch` more extensible.

realitix:
Ok I wait for the others but the actual SpriteBatch can't be easily overriden.

MobiDevelop:
I agree with @saltares on this one.

realitix:
How do you customize SpriteBatch while most of properties are private ?

realitix:
Some others complain too about the private properties:
https://github.com/libgdx/libgdx/issues/3345
https://github.com/libgdx/libgdx/issues/3186

Darkyenus:
Since most of the SpriteBatch usages actually expect interface Batch, copying SpriteBatch and customizing it there should be easy enough. It will probably be cleaner and faster for most customizations too.

MobiDevelop:
You should create your own implementation of the Batch interface. If that means copying the SpriteBatch and modifying it, so be it... At least it is open source.

realitix:
Sure but copying more than 1000 lines to change 10 lines...
Ok I close this pull request, too bad for developers.

MobiDevelop:
I'd be willing to bet there is a better way to accomplish what you are trying to accomplish that doesn't involve making the entire implementation of SpriteBatch a public api, or copying the SpriteBatch source.

realitix:
I am not sure how.

What I need:
```
public void flush () {
		if (idx == 0) return;

		renderCalls++;
		totalRenderCalls++;
		int spritesInBatch = idx / 20;
		if (spritesInBatch > maxSpritesInBatch) maxSpritesInBatch = spritesInBatch;
		int count = spritesInBatch * 6;

		lastTexture.bind();
		Mesh mesh = this.mesh;
		mesh.setVertices(vertices, 0, idx);
		mesh.getIndicesBuffer().position(0);
		mesh.getIndicesBuffer().limit(count);

		if (blendingDisabled) {
			Gdx.gl.glDisable(GL20.GL_BLEND);
		} else {
			Gdx.gl.glEnable(GL20.GL_BLEND);
			if (blendSrcFunc != -1) Gdx.gl.glBlendFunc(blendSrcFunc, blendDstFunc);
		}

		mesh.render(customShader != null ? customShader : shader, GL20.GL_TRIANGLES, 0, count);

		idx = 0;
	}
```

becomes

```
public void flush () {
		if (idx == 0) return;

		renderCalls++;
		totalRenderCalls++;
		int spritesInBatch = idx / 20;
		if (spritesInBatch > maxSpritesInBatch) maxSpritesInBatch = spritesInBatch;
		int count = spritesInBatch * 6;

		Texture t = lastTexture;
		tempDesc.set(t, t.getMinFilter(), t.getMagFilter(), t.getUWrap(), t.getVWrap());
		shader.setUniformi("u_texture", context.textureBinder.bind(tempDesc));

		Mesh mesh = this.mesh;
		mesh.setVertices(vertices, 0, idx);
		mesh.getIndicesBuffer().position(0);
		mesh.getIndicesBuffer().limit(count);

		if (blendingDisabled) {
			Gdx.gl.glDisable(GL20.GL_BLEND);
		} else {
			Gdx.gl.glEnable(GL20.GL_BLEND);
			if (blendSrcFunc != -1) Gdx.gl.glBlendFunc(blendSrcFunc, blendDstFunc);
		}

		mesh.render(customShader != null ? customShader : shader, GL20.GL_TRIANGLES, 0, count);

		idx = 0;
	}
```

The texture is binded with a RenderContext. I need to copy/paste all the function's content because it's in the middle of "flush". And lot of properties used in "flush" are private and can't be accessed in an other way.
Do you see another solution ?

xoppa:
TextureBinder doesn't use unit 0 by default, so it doesn't interferes with SpriteBatch. There's no need to modify SpriteBatch for that. See: https://github.com/libgdx/libgdx/wiki/ModelBatch#texturebinder.

SpriteBatch, by design, doesn't keep track of the opengl state. Using RenderContext for *only* the texture binds doesn't "fix" that. You'd need to use it for practically all gl calls (like e.g. the blend function in your snippet). 

SpriteBatch does use its own strategy to prevent duplicate texture binds. (although, for performance, the texture bind is not what's really important here, it is the draw call that makes the actual difference).

Typically you don't switch alot between ModelBatch and SpriteBatch in a single frame (in most cases only once). So for performance, using TextureBinder for SpriteBatch is probably not going to gain anything noticeable. Perhaps you have another reason to do this?

tl;dr: This might be a premature optimization. If you are trying to optimize here then make sure to profile, bench mark and compare.

realitix:
Indeed Xoppa, I did something special, I created a TextureBinder that binds only one time each texture. So after the second frame, all textures are binded and no more bind after that.
This increase performance but the drawback is that all game textures must be packed in Atlas.
It allows only 15 textures so I have to use  the unit 0 too (15 is not a lot!).
With the actual SpriteBatch, it would erase the unit 0 and all next renders using unit 0 (in ModelBatch loop) would use the wrong texture.
I hope I'm clear.
Like you say, maybe it's a premature optimization but I heard that binding texture can be expensive so I never bind textures after the first frame.
What do you think about that ?

xoppa:
Binding a texture on it self is not an expensive operation. Unless you are actually experiencing issues, profiled it and found the actual binding to be the culprit (which is very unlikely), you are probably focusing on the wrong things.

As general rule of thumb: never assume, always verify! Optimizing only makes sense if you can verify that it actually gains performance. Going through all those loops, because you *heard* that it *can* be expensive, doesn't sound like a good idea.

Btw, let's not deviate too much on the issue tracker. General discussions about performance and such are probably better suited for the forum or irc.

realitix:
I understand Xoppa but that's not resolve the current problem, LibGDX is hard to customize.
In my opinion, the main concern of a framework like LibGDX (video game with 1000000 possibilities, not a black box) would be the extendability. I think it's a java habit to not expose internal representation.

Like creators of python say, we are all adults:
http://docs.python-guide.org/en/latest/writing/style/#we-are-all-responsible-users

Like I read on the LibGDX forum, some classes have public field and it's a bad Java practice but it's justified with LibGDX because there is more important than design pattern or good practice in video games world...

In my opinion, all libGDX classes should avoid private or package-private and everything should be protected or public.

What do you think about it ?

saltares:
Anything that is not private becomes part of the public API and would force the team to support it **forever**. Surely we can break the API but Libgdx has **a lot** of users and they would, justifiably, become annoyed users if we kept breaking the API.

Yeah according to Python we're all adults, but are we really? Libgdx is use by both expert and novice developers.

Even if everyone knew what they were doing, in big projects it's very easy to break stuff. If we exposed API internals, it would become much easier to blow stuff up. We do not want that. Robust APIs are good APIs.

Not exposing the internal representation is not just a Java thing, it's is an anti-pattern in OOP generally speaking. It also applies to other languages. Python is not OO.

I'm pretty sure I can speak for all core devs, we will not change paradigm and start exposing everything. I think most Libgdx users would agree with us as well.

realitix:
Ok

PR #3391, opened 2015-09-03
Color.valueof(String s)

alexcrist:
I was using libgdx's Color class, specifically to create a color from a hex string. In Java's standard Color class, you are able to enter hex strings that start with a `#`.

I think myself and many others may have been briefly confused because `Color.valueOf(String s)` does not accept strings that start with a `#`.

I believe the following code will fix this. Thanks for the great library! :-)


alexcrist:
Hmm, any chance somebody could explain to me why the PR Check failed? Does this mean the code doesn't work?

Tom-Ski:
`hex.charAt(0)` returns a primitive char.

PR #3392, opened 2015-09-03
Color.valueOf(String s)

alexcrist:
I modified Color.valueOf(String s) so it can now also accept hex strings that start with a '#'

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -331,6 +331,7 @@ public class Color {
 	/** Returns a new color from a hex string with the format RRGGBBAA.
 	 * @see #toString() */
 	public static Color valueOf (String hex) {
+		hex = hex.charAt(0) == '#' ? hex.substring(1) : hex;
 		int r = Integer.valueOf(hex.substring(0, 2), 16);
 		int g = Integer.valueOf(hex.substring(2, 4), 16);
 		int b = Integer.valueOf(hex.substring(4, 6), 16);


Gtoknu:
I'm not a huge fan of this solution. It recreates the string, when you could simply offset the other substrings by 1.

Like this: 

```java
		int offset = hex.charAt(0) == '#' ? 1 : 0;
		int r = Integer.valueOf(hex.substring(0+offset, 2), 16);
		int g = Integer.valueOf(hex.substring(2+offset, 4), 16);
		int b = Integer.valueOf(hex.substring(4+offset, 6), 16);
		int a = hex.length() != (8+offset) ? 255 : Integer.valueOf(hex.substring(6+offset, 8), 16);
```

This should be enough to read it without changing the input string!

MobiDevelop:
Note that would require to also apply the offset to the end index for the substring.

Gtoknu:
Oops, my bad! Thanks for pointing out :)

alexcrist:
@Gtoknu Nice solution, just curious, what's the disadvantage of recreating the string?

Darkyenus:
@alexcrist .substring creates a new string which has to be then freed by GC, creating too much objects can cause GC pauses, especially on mobile platforms

MobiDevelop:
I'm not sure this is a case where that really matters, given the 3 other substring calls in the method.

davebaol:
Yep, I have to agree with @MobiDevelop. 
If you want to avoid string allocations, which never hurts, you can do something like [this](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java#L316)

Gtoknu:
@davebaol that makes me wonder why it has been re-implemented on the `GlyphLayout` instead of improving the method on the `Color` class and using it.

davebaol:
Parsing color markup language is a bit different because colors are embedded in a larger text.
And the equivalent implementation of `Color.valueOf` would be simpler (no need to detect terminator char `]`). Something like this should be enough I guess (100% UNTESTED code)
````java
    public static Color valueOf(String hex) {
        int i = hex.charAt(0) == '#' ? 1 : 0;
        int n = hex.length();
        if (n != 8 + i)
            throw new IndexOutOfBoundsException("...");
        int colorInt = 0;
        while(i < n) {
            char ch = hex.charAt(i++);
            if (ch >= '0' && ch <= '9')
                colorInt = colorInt * 16 + (ch - '0');
            else if (ch >= 'a' && ch <= 'f')
                colorInt = colorInt * 16 + (ch - ('a' - 10));
            else if (ch >= 'A' && ch <= 'F')
                colorInt = colorInt * 16 + (ch - ('A' - 10));
            else
                throw new NumberFormatException("Illegal hex character");
        }
        return Color.rgba8888ToColor(colorInt);
    }
````

**Comment by SleeplessByte** [comment]
@davebaol you probably don't need to force RRGGBBAA, so `n - i == 6 || n - i == 8`.

Do you really need the sanity checks?

````Java
int colorInt = 0;
int size = (n - i)/2;
int[] rgba = new int[4];
rgba[3] = 0xff;
int a, b, x, y;

for(int j = 0; j < size; j++){
    a = hex.charAt(i + j*2);
    b = hex.charAt(i + j*2+1);
    x = a < 65 ? a - 48 : a - 55;
    y = b < 65 ? b - 48 : b - 55;
    rgba[j] = x * 16 + y;
    colorInt = colorInt * 16 + x;
    colorInt = colorInt * 16 + y;
}

// ret has your rgba components
// so you could separate these functions out, or not
return Color.rgba8888ToColor( colorInt );

**Comment by Gtoknu** [comment]
@SleeplessByte you're probabbly missing the point of the checks in loop. You have to make them anyway because you don't really know if the format is `RRGGBBAA` or `rrggbbaa` (upper or lower case). If you're doing them anyway, the last else case that throws an error comes practically free, since the comp has already been done anyway (someone correct me if I'm wrong).

If you'd make it different and remove the sanity check completely, you'd have to change the order of the checks (upper case is lower than lower case in the ascii table), so you could simply drop the upper bound checks:
```java
            if (ch >= '0')
                colorInt = colorInt * 16 + (ch - '0');
            else if (ch >= 'A')
                colorInt = colorInt * 16 + (ch - ('A' - 10));
            else if (ch >= 'a')
                colorInt = colorInt * 16 + (ch - ('a' - 10));
```
But I'm not sure we should drop security for 3 comps. They surely aren't as costy as getting substrings.

SleeplessByte:
True dat. I like any solution better than the substrings. I totally forgot
that people lowercase hexvalues.

xoppa:
If someone wants to contribute a better implementation then that would be great. Until then, I don't think there's a reason not to merge this (the extra substring is only for the '#' character. if you dont want that, then dont use it).

erromu:
The only thing I don't like about current logic is if you pass an invalid RRGGBB[AA] string it will throw an exception (e.g. a "fff" value).

Maybe it should return a default color as suggested above, so Color.valueOf could handle invalid strings instead of adding such logic everyplace where a call to Color.valueOf is done.

I do have a tiled map having many color attributes to control lights, so I need to care about such possible invalid strings params everywhere (often people editing a tile map color attribute does not care / know what a valid hex color string is)

What do you guys think?

Gtoknu:
I still prefer to get an exception than a default color. Is it that hard to:

´´´java
Color color;
try {
    color = Color.valueOf("#FFF");
} catch(Exception e) {
    //log it
    color = myDefaultColor;
}
```

?

On Sat, Oct 3, 2015 at 8:56 PM Erick J Romero <notifications@github.com>
wrote:

> The only thing I don't like about current logic is if you pass an invalid
> RRGGBB[AA] string it will throw an exception (e.g. a "fff" value).
>
> Maybe it should return a default color as suggested above, so
> Color.valueOf could handle invalid strings instead of adding such logic
> everyplace where a call to Color.valueOf is done.
>
> I do have a tiled map having many color attributes to control lights, so I
> need to care about such possible invalid strings params everywhere (often
> people editing a tile map color attribute does not care / know what a valid
> hex color string is)
>
> What do you guys think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/libgdx/libgdx/pull/3392#issuecomment-145299890>.
>

**Comment by erromu** [comment]
You are right.  In fact, that is what I am doing every where.
But it becomes a repetitive code so it seems logic to me to have such logic in one place only.

I could add a wrapper function that does such try - catch and get the desired result using one code line like this

```
	Color parseColorValueOf(String hex) {
		Color color;
		try {
			color = Color.valueOf(hex);
		} catch (Exception e) {
			// log it
			color = new Color(1,1,1,1);
		}
		return color;
	}
```

It could solve but adds a small overhead creating new variables for that.

So I did realize similar logic in libgdx already exists returning a default color - [above link](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java#L317-L341) - and it also does not overhead by calling substring when parsing the hex value.

I guess seems logic to be consistency  every where, don't you?

**Comment by davebaol** [comment]
@erromu 
o_O Where's the default color in the code you linked? 
That code just adds the parsed color to a stack and returns the length of the parsed string or  -1 in case of bad hex color.

**Comment by Darkyenus** [comment]
What about making the `valueOf(String)` a non-static method of `Color`,  which returns boolean: true - parsed successfully / false otherwise and the original color is kept? Thus solving the problem of needing defaults and at the same time being able to catch parse errors (without creating and throwing costly exceptions). Current static `valueOf` can be kept for backwards compatibility.

**Comment by erromu** [comment]
hello @davebaol 

Linked code reads 
```
// Parse hex color RRGGBBAA where AA is optional and defaults to 0xFF if less than 6 chars are used.
...
for (int ii = 0, nn = 9 - (i - start); ii < nn; ii++)
							colorInt = colorInt << 4;
						colorInt |= 0xff;
...
```
So seems it is setting default color when doing `colorInt |= 0xff;`
In fact I added such linked code from one of your comments above.

So please correct me if I am wrong but linked code is an example how to set default color and not how to return a default color.

Regarding tracking parsing errors, you are right, it could be not tracked at least you add a kind of logging error instead of throwing an exception.

In my case I don't need to get an exception but a default color instead (maybe including the logging error part)

If you guys need to track parsing errors through an exception then it is ok for me to use a wrapper function doing the try catch as pointed above.

Issue #2663, opened 2014-12-20
Nexus 7 Android 5.0.1 Music not playing


milos1290:
Just tested with latest nightlies and stable build music is not playing on 5.0.1 Nexus 7, i tested on Nexus 5 on 5.0 and it runs normally. Here is the log

```
E/MediaPlayer﹕ prepareAsync called in state 0
W/System.err﹕ java.lang.IllegalStateException
W/System.err﹕ at android.media.MediaPlayer._prepare(Native Method)
W/System.err﹕ at android.media.MediaPlayer.prepare(MediaPlayer.java:1125)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidMusic.play(AndroidMusic.java:85)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidMusic$1.run(AndroidMusic.java:177)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame(AndroidGraphics.java:409)
W/System.err﹕ at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1522)
W/System.err﹕ at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1239)

E/MediaPlayer﹕ Error (-38,0)
E/MediaPlayer﹕ Error (100,0)
W/AudioSystem﹕ AudioPolicyService server died!
W/System.err﹕ java.io.IOException: Prepare failed.: status=0x64
W/System.err﹕ at android.media.MediaPlayer._prepare(Native Method)
W/System.err﹕ at android.media.MediaPlayer.prepare(MediaPlayer.java:1125)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidMusic.play(AndroidMusic.java:85)
W/System.err﹕ at com.badlogic.gdx.scenes.scene2d.utils.ClickListener.touchUp(ClickListener.java:89)
W/System.err﹕ at com.badlogic.gdx.scenes.scene2d.InputListener.handle(InputListener.java:57)
W/System.err﹕ at com.badlogic.gdx.scenes.scene2d.Stage.touchUp(Stage.java:344)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidInput.processEvents(AndroidInput.java:360)
W/System.err﹕ at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame(AndroidGraphics.java:414)
[...]

PR #3399, opened 2015-09-06
Fix for Android music Nu Player issues

obigu:
Android Music has several issues on Android 5.0+ devices using the new Nu Player by default.

These are two of the issues specific to Nu Player this patch fixes

1- setPosition() not working after stop() is called

```
music.play();
// Wait
music.stop();
// Wait
music.play();
music.setPosition(30f);
[Music starts from the beginning]
```

2- Calling stop() before play() throws IllegalStateException()

```
music.stop();
music.play();
```

The problem seems to be related to the call to player.seekTo(0) on the AndroidMusic stop() method. This was added most probably as a workaround to MediaPlayer not restarting the music after a stop call in some cases (see http://stackoverflow.com/questions/5985874/android-mediaplayer-clipping-on-restart-after-seekto and http://stackoverflow.com/questions/6849045/mediaplayer-instance-stop-behaves-like-a-pause). Porbably what's happening is that because seekTo() is assynchronous (http://developer.android.com/reference/android/media/MediaPlayer.html) it's returning after stop() has been executed and that is messing with Nu Player state.

The fix has been tested on a Nexus 4 and a Nexus 5 for both default Nu Player and Awesome Player for backwards compatibility but it has not been tested on Froyo devices that are supposed to be buggy (does libGDX support that?).

Other issues have been reported that may be related like https://github.com/libgdx/libgdx/issues/2663 but I haven't tested.


radiocold:
Hi I same problem with the music , I used 1.9.3-SNAPSHOT but this fix don`t exist , Please any hellp?

PR #3401, opened 2015-09-06
Added support for name provider in SelectBox and List

kotcrab:
Current implementation of SelectBox and List is limited and always uses `toString()` method to show item as text. I added way to set custom name provider for SelectBox/List. This way it's possible to customize what text will be shown for given item (or just call different method than `toString()`). 
If provider is not set it defaults to `toString()` so it doesn't break old stuff.


NathanSweet:
This is a pretty heavy solution for such a simple problem. I'd prefer just have a protected SelectBox#toString(Object) that can be overridden to customize the display.

kotcrab:
That would only allow to customize text in selectbox itself, not in it's dropdown list. Notice that in `SelectBox#setItemNameProvider` I also must call `selectBoxList.list.setItemNameProvider`.

kotcrab:
Updated, if that change is OK tell me and I'll rebase.

NathanSweet:
Ah oops, merged before you rebased. Ah well.

Thanks!

PR #3404, opened 2015-09-07
Update extensions.xml

azakhary:
Version 0.1.1 supporting latest editor and latest libGDX released

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
@@ -2,10 +2,10 @@
 <extensions>
 	<extension>
 		<name>Overlap2D</name>
-		<description>Level and UI Editor Runtime</description>
+		<description>Runtime for Overlap2D ui and level editing toolkit</description>
 		<package>com.underwaterapps.overlap2druntime</package>
-		<version>0.1.0</version>
-		<compatibility>1.6.2</compatibility>
+		<version>0.1.1</version>
+		<compatibility>1.6.4</compatibility>
 		<website>http://overlap2d.com/</website>
 		<gwtInherits></gwtInherits>
 		<projects>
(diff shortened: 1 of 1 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
```diff
@@ -2,10 +2,10 @@
 <extensions>
 	<extension>
 		<name>Overlap2D</name>
-		<description>Level and UI Editor Runtime</description>
+		<description>Runtime for Overlap2D ui and level editing toolkit</description>
```
saltares: Lowercase for UI? Shouldn't it be uppercase?

extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
```diff
@@ -2,10 +2,10 @@
...
 		<package>com.underwaterapps.overlap2druntime</package>
-		<version>0.1.0</version>
-		<compatibility>1.6.2</compatibility>
+		<version>0.1.1</version>
+		<compatibility>1.6.4</compatibility>
```
saltares: Is it not compatible with 1.6.5?

azakhary:
Oh 1.6.5 is up, forgot that. Well, I'll make it compatible and re-push. stand by :)

azakhary:
I think it should be compatible with 1.6.5 because nothing almost changed from previous version. but it says in it's dependencies 1.6.4. Will this work ?

azakhary:
yep, checked it works fine with 1.6.5. But it says in it's gradle dependency 1.6.4 and I can't change that without building new version to maven. So this should do right?

azakhary:
Should I put 1.6.5. here in the extension config?

saltares:
If there are no API breaking changes for you, go ahead.

azakhary:
Done.

azakhary:
I checked, all works, no API changes that can affect anything

saltares:
Can you squash the commits please?

azakhary:
I can, but in this particular case it's so much easier to do new PR if thats ok?

PR #3406, opened 2015-09-07
Add Gwt support for custom Cursors

intrigus:
Added Gwt support for custom Cursors.

Cursors should not work in IE according to
http://reference.sitepoint.com/css/cursor because we convert the canvas
to a data-url of the png format and IE only supports .cur and .ani
files.
IE would also not support setting hotspot coordinates, it only supports
hotspot coords in .cur files.
Not all opera version seem to support cursors using the "cursor:url()"
syntax

Any suggestions?


Tom-Ski:
Looks good, might be worth mentioning in the documentation its limitations based on the availability of cursor:url and hotspot coords.

Other than that, could you just squash to 1 commit.

MobiDevelop:
Looks good, thanks.

PR #3408, opened 2015-09-10
Add APK expansion support

sarkanyi:
The change is transparent to the user, Gdx.files.internal() can be used to access assets inside expansion files.
The application must call ((AndroidFiles)Gdx.files).setAPKExpansion() first, to open the expansion file(s). The function returns true if it was opened successfully.
This replaces https://github.com/libgdx/libgdx/pull/3381

### DIFF
--- a/tests/gdx-tests-android/src/com/badlogic/gdx/tests/android/AndroidTestStarter.java
+++ b/tests/gdx-tests-android/src/com/badlogic/gdx/tests/android/AndroidTestStarter.java
@@ -16,7 +16,6 @@
 
 package com.badlogic.gdx.tests.android;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import android.app.ListActivity;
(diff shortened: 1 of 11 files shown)


Tom-Ski:
test this

Tom-Ski:
test this please

tests/gdx-tests-android/src/com/badlogic/gdx/tests/android/AndroidTestStarter.java
```diff
@@ -38,6 +37,8 @@
 	public void onCreate (Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		GdxTests.tests.add(MatrixTest.class);
+		if (!GdxTests.tests.contains(APKExpansionTest.class))
```
intrigus: Why exactly do you do this?
If you want to add a new test you should add it to list in com.badlogic.gdx.tests.utils.GdxTests.

Tom-Ski: Its an android specific test

sarkanyi: And the check is needed because I noticed that otherwise it can happen that it will be added twice to the list if you do a back and start GDX Tests again.

MobiDevelop:
Looks OK to me. @Tom-Ski should merge if he agrees.

sarkanyi:
It seems that if there are lots of entries in the obb file, exists() is rather slow. For directories I have no better idea, but for files I think it can be optimized to this:
   return isDirectory() ? expansionFile.getEntriesAt(getPath()).length != 0 : read() != null;
Can I somehow make this patch without a new pull request?

sarkanyi:
And one more because of a stupid NullPointerException: https://github.com/libgdx/libgdx/pull/3473

Issue #2807, opened 2015-02-04
AndroidInput.onKey exception: java.lang.ArrayIndexOutOfBoundsException


titovmaxim:
Get new crash in-the-wild. Several devices with Android 4.0.3 and 4.2.2. Reports are from the field (China) so no minimal example. Stacktrace:

<pre><code>java.lang.ArrayIndexOutOfBoundsException: length=256; index=1027
       at com.badlogic.gdx.backends.android.AndroidInput.onKey(AndroidInput.java:496)
       at android.view.View.dispatchKeyEvent(View.java:5495)
       at android.view.ViewGroup.dispatchKeyEvent(ViewGroup.java:1246)
       at android.view.ViewGroup.dispatchKeyEvent(ViewGroup.java:1246)
       at android.view.ViewGroup.dispatchKeyEvent(ViewGroup.java:1246)
       at com.android.internal.policy.impl.PhoneWindow$DecorView.superDispatchKeyEvent(PhoneWindow.java:1921)
       at com.android.internal.policy.impl.PhoneWindow.superDispatchKeyEvent(PhoneWindow.java:1363)
       at android.app.Activity.dispatchKeyEvent(Activity.java:2324)
       at com.android.internal.policy.impl.PhoneWindow$DecorView.dispatchKeyEvent(PhoneWindow.java:1848)
       at android.view.ViewRootImpl.deliverKeyEventPostIme(ViewRootImpl.java:3327)
       at android.view.ViewRootImpl.handleFinishedEvent(ViewRootImpl.java:3300)
       at android.view.ViewRootImpl.handleMessage(ViewRootImpl.java:2460)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:137)
       at android.app.ActivityThread.main(ActivityThread.java:4424)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:787)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:554)
       at dalvik.system.NativeStart.main(NativeStart.java)
</code></pre>

Tom-Ski:
1027 is not defined by Android. We can check for invalid input key codes, but that's rather ugly and just hides the problem.

titovmaxim:
[...]

PR #3410, opened 2015-09-11
Fixes onKey ArrayIndexOutOfBoundsException on negative keycodes

Guy-kun:
Issue #2807 already partially fixed this issue but an additional check is not being made for negative keycodes in onKey causing crashes on some Chinese devices that send -2

```
java.lang.ArrayIndexOutOfBoundsException: length=260; index=-2
       at com.badlogic.gdx.backends.android.AndroidInput.onKey(SourceFile:498)
       at android.view.View.dispatchKeyEvent(View.java:7338)
```

(The issue also happens in a number of flavors further down onKey)

```
java.lang.ArrayIndexOutOfBoundsException: length=260; index=-2
       at com.badlogic.gdx.backends.android.AndroidInput.onKey(SourceFile:530)
```


saltares:
Thanks!

PR #3415, opened 2015-09-13
Add lerpAngle() and lerpAngleDeg() to MathUtils. Closes #3307.

### LINKED ISSUE #3307: Request: LerpAngle on MathUtils.
I would write a PR for this, since I found myself needing it more than once this week, but can't guarantee the solution I found to the problem is performant enough.

For more info, [this](http://docs.unity3d.com/ScriptReference/Mathf.LerpAngle.html) is what I'm requesting.

And [this](http://stackoverflow.com/a/14498790/1091440) is the best way that I found to do it . Anyone know a better way of doing it?

GreenLightning:


Tom-Ski:
Thanks!

PR #3416, opened 2015-09-14
Check status before binding to default frame buffer again.

code-disaster:
glCheckFramebufferStatus() was always called with the default frame buffer bound.


xoppa:
Thanks!

PR #3423, opened 2015-09-18
Add Input#isCatchMenuKey

intrigus:


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSInput.java
```diff
@@ -529,11 +529,16 @@ public void setCatchBackKey (boolean catchBack) {
 	@Override
 	public boolean isCatchBackKey () {
 		return false;
-	}
+	}
```
nooone: What happened here?

nooone:
Besides the strange tab being added, it looks good to me.

PR #3428, opened 2015-09-19
Fix false-positive of vibration support

intrigus:
We get our vibrator with
context.getSystemService(Context.VIBRATOR_SERVIBE).

The previous check tested if(vibrator != null), but according to 
http://developer.android.com/reference/android/content/Context.html#getSystemService(java.lang.String)  
it returns "The service or null if the name does not exist.". So it
should only return null if the name does not exist, which should be
never the case.
Instead of that we now use vibrator.hasVibrator()

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -712,7 +712,7 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 		if (peripheral == Peripheral.Compass) return compassAvailable;
 		if (peripheral == Peripheral.HardwareKeyboard) return keyboardAvailable;
 		if (peripheral == Peripheral.OnscreenKeyboard) return true;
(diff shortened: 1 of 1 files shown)


saltares:
We don't lose anything by checking for null here. I'm against removing the check.

intrigus:
Sure

xoppa:
[Vibrator#hasVibrator()](http://developer.android.com/reference/android/os/Vibrator.html#hasVibrator()) was added in api level 11.

intrigus:
Why did you close it @Tom-Ski ?

Tom-Ski:
It forces us to move up the lowest android api level supported by 3.

intrigus:
Shouldn't it be possible to do
```
if (Build.VERSION.SDK_INT >= 11) {
   vibrator.hasVibrator();
}
```

?

MobiDevelop:
Yes

intrigus:
So why does it forces us to move up the lowest android api level?

MobiDevelop:
It only does because you did not implement it with a version guard.

PR #3430, opened 2015-09-21
Raycast test for picking individual triangles on a collision shape.

jsjolund:
Useful for finding triangles on navigation meshes for example.


xoppa:
Thanks!

PR #3436, opened 2015-09-24
IsometricTiledMapRenderer amendment to accept a unit scale value for tile height

del-sol:
Currently, tiled maps in LibGdx only covers tiles that have the same width and length. In this case, the isometric renderer applies the same value for unit scale to a tile's height, which will then render the whole map vertically off-center. For applications that rely on screen coordinates matching to world coordinates, that match to an underlying structure such as a graph for pathfinding or a grid, this situation throws things out of whack. It forces users to restrict themselves to tilesets with tiles of the same width and height.

This simple modifications to the isometric renderers provides constructors to pass a unit scale for height, a separate field to reference it, and applies this value in the rendering process.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
@@ -95,9 +95,9 @@ public class IsometricStaggeredTiledMapRenderer extends BatchTiledMapRenderer {
 					TextureRegion region = tile.getTextureRegion();
 
 					float x1 = x * layerTileWidth - offsetX + tile.getOffsetX() * unitScale;
-					float y1 = y * layerTileHeight50 + tile.getOffsetY() * unitScale;
+					float y1 = y * layerTileHeight50 + tile.getOffsetY() * unitScaleY;
 					float x2 = x1 + region.getRegionWidth() * unitScale;
-					float y2 = y1 + region.getRegionHeight() * unitScale;
+					float y2 = y1 + region.getRegionHeight() * unitScaleY;
 
 					float u1 = region.getU();
 					float v1 = region.getV2();
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
@@ -28,20 +28,36 @@ import com.badlogic.gdx.maps.tiled.TiledMapTileLayer.Cell;
 
 public class IsometricStaggeredTiledMapRenderer extends BatchTiledMapRenderer {
 
+	private float unitScaleY;
+	
 	public IsometricStaggeredTiledMapRenderer (TiledMap map) {
 		super(map);
+		unitScaleY = 1.0f;
 	}
 
 	public IsometricStaggeredTiledMapRenderer (TiledMap map, Batch batch) {
 		super(map, batch);
+		unitScaleY = 1.0f;
 	}
 
 	public IsometricStaggeredTiledMapRenderer (TiledMap map, float unitScale) {
 		super(map, unitScale);
+		unitScaleY = unitScale;
 	}
 
+	public IsometricStaggeredTiledMapRenderer (TiledMap map, float unitScale, float unitScaleY) {
+		super(map, unitScale);
+		this.unitScaleY = unitScaleY;
+	}
+	
 	public IsometricStaggeredTiledMapRenderer (TiledMap map, float unitScale, Batch batch) {
 		super(map, unitScale, batch);
+		unitScaleY = unitScale;
+	}
+
+	public IsometricStaggeredTiledMapRenderer (TiledMap map, float unitScale, float unitScaleY, Batch batch) {
+		super(map, unitScale, batch);
+		this.unitScaleY = unitScaleY;
 	}
 
 	@Override
@@ -53,7 +69,7 @@ public class IsometricStaggeredTiledMapRenderer extends BatchTiledMapRenderer {
 		final int layerHeight = layer.getHeight();
 
 		final float layerTileWidth = layer.getTileWidth() * unitScale;
-		final float layerTileHeight = layer.getTileHeight() * unitScale;
+		final float layerTileHeight = layer.getTileHeight() * unitScaleY;
 
 		final float layerTileWidth50 = layerTileWidth * 0.50f;
 		final float layerTileHeight50 = layerTileHeight * 0.50f;
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
@@ -40,26 +40,44 @@ public class IsometricTiledMapRenderer extends BatchTiledMapRenderer {
 	private Vector2 topLeft = new Vector2();
 	private Vector2 bottomRight = new Vector2();
 
+	private float unitScaleY;
+	
 	public IsometricTiledMapRenderer (TiledMap map) {
 		super(map);
+		unitScaleY = 1.0f;
 		init();
 	}
 
 	public IsometricTiledMapRenderer (TiledMap map, Batch batch) {
 		super(map, batch);
+		unitScaleY = 1.0f;
 		init();
 	}
 
 	public IsometricTiledMapRenderer (TiledMap map, float unitScale) {
 		super(map, unitScale);
+		unitScaleY = unitScale;
 		init();
 	}
 
+	public IsometricTiledMapRenderer (TiledMap map, float unitScale, float unitScaleY) {
+		super(map, unitScale);
+		this.unitScaleY = unitScaleY;
+		init();
+	}
+	
 	public IsometricTiledMapRenderer (TiledMap map, float unitScale, Batch batch) {
 		super(map, unitScale, batch);
+		unitScaleY = unitScale;
 		init();
 	}
 
+	public IsometricTiledMapRenderer (TiledMap map, float unitScale, float unitScaleY, Batch batch) {
+		super(map, unitScale, batch);
+		this.unitScaleY = unitScaleY;
+		init();
+	}
+	
 	private void init () {
 		// create the isometric transform
 		isoTransform = new Matrix4();
@@ -87,7 +105,7 @@ public class IsometricTiledMapRenderer extends BatchTiledMapRenderer {
 		final float color = Color.toFloatBits(batchColor.r, batchColor.g, batchColor.b, batchColor.a * layer.getOpacity());
 
 		float tileWidth = layer.getTileWidth() * unitScale;
-		float tileHeight = layer.getTileHeight() * unitScale;
+		float tileHeight = layer.getTileHeight() * unitScaleY;
 		float halfTileWidth = tileWidth * 0.5f;
 		float halfTileHeight = tileHeight * 0.5f;
 
@@ -227,3 +245,4 @@ public class IsometricTiledMapRenderer extends BatchTiledMapRenderer {
 		}
 	}
 }
+	
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
@@ -143,9 +143,9 @@ public class IsometricTiledMapRenderer extends BatchTiledMapRenderer {
 					TextureRegion region = tile.getTextureRegion();
 
 					float x1 = x + tile.getOffsetX() * unitScale;
-					float y1 = y + tile.getOffsetY() * unitScale;
+					float y1 = y + tile.getOffsetY() * unitScaleY;
 					float x2 = x1 + region.getRegionWidth() * unitScale;
-					float y2 = y1 + region.getRegionHeight() * unitScale;
+					float y2 = y1 + region.getRegionHeight() * unitScaleY;
 
 					float u1 = region.getU();
 					float v1 = region.getV2();


gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
```diff
@@ -28,20 +28,36 @@
 
 public class IsometricStaggeredTiledMapRenderer extends BatchTiledMapRenderer {
 
+	private float unitScaleY;
```
intrigus: Why don't you add it to BatchTiledMapRenderer, and why did you make it private?

gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
```diff
@@ -40,26 +40,44 @@
 	private Vector2 topLeft = new Vector2();
 	private Vector2 bottomRight = new Vector2();
 
+	private float unitScaleY;
```
intrigus: Same here.

intrigus: What's with https://github.com/del-sol/libgdx/blob/master/gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java#L197-L200 and https://github.com/del-sol/libgdx/blob/master/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java#L145-L148 ?

gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java
adam-law: Since it's up for review, I decided to modify the code to fit my use case needs i.e. Isometric. Other tile maps predominantly use equivalent values for tile length and width, especially if a user intends to match a map up to a grid or graph, while keeping things as simple as possible.

Isometric maps are different because tiles with differing values for length and width, and tile sets that use only one value for these, are represented roughly equally. You need to supply the proper unit scales for the map to be centered on the default view, and match up to world coordinates.

As for making it private, all the other fields in the class are marked so, and the intent for each is to be used internally. 

Furthermore, in this case, since the right unit scale is coupled to a tile's pixel dimensions, there is no use to expose it for modification, unless the tiles' dimensions itself, is modified. Setting the value for unit scale is usually down outside the creation of the renderer, where one can infer the tile dimensions, and compute the value for unit scale.

gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java
adam-law: Thanks for the catch. I could have sworn I changed that part as well.

MobiDevelop:
I'm not sure I understand what the problem is that this is trying to solve. Why are pixels needing to be scaled differently depending on which axis they are on? Can you give a concrete example where pixels should be scaled differently depending on x or y?

adam-law:
Hi, I knew I should have used something different other than X and Y. It should be tile width and height. 

To summarize, this is to help make isometric maps draw centered on the screen, so that stuff like camera.unproject will match up to "locations" on the map.

There is only one unitScale field/value on all Tiled map renderers. It's applied in the rendering computation on both tile dimensions, and the tile map itself. This works well for maps using tiles with the same width and height e.g. 32 pixels wide by 32 pixels high. 

However, in isometric maps using tile sets that have tiles with a different width from height e.g. 64x32, the map will be drawn vertically off-center, considering the camera position is at default, and the correct unit scale is passed for tile width. 

For example, supplying 64 (or 1/64 in world coords) as the unit scale when instantiating an isometric tiled map renderer, will draw the map horizontally centered, and at the bottom of the screen. Since the iso map applies the same unit scale on the map dimensions itself, the rendered result is not quite in the correct scale, when compared to when viewed in Tiled.

MobiDevelop:
I'm going to need you to supply a code example (and possibly some images) showing the actual problem this is solving. I copied your changes and for our test iso map (having a 64x32 grid) and used different unitScale and unitScaleY values and it surely did not have any desired effect.

adam-law:
Okay, may I have the link to which map you used to test so I can write the code? Did this map's tiles have different values for width and height? Grid dimensions are not relevant, if the tile's "sides" are equal.

MobiDevelop:
You should more clearly state the problem conditions (or provide an actual example showing there is a problem, as I've been asking for).

The map I used is here: https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-android/assets/data/maps/tiled/isometric_grass_and_water.tmx
It does not have different tile image width and height, but I still don't understand how that matters (because you still have not clearly expressed what the actual problem is).

adam-law:
My sincere apologies. English is not my first language. I don't know how to say that a map made up off tiles having different width and height, is not vertically centered, and out of scale, any more clearer.

I used this map as well, and, the tiles that make up that map are each 64x32 (viewed in Tiled). Here's an image of how it projects on my application passing 1/64 for unit scale
([Isometric map centered horizontally, but not vertically](http://imgur.com/Qj9Dn6a).)

Ignore the small white box, it's just the mini-map. I cropped the image, but as can be seen, the map renders at the very bottom of the screen, since the 1/64 unit scale, is also applied to height of each tile, where it's supposed to be 1/32. 

I only use these camera settings to render:
    
    float pixelWidth = scene.getPixelWidth(); // tile dimensions queried from Tile map properties
    float pixelHeight = scene.getPixelHeight();

    OrthographicCamera camera = new OrthographicCamera();
    camera.setToOrtho(false, scene.getWidth(), scene.getHeight()); // also pulled from the tile map properties
    camera.position.set(scene.getWidth() * 0.5f, scene.getHeight() * 0.5f, 0);    
    ExtendViewport sceneViewport = new ExtendViewport(pixelWidth, pixelHeight * (Gdx.graphics.getHeight() / Gdx.graphics.getWidth()), camera);
    sceneViewport.apply();

With the code amendments, passing 1/64 for unitScale, and 1/32 for unitScaleY should render a centered map, with correct scale. Conversely, just passing 1/64 will result in what is seen in the image.

MobiDevelop:
I passed in 1/32f for the unitScaleY and it was nothing like it should be. The positioning has absolutely nothing to do with the unitScale.

So the actual issue is that you want to offset the map? Your code doesn't do that.

adam-law:
No, not offset. I want the map centered in the middle of screen, on both the x and y axes. With this in mind, shouldn't the unit scales for a 64x32 tile be 1/64 and 1/32, respectively? Isometric maps aside, here's my motivation. Since the renderers for "top-down" tile maps use one unit scale for both width and height, and have tiles with the same width and height, the rendered map is properly centered, a grid drawn starting from position (or "cell) 0, 0, matches perfectly to the tiles. 

On the same train of thought, to achieve the same centered look with isometric maps (which is what I want) shouldn't matching values for unit scale be used, when the sides aren't the same? Or is the screenshot I provided the way things are supposed to be (in the bottom middle of the screen)? If it's not too much to ask, may I ask for a screenshot of how the map looked like on your test? (sorry for the bother)

Thank you.

MobiDevelop:
Unit scale only exists to turn the pixel units of tmx maps into camera units. It is not tied to tiles unless you choose to make it so. In the case of a 64 x 32 tile (or grid) size, you may choose to make the width of one tile equal to one camera unit (1/64f) or the height of one tile (1/32f) or perhaps the width of two tiles is 1 unit (1/128f). Maybe you want the height of one tile to be equal to two camera units (1/16f). It is entirely up to you, it is just convenient to make one tile equal to one camera unit.

To vertically center the map, you need to understand where the drawing origin is, and the direction drawing goes from there:
The bottom left of the left most cell is at (0,0). Moving along the width of the map, each cell shifts downward by half the tile height and rightward by half the tile width. Moving along the height of the map, each cell shifts upward by half the tile hight and rightward by half the tile height.
screen shot 2015-09-29 at 9 27 41 pm
In this screenshot, I've illustrated how this looks - the red box simply starts at 0,0 and extends the size of the camera viewport. The green box shows the map bounds. For further clarification I've annotated the origin as well as the row and column axes.

Knowing this, it should be relatively easy to translate the camera so that the map is centered within the viewport.

adam-law:
Thank you for the detailed explanation, I really appreciate it :) 

I apologize that I didn't create this pull request with the adequate code and screenshots to display what I'm seeing. I still felt that there is a need for this code submission, so I started a new project to boil down to what's been bothering me. As a result, I did find something to consider.

When I'm ready to explain further, should I pastebin the test project, or should I just put it here? I'll have it up within the day since I'm not on my home PC.

Thanks for all your help!

MobiDevelop:
I can't imagine when you'd want to scale the y pixels differently than the x pixels (except maybe for a temporary visual effect), but I'm all ears. Go ahead and post a small example here if you have one.

In the meantime I'm going to close this and reopen if necessary.

nhydock:
Only reason why I can think someone would want taller Y scale than X is if they want to make a game with the visual style of Ultima

adam-law:
@nhydock On my part, I just want normal rendering. Thanks.

So I've stripped my rendering code to bare essentials, using only a camera, and no view port. Sorry if something's wrong in there.

Here's my main method:

    public class DesktopLauncher {
	public static void main (String[] arg) {
		LwjglApplicationConfiguration config = new LwjglApplicationConfiguration();
		config.width = 1920;
		config.height = 1080;
		new LwjglApplication(new Launch(), config);
	}
   }

...and my ApplicationListener

    public class Launch implements ApplicationListener {

	private IsometricTiledMapRenderer renderer;
	private OrthographicCamera camera;
	private SpriteBatch batch;
	private float w;
	private float h;
	
	@Override
	public void create() {
		batch = new SpriteBatch();
		AssetManager assetManager = new AssetManager();
    // I had some weird issue w/ the Android project after using the setup app, so I loaded the tmx from the classpath
		assetManager.setLoader(TiledMap.class, ".tmx", new TmxMapLoader(new ClasspathFileHandleResolver())); 
		assetManager.load("com/testbed/isometric_grass_and_water.tmx", TiledMap.class);
		assetManager.finishLoading();
		TiledMap tiledMap = assetManager.get("com/testbed/isometric_grass_and_water.tmx", TiledMap.class);

		w = (int) tiledMap.getProperties().get("width");
		h = (int) tiledMap.getProperties().get("height");
		camera = new OrthographicCamera();
		camera.setToOrtho(false, w, h);
		//renderer = new IsometricTiledMapRenderer(tiledMap); // no unit scale
		//renderer = new IsometricTiledMapRenderer(tiledMap, 1/64f); // width unit scale only
		renderer = new IsometricTiledMapRenderer(tiledMap, 1/64f, 1/32f); // my amendment
	}

	}

	@Override
	public void render() {
		Gdx.gl.glClearColor(0, 0, 0, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT | GL20.GL_DEPTH_BUFFER_BIT);

		camera.update();
		batch.setProjectionMatrix(camera.combined);
		batch.begin();
		renderer.setView(camera.combined, 0, 0, w, h);
		renderer.render();
		batch.end();
	}

	@Override
	public void pause() {
	}

	@Override
	public void resume() {
	}

	@Override
	public void dispose() {
	}
    }

Using this, I tried 3 different scenarios: 1. without passing a unit scale, 2. using the same unit scale for both width and height, and 3. using different values for unit scale. The 1st 2 are the default constructors LibGDX uses, of course.

Here are screen shots for the tests:

Default constructor, w/o unit scale.
renderer without passing unit scale

Default constructor, w/ unit scale (matching 1/64 for tile width)
same unit scale

Submitted code, w/ unit scales matching width and height (1/64 and 1/32 respectively)
different unit scales

As can be seen, the 1st two images shows the map being drawn weird. The 1st understandably because the camera is in world units, but I believe the second image is not isometric, whereas, the 3rd image is.

Thanks.

MobiDevelop:
That code is very wrong.

adam-law:
Okay, I'm more than willing to admit I'm only going by my limited knowledge. My code could also be cleaner (sorry I just pasted it here when I was able to simulate the problem). Here's the cleaner version:

    package com.testbed;

    import com.badlogic.gdx.ApplicationListener;
    import com.badlogic.gdx.Gdx;
    import com.badlogic.gdx.assets.AssetManager;
    import com.badlogic.gdx.assets.loaders.resolvers.ClasspathFileHandleResolver;
    import com.badlogic.gdx.graphics.OrthographicCamera;
    import com.badlogic.gdx.graphics.g2d.SpriteBatch;
    import com.badlogic.gdx.maps.tiled.TiledMap;
    import com.badlogic.gdx.maps.tiled.TmxMapLoader;
    import com.badlogic.gdx.maps.tiled.renderers.IsometricTiledMapRenderer;

    public class Launch implements ApplicationListener {

	private IsometricTiledMapRenderer renderer;
	private OrthographicCamera camera;
	private SpriteBatch batch;
    //	private ExtendViewport sceneViewport;
	private float w;
	private float h;
	
	@Override
	public void create() {
		batch = new SpriteBatch();
		AssetManager assetManager = new AssetManager();
		assetManager.setLoader(TiledMap.class, ".tmx", new TmxMapLoader(new ClasspathFileHandleResolver()));
		assetManager.load("com/testbed/isometric_grass_and_water.tmx", TiledMap.class);
		assetManager.finishLoading();
		TiledMap tiledMap = assetManager.get("com/testbed/isometric_grass_and_water.tmx", TiledMap.class);

		w = (int) tiledMap.getProperties().get("width");
		h = (int) tiledMap.getProperties().get("height");
		camera = new OrthographicCamera();
		camera.setToOrtho(false, w, h);
		camera.translate(0, -h * 0.5f); // just to see things clearer 
		camera.update();
		
    //		renderer = new IsometricTiledMapRenderer(tiledMap);
    //		renderer = new IsometricTiledMapRenderer(tiledMap, 1/64f);
		renderer = new IsometricTiledMapRenderer(tiledMap, 1/64f, 1/32f);
	}

	}

	@Override
	public void render() {
		Gdx.gl.glClearColor(0, 0, 0, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT | GL20.GL_DEPTH_BUFFER_BIT);

		renderer.setView(camera.combined, 0, 0, w, h);
		renderer.render();
		
	}

	@Override
	public void pause() {
	}

	@Override
	public void resume() {
		// TODO Auto-generated method stub
		
	}

	@Override
	public void dispose() {
		// TODO Auto-generated method stub
		
	}
    }

As for the intent of the test, I thought it was as vanilla as could be, in terms of  in order to draw a tile map and have things in world coordinates. If you would be so kind, what is the correct way then to do this?

MobiDevelop:
You are setting your camera viewport without any regard for the aspect ratio of the window, which will cause undesired scaling in one direction. Also, why are you using that version of `setView`?

adam-law:
As shown on an earlier post, I am using an ExtendViewport set to the aspect ratio in my app, but I assumed (wrongly, it seems) that a vanilla test is what's merited for testing, so I just used a camera.

I have no excuses for using this setView version, other than I was playing around with finding how to center the map, without translating the camera. I should have cleaned that up, sorry. I did go over material online and do know that doing "setView(camera)" suffices for the test code.

MobiDevelop:
Let's move to the forum for additional support on this subject. The comments section here isn't the correct place for support requests.

We won't merge this change as it doesn't actually solve a real problem.

adam-law:
I have no problem with that. Thanks @MobiDevelop for the extra effort :)

PS I've changed the images above to ones that show the map centered. It will be more obvious for people to see what I mean i.e. same unit scale = skewed map, different scale = correct isometric view

adam-law:
Sorry for the double post, I forgot to ask, isn't the fact that the map is skewed (see image above), a real problem? Is the 2nd image above really how an isometric map should look in LibGDX?

MobiDevelop:
It is skewed because you are setting up your viewport wrong. Please use the forum instead of more comments here.

adam-law:
sorry, got caught up in RL. posted [to this thread.](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=20774).

Thanks.

plunntic:
@MobiDevelop could you tell why using this version of `setView()` is bad? According to [LibGDX wiki](https://github.com/libgdx/libgdx/wiki/Tile-maps#map-renderer) there is no statement about when to use which version.

Sorry for posting here, but you told @adam-law to to forum, but no one ever responded, and I really couldn't find anywhere else info on that topic - so it at least could be a complete information for possible future readers.

MobiDevelop:
Using either version of setView isn't wrong, unless you use them wrong. Pretending that units are different in the x and y direction is what @adam-law was doing wrong.

adam-law:
You know, I'm always the kind of guy who'll be the first to admit a mistake, if there was one. You keep telling me I'm wrong, and that's fine and all, but, with all due respect, what then is the correct way to set the view, given the information I've already provided? 

Also, respectfully, I don't get why you have to call my attempts at finding a solution as "pretending". I get that you are a major contributor to LibGDX, and I'm also aware it was you who authored this part of the code. For all intents and purposes, you're obviously way better in this field than I am, but was I ever disrespectful to you to merit your condescension? People have to begin learning from somewhere, you know. I apologize if I'm not as good as you are, and I understand how providing technical knowledge and support to the masses can get repetitive and tiring, but aren't you laying it on a bit thick? In the first place, this wasn't even a question to start with, but a contribution to the code base, regardless of how small and insignificant it may be.

Anyway, this pull request displays the correct view for my use case, regardless of the critiquing it has received, whereas using the available constructors does not. That still remains the issue that needs to be solved. If it's really caused by setting the view wrong, again, then what's the correct way? The test code I supplied seems to be the simplest case for setting the view as it is, but what do I know, right?

I've posted to the forum as you've asked, but like @plunntic mentioned, no one answered, even after all this time, so what was the point? Did I somehow use the wrong format there as well? From experience, it seems it's either a hit or miss getting help there, ever since I started learning LibGDX, and I don't think I'm the only one who feels this way.

bornskilled200:
I will be posting an correct example for this in the [forum post](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=20774) above, please redirect any more questions there.

loogie:
Hello!

 

You have a new message, please read <http://suntech7.com/traffic.php?w>

 

loogie@basicelementdesign.com

loogie:
Hello!

 

You have a new message, please read <http://dotmybusiness.com/catholic.php?cy>

 

loogie@basicelementdesign.com

PR #3439, opened 2015-09-25
Hiero - add .json file format for webgl / javascript use

toloudis:
I write the .fnt data out to json, mimicking the bmfont text format as closely as possible.
Added a -j command line option to indicate output should go to json, and a new menu item to direct output to json.


MobiDevelop:
What is wrong with the text format?

toloudis:
In the interest of small code size for mobile web apps, it is very convenient to have a .json file that can be downloaded and parsed with no extra code.

darthdurden:
I'd also love for Hiero to output in the AngelCode xml format.  I have some tools that read that in already, and it'd be a lot easier to fix the output than to fix the input.  Is that already support somehow that I'm not aware of?

darthdurden:
After digging a bit more, it looks like XML is supported in the code, but not actually exposed via the UI (or commandline).  Am I missing something?

NathanSweet:
This PR has conflicts, it doesn't seem to use the libgdx formatter, and it adds a dependency to emit JSON when the libgdx JsonWriter could be used instead.

I'm not sure Hiero should output the data in JSON, since we'd be creating a new format that is unlikely to be supported by any tools. I'm not completely against it though.

If there is an AngelCode XML format it would make sense for Hiero to export to that. Probably it just needs a menu item, I don't remember the state of the XML export code.

nhydock:
Seeing as Hiero is a valuable tool for more than just LibGDX applications, is there a reason why it is still packaged directly with LibGDX instead of in a separate repository that depends on LibGDX?

nhydock:
> /me marks this as libgdx 2.0

Could we please have a roadmap or at least a milestone added to the project's issue tracker for planned libgdx 2.0 features/ideas?

nhydock:
Regarding the JSON exporting implementation, should we really be adding the javax.json dependency to the tools project when LibGDX already has its own [json parser](https://github.com/libgdx/libgdx/wiki/Reading-%26-writing-JSON) built-in?  

Usually I'm all for using more standard libraries to leverage dependency on the JVM reference implementation as opposed to nothing but internal/third-party libraries.  However, javax.json, even though it is the reference implementation of JSR 353, is targeted more for Java EE as opposed to Java SE.   Additionally, it competes with org.json which is what Android primarily uses.  

Even though gdx-tools is a desktop only project, LibGDX as a whole is multiplatform.  I recommend sticking with what LibGDX already has and uses instead of introducing competing standards into the project, as that was the purpose of having the parsers in the first place.

badlogic:
Closed out due to @NathanSweet 's comments. As for milestones/roadmaps, there's a libGDX 2.0 issue on the tracker where peeps can put together a wish list. Thing is, the core team can't commit to that.

toloudis:
I agree with the comments.  At some point I will revisit, fix conflicts, and use the internal json writer library. If I do those things, I'll resubmit.  It seems the interest in having this as a feature is not so high, so it'll stay for my internal use for the time being.

PR #3440, opened 2015-09-26
Add support for Vibration on GWT

intrigus:
This PR adds support for vibration on GWT.
Repeated patterns are emulated, because the api doesn't support that.

There is a problem:
It's not possible to detect whether a device has a vibrator.
We can only check if the browser supports the vibration API, so it's possible that the browser supports vibration, but the device has no vibrator.

That's why Input.isPeripheralAvailable(Peripheral#Vibrator) does return true, if the browser supports vibration. 

I also added a better test.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
@@ -28,6 +28,7 @@ import com.badlogic.gdx.utils.TimeUtils;
 import com.gargoylesoftware.htmlunit.javascript.host.Navigator;
 import com.google.gwt.core.client.JavaScriptObject;
 import com.google.gwt.core.client.JsArray;
+import com.google.gwt.core.client.JsArrayNumber;
 import com.google.gwt.dom.client.CanvasElement;
 import com.google.gwt.dom.client.Document;
 import com.google.gwt.dom.client.Element;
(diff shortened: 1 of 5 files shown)


Tom-Ski:
Nice, whats the support like for this?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
```diff
@@ -264,9 +323,23 @@ public boolean isPeripheralAvailable (Peripheral peripheral) {
 		if (peripheral == Peripheral.HardwareKeyboard) return true;
 		if (peripheral == Peripheral.MultitouchScreen) return isTouchScreen();
 		if (peripheral == Peripheral.OnscreenKeyboard) return false;
-		if (peripheral == Peripheral.Vibrator) return false;
+		if (peripheral == Peripheral.Vibrator) if (isVibratorAvailableJSNI()) {
```
MobiDevelop: Code like this makes me sad.

Tom-Ski: heh

intrigus: I shouldn't commit, when I'm tired :D
Sry^^

intrigus:
@Tomski http://caniuse.com/#feat=vibration

Tom-Ski:
So pretty awful then?

intrigus:
Support for vibration is ~65.5%, support for webgl is 80% so it's not extremely awful only 15% more awful ^^

Tom-Ski:
Thats API support rather than functionality though? It should only be able to vibrate if its supported in hardware, most of the time that means a phone/tablet which will be running the mobile browsers, for which support is way less.

MobiDevelop:
Even if functional support is low, it doesn't hurt to have it.

MobiDevelop:
What happened with Input.java and GwtInput.java? Did the line endings get changed or something?

intrigus:
I'm going to fix it, when I'm healthy again.

intrigus:
Line endings are fixed :)

Things to do:

- [ ] Clean/Refactor AdvancedVibratorTest
- [ ] Verify that vibration is exactly the same as on Android.

MobiDevelop:
The problem with "todo" lists on a PR is that they never get "todone" because nobody knows who the todo is intended for. Basically it makes the PR seem like a WIP.

intrigus:
The todo lists is intended for me ^^
Somehow it's a WIP, the testcases that I tried worked fine, but I'd like to verify that also complex testcases work.

I'll hope that I can finish the PR in about 1-2 weeks :)

badlogic:
Hey @intrigus wanna fix it up and resubmit? At your own pace, i'll close this one out. Please resubmit once you are done!

SimonIT:
@intrigus Are there any reasons for not merging this? I tested it myself and it seem to work fine

intrigus:
> @intrigus Are there any reasons for not merging this? I tested it myself and it seem to work fine

It should probably work although I'm not sure anymore.
Looking at the comments/docs I wrote, I can say that some of them are less than ideal :D
I'd love to overhaul everything and do a new PR but I don't have time.

In the current state I'd rather not merge it...

PR #3445, opened 2015-09-28
Fix false-positive of vibration support

intrigus:
See https://github.com/libgdx/libgdx/pull/3428

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -28,6 +28,7 @@ import android.hardware.Sensor;
 import android.hardware.SensorEvent;
 import android.hardware.SensorEventListener;
 import android.hardware.SensorManager;
+import android.os.Build;
 import android.os.Handler;
 import android.os.Vibrator;
 import android.service.wallpaper.WallpaperService.Engine;
@@ -712,7 +713,8 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 		if (peripheral == Peripheral.Compass) return compassAvailable;
 		if (peripheral == Peripheral.HardwareKeyboard) return keyboardAvailable;
 		if (peripheral == Peripheral.OnscreenKeyboard) return true;
-		if (peripheral == Peripheral.Vibrator) return vibrator != null;
+		if (peripheral == Peripheral.Vibrator)
+			return (Build.VERSION.SDK_INT >= 11 && vibrator != null) ? vibrator.hasVibrator() : vibrator != null;
 		if (peripheral == Peripheral.MultitouchScreen) return hasMultitouch;
 		return false;
 	}


MobiDevelop:
While we are making changes, does it make sense to also check whether permission has been granted to use the vibrator?

intrigus:
The "android.permission.VIBRATE"-permission is automatically granted if it's declared in the Manifest.xml.
Or what do you mean?
See http://developer.android.com/guide/topics/security/permissions.html#normal-dangerous and 
http://developer.android.com/guide/topics/security/normal-permissions.html.

MobiDevelop:
And if you don't add that permission to the manifest, then you don't have permission, and you don't have vibrator support.

intrigus:
You suggest that we check if the application has the permission?And if that's the case, vibrate?
I don't really see the advantage of permission checking.

MobiDevelop:
I'm suggesting that reporting support for Vibrator is incorrect if you can't use it (due to not having the correct permission to do so).

intrigus:
I don't think that we should return false, if we don't have the correct permission. 

Pseudoc
'''
if (isVibratorAvailable())
{
vibrate ()
}

MobiDevelop:
It was just a suggestion. If you want to pick and choose which false positives you want to catch, then it is what it is.

It doesn't lead to hard to find bugs, it leads to a correct indication of whether vibrate is supported. I'll leave it to others to weigh in.

xoppa:
I agree that the method `isPeripheralAvailable` should return whether the peripheral is available. Imo "available" means that you can actually use it. If there's no permission, then you can not use it, so i'd say it is not available. I don't see how this will lead to hard to find bugs, can you explain that @intrigus? Note that you can still call the vibrate method without checking whether it is available, if you prefer it to crash.

intrigus:
If you always check whether isPeripheralAvailable returns true and only vibrate if it returns true you will never vibrate. 
I think it's very hard to debug this,  because you have no crash, no exceptions, nothing. And the cause is not too obvious. 

I'd say that isPeripheralAvailable should return whether the peripheral is available. Available means that the peripheral exists. 

@Tom-Ski Could you please share your opinion on this?

MobiDevelop:
Knowing something exists does you absolutely no good if you cannot use it. It might as well not exist.

It isn't hard to debug because it isn't a bug... not vibrating when the vibrator is not available for use is the expected behavior.

xoppa:
Ow, you mean that not having the permission is clearly a programmer error. You should never have to check at runtime whether it has the permission, because it is always granted. If you don't have the permission then that's an indication that something is wrong with the manifest xml.

There's something to say for that indeed. But then, it would also indicate that the developer didn't even test the vibrate functionality he wrote.

However, if android, for some reason, decides in the future that the user does have to explicitly grand permission to use vibrate (and the user can decide not to grand that right but still use the application). Then I don't think that simply crashing is the desired behavior. Instead, not vibrating would be better in that case.

MobiDevelop:
Android 6.0 introduces the ability for users to grant permissions on demand.

intrigus:
Can you also revoke normal permissions like vibration? I'm not sure because normal permissions are automatically granted.

MobiDevelop:
I just installed the 6.0 update for my phone and it appears that vibration is not a permission you can revoke.

PR #3446, opened 2015-09-28
Add ShaderTester, it wraps Khronos shader compiler

intrigus:
This PR adds a thin wrapper around the the [official Khronos reference shader compiler](https://github.com/KhronosGroup/glslang), it can process
whole directories and shaders as Strings.

There are some open questions:
It should be also possible to directly use the compiler through JNI, this would remove the limitation, that e.g. fragment shader MUST have .frag as extension.
But I don't know anything about JNI and jnigen.

Mac support is not available because, there are no binaries available for Mac.

The biggest problem is https://github.com/intrigus/libgdx/blob/9a4e6620b0d96116803b521f994d6d447e084104/extensions/gdx-tools/src/com/badlogic/gdx/tools/shadertester/ShaderTester.java#L41-L105 this is stolen from SharedLibraryLoader.getExtractedFile(). The method I copied is private so I'd have to invoke SharedLibraryLoader.getExtractedFile() through Reflection, which would remove duplicated code, but I don't like relying on Reflection.


xoppa:
Thanks! I don't think there's a need for this though. Shader editors typically provide the same (and more) information in a more convenient way.

badlogic:
Yeah, thanks Intrigus, but what Xoppa said :)

PR #3449, opened 2015-10-02
ShadowSystem 1/6

realitix:
The shadow system pull request has been splitted in 6 parts.
This is the first.

1/6 -> Add interfaces.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
@@ -1,0 +1,97 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.system;
+
+import java.util.Set;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.Cubemap.CubemapSide;
+import com.badlogic.gdx.graphics.g3d.environment.DirectionalLight;
+import com.badlogic.gdx.graphics.g3d.environment.PointLight;
+import com.badlogic.gdx.graphics.g3d.environment.SpotLight;
+import com.badlogic.gdx.graphics.g3d.utils.ShaderProvider;
+
+/** Shadow system provides functionalities to render shadows
+ * @author realitix */
+public interface ShadowSystem {
+
+	/** Return number of pass
+	 * @return int */
+	public int getPassQuantity ();
+
+	/** Return shaderProvider of the pass n
+	 * @return ShaderProvider */
+	public ShaderProvider getPassShaderProvider (int n);
+
+	/** Return shaderProvider used for main rendering
+	 * @return ShaderProvider */
+	public ShaderProvider getShaderProvider ();
+
+	/** Add spot light in shadow system
+	 * @param spot SpotLight to add in the ShadowSystem */
+	public void addLight (SpotLight spot);
+
+	/** Add directional light in shadow system
+	 * @param dir DirectionalLight to add in the ShadowSystem */
+	public void addLight (DirectionalLight dir);
+
+	/** Add point light in shadow system
+	 * @param point PointLight to add in the ShadowSystem */
+	public void addLight (PointLight point);
+
+	/** Add point light in shadow system
+	 * @param point PointLight to add in the ShadowSystem
+	 * @param sides Set of side */
+	public void addLight (PointLight point, Set<CubemapSide> sides);
+
+	/** Remove light from the shadowSystem
+	 * @param spot SpotLight to remove in the ShadowSystem */
+	public void removeLight (SpotLight spot);
+
+	/** Remove light from the shadowSystem
+	 * @param dir DirectionalLight to remove in the ShadowSystem */
+	public void removeLight (DirectionalLight dir);
+
+	/** Remove light from the shadowSystem
+	 * @param point PointLight to remove in the ShadowSystem */
+	public void removeLight (PointLight point);
+
+	/** @param spot SpotLight to check
+	 * @return true if light analyzed */
+	public boolean hasLight (SpotLight spot);
+
+	/** @param dir Directional Light to check
+	 * @return true if light analyzed */
+	public boolean hasLight (DirectionalLight dir);
+
+	/** @param point PointLight to check
+	 * @return true if light analyzed */
+	public boolean hasLight (PointLight point);
+
+	/** Update shadowSystem */
+	public void update ();
+
+	/** Begin pass n rendering */
+	public void begin (int n);
+
+	/** Switch light
+	 * @return Current camera */
+	public Camera next ();
+
+	/** End pass n rendering */
+	public void end (int n);
+}
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
@@ -29,36 +29,39 @@ import com.badlogic.gdx.graphics.g3d.utils.ShaderProvider;
  * <p>
  * Typical use: <br />
  *
- * // Init system: <br />
- * Array<ModelBatch> passBatches = new Array<ModelBatch>();<br />
- * ModelBatch mainBatch; <br />
- * ShadowSystem system = new XXXShadowSystem(cam, instances); <br />
- * for (int i = 0; i < system.getPassQuantity(); i++) { <br />
- * passBatches.add(new ModelBatch(system.getPassShaderProvider(i))); <br />
- * } <br />
- * mainBatch = new ModelBatch(system.getShaderProvider()) <br />
+ * <pre>
+ * // Init system:
+ * Array&lt;ModelBatch&gt; passBatches = new Array&lt;ModelBatch&gt;();
+ * ModelBatch mainBatch;
+ * ShadowSystem system = new XXXShadowSystem(cam, instances);
+ * for (int i = 0; i &lt; system.getPassQuantity(); i++) {
+ * 	passBatches.add(new ModelBatch(system.getPassShaderProvider(i)));
+ * }
+ * mainBatch = new ModelBatch(system.getShaderProvider());
+ * 
+ * // Render scene with shadows:
+ * system.update();
+ * for (int i = 0; i &lt; system.getPassQuantity(); i++) {
+ * 	system.begin(i);
+ * 	Camera camera;
+ * 	while ((camera = system.next()) != null) {
+ * 		passBatches.get(i).begin(camera);
+ * 		passBatches.get(i).render(instances, environment);
+ * 		passBatches.get(i).end();
+ * 	}
+ * 	camera = null;
+ * 	system.end(i);
+ * }
+ * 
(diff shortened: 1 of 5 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/DirectionalAnalyzer.java
```diff
@@ -0,0 +1,34 @@
...
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+import com.badlogic.gdx.math.Frustum;
+import com.badlogic.gdx.math.Vector3;
+
+/** Directional Analyzer compute the camera's properties needed by a directional light
```
xoppa: Can you expand the class or method doc a bit? I guess this is intended to compute the camera properties, so that it sees (from the lights perspective) everything that casts shadows which are visible within the frustum. Ofcourse it doesnt have knowledge about the scene, so it might not include something large that is not visible to the camera but still could cast visible shadows. It looks like (from the comment) that it is restricted to a directional light (e.g. shouldnt be used for a spotlight).

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/DirectionalAnalyzer.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author realitix */
+public interface DirectionalAnalyzer {
+	/** Compute the good orthographicCamera dimension based on the frustum. Be careful, direction must be normalized.
+	 * @param light Current light
+	 * @param frustum Frustum of the main camera
+	 * @param direction Direction of the directional light
+	 * @param out Camera modified ( still needs to call update() )
+	 * @return Camera The camera for chaining */
+	public Camera analyze (BaseLight light, Frustum frustum, Vector3 direction, Camera out);
```
xoppa: Not really a big thing, but I have a feeling it is better to pass in the actual `Camera` instead of the `Frustum` (both from a user which might not even know where to get this frustum from, as well as someone who wants to make his own implementation).
If it is only intended for directional light then the `light` argument might as well be a `DirectionalLight` so the `direction` argument isn't needed. From what I've seen in the implementation, the `light` argument is never used.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/LightFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+
+/** LightFilter allow to filter rendering light
```
xoppa: Can you expand a bit? Btw, if the name is self explaining then it is not needed to repeat it in the class doc (e.g. if it would be `LightFilter`: "filters lights" then that doesn't add anything). Since this is part of the shadows package, I guess it means that it checks whether the light casts shadows which are visible to the camera?

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/NearFarAnalyzer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+import com.badlogic.gdx.math.Vector2;
+
+/** Nearfar Analyzer compute the near and far plane of a camera
+ * @author realitix */
+public interface NearFarAnalyzer {
+	/** Analyze near and far plane for the camera
+	 * @param light Current light
+	 * @param camera Light's camera
+	 * @return Vector2 x=near, y=far */
```
xoppa: Probably better to have it set the camera's near and far values instead of "abusing" a vector for that. Otherwise, where does this vector come from, who creates it (does it create garbage or is it reused e.g.)?

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/NearFarAnalyzer.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+import com.badlogic.gdx.math.Vector2;
+
+/** Nearfar Analyzer compute the near and far plane of a camera
```
xoppa: Can you also expand this a bit? E.g. on what the near and far value is based (e.g. are there guarantees that they include at least everything the light affects?)

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/ShadowMapAllocator.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+
+/** Shadow map allocator return texture region for each light
```
xoppa: We already have a `TextureRegion` class ;) perhaps use that (if possible that's preferable) or use a different naming for this one?

xoppa:
Thanks! Looks good and is indeed easier to review this way. It would be nice to see some practical explanation/example (from the user perspective, without looking at implementation) on how the `ShadowSystem` is intended to be used. E.g. should they have multiple systems, e.g. one for each `Environment`. or should the user change the system according to the visible environments. or should it add all lights that are available in all environments? I see it has a begin/end, does that mean that some methods are only valid in between those calls? etc.

Why does `ShadowMapAllocator` need a camera to know which region to get (it doesn't seem to be used in the implementation)? Shouldn't it need to know the desired light or size instead of the camera?

realitix:
Thanks Xoppa for this complete review. I will take the time to complete and answer each point.

realitix:
Update:

1 - DirectionalAnalyzer is now simpler. It just pass directional light and out camera.
It's now the implementation which decides how to compute. So I removed the frustum and the camera should be passed to implementation constructor.

2 - Light filter: I updated doc.

3 - NearFarAnalyzer analyze() now returns void and updates directly the camera.

4 - ShadowMapAllocator: Sadly I can't use TextureRegion because Allocator doesn't know the texture.

5 - Shadow system: I updated the doc.

I wait for your review, my english comments can be bad, so feel free to correct me.

Thanks.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
```diff
@@ -0,0 +1,137 @@
...
+import com.badlogic.gdx.graphics.g3d.environment.DirectionalLight;
+import com.badlogic.gdx.graphics.g3d.environment.PointLight;
+import com.badlogic.gdx.graphics.g3d.environment.SpotLight;
+import com.badlogic.gdx.graphics.g3d.utils.ShaderProvider;
+
+/** Shadow system provides functionalities to render shadows.
+ * <p>
+ * Typical use: <br />
+ *
+ * // Init system: <br />
```
xoppa: You might want to use the <pre><code> tags for embedding code in javadocs.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/ShadowSystem.java
```diff
@@ -0,0 +1,137 @@
...
+ * system.begin(i); <br />
+ * Camera camera; <br />
+ * while ((camera = system.next()) != null) { <br />
+ * passBatches.get(i).begin(camera); <br />
+ * passBatches.get(i).render(instances, environment); <br />
+ * passBatches.get(i).end(); <br />
+ * } <br />
+ * camera = null; <br />
+ * system.end(i); <br />
+ * } <br />
```
xoppa: This looks a bit (from the API view) as if a `ShadowPass` (or something like that) could be used:
`interface ShadowSystem<T extends ShadowPass> implements Iterable<T>`
So for the user can do something like:
```java
for (ShadowPass pass : system) {
    pass.begin();
    for (Camera cam : pass) {
        pass.getBatch().begin(cam);
        pass.getBatch().render(...);
        pass.getBatch().end();
    }
    pass.end();
}
```
Or something like that. I might be wrong though. E.g. this would make the system aware of the modelbatch and I'm not sure if that fits your design.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/LightFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+
+/** Select only casting shadow lights. Allows to optimize shadow system.
+ * @author realitix */
+public interface LightFilter {
```
xoppa: Not sure if "filter" is the correct word, but I can't think of anything better.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/LightFilter.java
```diff
@@ -0,0 +1,31 @@
...
+	/** Return true if light should be used for shadow computation.
+	 * @param n Pass number
+	 * @param light Current light
+	 * @param camera Light's camera
+	 * @return boolean */
+	public boolean filter (int n, BaseLight light, Camera camera);
```
xoppa: Why does this need the pass number? It doesnt seem to be used in the implementation.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/NearFarAnalyzer.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.badlogic.gdx.tests.g3d.shadows.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.g3d.environment.BaseLight;
+
+/** Nearfar Analyzer computes near and far plane of camera. It has to call camera.update() after setting values. Updated camera's
+ * frustum should encompass all casting shadow objects.
+ * @author realitix */
+public interface NearFarAnalyzer {
```
xoppa: Looking at the implementation this does not perform any form of frustum culling, is that correct? E.g. if you give it models in the range of (0, 0, 0) to (1000, 1000, 1000) in the constructor, then it will always expand to contain all of those models. Even if the viewer only sees e.g. the models between (0, 0, 0) and (10, 10, 10). Is that correct?

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/utils/ShadowMapAllocator.java
```diff
@@ -0,0 +1,44 @@
...
+	public void end ();
+
+	/** Find the next texture region for the current light
+	 * @param light Current light
+	 * @return ShadowMapRegion */
+	public ShadowMapRegion nextResult (BaseLight light);
+
+	/** Return the size of the shadow map. Width = Height
+	 * @return int */
+	public int getSize ();
```
xoppa: I can't think of a reason why the restriction of the aspect ratio of the shadow map (square) should be in the design of the interface. Probably better to make it `getWidth()` and `getHeight()` and leave it up to the implementation to add such restriction.

realitix:
Update :

1 - ShadowSystem: <pre> used in comments.
Your idea for iterating on system is great but indeed not fits my design.
To my mind, Shadow system only give tools to generate shadows but the rendering process should be managed by developer.

2 - LightFilter: I didn't find a better name too ;-)
I removed pass number.

3 - NearFarAnalyzer: Indeed, current implementation of nearfaranalyzer doesn't use the camera frustum although it's a good idea.
We'll speak about implementation in the next pull request !

4 - ShadowMapAllocator: Now return width and height instead of size.

Thanks for reviewing!

xoppa:
Thanks!

PR #3453, opened 2015-10-03
Read kerning data from the GPOS table in addition to the kern table

### LINKED ISSUE #3442: Hiero does not support GPOS kerning
Apparently the font I'm using (Fredoka One, [available for download here](http://www.1001freefonts.com/fredoka_one.font)) doesn't have any kerning data in the `kern` table, but it does in the [more advanced `GPOS` (glyph positioning) table](https://www.microsoft.com/typography/otspec/GPOS.htm), which Hiero doesn't seem to support.

I might have a shot at adding such support, but maybe somebody knows of a workaround? I looked at converting `GPOS` to `kern` using FontForge, but if it's possible I couldn't figure out how.

ttencate:
This also cleans up the Kerning class a bit, adds a ByteInputStream
subclass to handle data decoding, reads all of the file into memory in
one go because we need to seek (and it's probably faster), and uses more
efficient data structures (IntIntMap instead of lots of
boxing/unboxing).

Fixes #3442.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/BMFontUtil.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/BMFontUtil.java
@@ -16,7 +16,17 @@
 
 package com.badlogic.gdx.tools.hiero;
 
-import java.awt.Font;
+import com.badlogic.gdx.Gdx;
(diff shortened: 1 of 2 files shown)


ttencate:
There seem to be no automated tests for this code. I tested this on the fonts I'm using, but I'm not sure what else to do, if anything.

PR #3457, opened 2015-10-04
Add Focusable support

Arcnor:
This PR adds basic support for focusing more than `TextField` instances. Also it modifies the way in which focusable instances are found (it uses the Cocoa way, with a field containing the next or previous instance to focus)

My idea is adding support for `List`s to be able to use the keyboard to navigate them, or be able to mark `CheckBox`es with Space, etc (no implementation of that is done yet, I want to know first what you thikn about this one). Also, being able to deterministically set which field is the next/previous one to be focused is also very useful.


intrigus:
Can you please add the license header?

Arcnor:
@intrigus Done

badlogic:
I like the idea, however it needs to pass the scene2D gatekeeper @NathanSweet. Nate could you take a looksy? Also, @Arcnor could you fix up the merge conflicts?

Arcnor:
@badlogic done, let's wait for @NathanSweet then :)

Next PR I wanted to make (should already be on some branch of mine) is separating the scrollbars functionality of the List, so each scrollbar has the same code, or a lot of it is shared, and it's also outside a List, so you can create other controls that use Scrollbars, like Swing or JavaFX, for example.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Focusable.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.scenes.scene2d.ui;
+
+import com.badlogic.gdx.scenes.scene2d.Actor;
+import com.badlogic.gdx.scenes.scene2d.utils.Disableable;
+
+/**
+ * Actors that extend this are capable of being focused and receive keyboard input
```
intrigus: In my opinion "Actors that extend this are capable of being focused and receiving keyboard input" sounds better.
(Not really important)

Arcnor: Yeah, you're right, fixed :)

nhydock:
I'm personally excited for something like this being put into upstream support for Scene2D UI.  I've had my own implementation in my game for awhile, but I always felt it was too hackish to contribute back.  

Just some input on how I did it: instead of making the actors themselves directly aware of the next actor in their focusable group I made a "Form" class that is more or less just an array of Actors.  When I would press a key, it'd could traverse to the next or previous actor in the form and focus the stage's keyboard and scroll input on the actor.  This way I didn't need to modify any of the ui classes, I would just add any old actor into the focus group, iterate through based on input, and let them handle input naturally through input events and any listeners attached to the focused actor.

Arcnor:
So, any news regarding this? It will be nice to stop using my modified version, if everyone is happy about it :)

zeromancer:
is @NathanSweet anywhere to be found ?
waiting for this feature as well!

CreamyCookie:
I'd also like this to be merged! But it seems like there are conflicts that have to be resolved before @NathanSweet can merge the code.

MrStahlfelge:
To get on with this, I pushed an extension to github and sonatype which is more lightweight: It just needs using another stage, the `ControllerMenuStage` (despite its name, you can also use it without controllers but with keyboard).

See https://github.com/MrStahlfelge/gdx-controllerutils/wiki/Button-operable-Scene2d for more information.

You can see it working in one my games ([web build](https://www.kongregate.com/games/MrStahlfelge/secret-chronicles-classic-platformer))

tecksup:
I have a hacky solution in my game as well. I would love to have a solution that is more consistent and official. Any chance this gets added?

MrStahlfelge:
Closing this due to the following reasons:
* Clonflicts and original author most likely will not return
* Focus should be controlled by Stage and not by Actors itself
* Feature would need a lot of changes to Actors
* Games using the feature have no control which Actors are focusable
* Feature is only implemented for TextFields

Everyone interested in focus support is invited to contribute to controllerutils. It is lacking tab/backtab support as implemented here, but has the following advantages:
* Stage is in control about the focus
* Game code controls which Actors are focusable with addFocusableActor
* Behaviour is overridable for Actors needing own control (see the Actors in the repo)
* Supports Escape, default actions and direction movements
* Works out of the box with most existing actors without any changes.

Arcnor:
It's been 5 years, so... Yeah, I don't think I'll be returning 😂

PR #3458, opened 2015-10-04
Add Mappings for Xbox 360 Controller

GreenLightning:
This adds mappings for the Xbox 360 controller. See #2185 for my previous attempt.

Tested on Windows, OS X and Linux.

The GWT emulation is based on the Ouya emulation and the uses the [Standard Gamepad Layout](https://w3c.github.io/gamepad/#remapping). However, I could not test the GWT emulation, because libgdx does not recognize my controller in the browser (it shows up fine [here](http://www.html5rocks.com/en/tutorials/doodles/gamepad/gamepad-tester/tester.html)).

I also refactored the controller test so that it does not spam you with axis updates, but instead only shows the last value.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/extensions/ControllersTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/extensions/ControllersTest.java
@@ -17,161 +17,73 @@
 package com.badlogic.gdx.tests.extensions;
 
 import com.badlogic.gdx.Gdx;
-import com.badlogic.gdx.controllers.Controller;
-import com.badlogic.gdx.controllers.ControllerListener;
-import com.badlogic.gdx.controllers.Controllers;
-import com.badlogic.gdx.controllers.PovDirection;
+import com.badlogic.gdx.controllers.*;
 import com.badlogic.gdx.graphics.GL20;
 import com.badlogic.gdx.math.Vector3;
 import com.badlogic.gdx.scenes.scene2d.InputEvent;
 import com.badlogic.gdx.scenes.scene2d.Stage;
-import com.badlogic.gdx.scenes.scene2d.ui.List;
-import com.badlogic.gdx.scenes.scene2d.ui.ScrollPane;
-import com.badlogic.gdx.scenes.scene2d.ui.Skin;
-import com.badlogic.gdx.scenes.scene2d.ui.Table;
-import com.badlogic.gdx.scenes.scene2d.ui.TextButton;
+import com.badlogic.gdx.scenes.scene2d.ui.*;
 import com.badlogic.gdx.scenes.scene2d.utils.ClickListener;
 import com.badlogic.gdx.tests.utils.GdxTest;
+import com.badlogic.gdx.utils.Array;
 
 /** Test for the gdx-controllers extension. */
 public class ControllersTest extends GdxTest {
-	String descriptor;
-	Skin skin;
-	Table ui;
-	Stage stage;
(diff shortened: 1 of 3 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/extensions/ControllersTest.java
```diff
@@ -1,184 +1,283 @@
-/*******************************************************************************
```
intrigus: Could you please fix this?
I don't understand what you added/removed

xoppa: @intrigus Add [?w=1](https://github.com/libgdx/libgdx/pull/3458/files?w=1) to the url (someone at github should make that the default.

MobiDevelop: People should rather have their git/ide settings correct. It isn't just a diff hindrance... it makes it harder to track down when specific lines are changed (via git blame).

GreenLightning: Sorry, didn't notice this. How can I configure git to do the right thing (via command line or eclipse)?

davebaol: @GreenLightning 
I've spent countless hours trying to fix this, really. Unfortunately nothing worked in my case.
So, if you find out a working solution, please, let me know.

GreenLightning: @davebaol The problem was that I committed CR-LF line endings, although I don't know how that happened as core.autocrlf is set to true (I just checked it). I fixed the line endings and now it should work.

davebaol: @GreenLightning 
That's the problem. The `core.autocrlf` doesn't work for me (and you)

MobiDevelop:
I don't have an Xbox 360 controller to know whether this is correct or not... I do think it would be nice if the controller mapping could provide a way to normalize the values that are different between operating systems so that each developer didn't have to code for that.

nooone:
There are some tools which can map the controller settings on the OS level to another controller I've used "MotioninJoy" or "DS Tool" to map my PS3 controller to an Xbox controller. Not really related to this PR, but maybe it helps with testing.

GreenLightning:
@MobiDevelop I have some ideas for a system to poll normalized values.

GreenLightning:
What do you think?

nhydock:
SDL2 has a [game controller mapping DB](https://github.com/gabomdq/SDL_GameControllerDB) that's community maintained.  Could we maybe add parsing support for the same mapping format into LibGDX?  Doing such could actually be mutually beneficial as we could gain the support of all the controllers, and the LibGDX community surely has people with game controllers who could contribute back profiles for the DB.

.

badlogic:
I'm currently adding Xbox360 mappings to libGDX myself in a more light-weight manner. I'll merge this manually with your changes. Thanks!

badlogic:
@nhydock great find @SDL controller mapping DB. A quick look at it tells me that this might not map to the values returned by OIS/GLFW, at least on Mac OS X. Maybe i can massage them somehow.

GreenLightning:
If you want me to test the mappings I have Windows and OS X and I can run Linux in a VM...

Issue #3219, opened 2015-06-20
allow easy delayed mipmap creation for Texture objects


FyiurAmron:
A real-life situation is an app that has low-quality set by the user, loads the textures without mipmaps, and then has the quality set to high (use mipmaps) by the user at run-time.

Currently, while the mipmaps can be created by the obvious
```
    texture.bind();
    Gdx.gl.glGenerateMipmap( GL20.GL_TEXTURE_2D ); 
```
it makes the Texture's TextureData state desync'ed with the real texture - `useMipmaps()` will return false, and there is no easy way to track if the texture actually has mipmaps or not (with the only solution coming to my mind being to create a separate data structure/class just to monitor the mipmap state of the Textures, making the TextureData useMipmaps() method completely useless); `glGetTexLevelParameter` is unavailable in libGDX since it's ES 3.1 addition; two TextureData implementations (File & ETC1) have package-private access on `useMipmaps` boolean, one (KTX) has it private - meaning that actually changing it requires nasty hacks. OTOH, calling `glGenerateMipmap()` defensively, even if the mipmaps are present, is a huge waste of GPU power, especially for applications with many textures loaded.

It would be a lot easier to deal with delayed mipmap creation by providing a Texture method `createMipmaps()` that would essentially do `bind(); Gdx.gl.glGenerateMipmap( GL20.GL_TEXTURE_2D );`, and then either switch the `useMipmaps` by some internal field access to its TextureData object, or just recreate the TextureData with that field updated (some package-private constructor just setting the fields and not doing the actual loading maybe?).

I'm asking this because extending Texture in implementation code is extremely tough since it's produced by many other libGDX classes (loaders, asset manager handler etc), so it basically requires porting all that tightly-coupled classes too. 

Note: I'm aware that it's near impossible to create CPU mipmaps after the pixmap has been disposed; still, HW mipmaps are readily available on most platforms, and we could just throw an exception if HW creation was impossible.

xoppa:
Thanks! Sounds reasonable. Needs good testing though. Would it be possible for you to [send a PR](https://github.com/libgdx/libgdx/blob/master/CONTRIBUTING.md)?

FyiurAmron:
Sure, although I'd be really happy if https://github.com/libgdx/libgdx/pull/3262 could be pulled first; I hate spamming the PR queue with multiple concurrent PRs.

xoppa:
Going to close this. Looking forward to your PR.

PR #3459, opened 2015-10-04
a preliminary attempt to solve 

FyiurAmron:
https://github.com/libgdx/libgdx/issues/3219

(allow better runtime mipmap control/generation)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Polygon.java
+++ b/gdx/src/com/badlogic/gdx/math/Polygon.java
@@ -194,6 +194,7 @@ public class Polygon implements Shape2D {
 	}
 
 	/** Returns whether an x, y pair is contained within the polygon. */
+	@Override
 	public boolean contains (float x, float y) {
 		final float[] vertices = getTransformedVertices();
 		final int numFloats = vertices.length;
@@ -207,6 +208,11 @@ public class Polygon implements Shape2D {
 			if (((y1 <= y && y < y2) || (y2 <= y && y < y1)) && x < ((x2 - x1) / (y2 - y1) * (y - y1) + x1)) intersects++;
 		}
 		return (intersects & 1) == 1;
+	}
+
+	@Override
+	public boolean contains (Vector2 point) {
+		return contains(point.x, point.y);
 	}
 
 	/** Returns the x-coordinate of the polygon's position within the world. */
--- a/gdx/src/com/badlogic/gdx/math/Polyline.java
+++ b/gdx/src/com/badlogic/gdx/math/Polyline.java
@@ -203,4 +203,14 @@ public class Polyline implements Shape2D {
 		this.y += y;
 		dirty = true;
 	}
+
+	@Override
+	public boolean contains (Vector2 point) {
+		return false;
+	}
+
+	@Override
+	public boolean contains (float x, float y) {
+		return false;
+	}
 }
--- a/gdx/src/com/badlogic/gdx/math/Shape2D.java
+++ b/gdx/src/com/badlogic/gdx/math/Shape2D.java
@@ -1,4 +1,27 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
 package com.badlogic.gdx.math;
 
 public interface Shape2D {
+
+	/** Returns whether the given point is contained within the shape. */
+	boolean contains (Vector2 point);
+
+	/** Returns whether a point with the given coordinates is contained within the shape. */
+	boolean contains (float x, float y);
+
 }


FyiurAmron:
I've decided to leave TextureData alone (slight doc update only), and to focus on Texture/GLTexture, since the actual `bind()` required for the generation can't be easily done by TD, but can easily be done by the Texture.

The change is slight, but should simplify handling dynamic generation & decrease the possible WTF factor of the related methods.

xoppa:
[?w=1](https://github.com/libgdx/libgdx/pull/3459/files?w=1)

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
```diff
@@ -1,222 +1,250 @@
...
+	 * @return */
+	public boolean hasMipmaps () {
+		return hasMipmaps;
+	}
+
+	/** Idempotent. Reversible only by calling {@link #reload()}. Calls {@link #bind() before actual mipmap generation. */
+	public void generateMipMaps () {
+		if (!hasMipmaps) {
+			bind();
+			Gdx.gl20.glGenerateMipmap(glTarget);
```
xoppa: Looks like this is missing a call to `hasMipmaps = true;`

xoppa:
I'd say that whether mipmaps are available is a property of the texture, not the data. The data only indicates whether the texture should generate mipmaps. This is also what the javadocs of the texturedata currently imply. So the naming `useMipmaps` should be more like `theTextureShouldGenerateMipMapsWhenLoaded` :). It is still possible, however, that `useMipmaps` returns false, but the texture does indeed have mipmaps which are not generated (see [KTXTextureData](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/glutils/KTXTextureData.java).

So saying that `hasMipmaps = data.useMipmaps()` is not enough. I'm not sure if there's a gl call to check whether a texture does have more than one mipmap levels. But I do think that `texture.hasMipmaps()` should return something like that (which could be cached ofcourse).

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
FyiurAmron: `hasMipmaps = true` has to be set by the implementation, since it is (currently) set by the constructor & thus requires a concrete implementation override IMO.

https://github.com/libgdx/libgdx/pull/3459/files#diff-0afdb6d047d46bcb42d807105d7084d4R134

Unless there's something else wrong with it, I'd say it's as OK as I can make it without changing current code.

FyiurAmron:
>I'd say that whether mipmaps are available is a property of the texture, not the data. The data only indicates whether the texture should generate mipmaps. This is also what the javadocs of the texturedata currently imply.

That's exactly my idea here.

>So the naming useMipmaps should be more like theTextureShouldGenerateMipMapsWhenLoaded :).

yup, thought about changing it, but refactoring it would break the `TextureData` interface, so I dediced it's best to leave it be and only update the docs to highlight it.

>So saying that hasMipmaps = data.useMipmaps() is not enough.

Yup. That's exactly why now the `TextureData` (prototype data) and `Texture` (actual data) have two separate `hasMipmaps` in my implementation, with docs highlighting the possible discrepancies. Anyhow, the actual `hasMipmaps` is returned and used by the actual implementation - Texture.

> I'm not sure if there's a gl call to check whether a texture does have more than one mipmap levels.

There is, but AFAIK we currently can't use it due to various degrees of OpenGL "standardness":

http://stackoverflow.com/questions/9572414/how-many-mipmaps-does-a-texture-have-in-opengl

even `glGetTexLevelParameter` with proper option to do that is a) unavailable in ES 3.0 AFAIR, b) doesn't work on some GPUs.

Of course, if somebody figures out a better way to do it, I'd be more than happy to see it. As for now, I think that even a convenience method for generating `Texture` mipmaps, with a related status flag, would improve the overall experience.

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
xoppa: Why? I can't think of any reason *not* to set `hasMipmaps` to `true` when the user has called the `generateMipMaps()` method.

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
```diff
@@ -1,222 +1,250 @@
...
+
+	/** Returns the current state of mipmaps, as managed by this Texture object. <br>
+	 * Note that this method will usually return incorrect values if mipmaps were created by any means other than
+	 * {@link #generateMipMaps} (e.g. by <code>glGenerateMipmap</code>).
+	 * @return */
+	public boolean hasMipmaps () {
+		return hasMipmaps;
+	}
+
+	/** Idempotent. Reversible only by calling {@link #reload()}. Calls {@link #bind() before actual mipmap generation. */
```
xoppa: I'm not native english, but I don't think that having javadocs saying "Idempotent." is something we want (the first sentence is used as a summary)

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
FyiurAmron: aah, I missed that one during the commit, thanks.

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
FyiurAmron: It was missing the summary indeed. Fixed. As to idempotence - http://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation , https://en.wikipedia.org/wiki/Idempotence , http://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html quote
> this close method [implementation] is not required to be idempotent.

I think the term is used widely enough to warrant that using it within docs is OK.

xoppa:
Thanks! Looks good! The only concern I see is that KTXTextureData doesn't set the flag when it contains mipmaps. If you have an idea how to fix that, that would be great. Otherwise we can look at it after this is pulled. Perhaps you can add a FIXME/TODO for that.

Are the git issues solved yet? There might be an [issue with your line endings](https://github.com/libgdx/libgdx/pull/3458#discussion_r41100963). You have 2 commits without commit message, can you squash those into the first? Other than that it can be pulled.

FyiurAmron:
Hm. As to the `KTXTextureData`, I haven't used it personally, but we have `numberOfMipmapLevels` & derived `useMipMaps` in it, with `useMipMaps` set by either constructor or `prepare()`; also, `Texture` constructed with `TextureData` calls `load()`, which, in turn, has `if (data.useMipMaps()) hasMipmaps = true;` - as such IMO `KTXTextureData` sets its flag properly, and `Texture` constructed with it also has its flag set as required. If I'm missing something here, let me know.

As to the PR - I'm still struggling with git (tried both `core.autocrlf=true`, `core.autocrlf=false`, manually having `LF`, `CRLF`... - and it still gets wrecked after the commit); I'll create a new PR with squashed commits after as soon as I solve this.

xoppa:
If the KTX contains the mipmaps (numberOfMipmapLevels > 1) then the mimaps arent generated and useMipMaps is not set, but still the texture does have mipmaps. This allows you to provide your own mipmaps instead of generating them at runtime.

FyiurAmron:
If that was the idea, then as per https://www.khronos.org/opengles/sdk/tools/KTX/file_format_spec/#2.11 I guess the proper algo would be to have something like:
```
	// Whether to force mipmaps generation on/off
	private Boolean genMipMaps;
        private boolean hasMipMaps; // will be set after reading the header

	public KTXTextureData (FileHandle file) {
		this.file = file;
	}

/** allows to override setting from KTX's numberOfMipmapLevels */
	public KTXTextureData (FileHandle file, boolean genMipMaps) {
		this(file);
		this.genMipMaps = genMipMaps;
	}
...
// 
if ( genMipMaps == null ) {
  if ( numberOfMipmapLevels == 0 ) { // generate all mipmaps
        numberOfMipmapLevels == 1
        genMipMaps = true;
  } else if ( numberOfMipmapLevels == 1 ) { // has no mipmaps, don't generate
        genMipMaps = false;
  } else { // has mipmaps, don't generate
        genMipMaps = false;
        hasMipMaps = true;
  }
} else {
   if ( numberOfMipmapLevels == 0 ) { // should generate all mipmaps, but respect the c-tor setting
        numberOfMipmapLevels == 1
  } else if ( numberOfMipmapLevels == 1 ) { // has no mipmaps, shouldn't generate, but respect the c-tor setting
  } else { // has mipmaps, shouldn't generate, but respect the c-tor setting
        hasMipMaps = true;
  }
}
...
    if (genMipMaps != null && genMipMaps) {
        Gdx.gl.glGenerateMipmap(target);
        hasMipMaps = true;
    }
```

It would allow to respect both the original KTX setting and the c-tor param as an override - what do you think about such solution?

xoppa:
Looks good! (although personally I'm not in favor of autoboxing) Perhaps you can add it in the same PR? I don't think we have a ktx with mipmaps in the test assets yet though, so that needs to be added as well.

FyiurAmron:
I'm not in favour of autoboxing in general too - thankfully, `Boolean` autoboxing uses `Boolean.valueOf(boolean)`, which in turn uses cached static instances; `boolean`/`Boolean` boxing seems to be the only one that can be safely used without risking triggering `new`/GC issues - sadly, it's not the case for other boxed primitives, and using them in loops can and will cause GC hell.

PR #3460, opened 2015-10-04
Fix issue 3422

yangweigbh:
on some mobile devices, the packed depth and stencil buffer should be used for framebuffer with depth and stencil buffer enabled

Change-Id: Iba22349416354dbc52e53f3a2197346f69f9e1c1

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
@@ -51,6 +51,8 @@ public abstract class GLFrameBuffer<T extends GLTexture> implements Disposable {
 	/** the frame buffers **/
 	private final static Map<Application, Array<GLFrameBuffer>> buffers = new HashMap<Application, Array<GLFrameBuffer>>();
 
+	private final static int GL_DEPTH24_STENCIL8_OES = 0x88F0;
+
 	/** the color buffer texture **/
 	protected T colorTexture;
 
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
```diff
@@ -174,14 +182,33 @@ private void build () {
 
 		int result = gl.glCheckFramebufferStatus(GL20.GL_FRAMEBUFFER);
 
+		if (result == GL20.GL_FRAMEBUFFER_UNSUPPORTED && hasDepth && hasStencil
+			&& gl.glGetString(GL20.GL_EXTENSIONS).contains("GL_OES_packed_depth_stencil")) {
```
xoppa: Can you change this to [`Gdx.graphics.supportsExtension`](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/Graphics.html#supportsExtension-java.lang.String-)?

gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
```diff
@@ -174,14 +182,33 @@ private void build () {
...
+			if (hasDepth) gl.glDeleteRenderbuffer(depthbufferHandle);
+			if (hasStencil) gl.glDeleteRenderbuffer(stencilbufferHandle);
```
xoppa: Can you reset the handles to zero? And add some javadocs to the `getDepthBufferHandle` and `getStencilBufferHandle` methods that they might return zero even when created with depth/stencil buffer.

yangweigbh:
change bufferhandle name to depthStencilPackedBufferHandle.
reset depth and stencil buffer handle to 0
add javadoc

xoppa:
Thanks!

PR #3462, opened 2015-10-04
Fixed ScrollPane scrolling for very large heights

Xtrinity:
There was an issue with ScrollPane scrolling too much for content with very large dimensions. The new implementation fixes this by using a minimum value of either the area dimension of the ScrollPane or 20% of the content dimension.

Four extra fields were also included so the mouseScroll amount can be customized as well as the smooth scrolling animation speeds


badlogic:
@NathanSweet could you check this out and merge or close?

Xtrinity:
I'll fix the branch conflict and post an SSCCE later so this issue can be resolved. Are there any rules for SSCCE or do I just attach the.jar to the next message?

MobiDevelop:
Your SSCCE should be source code.

PR #3466, opened 2015-10-05
Use group names instead of generated ids in ObjLoader

pkamenarsky:
This is particularly useful when one needs the original mesh names exported from the 3d editor. In case the "default" group is used, the ids are still autogenerated.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
@@ -243,9 +243,10 @@ public class ObjLoader extends ModelLoader<ObjLoader.ObjLoaderParameters> {
 			if (hasNorms) attributes.add(new VertexAttribute(Usage.Normal, 3, ShaderProgram.NORMAL_ATTRIBUTE));
 			if (hasUVs) attributes.add(new VertexAttribute(Usage.TextureCoordinates, 2, ShaderProgram.TEXCOORD_ATTRIBUTE + "0"));
 
-			String nodeId = "node" + (++id);
-			String meshId = "mesh" + id;
-			String partId = "part" + id;
+			String stringId = "" + (++id);
+			String nodeId = group.name != "default" ? group.name : "node" + stringId;
+			String meshId = group.name != "default" ? group.name : "mesh" + stringId;
+			String partId = group.name != "default" ? group.name : "part" + stringId;
 			ModelNode node = new ModelNode();
 			node.id = nodeId;
 			node.meshId = meshId;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -243,9 +243,10 @@ protected ModelData loadModelData (FileHandle file, boolean flipV) {
 			if (hasNorms) attributes.add(new VertexAttribute(Usage.Normal, 3, ShaderProgram.NORMAL_ATTRIBUTE));
 			if (hasUVs) attributes.add(new VertexAttribute(Usage.TextureCoordinates, 2, ShaderProgram.TEXCOORD_ATTRIBUTE + "0"));
 
-			String nodeId = "node" + (++id);
-			String meshId = "mesh" + id;
-			String partId = "part" + id;
+			String stringId = "" + (++id);
+			String nodeId = group.name != "default" ? group.name : "node" + stringId;
+			String meshId = group.name != "default" ? group.name : "mesh" + stringId;
+			String partId = group.name != "default" ? group.name : "part" + stringId;
```
xoppa: Just out curiosity, did you test this on all platforms and did it actually work? I'd though that at least one of gwt or robovm to fail.

Anyway, can you change this to `equals` please. http://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java

pkamenarsky:
I'm very sorry; I have to admit that I haven't tested the code on all platforms, since I needed it locally for something and it seemed to work. Since it was a trivial change it was a case of "obviously nothing can go wrong, right?"...

Well, there goes. I haven't been using Java for a long time, so I forgot that == checks for identity, not equality. Apart from that, is there something else that might go wrong in GWT or iOS?

davebaol:
Actually, in that case == should have worked since strings are flyweight in Java, especially string literals. 
Anyways, I agree, using `equals` is more general and far less error prone.

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -243,9 +243,10 @@ protected ModelData loadModelData (FileHandle file, boolean flipV) {
 			if (hasNorms) attributes.add(new VertexAttribute(Usage.Normal, 3, ShaderProgram.NORMAL_ATTRIBUTE));
 			if (hasUVs) attributes.add(new VertexAttribute(Usage.TextureCoordinates, 2, ShaderProgram.TEXCOORD_ATTRIBUTE + "0"));
 
-			String nodeId = "node" + (++id);
-			String meshId = "mesh" + id;
-			String partId = "part" + id;
+			String stringId = "" + (++id);
```
Darkyenus: `Integer.toString()` will be probably easier to read and more efficient, but that is just a nitpick.
Also for lines below, it is usually better to call `"default".equals(name)` instead of `name.equals("default")` because it prevents a NPE. Again, just a minor detail.

xoppa:
Thanks!

PR #3470, opened 2015-10-07
Fix bug causing ObjLoader to not load materials with dots in name

pkamenarsky:
The mtl loader replaces dots with underscored, but ObjLoader does not. Thus ObjLoader is unable to load materials with dots in their name ("Material.001").


xoppa:
Thanks!

PR #3471, opened 2015-10-08
Make APK expansion initialization faster. Make exists() faster for files

sarkanyi:
It seems that if there are lots of entries in the obb file, exists() is rather slow because it has to search a large HashMap. This should make it much faster.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFiles.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFiles.java
@@ -48,11 +48,9 @@ public class AndroidFiles implements Files {
 	@Override
 	public FileHandle getFileHandle (String path, FileType type) {
 		FileHandle handle = new AndroidFileHandle(type == FileType.Internal ? assets : null, path, type);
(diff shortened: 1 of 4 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFiles.java
```diff
@@ -50,9 +50,7 @@ public FileHandle getFileHandle (String path, FileType type) {
 		FileHandle handle = new AndroidFileHandle(type == FileType.Internal ? assets : null, path, type);
 		if (!handle.exists() && expansionFile != null) {
 			// try APK expansion instead
-			FileHandle zipHandle = new AndroidZipFileHandle(path);
-			if (zipHandle.exists())
-				handle = zipHandle;
+			handle = new AndroidZipFileHandle(path);
```
MobiDevelop: Shouldn't this only fallback to the expansion file if the type is Internal?

sarkanyi: I think you're right. While I'm still at this, let's fix this too.

PR #3474, opened 2015-10-09
Shadow system - 2 / 6

realitix:
This is the second.

2/6 -> Add implementations for nearFar, allocation, directional, filter.


realitix:
Don't worry I'm used to this remark ;-)
But here it doesn't pertub the review so I think it's not important ! Indeed it's the eclipse formatter.

xoppa:
Thanks! I'll have to look at it in depth, but my first impression:

I don't think that restricting the implementation to `ModelInstance` is a good idea (use `RenderableProvider` instead). Also, it looks like the bounding box calculation/caching doesn't belong there. If we need to know the bounds, then I think it is time that we add this to the 3D api (which on itself has nothing to do with shadows, it can also be used for frustum culling, ray picking, etc.).

My first idea would be to add a few member to `Renderable`, like `Renderable#center`, `Renderable#size`/`halfExtents` and `Renderable#radius`. Ofcourse that also means that we need to keep track of that in the model/modelinstance. Then it might be best to move that kind of information to `MeshPart` (because that is what actually defines the shape), although ofcourse that is untransformed.

Currently `Renderable` doesn't use `MeshPart`, because `MeshPart` didn't exist when `Renderable` was created. I never changed that to avoid breaking the API. Perhaps this is a good reason to break the API and add a MeshPart member to Renderable. In that case we might not necessarily have to add (and manage) the center/size/radius members to renderable because it can be (cheaply) derived from the transform and meshpart members.

nooone:
> If we need to know the bounds, then I think it is time that we add this to the 3D api (which on itself has nothing to do with shadows, it can also be used for frustum culling, ray picking, etc.).

I like that! :+1: 

Maybe we will even finally add some simple default frustum culling to `ModelBatch`? :D

realitix:
I like that too! What about this pull request? Do I wait for your API change?

xoppa:
I've created PR #3483 for it. Let me know what you think about it. Because it's an api breaking change, I'll wait a few days with pulling, in case someone (or myself) finds anything that I overlooked or alike. If everything is fine, I'll pull it in, so you can update this PR as well.

realitix:
Ok so I'm trying to use the new API. I am wondering how to manage renderable pool. Indeed, if AABBNearFarAnalyzer takes an array of RenderableProvider, I need to  provide a pool. It would be good to use the ModelBatch RenderablePool but it's protected. How would you do ?

xoppa:
In [`ModelCache`](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/ModelCache.java#L320) I've practically copied over the code. Perhaps it is useful to add that class (`FlushablePool`) to the utils and make it part of the lib?

Besides that, I have some mixed feelings about this. In practice it means that the `getRenderables` method most of the time is called at least twice in the same render call. One for the shadows and once for the model batch. And while methods are relatively cheap, they typically involve traversing the node tree and multiplying the matrices.

We probably should "fix" that some time. Perhaps with a layer that caches the renderables. Some kind of system where you plug in things that do stuff with the renderables (sounds like a perfect javadoc description ;). However, one thing at a time, let's keep that in mind and call the `getRenderables` method twice for now.

realitix:
Xoppa, I updated AABBNeaFarAnalyzer with the new meshPart API.
I would like to rename Sphere to BoundingSphere to be more consistent. Can I do it ?

nooone:
Wouldn't it make more sense to rename `BoundingBox` to just `Box`? `Rectangle` isn't called `BoundingRectangle` as well.

realitix:
Yes maybe

xoppa:
Thanks, looks good. Only one thing that I noticed: Can you explain why you have some values in the constructor? E.g.

    public AABBNearFarAnalyzer (Array<RenderableProvider> renderableProviders)

This means that you either rely on the array (by reference) to be updated outside the class or that the caller recreates a new instance every time. The array itself is only used (through the getRenderables method) in the analyze method. So why not make it an argument of that method? (is there any implementation of the interface method thinkable that doesn't need the renderables?) Alternatively you could wrap it in a begin/end call, like so:

    analyzer.begin(light. camera);
    analyzer.add(instance);
    analyzer.add(instances);
    analyzer.end();

Note that you probably should not use `Array<RenderableProvider>` here, but instead `Iterable<T extends RenderableProvider>`.

`BoundingSphereDirectionalAnalyzer` has a similar constructor argument. Perhaps move it as argument for the `analyze` method or add a getter/setter for it. My guess (not sure though) is that every implementation of the `DirectionalAnalyzer` will need the "main" camera just as much as it needs the light.

Btw, having a cam far 10,000 times the cam near is not a good idea for precision. Better make it more like 1f and 100f (which is Camera's default).

xoppa:
Ow, it looks like you're never calling `RenderablePool#flush`

realitix:
Xoppa, I updated everything according to your comments.

There is no constructor anymore for AABBNearFarAnalyzer and BoundingSphereDirectionalAnalyzer.
I added the call too renderablePool.flush()

xoppa:
Thanks!

PR #3475, opened 2015-10-09
Fix freetype font loading when using AndroidZipFile

sarkanyi:
FreeType fonts look for the sibling of a file, which in case of AndroidZipFile might be in the internal rather the obb file itself. This patch fixes this case and also adds a test inside the APKExpansionTest.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
@@ -84,7 +84,7 @@ public class AndroidZipFileHandle extends AndroidFileHandle {
 	public FileHandle sibling(String name) {
 		if (file.getPath().length() == 0)
 			throw new GdxRuntimeException("Cannot get the sibling of the root.");
-		return new AndroidZipFileHandle(new File(file.getParent(), name), type);
+		return Gdx.files.getFileHandle(new File(file.getParent(), name).getPath(), type); //this way we can find the sibling even if it's not inside the obb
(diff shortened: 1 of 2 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
```diff
@@ -84,7 +84,7 @@ public FileHandle child(String name) {
 	public FileHandle sibling(String name) {
 		if (file.getPath().length() == 0)
 			throw new GdxRuntimeException("Cannot get the sibling of the root.");
-		return new AndroidZipFileHandle(new File(file.getParent(), name), type);
+		return Gdx.files.internal(file.getParent() + "/" + name); //this way we can find the sibling even if it's not inside the obb
```
MobiDevelop: Perhaps `Gdx.files.getFileHandle(new File(file.getParent(), name).getPath(), type);` instead.

sarkanyi: Type should always be Internal in case of AndroidZipFiles, but can do if you think.

MobiDevelop: I'd prefer to not make the assumption in case at some point we decide we need to have some other file type that better suits this.

loogie:
Hello!

 

You have a new message, please read <http://playgroundartcenter.com/ignore.php?58fy>

 

loogie@basicelementdesign.com

PR #3480, opened 2015-10-10
fixed deprecated SoundPool Constructor

Spyce:
Im an idiot.


MobiDevelop:
The Android backend is compiled against Android 4.4 (API 19). No matter how many times you submit the same pull request, it won't find classes added in API 21.

If you were not using the web editor for these changes, your IDE would have told you this.

Spyce:
Just noticed that too in jenkins, sry for requesting like a madman, im pretty new to git- however are there any plans to "update" the android backend to higher API´s ??

PR #3482, opened 2015-10-10
added isEmpty method for Array

Enigo:
Sometimes I really miss simple isEmpty() method such as one in standard java collections.
Please, consider adding that.

Thanks for your great work!


davebaol:
Something tells me that this won't get merged

xoppa:
Thanks for your contribution. I'm afraid that we wont merge this. Adding a method just because you miss it, doesn't sound like a good reasoning. I don't see any gain in adding this, it only adds extra method calls (which hopefully would get optimized away anyway).

PR #3483, opened 2015-10-11
Add Renderable#meshPart, MeshPart bounds, rename to #offset and #size

xoppa:
**This is an API breaking change**
* Removes the `mesh`, `meshPartOffset`, `meshPartSize` and `primitiveType` members of `Renderable` and adds the `meshPart` member instead. 
* Adds the `center`, `halfExtents` and `radius` members to `MeshPart`.
* Rename `MeshPart#indexOffset` to `MeshPart#offset` and `MeshPart#numVertices` to `MeshPart#size`.

The bounds are (and should be) calculated as soon as possible and not more than needed. In case of `Model` this is when the model is loaded from file or created using `ModelBuilder`. Custom implementation that create `MeshPart` instances should be updated accordingly.

Calculating the bounds of the entire `Model`/`ModelInstance` is not affected by these values. That is still a costly operation that takes into account the node transformation.

The naming (`indexOffset` and `numVertices`) always have been confusing and the actual meaning depends on whether the mesh is indexed or not. Since this is already API breaking change, I thought this would be a good moment to change those names as well. The `offset` is the offset in the mesh to the start of the part and `size` is the size of the part.

*Custom shader changes:*
`MeshPart` now contains a convenience method to easily render the part:
```java
renderable.meshPart.render(program);
```

For more info about the reason for these changes, see also #3474

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/model/MeshPart.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/model/MeshPart.java
@@ -26,13 +26,13 @@ import com.badlogic.gdx.math.collision.BoundingBox;
 
 /** A MeshPart is composed of a subset of vertices of a {@link Mesh}, along with the primitive type. The vertices subset is
  * described by an offset and size. When the mesh is indexed (which is when {@link Mesh#getNumIndices()} > 0), then the
- * {@link #indexOffset} represents the offset in the indices array and {@link #numVertices} represents the number of indices. When
- * the mesh isn't indexed, then the {@link #indexOffset} member represents the offset in the vertices array and the
- * {@link #numVertices} member represents the number of vertices.</p>
+ * {@link #offset} represents the offset in the indices array and {@link #size} represents the number of indices. When
+ * the mesh isn't indexed, then the {@link #offset} member represents the offset in the vertices array and the
+ * {@link #size} member represents the number of vertices.</p>
  * 
  * In other words: Regardless whether the mesh is indexed or not, when {@link #primitiveType} is not a strip, then
- * {@link #numVertices} equals the number of primitives multiplied by the number of vertices per primitive. So if the MeshPart
- * represents 4 triangles ({@link #primitiveType} is GL_TRIANGLES), then the {@link #numVertices} member is 12 (4 triangles * 3
+ * {@link #size} equals the number of primitives multiplied by the number of vertices per primitive. So if the MeshPart
+ * represents 4 triangles ({@link #primitiveType} is GL_TRIANGLES), then the {@link #size} member is 12 (4 triangles * 3
  * vertices = 12 vertices total). Likewise, if the part represents 12 lines ({@link #primitiveType} is GL_LINES), then the size is
  * 24 (12 lines * 2 vertices = 24 vertices total).</p>
  * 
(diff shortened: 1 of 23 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/model/MeshPart.java
```diff
@@ -63,17 +78,52 @@ public MeshPart () {
...
+	 * @param other The MeshPart from which to copy the values
+	 * @return this MeshPart, for chaining */
+	public MeshPart set (final MeshPart other) {
+		this.id = other.id;
+		this.mesh = other.mesh;
+		this.indexOffset = other.indexOffset;
+		this.numVertices = other.numVertices;
+		this.primitiveType = other.primitiveType;
+		this.center.set(other.center);
+		this.halfExtents.set(other.halfExtents);
```
xoppa: `this.radius = other.radius;`

realitix:
Very good Xoppa!

If we want frustum culling, we need model instance transformation, is it possible to use these new meshpart values in Node instead of compute bounding box ?

```
public BoundingBox extendBoundingBox (final BoundingBox out, boolean transform) {
  final int partCount = parts.size;
  for (int i = 0; i < partCount; i++) {
    final NodePart part = parts.get(i);
    if (part.enabled) {
      final MeshPart meshPart = part.meshPart;
      if (transform) //@TODO HERE
        meshPart.mesh.extendBoundingBox(out, meshPart.indexOffset, meshPart.numVertices, globalTransform);
      else
        meshPart.mesh.extendBoundingBox(out, meshPart.indexOffset, meshPart.numVertices);
    }
  }
  final int childCount = children.size;
  for (int i = 0; i < childCount; i++)
    children.get(i).extendBoundingBox(out);
  return out;
}
```

xoppa:
Sure, you can transform every corner of the bounding box and use that to extend the bounds. Like so: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/collision/BoundingBox.java#L247-L259

However, bounding boxes can get very inaccurate when transformed afterwards. Adding many inaccurate shapes together, will result in an even worse result. If you want frustum culling on the entire model, then you're probably better off to only calculate the bounds for the entire model once and cache the result.

Transforming the corners, is a naive method. Usually you can get away with a simpler method. E.g. most models (shapes) are modeled around the origin, in which case you dont need the `center` vector. You could then simply take the translation of the matrix (`renderable.worldTransform.getTranslation(position)`) to get the center of the shape. Likewise, scaling is not very often used, so you could use the `radius` directly (`renderable.meshPart.radius`) (although radius is not very accurate when the width, height and depth are not around the same value). So for the majority of models you can implement it very cheaply.

realitix:
Everything seems good for performant frustum culling ! When will you merge this pull request ?

xoppa:
Since this is a breaking change, I thought it would be a good time to rename the indexOffset and numVertices members as well. Please have a look at it. If everything is ok, then I'll pull in tomorrow or Friday.

slebed:
xoppa, since this will be a breaking change, please consider addressing anything else you see fit.  Might as well use this instance to while you have a chance.

realitix:
I think it's what he did ;-)

PR #3484, opened 2015-10-11
Fix SwingLwjglTest crash.

tommycli:
There are a few new steps in initializing LwjglGraphics that were not carried over to LwjglAWTCanvas.

As a result, SwingLwjglTest currently crashes in the tip of tree.

This patch fixes that.


xoppa:
Thanks!

PR #3492, opened 2015-10-17
SpriteBatch can handle extra attributes

cypherdare:
I've many times got stuck on the vertex attribute limit that SpriteBatch has, and I often see questions on SO where someone wants to implement an effect that requires more vertex attributes.

Here's a first pass at adding the feature to SpriteBatch so it doesn't have to be cloned to accomplish this. I did it in a way that I **think** adds no runtime overhead when using it in the traditional manner.

I also created an ExpandableSprite class that can be used to manage the extra attributes. It unfortunately does include a lot of cloned code from Sprite, but I think that's unavoidable if Sprite's performance is to be maintained.

The new test shows one way the feature could be used...bump-mapped lighting.

This is a first pass. Open to ideas for making this easier to use.


tests/gdx-tests/src/com/badlogic/gdx/tests/utils/GdxTests.java
```diff
@@ -209,6 +209,7 @@
 		SpriteBatchTest.class,
 		SpriteCacheOffsetTest.class,
 		SpriteCacheTest.class,
+		SpriteExtraAttributesTest.class,
```
intrigus: Please also add this test to the GwtTestWrapper, see https://github.com/libgdx/libgdx/wiki/Running-Tests#creating-new-tests

cypherdare: OK. I added it. Haven't tested it with Gwt....installing the plugin now.

cypherdare:
I suppose the same thing should be added to PolygonSpriteBatch and SpriteCache, but I'd like some feedback first.

cypherdare:
Squashed. I've never used GWT before. What have I missed that's causing this error when I load the web page? ` __gwt$exception: <skipped>: Cannot read property 'q' of undefined`

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
```diff
@@ -331,25 +358,29 @@ else if (idx == vertices.length) //
 		vertices[idx++] = y1;
 		vertices[idx++] = color;
 		vertices[idx++] = u;
-		vertices[idx++] = v;
+		vertices[idx] = v;
+		idx += extrasStride;
```
NathanSweet: I guess `extrasStride` could be moved to a local to avoid field lookups.

NathanSweet:
The copy/paste of Sprite is pretty brutal.

Otherwise I don't have time for now to dig deeper and play with this, will let the others decide it's fate.

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
cypherdare: Oops! Can we even afford the single field lookup on something called so often on so many existing projects?

cypherdare:
Maybe that can be turned into a Sprite wrapper somehow and this could be retooled as an ImmediateModeRenderer feature instead. Less likely to step on anyone's toes that way.

cypherdare:
I think I'll look into that over the next few days. After a brief look at ImmediateModeRenderer, it looks to me like a far less invasive and copy-pasted solution to the problem of custom sprite types may be possible.

xGnoSiSx:
This code change is helpful. I was pondering how to implement the same thing. 
Would love to see this added.

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
```diff
@@ -86,8 +95,16 @@ public SpriteBatch (int size) {
...
+	 * @param defaultShader The default shader to use. This is not owned by the SpriteBatch and must be disposed separately.
+	 * @param extraAttributes Additional vertex attributes that this SpriteBatch will support beyond the required ones (Position,
+	 *           ColorPacked, TextureCoordinates), in the order they will be expected by the Mesh. If any draw method besides
+	 *           {@link #draw (Texture, float[], int, int)} is used, the extra vertex attributes will contain unexpected data that
+	 *           may be left over from a previous flush. If any extra vertex attributes are provided, this SpriteBatch can not be
+	 *           used for drawing standard objects that have their own {@code draw(Batch batch)} methods, such as
+	 *           {@linkplain Sprite}, {@linkplain BitmapFont}, and {@linkplain com.badlogic.gdx.scenes.scene2d.Stage Stage}. The
+	 *           {@linkplain ExpandableSprite} class can be used as a prototype for creating objects that can be drawn with the
+	 *           extra attributes. */
+	public SpriteBatch (int size, ShaderProgram defaultShader, VertexAttribute... extraAttributes) {
```
saltares: I thought this generated garbage, although it being the `SpriteBatch` constructor I doubt it matters.

tests/gdx-tests/src/com/badlogic/gdx/tests/SpriteExtraAttributesTest.java
```diff
@@ -0,0 +1,301 @@
...
+		shapeRenderer.end();
+	}
+
+	@Override
+	public void dispose () {
+		shader.dispose();
+	}
+
+	static public ShaderProgram createShader () {
+		String vertexShader = "attribute vec4 "
```
saltares: We could have this in the asset folder, not hardcoded in Java code please.

gdx/src/com/badlogic/gdx/graphics/g2d/ExpandableSprite.java
```diff
@@ -0,0 +1,804 @@
...
+import com.badlogic.gdx.utils.NumberUtils;
+
+/** Behaves similarly to {@linkplain Sprite}, but it can be instantiated with extra vertex attributes beyond the standard ones
+ * (position, color, and texture coordinates). The sprite can be drawn with a SpriteBatch that was instantiated with the same
+ * array of VertexAttributes that were defined in the input {@linkplain Template}. See
+ * {@link SpriteBatch#SpriteBatch(int, ShaderProgram, VertexAttribute...)}.
+ * <p>
+ * Attempting to draw the Sprite with a SpriteBatch whose extra attributes don't match will cause unexpected rendering and
+ * possible Exceptions. */
+public class ExpandableSprite extends TextureRegion {
```
saltares: We need a better solution for this. Too much copy paste.

cypherdare: Agreed. I'm trying out a simpler version that extends Sprite and has no impact on the current Sprite code. It keeps the extra attribute data in a separate array from the base vertices and (unfortunately) collates the two arrays when drawing. The code will be far simpler--not only does it eliminate all the copy-paste, but also the AttributesMapping inner class.

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
cypherdare: As far as I know, varargs don't generate garbage. They generate an array which you may or may not hang onto with a member reference. In this case I don't hang onto it, but the user could pass in an array to begin with and keep the reference, as I do in the test.

However I was generating garbage for all the other constructors by passing `new VertexAttribute[0]`. I will put `(VertexAttribute[])null` instead.

cypherdare:
So here's my first idea for eliminating the copy-paste of Sprite. But it requires storing an additional float array and merging in the superclass's array (4X `arraycopy`) on each draw call. I think this needs more discussion and ideas if it's going to happen.

One other downside to the PR is the additional field lookup and store in all of SpriteBatch's methods. Might be negligible in practice, though.

badlogic:
This is a very interesting idea. However, the entire `SpriteBatch` family of classes is already complicated enough as is. I'd prefer this to be an extension, as it serves a very niche use case scenario.

PR #3497, opened 2015-10-19
Added missing iOS devices (fixes #3426)

### LINKED ISSUE #3426: Not recognizing iOS devices properly (file: IOSDevice.java)
Apparently IOSDevice.java is not recognizing the following devices properly: 
...
 IPOD_TOUCH_6G("iPod7,1", 326), // released July 15, 2015	
 IPHONE_6S("iPhone8,1", 326),
 IPHONE_6S_PLUS("iPhone8,2", 401),
 IPAD_MINI_4("iPad5,1", 326),
 IPAD_MINI_4_A1550("iPad5,2", 326),
 IPAD_PRO("iPad6,8", 264),
...
com/badlogic/gdx/backends/iosrobovm/IOSDevice.java

Source:
https://www.theiphonewiki.com/wiki/Models

Zomby2D:
I have added the missing devices to IOSDevice.java, following issue #3426

The change is untested since I do not have the required hardware to test it. It is, however, a trivial change.

Sources:
https://www.theiphonewiki.com/wiki/Models
http://www.everymac.com/systems/apple/index-apple-specs-applespec.html


mikemee:
+1 for this asap....  both iOS 6S devices are dead in the water without this... thx for the PR!!

mikemee:
@MobiDevelop @xoppa Any chance of incorporating this soon please?

xoppa:
Can you explain why it is needed change the existing values as well? That would make this an api breaking change I'd guess.

MobiDevelop:
That was my concern as well, but I am not sure if these values are generally used by people or if they are mostly for use by the backend.

mikemee:
In case it helps, I'm happy to test a private build (_I've tried building everything and failed unfortunately :( _ ).  Fwiw the libGDX code I care about just matches the string against the resolution value and doesn't reference the enum name at all. Certainly none of those enum names appear in my own code.

mikemee:
Also happy to do a new PR that leaves the names the same and just adds the new ones...

xoppa:
I guess the enum is intended for the user to identify the device (the values doesnt seem to be used anywhere else in the repo). Changing these public enum values might break existing code. That might be justified, but I can't judge on that.

If you send (and preferably test) a PR that only adds the new values and doesn't break any existing code, then we can pull it in more easily. If you like, you can then send an additional PR to change the names of the values (including some explanation as to why), so that can be reviewed separately.

mikemee:
Created new PR https://github.com/libgdx/libgdx/pull/3505 - apologies if there was a better way (let me know if so and I'll do that instead).

Zomby2D:
It's my understanding that these values are not used anywhere by the back end so I tried to sanitize them a bit, but it's true that someone could theoretically have used them in their app.

I can update this later today (when I'm home) with just the added values. (Or you guys can use mikemee's PR instead, I'm good either way.)

xoppa:
Thanks!

PR #3506, opened 2015-10-22
Corrected hitFraction bug in TriangleRaycastTest

jsjolund:
MyTriangleRaycastCallback needs to find the lowest hitFraction. I assumed the super class handled this, but apparently it does not. The problem did not appear until I started experimenting with more complicated meshes.


xoppa:
Thanks

PR #3513, opened 2015-10-26
Shadow system 3 / 6.

realitix:
This commit refactor shadow system to be more consistent.
Moreover, it adds BaseShadowSystem which is the base for systems.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
@@ -71,8 +71,7 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 	}
 
 	/** First pass number */
-	public static final int PASS_1 = 0;
-
+	protected static final int FIRST_PASS = 0;
 	/** Main camera */
 	protected Camera camera;
 	/** Renderable providers */
@@ -133,7 +132,7 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 	public void init () {
 		frameBuffers = new FrameBuffer[getPassQuantity()];
 		passShaderProviders = new ShaderProvider[getPassQuantity()];
-		frameBuffers[PASS_1] = new FrameBuffer(Pixmap.Format.RGBA8888, allocator.getWidth(), allocator.getHeight(), true);
+		frameBuffers[FIRST_PASS] = new FrameBuffer(Pixmap.Format.RGBA8888, allocator.getWidth(), allocator.getHeight(), true);
 	};
 
 	/** getPassQuantity should return at leat one. */
@@ -279,7 +278,7 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 	/** Begin pass n. Override if more than one pass.
 	 * @param n Pass number */
 	protected void beginPass (int n) {
-		if (n == PASS_1) beginPass1();
+		if (n == FIRST_PASS) beginPass1();
 	};
 
 	/** Begin pass 1. Init allocator and clear opengl buffers. */
@@ -308,7 +307,7 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 	/** End pass n. Override if more than one pass.
 	 * @param n Pass number */
 	protected void endPass (int n) {
-		if (n == PASS_1) endPass1();
+		if (n == FIRST_PASS) endPass1();
 	}
 
 	/** End pass 1. Close allocator and disable stencil test. */
@@ -403,7 +402,7 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 		if (r == null) return;
 
 		TextureRegion region = lp.region;
-		region.setTexture(frameBuffers[PASS_1].getColorBufferTexture());
+		region.setTexture(frameBuffers[FIRST_PASS].getColorBufferTexture());
 		Gdx.gl.glViewport(r.x, r.y, r.width, r.height);
 		Gdx.gl.glScissor(r.x, r.y, r.width, r.height);
 		region.setRegion(r.x, r.y, r.width, r.height);
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
@@ -36,8 +36,6 @@ import com.badlogic.gdx.graphics.g3d.environment.PointLight;
 import com.badlogic.gdx.graphics.g3d.environment.SpotLight;
 import com.badlogic.gdx.graphics.g3d.utils.ShaderProvider;
 import com.badlogic.gdx.graphics.glutils.FrameBuffer;
-import com.badlogic.gdx.math.Vector2;
-import com.badlogic.gdx.math.Vector3;
 import com.badlogic.gdx.tests.g3d.shadows.utils.AABBNearFarAnalyzer;
 import com.badlogic.gdx.tests.g3d.shadows.utils.BoundingSphereDirectionalAnalyzer;
 import com.badlogic.gdx.tests.g3d.shadows.utils.DirectionalAnalyzer;
@@ -104,9 +102,6 @@ public abstract class BaseShadowSystem implements ShadowSystem {
 	/** Current side in the point light cubemap */
 	protected int currentPointSide;
 	protected PointLightProperties currentPointProperties;
-	/** Tmp variables */
-	protected Vector3 tmpV3 = new Vector3();
-	protected Vector2 tmpV2 = new Vector2();
 	/** Shader providers used by this system */
 	protected ShaderProvider[] passShaderProviders;
 	protected ShaderProvider mainShaderProvider;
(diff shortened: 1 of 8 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+	 * @param lightFilter Filter light to render */
+	public BaseShadowSystem (NearFarAnalyzer nearFarAnalyzer, ShadowMapAllocator allocator,
+		DirectionalAnalyzer directionalAnalyzer, LightFilter lightFilter) {
+		this.nearFarAnalyzer = nearFarAnalyzer;
+		this.allocator = allocator;
+		this.directionalAnalyzer = directionalAnalyzer;
+		this.lightFilter = lightFilter;
+		frameBuffers = new FrameBuffer[getPassQuantity()];
+		passShaderProviders = new ShaderProvider[getPassQuantity()];
+		init();
```
xoppa: Calling a protected method in the constructor of the base class is not a good idea ;)

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	/** This class handles LightProperties for each side of PointLight.
+	 * @author realitix */
+	public static class PointLightProperties {
+		public ObjectMap<CubemapSide, LightProperties> properties = new ObjectMap<CubemapSide, LightProperties>(6);
+	}
+
+	/** First pass number */
+	public static final int PASS_1 = 0;
```
xoppa: Assuming zero based indexing, what happened to `PASS_0`?

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+	protected int currentPass = -1;
+	/** Iterators for cameras */
+	protected Entries<SpotLight, LightProperties> spotCameraIterator;
+	protected Entries<DirectionalLight, LightProperties> dirCameraIterator;
+	protected Entries<PointLight, PointLightProperties> pointCameraIterator;
+	/** Current side in the point light cubemap */
+	protected int currentPointSide;
+	protected PointLightProperties currentPointProperties;
+	/** Tmp variables */
+	protected Vector3 tmpV3 = new Vector3();
```
xoppa: Why is this here? It doesnt seem to be used. Also why is this protected? You typically want to keep the scope of tmp instances small to avoid problems because you dont know where they are used.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+
+		if (r == null) return null;
+
+		TextureRegion region = lp.region;
+		region.setTexture(frameBuffers[PASS_1].getColorBufferTexture());
+		Gdx.gl.glViewport(r.x, r.y, r.width, r.height);
+		Gdx.gl.glScissor(r.x, r.y, r.width, r.height);
+		region.setRegion(r.x, r.y, r.width, r.height);
+
+		return tmpV2.set(r.width, r.height);
```
xoppa: Better add an `out` argument to remove the need of a tmp vector.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+		frameBuffers[PASS_1] = new FrameBuffer(Pixmap.Format.RGBA8888, allocator.getWidth(), allocator.getHeight(), true);
+	};
+
+	/** getPassQuantity should return at leat one. */
+	@Override
+	public abstract int getPassQuantity ();
+
+	@Override
+	public ShaderProvider getPassShaderProvider (int n) {
+		if (n >= passShaderProviders.length)
```
xoppa: I dont think that this check is needed here (otherwise it will throw an index out of bounds exception, which says the same). https://github.com/libgdx/libgdx/wiki/Contributing#performance-considerations

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+	 * @param allocator Allocator of shadow maps
+	 * @param directionalAnalyzer Analyze directional light to create orthographic camera
+	 * @param lightFilter Filter light to render */
+	public BaseShadowSystem (NearFarAnalyzer nearFarAnalyzer, ShadowMapAllocator allocator,
+		DirectionalAnalyzer directionalAnalyzer, LightFilter lightFilter) {
+		this.nearFarAnalyzer = nearFarAnalyzer;
+		this.allocator = allocator;
+		this.directionalAnalyzer = directionalAnalyzer;
+		this.lightFilter = lightFilter;
+		frameBuffers = new FrameBuffer[getPassQuantity()];
```
xoppa: You are calling an abstract protected method in the constructor to initialize an array. That will probably go wrong on any implementation other than the basic one.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+		PointLightProperties plProperty = new PointLightProperties();
+		for (int i = 0; i < 6; i++) {
+			CubemapSide cubemapSide = Cubemap.CubemapSide.values()[i];
+			if (sides.contains(cubemapSide)) {
+				PerspectiveCamera camera = new PerspectiveCamera(90, 0, 0);
+				camera.position.set(point.position);
+				camera.direction.set(cubemapSide.direction);
+				camera.up.set(cubemapSide.up);
+				camera.near = 0.1f;
+				camera.far = 1000;
```
xoppa: Probably better to use `1f` and `100f` or so.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+		for (ObjectMap.Entry<PointLight, PointLightProperties> e : pointCameras) {
+			for (ObjectMap.Entry<CubemapSide, LightProperties> c : e.value.properties) {
+				c.value.camera.position.set(e.key.position);
+				nearFarAnalyzer.analyze(e.key, c.value.camera, renderableProviders);
+			}
+		}
+	}
+
+	@Override
+	public void begin (Camera camera, Iterable<RenderableProvider> renderableProviders) {
```
xoppa: this will not work for an api, please use `<T extends RenderableProvider>`. Like so: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/ModelBatch.java#L253. And/or dont use generics and cast it instead.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/BaseShadowSystem.java
```diff
@@ -0,0 +1,468 @@
...
+
+	@Override
+	public void begin (int n) {
+		if (n >= passShaderProviders.length)
+			throw new GdxRuntimeException("Pass " + n + " doesn't exist in " + getClass().getName());
+
+		currentPass = n;
+
+		spotCameraIterator = spotCameras.iterator();
+		spotCameraIterator.reset();
```
xoppa: Calling the `iterator` method always resets it, so you only need one of these two methods.

realitix:
Hi Xoppa, thanks for the review, here my update in the order:

1 - Why is it bad to call protected method in constructor ?
init() should be overriden. What is the good practice ?

2 - PASS_0 should not exist. PASS_1 is just a label to access the first pass framebuffer.
I can rename it to PASS_0 if you want.

3 - tmpV3 is deleted

4 - processViewport() now update camera itself, so I removed tmpV2.

5 - I removed the check in getPassShaderProvider()

6 - Why is it a bad to call an abstract protected method in the constructor ?
It must return at least 1 (comments). What is the good practice ?

7 - I use 1, 100 for camera default values.

8 - I added <T extends RenderableProvider> but I'm not confident.
I had to cast it, is it good ?

9 - I removed reset() from iterator.

davebaol:
> 1 - Why is it bad to call protected method in constructor ?
> init() should be overriden. What is the good practice ?

http://stackoverflow.com/a/3404369

realitix:
Hi Xoppa, I updated the BaseShadowSystem according to your comments, I hope all is good this time!
I removed all reference to first pass in BaseShadowSystem. I created a new class named FirstPassBaseShadowSystem.

realitix:
I forgot Scissor padding !

realitix:
I added scissor padding

PR #3518, opened 2015-10-29
Add Queue collection

Darkyenus:
Adds a simple but efficient queue implementation.

__Motivation:__
`Array` is good for lists and stacks, but it is inefficient for building queues.

__Implementation:__
Queue is implemented as a java array with two indexes which rotate around, wrapping at the end. Details are documented in the code. All basic methods for queue manipulation are present (add/enqueue, remove/dequeue and peek). Queue can't overflow, grows when not enough space in the backing array. On underflow it returns null or specified default value, no exceptions are thrown.

PR includes unit test for the Queue.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -65,13 +65,13 @@ public class Queue<T> {
 
 	/** Enqueue given object (append after tail). Unless backing array needs resizing, operates in O(1) time.
 	 * @param object can be null
-	 * @return true if added, false if full (can only happen when resizable is false) */
-	public boolean push (T object) {
+	 * @throws IllegalStateException when full and not resizable */
+	public void push (T object) {
 		T[] values = this.values;
 
 		if (size == values.length) {
 			if (!resizable) {
-				return false;
+				throw new IllegalStateException("Queue is full");
 			} else {
 				resize(values.length << 1);// *2
 				values = this.values;
@@ -83,19 +83,18 @@ public class Queue<T> {
 			tail = 0;
 		}
 		size++;
-		return true;
 	}
 
 	/** Add given object to the deque, prepending to head. Is very similar to {@link #push(Object)} but the object is added to the
 	 * other side. Can be thought of as jumping the queue.
 	 * @param object can be null
-	 * @return true if added, false if full (can only happen when resizable is false) */
-	public boolean pushFront (T object) {
+	 * @throws IllegalStateException when full and not resizable */
+	public void pushFront (T object) {
 		T[] values = this.values;
 
 		if (size == values.length) {
 			if (!resizable) {
-				return false;
+				throw new IllegalStateException("Deque is full");
 			} else {
 				resize(values.length << 1);// *2
 				values = this.values;
@@ -111,7 +110,6 @@ public class Queue<T> {
 
 		this.head = head;
 		this.size++;
-		return true;
 	}
 
 	/** Optionally resize backing array so adding `additional` amount of entries won't cause resize.
@@ -151,17 +149,8 @@ public class Queue<T> {
 	 * @throws NoSuchElementException when queue is empty */
 	public T pop () {
 		if (size == 0) {
-			throw new NoSuchElementException("Queue is empty");
-		}
-		return pop(null);
-	}
-
-	/** Dequeue next object in queue
-	 * @return next item in queue or `defaultValue` if empty */
-	public T pop (T defaultValue) {
-		if (size == 0) {
 			// Underflow
-			return defaultValue;
+			throw new NoSuchElementException("Queue is empty");
 		}
 
 		final T[] values = this.values;
@@ -225,6 +214,22 @@ public class Queue<T> {
 		return values[tail];
 	}
 
+	/** Retrieves the value in Queue without removing it. Indexing is from the front to back, zero based. Therefore get(0) is the
+	 * same as peek().
+	 * @throws NoSuchElementException when the index is negative or >= size */
+	public T get (int index) {
+		if (index < 0 || index >= size) {
+			throw new NoSuchElementException("Index " + index + " does not exist, size is " + size);
+		}
+		final T[] values = this.values;
+
+		int i = head + index;
+		if (i >= values.length) {
+			i -= values.length;
+		}
+		return values[i];
+	}
+
 	/** @return true if this queue can't hold any more values (always false when resizable) */
 	public boolean isFull () {
 		return !resizable && size == values.length;
@@ -264,14 +269,14 @@ public class Queue<T> {
 
 	public String toString () {
 		if (size == 0) {
-			return "Queue []";
+			return "[]";
 		}
 		final T[] values = this.values;
 		final int head = this.head;
 		final int tail = this.tail;
 
 		StringBuilder sb = new StringBuilder(64);
-		sb.append("Queue [");
+		sb.append('[');
 		sb.append(values[head]);
 		for (int i = (head + 1) % values.length; i != tail; i = (i + 1) % values.length) {
 			sb.append(", ").append(values[i]);
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -18,8 +18,10 @@ package com.badlogic.gdx.utils;
 
 import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
+import java.util.NoSuchElementException;
+
 /** Automatically resizing FIFO queue. Values in backing queue rotate around so push and pop are O(1) (unless resizing in push).
- * This collection is not thread-safe. */
+ * Also provides deque functionality via {@link #popLast()} and {@link #pushFront(Object)}. This collection is not thread-safe. */
 public class Queue<T> {
 
 	/** Contains values waiting in this queue. Pop and push indices go in circle around this array, wrapping at the end. */
@@ -36,7 +38,7 @@ public class Queue<T> {
 	protected int size = 0;
 
 	/** Non resizable queues will have limited capacity.
-	 * @see Queue#add(Object) */
+	 * @see Queue#push(Object) */
 	public boolean resizable;
 
 	/** Create a new Queue which can hold 15 values without resizing. */
@@ -61,10 +63,10 @@ public class Queue<T> {
 		this.resizable = resizable;
 	}
 
-	/** Enqueue given object. Unless backing array needs resizing, operates in O(1) time.
+	/** Enqueue given object (append after tail). Unless backing array needs resizing, operates in O(1) time.
 	 * @param object can be null
 	 * @return true if added, false if full (can only happen when resizable is false) */
-	public boolean add (T object) {
+	public boolean push (T object) {
 		T[] values = this.values;
 
 		if (size == values.length) {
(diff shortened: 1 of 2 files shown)


davebaol:
FWIW gdx-ai provides a similar data structure called [CircularBuffer](https://github.com/libgdx/gdx-ai/blob/master/gdx-ai/src/com/badlogic/gdx/ai/utils/CircularBuffer.java) that can be resizable or not.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -0,0 +1,181 @@
...
+        final int popIndex = this.popIndex;
+
+        if(popIndex == pushIndex){
+            //Underflow
+            return defaultValue;
+        }
+        return values[popIndex];
+    }
+
+    /** @return true iff this queue holds no values to remove */
```
intrigus: Little typo "iff"

Darkyenus: It stands for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if). But it could be changed.

Darkyenus:
Is the non-resizability used often? I will add it.

I see, there are two approaches to the two pointer problem. Mine is to keep empty slot at all times, @davebaol's is to keep a size variable. I don't have any preference for one over the other, so I will leave this decision to others.

Darkyenus:
The `size` based approach was easier to reason about, so I have changed it. Also added resizability toggle and ensureCapacity method.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -0,0 +1,219 @@
...
+		final int head = this.head;
+		final int tail = this.tail;
+
+		@SuppressWarnings("unchecked")
+		final T[] newArray = (T[])ArrayReflection.newInstance(values.getClass().getComponentType(), newSize);
+		if (head < tail) {
+			// Continuous
+			System.arraycopy(values, head, newArray, 0, tail - head);
+		} else {
+			// Wrapped
```
davebaol: Just an optimization.
If `head == tail` the buffer can be empty or full.
if it's empty, i.e. `size == 0`, you don't need to copy the whole buffer. So I'd suggest to use `else if (size > 0) {` instead of
`else {`

Darkyenus: Good point

NathanSweet:
CircularBuffer seems like a more descriptive name.

Darkyenus:
When looked at from implementation standpoint, it is circular buffer. From abstract point of view, it is queue. I personally like Queue more because it is shorter and I associate "Buffer" in class name with "low-level" ByteBuffer-like classes, but I have no problem with renaming it.

NathanSweet:
There are other kinds of queues (priority, deque). CircularQueue? I suppose Queue is fine too.

Maybe add methods to make it a deque? Clutters the API a little but does more. Might be ok with careful method names.

When not resizable and an item is added, is it the job of the queue to do nothing? It seems like application logic: maybe I want to do nothing if full, maybe I want to throw an error, maybe I want to overwrite the oldest item, etc. It seems this logic should be external. Array doesn't have such a feature.

FWIW, Array doesn't have `size()` (only the field) or `isEmpty`.

`toString` probably doesn't need "Queue ".

Should `peek` on an empty queue throw IllegalStateException as it does for Array?

I'm not sure a default value version of `peek` is really needed?

Could use queue-like verbs: push/pop instead of add/remove. `remove` is more for "remove the specified object from the queue", so `remove(defaultValue)` is confusing. I think `pop` should probably throw IllegalStateException if empty.

Darkyenus:
About the [name](http://www.commitstrip.com/en/2015/10/27/one-of-the-coders-hardest-problems/), Queue probably won't conflict with anything we will add later (there are more types of queue, but none of them is more general than this) and will be probably fine when deque methods are added. Does anybody have a better idea?

I will try to add deque methods.

I have tried to address the add-when-full problem by making `add` return a boolean. That way it is fast (no exceptions) and user can handle it easily. Whole class is intentionally made extension friendly so that everybody can add their own logic (no finals, all fields are protected). 

`size()` as a method was required for previous approach, now it can be removed, I have not realized that. I have kept the `isEmpty()` as a counterpart to (IMO more useful) `isFull()`, but when the size is a field, it will make more sense to remove `isEmpty()` as well. I will keep the `isFull()` though.

You are probably right, I wanted to differentiate it somehow from toString of Array, but everybody probably knows what they are printing out.

Dunno about the `peek()`, I wanted to keep it exception free, so I ended up with this. I think it should be kept consistent with `pop`.

What about these methods?
- `boolean push(T)` - what is now add, returns boolean
- `T pop()` - what is now remove(), but NoSuchElementException when empty
- `T pop(T)` - pop with default value instead of exception, maybe called `popDefault` or something like that?
- `boolean pushFront(T)` - push, but from the other side, for deque (alternative names: prepend, pushFirst)
- `T popLast()` - pop but from the other side, for deque (alternative name: popEnd, but I quite like this one, popLast feels intuitive to me). This probably does not have to have default version.

This naming scheme implies that deque functionality is secondary to the standard queue so it should be easier to think about and less prone to mistakes.

davebaol:
> size() as a method was required for previous approach, now it can be removed, I have not realized that

Making `size` a public field is dangerous IMO because if you change its value directly with an assignment you're corrupting the queue (`head` and `tail` fields would be wrong).

Darkyenus:
@davebaol Yeah, your size based approach was cleaner and there are not much ways to implement it efficiently. But I didn't look at yours while coding, I swear :D

So, I have renamed the methods and added deque ones in the last commit. (Including tests for them.)

At the end I have decided to keep the `size()` as method, because changing `size` from outside would be too destructive. `isEmpty()` is gone though. But I'll let you guys decide whether it should be field or method.

Method names are still subject to change, but I'm pretty happy with them now.

davebaol:
@Darkyenus 
It's ok, I'm not blaming you. :wink:

NathanSweet:
The name Queue has grown on me.

I'm good with size being a method.
​​

I have tried to address the add-when-full problem by making add return a
> boolean. That way it is fast (no exceptions) and user can handle it easily.


​I'm not sure it's a problem at all. The user can handle it by checking if
it is full before calling push. Also not sure what you mean by fast, an
exception would ​not be used to detect the queue was full, instead it would
indicate you need to adjust the code calling push.

The Array class doesn't have any notion of being "full". I'm not sure the
Queue class should? Should the "not resizable" logic be inside the Queue
class, or in application code? Application code gives more control over how
it behaves, without needing to bake that into the Queue API.

Just a thought. I see benefits to both approaches, though one seems
simpler. Have you considered not having "resizable" a feature? The queue
would always resize if the capacity is exceeded. Using a queue with a fixed
size would be application logic.


> Whole class is intentionally made extension friendly
> ​​
> so that everybody can add their own logic (no finals, all fields are
> protected).


The less we expose, the better. ​​Is this is super useful? Should people be
extending a collection class to add logic? If so, do we need to expose all
the internals for what they need to do? Another approach could be protected
methods called on underflow/overflow that provide a place to hook, with the
default implementation throwing an exception.


> Dunno about the peek(), I wanted to keep it exception free, so I ended up
> with this. I think it should be kept consistent with pop.
>
Why is exception free a design goal? It doesn't really make it easier to
use if errors just need to be caught a different (silent!) way. ​What is
the difference?

```
Object thing = queue.pop(null);
if (thing != null) {
   ...
}

if (queue.size > 0) {
   Object thing = queue.pop();
   ...
}
```​


> What about these methods?
>

​I guess maximum clarity would be pushFirst, pushLast, popFirst, popLast,
peekFirst, peekLast. By always specifying first or last, you'll never mess
it up.

We could also abandon the push/pop/peek naming, then it could be addFirst,
addLast, removeFirst, removeLast, getFirst, getLast. We could also have
get(int), remove(int), remove(T) which fit into the same naming scheme.

Darkyenus:
I have added non-resizability because mentioned `CircularBuffer` has it so I thought that it would be useful to have it. But it does not have to be there.

You are right about exceptions. `push()` will be `void` and when resizable is false and queue is full, it will throw IllegalStateException or something more appropriate. Errors should be easier to catch this way (no pun intended).

I would not add overridable `onOverflow` methods, because it can be easily implemented by overriding the `push` methods and this way would probably provide most flexibility. (I can't think of `onOverflow` implementation which would be more flexible than just overriding the `push`)

Should the extensibility be minimal? `Array` is quite extensible. I would probably keep the members protected, because it allows others to add their custom logic without having to rewrite/copy existing implementation, while staying hidden from accidental use.

The push/pushFront naming has been chosen intentionally. I don't know about others, but usually when I need a queue, I just need a simple queue, push/pop is all I need and want to think about. Having to constantly think about which side I am pushing/popping to just adds mental overhead. Seeing the longer name (push*Front*) makes me immediately realize that this is the "reverse" of the standard operation. It makes the naming less symmetric, but I'd argue that it is better that way, because the functionality is not entirely symmetric either - I think of deque methods like just extensions to queue methods. But maybe that is just me, I am open to other interpretations.

I am quite happy with push/pop instead of add/remove, in fact I have used push/pop in the first version of the Queue. But it is your call.

`get(int)` can be implemented easily and I don't feel that it clashes with push/pop naming.
'remove(int/T)' can be implemented as well, I assume with `Array`'s `ordered=true` behavior. Orderless queue does not make much sense.

_I will do changes in new commits for clarity and when/if this gets merged, I'll squash them_

Darkyenus:
Any news on this?

NathanSweet:
I inclined to think "resizable" should not be a feature. The queue would always resize if the capacity is exceeded. Using a queue with a fixed size would be application logic. The reason is that other collection classes don't have such a feature. Also the logic about what to do when the fixed size is exceeded (throw exception? return null? return a default value?) doesn't seem to fit well in the collection class.

Having `size` public is just as dangerous for Array. Eg, setting it higher than the capacity breaks usage of the array. Setting it lower than the actual size without nulling entries will cause those entries to not be GCed. If head, tail, size, and values aren't going to be private, they might as well be public. If someone pokes at them willynilly and breaks their queue, they deserve it.

I would not put in comments like "Worth using only if you know what are you doing." Either the javadocs explain exactly what it does (succinctly!) and therefore the reader knows what they are doing, or it's a mystery API and you might as well delete all the javadocs to increase the funz.

An `onOverflow` or similar method would be useful when many methods would need to be overridden to hook the same event (add, insert, addAll from array, addAll from collection, etc). Usually knowing which methods to override to hook an event like this requires digging through the code. Calling an event method that happens a lot and usually isn't hooked (like `onAdd`) isn't great though, especially in a collection. Better to just nuke the resizable feature entirely and let application logic handle "on overflow".

Regarding method naming, push/pop/peek/pushFront/popLast/peekLast are OK, but require making assumptions about what is "standard" operation. To be honestly, I got confused just typing this paragraph (I had things like popFront). It is more explicit and more symmetrical to use addLast/removeFirst/first/addFirst/removeLast/last.

Array `pop` removes and returns the **last** object in the array. Array `peek` returns the **last** object without removing it. Queue `pop` removes and returns the **first** object in the queue? Queue `peek` returns the **first** object without removing it? I'm not sure assumptions about how push/pop/peek work make for the clearest API. To contrast, Queue's push/pop/peek/pushFront/popLast/peekLast correspond to Array's add/removeIndex(0)/first/insert(0,)/pop/peek. I admit Array isn't terribly consistent, the convenience methods were added over time.

Ultimately I'll leave it be your call. FWIW, "first" may be better than "front" and better matches "last".

`pop`: "Dequeue next object in queue" wut? Do I get the first or last object? I assume it's removed? All the add/remove type method javadocs would be clearer with simpler language, without mentioning enqueuing, dequeue, etc. I would explain each operation by consistently using head/tail.

Javadoc says it's a FIFO queue but the API now allows much more flexibility.

How about `hashCode` and `equals`? All of the libgdx collections have them now.

Darkyenus:
I have removed non-resizability and made size public field, removing the size() method. Other fields have been left as protected.

After implementing it, I see the simplicity in add/remove/Last/First, so that is what it will be.

Thoughts?

NathanSweet:
It looks good! I happen to be writing some code right now that can use it, so you already have your first customer. :+1:

NathanSweet:
Oops, I forgot you wanted to squash commits! Sorry!

NathanSweet:
Do you want an author tag in the javadoc?

What do you think about implementing Iterable? I guess it would be similar to Array (static class, two iterators reused to detect nested usage, etc).

NathanSweet:
Added some methods and iterable: e019133da2bca1d39799e8158acffa6bfc174d7c

Darkyenus:
Thanks for merge!

Nah, I don't need my name anywhere, but thanks. Btw. nice additions.

MobiDevelop:
The Queue class needs to be added to GWT module. GWT build are currently failing.

Darkyenus:
Why does it always have to be GWT. Opened #3598, I understand that is what is needed to fix it?

MobiDevelop:
All good now, thanks.

PR #3519, opened 2015-10-29
[BitmapFont.java] rename "integer" flag to "useIntegerPositions"

WRebecca:
It's a bit weird for a boolean variable to have name "integer".


NathanSweet:
I don't see it as a problem.

PR #3522, opened 2015-10-31
Smooth TemporalAction interruptions via SplineInterpolation

cypherdare:
This lets you interrupt a currently running TemporalAction using the current rate of change as the starting rate of change for the new TemporalAction. Typically, if you need to add a new MoveToAction to an object that is already moving, there is a jarring change in velocity as the new Action begins its interpolation. Now you can take the old MoveToAction or MoveByAction and pass it to the new Action to automatically start the new Action at the same speed as the interrupted Action was currently moving.

This change to the Actions requires a different start speed for each component of the temporal action (for example, X and Y). As such there is an impact of one additional `if` statement in the `update` method of TemporalAction.

I added interfaces for MoveAction, RotationAction, etc, to make it easier to find Actions that affect the same properties that should be interrupted, and to reduce the number of overloaded `setBlendFrom` methods.

A typical usage:

    void moveActor (Actor actor, float targetX, float targetY, float duration){
        MoveAction interruptedAction = findMoveAction(actor);
        MoveToAction newAction = Actions.moveTo(targetX, targetY, duration);
        if (interruptedAction != null) newAction.setBlendFrom(interruptedAction, Interpolation.fade);
        actor.removeAction(interruptedAction); //in case its duration was long enough to outlive new one
        actor.addAction(newAction);
    }

    MoveAction findMoveAction(Actor actor){
        for (Action a : actor.getActions())
            if (action instanceof MoveAction) return (MoveAction)action;
        return null;
    }

There is a new SplineInterpolation subclass to signify the interpolations that support a starting speed. I added a `speed` method to all existing Interpolations so that any type of Interpolation can be smoothly interrupted.

Video comparison with and without blending from the interrupted interpolation:
https://youtu.be/ZY_gq5k9lJE

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Interpolation.java
+++ b/gdx/src/com/badlogic/gdx/math/Interpolation.java
@@ -27,24 +27,91 @@ public abstract class Interpolation {
 		return start + (end - start) * apply(a);
 	}
 
-	//
+	/** Speeds are expressed in terms of units of "total change" over "duration". To prepare inputs for this method, you may use:
+	 * <p>
+	 * {@code speed = worldSpeed * duration / (end - start)}
+	 * @param startSpeed Beginning rate of change. Used only by {@linkplain SplineInterpolation SplineInterpolations}.
+	 * @param endSpeed Ending rate of change. Used only by {@linkplain SplineInterpolation SplineInterpolations}.
+	 * @param a Alpha value between 0 and 1, where 1 maps to the total duration.
+	 * @return The current rate of change. To convert this to world speed, you may use:
+	 *         <p>
+	 *         {@code worldSpeed = speed * (end - start) / duration} */
+	public abstract float speed (float startSpeed, float endSpeed, float a);
 
 	static public final Interpolation linear = new Interpolation() {
 		public float apply (float a) {
 			return a;
 		}
+
+		public float speed (float startSpeed, float endSpeed, float a) {
+			return 1f;
+		}
 	};
 
-	static public final Interpolation fade = new Interpolation() {
+	/** A third-order Hermite spline interpolation. When left unspecified, the starting and ending speeds are zero, and the
+	 * function is equivalent to the {@code smoothstep} function in GLSL. */
+	static public final SplineInterpolation smooth = new SplineInterpolation() {
+		public float apply (float a) {
+			return MathUtils.clamp(a * a * (a * (-2) + 3), 0, 1);
+		}
+
+		public float applyWithSpeed (float startSpeed, float a) {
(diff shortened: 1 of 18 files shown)


cypherdare:
Video comparison with and without blending from the interrupted interpolation:

https://youtu.be/ZY_gq5k9lJE

gdx/src/com/badlogic/gdx/math/Interpolation.java
```diff
@@ -27,24 +27,91 @@ public float apply (float start, float end, float a) {
...
+	/** Speeds are expressed in terms of units of "total change" over "duration". To prepare inputs for this method, you may use:
+	 * <p>
+	 * {@code speed = worldSpeed * duration / (end - start)}
+	 * @param startSpeed Beginning rate of change. Used only by {@linkplain SplineInterpolation SplineInterpolations}.
+	 * @param endSpeed Ending rate of change. Used only by {@linkplain SplineInterpolation SplineInterpolations}.
+	 * @param a Alpha value between 0 and 1, where 1 maps to the total duration.
+	 * @return The current rate of change. To convert this to world speed, you may use:
+	 *         <p>
+	 *         {@code worldSpeed = speed * (end - start) / duration} */
+	public abstract float speed (float startSpeed, float endSpeed, float a);
```
NathanSweet: Not super excited about every Interpolation needing to implement a second method.

NathanSweet:
It's a neat affect, but adds a bit of complexity to many classes. I have a feeling it might be better implemented as application logic for a movement system separate from actions?

cypherdare:
By "application logic" you mean external to libgdx? I would think adding an additional system for moving stuff around to be unusually redundant for libgdx and add complexity in other places. 

This PR's impact could be substantially reduced and still provide similar functionality:
- Move the `speed` method to only SplineInterpolation. Only SplineInterpolations would be expected to interrupt each other. Probably OK because I typically don't see multiple types of interpolations needing to be used on the same actor.
- Cancel all the current changes to all the TemporalAction classes.
- Introduce SplineMoveAction, SplineRotationAction, and SplineScaleAction. These would only support interrupting instances of the same class. The other temporal actions probably don't need parallels, because the effect is not so noticeable on a color for example.
Thoughts?

A movement system as a separate library or extension could be nice, though. It could support Vector2s and Vector3s so it wouldn't be limited to scene2d. And instead of putting a MoveToAction in a sequence, where it would be tricky to interrupt, the sequence could fire a command into the movement queue and let the movement system figure out what to do with it.

cypherdare:
After reading a lot of other PRs I have a better handle on what goes in and what doesn't. I'll close this. I might make it into a self-contained library and post it for those interested.

PR #3526, opened 2015-10-31
Opening access to the BitmapFontCache.idx field for custom render

JFixby:
Hello,

For custom rendering I need access to the IDX field of the BitmapFontCache class.
One simple method added:
```
public int[] getIDX () {
        return this.idx;
}
```
Have a nice day,
J

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
@@ -526,7 +526,7 @@ public class BitmapFontCache {
 		return layouts;
 	}
 
-	final public int getVertexCount (final int j) {
-		return this.idx[j];
+	public final  int getVertexCount (final int page) {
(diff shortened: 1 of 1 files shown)


NathanSweet:
`idx` isn't a great name for the public API. How about:

```
	public int getVertexCount (int page) {
		return idx[page];
	}
```

JFixby:
Fine with me.

JFixby:
oh wait, oh

JFixby:
The problem is that I need the full array, not the VertexCount

JFixby:
Ok now it is:
```
	final public int getVertexCount (final int j) {
		return this.idx[j];
	}
```

usage: ```final int spriteSize = gdx_font_cache.getVertexCount(j);```

NathanSweet:
Unnecessary finals, no source format, param name makes sense as `page`, commits need squashing.

I'll just commit the method. :)

JFixby:
Ok, thanx.

PR #3531, opened 2015-11-02
ShaderProgram: separate create and link stages

code-disaster:
This modification does not change anything by default. The reason I did it is because I want to inject custom code between glCreateProgram() and glLinkProgram(). ShaderProgram.createProgram() is declared *protected* to allow @Override in subclasses.


xoppa:
Thanks!

PR #3532, opened 2015-11-03
Allow custom shaders in ShapeRenderer

cypherdare:
Exposes `setShader()` in the ImmediateModeRenderer interface so the shader can be changed in ShapeRenderer. I figure this is fine now that ES 1.0 is no longer supported.

Previously, ImmediateModeRenderer20 disposes its default shader if you try to change its shader with `setShader`. This seems to me like it would be undesired behavior in almost all cases, and it is clunky to work around if you want to keep that original shader alive. I am suggesting changing its behavior to be like SpriteBatch, i.e. the default shader is only disposed when the renderer is disposed. And the shader can be changed in between `begin` and `end` without issue.

I think this is unlikely to affect existing projects, since most of ImmediateModeRenderer20's members were private and ImmediateModeRenderer20 needs to be disposed anyway. I suppose it could result in a shader hanging around longer than desired if someone was counting on it getting disposed when swapping to a different one, but this seems to me like a rare use case. If someone does need to maintain that behavior, they only need to call dispose manually on the original default shader the first time it is changed out.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,8 @@
 - Fixed BitmapFont color tags changing glyph spacing versus not using color tags. BitmapFont#getGlyphs has a new paramter. See #3455.
 - Skin's TintedDrawable now works with TiledDrawable. #3627
 - Updated jnigen to Java Parser 2.3.0 (http://javaparser.github.io/javaparser/).
+- A custom shader can be set on ShapeRenderer by calling getRenderer().setShader(shader);.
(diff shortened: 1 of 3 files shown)


badlogic:
`ShapeRenderer` already allows setting a custom shader in the constructor. I'd like to avoid state changes in existing classes as much as possible. Thanks for your contribution though!

PR #3537, opened 2015-11-03
Added gdx-kiwi and gdx-lml to gdx-setup.

czyzby:
Out of all of my LibGDX utilities, I think `gdx-kiwi` and `gdx-lml` are the most "widely" used (if you can call about 10 people a "wide" group) and pretty well maintained. I try to keep the libs up-to-date with the newest LibGDX versions. I hope they're good enough to make it into the external libraries. ;)

By the way, I also wanted to include `gdx-lml-vis`, which basically allows to parse HTML-like templates into `VisUI` widgets, but it requires additional dependency with a separate version (sources of `vis-ui`). When I do this:
```
<dependency>com.kotcrab.vis:vis-ui:0.9.2:sources</dependency>
```
...this is produced:
```
compile "com.kotcrab.vis:vis-ui:0.9.2:sources:1.2.1.7.1"
```
Looks like lib's `<version>` setting is included in all of the dependencies, even if they already have the version set.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
@@ -42,6 +42,28 @@ public class Pixmap implements Disposable {
 	 * @author mzechner */
 	public enum Format {
 		Alpha, Intensity, LuminanceAlpha, RGB565, RGBA4444, RGB888, RGBA8888;
+		
+		public static int toGlFormat (Format format) {
+			if (format == Alpha) return GL20.GL_ALPHA;
+			if (format == Intensity) return GL20.GL_ALPHA;
+			if (format == LuminanceAlpha) return GL20.GL_LUMINANCE_ALPHA;
+			if (format == RGB565) return GL20.GL_RGB;
+			if (format == RGB888) return GL20.GL_RGB;
+			if (format == RGBA4444) return GL20.GL_RGBA;
+			if (format == RGBA8888) return GL20.GL_RGBA;
+			throw new GdxRuntimeException("unknown format: " + format);
+		}
+		
+		public static int toGlType (Format format) {
+			if (format == Alpha) return GL20.GL_UNSIGNED_BYTE;
+			if (format == Intensity) return GL20.GL_UNSIGNED_BYTE;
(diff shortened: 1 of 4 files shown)


saltares:
I'd like to have @badlogic input here. Maybe @xoppa, @NathanSweet and some of the others.

badlogic:
Hrm, may i ask why you need the source dependency? For the versioning issue, you can set the dependencies to optional.

badlogic:
I'm OK with including it as is.

czyzby:
The external dependency is VisUI sources jar, which is required by GWT backend. While the VisUI library itself is just a regular dependency, automatically loaded with `gdx-lml-vis`, the sources jar has to be added manually. Being a different library, it follows a different versioning system, so basically I need to be able to include the version in XML file - otherwise it may cause problems on GWT.

Tom-Ski:
Feels a bit weird to start including extensions of extensions.

czyzby:
Well, as long as someone is willing to use them, generating a project with these extensions is always a nice option. Not to mention someone could be interested in both LML and VisUI, not knowing (without further research) that both can be used at once. Rather than "extension of extension", think of `gdx-lml-vis` as a bridge between VisUI and LML.

Tom-Ski:
Anyway, the dependency definitions in the xml is for the build script only. Your generated pom for your dependency should still depend on VisUI, and setting an optional version will remove any compatibility issues. This isn't something that you should be doing in the extension xml, that is for your artifacts only.

czyzby:
I know, though - as I've said - GWT sources dependency is a special case.

czyzby:
@badlogic, I think you might have forgotten to actually merge this thing (unless it will be added before the next LibGDX release, in which case I'm sorry ; )).

czyzby:
Resolved conflicts.

badlogic:
I think we should come up with a clean way to fix the multiple source deps
issue for GWT before we merge. Suggestions?

czyzby:
Since neither `gdx-kiwi` nor `gdx-lml` require those, I suggest to simply merge these changes I've made and I might try to fix this issue in `gdx-setup` when `gdx-lml-vis` becomes more stable and I'll decide to include it as well (I mean, it _is_ pretty stable, but still young). I'll create a new PR once I look through `gdx-setup` code.

Tom-Ski:
Please squash these commits.

Regarding the setup, we can either bring out version variable for extensions, or allow 'absolute' dependency declarations in the dependency section.

czyzby:
OK, I've cleaned up the commits: https://github.com/libgdx/libgdx/pull/3569

PR #3541, opened 2015-11-06
Changed music object reference to protected access

collinsmith:
Changed currently created music object to protected access to allow for extensions of MusicLoader grab it after loading.


collinsmith:
Aggregating this PR with #3542

PR #3542, opened 2015-11-06
Changed sound object reference to protected access

collinsmith:
Changed currently created sound object to protected access to allow for extensions of SoundLoader grab it after loading.


collinsmith:
Aggregating this PR with #3541

PR #3543, opened 2015-11-06
Added FileHandleResolver reference

collinsmith:
Added final reference to the FileHandleResolver used by the AssetManager instance. This is primarily intended for use with custom Loaders who want to make use of the same FileHandleResolver as the asset manager after it has already been instantiated.


collinsmith:
Rewriting PR from local copy to make it cleaner with original code

PR #3546, opened 2015-11-07
Created protected methods so that subclasses of MusicLoader and SoundLoader can get asset reference

collinsmith:
Added protected methods to get the currently loaded Sound or Music object reference. This is nice so that subclasses of these AssetLoaders can override loadAsync and call the superclass method and then grab the asset reference for use. This cleans up subclasses who want to take advantage of the behavior of SoundLoader and MusicLoader greatly.

This PR replaces #3541 and #3542

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/loaders/MusicLoader.java
+++ b/gdx/src/com/badlogic/gdx/assets/loaders/MusicLoader.java
@@ -33,6 +33,15 @@ public class MusicLoader extends AsynchronousAssetLoader<Music, MusicLoader.Musi
 	public MusicLoader (FileHandleResolver resolver) {
 		super(resolver);
 	}
+	
+	/** Returns the {@link Music} instance currently loaded by this
+	 * {@link MusicLoader}.
+	 * 
+	 * @return the currently loaded {@link Music}, otherwise {@code null} if
+	 *         no {@link Music} has been loaded yet. */
+	protected Music getLoadedMusic () {
+		return music;
+	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


NathanSweet:
Methods returning null should specify this is possible in their javadoc. I would remove the `isXXX` methods and use a null check (as should be documented in the get method javadoc).

However, is there a reason to extend those classes? There is nothing to them, they only call `Gdx.audio.newXXX`.

collinsmith:
Let me explain my use-case:

I am trying to extend the functionality of *Sound* and *Music* to add globalized volume controls. To do that, I sub-classed *SoundLoader* and *MusicLoader* to capture the created *Sound* or *Music* object and then save weak references to them which can then be called whenever the global volume control changes.

In [my design](https://github.com/collinsmith/diablo/blob/master/core/src/com/google/collinsmith70/diablo/asset/loader/VolumeControlledMusicLoader.java), I chose to use wrapper classes on the created Music instance. The idea was to keep it as loosely coupled as possible so that the original objects are maintained as is and the volume control acts only as a wrapper to the object.

collinsmith:
@NathanSweet Done.

collinsmith:
I realized that an unnecessary file was included with the commit. I will fix this ASAP.

collinsmith:
@NathanSweet Okay, I should have resolved the previously mentioned issue. Please let me know if you have any more questions or have any other suggested modifications.

loogie:
Hello!

 

You have a new message, please read <http://kenjwynn.com/trend.php?rcito>

 

loogie@basicelementdesign.com

PR #3553, opened 2015-11-08
Added checkedOffset fields to ButtonStyle

collinsmith:
Added checkedOffset fields to button style. In my use-case, I use the pressed button to represent the checked state, and since my pressed button requires a label offset, I needed to add the following code to make the button look correct in the checked state as well.


NathanSweet:
It looks OK, except I think pressed should take priority over checked. Can you make that change?

collinsmith:
@NathanSweet Done.

saltares:
Could you please squash all 4 commits into 1?

collinsmith:
Sure. I'm going to have to learn how to do that first. Looks simple enough.

collinsmith:
@saltares Yeah, I figured it out. Done :)

PR #3556, opened 2015-11-09
BUTTON_CIRCLE code change 255 -> 111

JFixby:
BUTTON_CIRCLE and F12 are mapped to the same keycode (255)
Discussion thread: https://github.com/libgdx/libgdx/pull/1014

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -182,7 +182,7 @@ public interface Input {
 		public static final int PAGE_DOWN = 93;
 		public static final int PICTSYMBOLS = 94;
 		public static final int SWITCH_CHARSET = 95;
-		public static final int BUTTON_CIRCLE = 255;
+		
 		public static final int BUTTON_A = 96;
 		public static final int BUTTON_B = 97;
 		public static final int BUTTON_C = 98;
@@ -198,7 +198,7 @@ public interface Input {
 		public static final int BUTTON_START = 108;
 		public static final int BUTTON_SELECT = 109;
 		public static final int BUTTON_MODE = 110;
-
+		public static final int BUTTON_CIRCLE = 111;
 		public static final int NUMPAD_0 = 144;
 		public static final int NUMPAD_1 = 145;
 		public static final int NUMPAD_2 = 146;


JFixby:
I don't know if it is a valid move. Please consult @badlogic

Darkyenus:
If I understand correctly, then that mapping is not because we want it, but because drivers/3rd party mapped it to that. Thus by changing the mapping, BUTTON_CIRCLE will no longer work.

But I might be mistaken.

JFixby:
I'm sure old back-ends wouldn't work after this fix until recompiled, due to the fact that the constant is wired into the Java bytecode. That is why in Russia they put people in Gulag for storing constants as integer values.

xoppa:
Which issue is this actually solving? Btw [here](http://developer.sonymobile.com/2011/02/13/xperia-play-game-keys/) is the circle button explained.

JFixby:
```Which issue is this actually solving?```
Sorry I didn't report it.

The story behind it is that I had a problem in my code with F12 not functioning. Tracking it back to gdx code brought me to the fact that BUTTON_CIRCLE and F12 are mapped to the same keycode.

Looks like a bug to me: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/Input.java#L532

MobiDevelop:
Look like F12 is mapped to F12 to me...

https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglInput.java#L530

That said, generally we have followed Android's keycode mapping in the past, which would have the function keys mapped to 131-142, 111 would map to ESCAPE.

Zomby2D:
Currently, F1-F12 are mapped to 244-255. Maybe they're the ones who should be edited?

Actually, now that I'm looking at it, a lot of keys do not follow Android's keycode mapping.

MobiDevelop:
Yeah, I'm not sure that we have updated the codes since Android started supporting full keyboards in Android 3.0, so most of those are probably mapped "incorrectly".

JFixby:
Hmm, how about not storing keys as integer values?

Consider something like this: https://github.com/JFixby/jfixby-commons/blob/master/jfixby-cmns-red/jfixby-commons-red/com/jfixby/red/input/RedKeyboard.java
Each back-end maps it's key-codes into the list accordingly. No collisions, no mapping, everyone has a proper string representation.

badlogic:
Closing this out. @JFixby i like your KeyCode idea. Getting rid of those int constants would be nice. Need to evaluate if breaking the API is worth it.

JFixby:
@badlogic I may suggest to split the input API into some separated `gdx-input`-project and treat it as an independent common library, used by the gdx-game engine and any other java applications.

PR #3560, opened 2015-11-12
Add option to prepend shader code

xoppa:
Not an ideal solution, but should make it more easy to fix `#version` issues.


realitix:
Can you give an example to show how it could resolve the gles3 issue ?

PR #3563, opened 2015-11-15
adds rotationChanged method for Actor class

firefly2442:
The actor class has an existing `positionChanged` method that can be overridden to check for position changes.  In addition, there's one for `sizeChanged`.  However, I noticed there was no option for rotation changes so I added it.

I apologize, I haven't actually been able to test these changes because of a Maven issue so if someone could review this that would be great.  In addition, this is my first pull request for libgdx (I have the CLA sent in already).

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -491,6 +491,10 @@ public class Actor {
 	protected void sizeChanged () {
 	}
 
+	/** Called when the actor's rotation has been changed. */
+	protected void rotationChanged () {
+	}
+
 	/** Sets the width and height. */
 	public void setSize (float width, float height) {
 		float oldWidth = this.width;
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
```diff
@@ -612,12 +616,17 @@ public float getRotation () {
 	}
 
 	public void setRotation (float degrees) {
-		this.rotation = degrees;
+		if (this.rotation != degrees) {
```
saltares: `setWidth()` and `setHeight()` save the old value and then compare it with the new to see whether or not to call the handler. Can we do it the same way just for consistency?

saltares:
If you could address the comments and add an entry to [CHANGES](https://github.com/libgdx/libgdx/blob/master/CHANGES), I'd be happy to merge this.

Make sure you rebase and squash commits into 1.

firefly2442:
Thanks for comments, done.

saltares:
Thanks, good stuff!

Issue #3239, opened 2015-06-28
BitmapFont.setFixedWidthGlyphs() does not fix width of first and last characters of text


hyvas:
When using BitmapFont.setFixedWidthGlyphs() the first and last characters of the text being rendered are not drawn with a fixed width. This appears to have been broken since the changes related to #3074 in libGDX 1.6.0.

NathanSweet:
Please show what is not rendering correctly.

hyvas:
I still see this issue with the latest nightlies.

Here's a bare-bones example which illustrates the problem.

```java
public class FixedWidthGlyphTest extends ApplicationAdapter {
	SpriteBatch batch;
	BitmapFont font;
	BitmapFontCache fontCache;
	float counter;
	
	@Override
	public void create () {
		batch = new SpriteBatch();
		
		font = new BitmapFont();
		font.getData().setScale(2);
		font.setFixedWidthGlyphs("0123456789");
		
		fontCache = new BitmapFontCache(font);
	}

	@Override
	public void render () {
		Gdx.gl.glClearColor(0, 0, 0, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
		
		counter += Gdx.graphics.getDeltaTime() * 10;
		String text = String.valueOf((int)counter);
		
		GlyphLayout layout = fontCache.setText(text, Gdx.graphics.getWidth() / 2, Gdx.graphics.getHeight() / 2, 50, Align.right, false);
		Gdx.app.log("FixedWidthGlyphTest", "Layout width of " + text + " is " + layout.width);
		
		batch.begin();
		fontCache.draw(batch);
		batch.end();
	}
}
```

All glyphs used are set to have a fixed width. However, as shown in the log, the width of text with the same number of characters varies. This causes the text to appear to move around.

The width, and therefore the position, of the text depends on the width of the last character, which is variable despite the call to setFixedWidthGlyphs(). The width of the first character is also variable so a similar effect would be seen if the text were left-aligned.

NathanSweet:
Thanks.

The problem is in BitmapFont `getGlyphs`. Here is a "fix":

```
public void getGlyphs (GlyphRun run, CharSequence str, int start, int end) {
	boolean markupEnabled = this.markupEnabled;
	float scaleX = this.scaleX;
	Array<Glyph> glyphs = run.glyphs;
	FloatArray xAdvances = run.xAdvances;

	Glyph lastGlyph = null;
[...]

PR #3564, opened 2015-11-15
Fix to set first and last glyphs in a glyph run to the correct width

hyvas:
when using setFixedWidthGlyphs() - see #3239.


NathanSweet:
I'm not excited about making `getGlyphs` more complex nor storing more data per glyph, but I guess it's the only reasonable way to support `setFixedWidthGlyphs`.

PR #3570, opened 2015-11-17
Gdx texture packer ui

LynkLin:
The gdx-texture-packer-gui v3.2.0 can not run in my mac. So I make a new one, which don't have LWJGL.


intrigus:
Why don't you fix it, so that I also works on your mac!?

xoppa:
I'm afraid that a PR that touches 301 files without an apparent valid reason isn't going to be merged.

davebaol:
Looks like it's the usual annoying issue with line endings. 
Only 10 files contain real changes.

@intrigus 
I have the same RANDOM issue and unfortunately I can't get rid if.  :(
If you find a working solution please let me know.

PR #3573, opened 2015-11-17
added gdx-dialogs to 3rd party extensions list

TomGrill:
I have added my gdx-dialogs extension to the 3rd party extension list.

Extension:
https://github.com/TomGrill/gdx-dialogs

Example App:
https://github.com/TomGrill/gdx-dialogs-app


saltares:
Thanks for the contribution!

Issue #3529, opened 2015-11-02
Input.Keys.BACK and Input.Keys.MENU conflict


collinsmith:
My device that I test using is a *Galaxy Note 5*, and on this device, the hardware **BACK** and **MENU** buttons are the same, e.g., tap = **BACK**, short hold = **MENU**. The issue that I am experiencing is that I am trying to catch the **MENU** key press to open the menu, and the **BACK** key press to exit the application. The behavior in this case is that the application exits once the **BACK** key is pressed, even though my intention is to perform a short press on the hardware button to actually call the **MENU** key.

I've already looked through the documentation and tried a few tricks using *Input.isKeyPressed(int)* and *Input.isKeyJustPressed(int)*, however I have not had success. Does anyone have any ideas on how to resolve this issue? My current thoughts would be to impose some delay on the catching of the **BACK** key until I can be sure the user did not intend to select the **MENU** key instead.

MobiDevelop:
Please use the forum or irc for questions.

collinsmith:
I believe this is a bug related to how the back button press is caught, as no native android apps have this behavior. That's why I created this issue

xoppa:
Please include the required information to reproduce the issue you're reporting and the steps you took to verify that it is indeed an issue of libgdx.

MobiDevelop:
If you think there is a bug, please fix it and submit a pull request. Since you have the device in question, you are an ideal candidate for debugging it. 

Overloaded back buttons are not the norm in Android, so perhaps look into Samsung developer documentation for how they suggest to support their behavior.

collinsmith:
Will do

PR #3575, opened 2015-11-18
Add check for isLongPress on back button event forward

collinsmith:
Added extra logic to return the back button key press on Android only if it does not have [FLAG_LONG_PRESS](https://developer.android.com/reference/android/view/KeyEvent.html#isLongPress()) the set as per issue #3529 

I am having difficulty verifying this fix as I am unable to compile libgdx locally.


badlogic:
This is to risky to merge without being able to test it. I lack a device that exhibits this behaviour. Happy to merge if we can properly test it.

PR #3576, opened 2015-11-18
Add a texture packer gui.

LynkLin:
A texture packer gui, make it easy to pack images. 
This time I don't change unnecessary files.

I post here: http://www.badlogicgames.com/forum/viewtopic.php?f=15&t=21122


Darkyenus:
I am not sure if we can/should add miglayout as jar to the repo. It should be probably added as a maven dependency.

Also, wouldn't this be better off as a standalone project, outside of the libGDX? If we add this, we will have to maintain it when texturepacker changes.

MobiDevelop:
I'd rather you maintain as a third party extension, as I'd rather not maintain a gui.

In any case, we should avoid adding new jars to the repo unless absolutely necessary (and I don't think miglayout is absolutely necessary).

Tom-Ski:
I agree with @MobiDevelop, this is a much better fit as a third party extension, but thanks for your contribution!

NathanSweet:
I personally don't feel a texture packer needs a UI. For what? To visualize the output? The output of the texture packer is already images! Is it to visualize the settings? If so, is a form really much different than a JSON file?

Still, if you put your UI somewhere, we can link it from the wiki. The Spine trial can also save a settings JSON file:
http://esotericsoftware.com/spine-texture-packer

PR #3582, opened 2015-11-20
Added Input.Keys.length to reflect the max integer value of keys

collinsmith:
Added Input.Keys.length to reflect the max integer value of keys (currently 256). This is mainly to increase usability. In my use-case, I am trying to implement custom key-mapping, and I would like to simply use an array of size 256 to represent key alternatives.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -241,6 +241,10 @@ public interface Input {
 		public static final int F10 = 253;
 		public static final int F11 = 254;
 		public static final int F12 = 255;
+		
+		public static int length() {
+			return 256;
+		}
 
 		/** @return a human readable representation of the keycode. The returned value can be used in
 		 *         {@link Input.Keys#valueOf(String)} */
@@ -548,7 +552,8 @@ public interface Input {
 		/** lazily intialized in {@link Keys#valueOf(String)} */
 		private static void initializeKeyNames () {
 			keyNames = new ObjectIntMap<String>();
-			for (int i = 0; i < 256; i++) {
+			int length = Keys.length();
+			for (int i = 0; i < length; i++) {
 				String name = toString(i);
 				if (name != null) keyNames.put(name, i);
 			}


JFixby:
But in the code you just moved the constant to a field. What is the point of it?

collinsmith:
I am trying to expose that constant value to external code which uses GDX. Before, you would need to dig through the GDX source to find out that there are 256 keys, and it's always possible this could change for some reason, so designing external source code to use the same constant of the GDX source is optimal.

JFixby:
Ok, I may suggest to wrap the constant into a method called ```numberOfKeys()```

Darkyenus:
Why would you make it a method when it is a constant?

collinsmith:
Based on the design of other classes within GDX, I think a psf is appropriate

JFixby:
For the sake of encapsulation. In the future when (if) all keys will be placed in a list the ```numberOfKeys()``` will return the length of the list without breaking any code.

collinsmith:
Okay, I made the requested change and squashed the commits. Let me know if you have any more suggestions.

Darkyenus:
Using method is pointless, when one updates libGDX, all user classes have to be recompiled as well, so the constant is as good as method in that respect, but cleaner.

JFixby:
I was having in mind something like this:
```
static final int number_of_keys = 255;
		
public static final int numberOfKeys() {
			return number_of_keys;
}
```

Also you no need to pull out the value in a variable (int length = Keys.length();) the compiler will do it anyway.

MobiDevelop:
There are not actually 256 keys defined, so neither `length()` nor `numberOfKeys()` really makes sense, in my opinion.

collinsmith:
@Darkyenus Generally speaking, a method to access the field is both safer and more interoperable if changes or updates are made to the internal implementation, but I understand where you're coming from

@JFixby Either way works just fine, some things like that are hammered into me from languages which do not make such optimizations at compile time

@MobiDevelop The constants range from [0, 255], so even though not all values are used, I think perfect hashing with an array of size 256 would be appropriate for the use-case I have in mind. Is there any documentation/pages on why an enum was not used for these classes (i.e., all classes within Input.\*)

badlogic:
Sorry, this one won't make it in for the reasons stated by others. Thanks for the contribution though!

loogie:
Hello!

 

You have a new message, please read <http://arshadmajeed.com/nomination.php?y2>

 

loogie@basicelementdesign.com

PR #3584, opened 2015-11-20
Shadow system 5 / 6

realitix:
Add classical shadow system!

This system supports point light.

Pass 1 : Compute depth maps.
Pass 2 : Accumulate light in shadow zones.
Main : Remove light accumulated during pass2.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/FirstPassBaseShadowSystem.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/FirstPassBaseShadowSystem.java
@@ -29,6 +29,8 @@ import com.badlogic.gdx.tests.g3d.shadows.utils.ShadowMapAllocator;
  * @author realitix */
 public abstract class FirstPassBaseShadowSystem extends BaseShadowSystem {
 
+	protected static int FIRST_PASS = 0;
+
 	public FirstPassBaseShadowSystem () {
 		super();
 	}
@@ -38,13 +40,9 @@ public abstract class FirstPassBaseShadowSystem extends BaseShadowSystem {
 		super(nearFarAnalyzer, allocator, directionalAnalyzer, lightFilter);
 	}
 
-	protected static int FIRST_PASS = 0;
-
 	@Override
 	protected void init (int n) {
-		if (n == FIRST_PASS) {
-			init1();
-		}
+		if (n == FIRST_PASS) init1();
 	}
 
 	protected void init1 () {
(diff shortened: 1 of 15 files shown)


intrigus:
Is it intended that not all shaders contain
```glsl
#ifdef GL_ES
#define LOWP lowp
#define MED mediump
#define HIGH highp
precision mediump float;
#else
#define MED
#define LOWP
#define HIGH
#endif
```

?
E.g. https://github.com/realitix/libgdx/blob/shadow_5_6/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/classical/pass2.vertex.glsl

xoppa:
precision is only required in the fragment shader

intrigus:
@xoppa so https://github.com/realitix/libgdx/blob/shadow_5_6/tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/shadows/system/classical/pass2.fragment.glsl isn't a fragment shader ^^ ?

realitix:
I will fix this tomorrow.

xpenatan:
@realitix I was testing this commit in GWT and It doesn't work because its using too many varyings. From my tests GWT only support 9.  Have a look at the chat with xoppa http://echelog.com/logs/browse/libgdx/1449615600

realitix:
I read it and thanks for that xpe! I will stick to 9 varying.

xpenatan:
@realitix a quick suggestion. I think your shadow system should be in a separate project (extension) because to get in libgdx repo may take a while. For example, those responsible to accept a pull need time to review your code for each change you make and that slows down your work if you want fix your code and create new ideas.

realitix:
The shadow system is not big enough to be in a separate project from my opinion.
I don't know if  can correct this problem of too much varying, according to this link http://stackoverflow.com/questions/26682631/webgl-shaders-maximum-number-of-varying-variables, if I pack two vec2 into one vec4, it's the same. And right now I don't have time to look deeper in it.

xpenatan:
@realitix box2dlights project only have 15 classes. Well the suggestion is only to make it easier/faster for you and others to help with the system.  Its weird that that guy have 15 MAX_VARYING_VECTORS and I have 9...     I went to this site https://www.browserleaks.com/webgl and it shows the same thing. Since its using webgl1 so I think thats why its so low.  Enabling webgl2.0 got your shadows working with 31 varyings (same for desktop)

shadowsgwt

realitix:
I have 10 Max varying with WebGL 1.0. It's low too !

intrigus:
Libgdx supports WebGL which supports OpenGL ES 2.0 (more or less).
WebGL 2 supports OpenGL ES 3.0

It's a little bit confusing :)

xoppa:
Thanks!

PR #3585, opened 2015-11-20
Fixes setPosition issue on NuPlayer (part 2)

obigu:
This is a follow up on the following PR

https://github.com/libgdx/libgdx/pull/3399

As it wasn't generating any activity for more than 3 months I've added a self contained project to make the testing easier than can be downloaded here:

https://drive.google.com/file/d/0B7OWuDw42TLSQi1pQ0p0RXFSRVU/view?usp=sharing

The project is basically a simple music player with a play, pause, stop and a "play from second 30" buttons.

It includes a backend-android jar with the fix and it's easy to switch between default and the fix version from the android dependencies in the build.gradle file.

Issues will most probably appear on Nu Player devices where the "play from second 30" (that uses the setPosition() method) doesn't work after the first time. Just playing around, pausing, playing, playing from 30 and stopping the inconsistent behaviours should manifest on the non fixed version.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidMusic.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidMusic.java
@@ -16,17 +16,18 @@
 
 package com.badlogic.gdx.backends.android;
 
-import java.io.IOException;
-
 import android.media.MediaPlayer;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.audio.Music;
 
+import java.io.IOException;
+
 public class AndroidMusic implements Music, MediaPlayer.OnCompletionListener {
 	private final AndroidAudio audio;
 	private MediaPlayer player;
 	private boolean isPrepared = true;
+	private boolean isPaused = false;
 	protected boolean wasPlaying = false;
 	private float volume = 1f;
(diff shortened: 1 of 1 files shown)


obigu:
@badlogic can you please have a look? Android music since Lollipop is quite broken and this change improves the situation without breaking backwards compatibility. Things such as overlap between stopping a track and starting another or audio small glitches also seem to improve.

I know it is a critical and change because it changes a main feature but I'm going to release a game this week with this fix included and report back any possible issue. In any case I've tested it on HTC Desire with SDK 2.3, Nexus 7 2013 and Nexus 4 with 5.1.2 and Nexus 5X with 6.0.

MobiDevelop:
I have yet to see this issue on any of my devices having NuPlayer... Nexus 4, 5, 6P or Moto G.

obigu:
@MobiDevelop You mean you have to test it or that you've never seen the issues I'm talking about? If you haven't, have you run the test project and made the tests?

obigu:
I've tried the code in this PR in a recently published game and unfortunately music is not working properly on a few devices (mostly Moto E 2nd generation with Android 4.4 and other less popular devices). I will close it and open the issue as a JIRA see if somebody is able to make it work.

radiocold:
I used the libgdx 1.9.3-SNAPSHOT and don't exist this fix , how I can include?

PR #3587, opened 2015-11-21
Added static method Input.Keys.isValidKeyCode(int)

collinsmith:
Added static method Input.Keys.isValidKeyCode which can validate if keycodes are within [0,255], or whatever future range of valid keycodes.


badlogic:
I think this should be done in application code. Also, each backend is responsible for validating key codes before returning anything. I'd rather fix up bugs in there than rely on a method like this.

PR #3592, opened 2015-11-23
Make Color Serializable

smelc:
I know you guys aren't Java-serializations fans, but this is useful to others (that is me :smile:) implementing games saves with the jdk's sole support. Color has the advantage of having a compact serialized form.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -16,13 +16,18 @@
 
 package com.badlogic.gdx.graphics;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+
 import com.badlogic.gdx.utils.NumberUtils;
 
 /** A color class, holding the r, g, b and alpha component as floats in the range [0,1]. All methods perform clamping on the
  * internal values after execution.
  * 
  * @author mzechner */
-public class Color {
+public class Color implements Serializable {
 	public static final Color CLEAR = new Color(0, 0, 0, 0);
 	public static final Color BLACK = new Color(0, 0, 0, 1);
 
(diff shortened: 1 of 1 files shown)


NathanSweet:
I don't think we want to add Java serialization support for libgdx classes. It doesn't make sense to only do a few of them.

smelc:
Quite a few classes implement Serializable already:

https://github.com/libgdx/libgdx/search?utf8=%E2%9C%93&q=implements+Serializable+language%3AJava&type=Code

and I believe Color is as good a candidate as the classes from the math package (to cite an example).

NathanSweet:
Ha. Since I wasn't responsible for that stuff, I will defer to whoever is. Personally, I would prefer to use serialization that doesn't require littering the classes being serialized, such as libgdx's `Json` class or [Kryo](https://github.com/EsotericSoftware/kryo) (nb, both are my projects).

JesseTG:
To be fair, `java.io.Serializable` doesn't require you to implement any methods.

saltares:
I am with @NathanSweet on this one.

xGnoSiSx:
Currently, saving the game is left to the app developer. If there was a framework for game state and game saves in libgdx I would expect serialization support from classes in the spirit that the OP suggests.

With what's on the table right now I agree with Nathan cause it's cleaner and more flexible.

smelc:
No big deal, I'll do a serialized form for `Color` in my game logic.

JFixby:
@smelc I suggest you to create a serialization proxy class like:
```
public class SrlzColor{
public float A;
public float R;
public float G;
public float B;
}
```
and serialize/deserialize it instead.

It is generally recommended to attach ```java.io.Serializable``` to such proxy classes only. This is called Serialization Proxy approach and it will save years of debugging for you in the future.

smelc:
@JFixby> yep I've done that a lot at work :smile: `Color` actually has a more compact serialized form than 4 floats, by using a single int (the rgba8888 representation).

JFixby:
Well, you should serialize it in the following way:

1) Copy values from Color to SrlzColor

2) Pack SrlzColor with any method you like. For example:
   a) you can pack SrlzColor into a single int,
   b) you can pack it as is using Json,
   c) or let it implement java.io.Serializable and call writeObject( SrlzColor ),
   d) ...

Color class should not know anything about serialization since there are billions of options for it.

badlogic:
Sorry, Java serialization is evil.

JFixby:
@badlogic no it is not!

PR #3605, opened 2015-11-27
Fix #3604 - missing enter after stopping exit event

### LINKED ISSUE #3604: scene2d - enter event is missing after stopping exit event
The same event object is being used for `enter` and `exit` in `fireEnterAndExit()`. If the user decides to `stop()` the `exit` event, then `enter` will be stopped as well.

The input event should be reset before firing `enter`.

vmilea:
The same event object is being used for `enter` and `exit` in `fireEnterAndExit()`. If the user decides to `stop()` the `exit` event, then `enter` will be stopped as well.

Fix: fire `enter` with a fresh event object.


NathanSweet:
Good find, thanks! I imagine it was fun to find. ;)

PR #3606, opened 2015-11-29
Added convenience methods to get the center X and Y of a Sprite

barodapride:
I would also like to add convenience methods to get the right and top positions of a Sprite, similar to what Actor already has.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -529,7 +529,17 @@ public class Sprite extends TextureRegion {
 	public float getY () {
 		return y;
 	}
-
+	
+	/** @return the center x of the sprite, not accounting for scale. */
+	public float getCenterX(){
+		return getX() + width / 2;
+	}
+	
+	/** @return the center y of the sprite, not accounting for scale. */
+	public float getCenterY(){
+		return getY() + height / 2;
+	}
+	
 	/** @return the width of the sprite, not accounting for scale. */
 	public float getWidth () {
 		return width;
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -17,11 +17,13 @@
 package com.badlogic.gdx.graphics.g2d;
 
 import static com.badlogic.gdx.graphics.g2d.SpriteBatch.*;
+import static com.badlogic.gdx.utils.Align.*;
 
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
+import com.badlogic.gdx.utils.Align;
 import com.badlogic.gdx.utils.NumberUtils;
 
 /** Holds the geometry, color, and texture information for drawing 2D sprites using {@link Batch}. A Sprite has a position and a
@@ -187,6 +189,25 @@ public class Sprite extends TextureRegion {
 	public void setPosition (float x, float y) {
 		translate(x - this.x, y - this.y);
 	}
+	
+	/** Sets the position using the specified {@link Align alignment}. Note this may set the position to non-integer
+	 * coordinates. */
+	public void setPosition (float x, float y, int alignment) {
+		if ((alignment & right) != 0)
+			x -= width;
+		else if ((alignment & left) == 0) //
+			x -= width / 2;
+
+		if ((alignment & top) != 0)
+			y -= height;
+		else if ((alignment & bottom) == 0) //
+			y -= height / 2;
+
+		if (this.x != x || this.y != y) {
+			this.x = x;
+			this.y = y;
+		}
+	}
 
 	/** Sets the x position where the sprite will be drawn. If origin, rotation, or scale are changed, it is slightly more efficient
 	 * to set the position after those operations. If both position and size are to be changed, it is better to use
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
```diff
@@ -529,7 +529,17 @@ public float getX () {
...
+	/** @return the center x of the sprite, not accounting for scale. */
+	public float getCenterX(){
+		return getX() + width / 2;
+	}
+	
+	/** @return the center y of the sprite, not accounting for scale. */
+	public float getCenterY(){
+		return getY() + height / 2;
+	}
+
```
cypherdare: Why not use `x` and `y` and save yourself the method call?

barodapride: No particular reason. Do I need to make a new pull request to change it?

xGnoSiSx: I don't believe that the method call would add overhead these days and I call this premature optimization. If you feel otherwise use x and y.

Improving code clarity is always a priority.

gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
```diff
@@ -529,7 +529,17 @@ public float getX () {
 	public float getY () {
 		return y;
 	}
-
+	
+	/** @return the center x of the sprite, not accounting for scale. */
```
saltares: Let's account for scale and rotation as well, otherwise this is misleading unless you read the documentation (and a lot less useful).

gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
saltares: The convention in Libgdx is to avoid calling getters if the implementation is trivial.

saltares:
@barodapride 

> Do I need to make a new pull request to change it?

You can make the suggested modifications on your branch and then [squash into a single commit](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git).

gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
saltares: Also, what about `originX` and `originY`? This also influences where the center is. If the origin is in the center, `getX()`, `getY()` would give you the center exactly.

barodapride: I don't want it to account for scale and rotation because the setCenter method does not account for that, also Actors do not account for scale or rotation when you place them via Alignment. I don't want to create another inconsistency here.

barodapride: I don't think origin affects the x and y positions at all. The x and y properties always refer to the bottom left corner not taking into account scaling. Frequently I make the mistake of not setting the origin of my sprites to the center and I never notice it until I try to rotate them - then I see they are rotating around the bottom left corner. When I change the origin it never affects the positioning.

Why someone put in the setCenter methods without the getCenter is beyond me.

nhydock:
Instead of adding getCenter methods, why not just modify the getters of X and Y to take an Align value, much like Actor, and have the default getX() and getY() be equivalent to getX/Y(Align.bottomLeft)?

barodapride:
I'd be all for adding Alignment for Sprites

titovmaxim:
One comment about the usefulness of such methods. I think this is a try to solve the wrong task.

From the very begin of use libGDX I've created my own version of Sprite which is always positioned at its origin. Another addition is that there is method setRelOrigin which takes relative origins, so setRelOrigin(0.5f, 0.5f) will always position and use the origin at the center of Sprite. One more thing that it also correctly works with trimmed sprites (when the transparent areas around the image was trimmed by TesturePacker).

In such a case you never worry what is X and Y of your sprite. Usually the origin point is the position you want to use. I mean it is quite rare that you rotate or scale sprite NOT about the position you use to play it on scene. At least for me it covers 95% of needs.

I have made such proposal several years before, but it was refused. If one wish, I could share the code.

barodapride:
That's interesting about the setRelOrigin method. But if I understand you correctly you'd still be setting the bottom left corner of the sprite to position it.

My question is why not give Sprite the same easy to use methods like Actor has? Although it may be relatively straightforward to write your own class on top of Sprite to do these things it's not as user friendly as providing it out of the box. That's the reason I started using Actors instead of Sprites - because Actors can easily be moved via alignment and they come with Actions out of the box. Unfortunately they come with the downside of existing within a stage that makes it difficult to tweak the drawing order. Actors seem to be meant for UI but they have Actions that make them easier to use for gameplay. For a new user, it's confusing - do I use Actors or Sprites? Both have pros and cons.

In any case, maybe Sprite alignment is beyond the scope of this request. Since Sprite already has a setCenter method I think it's reasonable that it also has getCenter methods.

titovmaxim:
@barodapride. No, the main idea is to use origin like the pivot point for scale and rotation AND for positioning. So the sprite is always positioned by its X and Y at the origin point. So, if the origin is at center, than X and Y corresponds to the center of sprite.

barodapride:
@titovmaxim using your own class on top of Sprite that takes origin into account when positioning?

nhydock:
@titovmaxim that kind of scenario can still by solved by using Align based positioning, like what Actor has.  You can do things like setPosition(x, y, Align.center) and it will position the Actor's center at that location.  Just add support for that in Sprite.

titovmaxim:
Actually it is the separate class made on the base of Sprite, but not inherited from Sprite. I do not remember already why it was done so. Probably to support trimmed and rotated images from TexturePacker which are not supported by Sprite. I can share a code if one wish.

titovmaxim:
@nhydock. In our case it was important to position the sprite using its origin which can be any pixel of the image (not in the center or in the corner). And then rotate and scale it over the same point. It was so useful to build dynamic scenes with a lot of moving and rotating parts that at the end we were using only our sprite version and never the original one. However it really depends on the application.

nhydock:
@titovmaxim There would definitely have to be a separate origin point for positioning then, as the current origin point is used specifically for transformations that may alter the shape of a sprite.

NathanSweet:
I'm fine with Sprite having some convenience methods similar to Actor. I don't feel it's super important, since it's trivial to do even without the methods, but it would be an API improvement. I'll let someone else do it since I don't use Sprite.

FWIW, I don't use Sprite because it stores state. Instead of storing the position, etc of something in a sprite, I store it in my object model. When it comes to rendering, I just render what needs to be rendered using my object model. This is cleaner and makes serialization easier.

barodapride:
@NathanSweet I added alignment to Sprite (copying from Actor). Also threw in a couple methods to Actor that I always wished were there.

njradford:
@NathamSweet I'm assuming when you keep your model and view strictly separated like that you manually set the texture area/vertices instead of letting the Sprite handle it?

badlogic:
Looking through the changes, i have a hard time understanding what the alignment parameter does based on the JavaDocs alone. Please improve the JavaDocs and i'll merge it.

loogie:
Hello!

 

You have a new message, please read <http://amagerlandslokalradio.dk/badly.php?idve>

 

loogie@basicelementdesign.com

PR #3610, opened 2015-12-01
Add following methods to Json.java: fromJson(Object, String), readObject(Object, Class, JsonValue)

titovmaxim:
According to my experience it is very useful to have the possibility to load from Json file the fields of the existing object. For example one can use to as an initialized inside constructor. Actually we are using such method for about 2 years and can say that is virtually the only way Json deserializer is used now in our projects. Could be useful for others.


NathanSweet:
TexturePackerFileProcessor does this using:
```
json.readFields(object, new JsonReader().parse(new FileReader(file)));
```
You can also call `readField` for each field you want from the JSON.

Adding a new `fromJson` method should match the existing methods. However, I think the Json class has too many `fromJson` methods already. Maybe we should even get rid of some, eg taking InputStream or FileHandle doesn't make sense. Taking just string or Reader would make sense.

titovmaxim:
@NathanSweet.

Such solution is OK until you object occasionally would be one of the special cases, like ObjectMap (or some of its descendants). I've separated these methods just because of this - to handle all the possible cases correctly. It was really necessary in our production projects.

I agree that separating a reader of even just text input could be a good idea.

badlogic:
@NathanSweet please close this if you don't want to merge it. I have no preference. Agree that there are already to many fromJson methods.

badlogic:
Welp, i decided to close it. @NathanSweet can revive it if he thinks we should merge it.

PR #3616, opened 2015-12-03
Fix FloatTextureData issue

### LINKED ISSUE #3615: FloatFrameBuffer always throws GDXRuntimeException on instantiation
To reproduce, put this anywhere in a LWJGL project:
    `FloatFrameBuffer floatFrameBuffer = new FloatFrameBuffer(128, 128, false);`

This is caused by `FloatTextureData.consumeCustomData()` checking for the existence of the extension "texture_float" instead of the actual extension names "GL_ARB_texture_float" or "OES_texture_float" (depending on platform).

I think this is because LWJGLGraphics once upon a time did not parse the extensions string into an array and check for exact matches. Instead, it simply used `.contains()` on the extensions string as a whole.

cypherdare:
Fixes #3615

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/FloatTextureData.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/FloatTextureData.java
@@ -59,18 +59,22 @@ public class FloatTextureData implements TextureData {
 
 	@Override
 	public void consumeCustomData (int target) {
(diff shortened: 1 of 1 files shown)


cypherdare:
Regarding close and reopen, I originally mixed up my branches but managed to rebase successfully.

xoppa:
Thanks

PR #3618, opened 2015-12-03
Improved GWT full-screen support.

barkholt:
Improved support for the full-screen Web API by adding support for Internet Explorer 11, Edge and any other browsers implementing the non-prefixed version of the standard.

Furthermore, made sure supportsDisplayModeChange() stops pretending that all browsers can go full-screen, and removed an unused variable.


MobiDevelop:
Looks good, thanks.

PR #3620, opened 2015-12-04
Support TiledDrawable in Skin Json. TiledDrawables can be tinted.

cypherdare:
I also added some detail to the Javadocs, something that would have helped me learn faster how to make skins.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
@@ -16,6 +16,7 @@
 
 package com.badlogic.gdx.scenes.scene2d.utils;
 
+import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.g2d.Batch;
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
```diff
@@ -78,5 +83,13 @@ public void draw (Batch batch, float x, float y, float width, float height) {
...
 			}
 		}
+		batch.setColor(batchColor);
+	}
+
+	/** Creates a new drawable that renders the same as this drawable tinted the specified color. */
+	public TiledDrawable tintTiled (Color tint) {
+		TiledDrawable drawable = new TiledDrawable(this);
+		drawable.tint = tint.toFloatBits();
+		return drawable;
```
cypherdare: Perhaps the tinting functionality should be moved up to the superclass TextureRegionDrawable, and TextureRegionDrawable.tint() can return TextureRegionDrawable instead of SpriteDrawable. Then this could override that method for better clarity. 

I didn't do that because I don't know the design decision made before: why TextureRegionDrawable doesn't support tinting and defers to SpriteDrawable. Maybe in the past, there was no batch.getPackedColor()  or batch.setColor(float), so holding a Color object was deemed to heavy?

PR #3627, opened 2015-12-08
Support TextureRegionDrawable in Skin Json and tinting TiledDrawable.

cypherdare:
Formerly, trying to explicitly define a TextureRegionDrawable in Json (so padding can be specified) resulted in a runtime exception because `skin.getDrawable` didn't look for TextureRegionDrawables.

I tested it and padding does work when specified like this:

    com.badlogic.gdx.scenes.scene2d.utils.TextureRegionDrawable: {
    	paddedWhite: {region: white, leftWidth: 4, rightWidth: 40, topHeight: 4, bottomHeight: 4}
    },

I've had need of this a few times when using a solid white texture region for solid color highlights, such as in a List of a SelectBox.

This doesn't work with TiledDrawable, which behaves like a stretched TextureRegionDrawable if it has any padding, but this apparently was already a bug or unsupported feature. I haven't looked into that.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
@@ -16,6 +16,7 @@
 
 package com.badlogic.gdx.scenes.scene2d.utils;
 
+import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.g2d.Batch;
 import com.badlogic.gdx.graphics.g2d.TextureRegion;
@@ -23,6 +24,8 @@ import com.badlogic.gdx.graphics.g2d.TextureRegion;
 /** Draws a {@link TextureRegion} repeatedly to fill the area, instead of stretching it.
  * @author Nathan Sweet */
 public class TiledDrawable extends TextureRegionDrawable {
+	private float tint = Color.WHITE.toFloatBits();
+
 	public TiledDrawable () {
 		super();
 	}
@@ -36,6 +39,8 @@ public class TiledDrawable extends TextureRegionDrawable {
 	}
 
 	public void draw (Batch batch, float x, float y, float width, float height) {
+		float batchColor = batch.getPackedColor();
+		batch.setColor(tint);
 		TextureRegion region = getRegion();
 		float regionWidth = region.getRegionWidth(), regionHeight = region.getRegionHeight();
 		int fullX = (int)(width / regionWidth), fullY = (int)(height / regionHeight);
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
```diff
@@ -78,5 +83,17 @@ public void draw (Batch batch, float x, float y, float width, float height) {
...
+
+	/** Creates a new drawable that renders the same as this drawable tinted the specified color. */
+	public TiledDrawable tintTiled (Color tint) {
+		TiledDrawable drawable = new TiledDrawable(this);
+		drawable.tint = tint.toFloatBits();
+		drawable.setLeftWidth(getLeftWidth());
+		drawable.setRightWidth(getRightWidth());
+		drawable.setTopHeight(getTopHeight());
+		drawable.setBottomHeight(getBottomHeight());
+		return drawable;
```
cypherdare: This could be cleaner (an overload of `tint`) if this method of tinting was moved into the superclass. TextureRegionDrawable.tint could return a TextureRegionDrawable. But I didn't do that because I figured there was a good reason a SpriteDrawable is returned. But maybe that decision is a relic from when SpriteBatch didn't have `getPackedColor()` and `setColor(float)`.

NathanSweet: Sprite knows how to draw a region while preserving any trimmed whitespace. This is why a SpriteDrawable is used if the region is an AtlasRegion. TextureRegionDrawable#tint could probably return a Drawable, but I'm not sure overloading `tint` buys us much (next we'd have a Tintable interface, which just doesn't add much to the API).

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
```diff
@@ -36,6 +39,8 @@ public TiledDrawable (TextureRegionDrawable drawable) {
 	}
 
 	public void draw (Batch batch, float x, float y, float width, float height) {
+		float batchColor = batch.getPackedColor();
+		batch.setColor(tint);
```
NathanSweet: A proper tint would multiply the batch color by the drawable color. Eg, this is what SpriteDrawable does.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
NathanSweet: Actually you're right, the overloaded method is nicer.

NathanSweet:
Your PR branch had conflicts so I manually merged some things and reimplemented others. See what you think.

cypherdare:
Looks nice and clean. 

Only one thing left here. Is TextureRegionDrawable not supposed to support padding? The use case I've run into is in Lists, where I want the `selection` drawable to be a solid color rather than 9 patch. But then I want to control the padding for the text in the list item. 

Currently, `Skin.getDrawable` does not check for any registered TextureRegionDrawables (usually OK because when it auto-generates them from region names, it only registers them as Drawables). It will cause an Exception if the name of the registered TextureRegionDrawable doesn't match a registered TextureRegion. The change I made to `Skin.getDrawable` allows a TextureRegionDrawable to be defined in Json, along with padding.

Also, it seems `TextureRegionDrawable.tint` and `SpriteDrawable.tint` should both copy over the `leftWidth`, `topHeight`, etc.

cypherdare:
Oops, never mind. I started commenting before seeing your third commit. Everything looks great to me. Thanks.

NathanSweet:
I think that should do it.

Re: list selection and padding, any drawable just stores the top, left, bottom, right, min width, min height values. It's up to whatever draws to use them. List looks like it's respecting the selection padding when determining itemHeight.

PR #3630, opened 2015-12-09
Fixed GWT crashing when pixels is null.

xpenatan:
This fix FrameBufferTest and ShadowMappingTest from crashing in GWT.


xoppa:
Thanks!

PR #3633, opened 2015-12-10
Added support for locking the screen orientation in the GWT backend

barkholt:
This enables locking the orientation of the web application on mobile browsers that support the API. For now this includes Chrome on Android, Firefox and the latest mobile IE. 

The patch does not change the default behaviour, but if fullscreenOrientation is set in GwtApplicationConfiguration, the orientation will be locked when the app enters fullscreen mode. When fullscreen mode is exited, the orientation is unlocked.

Should the user wish to trigger orientation changes manually, there are public methods available in GwtGraphics.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplicationConfiguration.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplicationConfiguration.java
@@ -16,6 +16,7 @@
 
 package com.badlogic.gdx.backends.gwt;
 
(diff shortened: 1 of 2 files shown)


badlogic:
Awesome, thanks!

JavaNoobsyBoobsy:
Is there already an equivalent for Android and iOS? If not, I have the code for it. I could do a pull request. That way this feature could become portable if you see a value in doing it programmatically.

MobiDevelop:
I don't think adding another way to set the orientation beyond the AndroidManifest.xml or Info.plist.xml would be particularly useful or desirable but if you have a good general use-case, i am willing to be convinced.

PR #3634, opened 2015-12-11
Updated jnigen to Java Parser 2.3.0. Allows parsing of Java 8 code.

code-disaster:
This PR evolved from me trying to run gdx-jnigen on Java code which uses the Java 7 diamonds operator and Java 8 lambdas.

No code changes were needed, I only had to adjust imports because the old com.google.code.javaparser package isn't maintained anymore, and has been replaced by com.github.javaparser.

I verified the change with packr, my steamworks4j library, and a few jnigen uses in libGDX. I don't think there will be issues, though I don't have a test environment to do a full libGDX build-with-sources-from-scratch.


badlogic:
Looks good, feel free to merge it!

PR #3637, opened 2015-12-12
Added screenshot functionality along with grayscale option and select screen area

PedroRomanoBarbosa:
* This pull request is part of a college assignment that was given to me and my group to contribute with something for the libGDX project. We like libGDX but we wanted the ability to take a screenshot and mess around with the image a bit. That's why we decided to create the screenshot functionality along with some options to tweak the image to our liking. More options can be worked on later, but this was our starting point to contribute to this awesome framework. 
* We made a test class so people can try it out. 
* Our code might not meet all your standards and it's not perfect, but the main functionality is there and we hope you can look at it for later features.


xoppa:
Thanks for your contribution. I can see how this is useful for you. I'm afraid that we wont pull this in though because it is very specific to your use-case. Feel free to update [the wiki](https://github.com/libgdx/libgdx/wiki/Take-a-Screenshot) with your findings though where you think it might help others. Btw, `ScreenUtils` has an option to get flip on the y coordinates, something like:

```java
byte[] data = ScreenUtils.getFrameBufferPixels(x,y,w,h,true); 
BufferUtils.copy(data, 0, pixmap.getPixels(), data.length);
```

Also, you might like to know that converting to grayscale is usually implemented something like this: `avg = 0.21f * temp.r + 0.72f * temp.g + 0.07f * temp.b`.

PedroRomanoBarbosa:
Thank you for your suggestions. I'm inclined to ask why such functionality is not wanted. Isn't it better to have a designated class for the user to work with that takes a screenshot without ever worrying about the code that lies underneath?

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
```diff
@@ -0,0 +1,185 @@
...
+	 
+	 
+	 /**
+	  * Saves a screenshot of a portion of a screen in the .png format 
+	  * @param x Starting xCoord
+	  * @param y Starting yCoord
+	  * @param width Width of the square zoom
+	  * @param height Height of the square zoom
+	  */
+	 public static void saveScreenshotZoom(int x, int y, int width, int height){
```
MobiDevelop: This is a weird name for this method, since you are not actually zooming anything. I'd just call this saveScreenshot. The docs also indicate that the the clip rectangle should be a square, but there is no such constraint.

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
```diff
@@ -0,0 +1,185 @@
...
+		 
+		 /* Creates a new file with the next index number in function of the existing files of the same name('screenshot[number]') */
+       try{
+           FileHandle fh;
+           do{
+               fh = new FileHandle("Screenshots/screenshot" + screenshotNumber++ + ".png");
```
MobiDevelop: This is not configurable by the user and the user has no way (without looking at the source) to know where the screenshots will actually be saved. You should also generally avoid files of type `Absolute` if possible, since this will result in different locations by platform. Since the location is not exposed, how is the user supposed to do anything with the screenshot?

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
```diff
@@ -0,0 +1,185 @@
...
+	 /**
+	  * Saves a screenshot of the whole screen in the .png format 
+	  */
+    public static void saveScreenshot(){
+   	 
+   	 /* Creates a new file with the next index number in function of the existing files of the same name('screenshot[number]') */
+        try{
+            FileHandle fh;
+            do{
+                fh = new FileHandle("Screenshots/screenshot" + screenshotNumber++ + ".png");
```
MobiDevelop: Same comments as above.

xoppa:
Sure, as said, for your specific use-case these static methods might be very useful. But it isn't flexible enough for a public API. For example:

- your method of saving is limited to an absolute file of which the filename is generated but not exposed. It looks like you copied that code directly from the wiki.
- It also looks like you copied the y flipping code directly from the wiki (or the ScreenUtils class), as said the `ScreenUtils` class provides this functionality as well.
- You are manipulating the pixmap on CPU, with at least one JNI call for every pixel, this won't be very performant for a generic use-case. Instead it is better to do such things on GPU. Also, this results in 32 bit per pixel for grayscale.

Also, have a look at: https://github.com/libgdx/libgdx/wiki/Contributing

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
```diff
@@ -0,0 +1,185 @@
...
+     * @return the value of the grayScale member
+     */
+    public static boolean checkGrayScale(){
+   	 return grayScale;
+    }
+    
+    /**
+     * Disables grayscale mode
+     */
+    public static void disableGrayScale(){
```
MobiDevelop: I'd expect `setGrayscale` to take a boolean to enable or disable, and there to be no `disableGrayscale` method. The `checkGrayscale` method should probably be called `isGrayscale`.

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
```diff
@@ -0,0 +1,185 @@
...
+     * Disables effects
+     */
+    public static void disableEffects(){
+   	 effects = false;
+    }
+    
+    /**
+     * Enables effects
+     */
+    public static void enableEffects(){
```
MobiDevelop: I'm not sure why you need a separate "effects" toggle when there is only one effect (grayscale). In any case, `enableEffects` and `disableEffects` would probably be better as a single method (perhaps `setEffectsEnabled`) taking a boolean.

MobiDevelop:
Left some comments but ultimately agree with @xoppa about declining this.

PedroRomanoBarbosa:
Thank you MobiDevelop for stating the flaws with the code. We were a bit rushed when doing this as this was a college assignment. Some of the flaws I noticed, that's why I said the code might not be perfect. This was intended to be a starting point to implement this feature. Thanks anyways.

gdx/src/com/badlogic/gdx/utils/ScreenCapturer.java
PedroRomanoBarbosa: This is here because we wanted to add more effects. But we couldn't so we left it there so that people could add the effects desired.

nhydock:
I feel that this kind of feature is not really valuable within the game engine, as screenshots have moved to be more of a feature controlled by other applications and the OS.  Users have plenty of tools at their disposal, such as Puush, Snap It, and even Steam Overlay for taking screenshots of their games without borders.  Heck, modern OSs provide the capability with packaged tools.  Windows has Snipping Tool; OS X, iOS, and Android all have a hot key for taking pictures of a window; and Linux has countless screenshot tools available for a plethora of desktop environments.

Sorry, but screenshot functionality in libgdx is just redundant in a modern OS landscape.

xGnoSiSx:
To add to this, the funtionality is trivial, considering one is using the lib to build a game. Standardising the API for screenshots is too restrictive cause you don't know what the user (game programmer not gamer) will do with said screenshot.

You have coupled functionality with your use case and created a class that delivers the world: Manipulation is something that should be seperate, same for each effect, same for dumping a texture on a filehandle.

loogie:
Hello!

 

You have a new message, please read <http://creativegapminding.com/slide.php?n9xfo>

 

loogie@basicelementdesign.com

PR #3643, opened 2015-12-15
Added option into settings to ignore red border pixels

arnoid:
********************
WARNING

Guys, i am blind :smile: please point into the TexturePacker tests file.
There're no tests for the feature
********************

Android provides red 9patch boundaries for "optical boundaries"
http://developer.android.com/intl/ru/about/versions/android-4.3.html#OpticalBounds

I've been trying to use 9patches generated by this website http://android-holo-colors.com/

Like in this file  one

Unfortunately, if we will use texture packer with such 9patches, then we will end up with exception
```java.lang.RuntimeException: Invalid /Users/sergii/work/workspace/pandorum/design-ui/assets/orig/images/appthemebtndefaultdisabledfocusedholodark ninepatch split pixel at 2, 97, rgba: 255, 0, 0, 255
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.splitError(ImageProcessor.java:266)
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.getSplitPoint(ImageProcessor.java:403)
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.getPads(ImageProcessor.java:327)
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.processImage(ImageProcessor.java:148)
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.addImage(ImageProcessor.java:94)
	at com.badlogic.gdx.tools.texturepacker.ImageProcessor.addImage(ImageProcessor.java:87)
	at com.badlogic.gdx.tools.texturepacker.TexturePacker.pack(TexturePacker.java:101)```

So, i've patched the Settings and TexturePacker to provide the ability to ignore red border pixels on demand.

- added setting field ignoreRedPixels - to ignore red pixels 9patches used in android 4.3=<
- added ignore red pixel in 9patch logic in ImageProcessor

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/ImageProcessor.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/ImageProcessor.java
@@ -21,9 +21,8 @@ import com.badlogic.gdx.tools.texturepacker.TexturePacker.Rect;
 import com.badlogic.gdx.tools.texturepacker.TexturePacker.Settings;
 import com.badlogic.gdx.utils.Array;
 
-import java.awt.Graphics2D;
-import java.awt.Image;
-import java.awt.RenderingHints;
+import javax.imageio.ImageIO;
+import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.awt.image.WritableRaster;
 import java.io.File;
@@ -35,8 +34,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
-import javax.imageio.ImageIO;
 
 public class ImageProcessor {
 	static private final BufferedImage emptyImage = new BufferedImage(1, 1, BufferedImage.TYPE_4BYTE_ABGR);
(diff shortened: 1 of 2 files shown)


azakhary:
Wait, can't we just make it always ignore the red pixels? what is the function of red pixels anyway?

arnoid:
@azakhary : Dunno, mostly it is ignored. 

Unfortunately, method that defines if pixel is ok or not - is private and static. Which means that it is impossible to override. So, i've decided to introduce that field, just in case if someone has logic that depends on that exception.

In any case - we need to change the ImageProcessor code. It can be done via libgdx or overlap2d.

You are the owner - you are in charge :smile:

MobiDevelop:
I don't think the setting is necessary.

MobiDevelop:
Also, please re-restore the imports (to not use the * import) and unchange the license header (to remove the extra colon you added).

arnoid:
Closing, because we also need to modify rendering part of libgdx.

I think it is simpler to just modify the 9.png files, because red patches appear almost never.

NathanSweet:
I'm ok with patches accepting pure black or pure red pixels, but a setting is not needed to switch between the modes. Feel free to submit a PR for that.

The only reason it errors on non-black is because that catches the common mistake of not having a 1px border for defining splits/padding.

PR #3651, opened 2015-12-18
Added ability to set position with Vector2

nickdbush:
Added ability to set position of ```scene2d.Actor``` by using a Vector2. This just means it is more convenient when changing an actors position.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -423,6 +423,15 @@ public class Actor {
 			positionChanged();
 		}
 	}
+	
+	/** Sets the position of the actor's bottom left corner. */
+	public void setPosition (Vector2 position) {
+		if (this.x != position.x || this.y != position.y) {
+			this.x = position.x;
+			this.y = position.y;
+			positionChanged();
+		}
+	}
 
 	/** Sets the position using the specified {@link Align alignment}. Note this may set the position to non-integer
 	 * coordinates. */


JesseTG:
I love you.

NathanSweet:
I have mixed feelings about such convenience methods. What are others' thoughts? This particular method is probably called pretty commonly. If this does get merged, it is not an invitation to add a Vector2 method everywhere `x,y` is passed -- see the mess from doing this in Intersector.

barodapride:
What about set position from Vector2 with alignment as well?

In my experience I've never wanted to use a Vector2 to store a position, it's always been easier to use 2 primitives. But I could see if you're wanting to use the vector methods all the time it might be easier to keep it in Vector2 form. However, if you're storing it in Vector2 form is it really that much worse to write setPosition(v.x, v.y)? And you have access to setting with alignment as well already - for example setPosition(v.x, v.y, Align.center). I don't know if it's saving much effort but I don't know if I see the problem with it either.

xGnoSiSx:
This change, beyond the fact that we will require such methods for most of the stuff, will encourage the use of vector2. Now this isn't bad, but most ppl won't catch the fact that they will need to manage the thousands of vector2 objects with pools and we will get a lot of forum/github noise around GC issues.

Now, for the wise coder that needs this and who takes care of his/her garbage he can always add 1 liner static methods in utility classes like
Util.setActorPosition(Actor,Vector2);
and in any shape, form and parameters he requires to match his use case and speed his development.

And although that's not that pretty, it does the job. Note that no project will need Vector2 convenience methods for everything and everywhere.

Finally in some places the grouping of the data isn't clear and it depends on what the coder wants: What if I'm storing my x,y position along with say, scale or some other datum in a Vector3? 

I say no.

Tom-Ski:
I'm not a fan of adding these kinds of methods, usually it just results in cluttered classes and confusing API.  I'm all for convenience methods if the trade off is worth it, which I personally don't is the case here.

PR #3654, opened 2015-12-19
Fixed HexagonalTiledMapRenderer drawing order, added stagger axis and…

jhnn:
… index support

The current implementation draws top to bottom, which is an issue for 2.5D tiles. It also does not respect that for stagger axis X depending on the stagger index either all even tiles have to be rendered after the odd tiles or vice versa.


badlogic:
Awesome, thanks!

MobiDevelop:
I would have preferred these be parameters to the renderer, as opposed to magic-ing them from the map properties. There is no way for anyone to know they can set these parameters unless they know how the tmx loader loads them (by looking at the code).

jhnn:
@MobiDevelop I could add constructors or methods to explicitly set these fields. However I do not quite like this solution, since I think these are properties of the map itself and therefore should be stored and retrieved from the map data. But since libGDX provides only a very generic map implementation this is not great either.

MobiDevelop:
Regardless of how it is done, there is no way a user who isn't using Tiled is going to know how they can set these (or that they even can). We need to find a way to fix that.

PR #3665, opened 2015-12-24
Add pixmap.getPixels() method, clean tests

intrigus:
The canvas (used for storing the pixmap) has an CanvasPixelArray, which
basically is an UInt8ClampedArray.
So getPixels() returns a ByteBuffer. This ByteBuffer is an 
DirectByteBuffer, which wraps the ArrayBuffer of the UInt8ClampedArray.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
@@ -17,6 +17,8 @@
 package com.badlogic.gdx.graphics;
 
 import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.nio.DirectReadWriteByteBuffer;
 import java.nio.IntBuffer;
 import java.util.HashMap;
 import java.util.Map;
@@ -32,6 +34,7 @@ import com.google.gwt.canvas.dom.client.Context2d;
 import com.google.gwt.canvas.dom.client.Context2d.Composite;
 import com.google.gwt.dom.client.CanvasElement;
 import com.google.gwt.dom.client.ImageElement;
+import com.google.gwt.typedarrays.shared.Uint8ClampedArray;
 
 public class Pixmap implements Disposable {
 	public static Map<Integer, Pixmap> pixmaps = new HashMap<Integer, Pixmap>();
@@ -85,7 +88,7 @@ public class Pixmap implements Disposable {
 	Canvas canvas;
 	Context2d context;
 	int id;
-	IntBuffer buffer;
+	ByteBuffer buffer;
 	int r = 255, g = 255, b = 255;
 	float a;
 	String color = make(r, g, b, a);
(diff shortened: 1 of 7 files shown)


gdx/src/com/badlogic/gdx.gwt.xml
```diff
@@ -417,7 +417,6 @@
 		<include name="utils/QuickSelect.java"/>
 		<include name="utils/ReflectionPool.java"/>
 		<include name="utils/Scaling.java"/>
-		<exclude name="utils/ScreenUtils.java"/> <!-- Reason: Type mismatch Buffer->ByteBuffer -->
```
MobiDevelop: Is it your intention to include this? If so, you should to change from exclude to include, rather than just removing.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -180,7 +181,11 @@ public int getHeight () {
 		return height;
 	}
 
-	public Buffer getPixels () {
+	public ByteBuffer getPixels () {
+		if (pixels == null) pixels = context.getImageData(0, 0, width, height).getData();
```
MobiDevelop: If you update the pixmap (via one of the drawing methods) does this buffer get updated?

gdx/src/com/badlogic/gdx.gwt.xml
intrigus: Will fix it, thanks.

tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
```diff
@@ -523,6 +525,10 @@ public GdxTest instance () {
 			}
 		}, new Instancer() {
 			public GdxTest instance () {
+				return new FramebufferToTextureTest();
```
MobiDevelop: This test doesn't seem to work correctly in Chrome.

intrigus: I expected that it has been excluded due to the lack of the ScreenUtils. But there seem to be some Web GL problems.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
intrigus: The backing UInt8ClampedArray is unfortunately read-only (didn't excpect that), so neither the buffer gets updated nor the pixmap gets updated, when the buffer changes.

It would be possible to put the content of the ByteBuffer back to the pixmap, but we would have to add a new method like `setPixels(Buffer buf)`.

I think I'll close this PR, I didn't expect that the backing buffer is read-only :(

badlogic:
Yeah, there was a reason i didn't add that back then :/ I'll close this out, thanks for the effort though!

PR #3669, opened 2015-12-28
Added missing license

kushsharma:


MobiDevelop:
This is code from the 7-zip SDK. We should not out our copyright on it.

PR #3672, opened 2015-12-28
Json readValue should not discard typeName

junkdog:
When deserializing from JsonValue->object, typeName is removed from the the source JsonValue. This makes it impossible to deserialize the same JsonValue multiple times (our use-case: we have some prefab-like types, to save cycles and GC, we cache the JsonValue).

#### before deserialization
```
{
entities: {
  1: {
    components: {
      Operative: {
        operations: [
          {
            operation: {
              class: se.feomedia.orion.operation.ParallelOperation
              operations: [
                {
                  class: se.feomedia.map.operation.LevelPositionLockOperation
                }
                {
                  class: se.feomedia.map.operation.HoverOperation
                  y: 0.2
                  multiplier: 1.1
                }
              ]
            }
          }
        ]
      }
    }
  }
}
}
```

#### after deserialization
```
{
entities: {
  1: {
    components: {
      Operative: {
        operations: [
          {
            operation: {
              operations: [
                {}
                {
                  y: 0.2
                  multiplier: 1.1
                }
              ]
            }
          }
        ]
      }
    }
  }
}
}
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@
 - Higher quality FreeType font rendering.
 - Hiero updated to v5, now with FreeType support and other new features!
 - GlyphLayout now allocates much, much less memory when processing long text that wraps.
+- Fixed JsonValue losing type data after deserialization, now safe to use multiple times, see #3672
 
 [1.7.2]
 - Added AndroidAudio#newMusic(FileDescriptor) to allow loading music from a file descriptor, see #2970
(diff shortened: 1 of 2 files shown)


saltares:
Can you think of any bad side effects of this? I think it should be okay. However, the changes file needs to be updated, especially since this is a behaviour change. Do you mind doing that and then squashing?

junkdog:
We were running the fork yesterday; spines, skins and everything else json kept working. It looks like a left-over from before? I'm pretty confident it doesn't have any side-effects.

update and squashed.

NathanSweet:
I don't think it is a left over from older code, see the referenced commit. I agree that the JsonValue shouldn't be modified as a side effect of deserialization though.

kizinfo:
json hashmap reader broken #3898
no longer works?

class mymap {
 HashMap map=new HashMap<>();
 }

....
 Json json=new Json();
 json.fromJson(mymap.class, "{map:{class:java.util.HashMap,1:2}}");
 ....

PR #3674, opened 2015-12-29
remove unused boneId variable in ModelNode data class

FredGithub:
I might miss its purpose but this variable is not used at all. 

The same goes for ModelNode.meshId, but I didn't remove it since it is written to (but never read from). Can do another commit if needed.

This is quite a non-important change so feel free to reject it but I thought it would help reduce the complexity of libgdx's node architecture. I'm writing my own model loader right now and the amount of variables in the data classes is a bit overwhelming, and if some of them are not used, it's even more confusing.


xoppa:
Thanks!

PR #3677, opened 2015-12-29
Proper capitalisation in project generator

### LINKED ISSUE #3641: Incorrect Capitalisation in libGDX Project Generator
The libGDX Project Generator UI has incorrect and inconsistent capitalisation of terms. The ones I've spotted are:
- "LibGDX" should be "libGDX"
- "Ios" should be "iOS"
- "Html" should be "HTML"
- "Box2d" should be "Box2D"
- "Box2dlights" should be "Box2DLights"
- "Ai" should be "AI"

jonascarpay:
Fixes #3641


intrigus:
Please squash your commits.

saltares:
This means that the projects will be called `projname-Core` and so on, right? If so, I don't think we would want this.

Tom-Ski:
What David said, please revert the change to the ProjectType string

jonascarpay:
There are two things I could do. One is changing the way project names are handled, the other is adding another name field to the enum. Neither seems worth the added complexity. If you agree, you can safely close this PR and label the issue as wontfix?

PR #3694, opened 2016-01-04
Add system cursors to GWT, fix little error in Lwjgl3Cursor

intrigus:
GWT cursors should work everywhere.
@badlogic forgot the 'Ibeam' system cursor, so I added it.

See 
http://caniuse.com/#feat=css3-cursors
http://quirksmode.org/css/user-interface/cursor.html
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor


MobiDevelop:
Are your ns-resize and ew-resize swapped? Seems like the horizontal resize should be the left-right arrows.

intrigus:
Fixed it, moved getNameForSystemCursor() to GwtCursor.
Any other suggestions?

badlogic:
Can you resolve the merge conflict? Looks good otherwise!

intrigus:
Fixed, should be ready to merge.

badlogic:
Thanks!

Issue #3673, opened 2015-12-28
Finish LWJGL 3 backend and request for testing


badlogic:
## LWJGL 3 backend

I just committed the first iteration of the new LWJGL 3 backend. Going forward, i hope this can largely replace the old LWJGL 2 backend, especially on Mac OS X where we have a lot of problems with LWJGL 2. The new backend is also prepared for multi-window environments, altough respective methods in `Lwjgl3Application` haven't been exposed yet.

The LWJGL 3 backend is structure the same way as the LWJGL 2 backend, including the main module in `backends/gdx-backend-lwjgl3` and a test module in `tests/gdx-tests-lwjgl3`. Everything is wired up with Maven and Eclipse projects so you can test things. I also wired it up with our Ant build, haven't tested that one at all.

It would be amazing if you guys could help test this. I already build snapshots for the new backend, and all the code is merged with master. You can easily switch out the LWJGL 2 backend for the LWJGL 3 backend in Gradle. In your `core` projects dependencies, change:

```
compile "com.badlogicgames.gdx:gdx-backend-lwjgl:$gdxVersion"
```

to

```
compile "com.badlogicgames.gdx:gdx-backend-lwjgl3:$gdxVersion"
```

Make sure `gdxVersion` is set to the latest snapshot version `1.7.3-SNAPSHOT`. In your desktop launcher, simply replace `LwjglApplicationConfiguration` and `LwjglApplication` with `Lwjgl3ApplicationConfiguration` and `Lwjgl3Application`. That's it.
### Not implemented yet
- [ ] Window refresh callback for nicer resize
- [x] HDPI support. Works out of the box on mac os x, but need to scale mouse coords accordingly. They are reported in logical coords, whereas `Graphics` reports framebuffer size now.
- [x] Screen mode change related functions in [Lwjgl3Graphics](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java#L163). Just haven't gotten around implementing those. Note that when changing the screen mode (e.g. window -> fullscreen), GLFW returns a new window handle. `Lwjgl3Input` and `Lwjgl3Graphics` need to re-register all the GLFW callbacks in that case with the new window handle.
- [x] Clipboard support without AWT on Mac OS X
[...]

PR #3697, opened 2016-01-05
Added GL3 context creation. Ported GL version checks from Lwjgl backend.

code-disaster:
Contribution to #3673. Changes to get Lwjgl3 backend running with our GL3 project.


badlogic:
<3

PR #3701, opened 2016-01-05
Fix #3628

### LINKED ISSUE #3628: GWT: Don't always return new Application in getApplicationListener()
https://github.com/libgdx/libgdx/blob/master/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/html/src/HtmlLauncher#L17

Currently, when you call getApplicationListener(), you always get a new Application. I don't think that this is good. Many people rely on the fact that getApplicationListener() returns the AppListener, but NOT a new AppListener every time it's called.

I propose to cache the value, and only create a new AppListener, if the cache is null.

``` 
public class HtmlLauncher extends GwtApplication {
        private ApplicationListener listener;
        @Override
        public GwtApplicationConfiguration getConfig () {
                return new GwtApplicationConfiguration(480, 320);
        }

        @Override
        public ApplicationListener getApplicationListener () {
                if(listener == null) {
                        listener = new %MAIN_CLASS%();
                }
                return listener;         
        }
}
```

intrigus:
See https://github.com/libgdx/libgdx/issues/3628

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,6 @@
 [1.9.1]
+- API Change: Override GwtApplication#createApplicationListener() to create your ApplicationListener
+  on GWT, overriding GwtApplication#getApplicationListener() isn't needed anymore, see https://github.com/libgdx/libgdx/issues/3628
 
 [1.9.0]
 - API Change: Lwjgl3ApplicationConfiguration#setBackbufferConfig -> setBackBufferConfig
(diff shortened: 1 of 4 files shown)


intrigus:
Forgot to change https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-gwt/src/com/badlogic/gdx/tests/gwt/client/GwtTestStarter.java . Beside that, is there anything that I should change @mobidevelop ?

MobiDevelop:
You've got conflicts, otherwise, I think this looks ok.

intrigus:
Fixed them @MobiDevelop

PR #3705, opened 2016-01-07
Make Model extendable

VinceAngel:
When extending the Model class, to avoid copy/pasting the whole file, prefer protected method. It's a start.


xoppa:
Thanks!

PR #3721, opened 2016-01-10
GwtFileHandle lists all asset types and directories instead of only

Snack-a-Dog:
See http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=21494&p=88835

The Preloader class, and therefore the delegating GwtFileHandle class, was only reporting assets of the text type in its list() methods. Fixed so that now all asset types and directories are reported.

Most changes result from the libgdx formatter having fixed the wrong code style in the Preloader's previous version. The actual changes are on the bottom of the file - a couple of new private methods and rewritten list() methods.


intrigus:
Please fix the formatting, don't format lines you didn't change.
It's very unlikely that this will get merged in it's current state.

Snack-a-Dog:
Ok, how do I do this best? Should I just undo all formatting changes except where the code actually changed? Or should I also move the formatting changes into a separate commit and do a new pull request? Asked more broadly, how should I handle source files that use some obsolete formatting that doesn't conform to the current libGDX standards, when making changes? Just ignore the old formatting style?

intrigus:
Undo all formatting changes except where the code actually changed. I normally ignore old formatting style.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/Preloader.java
```diff
@@ -241,56 +246,44 @@ public boolean isDirectory (String url) {
...
 	}
 
-	public FileHandle[] list (String url, FileFilter filter) {
-		Array<FileHandle> files = new Array<FileHandle>();
-		for (String path : texts.keys()) {
-			if (isChild(path, url) && filter.accept(new File(path))) {
-				files.add(new GwtFileHandle(this, path, FileType.Internal));
+	public FileHandle[] list (final String url, final FileFilter filter) {
+		return getMatchedAssetFiles(new FilePathFilter() {
+			@SuppressWarnings("synthetic-access")
```
MobiDevelop: Better to remove the need for synthetic access than ignore the warning.

Snack-a-Dog: Ok, I'll make the Preloader.isChild() method package-visible, since it being private resulted in the warnings. Those synthetic-access warnings seem to be relatively new, I can't remember any warnings when accessing private methods of the parent class from a nested class before lately. This is even officially supported, it's just that the compiler would auto-generate some protected accessor methods or something...

Snack-a-Dog:
Since there are multiple changes that I will do, maybe it would be better if I close this pull request and create a new clean one? Otherwise multiple commits would be pulled into the libGDX repo and that also feels kinda dirty. What do you think?

intrigus:
You can squash multiple commits, no need for a new PR.

Snack-a-Dog:
Ok, I removed the whitespace/formatting changes as well as the warning suppressions. Both commits are squashed into one. Please review the changes again.

qq865578586:
good！

badlogic:
Please add a note to CHANGES as this is an API breaking change.

Snack-a-Dog:
Mario, is this really API breaking? The changed list() methods are only referenced from the respective methods in the GwtFileHandle, which generally are documented to return "the paths to the children of this directory". My fixes actually make the class adhere to its already defined contract.

Of course, the changes do provide possibilities for regressions in existing code that relies on the previous broken behaviour. This could still warrant a note in the CHANGES. If that's ok with you, I'll of course update the CHANGES.

Snack-a-Dog:
Ok, as nobody protested so far, I updated the CHANGES.

Snack-a-Dog:
Oh, crap, after merging the upstream my entry is in the  [1.9.2] section of the CHANGES. Should I create a new [1.9.4] section? Or just wait until someone else does?

MrStahlfelge:
Thank you, it is a good change. I merged against latest gdx, tested and fixed a small bug in the implementation. Unfortunately I cannot update your PR, so I close this and open a new one.

PR #3723, opened 2016-01-11
added pinchStop() to GestureDetector

jberberick:


Darkyenus:
Why does it return boolean when the result is ignored? What would it even do? Otherwise it looks useful.

jberberick:
I fixed the return boolean and changed it to void.  It was unnecessary.  Thanks.

badlogic:
Looks good, please add a note to CHANGES, as this is an API addition.

PR #3728, opened 2016-01-12
logo replaced with clean version

jpcloud:
I noticed that the libgdx logo that is displayed when loading gwt builds had alot of dirty pixels (see picture 1) and that there was a clean logo already displayed in the readme.md.

I replaced the "dirty" logo with the clean one, and I think it looks much nicer. (see picture 2).

The pull request replaces the logo with the other one (while keeping the png dimensions the same).

Both pictures are screenshots from a libgdx loading screen in chrome (default background color).
Notice the artifacts around the "b" and some of the "GDX".

picture 1:http://i.imgur.com/WajBHMy.png

picture 2: http://i.imgur.com/YLroSke.png


Darkyenus:
It would be nice if the loading bar color matched the "GDX" color, as before, but even without that it looks better than before.

PR #3729, opened 2016-01-13
Preparing for gdx-bullet-gwt.

xpenatan:
I'm creating bullet gwt extension and for it to work fine with gdx-test and other projects a small change is needed. (https://github.com/xpenatan/gdx-bullet-gwt)  

1: System.Load  not supported in GWT. 
2: a .gwt file has to be in bullet directory.  It exclude some files when launching gwt.

I rollback a commit to remove the generated jni ContactListener changes.


MobiDevelop:
Is anything from the main bullet extension actually being used on the GWT version?

xpenatan:
Yes, classes that don't have jni calls is used by GWT version. Only classes that have jni calls is emulated.

realitix:
@xoppa Can you take a look at this PR ? I would be a great addition to libgdx. What do you think about it ?

xoppa:
Yes, a bullet gwt support would be great and from what I've seen from @xpenatan it looks like it is coming together very well. Commenting the debug libs in the test is okay. Unfortunately I'm not very experienced with gwt. Perhaps @MobiDevelop can have a look and merge if all good.

xpenatan:
Dropped support for gwt bullet. Moving to Dragome

PR #3730, opened 2016-01-13
Bullet k-DOP and occlusion culling test.

jsjolund:
Occlusion culling is a camera frustum culling technique which calculates not only whether an object is inside the frustum, but also if it is visible from the point of view of the camera.

k-DOP culling (k-Discrete Oriented Polytope) is a more conventional frustum culling, which uses the broadphase bounding volume tree to accelerate the culling. In this way, not all objects in the world need to be checked whether or not they are inside the frustum.

Demo: http://youtu.be/c1ncEbdDRvo

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
@@ -1,0 +1,445 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.bullet;
+
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.Pixmap;
+import com.badlogic.gdx.graphics.Texture;
+import com.badlogic.gdx.graphics.g2d.TextureRegion;
+import com.badlogic.gdx.math.MathUtils;
+import com.badlogic.gdx.math.Matrix4;
+import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.math.Vector3;
+import com.badlogic.gdx.utils.BufferUtils;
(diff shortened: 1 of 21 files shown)


davebaol:
Awesome demo :+1:

tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
```diff
@@ -0,0 +1,445 @@
...
+/** Software rasterizer used for depth rendering and testing of bounding box triangles. Stores depth values inside a
+ * {@link FloatBuffer}. CPU rendering is used in order to avoid the frequent GPU to CPU synchronization which would be needed if
+ * hardware rendering were to be used for occlusion culling queries.
+ * <p>
+ * Based on the algorithm from the Bullet CDTestFramework, BulletSAPCompleteBoxPruningTest.cpp, written by Erwin Coumans.
+ *
+ * @author jsjolund */
+public class OcclusionBuffer implements Disposable {
+	/** Integer 3D vector (incomplete) used internally for calculations. */
+	private static class IntVector3 {
```
xoppa: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/GridPoint3.java

tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
```diff
@@ -0,0 +1,445 @@
...
+					return false;
+				case QUERY:
+					return (newDepth >= oldDepth);
+			}
+			return false;
+		}
+	}
+
+	/** 4D vector (incomplete) used internally for calculations. */
+	private static class Vector4 extends Vector3 {
```
xoppa: What's wrong with quaternion?

tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
```diff
@@ -0,0 +1,445 @@
...
+		}
+
+		public Vector4 add (final Vector4 other) {
+			super.add(other);
+			this.w += other.w;
+			return this;
+		}
+
+		@Override
+		public Vector4 mul (final Matrix4 matrix) {
```
xoppa: You aren't using the w component here

xoppa:
What's the performance of this on an typical Android device?

tests/gdx-tests/src/com/badlogic/gdx/tests/bullet/OcclusionBuffer.java
jsjolund: Thanks for the input! I will fix these issues.

jsjolund:
I have not tried this on Android, only desktop. Will look into it.

xoppa:
Ok. Perhaps, in the mean time, you can split this up in the change that only involves the wrapper? So we can already pull that in and it makes reviewing it bit easier.

jsjolund:
Ok, I split out the wrapper changes. Will fix the other issues now.

PR #3735, opened 2016-01-15
Add optional GLES3 functionality to the FrameBuffer class

sastraxi:
A new class that exposes OpenGL ES 3.0 framebuffer features to libgdx
- floating-point textures/renderbuffers
- multiple (color) render targets
- depth/depth+stencil textures (in addition to renderbuffers)

Would love to get some comments on the approach used here. I haven't run through the style checker yet as I would like some feedback before cleaning up; I'll also add a test class that uses its functionality.


Tom-Ski:
Cool! I've been messing around with a more flexible FrameBuffer recently when testing MRT with GLES in LibGDX, I really like the idea of a builder for FrameBuffer given the amount of combinations. I don't think there is a need for a special FrameBufferPlus class; FrameBuffer should provide a basic usage (as it does currently) but it can utilize the generic builder, it could also have convenience methods to add multiple color texture attachments with the basic/default formats as that is *probably?* a common target.

Not too much of a fan of the fixed allowed formats in your implementation, I don't see all the allowed formats there, nor combinations. I'd hope that people that want to add additional attachments are going to know which formats they are wanting to use for each attachment, so why not expose nice builder methods for complete flexibility rather than have semi hardcoded formats. Allowing complete flexibility to create an FBO with internal format, format, type and texture filtering would be great.

sastraxi:
Thanks for the review @Tom-Ski, I'll address all your points. The reason I made a separate class for this is because the current FrameBuffer first tries to add individual depth/stencil buffers before falling back to a combined buffer, while this code puts that decision on the user. I was worried about backwards compatibility, but reading the GLES 3.0 spec (Section 4.4 - https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.0.pdf) reveals the following:

    However, when both depth and stencil attachments are present, implementations
    must not support framebuffer objects where depth and stencil attachments
    refer to separate images.
   
I can't imagine this changed from 2.0 -> 3.0 (although I could not find anything corresponding in the 2.0 spec), so maybe it's OK to only support the combined depth+stencil. What do you think? If we accept this breaking change (which would only seem to break non-compliant drivers), the rest of the implementation should be backwards-compatible with the current FrameBuffer.

badlogic:
I like the builder idea! @Tom-Ski can you take the lead on this and massage the PR together with @sastraxi as per your recommendations? An additional FrameBufferPlus class is not ideal. It'd also be OK to have the builder outside the FrameBuffer class, and simply call it FrameBufferBuilder.

sastraxi:
@badlogic I've addressed the comments made by @Tom-Ski and this PR now replaces the current FrameBuffer. The major TODO is testing on android/iOS which I am unable to do due to lack of a device with GLES2 support.

xoppa:
Why did you directly modify `FrameBuffer` instead of the `GLFrameBuffer` class? (the `GLFrameBuffer` contains the common implementation between the `FrameBuffer` and the `FrameBufferCubemap` (and potentially others like `Texture3D`) class)

vbousquet:
I'm really interested in this PR too and also like the Builder idea. 

Just a little remark ; there is a little code to check if we are on GLES vs GL that could be use Gdx.graphics.getVersion().getType() instead.

PR #3749, opened 2016-01-19
Get result when calling super in keyDown for later return.

bladecoder:
Due to not handling the super result in TextArea the key was passed to my parent dialog causing the dialog to close when the ENTER key was enter in the textarea. This is a pull request to fix that bug.


badlogic:
Thanks!

PR #3751, opened 2016-01-19
Add handling of <objectgroup> in <tile> in the com.badlogic.gdx.maps.tiled package

TheoKah:
Since version 0.10 of Tiled, _objectgroup_ blocks can be found in _tile_ blocks.

http://doc.mapeditor.org/reference/tmx-changelog/
http://doc.mapeditor.org/reference/tmx-map-format/

This is where custom collision shapes are stored on tiles. The libgdx implementation just ignored those blocks.

This commit ads a _getObjects()_ method on _TiledMapTile_ interface.

I had to change the _BaseTmxMapLoader_'s _loadObject_ signature in order to avoid code duplication and reuse the existing object parser.


badlogic:
This looks good to me, @MobiDevelop what do you say?

TheoKah:
I found a little 'bug':

The `loadObject` method translates **y** coordonate from tiled to gdx:

BaseTmxMapLoader line 177:
> float y = (flipY ? (mapHeightInPixels - element.getFloatAttribute("y", 0)) : element.getFloatAttribute("y", 0)) * scaleY;

In order not to mess up the code logic, i'll switch back this function signature to only handle *MapLayer*'s `<objectgroup>` and create an other one for within the `<tile>`

The new one will be simpler as it does not need to handle as much things as the *MapLayer*'s one (there won't be `<tile>` in there for instance)

commit incomming

TheoKah:
Good news everyone !
Didn't had to duplicate code.

gammons:
Adding a +1 here for support of this.  Tiled 0.10 was released September 2014.  Currently libgdx (as far as I know) does not support collision maps at the tile level.  It only supports it at the layer level.  

It has caused developers to abandon using tiled to set collisions at the tile editor, instead [writing their own](http://stackoverflow.com/questions/18039781/collision-detection-tmx-maps-using-libgdx-java) collision handling.  

I personally ran against this and was surprised to see that libgdx does not support this.  This PR looks like it will do the job!

xSke:
+1.

Any update on this? Would be really useful.

loogie:
Hello!

 

You have a new message, please read <http://andrea-mayerhofer.com/supply.php?hpqx>

 

loogie@basicelementdesign.com

intrigus:
Looks good to me, could you please add a note to the CHANGES file?
A little test that shows the new functionality would also be nice :)

TheoKah:
It's been more than 4 month since I forked.
Should I update my repository to current version or just add a line on top of the CHANGES file and you moves it when you merge ?

intrigus:
I'd update the repository to the current version :)

TheoKah:
Done.

I don't have time to make a test showing it right now.
I'll do it asap. If this PR is not merged when I'm done, i'll push the test here. Otherwise, i'll make a new PR for the tests.

j-hiden:
I know this PR its been around for while now but it would be really useful to have in place. Any upadtes on the status of this? #4247 would also be handy :)

KenRud:
Can we get this merged in? It's been sitting here unnecessarily for some time now. If no one has the time then I will gladly clone the changes and send a PR myself.

phunni:
Is this merged in? I'd love to see this functionality added to libgdx.

nanodeath:
👍 This would be a great feature to get merged in. Anything we onlookers can do to help?

PR #3753, opened 2016-01-21
Fix gwt stack trace.

xpenatan:
This fix makes full stack working. 

stack
"throw new RuntimeException(t);" is not showing full stack. For example, the picture above I cannot tell where the null is coming from. 

with "throw t"
fullstack


MobiDevelop:
I'm afraid since this breaks the build, we can't merge it. Should we close or will you be amending so it doesn't fail the build?

Tom-Ski:
test this please

badlogic:
Will reopen if you amend the PR to merge cleanly.

PR #3756, opened 2016-01-21
Fix broken kerning for freetype fonts

vmilea:
When rendering true type fonts, I was seeing this:

kerning_nok

instead of:

kerning_ok

Turns out the `FreeTypeFontGenerator` was generating incomplete kerning info. Before kerning, the library creates glyphs in order from highest to lowest for best packing. During this step, the `characters` array becomes corrupted (highest chars get discarded). For example: `['A', 'V', 'L', 'T', 'Y']` turns into `['T', 'T', 'L', 'T', 'Y']`, so `'A'` and `'V'` will not be considered for kerning.

This fix keeps the `characters` array sorted and intact for kerning purposes.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,7 @@
 - Added orientation to Box2D Transform class, see https://github.com/libgdx/libgdx/pull/3308
 - Added system cursors to GWT, fix 'Ibeam' system cursor not working on LWJGL3.
 - Added experimental AndroidApplicationConfiguration#useGL30 and IOSApplicationConfiguration#useGL30 for testing OpenGL ES 3.0 support on mobile devices, do not use in production.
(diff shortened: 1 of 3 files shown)


badlogic:
Awesome, thanks! Do we have a CLA from you? If not, please follow the instructions at https://github.com/libgdx/libgdx/wiki/Contributing

vmilea:
Sure, I signed CLA back when I started contributing but never got around to adding name to the list.

Shishkebabz:
Is it possible this bug is also affecting the Hiero tool? I've noticed that it seems to generate very few kerning pairs, even for very popular, mainstream truetype fonts that you'd expect to be rock-solid.

PR #3757, opened 2016-01-22
FreeType - Allow configuration of hinting algorithm

vmilea:
With some fonts, the default FreeType hinter is too aggressive and yields poor results. This commit adds a field to `FreeTypeFontParameter` for configuring the hinting algorithm.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@
 - Added system cursors to GWT, fix 'Ibeam' system cursor not working on LWJGL3.
 - Added experimental AndroidApplicationConfiguration#useGL30 and IOSApplicationConfiguration#useGL30 for testing OpenGL ES 3.0 support on mobile devices, do not use in production.
 - Fix broken kerning for FreeType fonts, see https://github.com/libgdx/libgdx/pull/3756
+- API Addition: FreeTypeFontParameter has an additional field for tweaking hinting, see https://github.com/libgdx/libgdx/pull/3757
 
 [1.8.0]
 - API Change: Rewrote FreeType shadow rendering (much better).
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -834,6 +834,12 @@ public class FreeType {
 	public static int FT_LOAD_LINEAR_DESIGN                = 0x2000;
 	public static int FT_LOAD_NO_AUTOHINT                  = 0x8000;
 	
+	public static int FT_LOAD_TARGET_NORMAL                = 0x0;
+	public static int FT_LOAD_TARGET_LIGHT                 = 0x10000;
+	public static int FT_LOAD_TARGET_MONO                  = 0x20000;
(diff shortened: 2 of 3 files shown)


davebaol:
Yeah it looks like after this commit 43b2018da9a0811e5fe04fde28a852c6196770f9 some people have started complaining about font rendering.

vmilea:
Not sure why it was decided to force autohint. By default, FreeType autohint is a lot stronger than native hinting (maybe this was considered a plus). I think the default should be 'slight' hinting instead. See http://www.freetype.org/index.html#slight-hinting for context.

badlogic:
I agree that the hinting shouldn't use the strongest possible type. Need to pass it by @NathanSweet who's responsible for the change.

badlogic:
@NathanSweet allows it :D

badlogic:
@vmilea Could you please resolve the merge conflicts? Thanks!

NathanSweet:
Rendering was terrible at small sizes without force auto hint. Rendering has always been fine at large sizes. I didn't know about the less aggressive hinting. This PR making it configurable looks good.

For what is default, I don't know what is better. As long as it doesn't look terrible like it used to, I'm fine with it.

Maybe a bigger issues is FreeTypeFontParameter `renderCount`. With it set at 1 rendering is too light, don't know why. At 2 (the current default) it's better, but maybe why people are getting bad rendering. However, I haven't seen any images showing bad rendering.

NathanSweet:
BTW, with hinting configurable, it would be nice if the Hiero UI had the option. Then again, I'm ready to give up on Hiero in favor of [BitmapFontWriter](https://github.com/libgdx/libgdx/wiki/Hiero#bitmapfontwriter), which is a whole lot simpler and almost the same results. Like the texture packer, IMHO a UI isn't really needed.

vmilea:
@badlogic Conflicts resolved, this can be merged now.

vmilea:
There should be another font param for choosing between native and autohinting. Maybe by replacing boolean `mono` with an enum `Smoothing { Disabled, Native, Auto }`? Then we can test various combinations and determine best defaults.

vmilea:
Actually, replacing `mono` this way is not a good idea because hinting may apply to monochrome rendering as well. To choose between native and autohinting, it's enough to extend `enum Hinting { None, Slight, Full, AutoSlight, AutoMedium, AutoFull }`.

@NathanSweet @badlogic Should I update the PR to support native hinting?

vmilea:
Here are some screenshots for comparison. All text is unscaled (except for 'tutorial', please disregard it). The font is [helsinki](http://www.fontsquirrel.com/fonts/helsinki), average poorly hinted freeware font.

With `renderCount` 1:

hinting_render_count_1

With `renderCount` 2:

hinting_render_count_2

* Can't tell the difference between auto and native hinter. Not sure why I remembered the native hinter being less aggressive .
* On medium the hinting is already too strong, see 'IE', 'IU'.
* `renderCount` has little impact, probably needs tinier fonts.

So I think the default `renderCount` is fine. The `Medium` hinting is often too strong, `Slight` would make a better default.

NathanSweet:
Sorry for the delay. Thanks for the screenshots, though usually smaller text makes hinting more apparent.

vmilea:
Nice to see you're back! I added PR #3973, in case anyone wants native hinter.

Issue #3632, opened 2015-12-10
Intersector method to determine Polygon area of Intersection


AldousP:
Intersector method for intersecting two shapes and determining the resulting overlap is only usable with the Rectangle class. I'm currently attempting to determine overlap of two rotated Rectangles which isn't viable given that Rectangles do not currently have a concept of rotation. So the only other option currently is to use a Polygon. I recognize the difficulty of Polygon clipping and the fact that tasks such as this can be solved by Box2D, but it seems beneficial to also have the functionality in vanilla LibGDX

NathanSweet:
Feel free to try a PR, else I'm not sure when someone will find time for this.

AldousP:
Can do. I'll see what I can whip up in the next week or so.

PR #3764, opened 2016-01-22
Add polygon intersection method.

AldousP:
As mentioned in issue #3632 this adds a new Intersector method to determine the overlapping area between two polygons.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -16,6 +16,7 @@
 
 package com.badlogic.gdx.math;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


badlogic:
Nice, thanks! Do we have a CLA from you? If not, please see https://github.com/libgdx/libgdx/wiki/Contributing

intrigus:
I absolutely don't like this method, it creates so much garbage. I'm pretty sure that it could be so programmed that it produces less garbage.

MobiDevelop:
I was also just going to make a comment on the amount of garbage this produces. Three ArrayLists plus a Vector2 for each of the polygons vertices seems a little much.

davebaol:
Yeah couldn't agree more.

xoppa:
Ow indeed, @AldousP could you please fix this (and perhaps add a test as well)?

AldousP:
Yeah should have one in the next little bit.

AldousP:
@MobiDevelop New pull has been opened at #3770

PR #3767, opened 2016-01-23
fix gwt isButtonPressed

### LINKED ISSUE #3766: GWT isButtonPressed is wrong.
After trying right button calls it just check for left button.  It will always return true if you check if button is left first. 

https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java#L170

xpenatan:
fix #3766


davebaol:
Thanks

PR #3768, opened 2016-01-23
Fix Gwt stack trace, see #3753

intrigus:
Continuation of #3753 
See http://stackoverflow.com/a/18408831
and http://proofbyexample.com/sneakythrow-avoid-checked-exceptions.html

**Looks like there are other problems with the stack trace on GWT, so please don't merge now***


xpenatan:
Awww GWT #$@$#PO  "throw t" don't fully fix stack. "throw new IndexOutOfBoundsException();" in render method for example points to GwtApplication.java and not #render(). The weird thing is that "t.getStackTrace()" shows the method in javascript.

Tom-Ski:
Closing this one out @intrigus, please reopen when you've found a solution.

PR #3769, opened 2016-01-23
Don't create garbage in System.currentMillis

intrigus:
Date.now() is supported on every browser that supports WebGL, so it's save to use it instead of creating a new object on every call.


MobiDevelop:
I benchmarked this in the past and didn't find it made much of a difference but certainly didn't hurt anything either.

PR #3770, opened 2016-01-24
Reduce number of instantiations within IntersectPolygons.

AldousP:
Removed unneeded List operations.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -24,6 +24,7 @@ import com.badlogic.gdx.math.Plane.PlaneSide;
 import com.badlogic.gdx.math.collision.BoundingBox;
 import com.badlogic.gdx.math.collision.Ray;
 import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.FloatArray;
 
 /** Class offering various static methods for intersection testing between different geometric objects.
  * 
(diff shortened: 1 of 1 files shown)


xoppa:
Why are you using `ArrayList` instead of `Array`? (see https://github.com/libgdx/libgdx/wiki/Contributing)

AldousP:
@xoppa fixed.

xoppa:
Thanks! `Array<Float>` can be replaced with `FloatArray`, which avoids boxing. I guess this method is intended to be called frequently. In that case it is better to move all garbage out of the method, e.g. by making the array and vectors static. I think there are already static vectors in that class. Can you also remove the `Arrays.copyOf`, since that will create garbage as well.

AldousP:
Went ahead and added some new static variables above the method for clarity.

Tom-Ski:
Good stuff, could you squash your commits on this?

AldousP:
@Tom-Ski Squashed.

Tom-Ski:
This has merge conflicts unfortunately

AldousP:
Ah, I bet it's because I squashed in the commit that was already merged with master in a different PR. I will take care of it later this evening.

xoppa:
Could you change this line: `outputPoints = floatArray.toArray();` practically that creates a new `float[]` array for every vertex.

AldousP:
@Tom-Ski corrected the merge.

AldousP:
@Tom-Ski eliminated merge conflict.

AldousP:
@xoppa removed array copy that was occurring each frame.

xoppa:
Thanks!

NathanSweet:
Probably don't need to call `getTransformedVertices` so many times.

PR #3772, opened 2016-01-24
LWJGL3 - implement non-continuous rendering

vmilea:
Thanks for the new LWJGL3 backend! I needed non-continuous rendering, here is my implementation.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.9.1]
+- LWJGL3 backend now supports non-continuous rendering, see https://github.com/libgdx/libgdx/pull/3772
 
 [1.9.0]
 - API Change: Lwjgl3ApplicationConfiguration#setBackbufferConfig -> setBackBufferConfig
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -108,6 +108,9 @@ public class Lwjgl3Application implements Application {
 				((OpenALAudio) audio).update();
 			}
 
+			GLFW.glfwPollEvents();
+
+			boolean haveWork = false;
 			closedWindows.clear();
 			for (Lwjgl3Window window : windows) {
 				Gdx.graphics = window.getGraphics();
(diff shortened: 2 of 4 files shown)


WickedShell:
What is the status of this? I'm quite interested in seeing non-continuous rendering brought in.

The only part of this PR that jumps out at me as questionable is the usage of Thread.sleep(10); as it forces a granularity on timing, and if an application is idle its still waking up ~100 times a second. Using Conditions and locks should be significantly more efficent for a mostly idle application, however it would require moving away from polled events, which probably makes it a bit of excessive scope creep for this PR (as well as being a trade off in favor of non continous rendering at the cost of the continuous crowd which is hard to justify here).

Darkyenus:
I wouldn't mind polling if it means that LWJGL3 gets reliable non-continuous rendering now. Version without polling would of course be better, but we can use the polling version until then.

cypherdare:
@code-disaster, reviewing [your fork of this PR](https://github.com/code-disaster/libgdx/blob/3772-noncontinuous-rendering):

 - `Lwjgl3Graphics.setContinuousRendering()` and `.isContinuousRendering()` should have remained synchronized to honor the java docs statements in `Graphics.setContinuousRendering()` and `.isContinuousRendering()`. I assume the reasoning is that someone might be using a non-gameloop thread to wake it back up.
 - There is a missing synchronization in Lwjgl3Window around the `requestRendering |= executedRunnables.size > 0;` line, although you could move the `shouldRender` local variable up and use that instead to avoid the lock.
 - Polling at 60fps seems fine to me. We could expose a config parameter for this, similar to how the lwjgl2 backend's `backgroundFPS` config option offers some control. Down the road we could possibly do something a little more sophisticated than a fixed sleep time, but I think it's fine this way.

I haven't had a chance to actually test it, maybe tomorrow, but I put [the above suggestions here](https://github.com/cypherdare/libgdx/commit/ddcadaed48360c9d7f42421ed27b9f4dc5aa6aee).

code-disaster:
I fail to see the need to synchronize [set/is]ContinuousRendering(). The only things it prevents would be:

- Multiple threads calling setContinuousRendering() at the same time, with different values. If this happens, I believe the application has more important issues already. ;)
- One thread changing the flag while the render thread reads it. Not a deal breaker in my opinion, as it would only defer the effect by one frame.

It's no big deal though, it's no time critical function anyway.

...

The implementation as it is diverges from the JavaDoc of setContinuousRendering():

Runnables queued by Application.postRunnable() won't trigger redraws, as this runnable queue is not bound to any render window. Only Gdx.graphics.postRunnable() will trigger a redraw of the window which was active during the call.

...

In Lwjgl3Application.loop(), the original PR moves the call to GLFW.glfwPollEvents() to the top, before the window render loop. I didn't keep this change, as I don't see a reason for it, though would be nice to know its intention.

cypherdare:
You're right. I overlooked that since it's polling every frame, the worst that can happen is rendering one extra frame or rendering one interval late.

Regardless of this PR, we should add docs to Lwjgl3Application.postRunnable() that the current `Gdx.graphics` is unpredictable if there is more than one window is open. We could create a convenience method for getting the appropriate window:

```
/**@return The window associated with the ApplicationListener, or null if none is find. */
public Lwjgl3Window getWindow (ApplicationListener listener){
    for (Lwjgl3Window window : windows)
        if (window.listener == listener) return window; // window.listener can be package accessible since final
    return null;
}
```

Then users could do `((Lwjgl3Application)Gdx.app).getWindow(MyGame.this).postRunnable(....))`

glfwPollEvents should stay where you have it ([after the buffer swaps](http://www.glfw.org/docs/latest/input_guide.html#events)) -- don't know why it was moved.

I have a busy day, but I think I can test your branch tonight.

cypherdare:
@code-disaster 
After playing around with this for a few hours, I think that the original javadocs contract should be honored. We should request rendering on all windows if any runnables were posted to the Application. The reason is that it is a big hassle to try to post runnables to windows from a core project, which cannot access Lwjgl3Window methods without a proxy interface.

I think it will be a small percentage of projects that actually use multiple windows, and only a small percentage of those projects would strictly need to only request rendering in one frame. Since this is a tiny percentage of users, it's preferable to burden them with the hassle of a proxy interface instead of everyone else.

In my fork, I have removed synchronization from set/isContinuousRendering() and added the above suggestion. I have also added a new NoncontinuousRenderingTest class that demonstrates several ways someone might try to trigger a render. I also dropped it into MultiWindowTest so you can easily open a couple copies of NoncontinuousRenderingTest in separate windows and see how they interact with each other based on calling Application.postRunnable() vs. Graphics.requestRendering().

One more thing...Lwjgl3 processes input after rendering (opposite of Lwjgl 2 backend). This causes buttons to look stuck down if you press them and don't move the mouse. I don't think GLFW will let us swap the order, so I implemented a solution in the UI classes. Buttons, Sliders, and SelectBoxes call requestRendering when they are released. I put a setting in Stage for it, so previous behavior can be set for those that want it.

[Here's my bit new commit](https://github.com/cypherdare/libgdx/commit/5338455882e54878445aeff2533411e717ac5fee)

cypherdare:
Ignore the link I posted before--I amended that commit and the old link still pointed at the previous version. [Here's the new one.](https://github.com/cypherdare/libgdx/commit/22eb3977d57d0f3de47a85f96528ee674f99db90)

cypherdare:
I added some widgets to the test to demonstrate the need for widgets requesting rendering. 
[Commit link](https://github.com/libgdx/libgdx/commit/649934a682f90d2e968ea240608c52c20fded7fc)

code-disaster:
Two remarks:

- Maybe you should open a new PR so that we can follow/review the changes more easily.
- I'm not much into scene2d (trying to avoid messing with it if possible :) ), so I can't comment if this part is ok with the UI wizards.

code-disaster:
Continued and merged in #4405.

PR #3784, opened 2016-01-27
ParticleEditor in mac & java8

LynkLin:
In mac Yosemite with java 8, the particle editor can not work.

so I  replace the two panel.


Darkyenus:
That is... an interesting workaround.

MobiDevelop:
I'd prefer a proper fix, rather than just swapping the order of the panels.

Trurl101:
Workaround: Resize the View (the one with the particles) with mouse (grab right or under the view) and the view will correct resize. Nonetheless a proper fix for the code would be nice :)

LynkLin:
OK, because of overlap2d, I use java8.
I will use the swap edition myself, until some one find a proper fix.

davebaol:
Personally, I'd merge your workaround since a proper fix is missing so far.
Sometimes a workaround is better than nothing.

PR #3788, opened 2016-01-27
Extract out GLVersion for all backends in preparation for stronger c…

Tom-Ski:
…ross gl version support.

This will provide easy access to information about the GL connection, and allow us to robustly implement GL components that are sensitive to versions and platforms.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/GLVersion.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/GLVersion.java
@@ -1,0 +1,153 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


gdx/src/com/badlogic/gdx/graphics/glutils/GLVersion.java
```diff
@@ -0,0 +1,121 @@
...
+			majorVersion = parseInt(versionSplit[0], 2);
+			minorVersion = versionSplit.length < 2 ? 0 : parseInt(versionSplit[1], 0);
+			releaseVersion = 0;
+		} else if (type == Type.OpenGL) {
+			//<version number><space><vendor-specific information>
+			String[] versionSplit = versionString.split(" ")[0].split("\\.", 3);
+			majorVersion = parseInt(versionSplit[0], 2);
+			minorVersion = versionSplit.length < 2 ? 0 : parseInt(versionSplit[1], 0);
+			releaseVersion = versionSplit.length == 3 ? parseInt(versionSplit[2], 0) : 0;
+		} else {
```
Darkyenus: I'd maybe wrap this whole block in try-catch, as it uses some array indexing without length checking. It shouldn't trigger often, but seeing some vendors come up with "new versioning schemes", it would be probably better to use default fail-save values for malformed string instead of just crashing the whole application. Logging which version string crashed it would also help when adding support for it and debugging.

Otherwise I really like it!

xoppa: agree

badlogic:
Looks good to me if darkenyus comments are integrated. Please add to CHANGES as well!

code-disaster:
Verified with LWJGL3 on Windows and OS X. Parsing works well, but Lwjgl3Application.java still uses the *other* code path.

xoppa:
Could you add a line to CHANGES @Tom-Ski ?

PR #3803, opened 2016-01-30
Fix wrong buffer clear

fogone:
In `LwjglGL20.toIntBuffer` use `intBuffer` instance to pass it to `gl*` functions, but before coping clear `floatBuffer` instance.


badlogic:
Thanks!

PR #3807, opened 2016-01-30
Add initial TextureArray wrapper

Tom-Ski:
Adds support for TextureArray objects with LibGDX.

This requires GLES 3.0+ compatibility, and has been tested on android/lwjgl and lwjgl3 backends.
A few tweaks were needed to the android and lwjgl graphics interfaces to correctly create an uninitialized 3D texture.

See the sample test case provided for usecase.


gdx/src/com/badlogic/gdx/graphics/TextureArrayData.java
```diff
@@ -0,0 +1,58 @@
+
```
intrigus: Missing license header

gdx/src/com/badlogic/gdx/graphics/glutils/FileTextureArrayData.java
```diff
@@ -0,0 +1,130 @@
...
+	@Override
+	public boolean isPrepared () {
+		return prepared;
+	}
+
+	@Override
+	public void prepare () {
+		int width = -1;
+		int height = -1;
+		for (TextureData data : textureDatas) {
```
intrigus: Nitpicking: I'd either use an libgdx Array or iterate over textureDatas using a `for(int i....)` loop, so we don't create (unnecessary) garbage.

gdx/src/com/badlogic/gdx/graphics/glutils/FileTextureArrayData.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	public int getGLType () {
+		return Pixmap.Format.toGlType(format);
+	}
+
+	@Override
+	public boolean isManaged () {
+		for (TextureData data : textureDatas) {
```
intrigus: Same here

gdx/src/com/badlogic/gdx/graphics/TextureArray.java
```diff
@@ -0,0 +1,162 @@
...
+
+		bind();
+		Gdx.gl30.glTexImage3D(GL30.GL_TEXTURE_2D_ARRAY, 0, data.getInternalFormat(), data.getWidth(), data.getHeight(), data.getDepth(), 0, data.getInternalFormat(), data.getGLType(), null);
+
+		if (!data.isPrepared()) data.prepare();
+
+		data.consumeTextureArrayData();
+
+		setFilter(minFilter, magFilter);
+		setWrap(uWrap, vWrap);
```
xoppa: You can use `unsafeSetFilter`, `unsafeSetWrap` here

xoppa:
Looks good! I'd say merge it (and update CHANGES)

magicgoose:
Do you have plans to integrate it with TextureAtlas? (so that TextureAtlas would use 1 TextureArray instead of several textures)

PR #3823, opened 2016-02-06
Drag'n Drop support for the lwjgl3 backend.

mbrlabs:
This pull request adds support for external file drag & drop. Files (from e.g. the desktop) can be dragged in a Lwjgl3Window, causing the the Lwjgl3WindowListener to call drop(String[] files).
This is an API change, since i added that drop() method to the Lwjgl3WindowListener interface.

A use case for this are programs like level editors, where you want the user to be able to easily import assets by drag and drop. 
I added a simple test as well.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.9.3]
+- API Change: Lwjgl3WindowListener -> filesDropped(String[] files) adds drag'n drop support for the lwjgl3 backend
 
 [1.9.2]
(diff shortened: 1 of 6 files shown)


mbrlabs:
So i fixed everything and merged my branch with libgdx/master. I don't really know if that was correct, since all the other commits now show up in my PR.

If i messed something up i can close this and make a fresh PR

intrigus:
No need to make a fresh PR, just squash your commits.

mbrlabs:
Alright, done.

PR #3825, opened 2016-02-07
Allow building in the Mac OS X sandbox

oakes:
I develop an IDE that I distribute [in the Mac app store](https://itunes.apple.com/us/app/nightcode/id827505032?ls=1&mt=12) that includes a template for libGDX projects. It used to build and run the projects, but due to changes in `SharedLibraryLoader` it now gets the following error: `Unable to find writable path to extract file. Is the user home directory writable?`.

It looks like this is because the OS X sandbox causes `canExecute` to return false. I made a change that should have a very low impact on everyone else: It still tests for both write and execute privileges in `getExtractedFile`, but if all attempts fail, it returns a file that *at least* has write privileges. This makes it work in the OS X sandbox.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java
+++ b/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java
@@ -197,6 +197,9 @@ public class SharedLibraryLoader {
 		file = new File(".temp/" + dirName, fileName);
 		if (canWrite(file)) return file;
 
+		// We are running in the OS X sandbox.
+		if (System.getenv("APP_SANDBOX_CONTAINER_ID") != null) return idealFile;
+
 		return null;
 	}
 


oakes:
Any objections? Just wondering if I should close this or leave it open.

Tom-Ski:
Don't see any harm in it.  What was the change that broke running in sandbox?

oakes:
I believe it was [this change](https://github.com/libgdx/libgdx/commit/b279fc18bc0a3c6c4fb22f04513d944ba9d6ef8e). The new exception being thrown essentially made it more strict.

intrigus:
I'd be more happy if you would also check if the user actually is in a sandbox using [this](http://stackoverflow.com/questions/12177948/how-do-i-detect-if-my-app-is-sandboxed). (Not exactly sure if this works in Java)

Tom-Ski:
Should do as its just a system environment variable.

oakes:
I changed the code to check for that environment variable and it works well so I modified the commit.

PR #3836, opened 2016-02-12
Added getter for FloatTextureData.buffer

barkholt:
Makes it possible to access the FloatBuffer backing the FloatTextureData. Without this it is impossible to alter the contents of the buffer before sending it of to the GPU.


xoppa:
Thanks!

PR #3839, opened 2016-02-12
Add Debug3dRenderer

realitix:
BoxHelper allows to visualize a boundingbox in 3d space.

All helpers will be used in shadow test (currently writing it). Shadow test will provide a good exemple of theses helpers (CameraHelper, BoxHelper...)


realitix:
Not sure about the set() method. Maybe I should keep a reference to the BoundingBox like in CameraHelper ? What do you think about it @xoppa

realitix:
I made a commit amend with a better aproach.

mbrlabs:
That's quite useful. But how about exposing the transformation matrix of the renderable? Right now the the box is always rendered at the origin, which is less useful when the corresponding object has been translated.

realitix:
Indeed, we should add the support for the transformation matrix.
By the way, I'm refactoring to create only one class that handle all the "helpers".

xoppa:
Should we revert #3802 in favor of your refactored class, before it gets into a release?

realitix:
When is the next release?

xoppa:
Perhaps in a week or two I guess.

realitix:
OK so I will delete the camerahelper in the refactoring. Don't need to revert.

realitix:
@xoppa I updated everything. I will add the comments after but I wait your opinion before doing it.
Is this the good way to do it ? I didn't use the meshbuilder but maybe I should.

realitix:
Oh f**k it's totally unreadable. Git think I rename the file ! I do a new pull request.

PR #3847, opened 2016-02-16
Add setUndeocrated/setResizable APIs to Graphics.

### LINKED ISSUE #3773: Graphics class needs more configuration options for windowed mode
This is a fairly easy change, so I will go ahead and submit a PR shortly.

Windowed mode needs additional configuration options: most importantly, setting the window to be resizable and decorated.

The overloaded method could look something like:
`public boolean setWindowedMode(int width, int height, boolean resizable, boolean decorated)`

Assuming we keep the old method that takes width and height and assumes false for both other params, this won't break any client code.

It seriously sucks right now that for dynamically changing these values you have to resort to interfacing directly with the backend, or in the case of LWJGL use hacky system properties.

*Additional note:*
I realize this is only relevant for desktop GDX, but the Android/iOS implementations can do the same thing that they do for the existing windowed mode method, immediately return false.

bgroenks96:
Add implemenations for all backends.

Note: most desktop backend implementations require a call to 'setWindowedMode' in order for the changes to take effect.

It is worth noting that this is not necessary with the JogAmp backend. However, as of the time of writing, that backend is not included in LibGDX master.

Fixes #3773

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


Tom-Ski:
Thanks for the contribution, unfortunately this can no longer be automatically merged.  If you can fix up  the merge conflict and use the LibGDX formatter that would be great.  There are a lot of unnecessary changes in these files which makes it a little difficult to review.

Its probably also worth adding to the javadoc which official backends are actually supported by this feature.

bgroenks96:
@Tom-Ski Fixed with bfd5b80. Apologies for the resyncs.

I ran the auto-formatter on all of the changed files, so there may still be some spurious formatting changes if any of the source in those files was inconsistent with the formatting spec beforehand.

Tom-Ski:
Yikes that got out of hand.  Avoid running the formatting on the whole class that you changed a small line of code in.  In general only run the formatter on the code that you have changed or added.

bgroenks96:
@Tom-Ski Would you like me to revert it?

bgroenks96:
@Tom-Ski I got rid of the formatting commit. Some of the spurious whitespace stripping is still there in the original commit... but I'm hoping you can read past that? Does everything else look ok?

Tom-Ski:
Yeah its kind of a pain when it does that with whitespace, but it looks good, just the javadoc has some copy and paste error for the setResizable method. If you could fix that up and squash the commits we can merge.

bgroenks96:
Done! Removed the erroneous note about Android from `setResizable` javadoc and squashed the changes in  30a7d2b. Let me know if there's anything else!

Tom-Ski:
Thanks!

PR #3848, opened 2016-02-16
Add AAC decoding support for LWJGL and LWJGL3 backends

bjsn:
LibGDX already supports playing AAC audio files on iOS and Android. However, so far there was no decoder for the desktop backends. I used the JAAD open source library (http://jaadec.sourceforge.net) to write a new AAC decoder for the LWJGL and LWJGL3 backends.

Please note that I am no expert in audio decoding. I was merely writing this code by following documentation and examples.


tyrondis:
@NathanSweet would you find some time to review this?

backends/gdx-backend-lwjgl/pom.xml
```diff
@@ -12,6 +12,13 @@
   <packaging>jar</packaging>
   <name>libGDX LWJGL Backend</name>
 
+  <repositories>
+    <repository>
+      <id>athou-repo</id>
+      <url>https://github.com/Athou/maven-repository/raw/master/repo</url>
```
intrigus: Why don't you use 
```xml
<id>central</id>
    <name>bintray</name>
    <url>http://jcenter.bintray.com</url>
````
?
See https://github.com/DV8FromTheWorld/JAADec for more information :)

Comment (tyrondis) [inline backends/gdx-backend-lwjgl/pom.xml]: Oh, I simply wasn't aware of that fork. I will check it out!

**Comment by tyrondis** [comment]
Can I get some feedback on the actual implementation, please?

File: backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/M4a.java [hunk 2]
```diff
@@ -0,0 +1,159 @@
...
+        public Sound(OpenALAudio audio, FileHandle file) {
+
+            super(audio);
+            if (audio.noDevice) return;
+
+            stream = new Stream(file);
+
+            ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+            byte[] buffer = new byte[40960];
```
Tom-Ski: Is this a typo, or did you actually want 40960? If so, why?

Tom-Ski:
Nice, looks good. Not sure about the library though. According to the apache, Apache2 software can be included in GPLv3 projects, but not the other way around.  If someone has more information on the compatibility of these two licenses, ye are summoned.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/M4a.java
tyrondis: TBH I don't remember. Perfectly willing to change, what is an advisable size? 4096 bytes?

tyrondis:
Thank you so much for the review, @Tom-Ski!

Where did you see that the library is GPLv3? Source Forge just says "Public Domain" and the author removed the LICENSE file with one of the latest commits (https://sourceforge.net/p/jaadec/code/157/)

Either way, found the following on http://www.gnu.org/licenses/license-list.html#apache2:

> This is a free software license, compatible with version 3 of the GNU GPL.

Tom-Ski:
https://github.com/jsiembida/jaad here.
And Apache's comments on compatibility http://www.apache.org/licenses/GPL-compatibility.html

tyrondis:
That fork seems to be outdated, not containing the commit I copied above where he removed the license.

Tom-Ski:
Ah I didn't see your message.  Yeah in that case its probably fine.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/M4a.java
NathanSweet: Ogg#Sound is using 2048. Wav#sound uses StreamUtils which defaults to 4096.

NathanSweet:
It looks fine to me, assuming there aren't license issues with JAAD.

backends/gdx-backend-lwjgl/pom.xml
tyrondis: We cannot use this fork, because it does not contain the commit of the original repository that removes the LICENSE file.

tyrondis:
Changed the buffer size and used the code formatter. There should be no licensing issue, as the license has been removed by the author, making it Public Domain.

loogie:
Hello!

 

You have a new message, please read <http://libertyapparels.com/primarily.php?75>

 

loogie@basicelementdesign.com

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/M4a.java
loogie: Hello!

 

You have a new message, please read <http://viettai.com/horror.php?40xor>

 

loogie@basicelementdesign.com

tyrondis:
It's been a month. Is there anything else you'd like me to change @NathanSweet @Tom-Ski @badlogic?

tyrondis:
@NathanSweet @Tom-Ski @badlogic Can you please merge this? I want to get rid of my fork.

tommyettinger:
Having AAC support would be great, but...

The license being removed doesn't make code public domain; it makes it unlicensed, just with what happens to be visible source. You could also see the source if you looked over the shoulders of programmers making Norton Antivirus, or even if some of them showed you what they were doing, but that product would still be very closed-source. In that case of unlicensed code with visible source, copyright doesn't change and the code is on unfortunately shaky legal grounds. If there's a copyright waiver like CC0 or the Unlicense in the original author's code, that would in most jurisdictions make it public domain, but not all, so in case of your code being used in North Korea or some other legal bizarro-land, there's usually an extra measure taken to avoid the possibility of legal events. I think Unlicense recommends dual-licensing with MIT, which is unambiguously allowed for commercial and non-commercial usage and similarly short. Licensing as BSD, MIT, Apache, or a number of other common "permissive software licenses" would allow libGDX to use this without question; MPL (Mozilla's older license) I think specifically has a clause to prevent its use with GPL code. Incorporating GPL code is not good if your lib is not already GPL, and for a product used in commercial closed-source apps, like almost any game framework, using the GPL is not an option.

I also don't know if GPL can be removed except by the permission of all contributors, it affects all code written with that license and that doesn't change authorship, so each author of original code in the library needs to declare, somehow, that they want the license changed.

Note, I am not a lawyer, but I also don't want to have to involve one to use libGDX.

tyrondis:
Yeah, I can see now, how this is tricky, though the library's author's commit that removed the license file says "[r157] removed license, changed to public domain" we still need a license, rather than none at all.

Let me get in touch with the author and see what we can do about that.

tommyettinger:
It looks like AAC support requires a pretty sizable library, though I may just be counting folders and not getting a reliable estimate of the size increase for an APK. Are there any alternatives out there, or any feasible ways to write a small AAC decoder from scratch or based off the reference documentation? I'm mostly wondering if there's a small, say, Gist out there somewhere, so that we could ask its author if we can incorporate it into libGDX. Many of libGDX's classes are unusually small for Java implementations that do similar things, which is really good for JAR/APK size and download time over the air or on GWT. It's also possible that the AAC lib, if the author is the only contributor, and since it looks like he's trying to declare it as public domain, then I think it should be OK to pull this in once we can show he has no interest in keeping the GPL on it or in suddenly enforcing a license he seems to want to drop.

Unless the author is in a place with an unusual legal system, a CC0 or Unlicense waiver should be all he needs, in one place, usually the root of the project (or the root of each module, if he has a project like libGDX split up into multiple download-able parts, and any build files that include licenses need updating, like pom.xml files sometimes have it). 

RE: unusual legal systems, not all countries allow copyright to be released to the extent the US does; in EU countries, you can give up copyright with a waiver like CC0 or the Unlicense, but you can't ever your give up your claim to authorship, which is weirdly different in subtle ways, even though those licenses still do their job just fine there. It's possible Cuba, Vietnam, and other countries that at least partly reject capitalism may not have signed any international treaties standardizing their positions on copyright law, and many places treat government employees' work differently (fun fact, just about every photo taken by NASA or the US Fish and Wildlife service, unless it's classified like NASA's likely satellite photos of war-zones, is public domain). If the author is in a country that may be still thinking the USSR will take them back to glorious communism some day, like Turkmenistan (which was very recently a dictatorship, I don't know about now), then we might need to be slightly more concerned about legal shenanigans, but if they're in almost anywhere in Europe, the US, a recent United Kingdom holding, most South American countries, probably Russia and China, etc. libGDX should be fine with pulling it once there's a copyright waiving (un?)license of some kind.

I'm not a dev so I can't say whether they'd actually want to pull this, but I think with the hassles people keep having with other audio formats and bitrates on a per-device basis, this could be great.

nhydock:
> It looks like AAC support requires a pretty sizable library, though I may just be counting folders and not getting a reliable estimate of the size increase for an APK

This should have no effect on the size of APK, as android already has AAC decoding.  This should be strictly for desktop/LWJGL.  

I half feel that something like this should be a pull request on upstream LWJGL3 instead of libgdx.

MrStahlfelge:
While it would be good to have more audio files supported, this clearly seems like a case it where it should be done with a 3rd party lib. It doesn't make sense to pull in a lib for everyone with its downsides, when most devs using libgdx don't need the feature. Additionally, the unclear license is a major showstopper.

Closing this, but feel free to publish this as an extension lib.

PR #3851, opened 2016-02-17
Fixed memory leaks on android

SaeedMasoumi:
I created [this sample project ](https://github.com/SaeedMasoumi/libgdx-android-leak
)to show that how libgdx creates memory leaks on android.

 In [LeakFragment](https://github.com/SaeedMasoumi/libgdx-android-leak/blob/master/android/src/io/saeid/libgdx/androidleak/fragment/LeakFragment.java) and [LeakActivity](https://github.com/SaeedMasoumi/libgdx-android-leak/blob/master/android/src/io/saeid/libgdx/androidleak/activity/LeakActivity.java) classes, i created a method called `whatShouldBeDoneOnSuperMethod` without this method, leak canary will show memory leak.


Tom-Ski:
THanks!

PR #3856, opened 2016-02-17
Add debug3dRenderer

realitix:
Follow this pull request
https://github.com/libgdx/libgdx/pull/3839

@xoppa

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
@@ -1,0 +1,220 @@
+package com.badlogic.gdx.graphics.g3d.utils;
+
+import com.badlogic.gdx.graphics.Camera;
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.Mesh;
+import com.badlogic.gdx.graphics.VertexAttribute;
+import com.badlogic.gdx.graphics.VertexAttributes.Usage;
+import com.badlogic.gdx.graphics.g3d.Material;
+import com.badlogic.gdx.graphics.g3d.Renderable;
+import com.badlogic.gdx.graphics.g3d.RenderableProvider;
+import com.badlogic.gdx.math.Vector3;
+import com.badlogic.gdx.math.collision.BoundingBox;
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.Disposable;
+import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.Pool;
+
+/** @author realitix */
+public class Debug3dRenderer implements Disposable, RenderableProvider {
+	private float vertices[];
+	private short indices[];
+	private int idVertice;
+	private int idIndice;
+	private Vector3 tmp = new Vector3();
+	private Vector3 tmp2 = new Vector3();
+	private Renderable renderable = new Renderable();
+	private boolean building;
+	private Color color0 = new Color();
+	private Color color1 = new Color();
+	private Color color2 = new Color();
+	private Color color3 = new Color();
+	private Color color4 = new Color();
+
+	public static int maxVertices = 5000;
+	public static int maxIndices = 5000;
+
+	public Debug3dRenderer () {
+		// Init mesh
+		Mesh mesh = new Mesh(false, maxVertices, maxIndices, new VertexAttribute(Usage.Position, 3, "a_position"),
+			new VertexAttribute(Usage.ColorPacked, 4, "a_color"));
+		// Init vertices
+		vertices = new float[maxVertices * (mesh.getVertexSize() / 4)];
+		// Init indices
+		indices = new short[maxIndices];
+
+		// Init renderable
+		renderable.meshPart.mesh = mesh;
+		renderable.meshPart.offset = 0;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,220 @@
...
+	private Renderable renderable = new Renderable();
+	private boolean building;
+	private Color color0 = new Color();
+	private Color color1 = new Color();
+	private Color color2 = new Color();
+	private Color color3 = new Color();
+	private Color color4 = new Color();
+
+	public static int maxVertices = 5000;
+	public static int maxIndices = 5000;
```
xoppa: These members are only used in the constructor and might as well be constructor arguments instead of static members.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,220 @@
...
+	private void vertice (Vector3 point, Color color) {
+		vertices[idVertice++] = point.x;
+		vertices[idVertice++] = point.y;
+		vertices[idVertice++] = point.z;
+		vertices[idVertice++] = color.toFloatBits();
+	}
+
+	/** Fill indices array with values
+	 * @param values */
+	private void setIndices (int... values) {
```
xoppa: Using varargs will create an array, let's not do that.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,220 @@
...
+
+	public void begin () {
+		if (building) throw new GdxRuntimeException("Call end() after calling begin()");
+		building = true;
+
+		idVertice = 0;
+		idIndice = 0;
+	}
+	
+	public void add (Camera camera) {
```
xoppa: I have a feeling that it should be able to accept a `Frustum` directly and rely on the camera (so a signature `add(Frustum frustum)` should be possible and this one could call `add(camera.frustum);`.

xoppa:
Thanks! So practically this is a class with a `begin()` and `end()` method and in between you can add "shapes" like a camera (frustum) or bounding box. That sounds very similar to `MeshBuilder`, except that `MeshBuilder` doesn't implement `RenderableProvider`. In fact, I can imagine that every "shape" method like sphere and such of `MeshBuilder` could fit in this class as well.

Let's break this into two parts:
- a class that bridges between `MeshBuilder` and `RenderableProvider`, but is more lightweight than `ModelBuilder` (no garbage)
- a method to add a frustum shape and bounding box to `MeshBuilder` (`MeshPartBuilder`)

For the first part I think this class could be a bit simplified. The `begin()` method could return the `MeshBuilder` (`MeshPartBuilder`), so the builder methods can be removed. That wouldn't share the mesh amongst multiple renderables, but that's fine in this case. Optionally it could use a `MeshPool` like `ModelCache` does.

For the second part, the methods could simply be added to `MeshBuilder`. However the number of methods of that class (and interface) keeps growing and I can imagine there are many other "shape" methods that could be useful.

Perhaps we should look into an alternative way. E.g. where `MeshBuilder` only contains the bare minimum to abstract away the primitive and such (so that it could be moved to jni e.g.). And some "create shape mesh" classes that implement the algorithm to create the specific shape. So that it would look something like this:

```java
MeshPartBuilder mpb = debugThing.begin();
ClassThatCreatesTheShapeOfTheCamera.DoYourThing(mpb, camera);
ClassThatCreatesTheShapeOfTheBounds.DoYourThing(mpb, boundingBox);
debugThing.end();
modelBatch.render(debugThing);
```

What do you think?

realitix:
@xoppa I think it's a very good idea mainly the second point. Indeed, meshBuilder is a very big file difficult to read. Moreover, it would allow to easily ad new shape type.

For your first point, I'm good with it too.

So wee need to refactor MeshBuilder and MeshPartBuiler to do this. Maybe it wouldn't be ready until the next release so you should revert the cameraHelper pull request.

Do yo want me to begin an implementation or do you want to do so ?

xoppa:
I've reverted #3856. It doesn't necessarily mean we have to refactor MeshBuilder right away. The old methods can still exist while moving towards the new methods. We could start by adding a camera frustum shape builder class with a static method that takes a MeshPartBuilder and Frustum/Camera. Could you do that?

realitix:
Yes I can Do that. 

- One class for the frustum (only frustum).
- One class for the camera which use the frustum class and add the up vector and camera position.

It's ok for theses two classes ?
It will take a little time because I have to study meshbuilder deeply.

xoppa:
Great. Thanks! Although I wouldn't mind it being one class, but that also depends on the amount of code and separation of concerns of course.

realitix:
@xoppa If I use the mesh builder, it will create a new mesh. So if I want for example to draw my camera every frame (a moving camera), it will create a new mesh every frame. What do you think ?

xoppa:
`MeshBuilder` can reuse the `Mesh` you provide it, you don't have to create a new `Mesh` every frame. Just pass the Mesh in to the call to `end`. Have a look at [`ModelCache`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/ModelCache.java) for an example.

xoppa:
Closing this. Please send a new PR when you find the time.

PR #3857, opened 2016-02-17
Prepare for gwt-freetype and other emscripten compiled apis

intrigus:
**(The diff of GwtApplication is a little bit confusing so please look at the result and not at the direct diff or use github's split view)**

For gdx-freetype-gwt or gwt-bullet we have to inject ceratin .js files but we have to do it before the main loop is running.

(Injection is async that's why we need this PR if it were sync. we would not have any problem :) )

The generator collects all `Injectable`s and injects them.
We wait 5s (50*100ms) for injection if the scripts aren't injected yet, we don't care, because the scripts are injected at the first possible time and if it isn't injected after 5 seconds it's unlikely that it will be injected.

The Timer code is a little bit ugly, I know, but I couldn't use libgdx Timer because it can only used if the mainloop is running.

The constructors of `DirectReadWriteByteBuffer`have been made public to easily transfer data from the emscripten heap to GWT/libgdx.

The example mentioned in the javadoc can be found [here](https://gist.github.com/intrigus/ee87be2c8427f62b754b).

Please add your suggestions @MobiDevelop @badlogic


scientistnm:
Any news when this is going to be merged?

MobiDevelop:
When someone has reviewed it and decided it is good to be merged.

scientistnm:
Very useful response :)

MobiDevelop:
That's the exact answer to the question you asked. We all do this in our spare time, the ETA for anything is when we have time so it is usually better to refrain from asking for an ETA.

vinou31:
Hi, does this PR would allow freeTypefontGenerator to be used in gwt?
If so, it is a very usefull feature. Indeed, we can't use this feature yet because we aim every platform including gwt. So we used only feature that are compatible with everything. Thanks for your time and your job.

vinou31:
I try to use your PR but when compiling gwt I encounter this issue, do you have any ideas?

```
> :tests:gdx-tests-gwt:compileGwt
> Compiling module com.badlogic.gdx.tests.gwt.GdxTestsGwt
>    Tracing compile failure path for type 'com.badlogic.gdx.backends.gwt.inject.JsInjectorGenerator'
>       [ERROR] Errors in 'file:/D:/Sources/libgdx/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/inject/JsInjectorGenerator.java'
>          [ERROR] Line 35: The method getPackage() is undefined for the type Class<JsInjector>
>    [ERROR] Aborting compile due to errors in some input files
> :tests:gdx-tests-gwt:compileGwt FAILED
```

MrStahlfelge:
The PR has nothing to do with emscripten, it is generally for any code that is in a js file. The normal process of including a js file is to place it in the webapp folder and include it in index.html. This approach was used by libgdx itself for the soundmanager.

Another approach is to inject js files with GWT's ScriptInjector 

         final String js = GWT.getModuleBaseForStaticFiles() + "freetype.js";
			ScriptInjector.fromUrl(js) 

It must be done at the very start of the application in `onModuleLoad`. 

The PR makes the process a bit more convinient, it's not needed to change HTMLLauncher or other files any more when including a lib supporting the process. On first sight, it looks a bit too complex for me. A lot of code is made to gather all available 3rd party libs using a certain interface and  generate code to inject their js files on gwt compilation. I am not sure if this is really needed: An overridable method in GwtApplication letting the game developer define a list of js files to inject should be okay, too. That's a matter of taste, but it could be used by game developers to inject other js file stuff (that is sometimes needed) with the same method.
But as said before, this is convinience - there's no need to have this. [FreeTypeFont can be used on GWT without this PR](https://github.com/intrigus/gdx-freetype-gwt) (with an addition to onModuleLoad), and [gdx-gameanalytics](https://github.com/MrStahlfelge/gdx-gameanalytics#gwt-project-setup) let the dev do the first approach.

Please correct me if I understood something wrong.

SimonIT:
I agree with @MrStahlfelge. It's a far to complex solution for just inserting a js file.
Maybe a slim version can be sufficient? Your FreetypeInjector seem to work without it. Something like you did there would be fine in my opinion. I like the idea of adding a js file without adding it to the html file by hand

MrStahlfelge:
@MobiDevelop is working on #5948, let's see if he needs this. If not, we can safely close it.

MobiDevelop:
I think, due to the async nature of loading and compiling the WebAssembly module we might need something like this. In my experiments with migrating box2d to use emscripten, only about 30% of the time have I seen it load quickly enough to be ready by the time the main loop is starting.

PR #3858, opened 2016-02-17
GWT support for sending float data to a texture.

barkholt:
Trying to use FloatTextureData on GWT failed due to the OpenGL wrapper around glTexImage2d unpacking the data into bytes.


badlogic:
Thanks

PR #3849, opened 2016-02-16
Add support for selecting modules through command line

Fr0sk:
This would help integrate LibGDX with other tools or game engines that leverage from it.

PR #3862, opened 2016-02-18
Add support for selecting extensions through command line

Fr0sk:
Together with my last pull request, this would help integrate LibGDX with other tools or game engines even further. If no extension is provided then it would have the default behaviour, so wouldn't break existing setups.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetup.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetup.java
@@ -477,13 +477,14 @@ public class GdxSetup {
 
 	private static void printHelp () {
 		System.out
-			.println("Usage: GdxSetup --dir <dir-name> --name <app-name> --package <package> --mainClass <mainClass> --sdkLocation <SDKLocation> [--excludeModules <modules>]");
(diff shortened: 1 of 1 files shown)


Fr0sk:
About this change and also the last one I made, #3849, when should I (or anyone for that matter) make the changes to the [Wiki page](https://github.com/libgdx/libgdx/wiki/Project-Setup-Gradle)?

MobiDevelop:
Generally speaking, once a change is merged that warrants an update to the wiki, the person who made that change should go update the wiki soon after.

Fr0sk:
Even if the changes are not available for the user? I mean, the updated setup is not available to download yet. If so, I would make the changes right away

MobiDevelop:
The updated setup is immediately available to users.

Fr0sk:
I may be doing something wrong, but I downloaded the gdx-setup.jar and it has not the new arguments implemented.

MobiDevelop:
Looks like the build has been hanging for a day or two.

Fr0sk:
Ok, I'll wait until the build comes out and make the changes to the wiki page. Thank you for your time and help :)

Tom-Ski:
Thanks!

PR #3865, opened 2016-02-20
Add updateIndices to IndexData

realitix:
Used with IndexBufferObjectSubData, it increases performance and allows
to update a portion of the indices.


realitix:
Oups Gwt not working. Why there is an emulation in GWT ?

realitix:
OK it's because of the byte buffer

xoppa:
Thanks!

PR #3870, opened 2016-02-22
Fixed scrollpane for large heights and exposed some variables

Xtrinity:
I had some issues with my fork and so this is a re-pull from #3462.

SSCCE: https://gist.github.com/Xtrinity/8b208181fdf05d562ea5

Explanation of issue:
If you scroll in the SSCCE you will see that the scroll overscrolls the areaHeight. I.E the topmost ID is 1 and the greatest ID that's viewable on the screen would be 14. From a UX standpoint I would expect a single scroll down from the mouse to show 15 as the next id instead of jumping to a non-consecutive number like 24.

Explanation of changes:
Exposed existing hard coded values and created two fields for them `smoothingFactor` and `smoothingSpeed`
Created two new fields to define the amount to scroll when the mouse is scrolled, `mouseWheelX` and `mouseWheelY`. If it is not changed by the user then it defaults to scrolling the width or height of the area. If the ScrollPane is not big enough it will automatically not scroll.

Note: I did not test Android or iOS but they shouldn't be affected from the changes committed.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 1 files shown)


NathanSweet:
I agree mouse wheel scroll should be limited to the area height. However, this PR always scrolls by one page, even when there are few items in the scrollpane. Change the items from 1000 to 100 to see what I mean.

Please see if the commit I just made sufficiently meets your needs.

Xtrinity:
@NathanSweet Yes, it does. Thanks :)

PR #3877, opened 2016-02-24
64-bit flag updated

entangledloops:
armv8 (64-bit) android accounted for


intrigus:
Looks like it won't always work, see http://stackoverflow.com/a/33752302.
It may be better to use http://developer.android.com/reference/android/os/Build.html#SUPPORTED_64_BIT_ABIS.

entangledloops:
@intrigus That stackoverflow user's post looks incorrect (i.e. their device is 64-bit according to the manufacturer, not sure how they determined it's 32). I've responded on stackoverflow with the relevant links.

intrigus:
i

xoppa:
Please verify your code before sending a pull request. Looks like you used github web editor for this?

xoppa:
Closing this since it doesn't compile. Please reopen when you fixed it.

PR #3881, opened 2016-02-27
Hiero GPOS extension support + pair format 2 fix

ccbrown:
Adds support for lookup type 9, which is just a way to address subtables with a 32-bit offset in large font files.

Also fixes pair format 2, which wasn't correctly populating class 0, which is a special case: For pair positioning, it gets everything in the coverage that isn't in any other class.

My testing for this has been with Apple's San Francisco fonts, which can be found here: https://developer.apple.com/fonts/

In particular, San Francisco UI display ultralight is one that requires both of these changes to export kerning information properly.

CLA will be signed and emailed shortly.


ccbrown:
CLA sent.

PR #3895, opened 2016-03-02
Fix Music.isPlaying() on GWT backend

ifcaro:


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtMusic.java
```diff
@@ -74,7 +74,7 @@ public void stop () {
 
 	@Override
 	public boolean isPlaying () {
-		isPlaying &= sound.getPlayState() == 1;
+		isPlaying = sound.getPaused() == false & sound.getPlayState() == 1;
```
davebaol: I'd say that `&&` should be used instead.
Also, `sound.getPaused() == false` is just `!sound.getPaused()`

ifcaro: changed

PR #3902, opened 2016-03-05
Fix #3838

### LINKED ISSUE #3838: Json integer parsing
In Json the last attribute of an object doesn't have a comma:

```
{
	"x": 17,
	"y": 0
}
```

When the last attribute is an integer the JsonValue getInt method throws an exception, because it thinks the blank space is part of a String. 

The exception is the following:
```
java.lang.NumberFormatException: For input string: "0	"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at com.badlogic.gdx.utils.JsonValue.asInt(JsonValue.java:237)
	at com.badlogic.gdx.utils.JsonValue.getInt(JsonValue.java:680)
```

I think this can be fixed using the trim() method, in order to delete blank spaces of the String.

tuslatom:
Parser was checking only spaces, not other whitespace characters and if I understand it correctly Character.isWhiteSpace() is not supported by GWT, but isSpace() is.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/JsonReader.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonReader.java
@@ -92,13 +92,13 @@ public class JsonReader implements BaseJsonReader {
 
 		try {
 
-			// line 3 "JsonReader.java"
+			// line 97 "JsonReader.java"
 			{
 				cs = json_start;
 				top = 0;
 			}
 
-			// line 8 "JsonReader.java"
+			// line 103 "JsonReader.java"
 			{
 				int _klen;
 				int _trans = 0;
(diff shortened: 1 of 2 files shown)


xoppa:
> `// Do not edit this file! Generated by Ragel.`

davebaol:
@tuslatom 
You should change `JsonReader.rl` and use Ragel [like that](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/JsonReader.rl#L2) to generate the new `JsonReader.java`.

tuslatom:
thanks for the hint, I've changed .rl file and generated new java file

PR #3903, opened 2016-03-06
Lwjgl2 backend displayMode tweaks

IonAgorria:
Seems that getDisplayModes(Monitor) only returns the getDisplayMode() result
which in turn getDisplayMode() doesn't return the current mode but the initial one that was the window created with (using getDesktopDisplayMode())


Tom-Ski:
Thanks!

PR #3906, opened 2016-03-07
Fix #1775 ensure loading of sound on Android

### LINKED ISSUE #1775: No way to know if a Sound has finished loading.
On Android, if a Sound is not fully loaded and you try to play it, it will tell you "Sample not ready" in the log and not do anything. This causes problems as in some cases, like the one I have, I need to be able to load sounds, wait for them to finish loading, and play them, and I can't load them ahead of time. It would be helpful if a method like "isReady" was added to Sound.

intrigus:
@badlogic I hope that is what you had in your mind :)

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
@@ -27,8 +27,10 @@ import android.content.res.AssetFileDescriptor;
 import android.media.AudioManager;
 import android.media.MediaPlayer;
 import android.media.SoundPool;
+import android.media.SoundPool.OnLoadCompleteListener;
 
 import com.badlogic.gdx.Audio;
+import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Files.FileType;
 import com.badlogic.gdx.audio.AudioDevice;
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Not a huge fan of this being the default.  Have you run any tests to see how this impacts loading times? Especially when batch loading audio files?

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
```diff
@@ -174,21 +177,46 @@ public Sound newSound (FileHandle file) {
...
+			public void onLoadComplete (SoundPool soundPool, int otherSampleId, int status) {
+				if (sampleId == otherSampleId && status == 0)
+					loaded[0] = true;
+				else if (sampleId == otherSampleId && status != 0) { // Should theoretically never happen
+					Gdx.app.log(TAG, "Sample with id " + sampleId + "has been loaded with unexpected status code " + status + " !");
+					failed[0] = true;
+				}
+			}
+		});
+		while (loaded[0] || failed[0]);
```
xoppa: Forgot a Sleep in here?

badlogic:
Interesting solution, but i'd say this is to risky cause Android OEMs tend to fuck audio up so badly, that i don't trust them to actually call the listener in every case.

PR #3909, opened 2016-03-08
+ETC1 Atlas Compressor

JFixby:
Back-story: https://twitter.com/badlogicgames/status/643082568334745600

```ETC1AtlasCompressor``` takes atlas produced by the ```TexturePacker``` and converts it to ```ETC1 ``` format by:
 - compressing textures from ```PNG``` to ```ETC1```,
 - changing texture-file names accordingly in the atlas description file.

To test the ```ETC1AtlasCompressor``` please run the ```ETC1AtlasCompressorTest```, the test will:
 - prepare example atlas using sprites in the sprites folder,
 - compress the atlas,
 - load and render compressed sprites on the screen using gdx desktop backend.

### CODE REVIEW

**Comment by xoppa** [comment]
Thanks for your contribution @JFixby. It looks like it can be very useful. I recall though that etc1 doesn't support transparency. The majority of texture packer usage probably needs transparency. Even if all images are opaque then the padding still is transparent to prevent bleeding of one region into the other (unless a pixel perfect projection of course). You could work around that by packing the transparency into the color or using a separate alpha image. But that would require a different shader and perhaps an extra image or less precision, which might or not defeat the purpose of using etc1 in the first place. So I'm entirely convinced of a use-case of this. Can you provide some more insight on this?

**Comment by JFixby** [comment]
@xoppa 
ETC1 does not support transparency, at least in the scope of this PR. Before packing the ETC1Compressor automatically converts textures to RGB888 or RGB565.

Padding:
Since ETC1AtlasCompressor  takes packed atlas as input, the textures suppose to be already padded at that point.

Transparency: 
I think it is possible to implement transparency using a special shader. You no need a separate alpha image if you paint transparent pixels with fuxia color, like it was done the first StarCraft.

**Comment by xoppa** [comment]
Sorry, I meant to ask whether you can provide more insight in a practical use-case of this. Why would anyone use this and how does it compare to not using it? Perhaps you can provide your reference images or benchmark results? Perhaps you can give a practical example of a valid use-case, to clarify/estimate how widely this would be used?

**Comment by JFixby** [comment]
Well I was intended to render huge animated background hoping to save some memory.

**Comment by MobiDevelop** [comment]
It seems like if you have a huge background, you probably don't want to use an atlas for it.

**Comment by JFixby** [comment]
Ok I just did some bench-marking:
1) I have an atlas file with number of textures 2.627 Mb each, 112 Mb in total
Loading time for the regular atlas is 2.402 sec,

2) The same atlas, but etc1 compressed: 0.362 Mb each, 15 Mb total
Compressed - 0.835 sec

**Comment by xoppa** [comment]
Thanks @JFixby. So, if I understand correctly, the practical example use-case is: *to reduce the disk space and time to load of a huge animated background*. As @MobiDevelop said, you probably don't want to use an atlas for that (it will not help in sprite batching, which is the primary reason to use an atlas).

Although I understand the technically reasoning behind your approach, I am not yet convinced of the practical added value of this functionality. That said; if there is a wide enough use-case for this then I think that it might as well be integrated in TexturePacker as option. I don't see why this should be a separate tool.

**Comment by JFixby** [comment]
Thank you @xoppa Let me sort it out.

- The purpose of this PR is to introduce a PNG-to-ETC1 compressor for atlases. It takes packed PNG-atlas for input and produces ETC1-atlas on output.

- The reason ```ETC1AtlasCompressor``` is not merged with the ```TexturePacker``` is because in this way it is possible to compress already packed atlases without repacking them from the start. 

- How can someone benefit from PNG-to-ETC1 compressor. The answer is: the same way anyone can benefit from texture-compression in general. Texture compression intended to reduce the disk space usage and load-time of a game.

- This PR is not addressing transparency because it is outside of it's scope and gdx-engine's in general. ETC1 by it's definition suppose to be used with opaque textures.

Transparency can be "introduced" in ETC1 by preliminary conversion of PNG-textures with transparent pixels to PNG-textures with fuxia color instead of the transparent pixels. Only then the output of this conversion can be compressed with ETC1. Anyway all of it is happening before applying ```ETC1AtlasCompressor``` so it is not related to this PR.

Not sure if it is relevant here, but I have some problems with animated background and loading time on Android, plus a heavy assets folder. So I decided to use compressed textures and atlases on Android back-end; and the regular atlases on desktop. Applying ```SpriteBatch``` wouldn't increase render performance, but I probably no need that. My main concern is to reduce assets folder weight and game-load time.

**Comment by JFixby** [comment]
On the second thought I probably can add ```replaceTransparentPixelsWithFuxia``` - flag to the ```ETC1AtlasCompressorSettings```

**Comment by JFixby** [comment]
I think ```ETC1Compressor.process()``` should actually replace transparent pixels with the fuchsia.

File: extensions/gdx-tools/src/com/badlogic/gdx/tools/etc1/ETC1AtlasCompressorSettings.java [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+ ******************************************************************************/
+
+package com.badlogic.gdx.tools.etc1;
+
+import com.badlogic.gdx.graphics.Color;
+
+public class ETC1AtlasCompressorSettings {
+
+	private String atlasFilePathString;
+	private Color transparentColor;
```
JFixby: @Darkyenus and here we benefit from the settings class.

JFixby:
Now ```ETC1AtlasCompressor``` and ```ETC1Compressor``` can specify what color to use to replace transparent pixels of the original texture. Before it was just black.

extensions/gdx-tools/src/com/badlogic/gdx/tools/etc1/ETC1AtlasCompressorSettings.java
Darkyenus: But not from factory pattern

extensions/gdx-tools/src/com/badlogic/gdx/tools/etc1/ETC1Compressor.java
```diff
@@ -58,17 +63,31 @@ protected void processDir (Entry entryDir, ArrayList<Entry> value) throws Except
...
+				this.transparentColor.set(transparentColor);
+			}
+		}
 	}
 
-	public static void process (String inputDirectory, String outputDirectory, boolean recursive, boolean flatten)
-		throws Exception {
+	public static void process (String inputDirectory, String outputDirectory, boolean recursive, boolean flatten,
+		Color transparentColor) throws Exception {
 		GdxNativesLoader.load();
```
JFixby: @Darkyenus and here we don't.
Java Rule #44: "Avoid using more than one method/constructor argument whenever is possible."

extensions/gdx-tools/src/com/badlogic/gdx/tools/etc1/ETC1AtlasCompressorSettings.java
JFixby: I don't have to fix the whole call-tree though.

JFixby:
@xoppa Something like this: https://github.com/JFixby/TextureCompressor No shades used.

PR #3913, opened 2016-03-09
FreeTypeFontGenerator-texture-bleeding-#3521-fix 

JFixby:
Fixing https://github.com/libgdx/libgdx/issues/3521
- ```PixmapPacker``` has a new parameter ```Color currentTransparentColor;``` used by ```Page``` constructor to fill transparent pixels.
- ```FreeTypeFontGenerator``` sets packer to use color of the font but with zero alpha for transparent pixels to avoid leaks.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -345,6 +345,8 @@ public class FreeTypeFontGenerator implements Disposable {
 			}
 			ownsAtlas = true;
 			packer = new PixmapPacker(size, size, Format.RGBA8888, 1, false, packStrategy);
+			packer.setCurrentTransparentColor(parameter.color);
+			packer.getCurrentTransparentColor().a = 0;
 		}
 
 		if (incremental) data.glyphs = new Array(charactersLength + 32);
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -347,6 +347,10 @@ public class FreeTypeFontGenerator implements Disposable {
 			packer = new PixmapPacker(size, size, Format.RGBA8888, 1, false, packStrategy);
 			packer.setCurrentTransparentColor(parameter.color);
 			packer.getCurrentTransparentColor().a = 0;
+			if (parameter.borderWidth > 0) {
+				packer.setCurrentTransparentColor(parameter.borderColor);
+				packer.getCurrentTransparentColor().a = 0;
+			}
 		}
 
 		if (incremental) data.glyphs = new Array(charactersLength + 32);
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -101,6 +101,7 @@
 	Format pageFormat;
 	int padding;
 	boolean duplicateBorder;
+	Color currentTransparentColor = new Color(0f, 0f, 0f, 0f);
```
MobiDevelop: Why not just transparentColor/getTransparentColor/setTransparentColor? I think that one can reasonably assume that they are getting/setting the current value.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -345,6 +345,8 @@ public FreeTypeBitmapFontData generateData (FreeTypeFontParameter parameter, Fre
 			}
 			ownsAtlas = true;
 			packer = new PixmapPacker(size, size, Format.RGBA8888, 1, false, packStrategy);
+			packer.setCurrentTransparentColor(parameter.color);
```
MobiDevelop: If a border and/or shadow is being used, it might make sense to use one of those instead of the glyph color.

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
JFixby: - ```getTransparentColor``` means that it belongs to the ```PixmapPacker```.
- ```getCurrentTransparentColor``` means that it belongs to the current state of the ```PixmapPacker```, you can vary it between glyphs/pages/etc for your reasons.

MobiDevelop: Not sure I agree with your reasoning. This is not really any different than SpriteBatch's getColor/setColor.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
JFixby: Ok I added 
```
packer.getCurrentTransparentColor().a = 0;
			if (parameter.borderWidth > 0) {
				packer.setCurrentTransparentColor(parameter.borderColor);
				packer.getCurrentTransparentColor().a = 0;
			}
```
for border.

JFixby: Shadow cannot be simply fixed. That will require adding major feature like TextureBleeder (https://github.com/JFixby/TextureBleeder) to the ```FreeTypeFontGenerator```. 

I may suggest gdx-users never to use shadow feature and draw your string twice instead: one pass for the shadow and one for the actual font.

JFixby:
I may suggest the following solution to fix the font shadow.
1) Merge this PR
2) Modify ```FreeTypeFontGenerator``` so it will
- generate regular font pages
- then switch ```packer.setCurrentTransparentColor(shadow color);``` 
- generate shadow font pages

3) Modify font render so it will
- render glyph shadow
- then render the actual glyph.

JFixby:
Should be good now.

PR #3919, opened 2016-03-12
Fixed MapObject x/y properties being scaled twice

andrejp88:
The x and y local variables are already scaled on lines 175 and 176 in the BaseTmxMapLoader, and scaling them again when modifying the properties ObjectMap makes the x and y properties too small (and inconsistent with the coordinates of Rectangle/Ellipse/etc if the MapObject in question is an instance of one of those objects).


MobiDevelop:
Thanks

PR #3923, opened 2016-03-12
ETC1Data output streams fix

JFixby:
About this PR:
 - closes ```DataOutputStream ```  and ```DataInputStream ```
 - closes ```GZIPOutputStream``` and ```GZIPInputStream```
 - closes ```FileOutputStream``` and ```FileInputStream``` to avoid locking the file. In Windows not closing file streams (e.g. ``` new DataOutputStream(new GZIPOutputStream(file.write(false)));```) locks the file folder until reboot or else.
 - adds  ```ETC1Data .write(OutputStream os)```
 -  ```ETC1Data.write(FileHandle file)``` now wraps the ```ETC1Data.write(OutputStream os)```
 - adds  ```static ETC1Data.read(InputStream os)```
 -  ```ETC1Data(FileHandle file)``` now wraps the ```ETC1Data.read(FileHandle file)``` to allow ``` ETC1Data .read(InputStream os)```


gdx/src/com/badlogic/gdx/graphics/glutils/ETC1.java
```diff
@@ -114,10 +129,13 @@ public void write (FileHandle file) {
...
 				}
+
+				write.close();
+
+				gzip.close();
+
 			} catch (Exception e) {
-				throw new GdxRuntimeException("Couldn't write PKM file to '" + file + "'", e);
-			} finally {
-				StreamUtils.closeQuietly(write);
```
intrigus: Why did you remove this?
Now an exception will be thrown when closing the file failed.
Previously nothing would have happened. I don't see any reason to change this behavior.

JFixby: Why would closing the file failed?
In Windows not closing file output stream locks the output folder until reboot or else.

Darkyenus: Stream closing should still be in a finally block though. As you said, it is important to close it. And when it is there, closeQuietly has the benefit that it does the null check for you.

JFixby: Can you please explain my why to close stream, when it is already closed by ```write.close();```?

Darkyenus: If anything between 123 (`gzip = new GZIPOutputStream(outputStream);`) and 133 `gzip.close();` throws an exception, the streams will be left unclosed. And even worse, in `write(FileHandle)`, if anything on 106 in `write(OutputStream)` throws an exception, the file stream won't get closed.

JFixby: Ok I'll fix that.

Darkyenus: No problem, we're all always learning something new.

Btw. `os.close();` on 107 is now redundant and the finally should be probably used in `write(OutputStream)` as well.

JFixby:
@Darkyenus @intrigus 
I'm closing only the file-```OutputStream``` because it can affect the file system state.

Darkyenus:
It is a good practice to explicitly close other streams as well. For example in this case, GZIPOutputStream will very likely compress and flush buffered data on close. So if it isn't flushed (=closed in this case, they are not the same, some streams may append a footer on close) the file will be missing last few bytes (or in extreme case the file will be empty).

Since these bugs are hard to find, it is always a good practice to close streams properly, even if you are sure that it is a no-op. (And it is most likely not in this case)

JFixby:
I did some tests. Everything looks fine.

JFixby:
Should be good now.

JFixby:
Ok, I also updated reading from file and stream.

Darkyenus:
This is getting pretty big and out of original scope. Smaller PRs are easier to review and thus more likely to be merged :)

JFixby:
Sure, I'll resubmit it.

PR #3927, opened 2016-03-13
remove texture artifacts for FreeTypeFontGenerator

JFixby:
Fixing https://github.com/libgdx/libgdx/issues/3521
- ```PixmapPacker``` has a new parameter ```Color currentTransparentColor;``` used by ```Page``` constructor to fill transparent pixels.
- ```FreeTypeFontGenerator``` sets packer to use color of the font but with zero alpha for transparent pixels to avoid leaks.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@
 - Added TextureArray wrapper see https://github.com/libgdx/libgdx/pull/3807
 - Fixed bug in AndroidGL20.cpp which cast a pointer to a 32-bit int. Crash on 64-bit ARM, but only for a specific code path and address...
 - Fixed multiple controllers registering on same index with LWJGL3, see https://github.com/libgdx/libgdx/issues/3774
+- Fixed the FreeTypeFontGenerator texture bleeding, see https://github.com/libgdx/libgdx/issues/3521
 
 [1.9.1]
 - API Change: Override GwtApplication#createApplicationListener() to create your ApplicationListener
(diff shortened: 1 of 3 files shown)


MobiDevelop:
Can you please try not to open and close 3 PRs per change request?

JFixby:
I am sorry for disturbing. Still learning the craft.

intrigus:
Then I suggest that you first learn git and after that send PRs.

Create a sample repo and try git there. 

It's nothing personal, but 6 PRs per change are a little bit annoying.

badlogic:
Please add docs, please add an entry to CHANGES. Why do youmset alpha to 0 on the transparent color?

JFixby:
Transparent color is intended to replace all the pixels with alpha=0; When new glyphs page created it is filled with the ```transparentColor``` before drawing glyphs on top.

NathanSweet:
PR seems OK to me. Could leave alpha, please can be expected to set a valid transparent color.

I expect it you use PMA the problem would go away entirely.

JFixby:
I'm sorry @NathanSweet but I didn't understand your comments.

Let me give some details about the PR. Current master code does the following:
1) FreeTypeFontGenerator creates PixmapPacker
2) PixmapPacker creates pages with default ARGB=#00000000
3) FreeTypeFontGenerator  draws glyphs with the specified font color.
4) When the specified font color is not equal to the default color from the step (1) the glyphs leak, because alpha mask border is getting blurred and reveals transparent pixels.

In the PR:
1) FreeTypeFontGenerator creates PixmapPacker and sets it's default color ARGB=#00RRGGBB where RRGGBB are taken from the font color - https://github.com/libgdx/libgdx/pull/3927/files#diff-0b6e873945e2b28ba82f95c11ecf9667R348
2) PixmapPacker creates pages with the default ARGB=#00RRGGBB - https://github.com/libgdx/libgdx/pull/3927/files#diff-7d471251b67a93e5a37864c211ffd868R353
3) FreeTypeFontGenerator  draws glyphs with the specified font color.
4) The font color is equal to the default color from the step (1) so the glyphs do not leak, because revealed transparent pixels has the same color as their non-transparent neighbors.

So
A) You can set any transparent color here: https://github.com/libgdx/libgdx/pull/3927/files#diff-7d471251b67a93e5a37864c211ffd868R598 alpha is not affected or anything.
B) Not sure if the  premultiplied alpha is relevant here.

PR #3933, opened 2016-03-16
Modify at VertexAttribute,want more control

leeho0108:
Change the private constructors to public.Let users  have more control on VertexAttribute,like to set attribute data type ....
I saw the code of `com.badlogic.gdx.graphics.glutils.VertexArray.bind(ShaderProgram, int[])`,It seems supported GL_FLOAT and `GL_UNSIGNED_BYTE`,But when I what to create a `VertexAttribute` with type `GL_UNSIGNED_BYTE`,I found the constructors is private,so I think, it may be public to provide more control on Vertex Attribute?
Or,there has some reason must keep this constructors private?

Sorry for my english.
best regards

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
+++ b/gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
@@ -64,11 +64,11 @@ public final class VertexAttribute {
 				usage == Usage.ColorPacked, alias, index);
 	}
 	
-	private VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias) {
+	public VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias) {
 		this(usage, numComponents, type, normalized, alias, 0);
 	}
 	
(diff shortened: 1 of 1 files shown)


leeho0108:
Maybe to support this,`com.badlogic.gdx.graphics.glutils.VertexData` have to add method to return the `ByteBuffer`?If not do that, even I can create a `GL_UNSIGNED_BYTE ` Vertex Attribute,but I can not set it's value?Because I can only get a `FloatBuffer`?
Sorry for my english.
Best regards

xoppa:
Theoretically this has quite some uses. But Java doesn't support structs, so without hacking around that (like is done for color), it won't be practical and probably only introduces problems. What is your practical real world use-case?

If we include this then at least it should have sufficient javadocs which includes a warning.

leeho0108:
@xoppa 
Like you say,Simply make `public` may just make more work due to support what i want.
make them `protected` would be nice,since constructors be `protected`,someone who want more control of `VertexAttribute` could find a way to do some extensions.
Thank you for think about this!
Sorry for my english.
Best regards

PR #3936, opened 2016-03-17
Modify at VertexAttribute

leeho0108:
Like say in #3933 ,just make `private` to `protected`

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
+++ b/gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
@@ -64,11 +64,11 @@ public final class VertexAttribute {
 				usage == Usage.ColorPacked, alias, index);
 	}
 	
-	private VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias) {
+	public VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias) {
 		this(usage, numComponents, type, normalized, alias, 0);
 	}
 	
-	private VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias, int index) {
+	public VertexAttribute (int usage, int numComponents, int type, boolean normalized, String alias, int index) {
 		this.usage = usage;
 		this.numComponents = numComponents;
 		this.type = type;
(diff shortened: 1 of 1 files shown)


xoppa:
Since you are making these constructors part of the api, please add the necessary javadocs. Including a warning.

xoppa:
Whats the status of this @leeho0108?

leeho0108:
@xoppa 
Sorry, I almost forgot this ,because I am recently to busy.
I add some docs to this two protected constructor in my fork
https://raw.githubusercontent.com/leeho0108/libgdx/master/gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
But my English is too bad...
So, It will be great if you could look at it first and correct any words. 
Thanks

badlogic:
@xoppa can you fix up the docs and merge? Looks OK to me.

gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
```diff
@@ -16,8 +16,13 @@
 
 package com.badlogic.gdx.graphics;
 
+import java.nio.FloatBuffer;
+
+import javafx.beans.binding.FloatBinding;
```
Darkyenus: What's up with this?

leeho0108:
@Darkyenus  It's my mistake,I had remove this useless import.

cypherdare:
@xoppa 
`usage` is an obsolete parameter since we no longer support fixed pipeline. I think it can remain in the constructor for backward compatibility, but there's no reason for storing its value or using it in `hashCode()`. And `getKey()` could use `type` instead.

Then, all cases (such as in VertexAttributes.calculateOffsets()) of `if (attribute.usage == VertexAttributes.Usage.ColorPacked)` could be replaced with `if (attribute.type == GL20.GL_UNSIGNED_BYTE)`, and these protected constructors would be safer to use. They could even be public. And we then we could simply copy-paste the docs from the other constructors.

cypherdare:
@xoppa Hold on, I see that you're using Usage for additional info in the 3D api. I guess it should be kept, but docs updated since they mention in multiple places that they're only for fixed function. I still think using `if (attribute.usage == VertexAttributes.Usage.ColorPacked)` anywhere should be changed.

xoppa:
Thanks @leeho0108! Sorry for the delay.

Care to send a PR for that @cypherdare?

cypherdare:
@xoppa, take a look at #4395. It could use your input.

PR #3937, opened 2016-03-17
LWJGL3 Added the possibility to create windows initially hidden

ClaasJG:
Hey folks,

in the moment it is possible to hide windows using Lwjgl3Window#setVisible. This pull requests adds the option to configure windows to be hidden on creation by default. This is handy if the Lwjgl3Application is supposed to only 'spawn' windows using #newWindow. Or (the case why I made the request) if you desire to do offscreen rendering.

While I am not sure if this would be a widely used feature it is a small change and therefor worth considering imho.

-ClaasJG

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -320,7 +320,8 @@ public class Lwjgl3Application implements Application {
 		appConfig.setWindowListener(config.windowListener);
 		appConfig.setFullscreenMode(config.fullscreenMode);
 		appConfig.setTitle(config.title);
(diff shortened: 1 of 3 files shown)


intrigus:
Could you please apply the formatter to your changes?

ClaasJG:
It seems like the formatter changed a lot of code I never touched. I dont know how much of an issue that is? (And I am sorry I forgot the formatter, it was not my intention to make Nate upset. :)

js78:
@ClaasJG Why don't you use Framebuffer to do offscreen rendering ?

Darkyenus:
@js78 If I understand correctly, this is not for offscreen rendering when you have window up but when you want to do some OpenGL rendering without _any_ windows visible (sort of headless). If it is as I understand it, it could be useful for tools which need OpenGL but don't want to display anything. (And I could use just that!)

ClaasJG:
@Darkyenus exactly

badlogic:
Awesome, thanks!

Issue #3938, opened 2016-03-17
UBJsonWriter writes wrong types for long and double arrays


MartinWeindel:
The methods
public UBJsonWriter value (long[] values)
and 
public UBJsonWriter value (double[] values)
of the class com.badlogic.gdx.utils.UBJsonWriter
write wrong type information. The resulting stream is not readable, neither by UBJsonReader nor by 
other UBJSON implementations.

#### Issue details
According to the UBJson specification, long arrays must start with "$L", double arrays with "$D".
The UBJsonWriter class wrongly uses "$I" (short values) and "$d" (float values).

#### Version of LibGDX and/or relevant dependencies
1.9.2

Pull request for fixing these bugs will follow.

MartinWeindel:
Bug fix see pull request #3939

PR #3939, opened 2016-03-17
UBJSON reader/writer: fix for writing long/double/char arrays

MartinWeindel:
This is a bug fix for writing and reading long, double and char arrays in UBJSON.

see Issue #3938


MobiDevelop:
Thanks

PR #3940, opened 2016-03-17
Fix issue with underscores in method names

intrigus:
Methods with a underscore, e.g. `_foo()` can't be used with jnigen, because they are 
transformed to `__1foo()` according to the rules defined in table 2-1 at 
http://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/design.html#wp615
They can't be used, because jnigen always looks for "_" + methodName so jnigen looks for a 
function `__foo()` which doesn't exist.


Tom-Ski:
To confirm you want jni-gen to support people not adhering to java style standards?

intrigus:
Yes, because it's convenient to have `void foo ()` and `native void _foo ()`.
I don't care if you don't merge it, it's already fixed in my fork :)

PR #3942, opened 2016-03-17
Fixes #3509 and adds a simple test program

### LINKED ISSUE #3509: Bullet wrapper issue
The Bullet wrapper currently doesn't allow a btPairCachingGhostObject to getAllContactManfolds. I've tracked this bug down to how a btBroadphasePair temporary instance is stored when trying to access an index in btBroadphasePairArray. It's an easy fix by storing the temporary btBroadphasePair instance in Java instead of C++. This doesn't change how the API is exposed and it looks like this is the desired behaviour anyway (see btBroadphasePair.java).

```
	/** Temporary instance, use by native methods that return a btBroadphasePair instance */
	protected final static btBroadphasePair temp = new btBroadphasePair(0, false);
	public static btBroadphasePair internalTemp(long cPtr, boolean own) {
		temp.reset(cPtr, own);
		return temp;
	}
```

I'll add to this tomorrow with an example to reproduce the results, but the following three file changes fixes the issue ([this guy](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=19830) had a similar issue). I didn't make a pull request because I'm having no luck with swig. Any tips would be appreciated.

*jni/swig-src/collision/collision_wrap.cpp*:
```
SWIGEXPORT jlong JNICALL Java_com_badlogic_gdx_physics_bullet_collision_CollisionJNI_btBroadphasePairArray_1at(JNIEnv *jenv, jclass jcls, jlong jarg1, jobject jarg1_, jint jarg2) {
  jlong jresult = 0 ;
  btAlignedObjectArray< btBroadphasePair > *arg1 = (btAlignedObjectArray< btBroadphasePair > *) 0 ;
  btBroadphasePair *result = 0 ;
  int arg2 ;

  (void)jenv;
  (void)jcls;
  (void)jarg1_;
  arg1 = *(btAlignedObjectArray< btBroadphasePair > **)&jarg1;
  arg2 = (int)jarg2;
  result = (btBroadphasePair *) &((btAlignedObjectArray< btBroadphasePair > const *)arg1)->at(arg2);
  *(btBroadphasePair **)&jresult = result;
  return jresult;
}
```

*jni/swig-src/collision/com/badlogic/gdx/physics/bullet/collision/btBroadphaseArray.java:*
```
public btBroadphasePair at(int n) {
  return btBroadphasePair.internalTemp(CollisionJNI.btBroadphasePairArray_at(swigCPtr, this, n), false);
}
```

*jni/swig-src/collision/CollisionJNI.java:*
```
public final static native long btBroadphasePairArray_at(long jarg1, btBroadphasePairArray jarg1_, int jarg2);
```

jsjolund:
Tested on Desktop and Android platforms.

The test program is not very efficient, since the ```btOverlappingPairCache```, ```btCollisionAlgorithm```, ```btPersistentManifold``` and ```btManifoldPoint``` classes do not use pooling.

### DIFF
--- a/extensions/gdx-bullet/jni/swig-src/collision/collision_wrap.cpp
+++ b/extensions/gdx-bullet/jni/swig-src/collision/collision_wrap.cpp
@@ -2368,6 +2368,9 @@ SWIGINTERN bool ContactListener_setEvents(ContactListener *self){
  /*SWIG_JavaArrayArgout##Longlong(jenv, jarr$argnum, (long long *)$1, $input);*/ 
  /*SWIG_JavaArrayArgout##Float(jenv, jarr$argnum, (float *)$1, $input);*/ 
  /*SWIG_JavaArrayArgout##Double(jenv, jarr$argnum, (double *)$1, $input);*/ 
+SWIGINTERN btBroadphasePair *btAlignedObjectArray_Sl_btBroadphasePair_Sg__at(btAlignedObjectArray< btBroadphasePair > *self,int n){
+		return &(self->at(n));
+	}
 SWIGINTERN int btAlignedObjectArray_Sl_btBroadphasePair_Sg__getCollisionObjects(btAlignedObjectArray< btBroadphasePair > *self,int result[],int max,int other){
 		static btManifoldArray marr;
 		const int n = self->size();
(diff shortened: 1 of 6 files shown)


xoppa:
Thanks!

PR #3944, opened 2016-03-18
Fix reference leak in ComparableTimSort

vbousquet:
At the moment ComparableTimSort use a ~~static~~ temp array for performing the sort. The array is not clean up after sorting, therefore maintaining ~~static~~ references to the sorted objects.

This pull request simply mimics what is being done in TimSort which correctly cleanup these references after sorting.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
+++ b/gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
@@ -51,6 +51,7 @@ class ComparableTimSort {
 
 	/** Temp storage for merges. */
 	private Object[] tmp;
+	private int tmpCount;
 
 	/** A stack of pending runs yet to be merged. Run i starts at address base[i] and extends for len[i] elements. It's always true
 	 * (so long as the indices are in bounds) that:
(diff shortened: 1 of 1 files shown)


JFixby:
What is the purpose of the ```tmpCount```?

vbousquet:
tmpCount holds the size of the array that is effectively used (it grows non linearly see ensureCapacity) to only set to null what is needed.

Note that the proposed change is just to sync with the fix that was added to TimSort for the same problem

JFixby:
Which one is the static temp array?

vbousquet:
`tmp `is not static but `ComparableTimSort` is used statically by `Sort`. `Sort` is used internally inside LibGdx for example to implement `Array.sort`. Therefore, as soon of at least one call to sort is made, a static reference is created and persisted.

The tmp array is resized by jumping to next power of 2. It may jump to fairly igger than needed size (ot avoid frequent reallocation). The `tmpCount `avoids nulling out an oversized array ; it just null out the used part of the array.

xoppa:
Can you include a link to the source you based it on?

vbousquet:
`TimSort` in `com.badlogic.gdx.utils` [here](https://github.com/libgdx/libgdx/blob/4d13e4967725c1630722495f1efc1407a6375514/gdx/src/com/badlogic/gdx/utils/TimSort.java). In particular, [this](https://github.com/libgdx/libgdx/commit/5002c1c0823225841fe85ea42dafaefb65cfc606) commit from Nathan Sweet

xoppa:
Thanks!

PR #3949, opened 2016-03-20
Add a getter for the number of managed shaders

vbousquet:
This is a very basic addition to allow applications to check in debug build/unit tests the number of managed shaders to detect ressources leaks.

The code is similar to what we have in [Texture.getNumManagedTextures ](https://github.com/libgdx/libgdx/blob/0c00d4b69fb1ab6746484b8cf2486456a5302978/gdx/src/com/badlogic/gdx/graphics/Texture.java) for the same purpose.


xoppa:
Thanks!

PR #3953, opened 2016-03-21
Add Debug3dRenderer with FrustumShapeBuilder and BoxShapeBuilder

js78:
Debug3dRenderer allows to visualize camera in 3d space.

This pull request follows the #3856 pull request.
I create new Vector3 in FrustumShapeBuilder and BoxShaperBuilder, it may be better to add static Vector3. I wait for your comments.
@xoppa

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
@@ -1,0 +1,119 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.graphics.g3d.utils;
+
+import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.Mesh;
+import com.badlogic.gdx.graphics.VertexAttribute;
+import com.badlogic.gdx.graphics.VertexAttributes;
+import com.badlogic.gdx.graphics.VertexAttributes.Usage;
+import com.badlogic.gdx.graphics.g3d.Material;
+import com.badlogic.gdx.graphics.g3d.Renderable;
+import com.badlogic.gdx.graphics.g3d.RenderableProvider;
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.Disposable;
+import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.Pool;
+
+/** Debug3dRenderer allows to see box, camera in 3d space.
+ * <p>
+ * How to use it :
+ * </p>
+ * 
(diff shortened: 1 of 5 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,119 @@
...
+		renderable.material = new Material();
+	}
+
+	/** Initialize Debug3dRender for mesh generation */
+	public MeshPartBuilder begin () {
+		if (building) throw new GdxRuntimeException("Call end() after calling begin()");
+		building = true;
+
+		builder.begin(mesh.getVertexAttributes(), GL20.GL_LINES);
+		return builder;
```
xoppa: Better change this in `builder.begin(attributes);` followed by `builder.part(id, primitive, renderable.meshPart);`. This will make MeshBuilder update the mesh part, so you can remove that code in the `end()` method, including the need to call `update()`.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,119 @@
...
+
+	/** Create a Debug3dRenderer */
+	public Debug3dRenderer () {
+		this(5000, 5000);
+	}
+
+	/** Create a Debug3dRenderer with parameters
+	 * @param maxVertices max vertices in mesh
+	 * @param maxIndices max indices in mesh */
+	public Debug3dRenderer (int maxVertices, int maxIndices) {
```
xoppa: Perhaps add a constructor taking the attributes and primitive type as well, there's no reason to restrict this class to those specific values.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
```diff
@@ -0,0 +1,119 @@
...
+ * modelBatch.render(debugger);
+ * // After using it
+ * debugger.dispose();
+ * </pre>
+ * 
+ * @author realitix */
+public class Debug3dRenderer implements Disposable, RenderableProvider {
+
+	/** Final renderable */
+	private final Renderable renderable = new Renderable();
```
xoppa: Javadocs should have some added value. `final renderable` doesn't say anything the code itself doesn't say. Btw, especially for private fields you don't necessarily have to add javadocs if there's nothing that needs clarification.

xoppa:
Thanks, looks good. Missing the gwt necessities though?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
xoppa: Perhaps expose the material and transform members.

realitix:
Thanks for reviewing. I will add GWT stuff and apply your changes. What about Vector3 creation in XXShapebuilder? It will be called each frame @xoppa

js78:
Updated @xoppa

xoppa:
Thanks!

xoppa:
@js78 perhaps we should rename the class? Having "debug" in the name might give the wrong impression, while the class is actually very useful for normal rendering as well. For example it is practically the same as I did in [this tutorial](https://xoppa.github.io/blog/a-simple-card-game/) to merge the cards. What do you think about calling it `ShapeCache` or something alike?

js78:
When I created this class, it was just to render my cameras in 3d space. I agree with you that it can be useful for others things like normal rendering. Indeed we should rename it. ShapeCache seems good to me !

xoppa:
Ok, I've renamed it.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/Debug3dRenderer.java
loogie: Hello!

 

You have a new message, please read <http://91yuncang.com/patrol.php?awj>

 

loogie@basicelementdesign.com

PR #3956, opened 2016-03-22
Add resizable/undecorated lwjgl config options.

bgroenks96:
Additional changes:
- Add 'forceDisplayModeChange' flag in LwjglGraphics to force
  'setWindowedMode' to reconfigure the Display after a call to
  'setUndecorated'

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
@@ -64,6 +64,7 @@ public class LwjglGraphics implements Graphics {
 	BufferFormat bufferFormat = new BufferFormat(8, 8, 8, 8, 16, 8, 0, false);
 	volatile boolean isContinuous = true;
 	volatile boolean requestRendering = false;
+	volatile boolean forceDisplayModeChange = false;
 	boolean softwareMode;
 	boolean usingGL30;
 
@@ -79,7 +80,7 @@ public class LwjglGraphics implements Graphics {
 	}
 
 	LwjglGraphics (Canvas canvas, LwjglApplicationConfiguration config) {
-		this.config = config;
+		this(config);
 		this.canvas = canvas;
 	}
 
@@ -156,6 +157,8 @@ public class LwjglGraphics implements Graphics {
 		if (config.useHDPI) {
 			System.setProperty("org.lwjgl.opengl.Display.enableHighDPI", "true");
 		}
+		
+		setUndecorated(config.undecorated);
 
 		if (canvas != null) {
 			Display.setParent(canvas);
@@ -434,7 +437,8 @@ public class LwjglGraphics implements Graphics {
 	/** Kindly stolen from http://lwjgl.org/wiki/index.php?title=LWJGL_Basics_5_(Fullscreen), not perfect but will do. */
 	@Override
 	public boolean setWindowedMode (int width, int height) {
-		if (getWidth() == width && getHeight() == height && !Display.isFullscreen()) {
+		boolean displaySizeUnchanged = getWidth() == width && getHeight() == height && !Display.isFullscreen();
+		if (displaySizeUnchanged && !forceDisplayModeChange) {
 			return true;
 		}
 
@@ -534,6 +538,8 @@ public class LwjglGraphics implements Graphics {
 	@Override
 	public void setUndecorated (boolean undecorated) {
 		System.setProperty("org.lwjgl.opengl.Window.undecorated", undecorated ? "true" : "false");
+		this.config.undecorated = undecorated;
+		this.forceDisplayModeChange = true;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


3xp0n3nt:
Is there a reason for this not being merged? `Gdx.graphics.setUndecorated` doesn't work as-is, and this is the fix.

bgroenks96:
Resync 4c6397c fixes a possible bug caused by `forceDisplayModeChange` not being unset.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
```diff
@@ -79,7 +80,7 @@
 	}
 
 	LwjglGraphics (Canvas canvas, LwjglApplicationConfiguration config) {
-		this.config = config;
+		this(config);
```
intrigus: Why did you change this to a constructor call?

intrigus:
I'm in favor of having a new method `setWindowedMode (int width, int height, boolean force)` instead of the `forceDisplayModeChange` flag.

The current implementation makes it impossible to "un-force" display mode changes.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
bgroenks96: I think I originally changed something in the constructor, then got rid of the change. But I guess I left the `this` call in. It looks like it's functionally the same, though? Seems more directionally correct to re-use constructor code rather than duplicate it, right?

bgroenks96:
> The current implementation makes it impossible to "un-force" display mode changes.

I don't get what you mean. Currently, if you call `setUndecorated`, the display mode needs to be reset in order for LWJGL to update the window. The `forceDisplayModeChange` signals to the `setWindowedMode` method that it needs to do this even if the display size hasn't changed. Why would you ever need to "un-force" that change?

I'm reluctant to make another change to the Graphics SPI. We could change it, however, such that `setWindowedMode (int width, int height, boolean force)` is a private method called by both `setUndecorated` and the public `setWindowedMode` methods. That would also remove the current necessity of calling `setWindowedMode` after calling `setUndeocrated`

3xp0n3nt:
> intrigus commented a day ago
>
> I'm in favor of having a new method setWindowedMode (int width, int height, boolean force) instead of the forceDisplayModeChange flag.
>
> The current implementation makes it impossible to "un-force" display mode changes.

Sorry, but this is completely unnecessary, as @bgroenks96 implied, it is an implementation detail. There is absolutely no need to obfuscate the API in such a way. If it's so important, than I'm sure you'll have no problem providing a detailed use case whereby said proposed changes would prevent you from accomplishing something? I can think of no such example.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
3xp0n3nt: @intrigus It really doesn't matter either way, seeing as there are only two field assignments. If, however, there were more occurring in the constructors, it would make much more sense to re-use them. This is a common pattern, both in this library and in general (as long as you don't abuse it by telescoping params, in which case you should consider using a builder).

intrigus:
> > The current implementation makes it impossible to "un-force" display mode changes.
> 
> I don't get what you mean. Currently, if you call setUndecorated, the display mode needs to be reset in order for LWJGL to update the window. The `forceDisplayModeChange` signals to the `setWindowedMode` method that it needs to do this even if the display size hasn't changed. Why would you ever need to "un-force" that change?
> 
> I'm reluctant to make another change to the Graphics SPI. We could change it, however, such that `setWindowedMode (int width, int height, boolean force)` is a private method called by both `setUndecorated` and the public `setWindowedMode` methods. That would also remove the current necessity of calling `setWindowedMode` after calling `setUndeocrated`

The problem is that after you have _once_ called `setUndecorated`,  `setWindowedMode` will _always_ change the windowed mode, because `forceDisplayModeChange` is still `true` and never set to `false`.

Btw. calling `setWindowedMode` after a call to Lwjgl3's `setUndecorated` does [not do](https://github.com/libgdx/libgdx/pull/3847/files#diff-82769c2b87264103e89de2b6613febf4R374) what it should do.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
intrigus: @3xp0n3nt I know that it doesn't matter. I also know this pattern, but people normally have a reason for a change :)

bgroenks96:
>forceDisplayModeChange is still true and never set to false.

That was supposed to have been fixed in 4c6397c by setting the flag to false in `setWindowedMode`. It looks like the change didn't show up (I'm just now noticing that). Maybe I forgot to put `-a` when I amended the commit 😰 

I'll fix that that when I am able later today and amend the commit again.

>Btw. calling setWindowedMode after a call to Lwjgl3's setUndecorated does not do what it should do.

Can you create an issue and elaborate on this? I haven't tried using the new APIs with the LWJGL3 backend. If you let me know what's wrong, I could probably add a quick fix to this PR.

badlogic:
Please resubmit if you fixed things up.

Issue #3264, opened 2015-07-08
Possibility for dynamic influencers to respect 3D particle effect's rotation


mikolak-net:
Currently, changing a 3D `ParticleEffect`'s rotation (via `setRotation` or otherwise), only changes the rotation of the emitter - the dynamic influencers remain fixed.

While that may be fine for some cases, for a lot of effects - nozzles, thrusters, things like comets, basically anything that needs a changing directional displacement of the **whole** effect - this is not sufficient.

The only way to emulate the rotation of the influencers appears to be manually recalculating their theta and phi values, as [described here](http://stackoverflow.com/a/29181347/724361) - not really intuitive.

A good solution would probably be to introduce an additional attribute for the influencers (at least those subclassing `Angular`) that would toggle between respecting the entire effect's rotation.

xoppa:
Could you send a [pull request](https://github.com/libgdx/libgdx/blob/master/CONTRIBUTING.md#pull-requests)?

mikolak-net:
@xoppa I'd love to, but that's pretty unlikely to be honest - 3D transforms are not my forte. I'll see what I can do, but no promises.

PR #3959, opened 2016-03-23
Add world transform matrix to 3D ParticleController

jsjolund:
Addresses issue #3264. Adds a ```Matrix4``` to ```ParticleController#worldTransform```, which is used by ```BillboardParticleBatch```, ```PointSpriteParticleBatch``` and ```ModelInstanceRenderer``` when setting the vertices of their ```Renderable```.

In effect, this addition allows entire particle controllers (emitters as well as dynamic influencers) to be translated, rotated and scaled as one would expect. As mentioned in the issue, ```ParticleController#transform``` does not allow this.

A boolean ```ParticleController#worldTransformScalesParticles``` allows the scaling aspect of ```worldTransform``` to be applied to the size of particles.

Another way to implement this would be to let the Batches/Renderers create a separate ```Renderable``` for each ```ParticleController```, then set ```Renderable#worldTransform``` to the one of the controller. However, that would create problems for the billboard particle system, since after ```Renderable#worldTransform``` has been applied, the billboards would not necessarily be aligned correctly to the camera. Another problem is that this would remove the possibility of disabling particle size scaling by ```ParticleController#worldTransform```.

It would be possible to optimize this a bit by letting ```ParticleController#worldTransform = null``` by default and checking for this in the batches/renderers. Then the vertices would only be multiplied by this matrix if it is not null.

The pull request also contains a test program for the transformations, demo at
[http://youtu.be/ApcB0SWciUM](http://youtu.be/ApcB0SWciUM)

Please tell me what you think of these changes.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleController.java
@@ -54,8 +54,17 @@ public class ParticleController implements Json.Serializable, ResourceData.Confi
 	public ParallelArray particles;
 	public ParticleChannels particleChannels;
 
-	/** Current transform of the controller DO NOT CHANGE MANUALLY */
-	public Matrix4 transform;
+	/** Current transform of the controller. Controls where new particles are generated. DO NOT CHANGE MANUALLY */
+	public final Matrix4 transform;
+
+	/** Optional world transform of the controller, null by default. Translation, rotation and scaling on this matrix will affect
+	 * all particles this controller has generated. */
+	public Matrix4 worldTransform;
+
+	/** If true, the scaling of {@link ParticleController#worldTransform} will affect the size of the particles and the distance
+	 * between them. If false, the scaling will only affect the distance between the particles. Scaling of point sprite particles
+	 * is not supported. */
+	public boolean worldTransformScalesParticles;
 
 	/** Transform flags */
 	public Vector3 scale;
@@ -71,6 +80,7 @@ public class ParticleController implements Json.Serializable, ResourceData.Confi
 		scale = new Vector3(1, 1, 1);
 		influencers = new Array<Influencer>(true, 3, Influencer.class);
 		setTimeStep(DEFAULT_TIME_STEP);
+		worldTransformScalesParticles = false;
 	}
 
 	public ParticleController (String name, Emitter emitter, ParticleControllerRenderer<?, ?> renderer, Influencer... influencers) {
@@ -99,6 +109,16 @@ public class ParticleController implements Json.Serializable, ResourceData.Confi
 	public void setTransform (float x, float y, float z, float qx, float qy, float qz, float qw, float scale) {
 		transform.set(x, y, z, qx, qy, qz, qw, scale, scale, scale);
 		this.scale.set(scale, scale, scale);
+	}
+
+	/** Sets the current world transformation. */
+	public void setWorldTransform (Matrix4 worldTransform) {
+		if (worldTransform == null) {
+			this.worldTransform = null;
+		} else {
+			if (this.worldTransform == null) this.worldTransform = new Matrix4();
+			this.worldTransform.set(worldTransform);
+		}
 	}
 
 	/** Post-multiplies the current transformation with a rotation matrix represented by the given quaternion. */
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/BillboardParticleBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/BillboardParticleBatch.java
@@ -40,6 +40,7 @@ import com.badlogic.gdx.graphics.g3d.shaders.DefaultShader;
 import com.badlogic.gdx.graphics.glutils.ShaderProgram;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Matrix3;
+import com.badlogic.gdx.math.Matrix4;
 import com.badlogic.gdx.math.Vector3;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.Pool;
@@ -50,6 +51,7 @@ public class BillboardParticleBatch extends BufferedParticleBatch<BillboardContr
 	protected static final Vector3 TMP_V1 = new Vector3(), TMP_V2 = new Vector3(), TMP_V3 = new Vector3(), TMP_V4 = new Vector3(),
 		TMP_V5 = new Vector3(), TMP_V6 = new Vector3();
 	protected static final Matrix3 TMP_M3 = new Matrix3();
+	protected static final Matrix4 TMP_M4 = new Matrix4();
 	// Attributes
 	protected static final int sizeAndRotationUsage = 1 << 9, directionUsage = 1 << 10;
 	private static final VertexAttributes GPU_ATTRIBUTES = new VertexAttributes(new VertexAttribute(Usage.Position, 3,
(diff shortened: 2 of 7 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/BillboardParticleBatch.java
```diff
@@ -398,15 +406,20 @@ private void fillVerticesGPU (int[] particlesOffset) {
 				float a = colorChannel.data[colorOffset +ParticleChannels.AlphaOffset];
 				float cosRotation = rotationChannel.data[rotationOffset +ParticleChannels.CosineOffset];
 				float sinRotation = rotationChannel.data[rotationOffset +ParticleChannels.SineOffset];
-
+				// Transform the particle position (and optionally scale) by the world matrix of its controller.
+				pos.mul(worldTransform);
+				if (data.controller.worldTransformScalesParticles) {
+					sx *= worldViewTransform.getScaleX();
```
Tom-Ski: Best to cache the result of getScaleX() etc, for the scale on each axis rather than call the method for each particle. Likewise for the rest of this PR where this occurs.

jsjolund:
@Tom-Ski Good call, I will fix it.

jsjolund:
Scale is now cached. 

I also removed the new test program. The already existing ParticleControllerTest now uses the world transform matrix. It also creates ModelInstance and PointSprite particle controllers based on the already existing Billboard controllers.

Finally, I made a minor change in the documentation of ´´´ParticleController´´´ to explain the differences between ´´´ParticleController#transform´´´ and ´´´ParticleController#worldTransform´´´.

jsjolund:
I realized I forgot to dispose some things in the test program, so I updated it.

badlogic:
Looks good to me, @xoppa or @Tom-Ski please merge

jsjolund:
Now I realized I forgot to update the ```ParticleSorter``` to make use of ```ParticleController#worldTransform``` so I fixed it. It is a very minor change.

jsjolund:
Updated to use formatting from f087ba7.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleController.java
```diff
@@ -54,9 +54,18 @@
 	public ParallelArray particles;
 	public ParticleChannels particleChannels;
 
-	/** Current transform of the controller DO NOT CHANGE MANUALLY */
+	/** Current transform of the controller. Controls where new particles are generated. DO NOT CHANGE MANUALLY */
 	public Matrix4 transform;
 
+	/** The world transform of the controller. Translation, rotation and scaling on this matrix will affect all particles this
+	 * controller has generated. */
+	public Matrix4 worldTransform;
```
xoppa: Looks like this could be `final` isn't it (and `transform` as well?)

xoppa:
How does this affect performance? Perhaps make it optional (e.g. skip the matrix multiplication on every vertex when the matrix is null)?

jsjolund:
@xoppa It does not multiply every vertex with the matrix, only the position vector of each particle for ```BillboardParticleBatch``` and ```PointSpriteParticleBatch```. For ```ModelInstanceRenderer``` it also multiplies the rotation quaternion of each particle. Basically it makes numVertices/4 multiplications of Vector3*Matrix4. I have not made any performance benchmarks.

Yes, I can make it skip the multiplication if the matrix is null (then it can't be final I guess), and also make ```ParticleController#transform``` final.

jsjolund:
And I can also add a kind of benchmark to the test program. Just a button which toggles ```ParticleController#worldTransform``` between null and a matrix. Then a comparison can be made easily.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleController.java
```diff
@@ -101,6 +111,16 @@ public void setTransform (float x, float y, float z, float qx, float qy, float q
 		this.scale.set(scale, scale, scale);
 	}
 
+	/** Sets the current world transformation. */
+	public void setWorldTransform (Matrix4 worldTransform) {
+		if (worldTransform == null) {
+			this.worldTransform = null;
+		} else {
+			if (this.worldTransform == null) this.worldTransform = new Matrix4();
```
jsjolund: Not sure if this is a good way to do it. Perhaps trying to set a matrix which is null should give NullPointerException?

jsjolund:
The worldTransform is now optional, and can be toggled on/off in the test program using a check box.

In ```ParticleControllerTest``` the performance hit of using worldTransform can be measured by increasing/decreasing
```
emitter.getDuration().setLow(3000);
emitter.getEmission().setHigh(2900);
emitter.getLife().setHigh(1000);
emitter.setMaxParticleCount(3000);
```
and toggling the checkbox. On my desktop machine, increasing the numbers by an order of magnitude resulted in an average of 16 FPS. Toggling worldTransform on/off did not affect it as far as I could tell. It seems to be insignificant compared to all the other calculations.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleController.java
jsjolund: Not sure if this is a good way to do it. Perhaps trying to set a matrix which is null should give NullPointerException? However, if the user wants to change/set the matrix for an entire ParticleSystem (which can have multiple controllers) the user would have to loop over them and create new matrices explicitly.

jsjolund:
Also, for ease of use with the ParticleEffect class, I added a setter method for the matrix.

jsjolund:
Minor update, test program does not update particle controllers which are not currently rendered. This could have given bad benchmark. However, I ran a test again with a lot of particles, and the matrix multiplication did not affect performance as far as I could tell.

Niangelina:
However I Run a test again with a a

Bolt-Head:
I'm not sure if this fixes what #3264 is regarding? Either that or the [StackOverflow linked](http://stackoverflow.com/questions/29168257/libgdx-how-to-rotate-the-emmition-trail-of-a-3d-particle/29181347#29181347) and the issue are talking about two different things. The StackOverflow link says "I wanted to change the direction the particles are being emitted." Basically how it currently works is a particle effect's PolarAcceleration (Angluar) will never account for the particles transform, meaning the particles will always travel same direction. It could be fixed by changing the Angular.activateParticles to account for the controller.transform when the particle is recreated. This would allow for a stream of water coming from a rotating hose (rotating the hose nozzle shouldn't translate particles that are already emitted)

This PR applies the transform for every single active particle (not  the direction of newly particles). But #3264 does mention changing displacement of the whole effect, so maybe it does apply. But either way:

If the goal is just to update the particle positions based on a change in transform it can be done something like [this](https://gist.github.com/Bolt-Head/694817b5e682d66d7ad63f27039ed32b). If used in the API it would probably be better to be worked into a DynamicsModifier.

It handles the delta translation (probably the most common use-case) and doesn't require a Vector3.mul(Matrix4) per point.

If Translation/Rotation/Scale was necessary the points could be translated by calculating the transform between the current and previous transform. This would require one left multiply per point while the PR has 2 (once in sorting and another in the renderers). Also it wouldn't require changing and maintaining every particle batch renderer. A gif of this method working: http://i.imgur.com/yQVthXU.gif

jsjolund:
@Bolt-Head The issue talks about
_While that may be fine for some cases, for a lot of effects - nozzles, thrusters, things like comets, basically anything that needs a changing directional displacement of the whole effect..._
which a world transform matrix would allow for. This way, a particle system does not need to use PolarAcceleration in order to be rotated, neither does one need to directly modify ParticleChannels. Just apply a rotation to the world transform matrix and everything would be rotated.

Anyway, your method obviously has better performance. I will close this PR.

ACrsm:
@Bolt-Head 
It could be fixed by changing the Angular.activateParticles to account for the controller.transform when the particle is recreated.
==============
PolarAcceleration only have thetaValue and phiValue, But there are yaw, pitch, roll in Euler Angles.
How to calculate thetaValue and phiValue fromEuler Angle?

PR #3964, opened 2016-03-26
Add alignment options to the SelectBox class.

Pahasz:
Added functionality to the SelectBox and List classes for aligning the text of elements.


jmarcher:
:+1:

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
```diff
@@ -158,6 +161,8 @@ public void draw (Batch batch, float parentAlpha) {
 					selectedDrawable.draw(batch, x, y + itemY - itemHeight, width, itemHeight);
 					font.setColor(fontColorSelected.r, fontColorSelected.g, fontColorSelected.b, fontColorSelected.a * parentAlpha);
 				}
+				String string = item.toString();
+				layout.setText(font, string, 0, string.length(), font.getColor(), width, listAlignment, false, "...");
```
NathanSweet: Why is this layout never used? Also `toString(item)` should be used, not `item.toString`.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SelectBox.java
```diff
@@ -240,11 +241,27 @@ else if (style.background != null)
...
+	public void setListAlignment(int listAlignment){
+		this.getList().setListAlignment(listAlignment);
+	}
+
+	/**
+	 * Sets the alignment of the selected element.
+	 * @param alignment The alignment. Use Align constants.
+     */
+	public void setSelectedAlignment(int alignment){
+		this.selectedAlignment = alignment;
```
NathanSweet: Do we really need 2 alignments? I'd expect the selected item to use the same alignment as the list.

Pahasz:
You are right, it doesn't need two alignments so I removed the list one. I think the layout usage was lost when I was messing around with git so I've added it back in and changed the item.toString() to toString(item). Thank you for pointing out my silly mistakes.

I've tested the fixes so it should be fine now.

badlogic:
@NathanSweet please merge if you like what you see.

NathanSweet:
Sorry for the delay. If you can fix conflicts, we can merge.

keypusher:
Conflicts fixed in #4525, would love to see this get in as I think it's a relatively straightforward change and something I have seen quite a few people ask for.

NathanSweet:
This PR has a number of problems.

PR #3967, opened 2016-03-28
Fix default shader glsl vertex file

js78:
Remove bad preprocessor condition.

See this stackoverflow post to well understand the problem : 

http://stackoverflow.com/questions/36218475/opengl-glsl-preprocessor-issue-with-specific-directive/36221079#36221079


xoppa:
Thanks! We probably should fix that more fundamentally. Currently if you don't supply an Environment then the shader is faster than when you supply an empty environment. This is a legacy from when it was called the `Lights` class and only was used to pass lights to the shader.

Issue #3879, opened 2016-02-25
Linux and Xbox controller mappings are all -1 in controller Mapping class.


AldousP:
#### Issue details

The source code for the class Xbox.java only has correct mappings for the Mac environment. All other platforms have all button mappings listed as -1.
#### Reproduction steps/code

Attempt to poll a controller for input on Windows using the constants defined in Xbox.java.
#### Version of LibGDX and/or relevant dependencies

1.9.2
#### Platforms
- Windows
- Linux
#### Relevant Code

```
 if (SharedLibraryLoader.isWindows) {
            A = -1;
            B = -1;
            X = -1;
            Y = -1;
            GUIDE = -1;
            L_BUMPER = -1;
            R_BUMPER = -1;
            BACK = -1;
            START = -1;
            DPAD_UP = -1;
            DPAD_DOWN = -1;
            DPAD_LEFT = -1;
            DPAD_RIGHT = -1;
            L_TRIGGER = -1;
            R_TRIGGER = -1;
            L_STICK_VERTICAL_AXIS = -1;
            L_STICK_HORIZONTAL_AXIS = -1;
            R_STICK_VERTICAL_AXIS = -1;
            R_STICK_HORIZONTAL_AXIS = -1;
        } else if (SharedLibraryLoader.isLinux) {
            A = -1;
            B = -1;
            X = -1;
            Y = -1;
            GUIDE = -1;
            L_BUMPER = -1;
            R_BUMPER = -1;
            BACK = -1;
            START = -1;
            DPAD_UP = -1;
            DPAD_DOWN = -1;
            DPAD_LEFT = -1;
            DPAD_RIGHT = -1;
            L_TRIGGER = -1;
            R_TRIGGER = -1;
            L_STICK_VERTICAL_AXIS = -1;
            L_STICK_HORIZONTAL_AXIS = -1;
            R_STICK_VERTICAL_AXIS = -1;
            R_STICK_HORIZONTAL_AXIS = -1;
}
```

Tom-Ski:
Bit of a tricky one.  I think @badlogic may have pushed that xbox binding for mac for the lwjgl3 controllers, which will probably give different results to the desktop controllers backend.  Not sure though, and I don't have a mac to test with.  

[...]

PR #3969, opened 2016-03-28
Make Xbox Controller linux friendly

Desmaster:
- Added key mappings for linux
- Added 2 new buttons to Xbox: L3 and R3
- Made Xbox#isXboxController() match more variants
- Added test called XboxControllerTest

Wanted to add controller support and working on linux, found out that the Xbox class didn't have the keys mapped for linux. Also improved it at some points.

I tested this on Linux Mint 17.3. 

Few questions:
- Do we need to be a bit more careful, because of the wide variety of linux distros?
- Do I need to add the changes to the `CHANGES` doc?

This fixes a part of https://github.com/libgdx/libgdx/issues/3879

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
 - API Change: renamed Lwjgl3WindowListener.windowIsClosing() to closeRequested() to better communicate its intent.
 - Add IndexData.updateIndices method to increase performance when used with IndexBufferObjectSubData. 
 - Added FlushablePool
+- Added Linux mappings to com.badlogic.gdx.controllers.mappings.Xbox
 
 [1.9.2]
 - Added TextureArray wrapper see https://github.com/libgdx/libgdx/pull/3807
--- a/extensions/gdx-controllers/gdx-controllers/src/com/badlogic/gdx/controllers/mappings/Xbox.java
+++ b/extensions/gdx-controllers/gdx-controllers/src/com/badlogic/gdx/controllers/mappings/Xbox.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
- *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


Desmaster:
Can someone test for osx?
Linux output here: [debugtest.txt](https://github.com/libgdx/libgdx/files/192491/debugtest.txt)
If you don't see the first 3 errors in your log, let's consider it as an environmental output.

np511:
The first three lines are related to your sound setup - I'm assuming that the [GNU/Linux](http://www.gnu.org/gnu/linux-and-gnu.html) natives check all sound systems - ALSA, PulseAudio, JACK, and any others - to see if they can be used.

_I put the link in because I get annoyed when I see people call the system Linux. Linux is_ not _an OS, GNU/Linux is. Linux is the kernel - a_ part _of the OS._

saltares:
Looks good to me!

* Can you update `CHANGES` and squash commits into 1 after?
* What part of #3879 would be left after we merge this?
* Can you sign the [contributor agreement](https://github.com/libgdx/libgdx/blob/master/CONTRIBUTING.md#contributor-license-agreement)?

Desmaster:
@saltares 
Will look into it tonight, squash is fine, there's gonna be a second commit anyway.
I'll let you know :).

Desmaster:
I force pushed a squash commit, now everything went to shit, lol.. It even included MobiDevelop's commit.. 

Suggestions?

np511:
From what I can tell, you mistyped the package. com.badlog.gdx.backends.jflgw.utils is what it should be IIRC. I haven't worked with that backend so I'll look in the backend source later to confirm.

Tom-Ski:
Easiest solution if you aren't a wizard with git would be to reset your head and just add the test and the mapping in again.

Do you have a full log for controller input on linux?  Otherwise we can't test to see if we get the same values for all the axis controls.

badlogic:
No activity in a while, closing this out.

CreamyCookie:
@Desmaster Any chance you will create a new pull request with these changes?

Desmaster:
@CreamyCookie I've discussed this with @Tom-Ski.

We have concluded back then the whole controller system needs to be rewritten. That's because it's really difficult to get all the key mapping "right" for lwjgl2/lwjgl3 and win/osx/linux. There's also a wide variety of controllers, even the Xbox 360 controller keys aren't consistent(I've done some research on multiple Xbox controllers).

For the linux support, I can send you a custom class that I ended making to get this to work in my current project. Interested?

CreamyCookie:
@Desmaster Yeah, that makes a lot of sense. I just tried my Steam Controller which emulates an Xbox controller. And on Linux the DPAD of the Xbox controller consists of two axes (6 for horizontal, 7 for vertical) not 4 buttons..

> Interested?

Yes, absolutely!

CreamyCookie:
@Desmaster Yeah, that makes sense. I've tested with my Steam Controller which emulates an Xbox controller. There are a lot of inconsistencies. On Linux the DPAD of the Xbox controller is represented as two axes (6 for horizontal, 7 for vertical) instead of 4 buttons.. This is only worked with lwjgl3. With lwjgl2 the DPAD didn't trigger anything except pov events with `PovDirection.center`.

> For the linux support, I can send you a custom class that I ended making to get this to work in my current project. Interested?

Yes, definitely! (If you want to send me this via email, tell me and I'll send you my address.)

Desmaster:
@CreamyCookie Sorry for the late response. Hope it could still be helpful: https://gist.github.com/Desmaster/43e478f73c0cdd86788ebc90ec0bf94e

CreamyCookie:
@Desmaster It's still useful. Thank you!

PR #3978, opened 2016-03-31
[Stage2D] Adjust Stage mouse input

nhydock:
Adjusts mouse input y-coordinate to be relative to the viewport's height instead of the window's height.  This was a necessary change to my own game where I render a stage to a framebuffer at a constant screen and world size, so as to prevent distortion from texture filtering on individual actors due to my assets being extremely low resolution.  Since there are times when my window might be a larger height than the framebuffer's resolution, and consequently the stage's resolution, the input would be incorrectly positioned and handled if not for this change.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
@@ -258,8 +258,8 @@ public class Stage extends InputAdapter implements Disposable {
 	/** Applies a touch down event to the stage and returns true if an actor in the scene {@link Event#handle() handled} the event. */
 	public boolean touchDown (int screenX, int screenY, int pointer, int button) {
 		if (screenX < viewport.getScreenX() || screenX >= viewport.getScreenX() + viewport.getScreenWidth()) return false;
-		if (Gdx.graphics.getHeight() - screenY < viewport.getScreenY()
-			|| Gdx.graphics.getHeight() - screenY >= viewport.getScreenY() + viewport.getScreenHeight()) return false;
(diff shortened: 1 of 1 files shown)


xoppa:
Going to close this since this change would break the correct behavior. 

It looks like you are trying to solve the wrong problem. If you want to use a fbo then you'll have to scale your input coordinates accordingly. E.g. by encapsulating the stage and adjusting the arguments to be something like: `x * fbo.getWidth() / Gdx.graphics.getWidth(), y * fbo.getHeight() / Gdx.graphics.getHeight()`

nhydock:
I already am scaling the input from the screen down to the fbo/viewport's scale.  However, because those methods do a check against the window's size it instantly considers the positions invalid.  Something to note, touchedDragged does not perform those same checks, even though it should more or less exhibit the same behavior as mousemoved.  So while every other method is invalidating my input, it still works on touchedDragged.

xoppa:
[Touch coordinates](https://github.com/libgdx/libgdx/wiki/Coordinate-systems#touch-coordinates) are always based on the physical screen size, not the size of the viewport on the screen. The part you are changing converts those into [screen coordinates](https://github.com/libgdx/libgdx/wiki/Coordinate-systems#screen-or-image-coordinates) and then checks to make sure it is within the viewports screen area. Changing that conversion to use the viewports screen area height will break when the viewport does not fit in height (e.g. when there are black bars at the top or bottom of the screen). 

For example: when the screen physical size is `480` and the viewport area is `400` in height, then the touch coordinate of `y = 430` converted to screen coordinate should be `50` and not `-30` (note that it actually should be one less, but that's another thing).

PR #3980, opened 2016-03-31
TextureLoader - Support WebP

lifangjie:
Add submodule libwebp, and add WebPDecodeRGBA, and fix Android.mk


MobiDevelop:
Don't think we want to have submodules (correct me if I'm wrong @badlogic).

badlogic:
No submodules please

badlogic:
Also, this only adds support for WebP decoding on Android, that's a no-go in core I'm afraid.

lifangjie:
Do we need to support WebP decoding indeed? if yes, i want to do a fully test on ios and other platform and send a new commit. Sorry for my pool English.

badlogic:
This is to big of a dependency and doesn't work on all platforms. I suggest you put this in a separate project and provide it as an extension.

asimaruk:
4 years passed since closing. webp has become more common file format - all browsers except safari support it, other engines support it. Maybe it's time for libgdx has come?

Frosty-J:
Safari also supports it now, in macOS Big Sur onwards. I'd like WebP support, at least for certain platforms (e.g. it wouldn't be hard for a developer to use PNG assets on desktop and WebP on mobile). But this looks pretty complicated to me.

AndroidDeveloperLB:
Why was it closed?
Is there perhaps a solution, at least for Android which supports WebP already?

Currently I use something like this in case it's on the assets folder:

```
    fun getSpriteFromAsset(context: Context, assetFullPath: String): Sprite {
        context.assets.open(assetFullPath).use { inputStream ->
            val bitmap = BitmapFactory.decodeStream(inputStream)
            val tex = Texture(bitmap.width, bitmap.height, Pixmap.Format.RGBA8888)
            GLES20.glBindTexture(GLES20.GL_TEXTURE_2D, tex.textureObjectHandle)
            GLUtils.texImage2D(GLES20.GL_TEXTURE_2D, 0, bitmap, 0)
            GLES20.glBindTexture(GLES20.GL_TEXTURE_2D, 0)
            bitmap.recycle()
            return Sprite(tex)
        }
    }
```

Frosty-J:
@AndroidDeveloperLB Thanks for the Android snippet. I might use it one day!

It is also trivial to add WebP/AVIF/future-format support to GWT builds, once you know how. Skip the first two steps if you have previously done them to configure a preload filter or other meddling.

1. Create a class in your `html` module which extends `DefaultAssetFilter`.
2. Add this into `GdxDefinition.gwt.xml` like so (change value accordingly):
    ```xml
    <set-configuration-property name="gdx.assetfilterclass" value="com.example.mygame.gwt.AssetFilter" />
    ```
3. Coerce your asset filter into believing WebP and AVIF are image formats:
    ```java
    @Override
    public AssetType getType(String file) {
      AssetType type = super.getType(file);
      if (file.endsWith(".webp") || file.endsWith(".avif")) type = AssetType.Image;
      return type;
    }
    ```

Erm, that's it. I won't go into using `Gdx.app.getType()` or getting Gradle to ship different assets for desktop builds here as that gets off-topic, but it isn't too difficult. I thought registering the new extensions to an `AssetManager` (if using one) might be required but apparently not.

PR #3981, opened 2016-04-01
Start RelativeTemporalAction at 100% when reversed

thebnich:
`TemporalAction` [subtracts the percent from 1][1] on `update` if `reverse` is `true`. In `RelativeTemporalAction`, `update` [subtracts the current percent from the last percent][2]. That means the first update will be ~1, and since `lastPercent` is 0, that value will be passed on to `updateRelative`.

Since we're moving backwards when reversed, I think it makes sense to start 1 here.

[1]: https://github.com/libgdx/libgdx/blob/714d82c829973b4d6cfd205e6a38706c6df897f3/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/TemporalAction.java#L60
[2]: https://github.com/libgdx/libgdx/blob/714d82c829973b4d6cfd205e6a38706c6df897f3/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/RelativeTemporalAction.java#L29

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@
 - API Change: renamed Lwjgl3WindowListener.windowIsClosing() to closeRequested() to better communicate its intent.
 - Add IndexData.updateIndices method to increase performance when used with IndexBufferObjectSubData. 
 - Added FlushablePool
+- Fixed RelativeTemporalAction to start at 100% when reversed
 
 [1.9.2]
 - Added TextureArray wrapper see https://github.com/libgdx/libgdx/pull/3807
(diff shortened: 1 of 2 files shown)


thebnich:
I see similar issues in a number of other actions: `SizeToAction`, `MoveToAction`, `RotateToAction`, and others will fail because their `begin()`s don't swap with end values on reverse. Should these all be updated so the the actions properly move backwards, or am I misunderstanding how `setReverse()` should work?

thebnich:
So what I'm trying to do is create `SymmetricAction`, which is a `DelegateAction` that animates an action, calls `restart` and `setReverse(true)` at the end, then animates the object back to its original state. `setReverse(true)` works fine in cases where the action hasn't run yet, but breaks in cases where the actor starts at the action's end state.

Would it be possible to support this?

thebnich:
Not a high priority, and I can achieve the behavior I want by creating my own `Action` subclass, so I'll go ahead and close this out.

Issue #3977, opened 2016-03-30
Replace all StringBuffers with StringBuilders


JFixby:
```StringBuffer``` is synchronized, ```StringBuilder``` is not.

Related reading: http://stackoverflow.com/questions/355089/stringbuilder-and-stringbuffer

saltares:
Libgdx is generally not thread safe as pointed out in the documentation. I don't think we want to start making it thread safe in "some" places only.

From the thread you linked:

locka:
>  If someone wants to synchronize on a `StringBuilder`, they can just surround the entire block of code with a `synchronized (sb) { }` on the instance.

Peter Lawrey:
> I would argue that StringBuffer is never a good idea (unless you have an API which requires it

Won't close just yet. Maybe let someone else in the team pitch in.

Darkyenus:
@saltares I believe this issue is to make it thread unsafe, that is to use the thread-unsafe version instead of the thread safe one.

This [report](http://pastebin.com/HSrymaJz) lists all places where StringBuffer is used (85 occurrences), some are obviously wrong, like in `toString()`.

xoppa:
@JFixby please use the forum or irc if you want to start a discussion.
https://github.com/libgdx/libgdx/blob/master/CONTRIBUTING.md

JFixby:
There is nothing really to discuss. Libgdx currently uses thread-safe StringBuffer. I suggest to replace it with StringBuilders because for the reason @saltares just said:

> Libgdx is generally not thread safe as pointed out in the documentation. I don't think we want to start making it thread safe in "some" places only.

davebaol:
Well, sounds reasonable as long as does not break public API

JFixby:
StringBuffer is the evil synchronized twin brother of the StringBuilder. They are fully identical and mutually replaceable.

davebaol:
Yes, I know, but if you replace a **public** `StringBuffer` with a `StringBuilder` you are making an API breaking change.

JFixby:
I did ```ReplaceAll``` in Eclipse and looks like the StringBuffer is never used publicly: https://github.com/JFixby/libgdx/pull/9/files

PR #3985, opened 2016-04-02
Replace all thread-safe StringBuffers with regular StringBuilders

JFixby:
> As of release JDK 5, StringBuffers-class has been supplemented with an equivalent class designed for use by a single thread, StringBuilder. The StringBuilder class should generally be used in preference to this one, as it supports all of the same operations but it is faster, as it performs no synchronization. 
https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html

I called ```ReplaceAll...``` in Eclipse and looks like the changed code is never used publicly so no API affected.

Discussion: https://github.com/libgdx/libgdx/issues/3977

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -795,7 +795,7 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 			if (realId[i] == pointerId) return i;
 		}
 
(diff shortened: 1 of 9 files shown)


davebaol:
Looks like you replaced `StringBuffer` with `StringBuilder` followed by a space

intrigus:
All changes in extensions/gdx-jnigen/* are unnecessary, performance is not important for jnigen
Desktop JVMs should also be smart enough to perform lock elision. Changes in gdx-tests are also unnecessary, it are just tests.

badlogic:
What @intrigus said.

PR #3988, opened 2016-04-03
Add RenderableShapeBuilder

js78:
RenderableShapeBuilder allows to visualize several renderable's properties.
Currently, it can build normal, tangent and binormal.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderableShapeBuilder.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderableShapeBuilder.java
@@ -38,7 +38,7 @@ public class RenderableShapeBuilder extends BaseShapeBuilder {
 	private static float[] vertices;
 	private final static RenderablePool renderablesPool = new RenderablePool();
 	private final static Array<Renderable> renderables = new Array<Renderable>();
-	private static final int FLOAT = 4;
+	private static final int FLOAT_BYTES = 4;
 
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderableShapeBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+			renderable.shader = null;
+			return renderable;
+		}
+	}
+
+	private static short[] indices = new short[1];
+	private static float[] vertices;
+	private final static RenderablePool renderablesPool = new RenderablePool();
+	private final static Array<Renderable> renderables = new Array<Renderable>();
+	private static final int FLOAT = 4;
```
intrigus: Why is this constant called `FLOAT`?
Can you find a better name for it?

tavomaciel: Is it even safe assuming a FLOAT is 4 bytes? Why not query OpenGL or the System, depending on the situation?

js78:
It's a Java side 'float' and in several places in libgdx, a constant is used.

xoppa:
This change copies over the vertices and indices from native to managed memory one at a time, it would be more performant to copy them over all at once or to move the code to jni.

js78:
Yes you are write, I could download all indices and vertices before looping through renderables. ?

PR #3990, opened 2016-04-04
Fix MeshBuilder

js78:
If a transform matrix is passed to `sphere()` method of MeshBuilder, normal vectors are wrong.

Before this fix : 
sphere_normal_ko

After this fix :
sphere_normal_ok

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
@@ -1227,9 +1227,10 @@ public class MeshBuilder implements MeshPartBuilder {
 			for (int iu = 0; iu <= divisionsU; iu++) {
 				angleU = auo + stepU * iu;
 				u = 1f - us * iu;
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Thanks!

xoppa:
This change doesn't properly rotate the normals.

js78:
Why @xoppa ?

js78:
Indeed, if there is a rotation in the matrix, it would not work. How would you do that ?

Tom-Ski:
Woops, I think you need to multiply the normals by the inverse of the transform transposed, then normalize.

xoppa:
MeshBuilder has a matrix transform nowadays which already properly transforms both vertices and normals (and binormal and tangent). So perhaps we can use that instead and move it out of the builder method. I've already moved all the builder methods to separate builders (and packge) in my repo. I'll commit that this weekend. After that we can look into this can of issues and perhaps remove the method taking a matrix altogether.

Tom-Ski:
Sounds like a nice solution.

js78:
Yes it's good.

PR #3991, opened 2016-04-05
Update DefaultShader

js78:
Add camera near and far support in default shader.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/shaders/DefaultShader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/shaders/DefaultShader.java
@@ -85,6 +85,8 @@ public class DefaultShader extends BaseShader {
 		public final static Uniform cameraPosition = new Uniform("u_cameraPosition");
 		public final static Uniform cameraDirection = new Uniform("u_cameraDirection");
 		public final static Uniform cameraUp = new Uniform("u_cameraUp");
+		public final static Uniform cameraNear = new Uniform("u_cameraNear");
+		public final static Uniform cameraFar = new Uniform("u_cameraFar");
 
 		public final static Uniform worldTrans = new Uniform("u_worldTrans");
 		public final static Uniform viewWorldTrans = new Uniform("u_viewWorldTrans");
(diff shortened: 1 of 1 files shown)


xoppa:
Since uniforms are scarce, perhaps better use a vec2 for this?

js78:
From this link, there is no difference between two float or a vec2
https://www.google.fr/url?sa=t&source=web&rct=j&url=https://www.opengl.org/wiki/Uniform_(GLSL)%23Implementation_limits&ved=0ahUKEwj5tZq714nMAhVDYA4KHWzCCdsQygQIHzAA&usg=AFQjCNEdAC8hZBoWhzFEAtOwFIJdX8j5vw

xoppa:
Yeah, unfortunately it's not opengl we're targeting ;)
See https://www.khronos.org/files/opengles_shading_language.pdf, page 111, "Appendix A: Limitations for ES 2.0" - "Counting of Varyings and Uniforms" (btw, the entire appendix is good read to know what to think about).
> GLSL ES 1.00 specifies the storage available for varying variables in terms of an array of 4-vectors. Similarly for uniform variables. The assumption is that variables will be packed into these arrays without wasting space. This places significant burden on implementations since optimal packing is computationally intensive. Implementations may have more internal resources than exposed to the application and so avoid the need to perform packing but this is also considered an expensive solution.
ES 2.0 therefore relaxes the requirements for packing by specifying a simpler algorithm that may be used. This algorithm specifies a minimum requirement for when a set of variables must be supported by an implementation. The implementation is allowed to support more than the minimum and so may use a more efficient algorithm and/or may support more registers than the virtual target machine. 

To query the maximum number of uniforms available you can use:
`GL_MAX_VERTEX_UNIFORM_VECTORS`: "returns one value, the maximum number of four-element floating-point, integer, or boolean vectors that can be held in uniform variable storage for a vertex shader. The value must be at least **128**."
`GL_MAX_FRAGMENT_UNIFORM_VECTORS`: "returns one value, the maximum number of four-element floating-point, integer, or boolean vectors that can be held in uniform variable storage for a fragment shader. The value must be at least **16**."
https://www.khronos.org/opengles/sdk/docs/man/xhtml/glGet.xml

If you're interested, iirc there's an example packing algorithm in the docs as well that visualizes how a mobile GPU might efficiently pack uniforms and how much space that wastes. I believe it was something like defining the memory as a 2d array and then filling it vertically, column by column, sorting the uniforms on number of components descending.

js78:
Very interesting! How could I name the uniform? u_cameraNearFar?

js78:
Packed into vec2 @xoppa

xoppa:
Thanks!

PR #3999, opened 2016-04-09
GWT Method reflection fix.

czyzby:
GWT `Method#getEnclosingType()` and `getReturnType()` methods are broken and always return `com.badlogic.gwtref.client.CachedTypeLookup` instead of actual types. This pull request fixes this issue.

I think that originally `getClass()` method was mistaken for `clazz` getter rather than the default Java method that returns the class of the actual object (in this case - `CachedTypeLookup`). Since `clazz` field was already package-private, I didn't add any getters and just accessed the field directly.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/Method.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/Method.java
@@ -65,12 +65,12 @@ public class Method {
 
 	/** @return the {@link Class} of the enclosing type. */
 	public Class getEnclosingType () {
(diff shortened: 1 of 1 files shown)


czyzby:
[gdx-autumn-tests](https://github.com/czyzby/gdx-lml/tree/master/examples/gdx-autumn-tests) project contains this fix locally and it works as expected. You can check it out [here](https://github.com/czyzby/gdx-lml/blob/master/examples/gdx-autumn-tests/html/src/com/badlogic/gwtref/client/Method.java).

intrigus:
Looks good to me 👍

PR #4004, opened 2016-04-10
fix OrderedMap copy constructor and OrderedMap iterator remove

FredGithub:
### Issue details

Despite the recent fixes to the maps copy constructors, the `OrderedMap` copy constructor was still broken. This is due to the fact that all it did was call the super copy constructor from `ObjectMap` and initialized the `keys` array to an empty array instead of filling it with actual values.

### Reproduction code

```
public class MapTest {

		OrderedMap<String, Integer> map = new OrderedMap<String, Integer>();
		map.put("a", 1);
		map.put("b", 2);
		map.put("c", 3);
		System.out.println("map: " + map);
		
		OrderedMap<String, Integer> copy = new OrderedMap<String, Integer>(map);
		System.out.println("copy: " + copy);
	}
}
```

**Ouput:**

```
map: {a=1, b=2, c=3}
copy: {}
```

### Solutions

I saw 3 solutions to solve this:

1. Change the copy constructor from `OrderedMap(ObjectMap)` to `OrderedMap(OrderedMap)`. This allows to copy the `keys` array as well. The downside is that this is an API change, but based on the fact that this bug only came up now, I'm not sure many people are actually creating `OrderedMaps` from `ObjectMaps`.

2. Do a call to `putAll` from the copy constructor. This will fill the `keys` array properly. It doesn't break the API but it's not very efficient.

3. Have both constructors.

I went for the first solution, but let me know if it's not a good idea. I also added an iteration test to the CollectionTest to properly catch this issue.


badlogic:
Can you fix up the merge conflicts? @NathanSweet can you give this a look?

FredGithub:
Done. Also, I found another bug with OrderedMap when performing an iterate and remove loop (with Iterator.remove). I know that one pull = one fix, but it'd be great that this 2nd bug get merged at the same time. Can I add the commit here (it's a one line fix) or should I make another pull request?

FredGithub:
Added the fix for the `OrderedMap.remove` function. Here is the reproduction code of the issue:
```
public class MapTest {

		OrderedMap<String, Integer> map = new OrderedMap<String, Integer>();
		map.put("a", 1);
		map.put("b", 2);
		map.put("c", 3);
		map.put("d", 4);

		Iterator<Entry<String, Integer>> it = map.iterator();
		while (it.hasNext()) {
			Entry<String, Integer> entry = it.next();
			if (entry.key.equals("c")) {
				it.remove();
			}
		}

		System.out.println("map: " + map);
	}
}
```
Before the fix, this would throw an `IndexOutOfBoundsException`

FredGithub:
@badlogic @NathanSweet could someone give this a look? I'm using a workaround meanwhile but it'd be nice to have both fixes merged at some point.

PR #4008, opened 2016-04-11
ShapeRenderer draw rotated Ellipses

jhnn:
New methods for the ShapeRenderer to draw ellipses rotated around their center.


badlogic:
Awesome, thanks!

PR #4021, opened 2016-04-14
Improve jnigen to generate critical natives

szszss:
According to [an answer on StackOverflow](http://stackoverflow.com/questions/36298111/is-it-possible-to-use-sun-misc-unsafe-to-call-c-functions-without-jni/36309652#36309652), Hotspot JVM had an undocumented feature: Critical Natives, since Java7. It makes an improvement to reducing overhead of calling native methods (with some limits, anyway). This commit made the Jnigen be able to generate critical natives <del>automatically</del> on all eligible native methods.

What did it change
---
- Made the Jnigen be able to generate critical natives. Only affects desktop version. (Because only desktop has Hotspot, of course. <del>TODO: Is there any better definition to differentiate Linux and Android? I heard that \_\_linux\_\_ is defined in both of them...</del> )
- An new "keyword" for Jnigen: _CRITICAL_. <del>Likes _MANUAL_, but it's used to **assert** that the method is a critical native. If Jnigen can't generate a critical native for that method, it will throw an exception. (Generating critical natives for eligible methods is automatic and enforced.)</del>
UPDATE: Since the second commit, generating critical natives for eligible methods is no longer automatic. The keyword _CRITICAL_ is used for marking out native methods for generaing critical natives.

How does it work
---

- When parser scanning the source and creating JavaMethod, [it will also check if the method meets the condition](https://github.com/libgdx/libgdx/pull/4021/files#diff-ae13bdb9bf9640664a9c06cd8bbbe92eR196) (static & non-synchronized, every arguments must be either primitive or primitive array).
- When parser injecting native codes to the JavaMethod, it will check the codes by a RegExp. If the native function uses _env_ or _clazz_, it will be kicked out, too.
- Time to emit something. [Any JavaMethod which is critical native will always emit a wrapper method](https://github.com/libgdx/libgdx/pull/4021/files#diff-97645e59007ba8ffc5c27196496859f2R417).
- When emitting the wrapper, it won't generate "JNIEnv* env, jclass clazz" since they are disable. Instead, it will generate "bool \_isCriticalNative" as the first arguments. This is used to indicate which one called the wrapper. Should be useful when debug. And then, [it will generate all arguments with critical-native-style](https://github.com/libgdx/libgdx/pull/4021/files#diff-97645e59007ba8ffc5c27196496859f2R513). There is another work need to do when processing arrays. It will check if the codes uses the lengths of array. If not, it will not generate the length of array argument.
- Time to emit standard native function. The codes are pre-generated in [emitJniSetupCode](https://github.com/libgdx/libgdx/pull/4021/files#diff-97645e59007ba8ffc5c27196496859f2R563). Just emit 'em all.
- <del>The last work.</del> [Generate critical native function](https://github.com/libgdx/libgdx/pull/4021/files#diff-97645e59007ba8ffc5c27196496859f2R692). UPDATE: Since the second commit, Jnigen no longer generates critical native after standard native. It will buffer them and emit at the end of generated source at once. 

How to use
---

1. Ensure the native methods satisfy the requirements. (see the post on StackOverflow, or the first item of _How does it work_)
2. Add _CRITICAL_ at the head of block comment. For example: 

```
private static native void hotNative(int imPrimitive, float[] imArray); /*CRITICAL
  someworks(imPrimitive, imArray, imArrayLength);
*/
```

Jnigen will generate that:

```
static inline void wrapped_Java_[package]_[class]_hotNative
    (bool _isCriticalNative, jint imPrimitive, jint imArrayLength, jfloat* imArray)
{
  someworks(imPrimitive, imArray, imArrayLength);
}

JNIEXPORT void JNICALL Java_[package]_[class]_hotNative
    (JNIEnv* env, jclass clazz, jint imPrimitive, jfloatArray obj_imArray)
{
float* imArray = (float*)env->GetPrimitiveArrayCritical(obj_imArray, 0);
jsize imArrayLength= env->GetArrayLength(obj_imArray);
wrapped_Java_[package]_[class]_hotNative(false, imPrimitive, imArrayLength, imArray);
env->ReleasePrimitiveArrayCritical(obj_imArray, imArray , 0);
}

// Some preprocessor to check if it's on desktop.
#if defined(__DESKTOP__)
extern "C" JNIEXPORT void JNICALL JavaCritical_[package]_[class]_hotNative
    (jint imPrimitive, jint imArrayLength, jfloat* imArray)
{
wrapped_Java_[package]_[class]_hotNative(true, imPrimitive, imArrayLength, imArray);
}
#endif 
```
As you see, if you want to get the length of array, you can use _[arrayName]Length_. (On the other side, if you already have an array argument named _XXX_, don't name any other argument as _XXXLength_) If you want to check if the code are called from the critical path, you can use _\_isCriticalNative_. (Don't forget that Hotspot has a _lazy critical entry logic_ for critical natives. The first time calling critical native will always run the normal native function (starts with "Java\_"). After some times (could be hundreds) the Hotspot will replace the normal function with critical one (starts with "JavaCritical\_").)

Pros:
---
- Gives a free performance improving (especially to Matrix4) for desktop. A critical matrix multiplication takes about 35ns when the vanilla one takes about 69ns. A rough SSE+Critical version only takes 24ns (It seems that MinGW can't optimize SSE well? In another native matrix multiplication implementation with MSVC it can takes only 15ns. Whether or not, it should be solved at another time, not now). For comparison, the JOML implementation takes 26ns.

Cons:
---
- Only works on desktop and Hotspot JVM. I don't know if ExcelsiorJET JVM supports this feature.
- Makes the libraries larger! (about 15kb pre library, \*laugh\*)
- Although critical natives is not an new technology, but it's new for using on unofficial projects.

Anyway, it should be treated and tested carefully since it affects all projects which are using Jnigen. I don't know if an engine which focuses on mobile need a desktop-only over-optimization, but it's good for desktop games which depend on heavy matrix operations, and should be enough to quiet the people who complain the performance of matrix of Gdx. 😅

### DIFF
--- a/extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
+++ b/extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
@@ -17,6 +17,8 @@
 package com.badlogic.gdx.jnigen.parsing;
 
 import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public interface JavaMethodParser {
 	public ArrayList<JavaSegment> parse (String classFile) throws Exception;
@@ -92,6 +94,10 @@ public interface JavaMethodParser {
 		public boolean isPlainOldDataType () {
 			return !isString() && !isPrimitiveArray() && !isBuffer() && !isObject();
 		}
+		
+		public boolean isCriticalNativeAllowed () {
+			return isPrimitiveArray() || isPlainOldDataType();
+		}
 
 		public String getBufferCType () {
 			if (!this.isBuffer()) throw new RuntimeException("ArgumentType " + this + " is not a Buffer!");
@@ -116,6 +122,19 @@ public interface JavaMethodParser {
 			if (this == LongArray) return "long long*";
 			if (this == FloatArray) return "float*";
 			if (this == DoubleArray) return "double*";
+			throw new RuntimeException("Unknown Array type " + this);
+		}
+		
+		public String getArrayJniType () {
+			if (!this.isPrimitiveArray()) throw new RuntimeException("ArgumentType " + this + " is not an Array!");
+			if (this == BooleanArray) return "jboolean*";
+			if (this == ByteArray) return "jbyte*";
+			if (this == CharArray) return "jchar*";
+			if (this == ShortArray) return "jshort*";
+			if (this == IntegerArray) return "jint*";
+			if (this == LongArray) return "jlong*";
+			if (this == FloatArray) return "jfloat*";
+			if (this == DoubleArray) return "jdouble*";
 			throw new RuntimeException("Unknown Array type " + this);
 		}
 
@@ -149,10 +168,14 @@ public interface JavaMethodParser {
 
 	/** @author mzechner */
 	public static class JavaMethod implements JavaSegment {
+		//Because Critical Native haven't JNIEnv* and jclass
+		private static final Pattern CRITICAL_NATIVE_CHECKER = Pattern.compile("\\b(env)\\W|\\b(clazz)\\W");
 		private final String className;
 		private final String name;
 		private final boolean isStatic;
 		private boolean isManual;
+		private boolean isCriticalNative;
+		private boolean assertCritical = false;
 		private final String returnType;
 		private String nativeCode;
 		private final ArrayList<Argument> arguments;
@@ -160,7 +183,7 @@ public interface JavaMethodParser {
 		private final int startIndex;
 		private final int endIndex;
 
-		public JavaMethod (String className, String name, boolean isStatic, String returnType, String nativeCode,
+		public JavaMethod (String className, String name, boolean isStatic, boolean isSynchronized, String returnType, String nativeCode,
 			ArrayList<Argument> arguments, int startIndex, int endIndex) {
 			this.className = className;
 			this.name = name;
@@ -170,6 +193,15 @@ public interface JavaMethodParser {
 			this.arguments = arguments;
 			this.startIndex = startIndex;
 			this.endIndex = endIndex;
+			isCriticalNative = isStatic && !isSynchronized;
+			if (isCriticalNative) {
+				for (Argument arg : arguments) {
+					if (!arg.type.isCriticalNativeAllowed()) {
+						isCriticalNative = false;
+					}
+				}
+				checkCriticalNative();
+			}
 			for (Argument arg : arguments) {
 				if (arg.type.isPrimitiveArray() || arg.type.isBuffer() || arg.type.isString()) {
 					hasDisposableArgument = true;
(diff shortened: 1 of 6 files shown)


code-disaster:
Nice one! I've got a couple of native functions which may benefit a lot from this. I'll check the PR as soon as possible. Just need to figure out a nice way to benchmark my use cases.

badlogic:
This is cool! However, may I ask that this functionality can be
disabled/enabled via a flag?

szszss:
@badlogic  No... because I thought it has few side effect (except a larger size). However, on second thought, as it's an experimental improvement, it should be disabled by default. I will make a second commit to disable it by default. The keyword "CRITICAL" will be no longer used for asserting, but used for enabling generating critical native.

extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
```diff
@@ -189,12 +221,64 @@ public boolean isStatic () {
...
+		 * 
+		 * public static native void foo(int i, float[] a1, float[] a2); //in somepackage.Someclass <br/><br/>
+		 * 
+		 * The normal native function is: <br/><br/>
+		 * 
+		 * Java_somepackage_Someclass_foo(JNIEnv*, jclass, jint, jfloatArray, jfloatArray); <br/><br/>
+		 * 
+		 * The critical native function is: <br/><br/>
+		 * 
+		 * JavaCritical_somepackage_Someclass_foo(jint, jint, float*, jint, float*); <br/><br/>
```
intrigus: Shouldn't this be `jfloat*`?

intrigus:
Looks nice, also wanted to implement this, but you've been faster than me ^^

`__APPLE__` is set for both OS X and iOS.
The best solution probably is to use something like [http://www.boost.org/doc/libs/1_60_0/libs/predef/doc/html/predef/reference/boost_os_operating_system_macros.html](this).

It also looks like critical natives with primitive arguments only (no arrays), [aren't faster than normal methods](https://github.com/LWJGL/lwjgl3/issues/175#issuecomment-204699966) with primitives arguments only, so I think we shouldn't generate critical natives in this case.

extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
szszss: Ah, a terrible typo... 😅

code-disaster:
My first attempt with this PR: so far I'm not able to get my JVM call the critical native function. This is on OS X with a Java 8 JDK.

The JNI code is generated correctly. I removed all the generated preprocessor code to rule out errors with wrong defines, so it's just the two pure C functions, critical and non-critical. I use the _isCriticalNative flag as return value, which is always false in my test.

My function doesn't use arrays, just jlong and jint parameters. Maybe that's why the JVM doesn't jump into the critical path?

code-disaster:
Ok, I got it working. In my version, I removed the 'extern "C"' too, but apparently this is important. The critical native function is used now.

Unfortunately, in my test scenario, there seems no immediate gain. On average, the performance seems to level out. I guess I need to test with a different one, using arrays and such.

> before: 181792 calls executed in 37 ms (37482000 ns)
> critical: 181792 calls executed in 37 ms (37596000 ns)

intrigus:
As I have said, there is maybe a difference of 1-2 clock cycles between critical and non critical natives, when the arguments are NOT primitive arrays.

intrigus:
@code-disaster the critical version is only called from compiled code i.e. bytecode after jit-ing.

szszss:
@code-disaster  @intrigus  Well, seems that it doesn't benefit the native methods without array arguments... It may not be a good news since only a few native methods (except matrix) require array arguments.

@code-disaster  And about the failure on OSX. I think the reason is name mangling, and [_extern "C"_ is used to stop it](https://github.com/libgdx/libgdx/pull/4021/files#diff-97645e59007ba8ffc5c27196496859f2R696). (By the way, the regular manner for _extern "C"_ should look like "#ifdef \_cplusplus \n #extern "C" { ... " The reason why I assert that it is compiled as C++, is because "env->" is C++ only. So the compiler couldn't be C compiler.) Anyway, did you figure out an appropriate preprocessor code for platform detection on OSX? Since I haven't a Mac, I had to code-and-pray all the time 😂

code-disaster:
> did you figure out an appropriate preprocessor code for platform detection on OSX?

No, I just removed the whole preprocessor block. In the worst case, this results in a small wrapper function which is never called. If you want to keep the platform check, maybe it'd be sufficient to reverse it, e.g. to "do not compile on Android and iOS".

szszss:
@code-disaster Maybe it's time for Plan B. Set up a compiler definition when generating build targets. It will 100% work if someone uses BuildTarget.newDefaultTarget to generate their build targets.
<del>(And on second (evil) thought, platform check is not such important, ha.)</del>

nhydock:
Use of critical natives came up in [LWJGL3's git](https://github.com/LWJGL/lwjgl3/issues/175) a few weeks ago as well.  Maybe contact Spasi for additional input on the subject

badlogic:
I really appreciate the effort here. But it seems there's little gain for potentially breaking changes. We have a metric ton of native code which would have to be tested against this (box2D, bullet, etc.). I'm not sure the benefits outweigh the negatives.

Issue #4022, opened 2016-04-14
Bullet - btActionInterface Inheritance


floralvikings:
#### Issue details
It would be very useful if it were possible to inherit from the `btActionInterface` class to create custom actions to be executed during a simulation step of a `btDynamicsWorld`; this is how the example `btKinematicCharacterController` class included in Bullet works.

I'm not terribly familiar with JNI, SWIG or how the native wrapping in libGDX works, but if I understand SWIG correctly it should be possible to generate a "director" class from the native C++ code that acts as a go-between for Java and C++.  According to the [Wiki Page](https://github.com/libgdx/libgdx/wiki/Bullet-Wrapper---Using-the-wrapper) on using the Bullet wrapper, there are already several callback classes that do something similar, but I'm completely clueless with regards to how those classes were created/written.

#### Reproduction steps/code
Not really a bug, as you wouldn't expect it to work in the first place, but inheriting from the `btActionInterface` and overriding the `updateAction` method has no effect on the native Bullet code.  

Without knowing that the Bullet classes are wrapped native C++, I would've assumed that the overridden `updateAction` method would execute during the `stepSimulation` call on a `btDynamicsWorld`. 

Using an anonymous override of `btKinematicCharacterController` as an example; as far as I'm aware, this and `btRayCastVehicle` are the only concrete implementations of this class:

```java
new btKinematicCharacterController(body, shape,  DEFAULT_SIZE.y * 0.5f) {
    @Override
    public void updateAction(btCollisionWorld collisionWorld, float deltaTimeStep) {
        super.updateAction(collisionWorld, deltaTimeStep);
        System.out.println("Test"); // Doesn't execute, no native calls to the Java code
    }
};
```

#### Version of LibGDX and/or relevant dependencies
All versions of gdx-bullet

#### Stacktrace
Not Applicable

#### Please select the affected platforms
- [x] Android
- [x] iOS
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

floralvikings:
I've actually got this working locally now, after adding this line to `dynamics.i`:

```c
%feature("director") btActionInterface;
```

I regenerated the swig files using the Ant build in `extensions/gdx-bullet/jni/swig` and rebuilt libGDX from source.

[...]

PR #4025, opened 2016-04-15
Add wrapper class for btActionInterface

floralvikings:
See #4022 

The first commit contains the non-generated files; `CustomActionInterface.h` wraps the btActionInterface C++ class and the changes to `dynamics.i` include this new class and specify it as a director class.

The second commit is the regeneration of the swig source files; swig 3.0.8 was used, most of the changes are just updates to the version number in the leading comment.


xoppa:
Can you remove (`%ignore` iirc) the old signature and debug drawer method from swig, so it never has to create those objects?

This change doesnt allow to customize `btKinematicCharacterController` though, which was the initial use-case you provided. Can you provide a new use-case or perhaps include one in the tests so others can see how to use it?

floralvikings:
I've added `%ignore` directives for those two methods.

The basic use case for overriding this class is to hook into the `stepSimulation` method of a `btCollisionWorld`, after collision checks have been made but before the motion states have been synchronized.  I'm using it to add listeners to respond to a few specific collisions, but basically anything that would benefit from hooking into that part of the world simulation.

As far as customizing `btKinematicCharacterController`: my initial thought here was to write my own implementation tailored more closely to my project, so customizing it wasn't really the idea so much as reimplementing.  If you think that would be a useful addition, though, I can add swig directives for it and make it extensible as well

xoppa:
Looks like the `debugDraw` is still included. Better implement it to prevent a pure virtual function as well. Perhaps you can do the same thing for that method as well, so that it meets the API as much as possible. Thus, make it call a new `virtual void debugDraw() {}` method (without any arguments, with default empty implementation). Let's not include directors unless there's at least one use-case, so no need to add one for `btKinematicCharacterController` as well.

floralvikings:
Done.  Thanks a lot for helping get this together, by the way, I really appreciate it 😄

floralvikings:
@xoppa does this still need some work?  Just let me know if it needs more tweaking.

xoppa:
Thanks!

Issue #3018, opened 2015-04-09
3D ParticleSystem crashes second time the game starts on Android


krial057:
Hey,

I'm using 3d Particle Effect in my Game on Android. However, the game will crash the second time I start it. Here is the error log:
```
04-07 15:32:10.680: E/AndroidRuntime(31308): FATAL EXCEPTION: GLThread 644
04-07 15:32:10.680: E/AndroidRuntime(31308): com.badlogic.gdx.utils.GdxRuntimeException: java.lang.NullPointerException
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at com.badlogic.gdx.assets.AssetManager.handleTaskError(AssetManager.java:540)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at com.badlogic.gdx.assets.AssetManager.update(AssetManager.java:356)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at lu.internet.alain_krier.ggame.World.update(World.java:239)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at lu.internet.alain_krier.ggame.screens.GameScreen.update(GameScreen.java:30)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at lu.internet.alain_krier.ggame.screens.GameScreen.render(GameScreen.java:35)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at com.badlogic.gdx.Game.render(Game.java:46)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at lu.internet.alain_krier.ggame.Ggame.render(Ggame.java:46)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame(AndroidGraphics.java:422)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1516)
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1240)
04-07 15:32:10.680: E/AndroidRuntime(31308): Caused by: java.lang.NullPointerException
04-07 15:32:10.680: E/AndroidRuntime(31308): 	at com.badlogic.gdx.assets.AssetManager.get(AssetManager.java:148)
[...]

PR #4027, opened 2016-04-17
Make it possible to use multiple ParticleSystems

intrigus:
This makes it possible to use multiple `ParticleSystems`.
There is no reason, for `ParticleSystems` being static.
This could also fix https://github.com/libgdx/libgdx/issues/3018 , although it's unclear whether `ParticleSystem` caused the issue at all.


badlogic:
Sounds sensible to me. @xoppa, please sign off on this and merge.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleSystem.java
```diff
@@ -12,6 +12,9 @@
 public final class ParticleSystem implements RenderableProvider{
 	private static ParticleSystem instance;
 	
+	/**
+	 * @deprecated Directly use the constructur
```
yuripourre: Typo: _constructur_ ?

PR #4028, opened 2016-04-17
Add test starter, allows to start all tests using LWJGL3 backend

intrigus:
This is a replacement for the old AWT test starter. It has been
necessary since AWT can't be used with LWJGL 3.
A new test is created in a new window, using the new window APIs from
LWJGL3.


js78:
I tested it and it works great ! Thanks @intrigus

intrigus:
Currently the last selected test from the previous run is highlighted during the whole run of the test starter. Do you think, that I should also select the last selected test from the current run and deselect the test from the previous run?

js78:
Indeed I have noticed that and I think you should always highlight the last test. It's more logical.

intrigus:
Fixed it.

js78:
@xoppa Could you test this pull request and merge it ? It's a step forward to deprecate lwjgl 2.

badlogic:
Thanks, merged!

PR #4032, opened 2016-04-18
Add getter for deadzone

Dibro89:


js78:
Hi @Dibro89 !
Maybe you could call the method like the property `getDeadzoneRadius`.
Moreover, you should squash your commits.

badlogic:
Will reopen if you follow @js78 recommendation of squashing the commits and calling it `getDeadzoneRadius`.

Dibro89:
@badlogic I called it like the setter `setDeadzone`. And sorry but I don't know how to squash commits.

js78:
@Dibro89 
First rename your method to `getDeadzoneRadius`. `setDeadzone` is a bad name and must be changed but not now because it breaks the api. Make a new commit. Now you can squash your commits.

To squash your commits, do that:

First determine the older commit id (add getter for deadzone) `git log`
Then `git rebase -i commitID^`
The `^` sign means : the one before

Then you have a list of commits, set all prefixes to `s` (meaning squash) except the first one like in the picture.
Then save and it will squash your commits, try to understand what you do by reading git help.
capture_git_rebase

Dibro89:
@js78 I think I did it...

js78:
Then you must push on your branch with --force option because you are rewriting the history.

PR #4050, opened 2016-04-24
moe backend fixes

igorushko:
IOS Preferences
	fix java.lang.ClassCastException: java.lang.String cannot be cast to ios.foundation.NSString
	Using NSAutoreleasePool on moe going to hangs
IOSInput fix some copy-paste bugs, merge with robovm backend

### DIFF
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSApplication.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSApplication.java
@@ -383,11 +383,11 @@ public class IOSApplication implements Application {
 		File finalPath = new File(libraryPath, name + ".plist");
 		String path = libraryPath + "/" + name + ".plist";
 
-		NSMutableDictionary<NSString, NSObject> nsDictionary = NSMutableDictionary.dictionaryWithContentsOfFile(path);
+		NSMutableDictionary<NSString, Object> nsDictionary = NSMutableDictionary.dictionaryWithContentsOfFile(path);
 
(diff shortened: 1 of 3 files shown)


Tom-Ski:
Thanks for the contribution, I'd like to fix these in different ways. Will be pushing fixes shortly.  
Not sure what your problem was with CoreMotion, could you explain why you removed it for the deprecated UIAccelleromter?

igorushko:
UIAccelleromter its just copy-paste from robovm. 

Main fix it remove breaks at this
`public boolean isCatchMenuKey() {
  		return false;
 @@ -544,15 +561,12 @@ void processEvents () {
  				if (event.phase == UITouchPhase.Began) {
  					if (inputProcessor != null) inputProcessor.touchDown(event.x, event.y, event.pointer, Buttons.LEFT);
  					if (numTouched == 1) justTouched = true;
 -					break;
  				}
  				if (event.phase == UITouchPhase.Cancelled || event.phase == UITouchPhase.Ended) {
  					if (inputProcessor != null) inputProcessor.touchUp(event.x, event.y, event.pointer, Buttons.LEFT);
 -					break;
  				}
  				if (event.phase == UITouchPhase.Moved || event.phase == UITouchPhase.Stationary) {
  					if (inputProcessor != null) inputProcessor.touchDragged(event.x, event.y, event.pointer);
 -					break;
  				}
  			}
  			touchEventPool.freeAll(touchEvents);`

and save touches `touchDown[event.pointer] = touch;` merge it only, now input is broken without this fixs

Tom-Ski:
Yes, will be pushing that soon, just testing.  Accel is already implemented through CMMotionManager

PR #4051, opened 2016-04-24
Added adapter class for Net.HttpResponseListener

nrallakis:


intrigus:
Could you present an use-case for your change?
I think that it's pretty uncommon to explicitly create a class that implements `HttpResponseListener` or extends your adapter.
Most people will probably just create an anonymous class.

MobiDevelop:
Anonymous subclasses of interface adapters are a pretty common use-case when you only care to handle one interface method without cluttering you code with empty methods.

nrallakis:
In my case, my app makes an http request to facebook, trying to retrieve a profile picture. However, it is not that important to succeed, because there is a default picture in place...

badlogic:
I'd rather not include this as doing it in your own project is trivial enough.

PR #4052, opened 2016-04-24
Allow pixmap to be loaded y-up.

js78:
I am not sure if I have to include jni files or we build them so I wait for your comments.
I talked with Tomski about this functionality which is included in stb_image and he was ok if the method was well documented and difficult to use without knowledge. Indeed, enabling y flipping make spritebatch not working (flipping uv).


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -184,6 +184,13 @@ public Buffer getPixels () {
 		return buffer;
 	}
 
+	/**
+	 * This method doesn't work with GWT.
+	 * @param flag
+	 */
+	static public void setFlipY(boolean flag) {
```
Tom-Ski: Please use the libgdx formatter on the java code that you change/add.  Keep `public static` order for consistency also.

gdx/src/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -161,6 +161,13 @@ public void setColor (int color) {
 		this.color = color;
 	}
 
+	/** Warning, only use this method if you know what you are doing. Set pixmap to Y-up if flag is true. This method doesn't work
+	 * with GWT and is not guarantee for jpeg files.
+	 * @param flag */
+	static public void setFlipY(boolean flag) {
```
Tom-Ski: Same goes for here

gdx/src/com/badlogic/gdx/graphics/g2d/Gdx2DPixmap.java
```diff
@@ -247,6 +247,13 @@ public int getGLType () {
 		return toGlType(format);
 	}
 
+	public static void setFlipY(boolean flag) {
```
Tom-Ski: LibGDX formatter has a space between method name and parameters.

Tom-Ski:
Looks good apart from the few formatting issues, what have you tested this on?

xoppa:
Perhaps rename the static method to `setDefaultFlipY` and make it also a constructor argument. E.g. I can imagine `ModelLoader.ModelParameters` having a option in for this, but you don't want that to conflict with normal texture (e.g. Skin) loading.

Would it be possible to emulate this in GWT? For example, copy the data line by line?

js78:
@Tom-Ski I'm testing it on Android (nexus 5) and it doesn't work for now, I get a `No inplementation found for GDX2DPixmap.flipY`. It works on desktop, does there is something else to build for android ?
@xoppa It would be great to add it as an constructor argument but in stb_image, it's static so with multi thread, it could happen bad behavior. I think it's possible to emulate this in GWT.
Moreover, why gdx2d.c uses jpgd whereas stb_image load jpeg files ?

xoppa:
Did you copy over the new .so files @js78? On android you have to copy those manually to the libs folder after building them.

Indeed, too bad it's static, Could you add a getter for it instead?

Iirc jpgd is used when stb_image fails: stb_image didn't support all formats (e.g. progressive). But they might be on par since last updates.

js78:
@xoppa I didn't copy the so file, it must be my problem.
I have another idea. I can copy the stb_image flipping method in gdx2d.c and like that I can pass an attribute to the load method. Moreover, it would work with jpgd. What do you think?

js78:
@xoppa I updated the commit. I can't make it run on Android, I think I don't really understand which .so file I have to copy and where.
Can you try on Android ?
Moreover, you can now use a constructor argument for loading y up.

gdx/src/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -127,28 +130,38 @@ public Pixmap (int width, int height, Format format) {
 	/** Creates a new Pixmap instance from the given encoded image data. The image can be encoded as JPEG, PNG or BMP.
 	 * @param encodedData the encoded image data
 	 * @param offset the offset
-	 * @param len the length */
-	public Pixmap (byte[] encodedData, int offset, int len) {
+	 * @param len the length
+	 * @param yUp load data y up */
+	public Pixmap (byte[] encodedData, int offset, int len, int yUp) {
```
Darkyenus: Please don't use `int` for what really is a `boolean`, at least in the public API, it is very confusing.

js78: Indeed I was hesitating but it make more sense to be a boolean.

js78:
@xoppa I can't make it work on android. I did everything you say:
- Launch GDXBuild.java in gdx
- ant -f build.xml and setting "build native" to true
- copy "dist/arm*" folders into gdx-test-android libs folder.

It doesn't work. In my java build path of eclipse I added a native library location to libs/armabi but it didn't help. It's strange because tests work on Android but it's just not reversing texture...

Could you try when you have time ?

badlogic:
Looks good to me in principle. Please also add an entry to CHANGES. @xoppa, pull the trigger if you feel like it.

js78:
@badlogic @xoppa I have to make a small modification (using boolean instead of int parameter) before merge. But I wait for xoppa to test it on android because I don't manage ot do it, shame on me !

Tom-Ski:
Hey @realitix if you could update this to master I can take a look at the android side for you.

realitix:
Hi @Tom-Ski.
Ok I will try to find time this week!

realitix:
@Tom-Ski I have rebased this pull request on master, I let you try on Android if you can make it works.

Tom-Ski:
All good @realitix. Can we do something about gwt?

realitix:
Yes it can be done with gwt but I don't know how because I don't use gwt.

PR #4055, opened 2016-04-28
Update Net.java

CypherEA:
Currently, the HTTPResponse implementation returns a Map<String,List<String>> of Headers.
However, HTTPRequests implements Headers as Map<String, String>

That way, it's impossible to addHeader to have multiple headers of the same name (like multiple Cookies).
Changed headers to Map<String, List<String>>
Changed getHeaders accordingly (for request)
Changed setHeader accordingly.
Added addHeader.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Net.java
+++ b/gdx/src/com/badlogic/gdx/Net.java
@@ -32,6 +32,7 @@ import com.badlogic.gdx.net.Socket;
 import com.badlogic.gdx.net.SocketHints;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.Pool.Poolable;
+import java.util.ArrayList;
 
 /** Provides methods to perform networking operations, such as simple HTTP get and post requests, and TCP server/client socket
  * communication.</p>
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/Net.java
```diff
@@ -175,8 +176,15 @@ public void setUrl (String url) {
 		 * @param name the name of the header.
 		 * @param value the value of the header. */
 		public void setHeader (String name, String value) {
-			headers.put(name, value);
+                    List<String> val = new ArrayList();
```
MobiDevelop: You should be doing a get from the headers map before creating and adding a new list.

MobiDevelop:
This is a breaking change so would need an entry in changes.

gdx/src/com/badlogic/gdx/Net.java
MobiDevelop: Oh, this is intended to replace any previously set value, never mind that last comment in that case.

MobiDevelop:
It doesn't compile because you didn't update JavaNetImpl to use it correctly.

Darkyenus:
Since it already is a breaking change, we could also transition from java.util to libGDX collections (not in this PR but in the same release).

MobiDevelop:
Since this doesn't compile, I'm going to close this out. Please fix it up and try again when you have time.

PR #4059, opened 2016-05-01
Increased fidelity of progress tracking.

psiegel:
Added a simple counter of total tasks created by individual asset loads,
which can be compared against the size of the current task stack to
determine fractional progress towards completion of a single asset.
This increases the fidelity of the progress value returned, which is
especially useful for assets with a large number of dependencies.

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -75,6 +75,7 @@ public class AssetManager implements Disposable {
 	AssetErrorListener listener = null;
 	int loaded = 0;
 	int toLoad = 0;
+	int peakTasks = 0;
         
 	final FileHandleResolver resolver;
 
@@ -318,6 +319,7 @@ public class AssetManager implements Disposable {
 		if (loadQueue.size == 0) {
 			loaded = 0;
 			toLoad = 0;
+			peakTasks = 0;
 		}
 
 		// check if an asset with the same name but a different type has already been added.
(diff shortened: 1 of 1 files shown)


psiegel:
In case it's not obvious what this is doing, here's a more in-depth description of how I ended up making this change:

In my own game, I created a custom loader for a high level data object called a "scene".  The scene file is essentially an xml file with a ton of references to other files (3d models, particle systems, etc.).  I then load the single file using the AssetManager and rely heavily on the underlying dependency tracker to load all the pieces.

The result is that my loading screen would show a progress bar that never progresses for a long period of time, then suddenly completes.  This is because the AssetManager was only ever asked to load one thing, so the progress value it returned was always 0 until it completed, then it suddenly became 1.  The fact is that my single asset had 32 dependencies and none of these were ever represented by the progress bar.

This change takes into account the task stack size for progress calculation.  When the AssetManager has loaded 16 of the 32 dependencies, the task stack drops to 16 while the new peakTasks counter remains at 32, thus the fractionalLoad value is calculated as 0.5, and the total progress then returns as 0.5  (0.5 / 1).  The end experience is a smoother progression to the loading progress bar during initial load of the game.

slebed:
Thank you so much for this.  I've been wanting this for a while.

badlogic:
Looks good to me, @NathanSweet please merge if you feel like it.

psiegel:
Did this get forgotten somewhere along the line?  Sure would like to see it merged in.

NathanSweet:
I never feel like looking at the beast that is AssetManager. I will trust @psiegel it works as advertised. Don't fail me! ;)

psiegel:
Thanks @NathanSweet!  Ping me if there are any problems, but I'm sure it'll be fine. :)

PR #4066, opened 2016-05-04
Classpath fonts can now be referenced in skin JSON

czyzby:
Currently, this is properly parsed by Desktop and GWT, which basically use the same mechanism for internal and classpath files:
```
  com.badlogic.gdx.graphics.g2d.BitmapFont: { 
    default: { file: { com/badlogic/gdx/utils/arial-15.fnt }
  }
```

However, this fails on Android. It would be nice if you could reference fonts located at classpath (for example - provided by external libraries), so you wouldn't have to include your own font files. Very useful for prototyping.


badlogic:
I'm afraid if it doesn't work for the other backends, we can't merge it. It's also an undocumented surprise if the Skin loader would all of a sudden also try to fetch things from the classpath.

PR #4070, opened 2016-05-08
Actor additions

jasoncarrete5:
### Overview
Added methods to the Actor class that have similar functionality to Actor#positionChanged(), Actor#sizeChanged(), and Actor#rotationChanged() but pass the change in quantity for the implementer to access when the method is called.

##### Example
Instead of overriding the positionChanged() method, the implementer could choose to override positionChanged(float deltaX, float deltaY). This would tell the implementer how much the Actor moved by rather than just simply notifying a change in position.

##### Hypothetical code
```
@Override
public void positionChanged(float deltaX, float deltaY) {
   if(Math.sqrt(deltaX*deltaX + deltaY*deltaY) > MAX_SPEED) {
      //insert logic to handle a speedy player
   }
}
```

##### Reasoning
I find myself implementing this simple feature often so I figured why not add it to libgdx :)? Its simple, it doesn't break anything, and it only adds a few primitive variables to the Actor class.


badlogic:
@NathanSweet what do you think? I'm not the biggest fan personally, as i believe an Actor implementation should keep track of change quantities itself. This would add additional house keeping for everyone that doesn't need this feature.

PR #4072, opened 2016-05-09
support both width, height scale for the tile cell

dotw:
Current `OrthogonalTiledMapRenderer` class only supports an uniform scale parameter for the cells.  Sometimes, we want to show the tile map in the whole screen, this usually need to use different scale values.


badlogic:
@MobiDevelop could you give this a looksy?

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
```diff
@@ -60,7 +60,9 @@
 	
 	protected TiledMap map;
 
-	protected float unitScale;
+//	protected float unitScale;
```
MobiDevelop: Don't comment out, just delete old code.

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
```diff
@@ -92,24 +98,26 @@ public Rectangle getViewBounds () {
 	}
 
 	public BatchTiledMapRenderer (TiledMap map) {
-		this(map, 1.0f);
+		this(map, 1.0f, 1.0f);
 	}
 
-	public BatchTiledMapRenderer (TiledMap map, float unitScale) {
+	public BatchTiledMapRenderer (TiledMap map, float unitScaleX, float unitScaleY) {
```
MobiDevelop: Would prefer to make this a new constructor, leaving the original to accommodate the far more common use case of uniform unit scale (and to avoid breaking existing code, if possible).

MobiDevelop:
As this does change the API, would need to add to the CHANGES file.

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
dotw: yeah. add a new constructor will be better.

dotw:
I found this is NOT integrated into the latest release.

MobiDevelop:
I am trying to think of why this was needed and the lack of a descriptive use case here isn't helping. If this is still desired, please open a new PR with only relevant changes and a descriptive use case as to why this change is needed

PR #4073, opened 2016-05-09
Implemented OS detection

kerberjg:
This PR adds the capability of querying the Application objects for the OS they're running on.
This also replaces the the `ApplicationType.Android` and `ApplicationType.iOS` with a `ApplicationType.Mobile`, as now the OS choice is delegated to `Application.getOS()`


MobiDevelop:
I'm not convinced this breaking change is worth it. Why did the ApplicationType enum have to change?

kerberjg:
Since the OS is would be reported by `getOS()`, I see no reason why having a redundant information in `getType()` would be needed, it would make no sense whatsoever.

This PR was devised taking into consideration how many people are looking for an effective way to detect the OS their game is running on, and as such they're currently forced to write their own custom solution, every time. It would make much more sense for `Application` to provide this information.

laubed:
I agree to MobiDevelop that we should keep the ApplicationtionType as it is to avoid breaking external code. 
Probably it is better to first introduce a second ApplicationType getter that acts like yours and make the old one deprecated and change that later.
I really thing your implementation of the ApplicationType is more intuitive but as MobiDevelop states it might not be worth breaking existing code

badlogic:
Knowing the specific ApplicationType has value, as it allows you to figure
out what backend you are running on.

Absolutely not a fan of deprecation. We either break it or not.

kerberjg:
@laubed I was just writing that, actually.
It could be a good idea indeed to introduce a new enum (let's say `PlatformType` for example) and keep back the ApplicationType as a deprecated enum, if we're going down the deprecation road.

kerberjg:
It all boils down to providing a OS detection feature to the users, while consolidating the `ApplicationType` enumerator.

laubed:
Then leave ApplicationType relatet things as they are and implement another Backend enum.
The getOS() thing is absolutely useful.

kerberjg:
@badlogic Not exactly, as currently `ApplicationType` doesn't differentiate between LWJGL, LWJGL3 and JGLFW. It's just an inconsistent hybrid between a "BackendType" and "DeviceType"

laubed:
So the best thing then is to clearly seperate between the OS, Backend and actual ApplicationType.
OS should return the specific OS (Windows, Linux, Mac, Android, iOS) and ApplicationType just the type of Application (Mobile, Desktop, Web).
The Backend should really point to the backend being used.

kerberjg:
Sounds good to me, I could do that in a few minutes if we all agree

kerberjg:
Actually it's done now. `Application` and all backends are ready.

MobiDevelop:
This doesn't even compile, are not testing your changes before pushing them up?

kerberjg:
It wouldn't compile since I cloned. There might be something wrong with my configs, I'll check it out now.

laubed:
Except from missing imports i don't see why this shouldn't compile.
From the logic point of view it looks far more intuitive than the other implementation.

MobiDevelop:
Click the details on the LibGDX PR Check then look at the build console logs - there are classes in libgdx itself that look for the existing ApplicationType values which would need to be updated to use these new methods/values.

kerberjg:
@laubed As I said, it's probably something with my Eclipse configs. Cloning everything again, I'll re-import all the projects.

@MobiDevelop Thanks, proceeding to fix ASAP

kerberjg:
This would be so much quicker if I could get gdx projects to build on my machine... (all libs/ folders missing in every project)

MobiDevelop:
You probably didn't run the fetch script as is described in the [wiki](https://github.com/libgdx/libgdx/wiki/Building-libgdx-from-source).

kerberjg:
Later I found it on the forum [here](http://badlogicgames.com/forum/viewtopic.php?f=11&t=7628#p34893), but thanks anyway :)

kerberjg:
BTW, since `Application` will now contain a `BackendType` enum, would be Graphics#GraphicsType still still a necessity? It might be just my impression, but at this point it seems redundant to me.

Here are the two for comparison:
```java
public enum BackendType {
	Android, RoboVM, Headless, LWJGL, LWJGL3, JGLFW, GWT
}
```

```java
public enum GraphicsType {
	AndroidGL, LWJGL, Angle, WebGL, iOSGL, JGLFW, Mock, LWJGL3
}
```

The only difference seems to be `Angle` in GraphicsType, but eventually GrahpicsType could be rewritten to something like this:
```java
public enum GraphicsType {
	GL, GLES, WebGL, Angle, Mock
}
```

laubed:
Im not that deep into the backend stuff but aren't backends bound to a
graphics type?
It's only redundant if one backend only uses one graphics type. If it might
use two that its necessary to keep them.

2016-05-09 20:30 GMT+02:00 Jakub Kerber <notifications@github.com>:

kerberjg:
@laubed They actually are, even with more backends sharing the same/similar graphics library, but that little `Angle` makes me perplexed.

laubed:
Then we probably leave that as it is.

2016-05-09 20:43 GMT+02:00 Jakub Kerber <notifications@github.com>:

kerberjg:
It finally builds!

Also I've actually searched all the files for a `GraphicsType.Angle`, but it's nowhere to be found in my local source. Running a search query on Github just to make sure.

kerberjg:
I've just found it in an abandoned repo which seems to be a fork of LibGDX from 5 years ago.
https://github.com/eerock/libgdx/blob/master/backends/gdx-backends-angle/src/com/badlogic/gdx/backends/angle/AngleGraphics.java

I guess this means that `GraphicsType.Angle` should be removed entirely, but at this point I might as well just refactor the whole GraphicsType

MobiDevelop:
We used to have an Angle backend, but it has been gone for a long time.

kerberjg:
Everything seems to work now, that leaves us with just a few questions:
 * `Graphics.GraphicsType` is used in some parts to detect the backend; should it be removed/refactored?
 * To merge or not to merge?

kerberjg:
Added BlackBerry OS 10 detection

laubed:
I really think this ApplicationType and SystemType is the wrong approach to a problem.
Different use cases need different granularity of detail. 
We might implement some kind of isOS(OSType) and isApplicationType(ApplicationType) which is based arount a tree type structure.

isApplicationType(Mobile) should return true as well as isApplicationType(Tablet) if running on a tablet because Tablet is just a specification of the Mobile ApplicationType. Same goes with OS

kerberjg:
That would actually make little sense, mostly in cases where a `switch(Gdx.app.getSystemType())` statement would be used (like  in [GLVersion](https://github.com/kerberjg/libgdx/blob/feature-devicetype/gdx/src/com/badlogic/gdx/graphics/glutils/GLVersion.java#L41)). We'd have to use as many function calls as supported OSes, instead of one call, in other words it'd be a code bloat.

It makes much more sense to let enums handle that, like it's been so far for `ApplicationType` and `GraphicsType`.

Tom-Ski:
This is getting a little out of scope.  
A few things, please use the LibGDX formatter, and don't change code that doesn't require changing. There is no benefit to switching GLVersion to a switch statement, and it just reduces the chances of your PR being merged.

All internal code that is backend specific, for example FrameBuffer on iOS should rely on a the fixed compile time constant, at the moment this is the ApplicationtType, but it makes sense to change this to some backend type. We should not rely on a query for System type for any backend specific code.

I agree with @laubed that this is not a good enough solution to the issue, especially with the changes to ApplicationType.  I suggest we just fix up ApplicationType to either respond with backends, or we make it a little bit more involved, and let ApplicationType be a container in which you can query the backend responsible for that type. e.g  Gdx.app.getType().getBackend().

kerberjg:
If don't consider a benefit avoiding double checks on exactly the same values and making the code more readable, I can revert it back. Though I agree this was out of scope.

Also the FrameBuffer looked [more OS-specific](https://github.com/libgdx/libgdx/pull/4073/commits/d7579f011dcf7419fe85f3210cdedf39252585cd#diff-e83c8b0f317a57e048027ce338aa53caL132) than backend-specific to me. Also given the fact that RoboVM got discontinued and the new iOS is on the horizon, I'm quite sure that querying for the OS type would make that particular line more future-proof, if the iOS backend was to ever change again.

Also looking at it again, it seems it's slowly getting out of hand indeed, I don't have any objections about replacing the `ApplicationType` values with `BackendType`'s values, but at this point we might just replace the name with `BackendType` as well, given that the former was far less specific or meaningful in this particular case.
Making a container class out of it would be an unnecessary bloat, we're talking about turning an enumerator into a class with more enumerated values inside. Even having 3 enumerators is better than that.

Tom-Ski:
It is backend specific, for example you could be running the GWT version of the app on an iPhone, giving the OS to be iOS yet running on WebGL and not the iOS backend specific GLES.

kerberjg:
I've just attempted to replace ApplicationType with BackendType, which pretty much didn't end well. Every 
```java
if(Gdx.app.getType() == ApplicationType.Desktop)
``` 
would need to be replaced with a 
```java
if(Gdx.app.getBackendType() == BackendType.LWJGL || Gdx.app.getBackendType() == BackendType.LWJGL3 || Gdx.app.getBackendType() == BackendType.JGLFW)
```
which doesn't seem to be a good idea in any case!

Knowing the kind of device the application is running on is much important as knowing the backend. If you're not sure of that, just read every line of libGDX code I had to change.

kerberjg:
Given the previous commit, I've reached the conclusion that removing any of the 3 enumerators (ApplicationType, BackendType, SystemType):
 - doesn't bring any benefits
 - makes the code much less clean
 - in the best case brings us back to square one

badlogic:
So, after watching this for a few days, i believe we are actually back at square one :)

I'll be the bad cop and close this out now. I see meager benefits compared to the breakage a change like this would bring.

kerberjg:
At this point I could actually take back everything breaking and just add
the OS detection, since it's what all this was really about.

PR #4075, opened 2016-05-11
Add window size limits to LWJGL 3 app and window configurations.

cypherdare:
Making very nice GLFW feature available.


badlogic:
Thanks!

PR #4077, opened 2016-05-11
OS detection (non-breaking)

kerberjg:
This PR adds OS detection to all backends, without breaking the previous enums like ApplicationType.

With this the OS can be queried by calling `Gdx.app.getSystemType()`. The supported systems are:
`Windows, Linux, OSX, Android, iOS, BlackBerry10`

The Android backend is able to differentiate between Android and BlackBerry10, just like GWT can detect all the systems in the enum.
The desktop backends (LWJGL, LWJGL3, JGLFW) are able to differentiate between Windows, OSX and Linux.


badlogic:
WHy not call it ``getOS()`?

kerberjg:
To keep the naming pattern between the enum and the getter.

Tom-Ski:
Lots of bloat in here, as all you are doing is parsing the system string to the enum values, the parsing could all be consolidated into core instead of littering the backends.

kerberjg:
Getter renamed, parsing moved to the core.
The OS type parsing in GWT has been kept inside the backend, as it uses a backend-specific class.

kerberjg:
I've reformatted the edited files with the Eclipse formatter and updated CHANGES as well.

intrigus:
Please only format things you have edited. 
Formatting everything makes reviewing harder.

kerberjg:
Merged the latest changes in `libgdx:master` thus resolving the conflicts, as well as adding the `getOS()` to the new MOE backend.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessApplication.java
```diff
@@ -189,6 +189,19 @@ public ApplicationType getType() {
 	}
 
 	@Override
+	public SystemType getOS () {
+		String os = java.lang.System.getProperty("os.name").toLowerCase();
```
Tom-Ski: Looks like you missed formatting on this block.

Tom-Ski:
What backends have you tested this on, and with what devices?

kerberjg:
Fixed now, thank you.

I tested it on a MacBook Pro with OSX 11 and a few PCs with Windows 7/8/10, Ubuntu 14.04 and Linux Mint 17. Returned the correct OS on LWJGL3, and thus it wasn't necessary to test it on LWJGL/JGLFW/Headless as this feature uses the same runtime-dependent method to get the OS name.
I didn't see the need to test it on either Android or iOS, as the method just returns a fixed value on these platforms.

Tom-Ski:
Android doesn't return a fixed type.

MobiDevelop:
What if I want to know that I am on a Kindle Fire? That's Android, but not.

kerberjg:
@Tom-Ski Right, but I do not have a BlackBerry device to test it on, nor I have installed an emulator yet. However I'm sure it's supposed to work, [theoretically](https://supportforums.blackberry.com/t5/Android-Development/How-to-Detect-if-an-Android-Application-is-Running-on-a/ta-p/1470313).

@MobiDevelop I could include the Fire OS detection at this point too. If you have a Kindle Fire device, could you kindly test it? ;)

MobiDevelop:
I have three Kindle Fire tablets, I could test it.

kerberjg:
Looks like the only correct way to verify if the app is running on a Fire OS device is to check the Build.MANUFACTURER variable, not Build.MODEL like many people had suggested; it turns out that the list of model identifiers is long, and they don't always include common or particular words or characters. Sources: 
 - https://developer.amazon.com/public/solutions/devices/fire-tablets/specifications/01-device-and-feature-specifications
 - https://developer.amazon.com/public/solutions/devices/fire-phone/docs/fire-phone-specifications
 - https://developer.amazon.com/public/solutions/devices/fire-tv/docs/device-and-platform-specifications

I realize the way it's written isn't the best. I'm planning on rewriting it at least partially once I get the confirmation that it works correctly.

kerberjg:
I've looked into the possibilities of rewriting the OS identification methods I've moved to the core, and the most logical solution would be to move the `ApplicationType.parseAndroidOS()` to the Android backend, however that would require me to either move this method to a new class somewhere in the Android backend, or into `AndroidApplicationBase`, thus turning it into an abstract class.
I personally don't think it's worth to do either, as it would unnecessarily bloat the backend.
If there aren't any further suggestions, I'll leave `ApplicationType.parseAndroidOS()` as it is.

Also, @MobiDevelop, have you had any chance to test the Fire OS detection?

MrStahlfelge:
Fire Devices can be recognized by the Manufacturer AMAZON. If it is non-touch, it is a FireTV. On Android, you sometimes want to know if it is Amazon or Google Play - but you should better use flavors anyway. And I don't see a reason to know the API level in core code.

For GWT it is uninteresting what OS the web browser is on. It is way more interesting if you run on Safari, Chrome or Firefox (or Electron). I don't see a solution for this here.

Closing this because I don't think we will ever get a solution that fits all needs. Games must use platform code here if there are differences between Manufacturers, Web browsers, and OS.

PR #4078, opened 2016-05-11
Make addAsset public to avoid inheritance just to access this method

VinceAngel:
Don't know what to say more, this is a simple change to make this method more easy to use.


xoppa:
There's a reason it is not public ;) see #270

VinceAngel:
Ok, I see. Thanks whatever.

PR #4080, opened 2016-05-13
Small fixes for JavaDoc in Body.java

lagebr:
Spell checking and some in-file consistence for the method headers.


intrigus:
I don't think that this javadoc is ever used, since it's an emulation for gwt.

lagebr:
Yeah, probably. I just had to fix it.

PR #4084, opened 2016-05-16
[GWT] Fix problem with logging, when app is not yet fully initialized

intrigus:


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
```diff
@@ -296,10 +297,29 @@ public Net getNet() {
 		return Gdx.net;
 	}
 
+	private void updateLogLabelSize () {
+		if (log != null) {
+			if (graphics != null) {
+				log.setSize(graphics.getWidth() + "px", "200px");
+			} else {
+				log.setSize(400 + "px", "200px"); // Should not happen at this point, use dummy value
```
czyzby: I think `"400px"` is sufficient, no need to concat int with a `px` string.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
```diff
@@ -296,10 +297,29 @@ public Net getNet() {
...
 			log = new TextArea();
-			log.setSize(graphics.getWidth() + "px", "200px");
+			
+			// It's possible that log functions are called
+			// before the app is initialized. E.g. SoundManager can call log functions before the app is initialized. 
+			// Since graphics is null, we're getting errors. The log size will be updated later, in case graphics was null
+			if (graphics != null) {
+				log.setSize(graphics.getWidth() + "px", "200px");
+			} else {
+				log.setSize(400 + "px", "200px"); // Dummy value
```
czyzby: Same here, `"400px"` seems sufficient.

MobiDevelop:
Out of curiosity, why not use the config's width instead of an arbitrary size?

intrigus:
That's a good questions. I somehow didn't even think about doing that.
Should I change or do you want to change it?

PR #4089, opened 2016-05-18
Auto search project for GdxTest

Ilya-Gazman:
Instead of maintaining a list of all the test files, I perform a recursive search over all the files in the project and pick them up automatically.
This adds few tests to the list.


Tom-Ski:
Thanks for the PR, but closing this out. Its often very handy to be able to disable on the fly without having to remove the classes from the file structure or add in some filter.  While it does not look nice on the eyes, it does have some functional use.

PR #4093, opened 2016-05-20
Add useful method

VinceAngel:
Useful method to manage quote replacement in paths


jsjolund:
Maybe you could provide a use case for the method? Also, your code is not correctly formatted:
https://github.com/libgdx/libgdx/wiki/Contributing#eclipse-formatter

VinceAngel:
This method is missing since it's in the jdk but not in gwt backend.

MobiDevelop:
Looks ok to me, one small optimization you could do is initialize the StringBuilder with a capacity at least the size of the string being passed in.

VinceAngel:
Found this, I forget it for a while... still ready to be merged :)

PR #4098, opened 2016-05-22
Add critical natives for some hot methods

intrigus:
Similar to https://github.com/libgdx/libgdx/pull/4021 but more explicit.

Adds support for critical natives, which reduces the overhead of jni-calls.
Critical natives are supported on Hotspot JVMs since Java7. See also [this answer on stackoverflow](http://stackoverflow.com/questions/36298111/is-it-possible-to-use-sun-misc-unsafe-to-call-c-functions-without-jni/36309652#36309652).

This PR shouldn't break any code.
Critical natives have been tested on Mac OS X 64bit, Linux 32 and 64bit, Windows 32 and 64bit.
# How to use
1. Ensure that your method can be used as a critical native. This means that it is **static**, **non-synchronized** and only has **primitives** or **primitives arrays** as arguments.
2. The method also can not use the normal `JNIEnv*` or  `jclass` arguments

Support for critical natives has to be _explicitly_ added.
So for example, if we have this method:

``` java
private static native void foo(int anInt, float[] anArray); /*
  foo(anInt, anArray);
*/
```

jnigen would create this:

``` c
JNIEXPORT void JNICALL Java_[package]_[class]_foo(JNIEnv* env, jclass clazz, jint anInt, jfloatArray obj_anArray)
{
  float* anArray = (float*)env->GetPrimitiveArrayCritical(obj_anArray, 0);
  foo(anInt, anArray);
  env->ReleasePrimitiveArrayCritical(obj_anArray, anArray, 0);
}
```
## **That's is the way it currently works.**

This PR adds support for generating critical natives.
To do so, we have to add a new (java) method:

``` java
private static native void foo_JNICritical(int anInt, int anArrayLength, float[] anArray); /*
  foo(anInt, anArray);
*/
```

this method is then used for generating a critical native, so it becomes this:

``` c
JNIEXPORT void JNICALL JavaCritical_[package]_[class]_foo(jint anInt, jint anArrayLength, jfloatArray anArray)
{
  foo(anInt, anArray);
}
```
## **What changed compared to the non critical version:**

Compare:
Normal version:

``` java
private static native void foo(int anInt, float[] anArray);
```

vs.
Critical version

``` java
private static native void foo_JNICritical(int anInt, int anArrayLength, float[] anArray);
```
1. A critical version has the same name plus `_JNICritical` to identify it as a critical native.
2. You **have to** add a new parameter of type `int` before every primitive array.
   The parameter contains the length of the primitive array. 
   **If you don't do this your critical native will probably _not_ be called!**
# Which functions benefits from it

Functions that do relatively little work in the native code benefit from it, since the jni overhead is greatly reduced.
That's why the various methods of `Matrix4` (`mul`, `rot`, ...),which all need to pass primitive arrays, are about 2 times faster.
Some methods of `BufferUtils` also gained support for critical natives, but I didn't benchmark how much they're faster and as is the case with `Matrix4` only methods that pass primitive arrays gained support.

### DIFF
--- a/extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/NativeCodeGenerator.java
+++ b/extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/NativeCodeGenerator.java
@@ -25,6 +25,7 @@ import com.badlogic.gdx.jnigen.parsing.CMethodParser.CMethod;
 import com.badlogic.gdx.jnigen.parsing.CMethodParser.CMethodParserResult;
 import com.badlogic.gdx.jnigen.parsing.JavaMethodParser;
 import com.badlogic.gdx.jnigen.parsing.JavaMethodParser.Argument;
+import com.badlogic.gdx.jnigen.parsing.JavaMethodParser.ArgumentType;
 import com.badlogic.gdx.jnigen.parsing.JavaMethodParser.JavaMethod;
 import com.badlogic.gdx.jnigen.parsing.JavaMethodParser.JavaSegment;
 import com.badlogic.gdx.jnigen.parsing.JavaMethodParser.JniSection;
@@ -183,6 +184,9 @@ public class NativeCodeGenerator {
 	JavaMethodParser javaMethodParser = new RobustJavaMethodParser();
 	CMethodParser cMethodParser = new JniHeaderCMethodParser();
 	CMethodParserResult cResult;
+	private ArrayList<JavaSegment> javaSegments;
+	private ArrayList<CMethod> cMethods;
+	private boolean logCriticalNativesGeneration;
 
 	/** Generates .h/.cpp files from the Java files found in "src/", with their .class files being in "bin/". The generated files
 	 * will be stored in "jni/". All paths are relative to the applications working directory.
@@ -257,12 +261,12 @@ public class NativeCodeGenerator {
 					}
 					String javaContent = file.readString();
 					if (javaContent.contains(JNI_METHOD_MARKER)) {
-						ArrayList<JavaSegment> javaSegments = javaMethodParser.parse(javaContent);
+						javaSegments = javaMethodParser.parse(javaContent);
 						if (javaSegments.size() == 0) {
 							System.out.println("Skipping '" + file + "', no JNI code found.");
 							continue;
 						}
-						System.out.print("Generating C/C++ for '" + file + "'...");
+						System.out.println("Generating C/C++ for '" + file + "'...");
 						generateHFile(file);
 						generateCppFile(javaSegments, hFile, cppFile);
 						System.out.println("done");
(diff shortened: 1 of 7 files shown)


extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
```diff
@@ -118,6 +118,19 @@ public String getArrayCType () {
 			if (this == DoubleArray) return "double*";
 			throw new RuntimeException("Unknown Array type " + this);
 		}
+		
+		public String getArrayJniType () {
```
intrigus: There is `getArrayCType()`, which is very similar, but the `jfoo`-types are there for a reason?

extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/parsing/JavaMethodParser.java
```diff
@@ -145,31 +158,49 @@ public String getName () {
 		public String toString () {
 			return "Argument [type=" + type + ", name=" + name + "]";
 		}
+
+		/** This only checks whether, the ArgumentType is the same, the name is ignored!
```
intrigus: This is not really ideal.
It's needed for checking whether a list of arguments is equal.
I could have created a method that compares a list of arguments instead of relying on `List#equals()`, but I'm not sure whether a method for this special kind of comparison (Only comparing the `ArgumentType`) would be better.

intrigus:
Added usage instructions and more information.

extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/NativeCodeGenerator.java
```diff
@@ -297,14 +301,41 @@ private void generateHFile (FileDescriptor file) throws Exception {
 	protected void emitHeaderInclude (StringBuffer buffer, String fileName) {
 		buffer.append("#include <" + fileName + ">\n");
 	}
+	
+	/** Emits a definition for desktop, so critical natives are only included in natives for desktop platforms */
+	private void emitDesktopDefinition (StringBuffer buffer) {
```
code-disaster: To be honest, I don't really like this block of definitions polluting every code generated. I'd prefer the alternative idea of the first PR, like having __LIBGDX_[_]DESKTOP__ defined by the initiator of the compilation.

intrigus: It is only generated for files which contain critical natives.
So I don't think it pollutes every code generated.

I'm also fine with passing `__LIBGDX__DESKTOP__` by the initiator.

PokeMMO:
LWJGL3 has currently disabled critical natives entirely due to GC issues. Adding explicitly opt-in support might be nice, but it's probably not worth the hassle atm.

Matrix methods were mostly moved back to java in #5883 making this a lot less necessary as well.

While a nice and neat addition, it probably won't get merged anytime soon.

Issue #3886, opened 2016-02-27
Framerate drops with gdx-controllers-lwjgl3 and NO connected controllers


saibotd:
#### Issue details

My game encounters extreme performance issues whenever I launch it without having any controllers attached. The framerate drops exactly from the point on where I call Controllers.getControllers() the first time.  
Game runs fine if at least one controller is attached or Controllers.getControllers()  never gets called.
#### Reproduction steps/code
- Enable LWJGL3
- Remove all gamepads from PC
- Call Controllers.getControllers() within you Application
#### Version of LibGDX and/or relevant dependencies

LibGDX 1.9.2 on Windows 10. Using a MS wireless receiver and Xbox 360 controller. It doesn't matter if the receiver is connected btw.
#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows 10
- [ ] Linux
- [ ] MacOS

intrigus:
Could you please define 'extreme performance issues'?

I don't have a controller so I can't test it.
The problem is probably that https://github.com/libgdx/libgdx/blob/master/extensions/gdx-controllers/gdx-controllers-lwjgl3/src/com/badlogic/gdx/controllers/lwjgl3/Lwjgl3ControllerManager.java#L33 takes a long time for enumerating all possible controllers (It's possible to have 16 controllers, iirc).
Since pollState() is called every frame we spend a long time enumerating all possible controllers.

GLFW is going to add connect and disconnect events, so we don't have to enumerate all possible controllers.

A possible fix would be to only check every second for new controllers, just like https://github.com/libgdx/libgdx/blob/master/extensions/gdx-controllers/gdx-controllers-desktop/src/com/badlogic/gdx/controllers/desktop/OisControllers.java#L61-L89 does for a lost window handle.

saibotd:
Thanks for taking the time. It goes from 60 fps to, I guess, 10-15 fps - I can measure it if you want to. I see no polling, etc. in the logs, but it sounds like that could be the problem. I checked with VisualVM (it helped me in the past with performance issues) and it shows an increased CPU usage (2% to 20%) when no controller is plugged in.

intrigus:
[...]

PR #4104, opened 2016-05-25
Update LWJGL3 to snapshot version #87

intrigus:
I updated LWJGL3 to snapshot version #87, so we don't have to fix so much, when it's officially released.
I also updated it, so we can fix https://github.com/libgdx/libgdx/issues/3886 since GLFW added controller connection and disconnection events.


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALAudio.java
```diff
@@ -87,14 +91,14 @@ public OpenALAudio (int simultaneousSources, int deviceBufferCount, int deviceBu
 		registerSound("mp3", Mp3.Sound.class);
 		registerMusic("mp3", Mp3.Music.class);
 		
-		device = ALDevice.create(null);
-		if(device == null) {
+		device = alcOpenDevice((ByteBuffer)null);
```
intrigus: See https://github.com/LWJGL/lwjgl3/issues/152#issuecomment-198144176

PR #4107, opened 2016-05-28
[WIP] Make scene2d groups generic

Pinpickle:
I wanted to subclass a `Group` that could only contain a certain kind of `Actor`. This would result in an unnecessary amount of casting from an Actor to a subclass. 

I thought about making my own implementation of group but that stops quite a few of the group parts of the API from working (e.g. I can no longer set `actor.parent`).

The logical conclusion of this is to make Groups generic.

Any implementation I can think of will lead to breaking BC in some way. At first I tried creating a `class GenericGroup<T extends Actor>` and then re-implemented Group to simply be:

```java
public class Group extends GenericGroup<Actor> { }
```

But that broke things in spectacular ways.

Changing `Group` to `Group<T extends Actor>` is far more stable and if things do break it's a little more obvious, so that's what this PR contains.

I'm still not sure that this is the best way to move forward or even if something like this is required. I wanted to see what other people thought about this.

## Reverse for Actors

The complement to this is to make an `ActorContainer` interface. The generic `Group` would implement this and then all types of Group can be replaced with this. This would allow for people to create the lighter implementations of groups and successfully set actor's parents.

This may be overkill, though.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
@@ -32,10 +32,10 @@ import com.badlogic.gdx.utils.SnapshotArray;
  * actors added earlier. Touch events that hit more than one actor are distributed to topmost actors first.
  * @author mzechner
  * @author Nathan Sweet */
-public class Group extends Actor implements Cullable {
+public class Group<T extends Actor> extends Actor implements Cullable {
 	static private final Vector2 tmp = new Vector2();
 
-	final SnapshotArray<Actor> children = new SnapshotArray(true, 4, Actor.class);
+	final SnapshotArray<T> children = new SnapshotArray<T>(true, 4);
 	private final Affine2 worldTransform = new Affine2();
 	private final Matrix4 computedTransform = new Matrix4();
 	private final Matrix4 oldTransform = new Matrix4();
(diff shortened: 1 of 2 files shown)


MobiDevelop:
I'm not sure this is something we want to do. I imagine it is a pretty specialized use-case to only want actors of a certain type to be contained in a group, and where casting is problematic.

Pinpickle:
> I imagine it is a pretty specialized use-case to only want actors of a certain type to be contained in a group, **and where casting is problematic**.

(emphasis mine)

I would argue that almost all cases of casting are problematic. If you are creating a group which should only contain a particular type of actor (which is useful as this means we can perform certain actions over the actors in this group), and you are casting to act on the actors it contains, you're adding a number of problems:

- You remove static type safety. Sure you're probably not going to add the wrong type of actor to this group but it would be nice to protect against it
- You're writing more code which gives you more room to make mistakes
- You are repeating yourself a lot (less DRY)
  - As an extension to this, JVM languages that make use of type inference lose those niceties

Having groups that only contain one kind of actor would be a fairly common use case I imagine. One of the [libgdx examples advertised for scene2d](https://github.com/alistairrutherford/libgdx-demos/tree/master/simple-shooter/core/src/main/java/com/netthreads/gdx/app/layer) makes heavy use of this (of course without actually restricting it with generics). Those "layer" objects don't really use any casting due to the way the game is designed, though.

NathanSweet:
Over the years I've had a couple situations where it would be nice to have Group use generics for the actors list. However, I'm not sure it happens often enough to warrant the change, and casting isn't really so bad. I think we can leave it without generics, at least for now.

PR #4130, opened 2016-06-07
Added access to application windows

kiture:
Now it's possible to hide window at start (using: setInitialVisible(false)) in LWJGL3, but later it's not possible to show it.


intrigus:
You say that it's not possible to show it later. 
But this works fine:
````java
Lwjgl3Window myWindow = Lwjgl3Application#newWindow(someListener, someConfig);
myWindow.setVisible(true);
````

kiture:
Yes, but using: Lwjgl3Application at start it's not possible. Lwjgl3Application do not give access to main window, created it this way. So that's why i added two methods to this class.

Tom-Ski:
Not sure whats gone on here?

PR #4131, opened 2016-06-07
Update LWJGL3 to release version 3.0.0 (build 90)

Zomby2D:
I also fixed a compile error in AwtTestLWJGL.java


Tom-Ski:
Thanks, I've updated the build.  Your changes to AwtTestLWJGL should not be required, looks like your IDE language compliance level is too low, it should be 1.8 for lwjgl3.

PR #4134, opened 2016-06-08
Implements properties types for tiled map loader

KyuBlade:
Added property cast for new tiled maps.
You can now use the methods :

- get (String key, Class<T> clazz)
- get (String key, T defaultValue, Class<T> clazz)

of the MapProperties class.
Support String, Boolean, Integer and Float types.


MobiDevelop:
Thanks for the contribution.

PR #4137, opened 2016-06-09
Add support for HTTP PATCH

pjot:
Setting the method to "PATCH" on a request causes it to be sent without a body but still with the PATCH verb.

This PR makes PATCH a proper HTTP verb in libgdx. The alternative would be to make the HTTP classes throw an exception when unsupported verbs are used instead of failing silently like it does now.


intrigus:
This does not work on GWT.
See https://github.com/libgdx/libgdx/blob/c7aaa1ef3c4745182d19776136acc77c05b96688/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java

PR #4140, opened 2016-06-10
Added support for typed TMX-properties.

wcmatthysen:
Modified the loadProperties() method by adding support for detecting the
types of the different properties. The type-information is stored as an
additional attribute with the name "type" in the property-element. If
present, this type-information is then used to convert the value-string
to the correct type (i.e. Boolean, Color, Float, FileHandle or Integer).


wcmatthysen:
Not sure if Gdx.files.internal() is the correct way of doing the file-handling part.

MobiDevelop:
We now have two PRs open for the same thing. See #4134.

wcmatthysen:
My bad. Sorry, didn't see that issue. I think the other pull request is more complete. I'll close this one.

PR #4146, opened 2016-06-11
Create AmazonFireGameController.java and fix an error in Json-Util

B4dT0bi:
Just a simple mapping of the Amazon Fire Game Controller (Amazon FireTV) and a fix of Json Util which got a SerializationException when trying to DeSerialize an ArrayList inside a TreeMap. (see JUnitTest)


Darkyenus:
Please don't create PR's with multiple independent changes, it is harder to review and a bit messy.

Also, could you elaborate more on the json fix? What did happen and what now happens?

(Please don't get discouraged by this :) )

B4dT0bi:
Ok, I was not familiar with PRs so I didnt know that if I still commit in my master-branch the PR will get updated, lesson learned ;)

Should I split it or is it ok for now?
I will make sure future PRs will come from new branches so there will be only one change in one PR.

Fix1 : when using a class which contains a Map<String, List<String>> the class name of the List (eg. java.util.ArrayList) is stored in the json Map definition so Json is able to recreate the correct List, but this class type is then also used as an element for the Map. So it tries to insert a String "java.util.ArrayList" as an element in the Map which then fails because it assumes a List.
This can be checked with the JUnitTests form the PR.

Fix2 : the type of key for a specific Map is ignored and always assumed to be String.
So if you have a Map<Integer, String> and the key is always assumed to be String then the hashvalues from Integer and String doesnt match so the values cannot be accessed.

Example:
key value pair 1 = "one" doesnt match "1" = "one"
Somehow the HashMap allows a String as Key even if the Map is declared as Map<Integer, String> and this leads to a different hash then the same value of the Integer. 

Both fixes make it easier to use Object->Json->Object conversions without the use of special Serializer.

Tom-Ski:
Yes, please split this PR.  Also, what backend did you test the mapping on?

PR #4147, opened 2016-06-12
Added the setInputProcessor.

LoveraSantiago:
Added the setInputProcessor. 
In the current class and the parent class, was not set neither in the input processor.

I do not know if the ApplyForce class is extended in some other example and that other example there is the set for inputprocessor. If ApplyForce really is a complete example then missed this set.


Tom-Ski:
It is part of the Box2DTestCollection (https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/Box2DTestCollection.java), which is responsible for setting the input processor.

PR #4153, opened 2016-06-16
private and public methods sorted by name to simplify syncing with Playn HtmlGl20

stefanhaustein:
Both variants have some independent improvements that may make sense to carry over.
Sorting simplifies the comparison.


stefanhaustein:
ping?

badlogic:
Pong :)

intrigus:
Sorting the glXyz methods makes sense, but sorting the special methods (ensureCapacity(), allocateUniform(), ... ) doesn't imo.

stefanhaustein:
I'll delete them in a followup

Issue #4111, opened 2016-05-29
Request gdx-jnigen


xpenatan:
Request a way to be able to add a custom tag so jniGen ignores it when generating c++ code.

Example in RobustJavaMethodParser.java:
```Java
private ArrayList<JniSection> getNativeCodeBodies (String classFile) {
	ArrayList<JniSection> sections = getComments(classFile);
	Iterator<JniSection> iter = sections.iterator();
	while (iter.hasNext()) {
		JniSection section = iter.next();
		if (section.getNativeCode().startsWith("JNI")) iter.remove();
		if (section.getNativeCode().startsWith("-{")) iter.remove();
		if (section.getNativeCode().startsWith(MyCustomtTAG)) iter.remove();
	}
	return sections;
}
```

#### Please select the affected platforms
- [X] Android
- [X] iOS
- [X] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

PR #4156, opened 2016-06-19
Custom ignore tag

xpenatan:
#4111


Darkyenus:
Please use standard Java naming conventions (field names should begin in lowercase)

xpenatan:
Interesting, so static and non static share the same rule?  o_O

PR #4165, opened 2016-06-23
Add resizable/undecorated options to LWJGL

bgroenks96:
- Add resizable/undecorated options to LWJGL config.
- Add 'forceDisplayModeChange' flag in LwjglGraphics to force 'setWindowedMode' to reconfigure the Display after a call to 'setUndecorated'

See #3956 for past submission.

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglApplicationConfiguration.java
@@ -70,6 +70,8 @@ public class LwjglApplicationConfiguration {
 	public boolean forceExit = true;
 	/** whether the window is resizable **/
 	public boolean resizable = true;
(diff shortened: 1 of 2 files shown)


crykn:
Both things that this PR adds work. The second change may need some explaining:

- Changing the display mode is needed to apply the undecorated setting. 
- At the moment, this can be achieved by calling `Gdx.graphics.setWindowedMode(...)` with a window size _different_ than the current.
- In contrast, with this PR, `Gdx.graphics.setWindowedMode(...)` can be called with the same size the window currently has (because the display mode change is forced anyway)

I'm not sure, whether this second change is all that useful: Since undecorated windows with lwjgl2 look like the following (at least on my machine), I don't think anybody will use the undecorated setting except when going to borderless fullscreen. And in that case, `Gdx.graphics.setWindowedMode(...)` has to be called with a different size anyway.

Nonetheless, both additions in this PR do their job, so I still think we should merge it.

lwgl_window

crykn (approved):

PR #4177, opened 2016-06-30
Added support to window icon on LWJGL3 backend

viniciuspjardim:
Now that GLFW has the glfwSetWindowIcon() method, I added support to window icon on LWJGL3 backend.
I tried to be as close as possible to LWJGL backend implementation.
I've tested with icons on png, jpg and bmp formats. But only on Windows.
I am not sure if the best place to put the GLFW.glfwSetWindowIcon() call is in the Lwjgl3Graphics constructor. Any suggestion?

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.backends.lwjgl3;
 import java.nio.IntBuffer;
 
 import com.badlogic.gdx.Application;
+import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.glutils.GLVersion;
 import org.lwjgl.BufferUtils;
 import org.lwjgl.PointerBuffer;
@@ -33,6 +34,8 @@ import com.badlogic.gdx.graphics.GL20;
 import com.badlogic.gdx.graphics.GL30;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.utils.Disposable;
+
+import org.lwjgl.glfw.GLFWImage;
 import org.lwjgl.opengl.GL11;
 
 public class Lwjgl3Graphics implements Graphics, Disposable {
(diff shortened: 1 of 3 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
```diff
@@ -82,6 +88,37 @@ public Lwjgl3Graphics(Lwjgl3Window window) {
...
+			// Array with pixmaps to be disposed
+			Pixmap[] pixmaps = new Pixmap[config.iconPaths.size];
+
+			for (int i = 0; i < config.iconPaths.size; i++) {
+				Pixmap pixmap = new Pixmap(Gdx.files.getFileHandle(config.iconPaths.get(i), config.iconFileTypes.get(i)));
+
+				if (pixmap.getFormat() != Pixmap.Format.RGBA8888) {
+					Pixmap rgba = new Pixmap(pixmap.getWidth(), pixmap.getHeight(), Pixmap.Format.RGBA8888);
+					rgba.drawPixmap(pixmap, 0, 0);
+					pixmap = rgba;
```
Tom-Ski: The original `pixmap` memory is leaked here

cypherdare:
Shouldn't the icon be applied in `Lwjgl3Application.newWindow()`, not in Graphics?

I think this would be more flexible and easier to use if the application configuration took Pixmaps instead of file names and file types. It would also match the precedent used for setting cursors.

There are a number of GLFW features that we currently only support at Window creation time--but potentially could also be applied to existing windows if we added methods to the Lwjgl3Window class. Not sure if anyone has need of changing a window icon on an existing window.

WickedShell:
Just chiming in that the lack of window icons is holding me up from completing the transition to the LWJGL3 backend.

Anecdotal - I don't have any need for changing icons at runtime, just titles.

cypherdare:
See #4212 for an alternate implementation. I took the basic algorithm from this, but moved it to where we can set different icons in different windows as well by using Lwjgl3WindowConfiguration. I also simplified the setup so you don't need to make multiple calls for all your icons. And I fixed a couple of potential bugs.

I take back what I said about Pixmaps. They are clunky to instantiate before the application is initialized.

viniciuspjardim:
I will close this. I think #4213 sounds better. Thanks for the replies.

Issue #4114, opened 2016-05-31
MultiTouch behaving incorrectly with MOE backend


majiddaeinejad:
#### Version of LibGDX

1.9.3
#### Version of Multi-OS Engine

1.0.648
#### Affected platforms

iOS (MOE)
#### Tested Devices

iPhone 5s and iPhone 6s

The issue is about touching with two fingers or more. The InputListener doesn't work correctly in iOS version after migration to Multi-OS engine. The 'touchup' for one of fingers has not been called in iOS version if display touched with two fingers or more together simultaneously, only 'touchdown' occured.
Simply you can test it by this code:

```
Gdx.input.setInputProcessor(new InputProcessor() {
            @Override
            public boolean keyDown(int keycode) {
                return false;
            }

            @Override
            public boolean keyUp(int keycode) {
                return false;
            }

            @Override
            public boolean keyTyped(char character) {
                return false;
            }

            @Override
            public boolean touchDown(int screenX, int screenY, int pointer, int button) {
                Gdx.app.log("touchdown", "touchdown");
                return true;
            }

            @Override
            public boolean touchUp(int screenX, int screenY, int pointer, int button) {
                Gdx.app.log("touchup", "touchup");
                return true;
            }

            @Override
            public boolean touchDragged(int screenX, int screenY, int pointer) {
                return false;
            }

            @Override
            public boolean mouseMoved(int screenX, int screenY) {
                return false;
            }

            @Override
            public boolean scrolled(int amount) {
                return false;
            }
        });

```

Tom-Ski:
Please fill in the Issue template you are shown before submitting an issue.  What version of LibGDX are you using? What version of MOE are you using? What device are you testing on?

majiddaeinejad:
@Tom-Ski I have edited this issue according to the issue template.

Tom-Ski:
Thanks, Ill have a look soon.  Looks like its an issue with touch up events, or maybe multi touch interfering somehow.

vbousquet:
I can confirm this bug since I have the same. I've ported my game to both RoboVM (MobiDevelop branch) and MOE and the bug only appears for the MOE branch. It happens for single touch as well as for multi touch.

vbousquet:
After investigation, there seems to be a bug in the MOE version of IOSInput which drops event.
The following extract shows that it use ```break``` instead of ```continue``` dropping the following events if any. It will happen for fast click (my case) or for multitouch (original poster case).

```

	void processEvents () {
[...]

PR #4181, opened 2016-07-02
iOS MOE Input fix

vbousquet:
This fix issue #4114 by removing the 'break' instruction that would discard multiple event processing.


Tom-Ski:
Thanks @vbousquet !

PR #4194, opened 2016-07-17
fix android multi-touch bug

yangweigbh:
in android multi touch, sometimes a new pointer-down event of
one pointer id come without corresponding pointer-up,
and when action_cancel event comes, all the previous pointer ids should be invalidated


Tom-Ski:
Do you have a test sample that shows the problem that this is trying to resolve? Invalidating all pointers on cancel could cause some unwanted behavior.

yangweigbh:
this fix want to fix the problem found when our tester quickly touch down and up the screen using four fingers. pointer down event of some pointer id without pointer up. finally find that the pointer down event company the cancel, and when cancel happens, all the pointer info in the android touch event is disposed, you can dump that info with event.getPointerCount() and see the value turn to 0 after ACTION_CANCEL

Tom-Ski:
I can't reproduce your issue, what device is this a problem with?

PR #4204, opened 2016-07-24
Change gl handle to webgl object hash maps to an efficient jso int map

stefanhaustein:
Change gl handle to webgl object hash maps to an efficient jso int mapp (based on sparse js arrays) tailored towards this purpose. Remove unneeded helpers


Tom-Ski:
Thanks!

PR #4212, opened 2016-07-31
Allow custom window icons in Lwjgl3

cypherdare:
This is an alternate to #4177.  I fixed some bugs it had, simplified usage, and added the ability to set icons in the window configuration, so different windows can have different icons.


cypherdare:
I messed up something with amending this, starting over.

PR #4215, opened 2016-07-31
Add respect-min-size functionality to SplitPane

cypherdare:
I've been using this in my project for a while and it has saved me a lot of time. This lets you set the SplitPane to respect the min size of the widgets in it, so the handle cannot shrink either inner widget to the point of cropping it. This is a lot easier than trying to manually calculate the split amount limits and also allows it to layout properly when you want this functionality.

I considered using prefSizes instead, but I think going down that path would make it ambiguous in usage. What I mean is, SplitPane crops widgets that are drawn smaller than their min sizes, so that's what we want to prevent. You can wrap it in a container if it doesn't already have a min size, or you want to use the pref size as a min size. Since the inner widgets are often Tables, this is usually not a concern anyway.

As far as I know, the only place where this might modify existing projects is where I started counting the min size of the handle drawable in the `getMinWidth` and `getMinHeight` methods. I think this is more technically correct since you shouldn't be able to draw a split pane so small that you can't see the handle.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,8 @@
 - API addition: Lwjgl3WindowListener.created() is called after a new window has been created. It's unsafe to call Lwjgl3Window functions in between Lwjgl3Application.newWindow() and this callback.
 - Updated LWJGL3 backend to 3.1.1. Lwjgl3Graphics.setUndecorated() and Lwjgl3Graphics.setResizable() now delegate their work to the respective GLFW functions.
 - API addition: ProgressBar.isVertical() - returns whether a progress bar is vertical or horizontal.
+- API Change: SplitPane now by default does not allow the split amount to shrink children below their minimum sizes (cropping them). This behavior can be reverted by overriding clampSplitAmount or wrapping the children in Containers set to minSize(0) and fill(). SplitPane also now correctly includes the handle min size in its own min size calculations.
+- API Change: SplitPane.getSplit() renamed to SplitPane.getSplitAmount() to match other getter and setter names.
 
 [1.9.6]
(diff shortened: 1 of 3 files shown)


NathanSweet:
TBH, I have never used SplitPane in my projects. Do we need `respectXxxMinSize` fields for configuration? I try to avoid classes getting too configurable, as it causes bloat and makes the API harder to use (see poor ScrollPane, which is nearly out of control). If we can make sane default functionality, I'd prefer that over customizability, especially if users can override a method and return zero (or whatnot) to customize the behavior.

cypherdare:
I see what you mean. I had cloned the class to add this feature in my
project. I'll look into what changes, if any, are needed to make it
customizable enough to do with a subclass.

NathanSweet:
I'm ok with any of these:

1) We adjust SplitPane so a subclass can customize the behavior.
2) We adjust SplitPane so the new default behavior always respects min size. People can use a container with min size of 0 to get old behavior.
3) You say the configurability is worthwhile and we merge your PR as is.

I have a feeling #2 is the way to go.

cypherdare:
I agree with 2. It seems to me like what would typically be expected, and more like how analogous widgets in Windows/Mac behave.

I made the `clampSplitAmount` protected. If someone has been extensively relying on the original behavior, they can override the method one time rather than wrapping everything in Containers.

Also fixed a mismatched getter name, as long as I'm breaking things. :)

cypherdare:
I updated this to also fix #4642.

NathanSweet:
Can you fix up this PR's merge conflicts so we can merge? Thank you!

cypherdare:
@NathanSweet 
Done

PR #4218, opened 2016-08-01
Allow explicit TextField focus traversal.

cypherdare:
For when you have separate columns of text fields, or a group that wants to capture traversal and loop around.


NathanSweet:
Would this be better solved by overriding `next` rather than making it configurable?

cypherdare:
This is a convenience feature to avoid littering your code with lots of
anonymous classes to do that. I personally subclass TextField once with
this feature and use that everywhere. The setMutual parameter also cuts the
configuration in half.

I see your point...overriding is potentially more flexible and keeps the
class cleaner.

PR #4219, opened 2016-08-01
Always respect explicit row() calls (fixes #4183)

### LINKED ISSUE #4183: Scene2d Table Not expected behavior
#### Issue details
When i create a Table after that i call row() and then i add an item is expected that that this item is in the next line. 

In Libgdx version 1.6.5 i create a TextButton with some text then i called row() to add a new label to this button. This label will be in a new line but after libgdx version 1.7.0 this label will be in the right side of the previous label. 

This image was taken using 1.6.5.
screen shot 2016-07-08 at 9 23 58 pm

And this one using the latest release 1.9.3.
screen shot 2016-07-08 at 9 26 00 pm

#### Reproduction steps/code
I created a simple code. Using the default skin. 

```java
public class TableBug extends ApplicationAdapter {
    
    public Skin skin;
    Stage stage;
    Table container;


    @Override
    public void create() {
        skin = new Skin(Gdx.files.internal("UI/uiskin.json"));

        container = new Table();
        container.setFillParent(true);


        stage = new Stage();
        stage.addActor(container);

    }

    @Override
    public void render() {
        Gdx.gl.glClearColor(1, 0, 0, 1);

        stage.act();
        stage.draw();

        if (Gdx.input.justTouched()) {
            TextButton newButton = new TextButton("Button ", skin);
            container.row();
            container.add(newButton);
        }
    }
}
```

version 1.7.0 and up

- [ yes] Android
- [ yes] iOS (robovm)
- [ yes] iOS (MOE)
- [ yes] HTML/GWT
- [ yes] Windows
- [ yes] Linux
- [ yes] MacOS

cypherdare:
If the user calls row() it can be assumed the user definitely wants a new row. #4183 describes a case where this is needed.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
@@ -341,6 +341,7 @@ public class Table extends WidgetGroup {
 	/** Indicates that subsequent cells should be added to a new row and returns the cell values that will be used as the defaults
 	 * for all cells in the new row. */
 	public Cell row () {
+		implicitEndRow = false;
 		if (cells.size > 0) {
 			endRow();
 			invalidate();


NathanSweet:
Ooo, a table bug. The fix is close, but `row` needs to avoid calling `endRow` when `implicitEndRow` has already called it, else the row count increases too many times.

PR #4223, opened 2016-08-02
Small fix to error message generated by XmlReader.Element.getAttribute(String name)

byron-conroy:
The error message generated by this function when the attribute did not exist was was of the form
"Element $attribute doesn't have attribute: $attribute"
instead of
"Element $element doesn't have attribute: $attribute"


MobiDevelop:
That file is generated, you should edit the XmlReader.rl file (and regenerate the XmlReader.java file).

PR #4229, opened 2016-08-05
Add HitMode to Group to allow choosing the hit detection strategy

mclem:
When multiple actors are added to a group, the one that gets returned by the hit method is the first one in the children hierarchy.

I have a use case where the actor that should be returned when the screen is touched is the one nearest to the hit coordinates (when two actors are partially superimposed, the current logic where the topmost actor is always returned feels unnatural when playing the game).

This pull request adds a HitMode option on the the Group class, that defaults to the current implementation and can be changed to a "nearest actor" mode.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
@@ -23,6 +23,7 @@ import com.badlogic.gdx.math.Matrix4;
 import com.badlogic.gdx.math.Rectangle;
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.scenes.scene2d.utils.Cullable;
+import com.badlogic.gdx.utils.Align;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.SnapshotArray;
 
@@ -33,6 +34,15 @@ import com.badlogic.gdx.utils.SnapshotArray;
  * @author mzechner
  * @author Nathan Sweet */
 public class Group extends Actor implements Cullable {
+
+	/** Determines how to choose the child that has been hit in a group. */
+	public enum HitMode {
+		/** Return the first child in the hierarchy that has been hit. */
+		first,
+		/** Return the child that is nearest to the hit coordinates. */
+		nearest
+	}
+
 	static private final Vector2 tmp = new Vector2();
 
 	final SnapshotArray<Actor> children = new SnapshotArray(true, 4, Actor.class);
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
```diff
@@ -237,17 +248,42 @@ public Rectangle getCullingArea () {
...
+		Actor hitActor = null;
 		Actor[] childrenArray = children.items;
 		for (int i = children.size - 1; i >= 0; i--) {
 			Actor child = childrenArray[i];
 			if (!child.isVisible()) continue;
 			child.parentToLocalCoordinates(point.set(x, y));
 			Actor hit = child.hit(point.x, point.y, touchable);
-			if (hit != null) return hit;
+			if (hit != null) {
+				switch (hitMode) {
```
cypherdare: Should the switch be outside the loop? There could be hundreds of children to check. Although maybe the difference is negligible due to branch prediction. Maybe make a local `final` reference to the setting and use that.

gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
```diff
@@ -33,6 +34,15 @@
  * @author mzechner
  * @author Nathan Sweet */
 public class Group extends Actor implements Cullable {
+
+	/** Determines how to choose the child that has been hit in a group. */
+	public enum HitMode {
+		/** Return the first child in the hierarchy that has been hit. */
+		first,
```
cypherdare: Libgdx's standard with Enums is camel case (not sure why we don't follow the standard and use all caps).

cypherdare:
Why not subclass Group rather than burden every WidgetGroup subclass?

gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java
```diff
@@ -237,17 +248,42 @@ public Rectangle getCullingArea () {
...
 			if (!child.isVisible()) continue;
 			child.parentToLocalCoordinates(point.set(x, y));
 			Actor hit = child.hit(point.x, point.y, touchable);
-			if (hit != null) return hit;
+			if (hit != null) {
+				switch (hitMode) {
+				case first:
+					return hit;
+				case nearest:
+					float dst = point.set(child.getX(Align.center), child.getY(Align.center)).dst(x, y);
```
cypherdare: Might as well use `dst2`.

NathanSweet:
There could be any number of ways you might want to handle hit detection (nearest, circle, ellipse, pixel perfect, etc) and subclasses can override `hit` to customize that. I don't think the base class needs modes, plus an enum is not extensible.

Issue #3782, opened 2016-01-26
Loading images on GWT is lossy (particularly visible when using premultiplied alpha)


ttencate:
This is a follow-up to #3316, which was closed, I guess partly due to lack of information. However, I did some more digging and found the root cause of the alpha banding problems in GWT. The summary is: storing pixel data in `Pixmap` is lossy on GWT, and should be avoided during `Texture` loading when possible.

A JSFiddle to demonstrate the issue: https://jsfiddle.net/gg9tbejf/

In my [answer to the corresponding StackOverflow post](http://stackoverflow.com/a/35015581/14637), I have described a workaround. Copying here for convenience:

```
public class GwtTextureLoader extends AsynchronousAssetLoader<Texture, TextureLoader.TextureParameter> {
    TextureData data;
    Texture texture;

    public GwtTextureLoader(FileHandleResolver resolver) {
        super(resolver);
    }

    @Override
    public void loadAsync(AssetManager manager, String fileName, FileHandle fileHandle, TextureLoader.TextureParameter parameter) {
        if (parameter == null || parameter.textureData == null) {
            Pixmap.Format format = null;
            boolean genMipMaps = false;
            texture = null;

            if (parameter != null) {
                format = parameter.format;
                genMipMaps = parameter.genMipMaps;
                texture = parameter.texture;
            }

            // Mostly these few lines changed w.r.t. TextureLoader:
            GwtFileHandle gwtFileHandle = (GwtFileHandle) fileHandle;
            ImageElement imageElement = gwtFileHandle.preloader.images.get(fileHandle.path());
            data = new ImageTextureData(imageElement, format, genMipMaps);
        } else {
            data = parameter.textureData;
            if (!data.isPrepared()) data.prepare();
            texture = parameter.texture;
        }
    }

    @Override
    public Texture loadSync(AssetManager manager, String fileName, FileHandle fileHandle, TextureLoader.TextureParameter parameter) {
        Texture texture = this.texture;
        if (texture != null) {
            texture.load(data);
        } else {
            texture = new Texture(data);
        }
        if (parameter != null) {
            texture.setFilter(parameter.minFilter, parameter.magFilter);
            texture.setWrap(parameter.wrapU, parameter.wrapV);
        }
        return texture;
    }

    @Override
    public Array<AssetDescriptor> getDependencies(String fileName, FileHandle fileHandle, TextureLoader.TextureParameter parameter) {
        return null;
    }
}
```

```
[...]

Issue #3316, opened 2015-07-29
WebGL banding on smooth alpha gradients


WillCalderwood:
I've got a LibGDX game with cartoon clouds with a smooth gradient. There are other examples of gradients in the game that have a similar issue, but the clouds are the most obvious example. They look fine in Android, on iOS and on the Desktop version of the game, but on the WebGL version the gradients are not drawn as smooth. It only appears to be alpha gradients that have the problem. Other gradients look ok.

I've used a custom shader to pin down the cause of the banding, and the banding occurs in the colour channels, not in the alpha, but only occurs on textures with an alpha gradient. I'm wondering if it's some kind of precision issue with alpha premultiplication, as the alpha channel itself isn't effected. But if that's the case I don't understand why it's only occurring in WebGL and why the brightness cycles from light to dark across the bands rather than just being a flat colour in the bands. If it was a precision issue within the texture in VRAM then that also wouldn't explain why the problem disappears when using `GL20.GL_SRC_ALPHA, GL20.GL_ONE_MINUS_SRC_ALPHA` as the blend function when still using the premultiplied png.

I've tried on 3 different devices in Chrome and IE, and all 3 produce the same results. You can find a test of the HTML5 version here.

https://wordbuzzhtml5.appspot.com/canvas/

I've added an example IntelliJ project on github here

https://github.com/WillCalderwood/CloudTest

The critical code is `render()` [here](https://github.com/WillCalderwood/CloudTest/blob/master/core/src/com/mygdx/game/MyGdxGame.java)

The bottom image here is the desktop version, the top is the WebGL version, both running on the same hardware.

![enter image description here][1]

There's nothing clever going on with the drawing. It's just a call to

        spriteBatch.draw(texture, getLeft(), getBottom(), getWidth(), getHeight());

I'm using the default shader, textures packed with premultiplied alpha with the blend function set as

        spriteBatch.setBlendFunction(GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_ALPHA);

This is the actual image, although alpha not premultiplied as that's done by my packer.

![enter image description here][2]

Does anyone know a possible reason for this and how I might resolve it?

This only appears to happen when using the blending mode `GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_ALPHA`

[...]

PR #4234, opened 2016-08-07
Fixes problems with lossy image data when using premultiplied alpha on GWT for textures

barkholt:
Fixes problem reported in #3782 and #3316.

Texture loading in the GWT backend makes use of the Pixmap class (as in the other backends). The GWT Pixmap is backed by a JavaScript Canvas element in order to make it mutable, as required by the Pixmap class. This is a problem since the round-trip to the Canvas in certain cases is lossy (as documented in the official specs), creating visual artifacts when using the alpha channel (see referenced tickets and supplied test case).

The PR proposes to delay the creation of the Canvas until absolutely necessary. In the typical texture loading case, there is no need for this extra representation, and we can just use the already created ImageElement when uploading the texture using OpenGL.

External users of the Pixmap class should be unaffected


Tom-Ski:
Not a huge fan of the lazy loading of the canvas.   What about a private constructor path to an ImageElement only Pixmap? TextureData is emulated, so a custom constructor for Pixmap could be used that skips the canvas when uploading for a Texture.

WillCalderwood:
@barkholt Any response to @Tom-Ski ? I'd love to see this fixed as I've just hit problems again.

@Tom-Ski Why do you not like the lazy loading of the canvas? You don't like the existance checks?

Thanks

Tom-Ski:
The existance checks aren't the worse, I'd just prefer clean routes.  This is OK though, can be merged as a fix for now, just needs testing. I can do that this weekend, unless someone beats me to it.

barkholt:
@Tom-Ski I'd love if you could do the testing and merging, or alternative throw the solution out definitively. I have several projects that suffer due to this issue. 
There might be a better way, and would love to investigate it at some point. In the mean time this solution is designed to minimize the changes in the back-end, and avoid creating side effects.

PR #4238, opened 2016-08-08
Fix json validity and missing section in the uiskin.json

ncareau:
I edited both uiskin.json present in the repo to be valid JSON. Double quotes everywhere and remove trailing comma.  

You can see that even github had trouble parsing it : https://github.com/libgdx/libgdx/blob/055b9f762e90303d3d5367b2e974bbaf595a7c64/extensions/gdx-tools/assets/uiskin.json

Also added the tooltip section to the gdx-tools/assets/uiskin.json file.

### DIFF
--- a/extensions/gdx-tools/assets/uiskin.json
+++ b/extensions/gdx-tools/assets/uiskin.json
@@ -1,59 +1,167 @@
 {
-com.badlogic.gdx.graphics.g2d.BitmapFont: { default-font: { file: default.fnt } },
-com.badlogic.gdx.graphics.Color: {
-	green: { a: 1, b: 0, g: 1, r: 0 },
-	white: { a: 1, b: 1, g: 1, r: 1 },
-	red: { a: 1, b: 0, g: 0, r: 1 },
-	black: { a: 1, b: 0, g: 0, r: 0 }
-},
-com.badlogic.gdx.scenes.scene2d.ui.Skin$TintedDrawable: {
-	dialogDim: { name: white, color: { r: 0, g: 0, b: 0, a: 0.45 } }
-},
-com.badlogic.gdx.scenes.scene2d.ui.Button$ButtonStyle: {
-	default: { down: default-round-down, up: default-round },
-	toggle: { down: default-round-down, checked: default-round-down, up: default-round }
-},
-com.badlogic.gdx.scenes.scene2d.ui.TextButton$TextButtonStyle: {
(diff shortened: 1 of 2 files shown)


Darkyenus:
libGDX uses custom lightweight JSON-like format so it is not wrong as it is. If we want to use the standard format is a different question though and depends on the original reason to use the current, simplified, format.

ncareau:
I understand, Is there somewhere i could read up on the format used ?

However, wouldn't it be better if the format was JSON valid ? Both IDE and custom script could be used to read and modify the file. (ex, i could use custom python or whatever else to generate a new file).

Darkyenus:
> [The Json class can write and read both JSON and a couple JSON-like formats. It supports "JavaScript", where the object property names are only quoted when needed. It also supports a "minimal" format (the default), where both object property names and values are only quoted when needed.](https://github.com/libgdx/libgdx/wiki/Reading-%26-writing-JSON)

I agree that the minimal format may not be necessary, but it is already widely used. In any case, only core contributors can correctly decide this, as I am not familiar with all reasons for introducing this format in the first place.

ncareau:
Thanks for the comment. 

I will just add that for me (and probably a lot of new libgdx users) having a valid JSON from the libgdx repo would be easier to pickup the scene2d UI stuff. A non valid JSON is prone to errors since the IDE cannot check for validity and no code completion is available. 
Just my 2c.

obigu:
Same as @ncareau I also find this a bit confusing as my IDE also highlights syntax errors on uiskin.json. 

I can see the potential benefit of using a lightweight custom serialization format but if it's not valid JSON  then it's not JSON (even if it's similar) and it should probably not have the .json extension.

Scellow:
I agree with everyone the current JSON format used makes it impossible to correctly manipulate it from an IDE, or JSON editor/linter

The skin file is only loaded once, i wonder why the team decided to use a non standard format..

MobiDevelop:
You are, of course, welcome to format your own skin files any way you like. If proper json is your desire, by all means use it. The test uiskin is a test asset, for some reason people think of it as a "default" skin.

ncareau:
> Skin files from the libgdx tests can be used as a starting point. 
From the Wiki/Skin page. And google points to these files when I search for "default skin.json".

Would a possible solution be that we add a default skin in the library or wiki ?

cypherdare:
@Tom-Ski I'm pretty sure this can be closed, as there is no bug, and LibGDX accepts valid JSON if you want to use the IDE's JSON review.

Issue #3659, opened 2015-12-21
Add the ability to clear all pools in Pools


asherbar:
I think it would be useful to add the ability to clear all pools that are stored internally in [Pools](https://github.com/libgdx/libgdx/blob/fde893cec5448fdd09e778d1c103aac474950873/gdx/src/com/badlogic/gdx/utils/Pools.java) via the `static` variable `typePools`. 
The implementation seems to be rather straightforward: Simply iterate over all the `pools` and call their `clear` method. Something like this: 

``` java
public class Pools {
    // ...
    public static void clear() {
        for (ObjectMap.Entry<Class, Pool> entry : typePools)  {
            entry.value.clear();
        }
    }
}
```

xoppa:
Please include an example of a practical use-case where this would be useful.

asherbar:
Consider the following use case:
On Android, due to the App's lifecycle, `static` references are kept alive when the App is hidden. So say we use `Pools` to obtain objects which reference a [Disposable](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/utils/Disposable.html) resource. This means that indirectly, these resources are statically referenced, because they are kept inside the `static` variable `typePools`.

It is considered (or so I understand) good practice to dispose resources in [ApplicationListener's](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/ApplicationListener.html) `dispose`. So if I do that, this means that after returning to the game, i.e., after `disposed` has been called, recycled objects that were kept alive due to the `static` referencing, are now referencing `disposed` resources. When I obtain these objects from `Pools`, I get unexpected behaviour.

With the above request, this problem can be easily be solved by clearing all pools in `Pools`.

xGnoSiSx:
I didn't quite understood the use case asherbar, but:

this would save memory while increasing app switching speed since you have to recreate the pools once you switch back to the application. I'd say to add the ability, but without testing in each different app for performance impact there might be trouble.

asherbar:
@xGnoSiSx I agree the use case I wrote is not too simple, but roughly, any use case you can imagine that would require [clearing a Pool](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/utils/Pool.html#clear--), can be applied to this proposed feature as well.

> ... but without testing in each different app for performance impact there might be trouble

What sort of trouble do you have in mind?

cypherdare:
[...]

PR #4249, opened 2016-08-15
Fixed issue 3659

likeanowl:
Related to #3659 issue.
Added clear() method.


cypherdare:
This doesn't actually fix the only use case that the referenced issue mentions, because all the Disposables would be leaked.

likeanowl:
Thank you, @cypherdare , for your response and noticing me being wrong. I'm just recently began using libgdx, so doesn't understand whole code yet, but rly want to make it better and do everything I can.
I'll close this request and probably reopen it a bit later.

PR #4261, opened 2016-08-20
Allow custom window icons in Lwjgl3

cypherdare:
You can set up your icons as image files and pass them into the app configuration. It takes image filepaths because Pixmaps are difficult to load from files before LibGDX is initialized.

You can also pass Pixmaps to an existing Lwjgl3Window to change the icon dynamically.

And I also added a method to change the title of an existing Lwjgl3Window.

Finally, I fixed a GLFW window size limits bug with a work-around, since I was in the area.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,5 @@
 [1.9.5]
-- Allow window icons to be set in Lwjgl3ApplicationConfiguration or LwjglWindowConfiguration.
+- Allow window icons to be set in Lwjgl3ApplicationConfiguration or Lwjgl3WindowConfiguration.
 - Allow window icon and title to be changed in Lwjgl3Window
 - API Addition: ApplicationLogger interface, allowing easier access to custom logging
 - DefaultRenderableSorter accounts for center of Renderable mesh, see https://github.com/libgdx/libgdx/pull/4319
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,6 @@
 [1.9.5]
+- Allow window icons to be set in Lwjgl3ApplicationConfiguration or LwjglWindowConfiguration.
+- Allow window icon and title to be changed in Lwjgl3Window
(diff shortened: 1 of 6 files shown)


viniciuspjardim:
I would like to have this feature in Lwjgl3. I need to use Lwjgl3 because it has the controller connect / disconnect events. If I could also use the window icon, would be very nice.

Quexten:
I'd also like this to be pulled.

CreamyCookie:
I'd also like to get this fixed, but the [GLFW bug](https://github.com/glfw/glfw/pull/805) is fixed, so that part should be removed from the pull request.

cypherdare:
@CreamyCookie, It's fixed, but given Lwjgl's slow release schedule, I could see it taking more than a year to trickle down to an Lwjgl release build. I'd vote for keeping the hack until that happens, so this non-obvious bug doesn't waste anyone's time.

cypherdare:
@code-disaster, Your consolidation makes very good sense. I merged it in. I also added a convenience method to the config that assumes Internal files, since that's almost always going to be the case. There are similar convenience constructors for may core classes, so I think it makes sense here.

The GLFW size limits bug fix didn't make it into Lwjgl 3.1, so I think the workaround should stay.

cypherdare:
@code-disaster 
OK, turns out the GLFW update did make it in, so I removed the workaround. Window size limits seem to be entirely non-functional now in Lwjgl 3.1.0, but that's a separate issue from this PR.

cypherdare:
I realized there was a ridiculous amount of copy-pasted code from Lwjgl3WindowConfiguration in Lwjgl3ApplicationConfiguration, so I fixed it.

code-disaster:
I probably would have done the duplicate code removal in a separate PR, but overall it looks fine to me now.

cypherdare:
Probably should have. I separated them.

PR #4263, opened 2016-08-21
Fixed #4262 by skipping typename while deserializing map entries.

### LINKED ISSUE #4262: Map type information in deserialized map
#### Issue details
Type information of serialized map is preserved after the map is deserialized.

#### Reproduction steps/code
```

public class TestObject{

    private String id = "id";
    private Map<String, String> map = new HashMap<>();

    TestObject(){
        map.put("a", "b");
        map.put("c", "d");
    }

    public String getId() {
        return id;
    }

    public void setId(String id) {
        this.id = id;
    }

    public Map<String, String> getMap() {
        return map;
    }

    public void setMap(Map<String, String> map) {
        this.map = map;
    }
}

```

```

public class JsonJavaTest {
    public static void test(){

        Json json = new Json();
        json.setUsePrototypes(false);
        json.setIgnoreUnknownFields(true);
        json.setOutputType(JsonWriter.OutputType.json);

        String jsonStr = json.prettyPrint(new TestObject());
        // Expecting 2 items in map
        TestObject testObj = json.fromJson(TestObject.class, jsonStr);

        for(Map.Entry<String, String> entry : testObj.getMap().entrySet()){
            System.out.println(entry.getKey() + " : " + entry.getValue());
        }
    }
}
```

In the above example the expected result would be:
> a : b
> c : d

But deserialization produces

> a : b
> c : d
> class : java.util.HashMap

1.9.4-snapshot

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [x] MacOS

ollipekka:
Related to the issue:
- https://github.com/libgdx/libgdx/issues/4262

ArrayMap and ObjectMap are not affected, but the added test should cover this scenario for them as well.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Json.java
+++ b/gdx/src/com/badlogic/gdx/utils/Json.java
@@ -909,18 +909,24 @@ public class Json {
 					ObjectMap result = (ObjectMap)object;
 					for (JsonValue child = jsonData.child; child != null; child = child.next)
 						result.put(child.name, readValue(elementType, null, child));
+					
 					return (T)result;
 				}
 				if (object instanceof ArrayMap) {
 					ArrayMap result = (ArrayMap)object;
 					for (JsonValue child = jsonData.child; child != null; child = child.next)
 						result.put(child.name, readValue(elementType, null, child));
+					
 					return (T)result;
 				}
 				if (object instanceof Map) {
 					Map result = (Map)object;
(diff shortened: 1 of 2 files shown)


Tom-Ski:
Thanks!

Issue #4251, opened 2016-08-15
Feature: Allow adding/modifying log destination


MrPlow442:
Libgdx exposes it's built-in logger through `Gdx.app.log` methods. From what I can gather, these log messages are always sent to stdout as that is the default behavior of the lwjgl logger. 

I suggest adding a feature which would allow developers to change that default behavior and set a different output stream/writer, add new writers, or hook onto the current log methods so that they might also be output somewhere else (such as a file or dev console).

For example, let's say that the game has a dev console and we wish to redirect all logging calls to the in-game dev console instead of stdout. Ideally it would look something like this:

```
// Create a custom output writer
public class ConsoleWriter extends Writer     
     //........
     public void write(char cbuf[], int off, int len) throws IOException {
           console.write(cbuf, off, len);
     }
     //........
}

Gdx.app.setLogWriter(new ConsoleWriter()); 
// from this point on all Gdx.app.log/error/debug invocations will write to the ConsoleWriter instead of stdout
```

xoppa:
Possible duplicate of #1311

PR #4266, opened 2016-08-23
Feature: Added possibility to change log destination

pecrom:
Every class which implements Application interface, is able to set its own writer where the logs will be written to. Using setLogWriter. As it was requested in #4251.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+[1.9.5]
+
 [1.9.4]
 - Moved snapping from ProgressBar to Slider to prevent snapping when setting the value programmatically.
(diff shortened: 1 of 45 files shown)


Tom-Ski:
Thanks for the PR @pecrom, going with a different implementation to expose a logging interface. You can now provide a [ApplicationLogger](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/ApplicationLogger.java) implementation yourself to redirect/manage logging yourself.

pecrom:
Thanks for the feedback. This was the other way I was thinking about:-)

PR #4280, opened 2016-08-29
[WIP]Support for gradle 2.5+

arnoid:
!!!WARNING!!!!! TEST WERE NOT RUN YET

- renamed gdx-box2d module to gdx-box2d-parent
- renamed gdx-controlled module to gdx-controller-parent
- upgraded android plugin from 1.2.3 to 2.1.0


badlogic:
Uhm, I'm not sure what you are trying to do, but this is not something we'd merge even if it passed any tests.

arnoid:
I've participared in recent chat in IRC channel with mentions that current libgdx src is not buildable with latest gradle.
So, i was trying to fix this "issue".

**I do agree with your decision**, because I made changes that affect the whole project structure and require HUGE amount of tests to be done.

Issue #3994, opened 2016-04-07
xoffset set to minus padding value for all letters


wrobelp:
I'm using Hiero to create Distant Field fonts for Java.
When I create the font all the values of xoffset are -8 when I set padding 8.
#### Issue details

I loading from File. The problem is for all files I've tried.
The problem occurs on Java version. For FreeType version it's working well.

I creating Distant Fields, but I think problem occurs for any Java version settings. I can test it.

I tried to debug it. The Glyphs are created ok when switching to FreeType, but they're NOT reloaded/creted when switching to Java and change somewhere along the way. Cause later when saving to file values in Glyphs are all the same.

I also tried using latest nightly but same problem occurs.
#### Reproduction steps/code

Load font from .ttf file. Choose Java, add Distant Fileds, delete Color. Set up padding. Saving to BMP.
#### Version of LibGDX and/or relevant dependencies

I use newest runnable-hiero.jar v5

For now I'm using two .fnt files to load one font - one for xoffsets and one for all other data.
#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [x] Linux
- [ ] MacOS

Thank You for Your Tools :)

Tom-Ski:
Im guessing you are using Hiero?!? Please don't delete the issue template, fill it out so we can fully understand what the issue you are reporting is.

wrobelp:
Is it enough information?

cheapbrain:
I'm having the same issue with hiero, and after some testing I can confirm that the issue is caused by selecting the java rendering system.

badlogic:
Related to #4106

NathanSweet:
Looks fixed by #4282, if not will reopen.

PR #4282, opened 2016-08-31
Fix for issue #3994.

bryanwagner:
Fixes issue #3994 where the Java renderer does not correctly set the BMFont xoffset because `charVector` was not being used.

I was experiencing this issue as well because I want to use the Java renderer in Hiero.  It looks like a simple bug; I compiled Hiero with this change and the fix corrects `xoffset`.  To test, I'm saving Arial using Java Rendering, the Extended character set, and setting 1 page of size 1024x512.  Attached are `.fnt` files before and after the fix (renamed to `.txt` for Github upload).  Notice `xoffset=2` for every character in the "before" file.  The "after" file with this fix renders correctly for me.
[arial-after-fix.txt](https://github.com/libgdx/libgdx/files/446252/arial-after-fix.txt)
[arial-before-fix.txt](https://github.com/libgdx/libgdx/files/446251/arial-before-fix.txt)

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/unicodefont/Glyph.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/unicodefont/Glyph.java
@@ -80,8 +80,8 @@ public class Glyph {
 			char[] chars = Character.toChars(codePoint);
 			GlyphVector charVector = unicodeFont.getFont().layoutGlyphVector(GlyphPage.renderContext, chars, 0, chars.length,
 				Font.LAYOUT_LEFT_TO_RIGHT);
-			GlyphMetrics charMetrics = vector.getGlyphMetrics(0);
-			xOffset = vector.getGlyphPixelBounds(0, GlyphPage.renderContext, 0, 0).x - unicodeFont.getPaddingLeft();
+			GlyphMetrics charMetrics = charVector.getGlyphMetrics(0);
+			xOffset = charVector.getGlyphPixelBounds(0, GlyphPage.renderContext, 0, 0).x - unicodeFont.getPaddingLeft();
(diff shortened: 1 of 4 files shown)


extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/unicodefont/Glyph.java
```diff
@@ -80,8 +80,8 @@
 			char[] chars = Character.toChars(codePoint);
 			GlyphVector charVector = unicodeFont.getFont().layoutGlyphVector(GlyphPage.renderContext, chars, 0, chars.length,
 				Font.LAYOUT_LEFT_TO_RIGHT);
-			GlyphMetrics charMetrics = vector.getGlyphMetrics(0);
-			xOffset = vector.getGlyphPixelBounds(0, GlyphPage.renderContext, 0, 0).x - unicodeFont.getPaddingLeft();
+			GlyphMetrics charMetrics = charVector.getGlyphMetrics(0);
+			xOffset = charVector.getGlyphPixelBounds(0, GlyphPage.renderContext, 0, 0).x - unicodeFont.getPaddingLeft();
 			xAdvance = (int)(metrics.getAdvanceX() + unicodeFont.getPaddingAdvanceX() + unicodeFont.getPaddingLeft()
```
qzole: Shouldn't charMetrics be used here? Right now charMetrics is unused

qzole:
Hit this bug too, when I was trying to generate distance field glyphmaps for my game, it was annoying how the text was off a bit. Kudos for the find!

bryanwagner:
Thanks!  Yeah I thought I was doing something wrong at first.  I'm also using distance fields.  Hiero is a great tool for generating them.

NathanSweet:
Thanks!

PR #4287, opened 2016-09-01
Add cascading style support to Skin

cypherdare:
This will help clean up skin json files considerably. A prime use case is ImageButtonStyle (or ImageTextButtonStyle)--I find it is common to have many ImageButtonStyles in my skin with all the same values except for the Image. But I also have many other types of Button subclass styles where I am using some of the same values over and over. It becomes error prone to make tweaks to these parameters and keep the visual look of all the buttons the same.

Usage: For any style definition, you can specify a `parent` in the Json that names another style to use for default values. The parent can be a superclass style, so a TextButtonStyle could use a ButtonStyle as a parent, for example. You then only need to specify parameters that have different values than those of the parent.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
@@ -44,6 +44,7 @@ import com.badlogic.gdx.utils.JsonValue;
 import com.badlogic.gdx.utils.ObjectMap;
 import com.badlogic.gdx.utils.SerializationException;
 import com.badlogic.gdx.utils.reflect.ClassReflection;
+import com.badlogic.gdx.utils.reflect.Field;
 import com.badlogic.gdx.utils.reflect.Method;
 import com.badlogic.gdx.utils.reflect.ReflectionException;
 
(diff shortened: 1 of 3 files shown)


Tom-Ski:
Would be a nice feature and the PR looks good, could you run the formatter over the lines you've added/changed.

cypherdare:
@Tom-Ski, 
Done, and squashed.

cypherdare:
Just added in checking for unknown fields so previous behavior is maintained even for classes that have a template. 

i.e. You still get an exception if you have unknown fields in your JSON.

Tom-Ski:
Thanks for the changes, Ill try to have a look at this and your other Skin PRs this weekend.

cypherdare:
@Tom-Ski Might be a good time to give this a review since we're probably a ways off from the next release version. I've been working on an app with heavy UI with this change in it for the past year and a half, but the implementation is slightly different because I did it in a Skin subclass so I could avoid building a modified LibGDX.

The same goes for #4293, but that one is pretty low-risk. If one of these two gets pulled, the other won't be auto-mergeable.

gdx/src/com/badlogic/gdx/utils/Json.java
```diff
@@ -817,6 +826,16 @@ public void readFields (Object object, JsonValue jsonMap) {
 		}
 	}
 
+	/** Override to allow subclasses to use fake field names for other purposes in {@link #readFields(Object, JsonValue)} without
+	 * triggering an exception for unknown fields.
+	 * @param type The object type being read.
+	 * @param fieldName A field name encountered in the Json map for which there is no actual field.
+	 * @return Whether the field name should be treated as known, i.e., not throw an exception when encountered in
+	 *         {@code readFields()}. */
+	protected boolean isNonexistantFieldKnown (Class type, String fieldName) {
```
NathanSweet: I'm not a fan of this function name. How about `ignoreUnknownField`? Javadoc can explain it is only called when a JSON name doesn't have a matching field and `ignoreUnknownFields` is false.

gdx/src/com/badlogic/gdx/utils/Json.java
```diff
@@ -76,6 +77,10 @@ public void setIgnoreUnknownFields (boolean ignoreUnknownFields) {
 		this.ignoreUnknownFields = ignoreUnknownFields;
 	}
 
+	public boolean isIgnoreUnknownFields () {
```
NathanSweet: Let's use `getIgnoreUnknownFields`. Using `is` for a prefix is often awkward, then we end up with a mix of `get` and `is` for boolean properties.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
```diff
@@ -427,6 +428,49 @@ protected Json getJsonLoader (final FileHandle skinFile) {
...
 				return super.readValue(type, elementType, jsonData);
 			}
+
+			static final String CASCADE_PARENT_TAG = "parent";
+
+			protected boolean isNonexistantFieldKnown (Class type, String fieldName) {
+				return fieldName.equals(CASCADE_PARENT_TAG);
+			}
+
+			public void readFields (Object object, JsonValue jsonMap) {
```
NathanSweet: I wonder if this can be simplified. Eg, if we assume `parent` is never a valid field name AND we have some way to copy the parent, then we could do this in Json `readValue`:

```
if (jsonData.isObject()) {
	String parentName = jsonData.getString("parent", null);
	if (parentName != null) {
		Object parent = findParent(parentName);
		Object copy = copy(parent);
		json.readFields(copy, jsonData);
		return copy;
	}
}
```

I like that this doesn't require exposing more of the Json class. For `copy()`, a hack would be to serialize to JSON and read it back as a copy. It would be better as a new Json function, since copying is very close to serialization (Kryo does both). It's also common to need a copy of a style and we have copy constructors for that, but it's probably awkward to use those here.

cypherdare: Having used this in my own UI-heavy project for the past year and a half or so, I'd say the cascading feature would be roughly half as useful if we remove the ability for a parent style to be an ancestor class (for example using a ButtonStyle as the parent of a TextButtonStyle).

I suppose Json could have one more `copy` method where you specify the subclass of parent you want the "copy" to be. But is that getting too specialized? I can tinker with this in the next day or two.

NathanSweet: @cypherdare see what you think about dfc9f4789171e03f468faaa81343795306569c34.

PR #4289, opened 2016-09-01
Fix MOE 1.1.x support

rolandvigh:
- Update Xcode project template in gdx-setup
- Use 1.1.+ version strings to pick up hotfixes
- Add ProGuard rules to proguard.append.cfg


build.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     VERSION = "1.9.5-SNAPSHOT"
     gdxVersion = VERSION
     robovmVersion = "2.2.1-SNAPSHOT"
-    moeVersion = "1.1.0"
+    moeVersion = "1.1.+"
```
Tom-Ski: Would rather have these as fixed values rather than introduce dynamic versions that may break things down the line.

Tom-Ski:
Thanks a bunch!

PR #4291, opened 2016-09-03
Can use Skin subclasses with SkinLoader

cypherdare:
Skin's Json-loading behavior can be highly customized by overriding `getJsonLoader`, but then the subclass needs a custom Loader class for use with AssetManager. So I exposed a method for instantiating the skin.

Usage:

```
manager.setLoader(Skin.class, new SkinLoader(manager.getFileHandleResolver()){
        public Skin generateSkin (TextureAtlas atlas) { 
            return new MySkinSubclass(atlas); 
        }
    });
```


cypherdare:
I revised the implementation to be cleaner. Usage is now to instantiate a SkinLoader with a Skin subclass type when assigning it to the AssetManager:

    manager.setLoader(SkinSub.class, new SkinLoader(manager.getFileHandleResolver(), SkinSub.class));
    manager.load("data/uiskin.json", SkinSub.class);

PR #4293, opened 2016-09-05
Support class name tags in Skin json

cypherdare:
I realize the auto-tagging of built-in styles might be overdoing it, but can we at least support tags in the first place so subclasses can take advantage?


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
```diff
@@ -525,9 +527,24 @@ public Object read (Json json, JsonValue jsonData, Class type) {
...
 		});
+		
+		for (Class cls : AUTO_TAGGED_STYLES){
+			json.addClassTag(cls.getSimpleName(), cls);
+		}
 
 		return json;
 	}
+	
+	private static final Class[] AUTO_TAGGED_STYLES = {
```
NathanSweet: This may be better as an ObjectMap<String, Class> field on Skin, which would allow users to customize the tags before loading Skin JSON. It could be populated with reasonable defaults.

cypherdare:
Good idea. Configuring it for use with SkinLoader could be done with #4291.

Issue #4295, opened 2016-09-05
Hiero writes BMFont file padding with incorrect order


bryanwagner:
#### Issue details

Hiero writes the padding array in the incorrect order, according to the [BMFont file specification](http://www.angelcode.com/products/bmfont/doc/file_format.html). Both the file tags and binary versions specify the ordering as:  up, right, down, left.

Hiero is writing out the padding ordering as:  [up, left, down, right](https://github.com/libgdx/libgdx/blob/f504a294a4aac57a3d43e2ab8cfa8c860ac90044/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/BMFontUtil.java#L72).

I believe this is an error, but please let me know if it was intentional for any reason (I'm patching my local build; will submit a PR).
#### Reproduction steps/code

Run Hiero, File > Save BMFont files (text)... and open the BMFont file.
#### Version of LibGDX and/or relevant dependencies

1.9.4
#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

badlogic:
Our `BitmapFont` assumes that order. Changing this in `BitmapFont` as well as `Hiero` would break a lot of code and assets our in the open. Worse, if people regenerate only some of their fonts with `Hiero` using the new order, while other fonts are saved with the old order, people would be very confused and later on upset. I'm afraid we won't change this at this point.

Issue #3074, opened 2015-04-30
no support for BMFont's (.fnt) padding - xHeight & capHeight invalid for padded fonts


FyiurAmron:
Current BitmapFont.java implementation just ignores the `padding=` field; it makes using fonts with e.g. shadow/glow very hard, since both *xHeight* & *capHeight* are calculated based on actual texture size, not the size minus padding one. Since for shadowed/glowed fonts the texture size is often 2-3x bigger than the proper *xHeight*/*capHeight* calculated from of the "core" of the glyph, this results in very poor vertical positioning of the glyphs during text compositing (i.e. it's almost impossible to calculate where the actual glyph will appear due to offseting from padding). Horizontal positioning is unaffected, since it's directly based on raw .fnt data.

The solution I used to work around this was replacing https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java#L472 line with
```
// private static final int paddingCount = 4;
                String line = reader.readLine();
                String keyword = "padding=";
                String substr = line.substring( line.indexOf( keyword ) + keyword.length() );
                String[] paddingStr = substr.substring( 0, substr.indexOf( ' ' ) ).split( ",", paddingCount );
                if ( paddingStr.length != paddingCount ) {
                    throw new IOException();
                }
                int[] padding = new int[paddingCount];
                for( int i = 0; i < paddingCount; i++ ) {
                    padding[i] = Integer.parseInt( paddingStr[i] );
                }
                // up, right, down, left
                int paddingHorizontal = padding[1] + padding[3], // currently unused
                        paddingVertical = padding[0] + padding[2];
```
and then, in lines 606 through 628, all occurences of `xGlyph.height`/`capGlyph.height`/`glyph.height` with the respective `.height - paddingVertical`.

I don't know if this is the best way to do this, but it seems that it fixes the issue. Any suggestions/patches are most welcome.

NathanSweet:
How can I see the issue?

FyiurAmron:
a) load a font with a significant amount of padding (example here http://www.filedropper.com/caviarshade (.fnt) & http://www.filedropper.com/caviarshade_1 (.png)),
[...]

PR #4297, opened 2016-09-05
Hiero BMFont padding order fix (#4295)

bryanwagner:
Fix for issue #4295.

Notice that there's actually a discrepancy in LibGDX with itself.
BitmapFontWriter writes:  up, down, left, right
BMFontUtil writes:  up, left, down, right
BMFont reads:  up, left, down, right

In order to make the padding ordering compatible with the [BMFont specification](http://www.angelcode.com/products/bmfont/doc/file_format.html), I normalized all orderings to up, right, down, left.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
@@ -34,12 +34,7 @@ public class BuildScriptHelper {
 		write(wr, DependencyBank.mavenLocal);
 		write(wr, DependencyBank.mavenCentral);
 		write(wr, "maven { url \"" + DependencyBank.libGDXSnapshotsUrl + "\" }");
-		if (projects.contains(ProjectType.HTML)) {
-			write(wr, DependencyBank.jCenter);
-		}
-		if (projects.contains(ProjectType.IOSMOE)) {
-			write(wr, "maven { url " + DependencyBank.moeLocalUrl + " }");
-		}
+		write(wr, DependencyBank.jCenter);
 		write(wr, "}");
 		//dependencies
 		write(wr, "dependencies {");
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/bmfont/BitmapFontWriter.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/bmfont/BitmapFontWriter.java
@@ -181,8 +181,8 @@ public class BitmapFontWriter {
 			.append(quote(info.italic ? 1 : 0)).append(" charset=\"").append(info.charset == null ? "" : info.charset)
 			.append("\" unicode=").append(quote(info.unicode ? 1 : 0)).append(" stretchH=").append(quote(info.stretchH))
 			.append(" smooth=").append(quote(info.smooth ? 1 : 0)).append(" aa=").append(quote(info.aa)).append(" padding=")
-			.append(xmlQuote).append(info.padding.up).append(",").append(info.padding.down).append(",").append(info.padding.left)
-			.append(",").append(info.padding.right).append(xmlQuote).append(" spacing=").append(xmlQuote)
+			.append(xmlQuote).append(info.padding.up).append(",").append(info.padding.right).append(",").append(info.padding.down)
+			.append(",").append(info.padding.left).append(xmlQuote).append(" spacing=").append(xmlQuote)
 			.append(info.spacing.horizontal).append(",").append(info.spacing.vertical).append(xmlQuote).append(xmlCloseSelf)
 			.append("\n");
 
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/BMFontUtil.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/BMFontUtil.java
@@ -69,8 +69,8 @@ public class BMFontUtil {
 		int pageHeight = unicodeFont.getGlyphPageHeight();
 		out.println("info face=\"" + font.getFontName() + "\" size=" + font.getSize() + " bold=" + (font.isBold() ? 1 : 0)
 			+ " italic=" + (font.isItalic() ? 1 : 0) + " charset=\"\" unicode=0 stretchH=100 smooth=1 aa=1 padding="
-			+ unicodeFont.getPaddingTop() + "," + unicodeFont.getPaddingLeft() + "," + unicodeFont.getPaddingBottom() + ","
-			+ unicodeFont.getPaddingRight() + " spacing=" + unicodeFont.getPaddingAdvanceX() + ","
+			+ unicodeFont.getPaddingTop() + "," + unicodeFont.getPaddingRight() + "," + unicodeFont.getPaddingBottom() + ","
+			+ unicodeFont.getPaddingLeft() + " spacing=" + unicodeFont.getPaddingAdvanceX() + ","
 			+ unicodeFont.getPaddingAdvanceY());
 		out.println("common lineHeight=" + unicodeFont.getLineHeight() + " base=" + unicodeFont.getAscent() + " scaleW=" + pageWidth
 			+ " scaleH=" + pageHeight + " pages=" + unicodeFont.getGlyphPages().size() + " packed=0");
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/unicodefont/Glyph.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/hiero/unicodefont/Glyph.java
@@ -80,8 +80,8 @@ public class Glyph {
 			char[] chars = Character.toChars(codePoint);
 			GlyphVector charVector = unicodeFont.getFont().layoutGlyphVector(GlyphPage.renderContext, chars, 0, chars.length,
 				Font.LAYOUT_LEFT_TO_RIGHT);
-			GlyphMetrics charMetrics = vector.getGlyphMetrics(0);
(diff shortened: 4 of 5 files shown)


cypherdare:
@Tom-Ski This one can be closed, not merged, as per Mario's comment on #4295.

RUSshy:
this is kinda bad to not fix issue that is present

libGDX approaches version 2.0

maybe time for a huge cleanup, and merge PR like this to fix old mistakes ?

RUSshy:
closed ? huh
this repo don't want more users.. sad to let libgdx die..

obigu:
It's a shame this is not merged, not everybody uses Hiero to generate their BMFont files. Not fixing this we are officially creating a fork of the BMFont format reducing our choices when it comes to glyph editors.

bryanwagner:
Friends, you can work around this. First, the padding is often the same value in all 4 fields. The only ones you need to consider are the middle ones, left and down. So your tools can check if left != down and issue a warning. Then, if you need to convert on import into LibGDX, you can make something like a gradle script that swaps the two values. If they change their source, then existing apps that dynamically link to the jar will possibly break. They should claim to be "almost" BMFont and issue a warning in their documentation (I haven't checked).

mgsx-dev:
libgdx wiki and javadoc states it rely on official BMFont specification. But current implementation is wrong about padding order.

I'm wondering why it's not acceptable to force users to migrate their BMFont files while it's acceptable to do so with their particle2D (because of recent changes).
We could add a "MIGRATION" tag in CHANGES.md to explain how to do it (either manually or with hiero). Maybe it would solve user's potential issues/frustrations discussed here : https://github.com/libgdx/libgdx/issues/4295#issuecomment-264701827
@badlogic what do you think?

Since there are discrepancy across 3 libgdx classes, I think it would be worth it to comply with official specification.

If you really think it souldn't be changed, we should at least : make these classes consistent, update javadoc and wiki to state it clearly.

Please reconsider closing this.

Tom-Ski:
We force breaking of Particles due to the terrible parsing that we do.  Its not on purpose, it would be much better if it was completely backwards compatible. The issue is how easily overlooked this change is, even when its in notes and in release notes. People will be using asset files that have been generated without version tag, from old versions of the tool, with no idea if the padding is correct/incorrect.  There is no way for the user to know from the BMFont file. 

Since there was 0 discussion on this for 3 years and on the original issue to combat @badlogic's decision, this was closed. Perhaps if people were that opinionated on it during those times it would have been merged.

I am indifferent to these changes, so we can reopen now everyone has come out of the woodwork.

obigu:
@Tom-Ski Thanks, you're right we should have shown support to the PR before instead of waiting for it to be closed. Would you consider merging something in the lines of giving the option to set the format to BitmapFont (for instance HIERO, BMFONT or even adding a version to open for future changes like HIERO_1.0) being Hiero format the default one?

Tom-Ski:
I think adding some versioning from Hiero would be great. At least that would give people a hint as to if their assets need an update.

I don't really like the idea of support both formats. Its relatively easy for people to update their fonts by just scripting over the font files, or just manually changing the order. As long as we present documentation on how to do this when we release, I think it will be ok. If we present multiple ways to load, it will still be an issue with old files, not knowing which format they are in.

Is there any meta we have to know if something is from hiero currently?

obigu:
There's no meta I can see unfortunately.

I don't see how my suggestion would still be an issue if we keep Hiero format as default. Even if we created a new 2.0 Hiero format with metadata in the future (which would also mean we break BMFont standard again afaik), when loading a font file we would first check if there's metadata to know the Hiero version and if there's not assume it's old current Hiero 1.0 format.

Adding the option to pass an alternative format to Hiero such as BMFont I don't see how it hurts in any way and makes non Hiero users life much easier (and for normal users is easier to understand the concept of formats without having to read a lot of docs on how to edit the files).

cypherdare:
I don't think meta data fixes the root issue you're attempting to fix,
which is that fonts not created by Heiro do not import correctly. Maybe we
can create a global static flag (similar to ShaderProgram.pedantic) in
BitmapFont for loading legacy Heiro fonts. And maybe we can get away with
defaulting it to off because most people probably aren't using asymmetrical
padding. When those people see their text looks different after updating,
surely they would look at CHANGES and see they can fix it with one line.

People who use BMFonts not created by Heiro are most likely not affected by
this change or they would have noticed something amiss when they were using
unusual asymmetrical padding, and would have been chiming in rigy here.
This is probably a fraction of a fraction of users, if they exist at all.

We could add a comment of the top of newly created Heiro fonts to indicate
the new version just for the sake of manually identifying them.

Tom-Ski:
The idea behind meta data would be to ensure backwards compatibility. If we could identify that a font was made with the incorrect standard via old Hiero, we can trivially load the font correctly for support for the old format.

I don't think its worth keeping the backwards compatibility, because _if_ the user even finds out they have a problem with their font, they will (hopefully eventually) look up the changelog, and just change them in a few seconds.  The only reason for keeping backwards compatibility imo is if we can automatically detect the font was made in the old format.  Id say finding the global flag that lets people load old formats is about the same effort as changing the paddings in the font files, and we dont have to keep code to support some invalid legacy format.

cypherdare:
That makes sense for simplicity. The number of affected people is probably quite small, and if it's mentioned in CHANGES we can provide a link to a Gist script for "upgrading" old fonts.

I don't see how meta data on new fonts can help us detect old ones without incorrectly labeling all old and new BMFonts from other font creators besides Heiro. And if we didn't care about those, there'd be no point in fixing this.

Tom-Ski:
Im talking about if there were any metadata in the original fonts that we could harness.

NathanSweet:
https://github.com/libgdx/libgdx/issues/5422#issuecomment-506944552 is related to this PR and @NumeronR is saying that we should not be doing anything with padding at all. #3074 by @FyiurAmron is the reason we started using padding values, which very well could have been wrong.

I'm generally terrified of changing font stuff, as it means all my apps break. Regenerating fonts isn't a big deal, but changes to how values are interpreted is much more pain to fix. App code is usually making assumptions about where text will be drawn based on these values. When it changes, fixing it usually means dorking with the fonts so they have whatever values they used to have, which isn't terribly easy. This [happened in the past](https://www.badlogicgames.com/wordpress/?p=3678) and you can see the code snippet provided there which massages the font metrics back to how they used to be.

If we must make breaking changes then we should be sure it is worth doing and that we're doing it right to avoid making this mess even bigger.

bryanwagner:
As the one who submitted the PR, I'm actually in favor of dropping the PR for those reasons. Maybe the `pedantic` flag suggested earlier would be a great idea. I've since moved on from the issue, but I do think it's important simply to update the documentation describing the caveats. I think it's a very minor issue overall.

bryanwagner:
Although, the internal inconsistency I found described in the OP between the two writers deserves a second look.

RUSshy:
if your layout is depending on font settings, then your layout sucks

building hacks on hack on hacks lead to this situation, fix it, and no more hacks needed

MrStahlfelge:
@mgsx-dev @obigu It's now your chance to get this done. :)

obigu:
I agree with Tom-Ski:
> I don't think its worth keeping the backwards compatibility, because if the user even finds out they have a problem with their font, they will (hopefully eventually) look up the changelog, and just change them in a few seconds. The only reason for keeping backwards compatibility imo is if we can automatically detect the font was made in the old format. Id say finding the global flag that lets people load old formats is about the same effort as changing the paddings in the font files, and we dont have to keep code to support some invalid legacy format.

I would merge this (editing CHANGES obviously) and follow the BMFont specification. This should affect a very small number of users (who use irregular paddings on fonts generated by Hiero) that will need to recreate the font if anything breaks. That's the whole point of versioning in my opinion but I understand it may scare some people. In any case it's a minor thing and I think we have @NathanSweet veto (if that's the case we can close it).

NathanSweet:
As mentioned in my comment above, maybe reading padding at all is the wrong thing to do. Assuming we really do need to read it, I'm indifferent on making the change. It would be better to follow the spec, but on the other hand the issue seems relatively minor and may not be worth the disruption. If you guys want to make the change, I don't mind.

tommyettinger:
I agree with @NathanSweet and @NumeronR that padding shouldn't affect the appearance of the font at all, since it's part of the `info` section of a .fnt file. I've personally made fonts with asymmetrical padding (hand-editing that section) and was very confused by the ordering; I almost always had to resort to trial and error to see which padding value affected which side of a glyph. Unfortunately, ignoring padding might also break the appearance of fonts that relied on its (incorrect) usage in libGDX. So my vote is to close without merging; using padding at all is incorrect, so other than the inconsistent writing order, Hiero can make .fnt fonts with padding that isn't to-spec and other tools shouldn't even notice. Inside libGDX, we can continue to be as incorrect as we always have been with .fnt loading and no code or assets should have to change.

obigu:
Ok, let's close this as it seems it's all a bit messed up and hard to change at this point.

@tommyettinger The problem I was referring to is not about opening .fnt files generated with Hiero on ther tools but .fnt files generated by other tools that follow BMFont specs (such as AngelCode BMFont tool itself) being loaded on libGDX and the confusions it may cause (considering that, even if it shouldn't, libGDX uses padding for positioning).

RUSshy:
this should be fixed imo

why not something like this:

- a new field in the info section (libGDXversion="1.9.12")

- for writing new fonts, add that info
- for reading fonts, check if that info is here
    - if it is here, check the version and handle the padding accordingly
    - if it is NOT here, then handle padding the old and wrong way

MrStahlfelge:
Would not help, this way we still wouldn't be compatible to the specification.

Issue #4300, opened 2016-09-08
IndexOutOfBounds on Queue.first() while size > 0


Robbiedobbie:
#### Issue details

An IndexOutOfBounds exception occurs when queue.first() is called. This seems to only happen when removing all/the last item in a queue through an iterator, and then adding a new item to the queue again.
#### Reproduction steps/code

```
import java.util.Iterator;

import org.junit.Test;

import com.badlogic.gdx.utils.Queue;

public class LibgdxQueueBug {

    @Test
    public void testQueue() {
        Queue<Integer> queue = new Queue<Integer>();

        //Simulate normal usage
        for(int i = 0; i < 100; i++) {
            queue.addLast(i);
            if(i > 50)
                queue.removeFirst();
        }

        Iterator<Integer> it = queue.iterator();
        while(it.hasNext()) {
            it.next();
            it.remove();
        }

        queue.addLast(1337);

        Integer i = queue.first();
    }
}
```
#### Version of LibGDX and/or relevant dependencies

Tested in LibGDX version 1.9.3, but looking at the git repo, the queue is not changed since.
#### Stacktrace

``` java
java.lang.ArrayIndexOutOfBoundsException: 64
    at com.badlogic.gdx.utils.Queue.first(Queue.java:256)
    at rails.LibgdxQueueBug.testQueue(LibgdxQueueBug.java:28)
    ...
```
#### Please select the affected platforms
- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

Darkyenus:
Thanks for great bug report! Fixed in #4302.

badlogic:
Merged #4302

PR #4302, opened 2016-09-11
Fix Queue bug #4300

Darkyenus:
Also adds a new test for this edge-case (verbatim from issue) and cleans up Queue and QueueTest (in separate commits for easy review).

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -238,7 +238,9 @@ public class Queue<T> implements Iterable<T> {
 			value = (T)values[index];
 			System.arraycopy(values, head, values, head + 1, index - head);
 			values[head] = null;
-			this.head++;
+			if (++this.head == values.length) {
+				this.head = 0;
+			}
 		}
 		size--;
 		return value;
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -19,7 +19,6 @@ package com.badlogic.gdx.utils;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
-import com.badlogic.gdx.utils.Array.ArrayIterable;
 import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
 /** A resizable, ordered array of objects with efficient add and remove at the beginning and end. Values in the backing array may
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -142,8 +141,7 @@ public T removeFirst () {
 
 		final T result = values[head];
 		values[head] = null;
-		head++;
```
gjroelofs: Personally, I'd refrain from this pattern as it just makes the code harder to read without any performance benefits.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -225,20 +223,22 @@ public T removeIndex (int index) {
...
-			value = (T)values[index];
+			value = values[index];
 			System.arraycopy(values, index + 1, values, index, tail - index);
 			this.tail--;
 		} else { // index is between head and values.length.
-			value = (T)values[index];
+			value = values[index];
 			System.arraycopy(values, head, values, head + 1, index - head);
 			values[head] = null;
-			this.head++;
```
gjroelofs: Same as l145

gdx/test/com/badlogic/gdx/utils/QueueTest.java
```diff
@@ -290,6 +313,6 @@ private void assertEqualsAndHash (Queue<?> q1, Queue<?> q2) {
 
 	private void assertValues (Queue<Integer> q, Integer... values) {
 		for (int i = 0, n = values.length; i < n; i++)
-			if (values[i] != q.get(i)) fail(q + " != " + new Array(values));
+			if (values[i] == null ? q.get(i) != null : !values[i].equals(q.get(i))) fail(q + " != " + new Array(values));
```
gjroelofs: Probably better not to have ternary statements inside if statements. Better to split it up.
Also, make sure you're using the LibGDX formatter.

Darkyenus: This is IntelliJ autogenerated null-safe equals check. I can split it if needed, but as it is just in Test, I didn't deem that necessary.

gdx/src/com/badlogic/gdx/utils/Queue.java
Darkyenus: I did it to be in line with libGDX's pattern of [bringing object fields into method variables first](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/Array.java#L162) before accessing them multiple times, for performance reasons (for Android, mainly, IIRC) (performance reason being, that due to multithreading, these two variants may have different semantics. It is true that JVM is free to change the semantics slightly for optimizations, I am not sure if it happens on Dalvik (or at least its earlier versions). Since this pattern is so common, I expect that it is just for this reason).

I'll let others decide on which "style" should be used, as similar pattern of inlining inc/dec operations is fairly common in libGDX's collections (just maybe not inside `if()`?).

gjroelofs: The main comment is on the if(++head ..) format which is an inc/dec operator inside the if loop.

Darkyenus: I have assumed so. My argument is that:
```
this.head++;
if(this.head == values.length) ...
```
requires two reads from the object, while:
```
if(++this.head == values.length) ...
```
has to read just once, which could be slightly faster, in some circumstances.

But it is not a big deal either way. I'll wait for core contributors to decide.

Robbiedobbie: So, I just want to jump in on this discussion.

First of all, I want to say that for readability, I prefer having the increment out of the if.

However, I mostly want to jump in because of the argument that this would create multiple reads. Fact is, you simply can't rely on what bytecode is generated for such small syntactic differences. I did test it out for the current java version on my system (javac 1.8.0_74), and when using an instance field it actually generated the exact same bytecode:

```
aload_0     //Load this pointer on stack
dup         //Duplicate this pointer on stack
getfield #2 // Field test:I
iconst_1    // Push 1 to stack
iadd        // Add top 2 elements
putfield #2 // Field test:I
aload_0     // Load this pointer on stack
getfield #2 // Field test:I
```
However, when using a static field, it would read the field only once, but duplicate it once more on the stack. It then depends on the jit compiler whether one is faster than the other. However, in the field instance case, performance will be exactly the same.

gjroelofs: @Darkyenus Apologies, I misinterpreted your explanation as I saw no assignment. To me, this level of optimization is rather done by the compiler if it comes a the expense of readability.

gdx/test/com/badlogic/gdx/utils/QueueTest.java
gjroelofs: After rereading; I'd just use Assert.assertEquals().

Darkyenus:
Fair points, thanks for comments, amended!

gjroelofs:
Also; thanks for the fix :-)

badlogic:
Thanks!

PR #4305, opened 2016-09-15
Fixed autoreleasepool usage

rolandvigh:


nhydock:
Is it really okay to make direct references to MOE here?  Wouldn't this break compatibility for people who still use the RoboVM fork, or is that officially no longer supported?

Tom-Ski:
There are two backends, one for MOE, and one for RoboVM.

nhydock:
Ah sorry, I missed that this change was only under iosmoe

PR #4314, opened 2016-09-18
Added md5 hash to file name for GWT preloader

stanislaw89:
Added MD5 hash to filename of asset files for easier dealing with browser cache.
Before this pull request the assets.txt file produced by PreloaderBundleGenerator may contain something like:

> t:uiskin/arial_wgl_b_13.fnt:27071:application/unknown
> 
> i:uiskin/arial_wgl_b_13.png:11533:image/png

After pull request:

> t:uiskin/arial_wgl_b_13.fnt:uiskin/arial_wgl_b_13-5af9a1bc28326501c446f7aafcd2e7fa.fnt:27071:application/unknown
> 
> i:uiskin/arial_wgl_b_13.png:uiskin/arial_wgl_b_13-1de0c7c7a0eb31e51db9393bfcc28695.png:11533:image/png

So, in addition to filepath, url for downloading was added which contains md5 of the file. Developer who uses it can access file by the filepath just like it was before.
These make working with browser cache easier. Having md5 of file in url is common practice (btw GWT itself uses it for js files) and saves developer from pain of configuring cache headers etc (which may not be easy if some CDN is used).


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/Preloader.java
```diff
@@ -117,8 +119,8 @@ public Preloader (String newBaseURL) {
 
 	public void preload (final String assetFileUrl, final PreloaderCallback callback) {
 		final AssetDownloader loader = new AssetDownloader();
-		
-		loader.loadText(baseUrl + assetFileUrl, new AssetLoaderListener<String>() {
+
+		loader.loadText(baseUrl + assetFileUrl + "?etag=" + System.currentTimeMillis(), new AssetLoaderListener<String>() {
```
MrStahlfelge, line 123: This looks like an approach to disable the cache completely. If used, md5 hash is not needed any more. 
I see the advantage of a md5 hash that the cache _can_ be used.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/Preloader.java
```diff
@@ -132,25 +134,25 @@ public void onSuccess (String result) {
...
+					if (tokens.length != 5) {
 						throw new GdxRuntimeException("Invalid assets description file.");
 					}
 					AssetType type = AssetType.Text;
 					if (tokens[0].equals("i")) type = AssetType.Image;
 					if (tokens[0].equals("b")) type = AssetType.Binary;
 					if (tokens[0].equals("a")) type = AssetType.Audio;
 					if (tokens[0].equals("d")) type = AssetType.Directory;
-					long size = Long.parseLong(tokens[2]);
+					long size = Long.parseLong(tokens[3]);
```
MrStahlfelge, line 145: (The problem was already here before this PR, I just note it here.) The token index numbers should be covered in constants named, so it is better understandable what each index is used for. Reference comment to PreloaderBundleGenerator should be added to these constants so it is better understandable what's happening here.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
```diff
@@ -116,17 +122,21 @@ public String generate (TreeLogger logger, GeneratorContext context, String type
...
+				if (pathMd5.startsWith("/")) pathMd5 = pathMd5.substring(1);
 				buffer.append(asset.type.code);
 				buffer.append(":");
-				buffer.append(path);
+				buffer.append(pathOrig);
+				buffer.append(":");
+				buffer.append(pathMd5);
 				buffer.append(":");
-				buffer.append(asset.file.isDirectory() ? 0 : asset.file.length());
+				buffer.append(asset.fileMd5.isDirectory() ? 0 : asset.fileMd5.length());
```
MrStahlfelge, line 137: Looks like a mistake on first glance, but it is just the naming that is confusing and should be improved.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
```diff
@@ -266,4 +279,15 @@ private String createDummyClass (TreeLogger logger, GeneratorContext context) {
 		sourceWriter.commit(logger);
 		return packageName + "." + className;
 	}
+
+	private static String md5(byte[] bytes) {
+		try {
```
MrStahlfelge, line 284: maybe fallback to timestamp would be good here to have this failsafe.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/Preloader.java
MrStahlfelge: Okay, I know better now. This is for the assets.txt. And needed.

MrStahlfelge:
Merged, fixed the conflicts and tested. There was a slight problem with files without extensions, but overall, it looks good to me.

MrStahlfelge:
What I did not test so far: if this really helps with the cache.

MrStahlfelge (approved):

Issue #4230, opened 2016-08-05
LibGDX project doesn't build with Gradle version over 2.9


obigu:
When trying to build the libGDX project from source using a Gradle version higher than **2.9** the following error message appears:

```Error:Gradle version 2.2 is required. Current version is 2.10. If using the gradle wrapper, try editing the distributionUrl in /XXXXXX/git/libgdx/gradle/wrapper/gradle-wrapper.properties to gradle-2.2-all.zip.```

This is because of a limitation on the version of the Android Gradle plugin (com.android.tools.build:gradle) that is set to **1.2.3**. The fix to support Gradle **2.10+** was introduced on version **1.5.0** of the Android plugin (this is the fix https://android.googlesource.com/platform/tools/base/+/b96ce169b385d006cfccc4c9f1e54ab47140f59d).

But if we set the **com.android.tools.build:gradle** version to **1.5.0** a new error appears this time:

```Error:(272, 0) Your project contains 2 or more modules with the same identification com.badlogicgames.gdx:gdx-box2d
at ":extensions:gdx-box2d" and ":extensions:gdx-box2d:gdx-box2d".
You must use different identification (either name or group) for each modules.
<a href="openFile:/XXXXXXX/git/libgdx/build.gradle">Open File</a>````

This error also appears just by setting the **com.android.tools.build:gradle** version to **1.3.0**.

**So, if I haven't made a mistake on my tests of course, libGDX is currently stuck on pretty old versions of Gradle and Android Gradle plugin versions until a fix for the box2d modules is found.**

Also it looks like the information on the versions page regarding the Android Gradle plugin is not correct https://libgdx.badlogicgames.com/versions.html.

Tom-Ski:
That information is what the setup tool uses to generate new projects, not the LibGDX gradle build.  There is no wrapper pushed for LibGDX source itself, but running with gradle version 2.5 is fine.  It will get updated eventually, what features are you wanting to use from 2.9?

obigu:
Ok.

None in particular, I just wanted to raise awareness about this limitation in case anybody tries to build it with a current Gradle version and comes across this issue.

3.0 is apparently going to be relased soon and comes with important performance improvements. May be a good time to update, it takes ages to build on my laptop... :(

Tom-Ski:
Ok, and yeah the updates are infrequent because its such a pita,  some plugins always breaks, and if we are lucky its in an obvious way. Then its testing on all IDEs and platforms to see if they are also behaving.

[...]

PR #4316, opened 2016-09-20
Updated to compile and run under Android Studio 2.2

Zomby2D:
With the release of Android Studio 2.2, you can no longer push your Android apps to the emulator or your hardware without first updating the Android Gradle plugin to version 2.2.0. This is of course dependant on Gradle 2.14.1+ so it breaks the build because of some modules and their parents sharing the same name. This was previously discussed in issue #4230 and we are now facing a situation where we need to address this, unless we want to get stuck in the past.

In order to resolve the issue, I have added the suffix "-common" to the main branch of ``gdx-box2d`` and ``gdx-controllers``. These fixes have allowed me to compile the source under Android Studio 2.2 and run the gdx-test.

So far it has only been tested under the latest version of Android Studio, with builds for lwjgl, lwjgl3, and Android. It would probably need to be tested under all IDE and on every platform before anyone could consider merging this.


tests/gdx-tests-android/AndroidManifest.xml
```diff
@@ -4,7 +4,7 @@
       android:versionCode="1"
       android:versionName="1.0"
       android:installLocation="preferExternal">
-	<uses-sdk android:minSdkVersion="8" android:targetSdkVersion="19"/>
+	<uses-sdk android:minSdkVersion="9" android:targetSdkVersion="19"/>
```
Tom-Ski: Why change this?

Zomby2D: The support libraries now require a minSdkVersion version of 9. The other option is to override each and every one of them in the manifest.

MobiDevelop: Good thing that the support libraries that are supposed to provide compatibility back to SDKs 4 and 7 now "require" a minimum of SDK 9. Good job Google.

MobiDevelop: I feel like, since we don't actually know if the user is going to use support libraries, we should default to the minimum required by libgdx itself, not some library that someone may or may not use.

Zomby2D: The tests should probably be able to run on the same platform set as libGDX itself. I went ahead with the overrides instead of upping the minSdk. I have done some quick testing with my old Archos 48 (which is on Android 2.2) as well as in the emulator and most test seems to run fine.

That being said, Google has deprecated SDK 8 (and the associated images) so I'm wondering if we should keep supporting it moving forward.

Zomby2D:
After playing around a bit with the build I realized that my initial solution caused a change in the distributable jar names.  I refactored the parent "module" name instead of the child. It's a bit of a messier change but it keeps the original names of the jar, not requiring any change to end-user apps.

PR #4318, opened 2016-09-20
Fixes #3742, 3D particles not rendered behind transparent models

### LINKED ISSUE #3742: 3d particles doesn't play nice with other transparent renderables
As you can see in the screenshot below, sometimes 3d particles are not draw behind semi transparent renderables. 

It seems that the particles renderable `worldTransform` field is never updated with the particles position. Therefore `DefaultRenderableSorter` can't sort the renderables properly and the particles aren't rendered when they are behind another renderable.

screenshot

jsjolund:
This fixes the problem mentioned in issue #3742 by using a separate ```Renderable``` for each ```ParticleController``` in  ```BillboardParticleBatch``` and ```PointSpriteParticleBatch```. The ```Renderable```s are obtained from a pool as needed.

Depends on #4317 to work correctly.

The test/demo loads up pre-generated 3D particle systems of all three types. Video at
[http://youtu.be/-LVBtB3fgpU](http://youtu.be/-LVBtB3fgpU)


jsjolund:
One thing to note is this does not fix the problem when a ```ParticleController``` originates inside a transparent ```Renderable```. The particles inside the transparent ```Renderable``` are not rendered. It only fixes the problem when the particles originate outside it. This is only a problem for PointSprite and Billboard systems,  ```ModelInstanceParticleBatch``` works correctly.

I believe that to solve this, Billboards and PointSprites would need to use a separate  ```Renderable``` for each particle, which is wasteful. Or perhaps it can be solved in another way.

mgsx-dev:
I simply rebased the PR against master, i need to review and test it.

mgsx-dev:
I made this change configurable because it may have noticeable impacts on performances. This way, user can decide when enable it or not depending on the scene.

I'm not used to libGDX Particle3D API and implementation, so i'm not fully confident on this PR. I think it'd need some reviews.

@jsjolund I don't know if you still using libGDX but if you have time to review my changes against your original PR and tell me what you think about it, it would be great. Thank you for the PR BTW.

jsjolund:
> @jsjolund I don't know if you still using libGDX but if you have time to review my changes against your original PR and tell me what you think about it, it would be great. Thank you for the PR BTW.

Cool. I have not used libGDX in many years and cannot help much unfortunately.

mgsx-dev:
@jsjolund that's fine, i understand :-)

mgsx-dev:
I rebased the PR against master and also fixed a regression with old behavior when all particles should be sorted all together. This fix introduce a breaking change with ParticleSorter but i don't think users have their own implementation of it.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/PointSpriteParticleBatch.java
```diff
@@ -96,34 +113,78 @@ public PointSpriteParticleBatch (int capacity, ParticleShader.Config shaderConfi
 			this.blendingAttribute = new BlendingAttribute(GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_ALPHA, 1f);
 		if (this.depthTestAttribute == null) this.depthTestAttribute = new DepthTestAttribute(GL20.GL_LEQUAL, false);
 
-		allocRenderable();
+		renderables = new Array<Renderable>();
+		renderablePool = new RenderablePool();
+		blendingAttribute = new BlendingAttribute(GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_ALPHA, 1f);
+		depthTestAttribute = new DepthTestAttribute(GL20.GL_LEQUAL, false);
```
JackRainy, line 119: I am not quite sure I understand this part. A few lines above we assign `this.blendingAttribute` and `this.depthTestAttribute` from the parameters. If this assignment brought `null` then we create new attributes there.
And at this point, we override the parameters with yet another new objects.
Is it needed? If we need to pass the attributes as out parameters, does it make sense here to assign recently created `this.blendingAttribute` and `this.depthTestAttribute`, not making new ones ?

mgsx-dev: Oh, i missed that during the rebase. Thank you very much!

PR #4319, opened 2016-09-20
DefaultRenderableSorter accounts for center of Renderable mesh

jsjolund:
Useful for cases when ```Renderable``` translation cannot be stored in ```Renderable#worldTransform``` but must be stored in ```Renderable#MeshPart#center```, for example in 3D ```ParticleSystem```. The performance penalty of this change can be tested in e.g. ```Basic3DTest```, but I had a hard time seeing any change. It seems to be insignificant compared to the time it takes to render the models.


xoppa:
Thanks!

cyberbach:
Need sorting by blending status. Models with blending attribute must be rendered after ALL models without blending attribute.

Issue #4267, opened 2016-08-23
Race condition on AndroidGraphicsLiveWallpaper


lucasdupin:
Hey all,
We found a deadlock that happens on Android live wallpapers.
#### Issue details

There was a race condition between the GL thread and the main thread, when resuming wallpapers. It happens in [here](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java#L138)

If GL thread is much quicker because of scheduling delays, it might end up drawing its frame before the main thread reaches `resume()`.

This is a problem when you are not drawing on continuous mode, which most people probably do to save some battery.

In that case, `synch.wait()` will just hang because nobody else requested a frame since rendering was set to _on-demand_.
#### Reproduction steps/code

It's kind of tricky to reproduce, since it's a concurrency problem, but can add:

``` java
Thread.sleep(3000);
```

[Before the `synch.wait()` call](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java#L138)
#### Version of LibGDX and/or relevant dependencies

Latest on `master`.
#### Stacktrace

``` java
"main" prio=5 tid=1 Waiting
  | group="main" sCount=1 dsCount=0 obj=0x74b2ff58 self=0x7ee2c95a00
  | sysTid=29955 nice=0 cgrp=default sched=0/0 handle=0x7ee6c21a98
  | state=S schedstat=( 146166665 78867030 518 ) utm=9 stm=4 core=1 HZ=100
  | stack=0x7fcc076000-0x7fcc078000 stackSize=8MB
  | held mutexes=
  at java.lang.Object.wait!(Native method)
  - waiting on <0x07dbeb25> (a java.lang.Object)
  at com.badlogic.gdx.backends.android.AndroidGraphicsLiveWallpaper.resume(AndroidGraphicsLiveWallpaper.java:139)
  - locked <0x07dbeb25> (a java.lang.Object)
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaper.onResume(AndroidLiveWallpaper.java:156)
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaperService$AndroidWallpaperEngine.onResume(AndroidLiveWallpaperService.java:402)
[...]

PR #4326, opened 2016-09-24
#4267 Fixing a deadlock problem in AndroidGraphics and AndroidGraphicsLiveWallpaper

marko-vasic:
There was a problem with a possible deadlock occurrence in AndroidGraphics and AndroidGraphicsLiveWallpaper classes.
The issue is that a running thread can execute _notify_ before another thread executes _wait_ on the synch object, resulting in a deadlock.
The solution I propose is using semaphores for synchronization instead of mutexes.
Benefit of semaphores is that they internally remember number of permits allowed, 
and thus thread that _signals_ (releases) a semaphore will allow a thread that later arrives to the semaphore to use that permit,
unlike mutexes where waiting thread will be unblocked only by future notifications.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -356,7 +356,7 @@ public class AndroidGraphics implements Graphics, Renderer {
 			pause = true;
 			while (pause) {
 				try {
-                    synchSem.acquire();
+					synchSem.acquire();
 					if (pause) {
 						// pause will never go false if onDrawFrame is never called by the GLThread
 						// when entering this method, we MUST enforce continuous rendering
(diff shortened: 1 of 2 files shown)


MrStahlfelge:
@cypherdare in #5312 you said you wanted to test this PR, where you able to do that?

SimonIT:
Did #6074 maybe fix this problem? Or this a different issue?

cypherdare:
@MrStahlfelge I recall testing it not by building it myself, but by copy-pasting a clone of this class and using it in my project. It continued to deadlock with the Semaphores.

But merged PR #6074 is supposed to fix this issue. I have a particular app that frequently has the deadlock on a Google Pixel. I've installed my app on that phone after building it with 1.9.11. I'll periodically check it over the next week or two and see if the #6074 fix holds up. If so, I can close these other PRs/issues related to the deadlock issue.

I don't know for sure if #5042 is the caused by the same problem, but I don't recall experiencing it in the past couple years, so I'll at least close that one.

ufoq:
> Did #6074 maybe fix this problem? Or this a different issue?

It should fix issue at least for non-Wallpaper version, however I haven't do extensive test with wallpaper-version, so I'm not sure, but by looking just at code it should work (or maybe you need to port the same logic to AndroidGraphicsLiveWallpaper, I don't remember if they share cleanup method code).

cypherdare:
I can study it more closely in a few days and see whether similar changes should be carried over to wallpapers.

cypherdare:
I haven't had a chance to look at the wallpaper solution yet, but so far have had no issues with #6074 so I'm closing this one.

PR #4331, opened 2016-09-26
Fixes #4322: Restore window on the monitor it was before going fullscreen.

### LINKED ISSUE #4322: Lwjgl3: When going from fullscreen to windowed, the window isn't restored on the correct monitor
#### Issue details
The window isn't restored on the correct monitor if going from fullscreen to windowed.

#### Reproduction steps/code
Execute, immediately drag the window to any non-primary monitor.

```java
import com.badlogic.gdx.*;
import com.badlogic.gdx.graphics.*;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;

public class BarebonesBatch extends ApplicationAdapter {
    SpriteBatch batch;
    Texture texture;

    public void create () {
        batch = new SpriteBatch();
        texture = new Texture(Gdx.files.internal("data/badlogic.jpg"));

    }

    int i = 0;

    public void render () {
        i++;
        if (i == 100000) {
            Gdx.graphics.setFullscreenMode(Gdx.graphics.getDisplayMode());
        } else if (i == 200000) {
            Gdx.graphics.setWindowedMode(400,400);
        }
        batch.begin();
        batch.draw(texture, 100, 100);
        batch.end();
    }

        new LwjglApplication(new BarebonesBatch());
    }
}
```

1.9.4

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

CreamyCookie:
Instead of centering the window when going from fullscreen to windowed, it is now placed at the position it was before we switched to fullscreen. This, I think, is a more intuitive solution, because if the user has positioned the window in a certain way, then this shouldn't be lost, when switching between windowed and fullscreen.

This works, because in Lwjgl3 (or GLFW) all monitors are [part of a virtual screen coordinate system](http://www.glfw.org/docs/latest/intro_guide.html#coordinate_systems). The stored positions are in that coordinate system.

-----

For cases, where developers want to center their window, we could do the following, which is based on the previous code in `setWindowedMode`:

```java
	public void centerWindowOnMonitor(Monitor monitor) {
		if (!isFullscreen()) {
			DisplayMode currentMode = getDisplayMode(monitor);
			int windowPosX = monitor.virtualX + (currentMode.width - getWidth()) / 2;
			int windowPosY = monitor.virtualY + (displcurrentModeayModeBeforeFullscreen.height - getHeight()) / 2;
			window.setPosition(windowPosX, windowPosY);
		}
	}
	
	public void centerWindowOnCurrentMonitor() {
		centerWindowOnMonitor(getMonitor());
	}
```

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
@@ -51,6 +51,9 @@ public class Lwjgl3Graphics implements Graphics, Disposable {
 	private long frameCounterStart = 0;
 	private int frames;
 	private int fps;
+	private int windowPosXBeforeFullscreen;
+	private int windowPosYBeforeFullscreen;
+	private DisplayMode displayModeBeforeFullscreen = null;
 
 	IntBuffer tmpBuffer = BufferUtils.createIntBuffer(1);
 	IntBuffer tmpBuffer2 = BufferUtils.createIntBuffer(1);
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Works great, thanks!

PR #4333, opened 2016-09-27
Add Matrix4::hasRotationOrScaling to GWT backend

Zomby2D:
Following #4319, a ``hasRotationOrScaling`` method was added to ``Matrix4``.

However, the method had not been replicated in the GWT backend, causing it to stop compiling properly.


xoppa:
Thanks!

PR #4337, opened 2016-10-01
Fix Gradle build problems cause by the retooling

Zomby2D:
Since the Gradle retooling, the source won't build under Gradle anymore.

This is the minimum changeset required to get the command-line Gradle build working again. It's not enough to Sync the project or run the demos under Android Studio but it's a step in the right direction, bringing back command-line tools.

### DIFF
--- a/extensions/gdx-controllers/gdx-controllers-android/build.gradle
+++ b/extensions/gdx-controllers/gdx-controllers-android/build.gradle
@@ -1,0 +1,20 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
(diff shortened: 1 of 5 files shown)


Tom-Ski:
This already builds on command line, in IDEA and eclipse in the way it should. Perhaps pastebin your error that shows what problems this is trying to solve.

Tom-Ski:
I've pushed the missing gradle files in my latest commit, and addressed the issue which I think you were trying to with lint.

If you have an issue importing, please create an issue.

Zomby2D:
Everything is working now under Android Studio 2.2. Thanks for applying the fixes.

PR #4342, opened 2016-10-02
TexturePacker$Settings now correctly copies array fields from another instance

metaphore:
`TexturePacker$Settings#set(Settings)` used to copy array fields by reference. Now it creates new arrays for own fields.


metaphore:
Hey @NathanSweet, can you please revisit and merge this one, if there is no problem?

NathanSweet:
Thank you! Sorry for the delay.

PR #4345, opened 2016-10-05
Add methods to calculate average for all numeric data types

TertyTree:


Darkyenus:
- Unrelated `@Override`s should be probably removed from this PR
- Methods themselves are IMHO too verbose, the comments are useless
- Storing sums inside corresponding type will be prone to overflows, especially in the case of byte (do we even need average of bytes?), it would probably be better to use _at least_ `int`s for small types
- It would be better to round properly and do the division using floating point numbers (for example `2,2,2,2,1` will now return average 1, when it is arguably 2 (1.8))

Otherwise looks good!

TertyTree:
@Darkyenus thanks for the comments, fixed the last 3 points I believe, as for the first point, how can I remove that commit?

I want to keep the byte average method for in case, I've had to use it several times but I probably use the byte data type a lot more often than a lot of people.

cypherdare:
A comment on usefulness...I can't see myself using these methods, since they require allocating an array on each call. My numbers tend to be in Collections, if in a collection/array at all.

xoppa:
Thanks for your contribution. Unfortunately it is indeed not very useful except for a very specific use-case (like rounding the result, using double precision, creating an array on every call). Averaging an array of numbers is pretty trivial though.

PR #4350, opened 2016-10-07
Added containsAll method for Array

Enigo:
Useful method to check if one array contains all values of another array.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -171,6 +171,16 @@ public class Array<T> implements Iterable<T> {
 		return false;
 	}
 
+	/** Returns if this array contains all values.
+	 * @param identity If true, == comparison will be used. If false, .equals() comparison will be used.
+	 * @return true if array contains all values, false if it doesn't */
+	public boolean containsAll (Array<T> values, boolean identity) {
+		for (T value : values) {
+			if (!contains(value, identity)) return false;
+		}
+		return true;
+	}
+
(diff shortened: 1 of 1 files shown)


cypherdare:
Doesn't look very optimized for something that could be iterating over many objects, although maybe we could expect the compiler to inline the `contains` method call and branch prediction would take care of the repeated checking of the `identity` parameter. 

Why not replace Array with Iterable so it's more flexible? Although in my opinion, this is still too niche for LibGDX. If I'm wrong about that, shouldn't this method be added to all of the util collections?

Darkyenus:
I agree with cypherdare, and to add: if the signature stays with `Array`, it should not use its iterator (as it now implicitly does) because it could lead to nasty surprises when using it inside loops already iterating through the parameter array, as the `Array#iterator()` does not support nesting. A minor problem, but it should be better optimized anyway.

Enigo:
Hi guys! Thank you for comments. Changed to Iterable. Not sure that this method should be added to the all util collections though.

NathanSweet:
I'm not sure about this method. Is it really useful often enough to warrant a method to replace 2-3 lines? What about contains none, or whatever particular behavior your app wants? I think I vote for not including the method.

Also, it uses an iterator, which it shouldn't do so that the caller keeps control over nested iterator (re-)use.

badlogic:
I'm not im favor of this for an additional reason besides the use of
iterators. It's complexity is quadratic. In many cases an application
specific context would allow a better way to test for intersection.

NathanSweet:
Good point. I think we won't merge.

PR #4358, opened 2016-10-11
Add splitFrames on TextureRegion

LucianoPC:
- This feature makes possible to split a frame by
  determinated rows and cols

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/TextureRegion.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/TextureRegion.java
@@ -284,4 +284,29 @@ public class TextureRegion {
 		TextureRegion region = new TextureRegion(texture);
 		return region.split(tileWidth, tileHeight);
 	}
+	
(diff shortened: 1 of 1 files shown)


krukhlis:
Hi, just wondering if it should be so streightforward. May be it should also expect some indentX, indentY params to the method -- e.g. if tiles are not glued pixel to pixel but have some border pads between them. 
Just to make API more functional and generic.

So instead of  `public static TextureRegion[][] splitFrames(Texture texture, int rows, int cols)
` there will be  `public static TextureRegion[][] splitFrames(Texture texture, int rows, int cols, int indentX, int  indentY)` or something like that.

cypherdare:
To me this seems a bit too niche to be in the framework, and it really only saves you one line of code. Also, there are four different likely orders people might have placed animation frames, assuming that's what this is for, and this is arbitrarily picking one of them. (across->down, down->across, across->up, across->down)

Tom-Ski:
Thanks for the PR, but as @cypherdare said its a little to niche. Its important to keep the API clean, and bloatless wherever possible.

PR #4361, opened 2016-10-12
Fixes #4164, adds FilterableVehicleRaycaster

### LINKED ISSUE #4164: btRaycastVehicle wheels never collide with objects part of collision group
#### Issue details
The btRaycastVehicle chasis correctly collides with other objects of the same collision group, however the wheels will always fall through the surface. This means the btRaycastVehicle only works if collision groups and filters are never used in the world.  This is briefly discussed on the Bullet forums:
http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=5494&start=0

#### Reproduction steps/code
This can be observed by creating a modified VehicleTest:
```
public class VehicleFilterTest extends VehicleTest {

	@Override
	public BulletWorld createWorld() {
		//Force all objects to same collision group and filter
		final short filterGroup = (1 << 11);
		final short filterMask = filterGroup;
		
		BulletWorld filterWorld = new BulletWorld() {
			@Override
			public void add (final BulletEntity entity) {
			super.add(entity);
			if (entity.body != null) {
				if (entity.body instanceof btRigidBody)
					((btDiscreteDynamicsWorld)collisionWorld).addRigidBody((btRigidBody)entity.body,filterGroup,filterMask);
				else
					collisionWorld.addCollisionObject(entity.body,filterGroup,filterMask);
				// Store the index of the entity in the collision object.
				entity.body.setUserValue(entities.size - 1);
			}
			}
		};
		return filterWorld;
	}
}
```

The fix is to override the rayCast() function of the btVehicleRaycaster that is given in the btRaycastVehicle constructor and set the collision group and filter of the ray before performing a rayTest.

I wrote out what _should_ be potential solution below. Only the new setCollisionFilterMask(int) and setCollisionFilterGroup(int) functions would have to be exposed through the JNI.
```
//------------------------------ btFilterableVehicleRaycaster.h-------------------------------------
#ifndef BT_FILTERABLE_VEHICLE_RAYCASTER_H
#define BT_FILTERABLE_VEHICLE_RAYCASTER_H

#include "LinearMath/btVector3.h"


class btFilterableVehicleRaycaster : public btDefaultVehicleRaycaster
{
public:

	short int m_collisionFilterMask;
	short int m_collisionFilterGroup;

public:
	virtual void setCollisionFilterMask(short int collisionFilterMask);
	virtual void setCollisionFilterGroup(short int collisionFilterGroup);

};

#endif // BT_FILTERABLE_VEHICLE_RAYCASTER_H


//------------------------------ btFilterableVehicleRaycaster.cpp-------------------------------------
#include "btFilterableVehicleRaycaster.h"

btFilterableVehicleRaycaster(btDynamicsWorld* world)
		:m_dynamicsWorld(world), 
		m_collisionFilterMask(btBroadphaseProxy::AllFilter),
		m_collisionFilterGroup(btBroadphaseProxy::DefaultFilter)
		
	{
	}
	
void btFilterableVehicleRaycaster::setCollisionFilterMask(short int collisionFilterMask)
{
	m_collisionFilterMask = collisionFilterMask;
}

void btFilterableVehicleRaycaster::setCollisionFilterGroup(short int collisionFilterGroup)
{
	m_collisionFilterGroup = collisionFilterGroup;
}

void* btFilterableVehicleRaycaster::castRay(const btVector3& from,const btVector3& to, btVehicleRaycasterResult& result)
{

	btCollisionWorld::ClosestRayResultCallback rayCallback(from,to);

	rayCallback.m_collisionFilterMask = m_collisionFilterMask;
    rayCallback.m_collisionFilterGroup = m_collisionFilterGroup;
		
	m_dynamicsWorld->rayTest(from, to, rayCallback);

	if (rayCallback.hasHit())
	{
		
		const btRigidBody* body = btRigidBody::upcast(rayCallback.m_collisionObject);
        if (body && body->hasContactResponse())
		{
			result.m_hitPointInWorld = rayCallback.m_hitPointWorld;
			result.m_hitNormalInWorld = rayCallback.m_hitNormalWorld;
			result.m_hitNormalInWorld.normalize();
			result.m_distFraction = rayCallback.m_closestHitFraction;
			return (void*)body;
		}
	}
	return 0;
}
```

I'm working on setting up an build environment for the native sources, but if someone already has one set up is able to submit a PR before then, it would be greatly appreciated.

jsjolund:
Fixes #4164. Adds a new ```FilterableVehicleRaycaster``` C/C++ class to the custom Bullet wrapper classes and a generated Java class by adding it to the ```dynamics.i``` interface file.

Includes a test/demo called ```VehicleFilterTest``` which extends the existing ```VehicleTest```.

The wrapper code compiles on the Linux and Android platform (Windows and OSX is untested). In order to compile for Android we had to run the ```BulletBuild.java``` program, which regenerated a file called ```Android.mk```. Presumably the automated build system will do this, otherwise we can add it to the PR.

The demo has been tested on desktop, but we were unable to test on Android because of problems when packaging the Bullet libraries. Neither has it been tested on the other platforms. If someone else could test on Android and the other platforms it would be great.


xoppa:
Thanks, looks good! Please include the files generated by BulletBuild indeed (everything which is already in the repo needs to be up to date).

jsjolund:
@xoppa Thanks for the input. The test could not be modified exactly as you suggested. ```VehicleTest#create()``` needs to create the car, and also call super method ```BaseBulletTest#create()```. If the ```VehicleFilterTest``` should extend ```VehicleTest```, there is no way to avoid creating the first car while also calling ```BaseBulletTest#create()```. Instead, we made both tests extend an  ```AbstractVehicleTest```, which does not override the ```BaseBulletTest#create()``` method, but leaves it to the extending classes.

jsjolund:
Changed the tests as per your suggestion @xoppa, thanks!

jsjolund:
Changed class name to FilterableVehicleRaycaster, added info to CHANGES

xoppa:
Thanks!

PR #4377, opened 2016-10-25
Optimized getFirstKeyframeIndexAtTime() to use a Binary Search

gf11speed:
This is a 3D animation optimization that helps the performance of animation updates that happen every frame. The function getFirstKeyframeIndexAtTime() was very slow when iterating through to get keyframe indices when called every frame. Converting this to a Binary Search helps animation performance, especially in more complex models with more joints/bones.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@
 # The email address is not required for organizations.
 Mario Zechner <badlogicgames@gmail.com>
 Nathan Sweet <nathan.sweet@gmail.com> 
+Greg Fields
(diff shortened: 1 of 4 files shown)


AUTHORS
```diff
@@ -8,3 +8,4 @@
 # The email address is not required for organizations.
 Mario Zechner <badlogicgames@gmail.com>
 Nathan Sweet <nathan.sweet@gmail.com> 
+Greg Fields
```
MobiDevelop: You should add your name to the contributors file instead.

gf11speed: Not a problem. It made that change automatically, but I reverted it.

MobiDevelop:
Alternative PR was merged, closing this one out.

PR #4382, opened 2016-10-28
Fixes #4495 (wrong Label pref height)

### LINKED ISSUE #4495: Wrong Label layout for scaled BitmapFonts
**Issue Details**
This issue is associated to the PR #4382 which provides a fix for it.

Label prefHeight is not correctly calculated if underlying BitmapFont scale Y is different to 1.

This affects any layout using Labels that use a BitMap font with a scaleY != 1.

An example:
example

**Reproduction steps/code**
Please see the SSCCE attached http://pastebin.com/EB7mv2AU

1.9.5

- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

obigu:
Pref height is not correctly calculated if BitmapFont scale Y is different to 1.

Label only takes into account fontScaleY correction for the descent value assuming the underlying Bitmap y font scale is 1. The correction should be the relation between the BitmapFont and Label scales.

Find SSCCE test attached that illustrates the issue

http://pastebin.com/EB7mv2AU

Before the fix

java 2016-10-30 14-24-51-818

After the fix

java 2016-10-30 14-23-17-652

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
@@ -230,7 +230,11 @@ public class Label extends Widget {
 
 	public float getPrefHeight () {
 		if (prefSizeInvalid) scaleAndComputePrefSize();
-		float height = prefSize.y - style.font.getDescent() * fontScaleY * 2;
+		float descentScaleCorrection = 1;
+		if (fontScaleChanged) {
+			descentScaleCorrection = fontScaleY / style.font.getScaleY();
+		}
+		float height = prefSize.y - style.font.getDescent() * descentScaleCorrection * 2;
 		Drawable background = style.background;
 		if (background != null) height += background.getTopHeight() + background.getBottomHeight();
 		return height;
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
@@ -231,9 +231,7 @@ public class Label extends Widget {
 	public float getPrefHeight () {
 		if (prefSizeInvalid) scaleAndComputePrefSize();
 		float descentScaleCorrection = 1;
-		if (fontScaleChanged) {
-			descentScaleCorrection = fontScaleY / style.font.getScaleY();
-		}
+		if (fontScaleChanged) descentScaleCorrection = fontScaleY / style.font.getScaleY();
 		float height = prefSize.y - style.font.getDescent() * descentScaleCorrection * 2;
 		Drawable background = style.background;
 		if (background != null) height += background.getTopHeight() + background.getBottomHeight();


tommyettinger:
I'm curious about this. I rely a lot on Labels in my codebase, and I'm hoping this doesn't undo the fix to the bug that got into 1.9.4 . I didn't notice any issues when running the nightly before this PR, but I might not be using preferred height much. Do you have any screenshots to show the changes, before and after?

tommyettinger:
Also, `style.font.getScaleY()` can never return 0, I hope, because if it can this could divide by 0.

obigu:
Don't mean to be rude but I expect a portion of the effort it takes to investigate the issue, create PR and the SCCEE on the comments on it. 

Have you tried to run the SCCEE or have a look at the file changes before asking?

I'm the one who created #4270 if that's the fix you are referring to, they're unrelated.

Preferred sizes are used most times you use a layout (i.e. put a Label inside a Table), that's how Scene 2D works.

No, style.font.getScaleY() can't return 0, again, have you even checked BitmapFont source code before asking? The argument check with the IllegalArgumentException 
is the first line on the setScale() method...

tommyettinger:
That sounds... like you mean to be rude. I'm only concerned since you're also the one who created #4233 , and while I thank you for addressing that in #4270, I don't want another subtle bug to creep into LibGDX. I don't have a suitable setup for building LibGDX itself, so I can't run this code on my own to check if my rendering is affected at all. So, I'm guessing that's a "no" on the screenshot, which to be honest would be helpful in judging exactly what this changes.

And a comment is not a pull request. It's a comment. There is no effect I can have on the LibGDX codebase with a comment, while there is with a pull request, and that's why there's substantial effort required for PRs. I want this merged if it works, but is it wrong to want it validated in case it *doesn't*?

obigu:
Really, I don't. You are questioning the PR just based on the user that has submitted, not on any work or effort on your side.

I've added the screenshots you request. Setting up libGDX as an IDE project takes 5 minutes https://github.com/libgdx/libgdx/wiki/Building-libgdx-from-source and you'll be able to run tests and contribute.

tommyettinger:
Screenshots look good, they definitely show the problem before and the fix after. I'm wondering where the devs are on this. The code doesn't seem like it has any possibilities for introduced bugs, assuming the font isn't some strange user-made subclass of BitmapFont that can have 0 for scale, and that would have its own issues elsewhere. Only feedback on code is that it could avoid the temporary variable by using a ternary conditional in place of descentScaleConditional, changing the code to
`float height = prefSize.y - style.font.getDescent() * (fontScaleChanged ? fontScaleY * 2f / style.font.getScaleY() : 2f);`

obigu:
Inlined conditional as per original source code style.

@NathanSweet Could you please have a look at this?

NathanSweet:
Thanks for the thorough PR! Cheers!

Issue #2820, opened 2015-02-08
ProgressBar Knob optional rounding?


Mirre:
I am not 100% sure if it is already possible to disable rounding on the Knob of the ProgressBar.

Doesn't seem so since it gets cast into a int when it it is drawn. 

https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ProgressBar.java#L157

https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ProgressBar.java#L186

The problem with the rounding is if you are gonna have a moving camera it rounds the position and so the knob "lags"/looks hacky.

For other classes like BitmapFont you can do:

font.setUseIntegerPositions(false);

And for some others I think there exists setRounding(boolean).

It is not a problem if it can't be added, Just that I spent some time trying fix the "hacky" looking knob before checking the source.

NathanSweet:
Most people don't move the UI camera around, so this probably hasn't been
needed yet.

Maybe you could make a PR to add a `setRound` method? It should round
everywhere that is appropriate for both horizontal and vertical progress
bars (and sliders!).

`setRounding` would have been a better name, but a number of scene2d.ui
classes already use `setRound`. At least it is still better than
`setUseIntegerPositions`! :D

NathanSweet:
Merged PR, thanks!

PR #4384, opened 2016-10-29
Fix ProgressBar rounding and allow it to be disabled

cypherdare:
ProgressBar currently always rounds positions and sizes towards zero when drawing its drawables. Also, there was a request to allow rounding to be disabled in #2820, which can be useful if the UI is not pixel perfect. For example, you might be drawing your UI elements enlarged on a high resolution screen, but want to keep the motion of the progress bar looking smooth. Or you might be drawing it scaled down with trilinear filtering and want smoother looking motion.

I kept the same rounding behavior as Table, which doesn't try to round its own bounds (leaving that to the parent). So in the case of  a vertical progress bar, this means the background drawable's vertical position and height are not rounded. And vice versa for horizontal.


NathanSweet:
Cheers!

PR #4385, opened 2016-10-29
Restore ParticleEffect scale when returning to pool (Fixes #2893)

### LINKED ISSUE #2893: ParticleEffect.scaleEffect(float scaleFactor) causes pooled FX to not reset() correctly
Hi.
i have constant issues with pooled particleEffects / emitters.

after some investigation i found out that ParticleEffect.scaleEffect(float scaleFactor) causes pooled FX to not reset() correctly.

here is a libgdx-barebone-desktop-app to test it:
http://pastebin.com/wmMBiiTv

here are the .png and .p files. put them in assets into a subdirectory "particleFX" 
https://www.dropbox.com/sh/o014eeg53y6cymn/AACJFE-wwFhQ_hgiowqNsrtea?dl=0
(not sure if this works)

alternative:
this is the .p file. (name it "test_explosion.p" and put it into assets/particleFX/
http://pastebin.com/826YZQnm

put the standard particleImage "particleExhaust.png" there, too.

app:
press SPACE
and activate / deactivate the line with setScale...

bye
phil

cypherdare:
Fixes #2893.

The scaleEffect() method is not easily reversible in the ParticleEffect class, unfortunately. But the PooledEffect class has easy access to the original parameters in the prototype effect in the pool, which is the common way ParticleEffects are used anyway.

Since the ParticleEffect super class is unchanged, this fix won't break code that was relying on `reset()` not affecting scale.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEffect.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEffect.java
@@ -38,6 +38,7 @@ public class ParticleEffect implements Disposable {
 	private final Array<ParticleEmitter> emitters;
 	private BoundingBox bounds;
 	private boolean ownsTexture;
+	protected float sizeScale = 1f, motionScale = 1f;
 
 	public ParticleEffect () {
 		emitters = new Array(8);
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEffect.java
```diff
@@ -226,7 +226,11 @@ public BoundingBox getBoundingBox () {
 		return bounds;
 	}
 
+	/**Scales all the parameters of all the emitters in this effect. The original values are not
+	 * recovered when {@link #reset()} is called, unless this effect came from a {@link ParticleEffectPool},
+	 * in which case it uses the pool's prototype effect to restore the original values.*/
 	public void scaleEffect (float scaleFactor) {
+		// If this method is ever modified,
```
Darkyenus: Forgotten comment?

cypherdare: That was intentional. It's kind of a hacky method. If anybody ever changes the way it works, I figure it wouldn't be obvious that the reset method in the subclass would also need to be updated.

cypherdare: Oops, now I see...I never finished writing the comment. I'll just remove it. The javadoc should lead future contributors to the right place.

cypherdare:
Forgot to comment before, but I came back and refactored this to make the `scaleEffect` method less of a manual hack.

PR #4391, opened 2016-11-01
Buttons keep skin reference

### LINKED ISSUE #4387: getSkin() returns null in Button subclasses
The documentation for `Table.getSkin()` says it will return the skin that was passed into the constructor. However, some/all of the subclasses of Button that have a Skin parameter in their constructors do not pass this skin to the super constructor in Button. This makes `getSkin()` return null, thereby contradicting the documentation and making these classes a bit less handy when subclassing them.

To reproduce: 
Create an ImageButton using a constructor that takes a Skin. Call `getSkin()` on it and observe that it returns null.

I'll take a stab at a PR in the next few days.

cypherdare:
Fixes #4387 

These few constructors (out of all the Button and Button subclass constructors) were missing this pass-through of the Skin reference, causing `getSkin()` to return null when it shouldn't (that is, when a skin was passed to the constructor).


xoppa:
Thanks!

PR #4395, opened 2016-11-02
Clean up VertexAttribute and expose expert constructors

cypherdare:
This builds off of #3936. The expert constructors can be made public with a bit of clean-up to avoid potential pitfalls and clarify the docs. Users will still need to understand the parameters that go into `glVertexAttribPointer`, but I don't think people will try to use any of the constructors unless they know what they're for.

@xoppa 
I think MeshBuilder might be incompatible with customized VertexAttributes, based on a quick look at its `begin()` method. I'm not very familiar with MeshBuilder, but you might want to add a doc clarifying that it only supports the built-in VertexAttributes. I didn't test it, but maybe it would be compatible so long as one of the built-in Usages is used.

Most people will never need these features, but they might open up some ways to optimize mesh vertex sizes. I've tried doing that before and this would have helped.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Mesh.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Mesh.java
@@ -1027,7 +1027,7 @@ public class Mesh implements Disposable {
 					if (a == null) continue;
 					for (int j = 0; j < a.numComponents; j++)
 						checks[++idx] = (short)(a.offset + j);
-					attrs[++ai] = new VertexAttribute(a.usage, a.numComponents, a.alias);
+					attrs[++ai] = new VertexAttribute(a.usage, a.numComponents, a.type, a.normalized, a.alias, a.unit);
 					newVertexSize += a.numComponents;
 				}
 			}
--- a/gdx/src/com/badlogic/gdx/graphics/Mesh.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Mesh.java
@@ -1027,7 +1027,7 @@ public class Mesh implements Disposable {
 					if (a == null) continue;
 					for (int j = 0; j < a.numComponents; j++)
 						checks[++idx] = (short)(a.offset + j);
-					attrs[++ai] = new VertexAttribute(a.usage, a.numComponents, a.type, a.normalized, a.alias, a.unit);
+					attrs[++ai] = a.copy();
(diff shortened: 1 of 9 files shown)


gdx/src/com/badlogic/gdx/graphics/Mesh.java
```diff
@@ -1027,7 +1027,7 @@ public Mesh copy (boolean isStatic, boolean removeDuplicates, final int[] usage)
 					if (a == null) continue;
 					for (int j = 0; j < a.numComponents; j++)
 						checks[++idx] = (short)(a.offset + j);
-					attrs[++ai] = new VertexAttribute(a.usage, a.numComponents, a.alias);
+					attrs[++ai] = new VertexAttribute(a.usage, a.numComponents, a.type, a.normalized, a.alias, a.unit);
```
xoppa: Looks like this would be better suited in a copy method. That would also hold when the class is extended.

gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
```diff
@@ -184,11 +147,27 @@ public int getKey () {
 		return (usageIndex << 8) + (unit & 0xFF);
 	}
 	
+	/** @return How many bytes this attribute uses. */
+	public int getNumBytes () {
```
xoppa: Perhaps getSize or getSizeInBytes would be a better name?

gdx/src/com/badlogic/gdx/graphics/Mesh.java
cypherdare: Agreed, that's cleaner and safer.

gdx/src/com/badlogic/gdx/graphics/VertexAttribute.java
cypherdare: I'll go with getSizeInBytes to avoid confusion with VertexAttributes.getSize() which returns number of components.

cypherdare:
I added support for shorts to VertexArray, and a test. Correct me if I'm wrong, but I didn't see a good reason for using the FloatBuffer internally in VertexArray--that's the only thing that was preventing use of shorts. I know the reference is needed for supporting the interface.

xoppa:
Looks great! Thanks @cypherdare !

PR #4405, opened 2016-11-08
Lwjgl3 non-continuous rendering.

cypherdare:
This is an evolved version of #3772 that I worked out with @code-disaster.


cypherdare:
UI change averted. I found that the input handling in the backend was using a sub-optimal order of events. Also, justTouched and justPressed were acting with a one frame delay.

cypherdare:
Rebased/squashed

Issue #4409, opened 2016-11-08
"unknown error 36059" when instantiating FrameBufferCubemap


JustinMullin:
When attempting to construct a simple FrameBufferCubemap on OS/X Yosemite (10.10.5), using LibGDX 1.9.4 stable on the LWJGL/Desktop backend, I'm getting the error below. I have not attempted to reproduce on other platforms.

OpenGL appears to return an error status from `glCheckFramebufferStatus` if it's queried before `glBindFramebuffer`. In the current state of GLFrameBuffer the status is queried before the FB bind, resulting in an error code.

This appears to be regression on https://github.com/libgdx/libgdx/issues/3227. See also the original fix commit at: https://github.com/libgdx/libgdx/pull/3233.

Stack trace:

```
Exception in thread "LWJGL Application" java.lang.IllegalStateException: frame buffer couldn't be constructed: unknown error 36059
at com.badlogic.gdx.graphics.glutils.GLFrameBuffer.build(GLFrameBuffer.java:230)
at com.badlogic.gdx.graphics.glutils.GLFrameBuffer.<init>(GLFrameBuffer.java:118)
at com.badlogic.gdx.graphics.glutils.FrameBufferCubemap.<init>(FrameBufferCubemap.java:88)
at com.badlogic.gdx.graphics.glutils.FrameBufferCubemap.<init>(FrameBufferCubemap.java:75)
at xyz.jmullin.fbcube.error.desktop.FbCubeError.create(FbCubeError.java:11)
at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:147)
at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:124)
```

Minimal application to reproduce:

```
import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
import com.badlogic.gdx.*;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.Pixmap;
import com.badlogic.gdx.graphics.glutils.FrameBufferCubemap;

public class FbCubeError extends ApplicationAdapter {
	public void create () {
		new FrameBufferCubemap(Pixmap.Format.RGB888, 1024, 1024, true);
	}

	public void render () {
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
	}

	public static void main (String[] args) throws Exception {
		new LwjglApplication(new FbCubeError());
	}
}
```

[...]

Issue #3227, opened 2015-06-24
"unknown error 36059" when trying to use a FrameBufferCubemap


haedri:
When trying to use a FramBufferCubemap on a OSX system I get the following error (from the FrameBufferCubemap test) :
```
Exception in thread "LWJGL Application" java.lang.IllegalStateException: frame buffer couldn't be constructed: unknown error 36059
	at com.badlogic.gdx.graphics.glutils.GLFrameBuffer.build(GLFrameBuffer.java:195)
	at com.badlogic.gdx.graphics.glutils.GLFrameBuffer.<init>(GLFrameBuffer.java:111)
	at com.badlogic.gdx.graphics.glutils.FrameBufferCubemap.<init>(FrameBufferCubemap.java:87)
	at com.badlogic.gdx.graphics.glutils.FrameBufferCubemap.<init>(FrameBufferCubemap.java:74)
	at com.badlogic.gdx.tests.g3d.FrameBufferCubemapTest.create(FrameBufferCubemapTest.java:44)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:143)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:120)
```

I noticed that moving the line that gets the frame buffer status a few line below in GLFrameBuffer solves the issue, and the test runs perfectly fine after that (same as on Windows/Linux)

Problem is I have no idea if it is the correct way to do that, so I'll let someone that knows this subject better do a PR :)

There is no problem with standard FrameBuffer

Here's the patch of the modification I did to make the test work :
```
diff --git a/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java b/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
index c9a3648..792f68b 100644
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
@@ -169,11 +169,11 @@
 			gl.glFramebufferRenderbuffer(GL20.GL_FRAMEBUFFER, GL20.GL_STENCIL_ATTACHMENT, GL20.GL_RENDERBUFFER, stencilbufferHandle);
 		}
 
-		int result = gl.glCheckFramebufferStatus(GL20.GL_FRAMEBUFFER);
-
[...]

PR #4411, opened 2016-11-08
Fix for #4409 - 

JustinMullin:
See also: https://github.com/libgdx/libgdx/issues/4409 and https://github.com/libgdx/libgdx/issues/3227.

Deferred frame buffer status query 'til after frame buffer bind. This prevents a FRAMEBUFFER_INCOMPLETE_DRAW_BUFFER_EXT error on some systems.


code-disaster:
As the one guilty of "fixing" this the last time, I need to ask: is there an explanation why this is a fix?

To my understanding, and by the GL docs, this function **"queries the completeness status of the framebuffer object currently bound to target"**. So this call now queries the default framebuffer, which let's assume is complete.

If this change is valid for some reason, may I suggest you leave a comment right there in the source code for the next person like me who tries to mess with it again? ;)

JustinMullin:
So with a fresh pair of eyes I'm looking through this again, and you pose a very good question. I believe I'm guilty of assuming because this resolved my symptoms (a nasty generic error), it must be the cure. My understanding is the same as yours on further consideration. It's probably just hitting the default framebuffer, seeing everything is fine, and moving along with no errors.

What's strange is the framebuffer does indeed work for me if I circumvent the error. I'm inclined to say this is an incorrect fix and there must be something else going on. I'll leave some extra comments on the issue I opened and see where we land.

Thanks!

Issue #4298, opened 2016-09-06
Add setMaximized to Lwjgl3ApplicationConfiguration


CreamyCookie:
Add `setMaximized(boolean maximized)` (and `isMaximized()`) to `Lwjgl3ApplicationConfiguration`, which would set the window to maximized. This should be possible using the window hint `GLFW_MAXIMIZED`. More information: https://github.com/glfw/glfw/blob/master/docs/window.dox
#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

code-disaster:
Added in 3ed4e06.

CreamyCookie:
Thanks @cypherdare !

PR #4412, opened 2016-11-09
Add Lwjgl3 maximized window features

cypherdare:
I put in a callback for maximized state, and options to maximize a window at will or start it maximized.

I renamed `window.deiconify()` to `window.restore()` to reflect that it can be used for two different purposes. Better to rename it now when the backend is young than to make it a quirk forever.

Addresses #4298. I didn't add a `isMaximized()` method to keep it clean, as we don't have `isFullScreen` or `isIconified` window methods, and the callback should be sufficient.


code-disaster:
There was a small memory leak - the GLFW callback needs to be free()'d when closing a window. I also took the liberty to merge maximized() and demaximized() into one single function.

cypherdare:
Good catch. I put `demaximized()` in to mirror `deiconified()`. I wonder if iconified/deiconified should also be condensed into one for symmetry.

code-disaster:
Personally, I'd say yes. Not only to keep the interface less cluttered, but also because demaximize and deiconify are really strange words. ;)

PR #4414, opened 2016-11-09
FrameBufferCubemap fixes

code-disaster:
see #4409 

One from the "how the heck did this ever work?" category.

I open a PR because I'd like some extra eyes on what I changed. Esp. if it's a good idea to rename an attribute alias like this - I reckon it may break user shaders.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/attributes/CubemapAttribute.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/attributes/CubemapAttribute.java
@@ -46,7 +46,7 @@ public class CubemapAttribute extends Attribute {
 
 	public CubemapAttribute (final long type, final Cubemap texture) {
 		this(type);
-		textureDescription.set(texture, texture.getMinFilter(), texture.getMagFilter(), texture.getUWrap(), texture.getVWrap());
+		textureDescription.texture = texture;
 	}
 
 	public CubemapAttribute (final CubemapAttribute copyFrom) {
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/attributes/CubemapAttribute.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/attributes/CubemapAttribute.java
@@ -22,7 +22,7 @@ import com.badlogic.gdx.graphics.g3d.utils.TextureDescriptor;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 
 public class CubemapAttribute extends Attribute {
-	public final static String EnvironmentMapAlias = "environmentMapTexture";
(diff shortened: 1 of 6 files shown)


gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
```diff
@@ -126,6 +126,12 @@ public GLFrameBuffer (Pixmap.Format format, int width, int height, boolean hasDe
 	/** Override this method in a derived class to dispose the backing texture as you like. */
 	protected abstract void disposeColorTexture (T colorTexture);
 
+	protected void bindFrameBufferColorTexture () {
+		GL20 gl = Gdx.gl20;
+		gl.glFramebufferTexture2D(GL20.GL_FRAMEBUFFER, GL20.GL_COLOR_ATTACHMENT0, GL20.GL_TEXTURE_2D,
```
xoppa: Better move 2d texture specific code to the 2d texture implementation and make this abstract instead.

gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
```diff
@@ -126,6 +126,12 @@ public GLFrameBuffer (Pixmap.Format format, int width, int height, boolean hasDe
 	/** Override this method in a derived class to dispose the backing texture as you like. */
 	protected abstract void disposeColorTexture (T colorTexture);
 
+	protected void bindFrameBufferColorTexture () {
```
xoppa: Perhaps use "attach" instead of "bind" to avoid confusion.

xoppa:
Good catch. The alias is not used (or at least should not be used) for anything else than debugging, changing it will not break user shaders. [The uniform itself is already named correctly](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/shaders/DefaultShader.java#L120).

gdx/src/com/badlogic/gdx/graphics/g3d/attributes/CubemapAttribute.java
```diff
@@ -46,7 +46,7 @@ public CubemapAttribute (final long type) {
 
 	public CubemapAttribute (final long type, final Cubemap texture) {
 		this(type);
-		textureDescription.texture = texture;
+		textureDescription.set(texture, texture.getMinFilter(), texture.getMagFilter(), texture.getUWrap(), texture.getVWrap());
```
xoppa: This change is not needed (and unwanted), please undo this.

xoppa:
Thanks!

PR #4416, opened 2016-11-09
Added overload for lookAt

Scellow:


cypherdare:
Doesn't make sense to me to mix floats and a Vector3. It should be all floats or all Vector3s in my opinion.

PR #4420, opened 2016-11-12
Lwjgl3 fix javadoc

cypherdare:
(I feel like the need for requesting rendering in this condition could be very easily overlooked, and there's no harm in doing it automatically.)

Edit: Only the Javadoc needs to be corrected. There is no need to request rendering since it happens anyway.


WickedShell:
You don't need to request rendering when posting a runnable. A runnable will always force a redraw. https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java#L380

cypherdare:
??? 

This has nothing to do with posting RunnableA.

cypherdare:
Runnables. Sorry, on phone.

cypherdare:
You're right--it's just the java doc that needs to be updated.

PR #4422, opened 2016-11-14
Made Array class Poolable for automatic clearing when free in a Pool.

WinterGuardian:
Arrays are used with pools at multiple location in the source code, including the Actor class.


WinterGuardian:
Sorry, messed up my commit

PR #4423, opened 2016-11-14
Added support for configuring statusbar visibility in MOE backend

liliomk:


cypherdare:
Shouldn't the default be false, since LibGDX is primarily used for games?

liliomk:
I was thinking the same, but opted to keep the current behavior.

Tom-Ski:
Lets change this so its hidden by default.

PR #4428, opened 2016-11-19
Added ShaderProgramLoader for AssetManager

cypherdare:
It always bothers me that ShaderPrograms can't be managed along with the rest of the assets. And since they're a synchronously loaded type of asset, it would be nice for animated loading screens if they could also be included in step-by-step loading via AssetManager.

Since typically two files are used, it provides various ways to specify the second file. The easiest (zero configuration) way would be to use .vert and .frag file name suffixes, and queue them to AssetManager.load() using either of the file names.

I derived from AsynchronousAssetLoader for future flexibility. I saw that some other synchronous loaders do the same.


voidburn:
Ty for this! I can take it off from the TODO :D

Issue #4419, opened 2016-11-12
[Improvement] Lwjgl3 winproc handling thread/main loop thread separation


voidburn:
Currently when the window application is dragged it freezes the mainloop, which stops everything that doesn't run in it's own thread with its own independent update loop. This behavior is non damaging for small games, but networked simulations that involve prediction and interpolation don't like large spikes in the delta time of a frame.

Is something like this conceivable?: https://github.com/LWJGL/lwjgl3-demos/blob/master/src/org/lwjgl/demo/opengl/glfw/Multithreaded.java

badlogic:
I've tested resizing and moving the window on macOS with the latest libGDX version from master and it works as intended. The rendering loop is being called as intended.

voidburn:
The problem is on windows (haven't tested linux). I PRed a threaded audio update implementation, but since I'm using a custom backend I can't contribute the renderer (mine does away with multiple windows support).

badlogic:
Will give it a try on Windows then.

voidburn:
I implemented my separate thread for the renderer pretty much like shown in the example above. 

Word of caution if you embark on this: Almost everything obvious works out of the box, but I had to proxy the *Graphics.setCursor() calls, since GLFW states they only work when called from the main thread (the window process if you decouple it from the renderer), until I did that its behavior was in fact unreliable.

I also was too much of a noob to retain multi-windows support, if you come up with a threaded renderer that includes that I look forward to check your implementation! I kept getting complains when trying to switch GL context.

code-disaster:
The stop happens on OS X when clicking and holding the mouse on the window border (like resize, without moving the mouse), or when navigating the application menu. On Windows it also happens when just clicking and holding the mouse button on the window title bar.

I believe it would be overkill (and out of scope) for libGDX to do a separate thread per window. It's already bad enough to properly handle GL resources shared among contexts, it'd be even worse if we invite multi-threaded windowing to the party.

What I could see working is to do GLFW setup on the main thread, then spawn a render thread, and do everything else - from window creation to application listener calls to window shutdown - inside this thread. Since glfwPollEvents() and therefor GLFW callbacks would still need to happen from the main thread, it'd be required to pass events to the render thread somehow, before calling the appropriate ApplicationListener functions.

voidburn:
[...]

PR #4435, opened 2016-11-23
Lwjgl3 backend's OpenAL audio on a dedicated thread

voidburn:
- Synchronized access for the music add/remove/update methods
- Moved dispose method in the dedicated thread (The MockAudio won't use natives that require disposal anyway)
- Testers welcome!


voidburn:
There is currently an issue with Music.setPosition(), looking into it, help appreviated :)

intrigus:
I'm pretty sure, that this is not properly synchronized. You would probably have to synchronize *all* methods.

voidburn:
Tested with AudioDeviceTest and MusicTest, seems to be working. Sadly I don't have a complex sound usage in my engine yet, would definitely like some feedback from someone that hits it hard :)

code-disaster:
I'm worried that we ask for trouble in the long run by doing synchronization like this. It would be a lot cleaner to let the public API write into some postMessage() style queue, and let the audio update thread take care of the execution.

voidburn:
@code-disaster If the multi-threaded render loop happens I'll implement your suggestion on top of that, unless Mario tackles it while looking into it (actually rendering and audio updates could just share the same thread, making this useless).

Anyway, I agree completely, that's how I'm handling Gdx.graphics.setCursor() calls from the APIs in my multi-threaded renderer implementation right now, far cleaner.

ollipekka:
+1 @code-disaster 

I would advise against merging this. This looks like potential source of problems and adds complexity when it could be simple. If you need multi-threaded audio support you could fork the libgdx audio API and roll your own.

sreich:
Silly question here, but when is multithreaded audio needed or desirable?

code-disaster:
@sreich it's one approach to avoid blocking audio updates in case the main thread blocks, e.g. what happens when you try to resize or move the application window (see #4419), or for example during screen transitions when you may have some heavy I/O work to do.

voidburn:
I'm definitely going to close this, since in the end it was not even the correct approach (though I tested it heavily with a LRU cache backing the sound and music banks, and I had zero problems even when the audio backend was concurrently fed multiple playbacks from multiple threads.

Anyway, this was all born from the necessity of having the winproc thread not interfere with the rest of the threads during window drag operations (on windows, and possibly on OSx under some circumstances, the winproc thread is completely frozen during the window drag and until you release it).

I started with the audio backend, then moved to the renderer (in the process losing multi-window rendering btw, just couldn't make it work), I made this pull request thinking it was a good stand-alone feature, and in part because someone left the "FIX THIS: MOVE TO SEPARATE THREAD" comment in the audio loop - or something like that-.

In the end I **haven't used this**, because I moved both the audio AND the renderer together in their own thread, effectively decoupling everything else from the winproc thread, using what I found in #4419 as implementation hint.

The reason I haven't posted that PR is because I ended up further customizing my backend (I don't use gradle for my project) to be able to switch arbitrary versions of LWJGL and plug LWJGL modules as I see fit, while wrapping them in handy Gdx.app.whatevermodule method calls.

Closing this PR, if you ever feel like you need something like this, consider instead using @code-disaster suggestion and provide a thread safe queue to push requests, while having the audio update loop (in whatever thread it lives) peek&poll it.

PR #4438, opened 2016-11-25
+ Added mappings for Xbox 360 Controller on Linux

WilliamHayward:
+ Added button code for pressing in thumbsticks
+ Fixed typo where bumpers were referred to as axis, instead of triggers

Also I started mapping out the Windows versions, but I ran into some issues. Namely that the guide button wouldn't register, and both triggers were "Axis 4", with one being negative (resulting in some weirdness and overlapping values). I can make a pull request with a partially complete set of Windows values if you want though.


DenialAdams:
This + the windows mapping would be very helpful! I hope this gets merged.

FWIW, these match what I found in my testing as well (linux).

Thanks for your work @WilliamHayward

WilliamHayward:
@Tom-Ski - Not sure what "Backend" means specifically in this context, so here's some general information on the circumstances:

OS: Xubuntu 16.04
Controllers: First party and third party wired Xbox 360 Controller
IDE: Eclipse Neon
Java version: 1.8

Hope that helps!

DenialAdams:
For what it's worth, I am using this code with the LWJGL3 desktop backend.

Tom-Ski:
Lwjgl vs Lwjgl3

PR #4441, opened 2016-11-26
Update CONTRIBUTORS.

3xp0n3nt:


3xp0n3nt:
@Tom-Ski This was supposed to be part of #4440.

Issue #2252, opened 2014-08-21
playbackDevice.writeSamples eating CPU


tnelsond:
On my system which is a 64-bit Linux box using Alsa for sound, the call to writeSamples on a AudioDevice maxes out a CPU core if the samplerate is 30khz (30000) or higher; however hardly any at all ( < 10% CPU) is used when the samplerate is 20khz (20000) or lower. 44.1khz and 48khz also maxes out a CPU.

Android seems to have no problem whatsoever.

mgsx-dev:
Issue is not related to audio driver but to java code (see my PR that fix it)

PR #4442, opened 2016-11-26
fixes libgdx/libgdx#2252 reduce CPU usage in OpenALAudioDevice

mgsx-dev:
I had the same problem as in libgdx/libgdx#2252. When writing to audio device, Thread is not sleeping because of a wrong time calculation which result to zero (no sleep at all) with sample rate of 44100 Hz or above.

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OpenALAudioDevice.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OpenALAudioDevice.java
@@ -145,7 +145,7 @@ public class OpenALAudioDevice implements AudioDevice {
 			}
 			// Wait for buffer to be free.
 			try {
-				Thread.sleep((long)(1000 * secondsPerBuffer / bufferCount));
+				Thread.sleep((long)(1000 * secondsPerBuffer));
 			} catch (InterruptedException ignored) {
 			}
 		}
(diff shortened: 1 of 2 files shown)


intrigus:
What exactly is this fixing and why is this a fix for #2252 ?

mgsx-dev:
Sorry, I thought my commit comment body was enough.

When writing to an AudioDevice, OpenALAudioDevice implementation wait for a free buffer to fill using Thread.sleep. With some configurations, wait duration is rounded to zero which lead to poll continuously OpenAL without sleeping and then eat 100% of CPU.

This issue is due to wait duration calculation : `(long)(1000 * secondsPerBuffer / bufferCount)`

Divide by bufferCount doesn't make sense, it should just wait 1000 * secondsPerBuffer which ensure to have a free buffer after this delay.

Default configuration is concerned by this issue : 
* stereo (2 channels)
* PCM 16 bits (2 bytes per sample)
* sample rate of 44100 Hz
* audioDeviceBufferSize = 512
* audioDeviceBufferCount = 9, 

secondsPerBuffer is computed like this : `secondsPerBuffer = (float)bufferSize / bytesPerSample / channels / sampleRate;`

with default configuration : `secondsPerBuffer = 512 / 2 / 2 / 44100 ~= 3 ms`. And thread sleep time is 3/9 rounded to zero.

Once fixed, thread sleep time is 3ms and doesn't eat 100% CPU anymore.

Both lwjgl and lwjgl3 implementations are concerned.

I mentionned Issue #2252 because it describe the problem.

mgsx-dev:
I was surprised this fix was not shipped with last LibGDX release. Do you need more information on this ? how can I help ?

voidburn:
Surprised as well, I'm using your patch in my custom backend and works as intended. So well, thanks!

badlogic:
Mostly because we haven't pulled it in before the release. Will be part of the next release.

mgsx-dev:
OK, I understand, PR was very close to 1.9.5 release.

PR #4443, opened 2016-11-29
OrthogonalTiledMapIterator

BenMcLean:
I wrote an iterator which cycles over visible tiles drawn by OrthogonalTiledMapRenderer. This makes it easy to draw things (like health bars or fire or whatever) on top of visible tiles in an orthogonal tile map.

Also included is a test which demonstrates this working.

It may be desirable to also create similar iterators for other tiled map renderers at some point.


gdx/src/com/badlogic/gdx/maps/tiled/iterators/OrthogonalTiledMapIterator.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public GridPoint2 next() {
+        x++;
+        if (x >= x2) {
+            y--;
+            x = x1;
+        }
+        return new GridPoint2(x, y);
```
MobiDevelop: This is producing a lot of garbage.

gdx/src/com/badlogic/gdx/maps/tiled/iterators/OrthogonalTiledMapIterator.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public OrthogonalTiledMapIterator(OrthographicCamera camera, TiledMapTileLayer layer, float unitScale) {
+        this.camera = camera;
+        this.layer = layer;
+        this.unitScale = unitScale;
+        reset();
+    }
+
+    public OrthogonalTiledMapIterator reset() {
```
MobiDevelop: Feels like the renderer should be providing the visible cells, rather than duplicating the logic to calculate them.

BenMcLean: Sometimes you want just the coordinates, (like when you're drawing things on top of the tiled map which aren't part of the cells, such as some health bars I drew which actually draw some distance above the cells) and if you've got the coordinates then you should be able to get the cells I think.

gdx/src/com/badlogic/gdx/maps/tiled/iterators/OrthogonalTiledMapIterator.java
BenMcLean: True. How would you suggest fixing this?

TEttinger's SquidLib project fixes this by caching x and y coordinates in his Coord class. https://github.com/SquidPony/SquidLib/blob/master/squidlib-util/src/main/java/squidpony/squidmath/Coord.java

gdx/src/com/badlogic/gdx/maps/tiled/iterators/OrthogonalTiledMapIterator.java
BenMcLean: I guess you might call that "coloring outside the lines"? :)

BenMcLean: The problem with just returning the cells is that then the user wouldn't know where the cell they're looking at actually is on the map, unless I'm mistaken.

MobiDevelop: The duplication of the logic is problematic; it would make more sense to add a method to the renderer to get the visible bounds using the logic it already has.

BenMcLean:
This is ready to be looked at again please. :)

BenMcLean:
Any thoughts on how this looks now?

BenMcLean:
Need to work on this more before PR

PR #4444, opened 2016-11-29
Add contains circle to math.Rectangle

tentone:
Hi

Added contains method to math.Rectangle to check if a circle is contained in the Rectangle
It might be helpful 

Cumps

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -199,13 +199,13 @@ public class Rectangle implements Serializable, Shape2D {
 	 * @return whether the circle is contained in the rectangle */
 	public boolean contains (Circle circle) {
 		float xmin = circle.x - circle.radius;
-		float xmax = xmin + 2f * circle.radius;
+		float xmax = circle.x + circle.radius;
 
 		float ymin = circle.y - circle.radius;
-		float ymax = ymin + 2f * circle.radius;
+		float ymax = circle.y + circle.radius;
 		
-		return ((xmin > x && xmin < x + width) && (xmax > x && xmax < x + width))
-			&& ((ymin > y && ymin < y + height) && (ymax > y && ymax < y + height));
+		return (xmin >= x && xmin <= x + width) && (xmax >= x && xmax <= x + width)
+			&& (ymin >= y && ymin <= y + height) && (ymax >= y && ymax <= y + height);
 	}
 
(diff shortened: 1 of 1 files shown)


obigu:
Just a couple of comments. 
1. I would include equality to the comparisons i.e. `<=`. I can see it follows the same logic that `contains(Rectangle)` which doesn't and I would also modify because `contains(Vector2)` does. I believe this is inconsistent.

2. It's minor but imo it is clearer to define 
`xmin = circle.x - circle.radius;`
`xmax = circle.x + circle.radius;`

instead of 

`xmin = circle.x - circle.radius;`
`xmax = xmin + 2f * circle.radius;`

tentone:
@obigu Thanks for the feedback.

Agree, its better as you suggested, it's cleaner, and it removes a completely unnecessary multiplication.

I had already noticed that, and i would prefer to have <= comparison on all three, i changed to <= on this one.

xoppa:
This might be better suited in the [`Intersector`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Intersector.java) class

tentone:
@xoppa Since all the other contains methods are inside the Rectangle class probably most people are going to look there first for this one as well (and maybe never look for it inside the Intersector class),

Also theoretically it's not an intersection verification, since its only true if the circle is fully contained in the rectangle

gdx/src/com/badlogic/gdx/math/Rectangle.java
```diff
@@ -195,6 +195,19 @@ public boolean contains (Vector2 point) {
...
+	 * @return whether the circle is contained in the rectangle */
+	public boolean contains (Circle circle) {
+		float xmin = circle.x - circle.radius;
+		float xmax = circle.x + circle.radius;
+
+		float ymin = circle.y - circle.radius;
+		float ymax = circle.y + circle.radius;
+		
+		return (xmin >= x && xmin <= x + width) && (xmax >= x && xmax <= x + width)
+			&& (ymin >= y && ymin <= y + height) && (ymax >= y && ymax <= y + height);
```
xoppa: since `min <= max`, this can be reduced to: `xmin >= x && xmax <= x + width && ymin >= y && ymax <= y + height`

tentone:
@xoppa True, thanks, since xmax, xmin, ymax and ymin were only used once i removed them as well.

xoppa:
Thanks

PR #4448, opened 2016-12-01
Generic Animation class

cypherdare:
I have found myself copy-pasting the Animation class a few times to support NinePatches or certain types of Drawables. I think PolygonRegions would be another common need. I'm also playing around with multi-texturing for normal-mapped sprites, so an animation made up of TextureRegionPairs would be useful. So it makes sense for the frame type to be generic.

This is one way to do it that shouldn't affect existing code, unless someone is serializing Animations for some reason (although I doubt anyone is trying to serialize TextureRegions, nor does it make sense to serialize assets). I created a new Animation class in the `utils` package (since this theoretically could be used for more than just 2D animation frames, think decals or some sort of state manager). The existing Animation class extends from it and has a pass-through PlayMode class to prevent code breakage, and having to fully qualify PlayMode if the g2d.Animation class has been imported.

Here are some alternative ways to do this. Any opinions?

1. Do it like this PR, except that the existing g2d.Animation class is untouched. Zero chance of code breakage, but now we'd have two classes with a bunch of duplicate code.
2. Do it like this PR, but give the utils.Animation class a different name to avoid confusion. Maybe could also remove the g2d.Animation.PlayMode pass-through class, although that would force people to fix up some code by importing the new class.
3. Keep this PR as is, deprecate the old g2d.Animation class, with a Javadoc explaining the change, and remove it in a year or two. In the long term this may be the cleanest.
4. Rip off the Band-Aid by deleting the g2d.Animation class now.
5. Forget the new class. Just make g2d.Animation generic, and force people to update code to specify &lt;TextureRegion&gt;. This is a nice middle-ground. Code will be broken, but is easy to fix, and it avoids adding a new class.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@
 - Lwjgl3Window now has a maximize() method, and windows can be started maximized using the window or app configuration's setMaximized() method.
 - NinePatch can now be drawn rotated or scaled.
 - NinepatchDrawable is now a TransformDrawable.
+- API Change: g2d.Animation is now generic so it can support Drawables, PolygonRegions, NinePatches, etc. To fix existing code, specify the TextureRegion type in animation declarations (and instantiations in Java 6), i.e. Animation<TextureRegion> myAnimation = new Animation<TextureRegion>(...);
 
 [1.9.4]
 - Moved snapping from ProgressBar to Slider to prevent snapping when setting the value programmatically.
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -20,13 +20,16 @@ import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.utils.Array;
 
 /** <p>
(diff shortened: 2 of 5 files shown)


gdx/src/com/badlogic/gdx/utils/Animation.java
```diff
@@ -0,0 +1,200 @@
+package com.badlogic.gdx.utils;
+
+import com.badlogic.gdx.graphics.g2d.PolygonRegion;
+import com.badlogic.gdx.graphics.g2d.TextureRegion;
```
obigu: Remove unused imports?

NathanSweet:
Having two classes in libgdx with the same name isn't great, especially when they do the same things. I'd say #5, break it for the greater good.

cypherdare:
Actually, it would have been the third Animation class!

I updated this to go with number 5. I'll monitor SO and the forums if this is pulled. :)

Tom-Ski:
This will probably break the demos, so lets do it.

cypherdare:
@Tom-Ski Do we fix demos after the next release or right away?

Tom-Ski:
Now

PR #4449, opened 2016-12-01
Pixmap method chaining

BenMcLean:
Adds method chaining to draw methods in Pixmap


MobiDevelop:
The same changes would be needed in the GWT emulation of Pixmap.

BenMcLean:
I added that change. :)

NathanSweet:
In general I'm not a big fan of method chaining unless it's a builder class (used only to create an instance). It tends to add noise to the API when mixed with non-chaining methods. In some cases it makes sense to do it anyway, eg Table. For Pixmap I'm relatively indifferent, though I definitely would not want to trigger a crusade where we add method chaining to all our APIs.

BenMcLean:
I'm planning on doing procedurally generated pixel art, and I think that's a case where Pixmap would get used kind of like a builder class.

PR #4452, opened 2016-12-03
Added KTX extensions: general utilities for LibGDX application written in Kotlin

czyzby:
See [KTX repository](https://github.com/czyzby/ktx).

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
@@ -249,8 +249,8 @@
 		<name>KTX</name>
 		<description>A set of Kotlin utilities for LibGDX applications.</description>
 		<package>ktx</package>
-		<version>1.9.4-b2</version>
-		<compatibility>1.9.4</compatibility>
+		<version>1.9.5-b1</version>
+		<compatibility>1.9.5</compatibility>
 		<website>https://github.com/czyzby/ktx</website>
 		<gwtInherits></gwtInherits>
 		<projects>
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/data/extensions.xml
@@ -245,4 +245,32 @@
 			</html>
 		</projects>
 	</extension>
+	<extension>
+		<name>KTX</name>
+		<description>A set of Kotlin utilities for LibGDX applications.</description>
+		<package>ktx</package>
+		<version>1.9.4-b2</version>
+		<compatibility>1.9.4</compatibility>
+		<website>https://github.com/czyzby/ktx</website>
+		<gwtInherits></gwtInherits>
+		<projects>
+			<core>
+				<dependency>com.github.czyzby:ktx-actors</dependency>
+				<dependency>com.github.czyzby:ktx-app</dependency>
+				<dependency>com.github.czyzby:ktx-assets</dependency>
+				<dependency>com.github.czyzby:ktx-collections</dependency>
+				<dependency>com.github.czyzby:ktx-i18n</dependency>
+				<dependency>com.github.czyzby:ktx-inject</dependency>
+				<dependency>com.github.czyzby:ktx-log</dependency>
+				<dependency>com.github.czyzby:ktx-math</dependency>
+				<dependency>com.github.czyzby:ktx-scene2d</dependency>
+				<dependency>com.github.czyzby:ktx-style</dependency>
+			</core>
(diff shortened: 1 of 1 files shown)


MobiDevelop:
To be honest, this doesn't feel like a great fit for the setup tool since it will only get you part of the way - there is still a buildscript class path dependency to add as well as applying the kotlin plugin to your builds and adding an additional dependency on the kotlin standard lib.

Maybe I am wrong but I'd think we should strive to have the setup only include options which can work out of the box.

czyzby:
Well, it _does_ work out of the box - you could potentially use Kotlin libraries with Java without any additional plugins or having to write a single Kotlin line. Although I do agree that in this particular case there's little reason to do so: KTX is basically syntax sugar for Kotlin applications using LibGDX. It's definitely not your typical extension that happens to be written in some other language beside Java.

I'll leave it up to you, but I honestly wouldn't mind seeing some extra extensions for other JVM languages in the official setup tool, even if they came with some warnings or links explaining how to fully set them up. It would also be nice if gdx-setup simply allowed the library creators to add extra plugins to the Gradle script - that would solve the problem altogether.

nhydock:
Maybe there could be a checkbox list of jvm language that can be applied through gradle added onto the gdx-setup.

czyzby:
It's certainly [doable](https://github.com/czyzby/gdx-setup). [Wiki](https://github.com/libgdx/libgdx/wiki/Using-libGDX-with-other-JVM-languages) is alright, but the ability to set it up with a few clicks would be great for newcomers.

BenMcLean:
You might want to suggest this over at the [alternative setup project](https://github.com/czyzby/gdx-setup).

czyzby:
Suggest what? KTX is already included there, as well as basic alternative JVM languages support and project templates. I wanted KTX to be present in the official gdx-setup, as I'm guessing it's the most popular project setup tool by far.

BenMcLean:
Oh good. I did not realize it was already included there. You may disregard my comment. :)

haxpor:
Current state, `gdx-setup` has option to generate kotlin project via Advanced->Use Kotlin, it's included in `master` branch already.

Like @MobiDevelop said, it needs to have configured dependencies thus `build.gradle` to be able to use kotlin in the project. But as `gdx-setup` now does that for you, so it's no need :)

FYI: Also related, to have all source file generated as kotlin https://github.com/libgdx/libgdx/pull/4812 waiting to be merged. So combine all of these together, surely better support for kotiln.

crykn:
I don't think it makes sense to include Kotlin libraries as extensions as long as Kotlin is not supported by all libGDX backends. I also agree with the reasoning [by MobiDevelop](https://github.com/libgdx/libgdx/pull/4452#issuecomment-268340394). Closing this.

czyzby:
@crykn Well, Kotlin itself is not supported by every LibGDX backend either, and yet it was added to the official setup tool. If you decided to officially include Kotlin as a language, I don't see why Kotlin-specific extensions couldn't be available via the tool only if the user selects Kotlin in the first place. Especially since KTX is basically an unofficial Kotlin wrapper for LibGDX.

crykn:
@czyzby thanks for pointing out that Kotlin support is nowadays offered by the setup tool.

I'm still not sure about adding Kotlin as extensions as in my opinion this implies Kotlin is fully supported. Whether there is HTML support for Kotlin is a very real source of confusion (see, for example, [here](https://github.com/libgdx/libgdx/issues/4933) or [here](https://www.badlogicgames.com/forum/viewtopic.php?f=11&t=22124)). 

But - we are currently evaluating switching to gdx-liftoff, which, as you know, already has a KTX checkbox. So this ticket is obsolete for another reason.

czyzby:
@crykn I see, thanks. That's pretty cool to hear, since I started this project. :)

If I were to write the alternative [`gdx-setup`](https://github.com/czyzby/gdx-setup/) today (which `gdx-liftoff` forked), I think I'd create a web tool instead - something along the lines of [Spring initializr](https://start.spring.io/). I originally wanted to create an app that would use Kotlin and showcase some of my LibGDX libraries, but I have to admit that a web tool would be much more accessible.

The biggest problems with the unofficial `gdx-setup` that I had were the manual maintenance of the extensions versions and the difficulty of contributing by new users - mostly due to LibGDX extensions that vastly change the project structure. I'm guessing `gdx-liftoff` has some of these problems as well, since AFAIK it still uses `gdx-lml`/`gdx-autumn`. Today I'd just go with a relatively simple web framework like Vue and host it via GitHub pages. Also, I'd include Maven lookup from day one, so that the version of the registered extensions would be fetched automatically, and you could search for any libraries to add them to your project as a user.

Just some thoughts on the subject.

Issue #3796, opened 2016-01-29
Textarea in a ScrollPane


voidburn:
This is basically the same problem discussed in #2320, which hasn't been investigated further. I tried to tackle it, and as suggested I found a way (pretty easy) to have the TextArea re-evaluate it's height based on the text. 

The TextArea pane resized as expected but, as MobiDevelop hinted, something does not work properly, since the ScrollPane where it was nested didn't seem to care about its child changing size, therefore not computing the new child height and updating the scrollbar. I've tried all possible methods that hinted to calling for a re-evaluation of the layout (invalidate, invalidateHierarchy, layout), but nothing made the ScrollPane aware of the TextArea new size.

I can't find a way to identify the problem, but if fixed, I'd be happy to contribute a mode switching for the TextArea which enables it to either behave as it is now (default) or switch to growing its size along with the contained text height. And perhaps a wordwrap on/off mode, since scrolling will be possible at that stage.

I've attached a test which only demonstrates how a TextArea currently fails to work compared to using a Label inside a ScrollPane (drop the class in the gdx-tests-lwjgl project sources and run it). 

Steps towards a solution:

**Step 1** is to make the TextArea resize itself when a new line gets added or removed (I checked the desired size every time the text had to be drawn, but it can easily be done only when needed, such as adding a new character to the text)

**Step 2** is to figure out why the ScrollPane ignores the new size.

[ScrollableTextAreaTest.zip](https://github.com/libgdx/libgdx/files/110058/ScrollableTextAreaTest.zip)

rami-alisawi:
if I understood you correctly, you want ScrollPane to grow in size together with the TextArea?  what is the purpose of ScrollPane then?

voidburn:
Nope, the textarea grows, and being nested inside the scrollPane it should trigger the scroll. This is what happens with almost anything embedded in the scroll pane. It clips the visible rect and computes the scroll amount.

rami-alisawi:
If you replace the ScrollPane actor with a new ScrollPane actor, would the new one realize the bigger size of TextArea and let you scroll?

voidburn:
Cannot try on the spot, I left that idea behind but will try again to embed the scrolling in the future

NathanSweet:
Try this:

```
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
[...]

PR #4454, opened 2016-12-04
Added test to check TextArea behaviour when nested in a ScrollPane

voidburn:
Showcases and tests the implementation of a TextArea inside a ScrollPane (As suggested by Nate in Issue #3796)

It currently cannot properly scroll the full TextArea height, I hope this will help iron out the small issues with such use case, and ultimately work as an example on how to implement a scrollable TextArea.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/TextAreaScrollTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/TextAreaScrollTest.java
@@ -1,0 +1,104 @@
+package com.badlogic.gdx.tests;
+
+import com.badlogic.gdx.Gdx;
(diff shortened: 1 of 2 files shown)


intrigus:
This needs to also be added to the gwt tests.

voidburn:
If it's the stuff in `GwtTestWrapper.java` I didn't add it there because there wasn't a TextAreaTest to begin with. Can you point me in the right direction?

voidburn:
Also, there's this in there, looks like the ScrollPane doesn't play nice with Gwt?

```Java
// new Instancer() {public GdxTest instance(){return new ScrollPaneTest();}}, // FIXME this messes up stuff, why?
```

NathanSweet:
Not sure about that FIXME. Seems OK to merge as is.

PR #4458, opened 2016-12-06
Fix RenderContext Javadoc

cypherdare:
The begin() method states that it assumes default OpenGL states, but actually it enforces them, regardless of what the state was before calling begin().

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderContext.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderContext.java
@@ -39,15 +39,7 @@ public class RenderContext {
 		this.textureBinder = textures;
 	}
 
-	/** Sets up the render context, must be matched with a call to {@link #end()}. Resets the following OpenGL states to their
-	 * defaults:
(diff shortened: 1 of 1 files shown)


NathanSweet:
"OpenGL state" is very vague, as its a huge state machine. The state is everything you can glEnable/Disable, bound textures, and lots more. It looks like this assumes the default OpenGL state, makes changes, then restores the state to the defaults. SpriteBatch does something similar.

cypherdare:
Yeah, I assumed it's talking about the specific states that it manages. A literal reading of the doc would make me think that it's up to me to ensure the affected states are set to defaults before I use this, but that's not the case.

Actually, it theoretically should be unnecessary to reset those parameters, since no LibGDX rendering class modifies these particular states without restoring them in `end()`. All the 2D batches assume as much.

cypherdare:
Added very explicit clarification. Should be useful when determining which parameters need to be changed when customizing in the 3D API. @xoppa

xoppa:
Let's remove the entire sentence.

xoppa:
Thanks!

Issue #4360, opened 2016-10-12
BoundingBox in 2d


Heerbann:
Less an issue but suggestion:

When using the provided methods for checking intersection or contain of the [BoundingBox](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/math/collision/BoundingBox.html) it always checks for validity first. 

`public boolean isValid () {
        return min.x < max.x && min.y < max.y && min.z < max.z;
    }`

this of course returns never true in a strict 2d environment and one has to set max.z > 0 manually. 

a solution could be

`public boolean isValid () {
        return min.x < max.x && min.y < max.y && min.z <= max.z;
    }`

but I dont know about the implications in 3d.

Cheers

xoppa:
Sounds reasonable. It should be consistent among all axes though. Also the javadoc would need to be updated and it should be noted in CHANGES.

Not sure about the implications indeed. Afaik all code paths use `.inf()` for invalid bounds, but it's best to verify that at least by running the tests.

Care to send a pull request?

Heerbann:
#isValid() is really only used for #contains() and #intersects() and in said methods the equal operator is already in use -> for reference: [here](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/collision/BoundingBox.java#L304) and [here ](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/collision/BoundingBox.java#L276) and [here ](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/collision/BoundingBox.java#L296).
I guess that was just overlooked. 

> Care to send a pull request?

not gonna lie, I have no idea how to do that. I would if I could.

kerberjg:
I agree about the need for consistency here, but If we opted for consistency among all axes, a `BoundingBox` with a volume of zero would be then considered valid.

If that's not a problem, I could take care of the PR.

badlogic:
Please submit a PR, closing this out.

PR #4460, opened 2016-12-06
Implements #4360

kerberjg:
Makes [BoundingBox](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/math/collision/BoundingBox.html) compatible with 2D by performing a "*greater or equal*" comparison between `min` and `max` in BoundingBox#isValid().

Naturally, this would cause a BoundingBox with an area of 0 to be considered valid, so a `&& !dim.isZero()` was added to the expression as well in order to prevent that.

EDIT: updated the dim size check


Darkyenus:
Am I missing something or would this mean that only zero-size bounding box is considered valid? Shouldn't it be `!dim.isZero()`?

kerberjg:
@Darkyenus You're right, it was a typo. Just fixed it.

xoppa:
Why did you make an exception for at least one axis to be not zero? If a zero volume (3D) is valid and a zero area (2D) is a valid then I guess that a zero length (1D) should be valid as well.

kerberjg:
@xoppa I actually had in mind that none of them are to be considered valid, but I guess that's a mistake. Upon further consideration it could be considered valid indeed.

gdx/src/com/badlogic/gdx/math/collision/BoundingBox.java
```diff
@@ -210,7 +210,7 @@ public BoundingBox clr () {
 	/** Returns whether this bounding box is valid. This means that {@link #max} is greater than or equal to {@link #min}, with {@link #dim} greater than zero.
```
moonwolf-github: I think you should update comment also.

kerberjg: Done now

xoppa:
Thanks

PR #4464, opened 2016-12-06
Improvements to barycentric methods in GeometryUtils

Heerbann:
I adapted this from the book Realtime Collision Detection by Christen Ericson (http://realtimecollisiondetection.net/) and this thread https://gamedev.stackexchange.com/questions/23743/whats-the-most-efficient-way-to-find-barycentric-coordinates
in theory this should work but my math might be off or even made a conceptual error. 
Furthermore, the caching might be an overkill for 99% of the use cases. And it poses the question if there should be a cached option for the 2d methods too. Finally, the bigger question is if it wouldnt be even a better idea to refactor it into a (new) Triangle class. (maybe a whole new family of Shape3D objects? there are some for bounding volumes, but for example no 3d polygon as far as I can see)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/GeometryUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/GeometryUtils.java
@@ -16,11 +16,131 @@
 
 package com.badlogic.gdx.math;
 
+import com.badlogic.gdx.utils.Pool.Poolable;
+
 /** @author Nathan Sweet */
 public final class GeometryUtils {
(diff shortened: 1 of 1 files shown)


Heerbann:
after some consideration I think I will retract this pull request. On the one hand some parts don't really fit in with the design philosophy, at least not in the current form. On the other hand I am currently working on refactoring the whole math.collision, intersector and all other collision black holes. 
I will reopen a new pull request with my other changes which will hopefully fit in better.

NathanSweet:
Sounds interesting. Note we do have some shape classes, the utils classes tend to be junk drawers where things got added as-needed.

PR #4465, opened 2016-12-07
[WIP] implementation of texture atlas and batch which use 2D Array texture

magicgoose:
This is implementation of texture atlas and batch which use [2D Array texture](https://www.opengl.org/wiki/Array_Texture) instead of multiple regular 2D textures. This allows to have fewer bind and draw calls when drawing regions from huge multi-page atlases.

They require OpenGL [ES] 3+, because TextureArray requires OpenGL [ES] 3+.

Things that do not work yet:

 * Generation of mipmaps
 * Loading of [Z]KTX files into TextureArray, neither from separate files nor from single file.

Things that do work right now:

 * Creation of `TextureArrayAtlas` from regular `.atlas` files which reference PNG images. Images MUST be of same size, this can be enforced in TexturePacker config.
 * Drawing `TextureArrayAtlas.ArrayAtlasRegion` instances with `ArrayRegionBatch`.

I omitted many `draw` call overloads in batch implementation, they can be added later if needed.

I also made a simple test for this stuff which works (tested on two different OS X machines, one with Nvidia, one with AMD) but it's written in Kotlin, should I port it to Java or what do I do?

Also, there's obviously a lot of code duplication against existing libgdx code files, I didn't make any attempts to solve that yet, but when the time comes to merge this work, it should probably be done to some extent too.

Also some code comments left from original files are probably off. In that case they need to be corrected or deleted.


magicgoose:
>All checks have failed
What does it mean? How do I fix it?

gdx/src/com/badlogic/gdx/graphics/g2d/TextureArrayAtlas.java
```diff
@@ -0,0 +1,476 @@
...
+import static com.badlogic.gdx.graphics.Texture.TextureFilter.Linear;
+import static com.badlogic.gdx.graphics.Texture.TextureWrap.ClampToEdge;
+import static com.badlogic.gdx.graphics.Texture.TextureWrap.Repeat;
+
+@SuppressWarnings("ClassWithTooManyConstructors")
+public class TextureArrayAtlas implements Disposable {
+    static final String[] tuple = new String[4];
+
+    private TextureArray textureArray;
+    private final Array<ArrayAtlasRegion> regions = new Array<>();
```
magicgoose: @intrigus done

cypherdare:
If we had a parent class for TextureRegion...

```
public abstract class GLTextureRegion<T extends GLTexture> {
	T texture;
	float u, v;
	float u2, v2;
	int regionWidth, regionHeight;

	//...
}

public class TextureRegion extends GLTextureRegion<Texture> {
    //...
}
```
 
And then created the following:
```
public class TextureArrayRegion extends GLTextureRegion<TextureArray> {
    int arrayIndex;
    //...
}
```

And changed Batch to accept GLTextures and GLTextureRegions instead of Textures and TextureRegions...

Then we wouldn't need a new Batch class and all that code duplication.

magicgoose:
@cypherdare yeah that's true I think.
And also the same for Atlas.

magicgoose:
Forgot about using formatter with special settings, will update the code soon.
Also I figured out how to write tests, so will port my test soon, too.

Tom-Ski:
Closed in favour of https://github.com/libgdx/libgdx/pull/4487

PR #4467, opened 2016-12-08
Removed usage of Pattern and Matcher in BitmapFont

monkeymanboy:
These wouldn't function properly in GWT 2.8.0 so I changed it to simply use substring.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -25,8 +25,6 @@ package com.badlogic.gdx.graphics.g2d;
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.util.StringTokenizer;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.files.FileHandle;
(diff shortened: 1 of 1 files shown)


NathanSweet:
Oops, sorry GWT guys. What if "id=" is not present? Also, indentation got whacked.

monkeymanboy:
@NathanSweet Yeah the I used githubs editor so I didn't notice the indentation. I fixed it and added exceptions.

NathanSweet:
`id` was optional, now it's not. I don't know why `id` wouldn't be there, but the new code isn't the same as the old code.

Have you tested this PR's code?

monkeymanboy:
@NathanSweet Alright made it optional now but I don't see why include it at all if its optional. Anyways yeah it works I've been using it for my own game.

MonsterOfCookie:
Hey!

I discovered this too, however the issue can be fixed by changing the matcher to call to:

`// Expect ID to mean "index".
Matcher matcher = Pattern.compile(".*id=(\\d+)").matcher(line);
if (matcher.find()) {
	String id = matcher.group(1);
	try {
		int pageID = Integer.parseInt(id);
		if (pageID != p) throw new GdxRuntimeException("Page IDs must be indices starting at 0: " + pageID);
	} catch (NumberFormatException ex) {
		throw new GdxRuntimeException("Invalid page id: " + id, ex);
	}
}

matcher = Pattern.compile(".*file=\"(.*)\"").matcher(line);
if (!matcher.find()) {
	Gdx.app.log("BF", "not found");
	throw new GdxRuntimeException("Missing: file");
}
String fileName = matcher.group(1);`

The reason for the issue is the "matcher.matches()" attempts to match the entire line against the pattern. I suspect with GWT emulation it will be the native JS regex attempt which will do the partial match - hence why the error occured in GWT but not in the Desktop module - matcher.matches() always returned false, which I reckon was ok for single page font files. However in the Java world we need to use the "matcher.find()" which will look for the character request.

I have tested that this works for Desktop and GWT.

Providing we want to keep the pattern/matcher solution

MobiDevelop:
Thanks for the effort on this. Decided to go with #4475. Please let us know if the issue is not resolved in the next snapshot.

PR #4468, opened 2016-12-09
Use only https for fetching binaries with Ant

magicgoose:
Trivial but important change. (Because who knows when someone will change binaries in transit in order to grow their botnet).


intrigus:
LGTM

PR #4473, opened 2016-12-10
Use varargs instead of lot of constructors

Scellow:


badlogic:
Varargs allocate an array under the hood. That's not a good idea on Android.

Scellow:
Oh i didn't know that, thanks for the info

PR #4475, opened 2016-12-11
Fix BitmapFont parsing

tdionizio:
#### Issue details
A recent commit https://github.com/libgdx/libgdx/commit/7420bd0f5111b73e2f1b9c5393640adfbc4aadbb changed a few regular expressions on the BMFont parser, which are now giving problems on GWT.

Desktop jvm expects to match the whole string when matching the expression to find the "id=", while GWT matches part of the string. With the current expression desktop fails to match the string (incorrectly) but gwt seems to find a match, leading to the problem when going through "id.substring(3)" when the id already contains the match group with the expected number and returns an empty string.

Desktop jvm also seems to be more forgiving when matching the expression used to find the "file=", but GWT throws an exception when it encounters "*+".

#### Reproduction steps/code
Just load a font in GWT.

1.9.5

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -528,8 +528,8 @@ public class BitmapFont implements Disposable {
 					if (line == null) throw new GdxRuntimeException("Missing additional page definitions.");
 
 					// Expect ID to mean "index".
-					Matcher matcher = Pattern.compile(".*id=(\\d+).*").matcher(line);
-					if (matcher.matches()) {
+					Matcher matcher = Pattern.compile(".*id=(\\d+)").matcher(line);
+					if (matcher.find()) {
 						String id = matcher.group(1);
 						try {
 							int pageID = Integer.parseInt(id);
@@ -539,8 +539,8 @@ public class BitmapFont implements Disposable {
 						}
 					}
 
-					matcher = Pattern.compile(".*file=\"?([^\"]+)\"?.*").matcher(line);
-					if (!matcher.matches()) throw new GdxRuntimeException("Missing: file");
+					matcher = Pattern.compile(".*file=\"?([^\"]+)\"?").matcher(line);
+					if (!matcher.find()) throw new GdxRuntimeException("Missing: file");
 					String fileName = matcher.group(1);
 
 					imagePaths[p] = fontFile.parent().child(fileName).path().replaceAll("\\\\", "/");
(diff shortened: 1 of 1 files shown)


tdionizio:
Forgot to check for a pull request to fix this before making my own... https://github.com/libgdx/libgdx/pull/4467

Anyway, feel free to close this if already solved.

MonsterOfCookie:
Hey!

I discovered this too, however the issue can be fixed by changing the matcher to call to:

```
// Expect ID to mean "index".
Matcher matcher = Pattern.compile(".*id=(\\d+)").matcher(line);
if (matcher.find()) {
	String id = matcher.group(1);
	try {
		int pageID = Integer.parseInt(id);
		if (pageID != p) throw new GdxRuntimeException("Page IDs must be indices starting at 0: " + pageID);
	} catch (NumberFormatException ex) {
		throw new GdxRuntimeException("Invalid page id: " + id, ex);
	}
}

matcher = Pattern.compile(".*file=\"(.*)\"").matcher(line);
if (!matcher.find()) {
	Gdx.app.log("BF", "not found");
	throw new GdxRuntimeException("Missing: file");
}
String fileName = matcher.group(1);
```

The reason for the issue is the "matcher.matches()" attempts to match the entire line against the pattern. I suspect with GWT emulation it will be the native JS regex attempt which will do the partial match - hence why the error occured in GWT but not in the Desktop module - matcher.matches() always returned false, which I reckon was ok for single page font files. However in the Java world we need to use the "matcher.find()" which will look for the character request.

I have tested that this works for Desktop and GWT.

Providing we want to keep the pattern/matcher solution

tdionizio:
@MonsterOfCookie yes, it makes sense to use matcher.find(). took your comments into account and adjusted the changes, seems cleaner this way too. thanks!

IMHO using regular expressions is easier to read than using String.contains/indexOf.

MonsterOfCookie:
Did you check that this regex worked in all cases?

`matcher = Pattern.compile(".*file=\"?([^\"]+)\"?.*").matcher(line);`

tdionizio:
The only change was to remove the extra * before the + when capturing the file name.
Before it had "[^\"]*+" which is an invalid sequence, removed the * and that makes the file name mandatory but I believe that was the intended purpose. Unless the file name is optional (empty string) that should be ok.

MobiDevelop:
Looks good to me and I agree that regex vs `indexOf/contains` is more readable in this particular case.

gaborauth:
Is there any workaround before release 1.9.6?

killermenpl:
@gaborauth You can switch to 1.9.6-SNAPSHOT. It seems to work fine, but there may be some bugs I haven't run into.

PR #4477, opened 2016-12-12
Update JSON formatting to RFC 4627 Standards

madigan:
The minimal JSON formatting is concise, but it has been confusing new Scene2D UI users who are trying to debug their code and wonder if 300+ errors in their JSON structure might have anything to do with it.

### DIFF
--- a/tests/gdx-tests-android/assets/data/uiskin.json
+++ b/tests/gdx-tests-android/assets/data/uiskin.json
@@ -1,70 +1,187 @@
 {
-com.badlogic.gdx.graphics.g2d.BitmapFont: { default-font: { file: default.fnt } },
-com.badlogic.gdx.graphics.Color: {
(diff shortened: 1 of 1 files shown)


gjroelofs:
The minimal JSON can be parsed by the library JSON Parser, which can also read and write to full specced JSON if so required. However, the gains given by minimal JSON cannot simply be ignored and are best used during development. Arguably, this file could then also serve as an introduction on the subject.

madigan:
I'm not saying it can't be parsed by libGDX; I wouldn't consider it a bug, I just haven't found "the gains given by minimal JSON" that you reference, and it does seem like it's more likely to cause confusion.
That being said, I'm happy to be corrected- what is the benefit to this minimal JSON?

xoppa:
Duplicate of #4238

PR #4487, opened 2016-12-15
TextureArrayRegion and TextureArrayAtlas for easier batch management

cypherdare:
**Problem:** With large multi-page atlases, it takes a lot of care to sort texture atlas source images and in-game sprites in a way that minimizes batch flushes.

**Solution:** TextureArrays make it possible for multiple pages of an atlas to be used in a single draw call, binding only one texture. Within a couple of years, a large percentage of devices will support OpenGL ES 3.0, so this is a viable solution.

I made parent classes GLTextureRegion and GLTextureAtlas so there is minimal copied code and so there is some flexibility toward future use of these classes. The TextureArrayAtlasTest includes a quick and dirty version of SpriteBatch that can accept TextureArrayRegions, or fall back to TextureRegions if GL 3.0 is not available.

This PR has very minor code breakage. I moved TextureAtlasData into the GLTextureAtlas superclass, so imports need to be updated. The vast majority of people are probably not using that class directly.

Long term, there are various ways we could make these usable in Batches. One possibility is changing all the Batch classes to accept GLTexture and GLTextureRegion instead of Texture and TextureRegion. Then someone could encode the TextureArray layer in the U or V component (since the layer is a whole number it can be separated from the fractional part of the coordinate in the shader).

Alternatively, we could create a new type of Batch class. There have been several requests for multi-texturing, or custom vertex attributes, etc. for SpriteBatch, but PRs and Issues for those kinds of features have been closed due to the bloat and overhead they would add to SpriteBatch. I think the solution here is to create a single FlexibleBatch class that can be customized with a sort of Batchable template class that defines its behavior. I have a PR for this in the works that I'm polishing and will post in the next week or so.

This PR is an alternative to #4465.


cypherdare:
@badlogic 
Your plan sounds cleaner. I'm a bit stuck on that last bullet point. TextureRegion references a Texture, not GLTexture, so a subclass of TextureRegion cannot reference a TextureArray as is. Seems like kind of a major change for such a widely used class. Maybe TextureRegion could keep the `getTexture()` method as is with the caveat that subclasses might return null.

Alternatively we could keep the hierarchy I've created with GLTextureRegion<T> as the parent of both, and add an option to my TextureArrayRegion for baking layer into the UVs (to be unpacked in vertex shader). Then Batch could be updated to use GLTexture and GLTextureRegion instead of Texture and TextureRegion.

cypherdare:
I remember now that I created the GLTextureAtlas superclass so the return type of regions can be generic. I'm not seeing how to otherwise make the return type pluggable in a TextureAtlas subclass without breaking existing projects.

But this wouldn't matter if we make TextureArray extend Texture.

The other related hurdle is that if TextureArrayAtlas extends TextureAtlas, it must be able to return some kind of AtlasSprite subclass. Not easily doable as long as Sprite extends TextureRegion. I don't think Sprite should extend TextureRegion (conflated game object with asset, newbies trying to pass Sprites to SpriteBatch methods, and memory waste when many Sprites are based on the same UVs), but that should be its own PR.

cypherdare:
I think maybe the solution is to make TextureRegion reference a <T extends GLTexture> instead of Texture.. Most g2d classes would need to be updated to reference a GLTexture instead of Texture. If anyone needs getTexture() to return specifically a Texture, they can specify the Texture type when declaring the TextureRegion -- I think this need will be rare since most of Texture's methods are from its superclass.

Then there can be a `TextureArrayRegion extends TextureRegion<TextureArray>` that has an option for baking the layer into the UVs. TextureAtlas could have a loading option to use TextureArrays, and it would use TextureArrayRegions with baked layer under the hood. Then there would be no need for a TextureAtlas subclass.

cypherdare:
(Deleted my multiple long comments)
@badlogic 
Your plan is cleaner, but seems to require TextureArray to extend Texture instead of GLTexture. I mocked this up [here](https://github.com/libgdx/libgdx/compare/master...cypherdare:texturearray3), and it's surprisingly elegant. 

Texture and TextureAtlas are barely touched at all. Existing Batch classes stay untouched and simply work with either type of region if you give them an appropriate shader. I'm going to test this some more (haven't tried scene2d, fonts, ninepatch, etc. with it yet) and then update this PR.

cypherdare:
@badlogic 
OK, I've updated the PR to avoid changing inheritance of existing classes. Existing classes were changed just a little to allow the TextureArrayAtlas subclass as requested.

I created a TextureRegion subclass called TextureArrayRegion for optionally handling the TextureArrayAtlas's extra data in the texture coordinates. I changed AtlasRegion to inherit from TextureArrayRegion, because this avoids the need for differentiation and finding some way to plug in a different type of returned region.

TextureArrayRegion does not enforce the type of Texture it owns, so AtlasRegion has merely gained an extra optional `layer` parameter and the option to bake this layer into the coordinates.

None of the batch classes had to be touched...they simply work with the new atlas type if you use an appropriate shader!

@Tom-Ski 
To make this happen, I changed TextureArray to extend from Texture. The solution works out pretty clean, and it reduced a bit of code, too. I also removed management features from TextureArray. Correct me if I'm wrong, but only Android <3.0 needs to deal with context loss, so there's no need for managing a TextureArray.

While I was at it, I added the option to create a blank TextureArray that you can draw to (via new PixmapTextureArrayData class).

And I've added loaders and updated SkinLoader so the new TextureArrayAtlas is fully usable with AssetManager.

cypherdare:
I eliminated TextureArrayRegion and replaced it with a utility method for baking layer into texture coordinates. This reduces the impact on existing classes further. There is now no need for changing the inheritance of AtlasRegion, or for TextureArrayAtlas to use a different class for its regions.

cypherdare:
I decided to add management of TextureArrays back in because [the Android documentation doesn't seem to guarantee no context loss](https://developer.android.com/reference/android/opengl/GLSurfaceView.html#setPreserveEGLContextOnPause(boolean)).

I merged the management functionality into the Texture class object as it is much cleaner to hook into. It now supports asynchronous reloading of TextureArrays via AssetManager, just like Texture does.

Tom-Ski:
Cool stuff. A downside is that due to the use of texture coordinate to encode the index, it will break using repeats in atlases, which is still useful. The alternative would be to add another attribute and use a separate batches, but that adds quite a bunch of class overhead :[.

cypherdare:
@Tom-Ski Yes, I didn't think it would be likely we'd want to add that much bloat for the limited use cases. In most cases, I don't guess there will be so many repeating textures in use that drawing them separately will have much impact. You could at least force the repeating ones onto the first page of the atlas and swap shaders to draw them.

I made the FlexBatch library because there were so many different ways I wanted to modify SpriteBatch, but it's too optimized justify putting that kind of stuff in. If this gets pulled, I'll probably update FlexBatch for compatibility with the TextureArrayAtlas, and maybe it can avoid stripping the integers off the UVs.

Tom-Ski:
Yeah, I agree. As long as we have an easy way to get this data and use it in custom implementations of batch its not a huge deal, just something that just be noted in the documentation.

PR #4513, opened 2016-12-29
Lwjgl3Input delegates to InputEventQueue for event timestamps.

### LINKED ISSUE #4512: Lwjgl3 backend report event timestamp instead of current time.
This is a feature request and PR that would affect the Lwjgl3 backend.  It's about reporting of input event timestamps to users.  

## Background
I think the expected way to access event timestamps is to call `Gdx.input.getCurrentEventTime()` during `InputProcessor` callbacks like `keyUp()`.  For example:
```
    @Override
    public boolean keyUp(int keycode) {
        System.out.println("Event time is: " + Gdx.input.getCurrentEventTime());
        return false;
    }
```
 From the [comments/docs](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/Input.java#L709), this should return the timestamp of the current event being handled.

## Current State
I think the current implementation of `Lwjgl3Input` is actually returning the time when the most recent event was enqueued for processing, rather than the time when the currently-being-handled event occurred.

This is because the `Lwjgl3Input.currentEventTime` is only updated as events are enqueued, and then reported to the user during `getCurrentEventTime()`.

So, the timestamps for events that were enqueued earlier might be ignored and replaced with a later time stamp.

## Suggested Change
I suggest that `Lwjgl3Input` should report event timestamps by delegating to its `InputEventQueue`, instead of keeping its own `currentEventTime` field.

This would take advantage of the fact that `InputEventQueue` updates its own `currentEventTime` field each time it dequeues an event and makes an `InputProcessor` callback.

Here is a PR with the specific change that I suggest: #4513.

## Happiness
Thank you very much for libgdx!  It is a great project and I am excited to be working with it!

benjamin-heasly:
Would close #4512.

Suggest that `Lwjgl3Input` should delegate to its `InputEventQueue` for event timestamp reporting, instead of keeping its own `currentEventTime` field.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Input.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Input.java
@@ -41,20 +41,14 @@ public class Lwjgl3Input implements Input, Disposable {
 	private int pressedKeys;
 	private boolean keyJustPressed;
 	private boolean[] justPressedKeys = new boolean[256];
-	private long currentEventTime;	
 	private char lastCharacter;
-	
-	private void updateCurrentEventTime () {
-		currentEventTime = System.nanoTime();
-	}
-	
+		
 	private GLFWKeyCallback keyCallback = new GLFWKeyCallback() {		
 		@Override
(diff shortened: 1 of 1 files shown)


badlogic:
Great, thanks!

PR #4517, opened 2017-01-04
Add missing methods to ClassReflection

intrigus:
Adds some missing methods to ClassReflection.
The methods already exist but haven't been exposed for general usage.


badlogic:
Thanks!

PR #4519, opened 2017-01-05
Improved MOE support in SharedLibraryLoader

rolandvigh:


intrigus:
This should not be necessary, because the `isIos` variable should be set for robovm-ios and moe-ios.
Maybe moe-ios is detected as android, since it's based on Android ART.

liliomk:
Yes, it is recognized as Android. In MOE 1.3.0-beta-1, we changed the behavior of System.loadLibrary to be compliant with VM specs, which introduced this issue. Would you like us to change the `isIos` variable instead of creating an `isMoe`?

intrigus:
Would be better to fix that moe is recognized as Android, so changing https://github.com/libgdx/libgdx/pull/4519/files#diff-02298d26f3a99802c7af1ef20425088fL51 to 
````java
// MOE is recognized as Android, since it's based on ART, so check that it's really Android
if (vm != null && vm.contains("Android Runtime") && !isMOE) {
````
and changing the check `isIos || isMOE` back to `isIos` should work, but ultimately @badlogic has to decide.
Just my 2 cent :)

liliomk:
Please ignore this pull request, we are working on a new one with propert Multi-OS Engine 1.3.0 (currently beta) support.

PR #4520, opened 2017-01-05
gdx-dialogs extension version is now 1.2.0 and supports GWT

TomGrill:
gdx-dialogs extension version is now 1.2.0 and supports GWT


Tom-Ski:
Missed this, sorry @TomGrill

PR #4521, opened 2017-01-06
Fix #4001

### LINKED ISSUE #4001: GWT's ClassReflection.isArray returns false instead of expected true
#### Issue details

`ClassReflection.isArray` returns false instead of expected true.
It could be possible that `isEnum` also returns wrong values, but I didn't test that.
#### Reproduction steps/code

``` java

package com.badlogic.gdx.tests;

import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.tests.utils.GdxTest;
import com.badlogic.gdx.utils.reflect.ClassReflection;
import com.badlogic.gdx.utils.reflect.ReflectionException;

public class IsArrayIssueTest extends GdxTest {

    public void create () {
        testIsArray();
    }

    public void testIsArray () {
        Class c = null;
        try {
            c = ClassReflection.forName("com.badlogic.gdx.math.Vector2[]");
        } catch (ReflectionException e) {
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
        boolean isArray= c.isArray();
        Gdx.app.log("ClassReflectionTest", "native is array: " + isArray);

        boolean gdxIsArray= ClassReflection.isArray(c);
        Gdx.app.log("ClassReflectionTest", "gdx is array: " + gdxIsArray);
    }
}
```
##### Expected results are

`native is array: true`
`gdx is array: true`
##### Actual results are

`native is array: true`
`gdx is array: false`

LibGdx 1.9.2 & Gwt 2.7.0
#### Suggested fix

Just use the `isArray` method of `Class`.
#### Where's the problem

The problem is that `c` is always `null`, when it's used to generate the reflection information for a `foo[]` class, so https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheSourceCreator.java#L537 will never be entered and [`isArray`](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheSourceCreator.java#L539) will never be set to `true`.
Instead, of using `c` we should probably use [`t`](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheSourceCreator.java#L483) which is of the type `JArrayType` and `t.isArray` is `!= null`, for `foo[]`.
- [ ] Android
- [ ] iOS
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

intrigus:
Fixes #4001 and some other issues.
Try out the test to see it working.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheSourceCreator.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheSourceCreator.java
@@ -534,16 +534,25 @@ public class ReflectionCacheSourceCreator {
 		pb("if(" + varName + "!=null) return " + varName + ";");
 		pb(varName + " = new Type(\"" + name + "\", " + id + ", " + name + ".class, " + superClass + ", " + assignables + ", " + interfaces + ");");
 
+		if( c == null && t.isArray() != null){
+			// if it's not a class or an interface, it may be an array instead
+			c = t.isArray();
+		}
+		
 		if (c != null) {
 			if (c.isEnum() != null) pb(varName + ".isEnum = true;");
-			if (c.isArray() != null) pb(varName + ".isArray = true;");
-			if (c.isMemberType()) pb(varName + ".isMemberClass = true;");
-			if (c.isInterface() != null) {
-				pb(varName + ".isInterface = true;");
+			if (c.isArray() != null) {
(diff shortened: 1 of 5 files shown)


intrigus:
Thanks

PR #4525, opened 2017-01-08
Add alignment options to the SelectBox class.

keypusher:
This takes similar approach to #3964 and fixes merge conflicts with current master.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
@@ -28,6 +28,7 @@ import com.badlogic.gdx.scenes.scene2d.utils.ArraySelection;
 import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener.ChangeEvent;
 import com.badlogic.gdx.scenes.scene2d.utils.Cullable;
 import com.badlogic.gdx.scenes.scene2d.utils.Drawable;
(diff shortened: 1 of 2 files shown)


keypusher:
@NathanSweet Appreciate any feedback you have, when time allows.

keypusher:
It's been a month, any feedback on this PR?

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/List.java
```diff
@@ -171,7 +174,8 @@ public void draw (Batch batch, float parentAlpha) {
 					selectedDrawable.draw(batch, x, y + itemY - itemHeight, width, itemHeight);
 					font.setColor(fontColorSelected.r, fontColorSelected.g, fontColorSelected.b, fontColorSelected.a * parentAlpha);
 				}
-				drawItem(batch, font, i, item, x + textOffsetX, y + itemY - textOffsetY);
+				layout.setText(font, toString(item), 0, toString(item).length(), font.getColor(), width, alignment, false, "...");
```
NathanSweet: You no longer call `drawItem`, which subclasses may rely on. Better to add `width` and `alignment` parameters to `drawItem`. Also, a GlyphLayout should not be needed, BitmapFont has methods to draw with alignment (which internally uses a temp GlyphLayout).

MrStahlfelge:
This is a duplicate of #3964 which has been merged already.

PR #4530, opened 2017-01-11
Add Mesh constructor for VertexAttributes

cypherdare:
I was finding it kind of hacky to work around this missing constructor. I wanted to work with a VertexAttributes object to calculate some stuff before instantiating the Mesh, so I ended up creating one and throwing it away before passing the same individual VertexAttribute objects to Mesh for it to generate another VertexAttributes.

Also, I could see someone possibly wanting to subclass VertexAttributes.


badlogic:
Makes sense, thanks!

Issue #4532, opened 2017-01-11
Narrow stereo field in LWJGL3 backend


ollipekka:
#### Issue details
LWJGL3 backend has noticeably narrower stereo field. Quick googling shows that it is probably regression of this:
http://townsgame.com/forums/viewtopic.php?f=13&t=10302 and  https://code.google.com/archive/p/libgdx/issues/1374

#### Reproduction steps/code
You need a stereo sample that is panned to left, middle and right. You can easily check the difference by playing the sample in any player and with lwjgl3 and lwjgl2 backends. Player and lwjgl2 play full stereo field where as lwjgl3 sound is muffled.

#### Version of LibGDX and/or relevant dependencies
1.9.5 snapshot and LWJGL3 backend

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [x] MacOS

Tested on windows and MacOS

Spasi:
The narrow stereo field is caused by the default OpenAL-Soft configuration and it is not a bug. [This thread](http://openal.org/pipermail/openal/2015-February/000325.html) should explain the details.

There are two possible "fixes":

- Using the `SOFT_direct_channels` extension and `alSourcei(source, AL_DIRECT_CHANNELS_SOFT, AL_TRUE)`. This is more appropriate for music tracks.
> allows a multi-channel source to play without virtualized output speakers. By default, OpenAL requires buffer channels to be down-mixed to the output channel configuration, possibly using HRTF or other virtualization techniques to give a sense of speakers that may not be physically present. Sometimes audio tracks are authored with their own spatialization effects, where the AL's virtualization methods can cause a notable decrease in audio quality.
- Using the `EXT_STEREO_ANGLES` extension:
```java
try (MemoryStack stack = stackPush()) {
	alSourcefv(source, AL_STEREO_ANGLES, stack.floats(
		(float)(+Math.PI / 3),
		(float)(-Math.PI / 3)
	));
}
```
The OpenAL-Soft default for stereo angles is +/- 30 degrees. Setting the virtual speakers to 60 or even 90 degrees produces a wider stereo field.

code-disaster:
Is there an explanation why this doesn't happen with the LWJGL2 backend?

Spasi:
[...]

PR #4534, opened 2017-01-12
#4532 AL_DIRECT_CHANNELS_SOFT fixes narrow stereo channel problem wit…

ollipekka:
Fix to this:
- https://github.com/libgdx/libgdx/issues/4532

It is not really possible to test this without listening to the difference. The difference is that without this fix, the left and right channels won't be cleanly separated and the audio sounds muffled. With this fix, there should be cleaner separation between left and right channels.


badlogic:
Thanks!

PR #4537, opened 2017-01-12
Fix #4536: HexagonalTiledMapRenderer staggering bug

### LINKED ISSUE #4536: HexagonalTiledMapRenderer staggering bug
I have found a issue with the staggering index on the HexagonalTiledMapRenderer. When loading a map from Tiled Editor the staggering index would be odd if the selected option from tiled would be even, and vice versa.

#### Reproduction steps/code
A loaded hexagonal tiled map will not look the same as the map exported from Tiled.

andrei-cosmin:
Fixed staggering on even indexes for HexagonalTiledMapRenderer (now the map has the same staggering as it appears in Tiled Editor)


badlogic:
Thanks!

PR #4538, opened 2017-01-12
get(String name) for hex Strings

1stKarkan:
If get(String name) is now called with a hex String of 6 or 8 characters, a new Color with specified RGB(A) channel values is returned.
For Example get("ff000088") will return a red color with transparancy.

With this change it is possible to use custom colors when printing text. Also the custom Color may have transparancy

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Colors.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Colors.java
@@ -38,12 +38,43 @@ public final class Colors {
 	/** Convenience method to lookup a color by {@code name}. The invocation of this method is equivalent to the expression
 	 * {@code Colors.getColors().get(name)}
 	 * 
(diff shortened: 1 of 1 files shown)


mgsx-dev:
Why ? This method already do the job (with less code) : https://github.com/1stKarkan/libgdx/blob/bd6aba7930eed45376ad82214f1f7822e15ce25b/gdx/src/com/badlogic/gdx/graphics/Color.java#L334 ?

1stKarkan:
... I'm sorry.
I just found out that what I was planning to do is already working when adding a tag like this: [#fae20089]
I just stumbled across the ability to use [RED], [BLUE], ... to text, not custom colors. But since this is actually already working this change is 100% useless. Sorry :(

MobiDevelop:
You should use the method linked by @mgsx-dev. The method you changes is for getting a named color, not a color by hex.

PR #4539, opened 2017-01-12
Refactor GLProfiler for use in OOP to suit better with Lwjgl3. Fixes #4515

### LINKED ISSUE #4515: [BUG] [LWJGL3] GLProfiler stats not counting after GLProfiler.reset()
#### Issue details
with the lwjgl3 backend, at init running GLProfiler.enable() and then in render, do a batch draw, print counters and call GLProfiler.reset. counters will not update after that, ONLY on lwjgl3 backend. on lwjgl2 backend it works fine.

#### Reproduction steps/code
* use lwjgl3 backend
* at init call GLProfiler.enable()
* render anything (spritebatch is fine)
* print the GLProfiler.drawCalls counter, for example
* GLProfiler.reset()
* counters never go beyond 0 anymore, after that reset

a workaround for this i found is after the reset to call GLProfiler.enable() -- obviously this isn't meant to be called, but it fixes it. i've no idea how lwjgl3 manages to affect this sort of thing, as it is just a rather simple interception of gl calls, which all go through libgdx first.

```

public class GLProfilerTestCaseIssue extends ApplicationAdapter {
    SpriteBatch batch;
    Texture texture;

    public void create() {
        GLProfiler.enable();
        batch = new SpriteBatch();
        Pixmap pixmap = new Pixmap(16, 16, Pixmap.Format.RGBA8888);
        pixmap.setColor(Color.RED);
        pixmap.fill();
        texture = new Texture(pixmap);
    }

    public void render() {
        batch.begin();
        batch.draw(texture, 100, 100);
        batch.end();

        //draw calls printed as 1 the first frame, 0 every frame after!
        System.out.println("draw calls:" + GLProfiler.drawCalls);
        GLProfiler.reset();

        //if we add:
        //GLProfiler.enable();
        //here every frame, it magically counts properly. obviously that isn't the right way
    }

    public static void main(String[] args) throws Exception {
        Lwjgl3ApplicationConfiguration config = new Lwjgl3ApplicationConfiguration();
        config.setWindowedMode(640, 480);
        new Lwjgl3Application(new GLProfilerTestCaseIssue(), config);
    }
}


```
lwjgl3 backend - libgdx 1.9.4 and 1.9.5 confirmed.

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [X] Linux
- [ ] MacOS

Tom-Ski:
Add setGL20 and setGL30 methods to the Graphics interface.
Fix GLES 3.0 support methods on ios


badlogic:
@Tom-Ski can you resolve conflicts? Feel free to merge, looks good.

PR #4541, opened 2017-01-14
Added a lint file to to prevent ununnecessary error markings

Leejjon:
Added a lint file to make IntelliJ / Android Studio stop doing Android validation on the desktop Java project (to prevent ununnecessary error markings).

Original forum topic with problem described:
http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=24943

Source of solution (see the comment of Róbert Papp):
https://youtrack.jetbrains.com/issue/IDEA-122904


badlogic:
This won't automatically copy the lint file to the desktop project. Could you please amend the PR and make the gdx-setup app copy it to the desktop project on creation?

Leejjon:
Oh sure, I'll see if I can find out how to do it this week. I just assumed adding it to this folder would include it.

Tom-Ski:
Would be nice to hack into the gradle import instead to stop the gradle android facet being added to everything.

Leejjon:
@Tom-Ski I've put some time into trying to modify my gradle script to make that happen, but I was unable to (might be a bit inexperienced with gradle). On the ticket I put in the PR description on the jetbrains ticket system multiple people have reported this problem and this was the best solution posted only recently.

Leejjon:
Tested and verified that when using the gdx-setup it copies the lint.xml file to the desktop project.

PR #4555, opened 2017-01-21
PolygonSprite minor improvements

### LINKED ISSUE #3722: PolygonSprite:  setRegion uses mutable Color instance without resetting to vertex values
In PolygonSprite, `getVertexColor` should be called inside `setRegion` before using the mutable `Color` instance, otherwise the color can be modified by side-effects.

Example:

``` java
import com.badlogic.gdx.*;
import com.badlogic.gdx.Input.Keys;
import com.badlogic.gdx.graphics.*;
import com.badlogic.gdx.graphics.g2d.*;

public class BarebonesBatch extends ApplicationAdapter {
    PolygonSpriteBatch batch;
    Texture texture;
    PolygonRegion region;
    PolygonSprite sprite;

    public void create () {
        batch = new PolygonSpriteBatch();
        texture = new Texture(Gdx.files.internal("data/badlogic.jpg"));
        region = new PolygonRegion(new TextureRegion(texture), new float[] { 0, 0, 0, 100, 100, 100 }, new short[] { 0, 1, 2 });
        sprite = new PolygonSprite(region);
        sprite.setColor(0, 1, 0, 1);
    }

    public void render () {
        // Example:  calling getColor and modifying it causes a side-effect
        if (Gdx.input.isKeyJustPressed(Keys.SPACE)) {
          Color color = sprite.getColor();
          color.set(1, 0, 0, 1);  // only setColor should actually change the color
          //sprite.getVertexColor();  // fix: getVertexColor should be called inside setRegion
          sprite.setRegion(region);
        }
        batch.begin();
        sprite.draw(batch);
        batch.end();
    }

        new LwjglApplication(new BarebonesBatch());
    }
}
```

cypherdare:
I removed a couple of redundant lines (I'll explain below) and added an optimization when setting a new region.

Resolves #3722 by correcting the docs for the `getColor` methods. I didn't simply say "don't modify the returned color" because it would be valid to pass the color back into `setColor`, which might be convenient if doing some sort of transformation on the existing color.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSprite.java
@@ -37,7 +37,6 @@ public class PolygonSprite {
 
 	public PolygonSprite (PolygonRegion region) {
 		setRegion(region);
-		setColor(1, 1, 1, 1);
 		setSize(region.region.regionWidth, region.region.regionHeight);
 		setOrigin(width / 2, height / 2);
 	}
@@ -62,7 +61,6 @@ public class PolygonSprite {
 		scaleX = sprite.scaleX;
 		scaleY = sprite.scaleY;
 		color.set(sprite.color);
-		dirty = sprite.dirty;
 	}
 
 	/** Sets the position and size of the sprite when drawn, before scaling and rotation are applied. If origin, rotation, or scale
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSprite.java
```diff
@@ -37,7 +37,6 @@
 
 	public PolygonSprite (PolygonRegion region) {
 		setRegion(region);
-		setColor(1, 1, 1, 1);
```
cypherdare: Redundant. `setRegion()` always applies the `color`.

gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSprite.java
```diff
@@ -62,7 +61,6 @@ public void set (PolygonSprite sprite) {
 		scaleX = sprite.scaleX;
 		scaleY = sprite.scaleY;
 		color.set(sprite.color);
-		dirty = sprite.dirty;
```
cypherdare: Removed for being redundant. Newly set region always leaves `dirty` true.

PR #4564, opened 2017-01-25
static overlaps for Rectangle

ghandhikus:


Tom-Ski:
Use Intersector.overlaps ()

PR #4565, opened 2017-01-25
raw overlaps to avoid unnecessary object creation

ghandhikus:


cypherdare:
I think this needs a javadoc to explain the parameters and that it's for rectangles.

ghandhikus:
Parameter count and names suggest that

MobiDevelop:
I'd agree that the method needs a javadoc and/or be named something with the word rectangles in it (e.g. `overlapRectangles(...)`).

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -887,6 +887,21 @@ public static boolean overlaps (Rectangle r1, Rectangle r2) {
 		return r1.overlaps(r2);
 	}
 
+	/**
+	 * @param r1x x of first {@link Rectangle}
+	 * @param r1y y of first {@link Rectangle}
+	 * @param r1w width of first {@link Rectangle}
+	 * @param r1h height of first {@link Rectangle}
+	 * @param r2x x of first {@link Rectangle}
```
Darkyenus: "second"

Nothing in the signature (except parameter names) suggest that this method deals with rectangles, I am in favor of renaming the method to something like `overlapRectangles`.

badlogic:
Agree with the others, the method name needs to indicate that this a rectangle overlap test. Please amend the PR or send a new one.

PR #4569, opened 2017-01-25
Add getters and setters to Rectangle class for its sides (top, right, bottom, left)

varomorf:
Add methods to Rectangle class for getting and setting its sides and center (without the need to pass a Vector param).

I've added this as I found useful said methods on MonoGame's Rectangle2D class and decided to port said functionality to libgdx while I'm working a port from MonoGame to libgdx after doing a Coursera course.

I still need to sign the CLA but as I'm not sure if this fits in the project, I'm asking beforehand to avoid lost time :)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -351,12 +351,6 @@ public class Rectangle implements Serializable, Shape2D {
 		this.x = x;
 	}
 
-	/** Calculates the center of the rectangle.
-	 * @return A vector with the center position. */
-	public Vector2 getCenter () {
-		return getCenter(new Vector2());
-	}
-
 	/** Calculates the center of the rectangle. Results are located in the given Vector2
 	 * @param vector the Vector2 to use
 	 * @return the given vector with results stored inside */
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -303,6 +303,60 @@ public class Rectangle implements Serializable, Shape2D {
 		return (height == 0) ? Float.NaN : width / height;
 	}
 
+	/** Returns the bottommost coordinate of the rectangle.
+	 * @return The bottommost coordinate of the rectangle. */
+	public float getBottom () {
+		return y;
+	}
+
+	/** Sets the Y coordinate so that the bottom of the rectangle is at the given value.
+	 * @param y The value for the bottom. */
+	public void setBottom (float y) {
+		this.y = y;
+	}
+
+	/** Returns the topmost coordinate of the rectangle.
+	 * @return The topmost coordinate of the rectangle. */
+	public float getTop () {
+		return y + height;
+	}
+
+	/** Sets the Y coordinate so that the top of the rectangle is at the given value.
+	 * @param y The value for the top. */
+	public void setTop (float y) {
+		this.y = y - height;
+	}
+
+	/** Returns the rightmost coordinate of the rectangle.
+	 * @return The rightmost coordinate of the rectangle. */
+	public float getRight () {
+		return x + width;
+	}
+
+	/** Sets the X coordinate so that the right of the rectangle is at the given value.
+	 * @param x The value for the right. */
+	public void setRight (float x) {
+		this.x = x - width;
+	}
+
+	/** Returns the leftmost coordinate of the rectangle.
+	 * @return The leftmost coordinate of the rectangle. */
+	public float getLeft () {
+		return x;
+	}
+
+	/** Sets the X coordinate so that the left of the rectangle is at the given value.
+	 * @param x The value for the left. */
+	public void setLeft (float x) {
+		this.x = x;
+	}
+
+	/** Calculates the center of the rectangle.
+	 * @return A vector with the center position. */
+	public Vector2 getCenter () {
+		return getCenter(new Vector2());
+	}
+
 	/** Calculates the center of the rectangle. Results are located in the given Vector2
 	 * @param vector the Vector2 to use
 	 * @return the given vector with results stored inside */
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/math/Rectangle.java
```diff
@@ -303,6 +303,60 @@ public float getAspectRatio () {
...
+
+	/** Sets the X coordinate so that the left of the rectangle is at the given value.
+	 * @param x The value for the left. */
+	public void setLeft (float x) {
+		this.x = x;
+	}
+
+	/** Calculates the center of the rectangle.
+	 * @return A vector with the center position. */
+	public Vector2 getCenter () {
```
Tom-Ski: These methods should be avoided, getCenter (Vector2 out) already exists and is the preferred method due to it not generating garbage internally. If the user doesn't care about the garbage, or simply wants to create a new vector2 isntance they should pass in new Vector2() to the method.

varomorf: I'm simply not new but plainly unknowledgeable about game programming, so please correct me if I'm saying something ridiculous.

But, in Java programming in general that code doesn't generate garbage (or I'm understanding garbage wrongly).

When the user calls that method, a new Vector object is created, populated (in the already existing method), returned, then the user does something with it (presumably getting its x and y) and unless it saves it in a permanent field or something it will be removed from memory the next time the garbage collector executes. So that's why I don't understand the part of the method generating garbage.

Again, I'm talking from a "normal Java coding" view and not a game programming view so maybe I'm really overstepping here and would like to know if so.

varomorf:
Hi @Tom-Ski 

The idea was to add a more semantic approach so that coders don't need to compute x + width every time they need to get the right part of a Rectangle (for example to know if the Rectangle is touching some bounds or something like that).

I'm currently using the Rectangle class as a Collision rectangle to know if an entity is touching some bounds (top, right, bottom and left) so having said "shortcuts" simplifies" my code quite a lot.

Again as with my other comment, I'm talking from a "normal Java coding" view and not a game programming view so maybe I'm really overstepping here and would like to know if so.

gdx/src/com/badlogic/gdx/math/Rectangle.java
agmcleod: @varomorf this one is generating garbage because you're creating a new object each time you call it. Ideally in your game code, you should have a single Vector2 property that you re-use to pass to getCenter().

MobiDevelop: Giving the garbage collector more work to do means it takes more time to do its work. Minimizing that time is important in games and in particular on mobile, where the garbage collection is generally not amazingly fast.

Tom-Ski:
Sure, do you have a code sample/pseudo code for your bounds check for collision you mentioned? It would be good to see how you are using it and how this benefits you. Particularly for the setLeft methods.

If something has definite utility then it can be added at the expense of bloat etc, I'm just not sure this does.

varomorf:
For example, I have the following inside a Target class. I use it to have several targets floating through the screen bouncing off the limits of the screen. This way I understand it better (is more semantic) than with the computations directly in the code.

```
/**
 * Bounces the target off the top and bottom window borders if necessary
 */
private void bounceTopBottom()
{
    if (drawRectangle.getTop() > GameConstants.WINDOW_HEIGHT)
    {
        // bounce off top
        drawRectangle.setTop(GameConstants.WINDOW_HEIGHT);
        velocity.y *= -1;
    }
    else if (drawRectangle.getBottom() <= 0)
    {
        // bounce off bottom
        drawRectangle.setBottom(0);
        velocity.y *= -1;
    }
}

/**
 * Bounces the target off the left and right window borders if necessary
 */
private void bounceLeftRight()
{
    if (drawRectangle.getLeft() < 0)
    {
        // bounce off left
        drawRectangle.setLeft(0);
        velocity.x *= -1;
    }
    else if (drawRectangle.getRight() > GameConstants.WINDOW_WIDTH)
    {
        // bounce off right
        drawRectangle.setRight(GameConstants.WINDOW_WIDTH);
        velocity.x *= -1;
    }
}
```

gdx/src/com/badlogic/gdx/math/Rectangle.java
varomorf: I think that I understand what you're saying. But still can't quite wrap my head around it. Let's say I have the following code in an EnemyShip class:

```
private void shoot(){
    Texture projectileSprite = Game1.GetProjectileSprite(ENEMY_SHIP_PROJECTILE_TYPE);
    Vector2 center = drawRectangle.getCenter();

    int projectileY = (int) (center.y + GameConstants.ENEMY_SHIP_PROJECTILE_OFFSET);
    Projectile projectile = new Projectile(ENEMY_SHIP_PROJECTILE_TYPE, projectileSprite, center.X, projectileY, -GetProjectileYVelocity());
    Game.AddProjectile(projectile);
}
```

This method will be called in every enemy ship that is currently in the game when a counter gets to a limit value (limit value that gets updated in an update method). 

This way the projectiles get shoot off the center of the enemy ships (and slightly down of its center).

I fail to see what improvement I could get creating a new Vector2 here and passing it to rectangle.getCenter(Vector2), instead of creating it inside rectangle.getCenter().

Again, I'm REALLY new to game programming so I may be doing this completely wrong. 

Maybe I should move this to another class and loop every ship using only one Vector2 for their centers?

Tom-Ski: The improvement would be not to create instances, so you wouldn't pass in a new Vector2, but pass in a class member variable instead, or a temporary variable that persists.

varomorf: Oooooooooooo. I think that now I'm starting to understand :)

Thank you all for the explanation, I'll remove said code as now I understand the problems it might produce.

cypherdare:
I'm of the opinion that this is bloat, and easily added by subclassing anyway. It doesn't make sense to have multiple methods for returning the exact same thing under a different name.

Tom-Ski:
I agree, the use case you have provided is a very specific example, usually with collision detection and response you have a more robust method of determining collisions between objects. Infinite size sides is a very small case. If you'd like to avoid having the additional math everywhere in your code just add a simple helper method in your own class.

Thanks for the PR.

Issue #4560, opened 2017-01-24
Default gitignore is too restrictive


mgsx-dev:
gitignore file generated by gdx setup have undesirable side effects : any folder you want to track with git (including java package) which is named "war", "build", "out" or "bin" will be ignored. Note that I discovered this issue creating a "war" game with a "war" package name somewhere, so other developers  may face the same problem.

IMO best practice is to avoid recursive path in gitignore as much as possible (all path without a leading slash are processed recursively).

My proposal would be to keep some global recursive ignore in the root file (*.class, *.project ...) and dispatch specific parts to related sub-project : For instance, "war/" would be replaced by "/war" and moved to web sub-project.

What do you think ?

Tom-Ski:
Sure, PR welcome, let's keep it at one master .gitignore though

----- Reply message -----

PR #4580, opened 2017-01-30
improve gitignore template fix libgdx/libgdx#4560

mgsx-dev:
as discussed in #4560, here is the fix (maybe not the final fix).

I'm not really confident with these modifications because I'm not using all OS/IDE/Targets. I'm using eclipse on linux and mostly target desktop and android so I think this fix should be validated by more contributors.

Has you can see, what I've done so far is : 
* add leading slashes and full path to avoid unwanted ignore.
* removed some duplicates : for instance html/war/WEB-INF/deploy/ is useless with /html/war
* removed some IMO useless : tmp/
* move local.properties to gradle section

Note that I could use single wildcard instead of path for every sub project (for instance /*/dist/) but it could conflicts with some project specific folder structure.

@Tom-Ski : I don't really understand why you want to keep just one big file instead of dispatching some parts in related sub projects. It would be cleaner way for many reasons : 
* make gitignore files more readable.
* avoid useless ignore when not targeting all platforms.
* easily manage sub folder renaming / moving.

Thanks to point me in the right direction in order to improve these changes.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gitignore
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gitignore
@@ -6,64 +6,108 @@
 hs_err_pid*
 
 ## Robovm
-robovm-build/
+/ios/robovm-build/
 
 ## GWT
-war/
-html/war/gwt_bree/
-html/gwt-unitCache/
+/html/war/
+/html/gwt-unitCache/
 .apt_generated/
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Mostly for the benefit of people not used to git. Just like with gradle we have all the dependencies declared in the root build.gradle file as that makes it easier for people new to gradle to see project definitions, compare them and add/remove what they need without having to forage around in an environment they aren't used to.

PR #4583, opened 2017-01-31
LwjglAWTCanvas: Fix for flickering cursor during swing/javafx drag & drop

Lusito:
Every render frame, canvas.setCursor(cursor); would be called:
https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglAWTCanvas.java#L254

When you have an application using swing or javafx drag & drop features, the above will cause the mouse-cursor to flicker during dnd.

LwjglCanvas suffers from the same issue:
https://github.com/libgdx/libgdx/blob/9b8f20b2d0e96e53f0f2b98dc8f6131c810aae71/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglCanvas.java#L216

I'm honestly not sure why this was done every frame..
@NathanSweet could you please give some insight?

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglAWTCanvas.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglAWTCanvas.java
@@ -76,7 +76,6 @@ public class LwjglAWTCanvas implements Application {
 	int logLevel = LOG_INFO;
 	ApplicationLogger applicationLogger;
 	final String logTag = "LwjglAWTCanvas";
-	Cursor cursor;
 
 	public LwjglAWTCanvas (ApplicationListener listener) {
 		this(listener, null, null);
@@ -251,7 +250,6 @@ public class LwjglAWTCanvas implements Application {
 		if (!running) return;
 
 		setGlobals();
-		canvas.setCursor(cursor);
 
 		int width = Math.max(1, graphics.getWidth());
 		int height = Math.max(1, graphics.getHeight());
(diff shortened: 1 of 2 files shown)


Lusito:
It's been almost two month.. is there a reason for this not to be merged?

yuripourre:
@Lusito, updating cursor every frame doesn't make sense, indeed. I didn't test the PR but the changes seems good to me.

yuripourre:
@Lusito now I got why it is this way. Setting cursor is necessary to change the cursor on the fly. If you only call it during configuration, your cursor will be defined but cannot change after set.

If this behavior produces flickering I would suggest to check if cursor is the same (or a variable) before setting it.

```
if (cursorChanged) {
  canvas.setCursor(cursor);
}
```

@SimonIT I didn't notice before but the current PR would produce a breaking change.

NathanSweet:
I don't remember why it was done the way it is currently. However, I tried only setting the cursor when it changes and the cursor does not change, or only changes for 1 frame. Something is probably overwriting the cursor, I'd guess this:
https://github.com/LWJGL/lwjgl/blob/2df01dd762e20ca0871edb75daf670ccacc89b60/src/java/org/lwjgl/opengl/WindowsDisplay.java#L633

yuripourre:
I see, the code there is not beautiful, thank you for pointing.
Since lwjgl2 will not be supported anymore, should we close this PR?

crykn:
Since the proposed fix is apparently not working, I think yes, we should.

Lusito:
Back then I tried the patch and it worked for me when using a snapshot version.
But I'm not using this anymore, so it's fine. Just surprised someone actually took a look after all this time.

NathanSweet:
I'd be up to make changes to LWJGL2 since I use it in some of my projects, except when I've tried their build system was... not terribly straightforward. It'd be easier to monkey patch the class files.

Issue #4590, opened 2017-02-03
GWT build will not load without Flash On Some Browsers


ajayyy:
Please ensure you have given all the following requested information in your report.

#### Issue details
When trying to run on a fresh Ubuntu install with Firefox, the game does not load, and an error appears in the console. (in stacktrace section)

I even have preferFlash to false:

```
public GwtApplicationConfiguration getConfig () {
        	GwtApplicationConfiguration config = new GwtApplicationConfiguration(getWindowInnerWidth(), getWindowInnerHeight());
        	config.preferFlash = false;
            return config;
        }
```

This appears: 

![](https://i.stack.imgur.com/yao56.png)

#### Reproduction steps/code
Running a game on Firefox on fresh Ubuntu. Here is a link to my test game in case you need it: http://ajayinkingston.com/splats/

#### Version of LibGDX and/or relevant dependencies
1.9.5 and 1.9.4 seem to have this issue.

#### Stacktrace

```
SoundManager: No Flash detected. Trying HTML5-only mode. soundmanager2-jsmin.js:32:357
SoundManager V2.97a.20130512 (AS3/Flash 9) + HTML5 audio soundmanager2-jsmin.js:32:357
SoundManager 2: soundManager: Fatal error: Flash is needed to play some required formats, but is not available. soundmanager2-jsmin.js:32:357
SoundManager 2 HTML5 support: mp3 = false (using flash), mp4 = false (using flash), ogg = true, opus = true, wav = true soundmanager2-jsmin.js:32:357
soundManager: No Flash response within expected time. Likely causes: Flash blocked or JS-Flash security error. soundmanager2-jsmin.js:32:357
soundManager: Waiting indefinitely for Flash (will recover if unblocked)...
```

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Mur4ik:
Have You tried **config.preferFlash = false** in **GwtApplicationConfiguration** ?

ajayyy:
@Mur4ik Yes, sorry I forgot to include it in the post. I have set preferFlash to false.

Mur4ik:
In which format your sound assets is (mp3, ogg)?

```
SoundManager 2: soundManager: Fatal error: Flash is needed to play some required formats, but is not available. soundmanager2-jsmin.js:32:357
[...]

PR #4596, opened 2017-02-07
GwtApplication refactoring + small GWT SoundManager improvements for better flexibility

Mur4ik:
GwtApplication refactoring + small GWT SoundManager improvements for better flexibility


badlogic:
There's no need to add this to GwtApplication, you can simply swap out the Gdx.audio static if you need a mocked implementation.

Mur4ik:
Ok, but how about rest of PR and this issue https://github.com/libgdx/libgdx/issues/4590?

PR #4599, opened 2017-02-10
MeshBuilder error using Short.MAX_VALUE vertices

Bolt-Head:
MeshBuilder will give an error when trying to use Short.MAX_VALUE vertices in a mesh.  This means meshes created using MeshBuilder can only use (Short.MAX_VALUE - 1) total vertices.

A use case requiring the change would be a mesh pooling implementation using MeshBuilder currently needs to check against the size of (Short.MAX_VALUE - 1) instead of the actual number of vertices allocated in the mesh.


xoppa:
Thanks!

Issue #4154, opened 2016-06-17
RotateToAction behavior missing wrap around logic


wiverson:
#### Issue details

RotateToAction does not properly deal with rotation cases around the 360 degree barrier - it always goes "the long way around".

```
com.badlogic.gdx.scenes.scene2d.actions.RotateToAction.update (float percent) {
        target.setRotation(start + (end - start) * percent);
    }
```

should probably be implemented as something like...

`target.setRotation(MathUtils.lerpAngleDeg(start, end, percent));
`
#### Version of LibGDX and/or relevant dependencies

  `gdxVersion = '1.9.3'
`
#### Affected platforms  - All

My first reported issue, so be kind.  :)  Seems awfully tiny for a pull request, but if that or anything else would be helpful just let me know.

xoppa:
So, you are suggesting that if an actor is currently unrotated and you apply a RotateToAction to 360 or 720 degrees to it, then it should do nothing (instead of fully rotating once or twice)? That doesn't sound right. Why do you think that?

wiverson:
Right now, if you have an actor currently rotated to 359, and you give it a RotateTo of 1, it goes all the way around (moving a full 358 degrees) instead of moving two degrees back to the 1.  That doesn't seem right either.

It might be that there is room for two Actions here instead, maybe a RotateTo and a RotateToDirect or something, or a parameter for RotateTo...?

wiverson:
Another idea - RotateTo (which takes a destination angle and rotates directly to the angle) and SpinTo (which explicitly spins to the target angle, multiple times if given values above 360 or below 0, clockwise if positive and counter-clockwise if negative).

Happy to do a pull if that is helpful.

badlogic:
This is about taking the shortest path for a rotation. Should be a parameter to RotateTo. Happy to merge PRs for that.

PR #4603, opened 2017-02-14
Adding LERP option to RotateToAction #4154

wiverson:
First PR, lmk if there is anything I should change.  :)

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/RotateToAction.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/RotateToAction.java
@@ -16,17 +16,41 @@
 
 package com.badlogic.gdx.scenes.scene2d.actions;
 
+import com.badlogic.gdx.math.MathUtils;
+
 /** Sets the actor's rotation from its current value to a specific value.
+ * 
+ * By default, the rotation will take you from the starting value to the specified value directly. For example, setting the start
+ * at 350 and the target at 10 will result in 340 degrees of movement.
+ * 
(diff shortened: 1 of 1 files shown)


mgsx-dev:
Interpolation (not only linear) is already supported. Use [setInterpolation](https://github.com/wiverson/libgdx/blob/c7bb9eeaa7acfa7bf4528275e3da6a73107d8320/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/TemporalAction.java#L122) method on any TemporalAction like RotateTo and RotateBy.

MobiDevelop:
Lerp isn't the right name for what this does, this would be something like useShorterDirection.

cypherdare:
How about `shorterPath`?

gdx/src/com/badlogic/gdx/scenes/scene2d/actions/RotateToAction.java
```diff
@@ -16,17 +16,41 @@
...
+		this.setLerp(useLerp);
+	}
+
 	protected void begin () {
 		start = target.getRotation();
 	}
 
 	protected void update (float percent) {
-		target.setRotation(start + (end - start) * percent);
+		if (isLerp())
```
cypherdare: Don't need a method call here. Though I suppose JIT will optimize it anyway.

wiverson:
useShortAngle?

wiverson:
FWIW, useLerp refers to the implementation, all of the others are for the user (which is better).  I don't have a strong option on the specific term, I'll submit a better user-centric name tomorrow with whatever seems like a good pick.  :)

badlogic:
Thanks!

PR #4604, opened 2017-02-14
Prevent potential crashes in IOSApplication due to audio being unavailable

CoderBaron:
I am seeing several `EXC_BAD_ACCESS KERN_INVALID_ADDRESS` crash logs happening when `OALSimpleAudio.sharedInstance().setUseHardwareIfAvailable(false);` is called. 

The issue seems to be that the shared instance of `OALSimpleAudio` is sometimes not available and returns `null`. There is already a null-check in https://github.com/libgdx/libgdx/blob/14d89225ae26afc1cc7f52e986e566bf19425275/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSAudio.java#L34 for the very same reason.

This PR adds the same checks to `didBecomeActive` and `willEnterForeground` for the `OALAudioSession` and `OALSimpleAudio` instances. 
I also made a very small change to robovm `IOSNet` to accompany the latest changes to moe `IOSNet`.


Tom-Ski:
Cheers @CoderBaron

PR #4609, opened 2017-02-17
Allow shader uniform customization

cypherdare:
Because why not?


badlogic:
I'm not sure what the purpose of this is. Users using a custom shader are supposed to set their uniforms outside the Batch implementation.

PR #4613, opened 2017-02-19
Avoid type-unaware Animation (fixes #4476)

### LINKED ISSUE #4476: ClassCastException in Animation.getKeyFrames()
#### Issue details
Affected calss com.badlogic.gdx.graphics.g2d.Animation

code:
`Animation<TextureRegion> animation = new Animation<TextureRegion>(...);`
`TextureRegion[] f = animation.getKeyFrames();`
cause:
Exception in thread "LWJGL Application" java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Lcom.badlogic.gdx.graphics.g2d.TextureRegion;

this code:
`Object[] f = animation.getKeyFrames();`
does not throw exceptions

LibGDX v1.9.5

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

cypherdare:
Fixes #4476.

An animation instantiated with an Array was using a backing Object array instead of the correct array type, causing possible CCE when calling `getKeyFrames()` and assigning it to its expected type.

Now it uses the same backing type as the Array that's passed in, so as long as the Array was instantiated with a type parameter, you're good to go. I updated TextureAtlas to return type-aware arrays in `findRegions()` since this is a common source of animation frames.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -18,6 +18,7 @@ package com.badlogic.gdx.graphics.g2d;
 
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
 /** <p>
  * An Animation stores a list of objects representing an animated sequence, e.g. for running or jumping. Each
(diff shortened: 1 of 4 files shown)


badlogic:
Have you tested this with GWT?

cypherdare:
No, I haven't, although I copied the array generation reflection code out of the Array class, which doesn't have an emulated version.

Not likely that I'll be able to test GWT soon. Just tinkered with it for an hour and couldn't get gwt-tests to run. There are various project setup issues, and I'm not sure how the backend tests project might be set up differently from a typical Libgdx GWT project, and I'm not aware of any documentation on running it. I don't use GWT, and don't know when I'll have time and motivation to get up to speed on it enough to know what I'm doing with it. Maybe someone else is willing to give this a spin?

Tom-Ski:
`./gradlew tests:gdx-tests-gwt:superDev` then go to http://localhost:8080/gdx-tests-gwt/

PR #4621, opened 2017-02-25
Fixed crash in Particle3D system.

Caresilabs:
I have tested the Particle3D system on three smartphones so far and two of them crashed on the:
`varying mat2 v_rotation;` part. One of them is the Nexus 7 2012 and the other is a Sony (not sure model, not mine).

I can only speculate because I have searched the web and did not find anything. So my hypothesis is that some GPUs does not support matrices sent as varying. I've tried the code in the pull request and it compiles fine on all three devices.


Tom-Ski:
Can reproduce with the nexus 7 2012

Tom-Ski:
Can't find any info on this specific problem that might help debugging, perhaps its just a really dodgy driver.  Either way we can have this workaround in the meantime, thanks.

PR #4622, opened 2017-02-26
Add fitScreen method for Actor

RUSshy:


Darkyenus:
This isn't a very good fit for the API. Primarily, many (most?) people don't use viewport of dimensions `Gdx.graphics.getWidth/Height()` and for those, this method will be misleading and incorrect.

NathanSweet:
Agreed.

PR #4627, opened 2017-03-01
InputProcessor Default methods

SimplyBallistic:
It annoyed me that there was no interface with default methods for the input processor. I didn't want to give up extending my classes to an InputAdapter or have a mess of methods I will not use. So I added default methods to the InputProcessor and depreciated the meaningless InputAdapter. 

You can rework this to your liking, but just get the general idea, use default methods for the InputProcessor and future interfaces of the same category.

### DIFF
--- a/gdx/src/com/badlogic/gdx/InputAdapter.java
+++ b/gdx/src/com/badlogic/gdx/InputAdapter.java
@@ -15,9 +15,10 @@
  ******************************************************************************/
 
 package com.badlogic.gdx;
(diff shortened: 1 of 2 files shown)


mgsx-dev:
default methods is a java 8 feature so it won't compile against java 7 or java 6 (libGDX target is java 6)

MobiDevelop:
As @mgsx-dev noted, we need to retain compatibility with Java 6/7.

PR #4630, opened 2017-03-02
Improve Performance, don't calculate known values

intrigus:
This improves performance a little bit by not calculating known values.
In my case it saved about 200ms.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -663,8 +663,18 @@ public class FreeType {
 					for (int y = 0; y < rows; y++) {
 						src.get(srcRow);
 						for (int x = 0; x < width; x++) {
-							float alpha = (srcRow[x] & 0xff) / 255f;
-							alpha = (float)Math.pow(alpha, gamma); // Inverse gamma.
+							float alpha;
+
+							// Zero raised to any power is always zero.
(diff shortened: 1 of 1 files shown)


extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
```diff
@@ -664,7 +664,13 @@ public Pixmap getPixmap (Format format, Color color, float gamma) {
 						src.get(srcRow);
 						for (int x = 0; x < width; x++) {
 							float alpha = (srcRow[x] & 0xff) / 255f;
-							alpha = (float)Math.pow(alpha, gamma); // Inverse gamma.
+							
+							// Zero raised to any power is always zero.
+							// One raised to any power is always one
+							// This means that we only have to calculate Math.pow() when alpha is NOT zero and NOT one
+							if (alpha != 0.0f && alpha != 1.0f) {
```
Darkyenus: The check should be probably done directly on `srcRow[x] & 0xff` (before the division), as "hard" float comparisons are discouraged and may fail on platforms with less precise floating point calculation (it shouldn't happen, but cheap phone vendors never fail to surprise). (Added bonus is that division is eliminated, making the code even slightly faster.)

Just a nitpick, otherwise 👍

badlogic: Jupp, integer check would be preferred.

intrigus:
@badlogic @Darkyenus Hey, I've updated my patch, please review again :)

PR #4631, opened 2017-03-02
fix particle emitter sprite update

### LINKED ISSUE #4496: Unable to change particles sprite after spawn
#### Issue details
Calling [setSprite](https://github.com/libgdx/libgdx/blob/1.9.5/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java#L601) on particle emitter does not change sprite of existing particles - only texture and origin are modified. Particle's region should also be updated. I think size should also be changed.

#### Reproduction steps/code
see https://github.com/serpent7776/gdx-particle-emitter-change-sprite-issue-sample

1.9.5

- [x] core
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

serpent7776:
When updating sprite with different region of the same texture, region of existing particles is not updated. This change will call `setRegion` on all active particles to update `TextureRegion`.

This should fix #4496


Tom-Ski:
Thanks!

PR #4632, opened 2017-03-02
avoid calls to bind() when creating texture

GoGonania:


xoppa:
Better just the already existing unsafe methods (see #558)

code-disaster:
It's a 2-lines change with the existing unsafe methods.

PR #4634, opened 2017-03-05
Container has got missing removeActor overload method

### LINKED ISSUE #4623: Container#removeActor(Actor, boolean) overridden method is missing
#### Issue details
`Container` overloads whole set of add/remove actor methods to define its specific singular child logic. But it has no override for `Group#removeActor(Actor, boolean)` which causes next bug: if child actor calls `remove()` on self, it triggers parent's `Group#removeActor(Actor, boolean)` and not `Group#removeActor(Actor)`, so `Container` will simply call super method (of `Group`) and keep reference to child actor.

Looks like regression bug, because javadoc of `Actor#remove()` also refers to `Group#removeActor(Actor)`, but actually calls only `Group#removeActor(Actor, boolean)`

#### Reproduction steps/code
``` java
Actor actor = new Actor();
Container container = new Container();
container.setActor(actor);
actor.remove();
assert(container.getActor() == null);
```

1.9.5

- [ ] Core

metaphore:
Container now has overridden Group#removeActor(Actor, boolean) method. Any child of Container can call Actor#remove() and it will work properly.

This patch fixes #4623

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Container.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Container.java
@@ -154,7 +154,7 @@ public class Container<T extends Actor> extends WidgetGroup {
 	public void setActor (T actor) {
 		if (actor == this) throw new IllegalArgumentException("actor cannot be the Container.");
 		if (actor == this.actor) return;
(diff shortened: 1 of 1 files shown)


NathanSweet:
Closed via e5d44539d146abbc2fd6203e494f27b36b1843d5, thanks!

Issue #4506, opened 2016-12-26
Create common interface for container classes in com.badlogic.gdx.utils


danijelz:
Looks like all Collection/Map alike utility classes have at least two things in common - size and ability to clear. Utility methods that question object 'emptyness' would be much easier to implement if all those classes would share a common interface e.g.:

    public interface Container {
        int size();
    
        void clear();
    }

Some of classes already implement interface Iterable and the 'emptyness' could be checked by Iterator, but this triggers creating Iterator instance if there was no need for one before. 

This interface could be implemented by:
 - Array
 - ArrayMap
 - BooleanArray
 - ObjectSet
 - ObjectMap
 - ObjectIntMap
 - ObjectFloatMap
 - LongMap
 - LongArray
 - IntSet
 - IntMap
 - IntIntMap
 - IntFloatMap
 - IntArray
 - IdentityMap
 - FloatArray
 - CharArray
 - ByteArray
 - BooleanArray
 - Queue
 - ShortArray
 - SortedIntList
 - PooledLinkedList

This could be a humble beginning for further API generalisation on this classes if needed.

PR #4638, opened 2017-03-09
Feature/container interface

Hogusong:
created a common interface container class with size and clear functions
implemented the container class in appropriate classes
for issue #4506

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -26,7 +26,7 @@ import com.badlogic.gdx.utils.reflect.ArrayReflection;
 /** A resizable, ordered or unordered array of objects. If unordered, this class avoids a memory copy when removing elements (the
  * last element is moved to the removed element's position).
  * @author Nathan Sweet */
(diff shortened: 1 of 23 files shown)


badlogic:
Could you give me a use case for this?

Hogusong:
I created this for issue #4506
@danijelz is there a specific use case you had in mind for this issue.

NathanSweet:
I haven't needed this, but I'm not against it. The Eclipse source code formatter should be run on classes changed. I'm not sure about the name `Container`. At least, libgdx already has a class called `Container` in scene2d. `Collection`? java.util.Collection is very rarely used in app code.

cypherdare:
Not sure what the point of creating this interface is if the only method it has is `size()`. With that single method, there's nothing you can do to the collection after you've checked its size. About the only other method you could add without generics is `clear()`. I can't think of any use case with just those two methods, but generics would make the primitive-based collections messy. 

If you add enough methods for this to have some use, it becomes clumsy to support both maps and lists/sets with the same interface.

Tom-Ski:
Don't see clear use case, or support for this, so closing it out. If the situation changes we can reopen.

PR #4639, opened 2017-03-09
Add map() function to MathUtils

Lonami:
Hey! First, I love `libGDX`. Amazing work :)

There's a function which I've been missing and I find *really* useful, `map()`. It takes an input value and two ranges, start range and end range. It then maps the value from the start to the end range. Pretty useful. I ported it from [Processing](https://github.com/processing/processing/blob/db4f1449e4097e08683a276c64db392f906362b4/core/src/processing/core/PApplet.java#L4369-L4373) because I would love to see it integrated into `libGDX` too.

There is another implementation for [JavaScript](http://stackoverflow.com/a/23202637/4759433), but I took the one from the Processing team (the algorithm to calculate the mapped value is not mine, but it's not something private anyway since it's also available for JavaScript in a different way etc.)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,7 @@
 - API Change: removed JGLFW backend
 - Fixed mixed up use of TexturePacker.Settings.stripWhitespaceX|Y.
 - Added joystick POV support to LWJGL3 controller backend.
+- Add map() function to MathUtils, ported from Processing
 
 [1.9.6]
 - Fix performance regression in LWJGL3 backend, use java.nio instead of BufferUtils. Those are intrinsics and quite a bit faster than BufferUtils on HotSpot.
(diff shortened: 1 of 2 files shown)


mgsx-dev:
I planned to PR this too when I'll have time do do it : map function and nor function as well (inverse of lerp function).

So I vote Yes for this PR !

Lonami:
> Please use the formatter on the lines you have changed.

The formatter using `eclipse-formatter.xml` really messes up the examples on the comments:
```java
/** Returns a re-mapped float value from one range to another. Examples: 2 in [ 0, 10] -> 6 in [ 5, 10]. 0 in [-5, 5] -> 20 in
 * [10, 30]. 9 in [ 0, 10] -> 1 in [10, 0]. value in [inRangeStart, inRangeEnd] -> re-mapped in [outRangeStart, outRangeEnd] */
static public float map (float value, float inRangeStart, float inRangeEnd, float outRangeStart, float outRangeEnd) {
	...
```

But the comment may as well just be:
```java
/** Returns a re-mapped float value from inRange to outRange. */
```

This is to follow the style of *"returns…"*; I think it's clearer if it simply said *"Remaps value from inRange to outRange"*, so let me know if you need this change, too.

Lonami:
That's fair enough.

PR #4643, opened 2017-03-11
Sprite.setTextureFit() added.

bogdann96:
Most 24fps animations consist of too many frames and they can't fit in a 4096x4096 sprite sheet. In that case, every frame has to reside in a separate file. setTextureFit (Texture tex) resizes the sprite when a texture with different dimensions from the current one is set. I've chosen that name because it will pop up on the auto-completion menu when someone decides to call setTexture().

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -128,6 +128,7 @@ public class Sprite extends TextureRegion {
 
 	/** Sets the texture and resizes the sprite to the size of that texture. */
 	public void setTextureFit (Texture texture) {
+		this.texture = texture;
 		setSize (texture.getWidth(), texture.getHeight());
 		setRegion (0, 0, texture.getWidth(), texture.getHeight());
 	}
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -126,6 +126,12 @@ public class Sprite extends TextureRegion {
 		dirty = sprite.dirty;
 	}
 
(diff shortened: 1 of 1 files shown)


cypherdare:
Seems like a rather specialized use case, and only saves you one line of code. The `setRegion` call would be unnecessary for the use case you described.

Even if your animation takes up several pages of an atlas, the atlas will load much faster than individual textures for each frame, especially on Android.

Tom-Ski:
If you end up using a separate file for each frame, which you definitely shouldn't be doing, you should use a new sprite object for each frame.

bogdann96:
cypherdare, you are wrong about setRegion() though. It is necessary because if I call only setSize(), then I get a black rectangle. I have to set both of them.

Tom-Ski, separate object for each frame is even worse because if my animation is, for example, 43 frames, then I would need to change my hero's position 43 times every time he moves.

But you are right, my game is currently loading 7-8 secs and I have only a few animations. It's better if I switch to spritesheets.

cypherdare:
I would suggest not using the Sprite class at all if you are doing Animations. Sprite is kind of a weird, specialized class that isn't really appropriate for this. It's sort of conflated asset / game entity. I suppose it makes sense to use Sprite for game objects that rarely move (or have their region changed), because it has some optimizations for that. 

But a better OO approach would be to create your own game entity class that tracks position, orientation, color, etc., and then draws a TextureRegion with those parameters. A TextureRegion is an asset. The game entity technically probably shouldn't reference one at all, although that is perhaps taking OO to a cumbersome extreme. What I do as a compromise is pass my Assets wrapper to my various draw methods, and my entities can directly access what they need by Assets field, having never needed to hold an assets reference. It keeps model and view and assets separated to a certain extent that makes refactoring quite easy if that's ever necessary.

bogdann96:
Okay, thanks a lot for the help. :)

PR #4657, opened 2017-03-17
Added ETC2 conversion tool

devjn:
This makes possible to use KTXProcessor for generating ETC2 textures with a help of [etc2comp](https://github.com/google/etc2comp)


intrigus:
Does this work with Mac OS X?

devjn:
Nope, somebody needs to build a binary

badlogic:
Where does the binary come from? What's the license?

devjn:
https://github.com/google/etc2comp   Apache License 2.0

devjn:
Updated with Mac OS binary 9f7624d, should now work probably

Darkyenus:
macOS executables don't have extension .app, in fact, .app is something completely different, which makes this naming very confusing. macOS executables don't have any extension, I suggest renaming it to "etctool-mac" or something similar.

Same could be done with Linux.

devjn:
Done 004e348.

mgsx-dev:
not a good idea to include binaries in the repo, users can build them and stay up to date.

PR #4659, opened 2017-03-21
improve OpenALMusic getPosition accuracy fix #4268

### LINKED ISSUE #4268: Fix for getPosition() on OpenALMusic
#### Issue details

After trying to sync some music on desktop, I found out that music.getPosition() does not always return accurate values. This is due to the field "renderedSeconds" being set to 0 once there isn't data to be put on one of the buffer, even with the other two buffers still not processed.
#### Reproduction steps/code

The music being used for this example was .wav, 10s long, 16bit, 44100Hz, mono --> 10s_2bytes_44100Hz = 882000 bytes of music. It was set to loop and, on the first loop, the method music.getPosition() did not go all the way to 10s, instead only went to 9.287981859s before going back to 0, and here is why: 

The class OpenALMusic uses three byte buffers of maximum size 40960 bytes each to transfer data to the audio device. In this example it is needed 21 buffers of size 40960 bytes and 1 of size 21840 bytes to pass all the 882000 bytes of music data on each loop. The problem is that a parameter that gives the information on music.getPosition() is reset once there is no data on ONE of the buffers, not all three.

In this example: the last three buffers sizes of the loop are: 1-40960 bytes; 2-40960 bytes; 3-21840 bytes. Once the first buffer is completely processed, the music is reset, taking music.getPosition() to 0 again, but there still was (40960+21840)bytes = 0.712018141s of data before the end of the loop.

The following code snippet is from OpenALMusic. The update() method calls fill(int bufferID) and resets renderedSeconds to 0 once there isn't data to be put on one of the three buffers, leading to a value of 0 being returned from getPosition() before the end of the loop.

``` java
public float getPosition () {
      if (audio.noDevice) return 0;
      if (sourceID == -1) return 0;
      return renderedSeconds + alGetSourcef(sourceID, AL11.AL_SEC_OFFSET);
   }

public void update () {
      if (audio.noDevice) return;
      if (sourceID == -1) return;

      boolean end = false;
      int buffers = alGetSourcei(sourceID, AL_BUFFERS_PROCESSED);
      while (buffers-- > 0) {
         int bufferID = alSourceUnqueueBuffers(sourceID);
         if (bufferID == AL_INVALID_VALUE) break;
         renderedSeconds += secondsPerBuffer;
         if (end) continue;
         if (fill(bufferID))
            alSourceQueueBuffers(sourceID, bufferID);
         else
            end = true;
      }
      if (end && alGetSourcei(sourceID, AL_BUFFERS_QUEUED) == 0) {
         stop();
         if (onCompletionListener != null) onCompletionListener.onCompletion(this);
      }

      // A buffer underflow will cause the source to stop.
      if (isPlaying && alGetSourcei(sourceID, AL_SOURCE_STATE) != AL_PLAYING) alSourcePlay(sourceID);
   }

private boolean fill (int bufferID) {
      tempBuffer.clear();
      int length = read(tempBytes);
      if (length <= 0) {
         if (isLooping) {
            loop();
            renderedSeconds = 0;
            length = read(tempBytes);
            if (length <= 0) return false;
         } else
            return false;
      }
      tempBuffer.put(tempBytes, 0, length).flip();
      alBufferData(bufferID, format, tempBuffer, sampleRate);
      return true;
   }
```

1.9.4
#### Stacktrace

``` java
//Please provide the stacktrace if applicable 
```
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

mgsx-dev:
This is a first shot, maybe more tests would be necessary.

As @riqueaalves explained, renderedSeconds variable is reset too early when looping. My fix stores futur  renderedSeconds values for queued buffers. This should fix position value when last buffer is being loaded and not played yet.

Suggestions welcome!


mgsx-dev:
Should be OK now.

Tom-Ski:
Thanks @mgsx-dev !

PR #4662, opened 2017-03-23
Invalidating addActorBefore()/addActorAfter() with Table as they do not work (Fixes #4602)

### LINKED ISSUE #4602: addActorBefore()/addActorAfter() should not be used with Table

#### Issue details
[Table](https://github.com/libgdx/libgdx/blob/db805953849780161df7c1fb96f81248c19cff44/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java) inherits (indirectly) from [Group](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java), which means Table has the following three methods for adding a child node:

* [add(Actor)](https://github.com/libgdx/libgdx/blob/db805953849780161df7c1fb96f81248c19cff44/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java#L194)
* [addActorBefore(Actor, Actor)](https://github.com/libgdx/libgdx/blob/db805953849780161df7c1fb96f81248c19cff44/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java#L290)
* [addActorAfter(Actor, Actor)](https://github.com/libgdx/libgdx/blob/db805953849780161df7c1fb96f81248c19cff44/gdx/src/com/badlogic/gdx/scenes/scene2d/Group.java#L304)

A developer using Scene2d *might* reasonably assume that all of these will work, when in fact, only the first method works -- the other two methods will execute and return normally, but will not layout the newly-added actor correctly (it'll be laid out underneath the last child of the Table).

#### Reproduction steps/code
_Even if you think your issue is trivial to reproduce, please supply a [SSCCE](http://sscce.org/) that demonstrates your issue.  This saves time on our end, and makes it much more likely that your issue will be fixed.
You can find barebones templates [here](https://github.com/libgdx/libgdx/wiki/Getting-help)_

I'm not sure this requires repro steps or code. Let me know if you want them.

LibGDX 1.9.5, but current mainline appears affected as well.

#### Stacktrace
No stacktrace.

#### Affected platforms
All.

#### What I'm looking for

I'm looking for anything that would have saved me an hour of fiddling with `pack()` and `invalidateHierarchy()` just to find that these methods just don't work presently. I'd be happy with any of the following:

1. Override `addActor{Before|After}()` in Table so they work as expected.
2. Override `addActor{Before|After}()` in Table and make them throw an exception.
3. Override `addActor{Before|After}()` in Table and have it call `super`, but *also* annotate the method as `@Deprecated` with a reason why.

I can see that the [documentation](https://github.com/libgdx/libgdx/wiki/Table#inserting-cells) does mention this exact issue, near the very bottom, but I didn't see this and I suspect others might miss it as well.

MartensCedric:
I did the suggestion 2. (throw an exception) as suggested in the bottom of Issue #4602

I added documentation to the methods and quoted the documentation [here](https://github.com/libgdx/libgdx/wiki/Table#inserting-cells) which explains that these are not supported and suggests an alternative way to accomplish the desired result.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
@@ -31,6 +31,7 @@ import com.badlogic.gdx.utils.Align;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.Pool;
 import com.badlogic.gdx.utils.Pools;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 /** A group that sizes and positions children using table constraints. By default, {@link #getTouchable()} is
  * {@link Touchable#childrenOnly}.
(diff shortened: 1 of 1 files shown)


NathanSweet:
While it is a slight API gotcha, in some cases it's convenient to add actors to a table without having them participate in table layouting.

PR #4664, opened 2017-03-23
Convenience Constructor Methods HorizontalGroup and VerticalGroup

zeromancer:
Same convenience constructor methods already exist for Stack.java so it would be nice (and consistent) to have them also in HorizontalGroup and VerticalGroup


Tom-Ski:
Varargs are generally avoided across the framework due to the additional garbage they create. If you want to do this please use your own utility method.

PR #4692, opened 2017-04-13
Fix GWT-Pixmap. Add missing format.

intrigus:
Pixmap's `format` is null, because #4234 incorrectly removed it.
See https://github.com/libgdx/libgdx/pull/4234/files#diff-193f92909143c2ad5b478ee1618ce414L126


Tom-Ski:
Thanks @intrigus

PR #4695, opened 2017-04-14
Add support for glBlendFuncSeparate blending.

TomGrill:
To make Batches blending more flexible I added glBlendFuncSeparate support.


NathanSweet:
FWIW, I use -1 for `blendSrcFunc` to prevent the batch from managing blending. If others like this PR, I'm OK with it.

PR #4701, opened 2017-04-16
Added setX and setY methods with alignment argument

nevabyte:
Found myself looking for these methods so decided to create them.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -408,12 +408,42 @@ public class Actor {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
```diff
@@ -408,6 +408,21 @@ public void setX (float x) {
 		}
 	}
 
+	/** Sets the x position using the specified {@link Align alignment}. Note this may set the position to non-integer
+	 * coordinates. */
+	public void setX (float x, int alignment) {
+
+		if ((alignment & right) != 0)
+			x -= width;
+		else if ((alignment & left) == 0) //
```
cypherdare: Should be `!=`?

PR #4710, opened 2017-04-21
Cleaned FileHandle#copyTo

LiliNear:
Line 583 was always false


xoppa:
Please carefully read https://github.com/libgdx/libgdx/wiki/Contributing and make sure to apply the formatter.

LiliNear:
Hello,
I applied the formatter but I had to modify the tab size from 3 (imported value) to 4 because I'm working on Intellij IDEA. (It was completing the tabs with spaces)
I also sent my contribution agreement.

NathanSweet:
It appears the PR isn't equivalent to the original code: `dest = dest.child(name());` doesn't happen for the outer `else`. This method is complex (see javadoc) so I'm hesitant to fix what ain't broke. I'll remove the `sourceDir` check which is always false.

PR #4715, opened 2017-04-23
Update GWT controller extension and add POV support

Zomby2D:
Due to changes in the way browsers handle controllers since the extension was written, the GWT controllers extension wasn't working anymore. Fortunately, all major browsers now pretty much conforms to the W3C working draft so the extension shouldn't break too much in the future.

This updates the extension to use the currently supported controller api. I've also added proper POV support for gamepads reporting a "standard" mapping. In my tests, only XInput gamepads reported as "standard" and everyting else had different ways of reporting their pov, which also differed between browsers.

This was tested as working on Chrome, Firefox and Opera.

### DIFF
--- a/extensions/gdx-controllers/gdx-controllers-gwt/src/com/badlogic/gdx/controllers/gwt/GwtControllers.java
+++ b/extensions/gdx-controllers/gdx-controllers-gwt/src/com/badlogic/gdx/controllers/gwt/GwtControllers.java
@@ -21,14 +21,21 @@ import com.badlogic.gdx.controllers.Controller;
 import com.badlogic.gdx.controllers.ControllerListener;
 import com.badlogic.gdx.controllers.ControllerManager;
 import com.badlogic.gdx.controllers.gwt.support.Gamepad;
+import com.badlogic.gdx.controllers.gwt.support.GamepadButton;
 import com.badlogic.gdx.controllers.gwt.support.GamepadSupport;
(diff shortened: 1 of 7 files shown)


extensions/gdx-controllers/gdx-controllers-gwt/src/com/badlogic/gdx/controllers/gwt/GwtControllers.java
```diff
@@ -140,67 +150,102 @@ public void onGamepadConnected(int index) {
...
-					float newButton = (float) buttons.get(i);
-					if (oldButton != newButton) {
-						if ((oldButton < 0.5f && newButton < 0.5f) || (oldButton >= 0.5f && newButton >= 0.5f)) {
-							controller.buttons.put(i, newButton);
-							continue;
+					float newButton = (float)buttons.get(i).getValue();
+					if (controller.isStandardMapping() && i >= 12 && i <= 15) {
+						int direction = controller.pov;
+						if (newButton < 0.5f) {
+							if (i == 12)
```
Tom-Ski: Can we get rid of these magic numbers?

Zomby2D: Sure thing... I'll replace them with proper constants and update the PR later today

Zomby2D:
The directional pad standard values have been converted to constants for clarity

Tom-Ski:
Thanks @Zomby2D

PR #4719, opened 2017-04-25
Added touch input delta to RemoteInput

Mikeywalsh:
Added two new arrays in RemoteInput, deltaX and deltaY. They record the difference in X and Y position between input readings.

Also updated the existing getDeltaX and getDeltaY methods, to return the corresponding new values.


Tom-Ski:
Thanks!

PR #4726, opened 2017-05-02
Added usefull method overloading of espilonEquals for Vector classes

WinterGuardian:
You are no longer obligated to provide a epsilon, the default parameter is now MathUtils.FLOAT_ROUNDING_ERROR making it easier to use and more readable


Tom-Ski:
Great, thanks!

PR #4736, opened 2017-05-10
Fix Ellipse docs

cypherdare:
Minor nitpick, but the units should be considered arbitrary. World size is confusing enough as it is for newcomers.


xoppa:
Thanks

PR #4753, opened 2017-05-21
Recognize X-Box 360 controller as Xbox controller

winniehell:
On Linux I get `Microsoft X-Box 360 pad` as `controller.getName()`, so currently `isXboxController()` returns `false` for me.

cc @WilliamHayward

### DIFF
--- a/extensions/gdx-controllers/gdx-controllers/src/com/badlogic/gdx/controllers/mappings/Xbox.java
+++ b/extensions/gdx-controllers/gdx-controllers/src/com/badlogic/gdx/controllers/mappings/Xbox.java
@@ -154,6 +154,6 @@ public class Xbox {
 	/** @return whether the {@link Controller} is an Xbox controller
 	 */
 	public static boolean isXboxController(Controller controller) {
-		return controller.getName().contains("Xbox");
+		return controller.getName().contains("Xbox") || controller.getName().contains("X-Box");
 	}
 }


winniehell:
@Tom-Ski Can you have a look at this please? :smiley:

tdgroot:
@winniehell Please take a look at an old PR I made some time ago, I encountered the same problem as you: https://github.com/libgdx/libgdx/pull/3969/files#diff-c282a5cbb1ce4dcd0eee21184296851eL146

winniehell:
@tdgroot Thank you! Where did you get the other values for the regex from?

tdgroot:
I tested it using several Xbox controllers, the controller names weren't really consistent.

WickedShell:
The proposed patch (or the regex) would also correct and add support for XBox one controllers which would be great. Under Linux I see "Microsoft X-Box One S pad" which does map out to the identical commands as the 360.

As a side note it would probably be worth listing the example strings that forced the regex/contains to look the way it does so that when looking at it in the future their is a clear list of the minimum things that need to pass.

DenialAdams:
I've had to do this exact fix inside of my own projects(under linux, at least), so +1 from me!

winniehell:
I agree that we should collect the existing names before we craft the regular expression. I only have access to one controller. Maybe can collect the different names here and I open a new pull request with the corresponding regular expression?

Also I am wondering if there isn't a more reliable criteria to identify a controller such as vendor ID?

Issue #4461, opened 2016-12-06
Aligned Particle Emitter not updating rotation on Gravity


azakhary:
#### Issue details
Trying to create a "sparks" effect. I am using emitter that emits particles up with velocity that decays over time. There is also a gravity set. Thus making particles change their angle over time. Because particle is set to be "aligned" this has to also change the rotation real-time. But only initial rotation is maintained.

#### Reproduction steps/code
This issue is easy to reproduce in particle editor. just use this effect: http://pastebin.com/pq2E7xAk

#### Version of LibGDX and/or relevant dependencies
1.9.4

#### Note
While example is with usage of gravity, the issue is not related to gravity. And is happening if you just try to change angle over time. rotation is not changing over time according to movement direction.
I am unsure if it's a bug or by design.

piotr-j:
It looks like aligned is not handled when gravity or wind are used
https://github.com/libgdx/libgdx/blob/9b8f20b2d0e96e53f0f2b98dc8f6131c810aae71/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java#L563
Probably an oversight.

PR #4764, opened 2017-05-26
Update wind and gravity first so they are taken into account by aligned

piotr-j:
Currently 'aligned' 2d particle effect setting doesnt take wind and gravity into account. This simple change corrects this issue. 
See #4461


piotr-j:
Actually, this doesnt quite work.

PR #4765, opened 2017-05-29
Fix mistake in TextButtonStyle copy constructor

Radnap:


intrigus:
Good catch +1

xoppa:
Thanks!

Issue #4758, opened 2017-05-23
Key events not firing on games within Facebook


WillCalderwood:
#### Issue details
`keyDown`, `keyUp` and `keyTyped` are not being fired when a game in played within Facebook. The certainly worked in the past but I'm not sure if something has changed Facebook side or the libgdx side.

I've added logging on the keypress events on the menu screen.

Open a game outside of Facebook here and display the console within the browser and press some keys and you'll see logging.
https://wordburst-html5.appspot.com/build/index.jsp

Open a game within Facebook (Facebook points to the above URL, so it's the same code) here and display the console within the browser. Press keys and you'll see no logging.
https://apps.facebook.com/scribble-wordburst/

(Tested in Chrome and Firefox)

#### Reproduction steps/code
```
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.InputProcessor;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.Texture;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;

public class MyGdxGame extends ApplicationAdapter implements InputProcessor {
    SpriteBatch batch;
    Texture img;
    private int x = 0;
    private int y = 0;

    @Override
    public void create() {
        batch = new SpriteBatch();
        img = new Texture("badlogic.jpg");
        Gdx.input.setInputProcessor(this);
    }

    @Override
    public void render() {
        Gdx.gl.glClearColor(1, 0, 0, 1);
        Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
        batch.begin();
        batch.draw(img, x, y);
        batch.end();
    }

    @Override
    public void dispose() {
        batch.dispose();
        img.dispose();
    }

    @Override
    public boolean keyDown(int keycode) {
        x += 10;
        return false;
    }

    @Override
    public boolean keyUp(int keycode) {
        y += 10;
        return false;
    }

    @Override
    public boolean keyTyped(char character) {
        if (character == 'x') {
            x = 0;
            y = 0;
        }

        return false;
    }

    @Override
    public boolean touchDown(int screenX, int screenY, int pointer, int button) {
        return false;
    }

    @Override
    public boolean touchUp(int screenX, int screenY, int pointer, int button) {
        return false;
    }

    @Override
    public boolean touchDragged(int screenX, int screenY, int pointer) {
        return false;
    }

    @Override
[...]

PR #4772, opened 2017-06-07
Fix issues with key events not firing in a frame

WillCalderwood:
I'm hoping I've changed this in the correct place.

See issue https://github.com/libgdx/libgdx/issues/4758


Tom-Ski:
Thanks!

PR #4776, opened 2017-06-09
Clean up Android setup files

cypherdare:
The more recent Android tools plugin 2.2.0 is now necessary for instant run.

Also, I tweaked a few things to make it more like a default Android project. I think it's best to keep it as similar to a default Android project that Android Studio would create, so it's easier to follow along with tutorials and what not for the non-LibGDX stuff.

- `proguard-project.txt` is now called `proguard-rules.pro`.
- `versionCode` and `versionName` are specified in `build.gradle` instead of the manifest.
- A default release buildType is available such that proguard can be enabled simply by switching `minifyEnabled` to true. The `proguardFiles` line matches a default Android Studio project now.
- ` project.properties` is obsolete. The target is specified in the `build.gradle` file.


Tom-Ski:
Great :] Have you tested this in eclipse?

cypherdare:
I ran the setup from Eclipse, imported a generated project, and tested it
on a device. I can run a release build later tonight to test the Proguard
if you think it's necessary. It's the same Proguard setup I manually
changed my own project to, and works for me.

Tom-Ski:
No, that looks fine and its disabled by default anyway which is good.  To clarify, did you use ADT? Or did you only rely on Gradle commands when installing the test device?

cypherdare:
I used the Gradle run command that is automatically available via the run
button in IDEA.

cypherdare:
Sorry, didn't understand what you were asking at first. Did you mean ADB? I have tested it with the IDEA run configuration that the android build.gradle automatically sets up. And I have run `gradlew installDebug` and that worked fine as well. But I thought either way it is running through ADB to connect to the device.

Tom-Ski:
Through https://developer.android.com/studio/tools/sdk/eclipse-adt.html, the android integration suite that most eclipse users still use. Gradle will work fine alone as it should as its standalone, but its important to check ADT as that is still a useful tool for many users.

cypherdare:
Ah, OK. ADT is a bad repressed memory for me. 

Yes, removing `project.properties` causes an ADT project to complain. I'll get it working again and update this PR. The reason I removed the file was that the comment about getting proguard working led me astray for an hour before I figured it out. But I can update the comment instead.

cypherdare:
@Tom-Ski 
I got the ADT project working fine. But I discovered that setting the Android build tools to 2.2.0 requires the Gradle version in the wrapper to be at least 2.14.1, whereas we've been using 2.10.

This is actually an issue I run into every time I start a new project via the Setup app. I need the features of the more recent build tools and have to go edit the gradle wrapper properties to update Gradle.

So I feel like this should be fixed. But I don't know what we're risking if we update the Gradle version. I put in the latest, 3.5, but I don't know if it will affect projects with platforms I don't work with (robovm, moe, gwt).

If we want to wrap up this PR, I can change the tools version back and the gradle files, and can debate updating the wrapped gradle version in a new issue.

Tom-Ski:
Thats not compatible across all IDEs currently, so should not be updated in this PR.

cypherdare:
OK, it's updated with wrapper 2.14.1.  I generated a fresh wrapper via `gradle wrapper --gradle-version 2.14.1`. We hadn't customized any of the gradle wrapper scripts, had we?

Tom-Ski:
No, looks good, thanks @cypherdare !

PR #4783, opened 2017-06-15
Added "Use Kotlin" option for setup tool

snordgren:
Supporting the new languages was primarily implemented through using a new Language enum containing the language-specific Gradle-related strings for the two supported languages (Java and Kotlin). Could be extended in the future to include other languages like Scala or Clojure.

I also edited the project build.gradle templates to be agnostic as to what language plugin they use.

This is feature that I personally felt I wanted to have, as tweaking new projects to compile Kotlin properly took a lot of time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ libgdx-*.zip.MD5
 /extensions/gdx-freetype/libs/
 /extensions/gdx-image/libs/
 /extensions/gdx-setup/gdx-setup.jar
+/extensions/gdx-setup/test/
 
 #the LWJGL3 libs are pulled via fetch.xml
 /backends/gdx-backend-lwjgl3/libs/lwjgl*.jar
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
@@ -27,8 +27,9 @@ public class BuildScriptHelper {
 
 	private static int indent = 0;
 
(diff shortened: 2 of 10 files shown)


Tom-Ski:
How are you handling kotlin and gwt?

Scellow:
> KOTLIN ("kotlin", "ext.kotlinVersion = '1.0.0'", 

Please..  latest version is 1.1.2

snordgren:
>How are you handling kotlin and gwt incompatibility?

As of right now it is ignored and a GWT build is created that can never run due to using Kotlin. Ideally the option should warn like other non-GWT compatible options do. 

>Please.. latest version is 1.1.2

True. My reasoning with using 1.0.0 was to leave version choice to the user, meaning the value doesn't have to be updated with each new version of Kotlin. But if you'd prefer that the value be 1.1.2 (1.1.2-5?), then sure, I will change that!

Any other thoughts, criticisms, etc?

Tom-Ski:
It would be better to disable the html project completely when generating with kotlin, and give warnings when selecting kotlin (or any other jvm language in future)

snordgren:
I find the Eclipse plugin [bundles 1.1.1](https://github.com/JetBrains/kotlin-eclipse/blob/ee56474801bd40f65767a6e59069423b4ec30b9e/pom.xml), and [so does Netbeans](https://github.com/JetBrains/kotlin-netbeans/blob/master/pom.xml). IDEA is usually up to date. So then 1.1.1 for the current version?

I will try to hook Kotlin to the same warning shown when selecting other GWT incompatible options, and prevent generation of GWT projects when a non-Java language is selected.

PR #4790, opened 2017-06-23
Add method resetScale to BitmapFont

yuripourre:


Tom-Ski:
What is the purpose of this method? This doesn't reset the font to the default value, (1)

yuripourre:
I am using this method to reset the font to the original scale. I thought it might be useful.

When I call setScale(2) the scale is applied to the font and it get bigger, if I call setScale(1) the scale is applied and the font keeps the same (bigger) size.

I just wanted a method to get the original size back. Please let me know if there is a method to do this already, I couldn't find it.

Maybe it is very specific to my needs.

PR #4792, opened 2017-06-25
Added useful utility methods to the Array and IntArray classes

jeremygillespiecloutier:
I added a shift method to Array that shifts every elements by a given offset, as well as a selectMin and selectMax method. I also added a sum method to IntArray that returns the sum of every element in the array.


jeremygillespiecloutier:
I included code formatting in my latest commit.

Tom-Ski:
Thanks!

PR #4798, opened 2017-06-27
Spinner UI element

jeremygillespiecloutier:
Spinner UI element (and SpinnerTest to demonstrate)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,210 @@
...
+				invalidate();
+				return true;
+			}
+
+			@Override
+			public void touchUp (InputEvent event, float x, float y, int pointer, int button) {
+				// Update spinner value, and fire change events if needed
+				if (getCurrentBackground() == getStyle().backgroundNext) {
+					if (getModel().next()) Spinner.this.fire(new ChangeListener.ChangeEvent());
+				} else if (getCurrentBackground() == getStyle().backgroundPrev) {
```
Darkyenus: This is a hack, logic should not depend on style.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,210 @@
...
+				if (x >= 0 && x <= getWidth() && y >= 0 && y <= getHeight()) {
+					if (amount > 0) {
+						if (getModel().previous()) Spinner.this.fire(new ChangeListener.ChangeEvent());
+					} else if (amount < 0) {
+						if (getModel().next()) Spinner.this.fire(new ChangeListener.ChangeEvent());
+					}
+					invalidate();
+				}
+				return false;
+			}
```
Darkyenus: Changing the value with up/down keys would also be nice.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,210 @@
...
+
+	public SpinnerModel getModel () {
+		return this.model;
+	}
+
+	/** This class defines the appearance of a Spinner. it includes the font used and the drawables for idle state, next button
+	 * pressed and previous button pressed. (See gdx-tests SpinnerTest) */
+	public static class SpinnerStyle {
+
+		public Drawable background, backgroundNext, backgroundPrev; // Backgrounds for the spinner states
```
Darkyenus: I am not sure if the way it is drawn (single background with buttons baked in) is the correct approach, but why not.
However it would be nice to also add on-hover images, to reach visual parity with libGDX's `Button`.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SpinnerNumberModel.java
```diff
@@ -0,0 +1,69 @@
...
+	 * @param max The maximum value the spinner can take
+	 * @param step The absolute increment when the next or previous button on the spinner is pressed
+	 * @param value The current value of the spinner */
+	public SpinnerNumberModel (int min, int max, int step, int value) {
+		if (max < min) throw new IllegalArgumentException("The maximum cannot be smaller than the minimum");
+		if (value < min || value > max)
+			throw new IllegalArgumentException("The current value must be between the minimum and the maximum (inclusive)");
+		this.min = min;
+		this.max = max;
+		if (step < 0) step *= -1;
```
Darkyenus: Nitpick: wouldn't `step = -step` be clearer? (and is this check even necessary?)

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
jeremygillespiecloutier: I have tried implementing all your proposed changes in my latest commit.

jeremygillespiecloutier:
I have tried implementing all your proposed changes in my latest commit.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,262 @@
...
+				updateBackgroundForCoords(x, y);
+				return true;
+			}
+
+			@Override
+			public void touchUp (InputEvent event, float x, float y, int pointer, int button) {
+				// Update spinner value if mouse is hovering over next or previous button
+				if (getState() == SpinnerState.NEXT)
+					updateValue(true);
+				else if (getState() == SpinnerState.PREVIOUS) updateValue(false);
```
Darkyenus: Did this go through formatter?

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,262 @@
...
+			currentBackground = getStyle().backgroundHover;
+		}
+		invalidate();
+	}
+
+	@Override
+	public void draw (Batch batch, float parentAlpha) {
+		currentBackground.draw(batch, getX(), getY(), getWidth(), getHeight());
+		// Center the text
+		style.font.draw(batch, model.getDisplayValue() + "", getX() + style.background.getLeftWidth(),
```
Darkyenus: You already set the display value text into the `layout`, you can (and should) also use it for drawing. Also, this allocates garbage even if `getDisplayValue` returns `String`.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	public SpinnerModel getModel () {
+		return this.model;
+	}
+
+	/** A spinner has four state. When the mouse is not over it, it is IDLE. When the mouse is over it, it is either over the NEXT
+	 * button, the PREVIOUS button or otherwise simply in a HOVER state (mouse is over the widget, but not over the buttons) */
+	enum SpinnerState {
+		IDLE, HOVER, NEXT, PREVIOUS;
```
Darkyenus: I am not sure on this, but Android, sadly, didn't like enums too much, IIRC. Since this is just internal state, byte constants (C-style) should be fine. Also, it should be possible to have just three states, without the `HOVER`, but I might be mistaken.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,262 @@
...
+		/** Constructor
+		 * @param background Drawable for the idle state
+		 * @param backgroundHover Drawable for the general hover state
+		 * @param backgroundNext Drawable for the next button hover state
+		 * @param backgroundPrev Drawable for the previous button hover state
+		 * @param font Font used for display */
+		public SpinnerStyle (Drawable background, Drawable backgroundHover, Drawable backgroundNext, Drawable backgroundPrev,
+			BitmapFont font) {
+			this.background = background;
+			this.backgroundHover = backgroundHover;
```
Darkyenus: I meant one hover for next and one for down. Sorry for not being clearer :/

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SpinnerArrayModel.java
```diff
@@ -0,0 +1,103 @@
...
+
+	/** Constructor
+	 * @param array The array of elements held by the spinner
+	 * @param displayValues Array containing the text displayed in the spinner for each element
+	 * @param decoupleDisplay If false, toString() will be used to display values, if true, otherwise elements in displayValues
+	 *           will be used.
+	 * @param index The index of the current element in the array
+	 * @param wrap If true, calling next() from the last element will move to the first element (instead of doing nothing). Similar
+	 *           behavior for previous(). */
+	private SpinnerArrayModel (Array<T> array, Array<CharSequence> displayValues, boolean decoupleDisplay, int index,
```
Darkyenus: Interestring approach. I would probably just let users to override `getDisplayValue` and default to `toString`. But this is your call, whichever approach you like more.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SpinnerDecimalModel.java
```diff
@@ -0,0 +1,79 @@
...
+
+package com.badlogic.gdx.scenes.scene2d.ui;
+
+import java.math.BigDecimal;
+
+import com.badlogic.gdx.scenes.scene2d.ui.Spinner.SpinnerModel;
+
+/** Spinner model used to display decimal numbers
+ * @author Jeremy Gillespie-Cloutier */
+public class SpinnerDecimalModel extends SpinnerModel<BigDecimal> {
```
Darkyenus: Games probably won't use BigDecimals much, floats (or doubles, if you insist) should be enough.

By the way, `toString` is much better than ` + ""` (for example `Integer.toString(int)` for primitives), since `toString` allocates only String, but `+ ""` has to allocate `StringBuilder` and then `String`.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SpinnerNumberModel.java
```diff
@@ -0,0 +1,75 @@
...
+	public boolean previous () {
+		if (value - step >= min) {
+			value -= step;
+			return true;
+		}
+		return false;
+	}
+
+	@Override
+	public Integer getValue () {
```
Darkyenus: Nitpick, this Model could also have separate getter with return value `int`, to avoid boxing.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Spinner.java
```diff
@@ -0,0 +1,262 @@
...
+ * on the mouse, or tap the "next" (up) and "previous" (down) portions of the background. A SpinnerStyle is made of 3 drawables,
+ * two for when the next or previous buttons are pressed, and one by default.
+ * 
+ * @author Jeremy Gillespie-Cloutier */
+public class Spinner extends Widget {
+
+	private SpinnerStyle style;
+	private GlyphLayout layout;
+	private SpinnerModel model;
+	private Drawable currentBackground; // The current background drawable, either default, next button pressed or previous button
```
Darkyenus: Nitpick: Even private fields can have javadoc-style comments :)

MrStahlfelge:
My thoughts:
* Name is a problem. For me, a spinner is an element indicating wait time. That's not the case here, it is some kind of carousel view.
* I am not sure if it is a good idea to add it to the main project, but if it is better to release as an own lib or gist.

Darkyenus:
I don't think that name is a problem as Java's swing [also names this component a spinner](https://docs.oracle.com/javase/tutorial/uiswing/components/spinner.html). Carousel is something different and not typically associated with (number) input.

PR #4803, opened 2017-06-29
Update build.gradle

athenacadence:
ios moe eclipse project will now be named appName+'ios-moe" instead of just "ios-moe". Otherwise, ios-moe eclipse projects do not match other platforms' project naming conventions


Tom-Ski:
Thanks!

PR #4813, opened 2017-07-06
Add Box2D createChain/Loop variants with offset and length

Darkyenus:
Currently I have to allocate a new `float[]` for each shape even though it would be possible to just use these and pass in my float buffer array.

I have never modified anything JNI related in libGDX, so please somebody check if I'm doing it right.


Tom-Ski:
retest this please

PR #4823, opened 2017-07-17
Add resetScale method to BitmapFont

yuripourre:
I am using this method to reset the font to the original scale. I thought it might be useful.

When I call setScale(2) the scale is applied to the font and it gets bigger, then if I call setScale(1) the font font keeps the same size (2).

I just needed a method to get the original size back. Please let me know if there is a method to do this already, I couldn't find it.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -246,6 +246,13 @@ public class BitmapFont implements Disposable {
 		cache.getColor().set(r, g, b, a);
 	}
 
+	/** Reset the font to the original size */
+	public void resetScale () {
+		data.setScale(1 / data.scaleX, 1 / data.scaleY);
+		data.scaleX = 1;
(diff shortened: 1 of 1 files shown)


slebed:
Why not use setScale(.5) to reset the size when you double it with setScale(2)?

Tom-Ski:
A setScale method should set the scale absolutely, which is apparently not what this method is doing currently?

Lestat:
yeah and another method called .scale should do it like this one is doing it.

obigu:
@yuripourre What libGDX version are you seeing this issue on? This should be fixed on 1.9.6.

yuripourre:
Hi guys, thank you for the answers and sorry for the late.

@slebed: I am calling 1/2 = .5 that's what I want to do with this method (a general way to reset the scale).

@Tom-Ski: I was probably a bug from versions older than 1.9.6, pointed by @obigu 

@Lestat I tried this method but it is relative to the font's current scale (the same problem as I pointed)

@obigu You are right, I updated to 1.9.6 and now it works perfectly!

PR #4825, opened 2017-07-21
3D Particles Extension

perseus93:
We introduced some additional functionality to the 3D particle system:

- Independent X and Y scaling for billboard particles
- Billboard particles can now be aligned to their movement on the screen (Using the camera's projection).
- Billboard particles' blending function can now be set.

To get a general idea of what the possibilities are with these additions I refer to this example video:
https://youtu.be/HJ-sTNyDDr0

(Showing the scaling and alignment additions)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleChannels.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleChannels.java
@@ -83,6 +83,20 @@ public class ParticleChannels {
 		}
 	}
 
+	public static class Scale3Initializer implements ChannelInitializer<FloatChannel> {
+		private static Scale3Initializer instance;
+
+		public static Scale3Initializer get () {
+			if (instance == null) instance = new Scale3Initializer();
+			return instance;
+		}
+
+		@Override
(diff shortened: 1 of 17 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/particles/ParticleChannels.java
```diff
@@ -83,6 +83,20 @@ public void init (FloatChannel channel) {
 		}
 	}
 
+	public static class Scale3Initializer implements ChannelInitializer<FloatChannel> {
```
Tom-Ski: Should this be Scale2?

perseus93: Correct! Sorry, I missed it during one of the renaming sessions I had.
I'll run a sweep through the other classes to check if I missed anything else!

perseus93: Okay, I fixed the naming and also found that the channel's width was still set to 3 (reduced it to 2).
Thank you for your time by the way! It's appreciated!

PR #4839, opened 2017-07-29
Edited a comment

GadWissberg:
After spending few hours to understand why I get the "Couldn't load dependencies of asset" when tried loading same font with different parameters, I've found out that the name cannot be fully arbitrary as it says in the comment - It must have the suffix of file format.


Lestat:
It will look better with no caps.

GadWissberg:
Changed the comment to a lower-case.

PR #4844, opened 2017-08-03
modify GestureDetector

McoyJiang:
fix bug about long press not call under some circumstances

### DIFF
--- a/gdx/src/com/badlogic/gdx/input/GestureDetector.java
+++ b/gdx/src/com/badlogic/gdx/input/GestureDetector.java
@@ -107,7 +107,7 @@ public class GestureDetector extends InputAdapter {
 				longPressFired = false;
 				tapSquareCenterX = x;
 				tapSquareCenterY = y;
-				if (!longPressTask.isScheduled()) Timer.schedule(longPressTask, longPressSeconds);
+				if (!longPressTask.isScheduled()) new Timer().schedule(longPressTask, longPressSeconds);
 			}
 		} else {
 			// Start pinch.


Darkyenus:
I am not sure what the issue is, but creating a new `Thread` on each click is probably not the right solution. When is the long press not called - what does this fix?

Tom-Ski:
Agreed. What is the issue? How can we reproduce?

McoyJiang:
### under this circumstance, the longPress method won't be call :
1 if there are two AndroidApplication classes, and within each of them there is a specific AndroidApplicationAdatper.  and make a event to jump from one Adapter to Another Adapter
2 in each AndroidApplication Adapter initialize the absolutely same ActorGestureListener, and addListener to Actor
3 the second ActorGestureListener's longPress won't be called

example link:
[https://github.com/McoyJiang/GestureDetectorTest](url)

Darkyenus:
libGDX is not designed to run multiple instances of itself on one VM. I'm surprised that this was the only thing that broke...

McoyJiang:
since Android can show many activities, there must be many occasions that we will use multiple Stage or ApplicationAdapter.  Then how could we solve this problem ?

Darkyenus:
This is exactly what [Game](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/Game.java) and [Screen(Adapter)](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/ScreenAdapter.java) is for :)

McoyJiang:
so in Android System, if we want to use Libgdx, then there must be only one Activity ?

Darkyenus:
Not necessarily one Activity, just only one libGDX activity. Making your whole game in a single activity is also better for portability - other platforms don't have any concept of Android Activities. If you structure your game correctly, you can even develop most of your game on desktop, which is arguably faster/easier.

(If you want to ask more, it would be better to use [forums or IRC](https://github.com/libgdx/libgdx/wiki/Community-%26-Support) as this discussion is not a good fit for the issue tracker)

Issue #4698, opened 2017-04-15
Gradle 4.0 Support 


raincole:
It's kind of sad that gdx-setup is still using Gradle 2.10 by default. Espeically the libGDX itself is built with [Gradle 3.1](https://github.com/libgdx/libgdx/blob/master/gradle/wrapper/gradle-wrapper.properties), which means one has to use two different major versions of Gradle if he works from source. Shouldn't the generated project use the same version as libGDX does?

[Gradle 3.5](https://docs.gradle.org/current/release-notes.html) just released with some good features. I'll suggest upgrading both libGDX and gdx-setup project to 3.5. I actually upgraded the Gradle wrapper of my project to Gradle 3.5 and it worked on desktop, but I haven't tested it on android/ios/gwt.

raincole:
After further testing, I found this upgrade breaks my IDEA scopes because [Gradle changed how scope mapping works](https://docs.gradle.org/3.4/release-notes.html#idea-mapping-has-been-simplified).

To fix this one has to add this in `build.gradle`:

```
subprojects {
    if(it.name != "android") {
        apply plugin: "java"
        apply plugin: "idea"
        idea {
            module {
                scopes.PROVIDED.minus += [configurations.compile]
                scopes.COMPILE.plus += [configurations.compile]
            }
        }
    }
}

```

To make them back to `COMPILE` scope again. Just a small heads-up for anyone who wants to upgrade his Gradle.

Tom-Ski:
Damn, looks like they broke integration quite well.  Its fixed in 2017.1, but not for us as we have an android project.  Your workaround should be good, and it's not too nasty.  Command line we are still good, just need to test AS, eclipse and netbeans also.

FlorentMasson:
I have successfully upgraded my libgdx project to gradle 3.5, but be aware of this issue : https://issuetracker.google.com/issues/37137480
I finally landed on gradle plugin 2.2.3 although Android Studio keeps asking you to upgrade more...

raincole:
Sorry for self-bumping, but is there any plan on this?

Scellow:
And upcoming gradle 4.0 looks even more interesting https://blog.gradle.org/blazing-fast-android-builds

Tom-Ski:
[...]

PR #4848, opened 2017-08-08
Updated the setup tool for the HTML backend

MonsterOfCookie:
Updated the setup tool to handle the deprecation of the Jetty plugin from the Gradle distribution (4+)
Moved over to using a simple file server to serve the content whilst the SuperDev mode is handled by GWT and it's internal Jetty
Changed the GWT plugin to a newer one, it's fork of what was there before but this is actively maintained.

This is what I use for my HTML5 based projects now - Thought I should give back to the community :)

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gradle/wrapper/gradle-wrapper.properties
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/html/GdxDefinitionSuperdev
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/html/GdxDefinitionSuperdev
@@ -1,12 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit trunk//EN" "http://google-web-toolkit.googlecode.com/svn/trunk/distro-source/core/src/gwt-module.dtd">
 <module rename-to="html">
-%GWT_INHERITS%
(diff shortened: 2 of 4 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
```
Darkyenus: This seems to update Gradle to 4.1 - there has been some discussion about that: #4698 (and maybe more)

MonsterOfCookie: Ah good point. This was just more out of habit from my projects.

extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/html/GdxDefinitionSuperdev
```diff
@@ -1,12 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit trunk//EN" "http://google-web-toolkit.googlecode.com/svn/trunk/distro-source/core/src/gwt-module.dtd">
 <module rename-to="html">
-%GWT_INHERITS%
```
Tom-Ski: Why is this removed?

MonsterOfCookie: Wasn't needed as it is inherited from the GdxDefinition.gwt.xml

If it is needed, I haven't found the use case where it breaks without it :/

Tom-Ski:
Looks good, will have a play with this implementation when I upgrade to 4 for the setup tool. I have updated to the replacement gretty plugin for the main libgdx repo, but I'm not entirely happy with it. This looks like a nicer option.

MonsterOfCookie:
Yea, when I took everything apart and realised Jetty was just being used to serve static content, I just thought why not KISS it.

Tom-Ski:
I agree, we'll use this when we upgrade. Just waiting on https://youtrack.jetbrains.com/issue/IDEA-122904 ....

MrStahlfelge:
Looks like this is already used now without a merge - can be closed now.

PR #4849, opened 2017-08-09
cache cubemap sides to reduce memory consumption

mgsx-dev:
Easy to reproduce by running FrameBufferCubemapTest with visualVM.


gdx/src/com/badlogic/gdx/graphics/glutils/FrameBufferCubemap.java
```diff
@@ -64,6 +64,9 @@
 	/** the zero-based index of the active side **/
 	private int currentSide;
 
+	/** cubemap sides cache */
+	private Cubemap.CubemapSide[] cubemapSides = Cubemap.CubemapSide.values();
```
code-disaster: If you make this `private static final`, you don't even have the per-instance overhead.

mgsx-dev: Your're right, thank you.

PR #4869, opened 2017-08-26
Added Scala support for GDX Setup

snordgren:
Made it possible to generate Scala-based Gradle projects using the setup tool. Uses Scala 2.12.3, Scala Gradle plugin does not play nice with Android Gradle plugin so only Java support on that platform.


snordgren:
In order to support Android, Scala 2.11 will have to be used. Will fix shortly.

MobiDevelop:
I don't think we are going to add Scala support to GDX Setup.

PR #4871, opened 2017-08-27
Add support for 2d particle sprite animation

FredGithub:
This makes it possible to set several images for a ParticleEmitter and set the sprite mode:
- Single: old behavior, single image
- Random: a random image will be picked from the list when a particle is activated
- Animated: particles will be rendered as a sprite animation, using the images from the list, based on their lifetime

The editor has been updated accordingly. The loader is backward compatible with the previous file format.

I added a demo particle effect (particle-animation.p) in the test assets folder.

Sadly the diff has gotten way bigger than I wanted to :/


Tom-Ski:
Looks good, can you resolve the conflict?

FredGithub:
Done!

PR #4872, opened 2017-08-27
Add support for 2d particle independant scale X and Y

FredGithub:
This makes it possible to set scale X and Y independently. When scaleY is not active, scaleX is used for both X and Y.

The editor has been updated accordingly. The loader is backward compatible with the previous file format.

This PR is mergeable with my other PR "animated particles"


Tom-Ski:
Thanks!

PR #4878, opened 2017-08-31
added RGB/HSV color conversion

mgsx-dev:
All is in the title.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -489,6 +489,90 @@ public class Color {
 		color.r = ((c & 0x000000ff)) / 255f;
 	}
 
+	/** Sets the RGB Color components using the specified Hue-Saturation-Value. Note that HSV components are voluntary not clamped
+	 * to preserve high range color and can range beyond typical values.
+	 * @param h The Hue in degree from 0 to 360
+	 * @param s The Saturation from 0 to 1
+	 * @param v The Value (brightness) from 0 to 1
+	 * @return The modified Color for chaining. */
+	public Color fromHsv (float h, float s, float v) {
+		float x = (h / 60f + 6) % 6;
+		int i = (int)x;
+		float f = x - i;
+		float p = v * (1 - s);
+		float q = v * (1 - s * f);
+		float t = v * (1 - s * (1 - f));
+		switch (i) {
+		case 0:
+			r = v;
+			g = t;
+			b = p;
+			break;
+		case 1:
+			r = q;
+			g = v;
+			b = p;
+			break;
+		case 2:
+			r = p;
+			g = v;
+			b = t;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -489,6 +489,94 @@ public static void abgr8888ToColor (Color color, float value) {
...
 	}
 
+	/** Sets the RGB Color components using the specified Hue-Saturation-Value. Note that HSV components are voluntary not clamped
+	 * to preserve high range color and can range beyond typical values.
+	 * @param color The Color to be modified.
+	 * @param h The Hue in degree typically from 0 to 360
+	 * @param s The Saturation typically from 0 to 1
+	 * @param v The Value (brightness) typically from 0 to 1
+	 * @return The modified Color for chaining. */
+	public static Color hsvToColor (Color color, float h, float s, float v) {
```
ttencate: Drive-by code review! If it requires a Color to be passed in, why is this static? Why not

    public Color setFromHsv(float h, float s, float v) {
      ...
      return this; // for chaining
    }

Or alternatively:

    public static Color fromHsv(float h, float s, float v) {
      Color color = new Color();
      ...
      return color;
    }

mgsx-dev: Indeed I made that choice to keep consistency of Color class : most of member methods clamp values where static method doesn't.
But I read the class javadoc again and clamping is expected for all color methods (for comparison purpose), so my "high range color" use case may not be valid after all.
I'll probably replace with your proposal and implements clamping. I wait for more comments and core team validation before rework and PR again.
Thanks for the review!

ttencate: For what it's worth: I once needed an unclamped `Color` and just created a new class `UnclampedColor extends Color` and overriding `clamp()` to do nothing. I guess, with none of the methods nor the class itself being `final`, that this is a supported use case, albeit not very explicitly.

Tom-Ski:
I'd remove the chaining from the static method as it doesn't follow the convention for the rest of the class. (I don't have a problem with chaining as long as the rest of the class also does this)

On clamping, I would create two methods, `hsvToColor` should return the clamped values, and perhaps add an additional method `hsvToColorUnclamped`. All methods are expected to return a clamped value, so having it in the method name for one specific method that violates this will avoid confusion.

mgsx-dev:
As discussed in the first review, I decided to forget about the "high range" feature since it breaks Color comparison (equals and hashcode). Color is properly clamped now.

Methods are not static anymore but I kept chaining since most of color method do chaining.

I hope it's OK for you.

hthomison1078:
https://github.com/libgdx/libgdx/pull/4878.patch
https://github.com/libgdx/libgdx/pull/4878.diff

PR #4882, opened 2017-09-04
GLFrameBuffer refactor

Tom-Ski:
Based around comments in this PR https://github.com/libgdx/libgdx/pull/3735

GLFrameBuffer exposes builders for customization of Frame Buffer Objects without the need to build a new class and manually call the gl functions. This includes MRT support, stencil, depth, and floating point textures.

Existing FrameBuffer classes use the builders internally to create simple configurations as they did before, however they have been moved to static methods rather than created via constructors.

---
Example can be seen in MultipleRenderTargetTest for a use case of a custom FrameBuffer object

```java
GLFrameBuffer.FrameBufferBuilder frameBufferBuilder = new GLFrameBuffer.FrameBufferBuilder(Gdx.graphics.getWidth(), Gdx.graphics.getHeight());
frameBufferBuilder.addColorTextureAttachment(GL30.GL_RGBA8, GL30.GL_RGBA, GL30.GL_UNSIGNED_BYTE);
frameBufferBuilder.addColorTextureAttachment(GL30.GL_RGB8, GL30.GL_RGB, GL30.GL_UNSIGNED_BYTE);
frameBufferBuilder.addColorTextureAttachment(GL30.GL_RGB8, GL30.GL_RGB, GL30.GL_UNSIGNED_BYTE);
frameBufferBuilder.addDepthTextureAttachment(GL30.GL_DEPTH_COMPONENT32F, GL30.GL_FLOAT);
FrameBuffer frameBuffer = frameBufferBuilder.build();
```

---
This also changes FloatFrameBuffer to allow custom formats, and to avoid creating a big float buffer cpu side when not required.

Ill write documentation after some comments, (probably)


tommyettinger:
You need to give this idea another year to form into a solid API, @Tom-Ski . The other PR was a mere 20 months ago!

This probably is useful though. If this one is merged, would the old PR be closed?

Tom-Ski:
Yes

cypherdare:
@Tom-Ski 
What's the reason for removing the public constructors and using factory methods? This makes it quite convoluted to do something simple like subclassing FrameBuffer.

Tom-Ski:
Creating something that is easy for people to create any combination of complex fbos at the same time as not exposing the complicated and delicate parts of creating complex fbos is important. 

It is not too convoluted to extend, you would just subclass FrameBufferBuilder instead of inline. See: https://gist.github.com/Tom-Ski/b513ef545342a850551e13f0e608e505 

Feel free to suggest a different design instead.

xGnoSiSx:
Do the existing GL functions still stay in place accessible or will I have to overhaul my in-house engine addons with 1.9.7 fro framebuffers? - if you plan to release this in 1.9.7

cypherdare:
@Tom-Ski 
I understand the reason for the GLFrameBuffer using a builder, and that refactor is awesome. I was asking specifically about its subclasses.

Your example is not bad for a subclass that doesn't have any configuration added. But if someone were adding any public methods to extend the functionality of FrameBuffer, they couldn't take the anonymous class shortcut but would instead need to fully subclass both the FrameBuffer and the Builder. And they would have to read carefully the source code of GLFrameBuffer (not just the docs) to understand how to do it.

I'm suggesting keeping GLFrameBuffer as is, but restoring the original FrameBuffer constructors. The new static factory methods don't add any convenience since they hide the builder under the hood, and they remove the ability to easily subclass FrameBuffer. This could be done by adding a package-private empty constructor to GLFrameBuffer, and having FrameBuffer configure its builder before setting `bufferBuilder` and calling `build()`.

Just a suggestion. I don't know if it's very common for anyone else to have subclassed FrameBuffer, and in my case I didn't add much to it. I just think that keeping classes easily extendible is a good practice for avoiding library bloat. Many minor or highly specific parameters and methods that don't really need to be in `core` can be left out if we make it very easy to configure the available classes. That's how most of the scene2d.ui widgets are designed, and it's made it very easy for me to customize everything in my very UI-heavy app I've been working on the past couple years.

Tom-Ski:
The convenience is that you dont need to use the builder to obtain the 'simple' verisons of the buffers.

Could you provide an example of what it is you are extending? 

I agree in general in terms of flexibility, it was done to avoid bloat in the first place by delegating as much logic as possible to the core object GLFrameBuffer. I will have a look and see what would be good fit to help flexibility.

cypherdare:
I mean convenience over the original constructor that uses the same arguments as the static factory method. What I'm suggesting is adding this to GLFrameBuffer:

```
GLFrameBuffer(){ // package-private
}
```

and replacing this block in FrameBuffer:

    public static FrameBuffer createFrameBuffer (Pixmap.Format format, int width, int height, boolean hasDepth, boolean hasStencil) {
        //...
    }

with

```
public FrameBuffer (Pixmap.Format format, int width, int height, boolean hasDepth, boolean hasStencil) {
    super();
    bufferBuilder = //...
    //...
    build();
    addManagedFrameBuffer(Gdx.app, this);
}
```

Now the 'simple' versions of the buffers can be instantiated *and* subclassed without a care for the builder or any of the details of how that works. Doubly convenient.

It's been a long time since I actually subclassed FrameBuffer in a project. I think I was making it into a post-processing manager, so I had added some parameters for controlling the effect. Maybe this is poor design and my manager should simply have referenced a FrameBuffer. More-experienced me is thinking it is. 

So feel free to ignore my comments, because I'm hard pressed to think of a *good* need for adding functionality directly to FrameBuffer. If I had noticed this change before the 1.9.7 release, the much bigger reason for my suggestion would be to avoid breaking everyone's existing code.

Tom-Ski:
Breaking changes are required to wake people up after 6 months on the same version. :]

I can't think of a good reason either, that wouldn't be better suited to composition or alternative. But I think the effort should be made to make it extensible in a nice way if there is some case.  What you suggested would probably work, Just adds a little bit of code duplication, but thats Ok.

Tom-Ski:
I've made changes as discussed here: https://github.com/libgdx/libgdx/commit/c5fed1006b816fddf899b9e16826a4578a373114

@code-disaster this also includes the customization of render buffers.

jrdzha:
@Tom-Ski Interesting change, but now my code is breaking on 1.9.7 using GL20. I've tried replacing parts using the FrameBuffer with FrameBufferBuilder, but I'm getting an error saying that GL20 doesn't support multiple render targets. Is there a way around this without updating to GL30? Thanks

Tom-Ski:
You'll have to show example code that does this. You should use 1.9.8-SNAPSHOT to do so.

Issue #4894, opened 2017-09-23
3d particle TangentialAcceleration axis doesn't rotate when ParticleController is rotated


ryanastout:
[MyGdxGame.java.txt](https://github.com/libgdx/libgdx/files/1326660/MyGdxGame.java.txt)
Please ensure you have given all the following requested information in your report.

#### Issue details
I have a 3d ParticleController which makes use of a TangentialAcceleration dynamics modifier. When the ParticleEffect is at the origin, the effect gives the appearance of a tornado where the axis of the tornado is the z-axis.
The TangentialAcceleration is not set to global.
If the ParticleController is translated, the tangential acceleration is applied to the particles as if the origin is still the center of the particle effect. If the particle effect is rotated such that the axis of the tornado is no longer (parallel to) the y-axis, the tangential acceleration is still applied using the y-axis. 
The desired effect is that translating or rotating the ParticleController should result in the tornado still remaining as a tornado, but that its center and orientation would change. 
#### Reproduction steps/code
I've attached a file (remove the .txt extension I was forced to add for the upload and it's a valid java file) which demonstrates the issue. At lines 91 and 92, if there is no rotation or translation (comment out lines 91 and 92), then the particles look somewhat like a tornado which has its center axis aligned with the z-axis (which is where the camera is). However, if the particle controller is translated and/or rotated, the tangential acceleration is still applied. Note that isGlobal is set to false, so the expected behavior is that the acceleration would be applied with respect to the center of the particle controller, not with respect to point (0,0,0).

Is this behavior intentional? If so, why does it make sense that the tangential acceleration only appears as it does in the Flame editor when the controller is unrotated and at the origin?

If not, would a reasonable fix be to store the center of the particle controller and the rotation, and apply the rotation to TMP_V3 in TangentialAcceleration.upate() and to subtract the center of the particle controller from the positionChannel values before performing the cross product?

#### Version of LibGDX and/or relevant dependencies
LibGDX version 1.9.6. I also pulled the code from git on 09/23/2017 and the issue is present there as well. 

#### Stacktrace
There's no stacktrace. The code runs fine.

#### Please select the affected platforms
I only tested on Linux, but I would be surprised if this was platform specific

https://stackoverflow.com/questions/29168257/libgdx-how-to-rotate-the-emmition-trail-of-a-3d-particle seems to be a similar issue

ryanastout:
[...]

PR #4896, opened 2017-09-23
Make TangentialAcceleration and PolarAcceleration respect the global setting

ryanastout:
Before this change, if a TangentialAcceleration or PolarAcceleration were used, the forces they'd apply particles didn't account for the location of the ParticleController. So if the ParticleController was moved or rotated, the TangentialAcceleration and PolarAcceleration both still acted as if the ParticleController hadn't been translated or rotated. 
After this change, if isGlobal is false, the direction of the PolarAcceleration's force is rotated when the ParticleController is and the axis about which TangentialAcceleration is applied is rotated. The particle to which the force is being applied is treated with its position relative to the ParticleController rather than relative to the origin. 
Tangential and Polar acceleration didn't respect the global/relative setting in Flame. this fixes https://github.com/libgdx/libgdx/issues/4894

Additionally, when using a controller of controllers, it is necessary to pass on the rotation information.

In order to test, I have added three tests: TangentialAccelerationTest.java, PolarAccelerationTest.java and ParticleControllerInfluencerSingleTest.java.
 The behavior before this change can be seen if isGlobal is set to true in either of the first two cases. In the final case, the correctly rotated ParticleController is shown. All 3 tests were based on ParticleControllerTest.java.

### DIFF
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSDevice.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSDevice.java
@@ -40,7 +40,13 @@ public enum IOSDevice {
 	IPHONE_7("iPhone9,3", 326),
 	IPHONE_7_PLUS("iPhone9,4", 401),
 	IPHONE_SE("iPhone8,4", 326),
-
+	IPHONE_8_CDMA_GSM("iPhone10,1", 326),
+	IPHONE_8_PLUS_CDMA_GSM("iPhone10,2",401),
+	IPHONE_X_CDMA_GSM("iPhone10,3", 458),
+	IPHONE_8("iPhone10,4", 326),
+    	IPHONE_8_PLUS("iPhone10,5", 401),
+	IPHONE_X("iPhone10,6", 458),
+	
 	IPOD_TOUCH_1G("iPod1,1", 163),
 	IPOD_TOUCH_2G("iPod2,1", 163),
 	IPOD_TOUCH_3G("iPod3,1", 163),
@@ -80,6 +86,12 @@ public enum IOSDevice {
 	IPAD_PRO("iPad6,8", 264),
 	IPAD_PRO_97_WIFI("iPad6,3", 264),
 	IPAD_PRO_97("iPad6,4", 264),
+	IPAD_5_WIFI("iPad6,11", 264),
+	IPAD_5_WIFI_CELLULAR("iPad6,12", 264),
+	IPAD_PRO_2_WIFI("iPad7,1", 264),
+	IPAD_PRO_2_WIFI_CELLULAR("iPad7,2", 264),
+	IPAD_PRO_10_5_WIFI("iPad7,3", 264),
+	IPAD_PRO_10_5_WIFI_CELLULAR("iPad7,4", 264),
 
 	SIMULATOR_32("i386", 264),
 	SIMULATOR_64("x86_64", 264);
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSDevice.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSDevice.java
@@ -40,7 +40,13 @@ public enum IOSDevice {
 	IPHONE_7("iPhone9,3", 326),
 	IPHONE_7_PLUS("iPhone9,4", 401),
 	IPHONE_SE("iPhone8,4", 326),
-
+	IPHONE_8_CDMA_GSM("iPhone10,1", 326),
+	IPHONE_8_PLUS_CDMA_GSM("iPhone10,2",401),
+	IPHONE_X_CDMA_GSM("iPhone10,3", 458),
+	IPHONE_8("iPhone10,4", 326),
+    	IPHONE_8_PLUS("iPhone10,5", 401),
+	IPHONE_X("iPhone10,6", 458),
+	
 	IPOD_TOUCH_1G("iPod1,1", 163),
 	IPOD_TOUCH_2G("iPod2,1", 163),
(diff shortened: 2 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/particles/influencers/DynamicsModifier.java
```diff
@@ -415,11 +421,17 @@ public void update () {
 
 				float cosTheta = MathUtils.cosDeg(theta), sinTheta = MathUtils.sinDeg(theta), cosPhi = MathUtils.cosDeg(phi), sinPhi = MathUtils
 					.sinDeg(phi);
-				TMP_V3
-					.set(cosTheta * sinPhi, cosPhi, sinTheta * sinPhi)
-					.crs(positionChannel.data[positionOffset + ParticleChannels.XOffset],
-						positionChannel.data[positionOffset + ParticleChannels.YOffset],
-						positionChannel.data[positionOffset + ParticleChannels.ZOffset]).nor().scl(strength);
+				TMP_V3.set(cosTheta * sinPhi, cosPhi, sinTheta * sinPhi);
+				TMP_V1.set(positionChannel.data[positionOffset + ParticleChannels.XOffset],
```
ryanastout: TMP_V1 now contains the (global) position of the particle to which we're applying the force. TMP_V3 is the vector representing the axis of rotation for.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/influencers/DynamicsModifier.java
```diff
@@ -415,11 +421,17 @@ public void update () {
...
+					positionChannel.data[positionOffset + ParticleChannels.YOffset],
+					positionChannel.data[positionOffset + ParticleChannels.ZOffset]);
+				if (!isGlobal) {
```
ryanastout: if isGlobal == false, then we need to get the position of the particle relative to the center rather than the origin, so we subtract the controller's position from TMP_V1.
TMP_V3 needs to be rotated if the controller was so that the axis about which we apply our force is relative to the controller rather than to the origin.

gdx/src/com/badlogic/gdx/graphics/g3d/particles/influencers/SpawnInfluencer.java
```diff
@@ -68,6 +70,13 @@ public void activateParticles (int startIndex, int count) {
 			positionChannel.data[i + ParticleChannels.YOffset] = TMP_V1.y;
 			positionChannel.data[i + ParticleChannels.ZOffset] = TMP_V1.z;
 		}
+		for (int i = startIndex * rotationChannel.strideSize, c = i + count * rotationChannel.strideSize; i < c; i += rotationChannel.strideSize) {
+			controller.transform.getRotation(TMP_Q, true);
```
ryanastout: When we have a controller of controllers, if the parent controller was rotated, we need to communicate the rotation to the children when they're spawned.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/ParticleControllerInfluencerSingleTest.java
```diff
@@ -0,0 +1,197 @@
+
```
ryanastout: This file demonstrates the correct behavior when the rotationChannel is properly updated and exhibits incorrect behavior without that change. 
In particular, the ParticleController is correctly rotated when the SpawnInfluencer properly accounts for rotation.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/PolarAccelerationTest.java
```diff
@@ -0,0 +1,180 @@
...
+		polarAcceleration.thetaValue.setHigh(90);
+		polarAcceleration.phiValue.setActive(true);
+		polarAcceleration.phiValue.setTimeline(new float[] {0});
+		polarAcceleration.phiValue.setScaling(new float[] {1});
+		polarAcceleration.phiValue.setHigh(0);
+		polarAcceleration.strengthValue.setActive(true);
+		polarAcceleration.strengthValue.setHigh(10);
+		polarAcceleration.strengthValue.setTimeline(new float[] {0});
+		polarAcceleration.strengthValue.setScaling(new float[] {1});
+		polarAcceleration.isGlobal = false;
```
ryanastout: Running the test in this file demonstrates the correct behavior when the PolarAcceleration's direction is properly updated and exhibits incorrect behavior without that change. If isGlobal is set to true here, the old behavior is retained, so changing this to true and running the test demonstrates the old behavior that's being corrected.

tests/gdx-tests/src/com/badlogic/gdx/tests/g3d/TangentialAccelerationTest.java
```diff
@@ -0,0 +1,180 @@
...
+		tangentialAcceleration.thetaValue.setHigh(90);
+		tangentialAcceleration.phiValue.setActive(true);
+		tangentialAcceleration.phiValue.setTimeline(new float[] {0});
+		tangentialAcceleration.phiValue.setScaling(new float[] {1});
+		tangentialAcceleration.phiValue.setHigh(0);
+		tangentialAcceleration.strengthValue.setActive(true);
+		tangentialAcceleration.strengthValue.setHigh(10);
+		tangentialAcceleration.strengthValue.setTimeline(new float[] {0});
+		tangentialAcceleration.strengthValue.setScaling(new float[] {1});
+		tangentialAcceleration.isGlobal = false;
```
ryanastout: Running the test in this file demonstrates the correct behavior when the TangentialAcceleration's direction and position is properly updated and exhibits incorrect behavior without that change. If isGlobal is set to true here, the old behavior is retained, so changing this to true and running the test demonstrates the old behavior that's being corrected.

Issue #4899, opened 2017-09-25
Gdx setup for Robovm, MOE backend should use Asset Catalog for App Icon


keesvandieren:
#### Issue details

Details discussed in https://github.com/MobiVM/robovm/issues/210

The Robovm app generated by the GDX setup utility (and probably also the MOE variant) use CFBundleIcons in Info.plist.xml to define app-icons available.

Since the Xcode 9 release (and related App Store upgrades) this is no longer supported; asset catalog AppIocn needs to be used.

#### Reproduction steps/code

* Generate a new libgdx project with ios (Robovm backend)
* Register app id in iTunes Connect, Apple Member Center.
* Submit the app.

It will fail with the following error:

    Missing Info.plist value - A value for the "Info.plist" key CFBundleIconName is missing in the 
     bundle 'com.rosapp.choinkafull'. Apps that provide icons in the asset catalog must also provide 
    this Info.plist key. For more information see 
    http://help.apple.com/xcode/mac/current/#/dev10510b1f7.

#### Version of LibGDX and/or relevant dependencies
All versions (Xcode / App Store change)

#### Please select the affected platforms
- [ ] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

keesvandieren:
If anyone points me to a  Libgdx logo in SVG format (Inkscape compatible) I can easily generate the logos in all required formats and attach them to this issue

piotr-j:
This works? https://github.com/libgdx/libgdx/blob/master/libgdx_logo.svg

piotr-j:
I dont think there is a square one. You can rotate this one i guess https://github.com/libgdx/libgdx/blob/master/libgdx_logo.svg

keesvandieren:
Pull request for Robovm has been accepted. MOE still needs to be done.

PR #4902, opened 2017-09-26
Use App Catalog for ios (Robovm) app icon

keesvandieren:
Use App Catalog for Robovm app icon.

Partial fix for #4899 (moe still needs to be done).

Icon is made with attached Icon.svg.(.txt) Not yet added that in Pull Request, not sure where it should be located.

[Icon.svg.txt](https://github.com/libgdx/libgdx/files/1334606/Icon.svg.txt)


Tom-Ski:
Looks great, thanks @keesvandieren.

piotr-j:
Is that partial transparent border in the icons intended? Looks a bit weird.

keesvandieren:
Thanks for testing.

No, its not intended. At first whole icon was transparant. Added a white square behind it, but probably smaller than the document. Will fix it.

PR #4906, opened 2017-09-29
Make desktop input dialog modal

fercarcedo:
Make desktop input dialog for Gdx.input.getTextInput modal and set it to be always on top, because currently when executing getTextInput on the desktop sometimes the input dialog gets under the game window, so it's invisible.


NathanSweet:
Are those spaces for indentation?! :o

fercarcedo:
Switched to tabs  :)  (and also squashed commits into one)

NathanSweet:
Thanks. FWIW, no need to squash. The diff is easy to see even with multiple commits and GitHub can squash when a PR is accepted.

PR #4907, opened 2017-09-29
Show keyboard automatically on Android getTextInput and move text selection to the end

fercarcedo:
Make Gdx.input.getTextInput show keyboard automatically on Android as soon as the dialog appears, and the EditText cursor position is moved to the end, so that the user can start to write instantly, instead of having to first tap on the EditText to show the keyboard and then tap on the end of the text to write after the current text.


intrigus:
Hi, could you please only format the parts that you changed?

fercarcedo:
Fixed. It seems there was some problem with my editor the first time

Leejjon:
Yes please. I also have build this in myself by extending AndroidInput. With this change that is no longer needed.

MobiDevelop:
After nearly 3 years of no activity, I'm going to close this out. If for some reason this is still desired, a new PR can be opened.

PR #4910, opened 2017-10-01
2d particle effect rotation and better scaling

FredGithub:
Particle emitters can now be rotated. To achieve this, I made particles use a local transform (position, scale, rotation). The sprite's transform is then set to the world transform of the particle, using the transform of the parent (the emitter).  Here is a [small demo](https://imgur.com/a/z9bL0).

The editor has been updated to allow for rotation and scaling of the effect.

I originally planned to add rotation only, but since I was using a `Matrix3` for the parent transform, adding scale was easy. In fact, I believe this way of scaling is much simpler and works better than the current `scaleEffect` API which scales every parameter related to position/movement, and always felt more like a workaround.

I know that this is a significant change, so I would understand if it can't be merged. 

If it can however, please let me know if I should also push [that other commit](https://github.com/FredGithub/libgdx/commit/bc8a26fd76f1ed852349b041940c3b732782ab44), which reduces code complexity quite a bit by removing everything related to the current `scaleEffect`. Removing various public APIs (and the work of other people) is not the best though, but having 2 APIs for scaling is also not ideal. Not sure what to do at that point.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/particleeditor/EffectPanel.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/particleeditor/EffectPanel.java
@@ -34,6 +34,7 @@ import javax.swing.table.DefaultTableModel;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.g2d.ParticleEffect;
 import com.badlogic.gdx.graphics.g2d.ParticleEmitter;
+import com.badlogic.gdx.tools.particleeditor.ParticleEditor.TransformMode;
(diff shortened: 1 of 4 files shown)


obigu:
This is really cool, I totally missed it. It's been a while but if you are still active I'd happily test it if the conflicts are resolved.

FredGithub:
I'm away from home for a week or so, but I'll try to solve the conflicts when I come back.

MrStahlfelge:
Closing this, although this looks like a good enhancement. @obigu feel free to reopen a new PR without conflicts if you need the change.

PR #4914, opened 2017-10-02
make TangentialAcceleration and PolarAcceleration respect global flag

ryanastout:
Tangential and Polar acceleration didn't respect the global/relative setting in Flame. this fixes https://github.com/libgdx/libgdx/issues/4894

Also fixed a bug in which rotation information wasn't passed to from SpawnInfluencer to ParticleControllers


ryanastout:
@Tom-Ski Is there anything I can do to make this pr more easy to review? The actual bug-fixes are about 20 lines, and the majority of the lines changed here are in 3 new test files to demonstrate that the fix works as intended.

PR #4916, opened 2017-10-05
Updated LWJGL3 backend to 3.1.3

Zomby2D:
LWJGL 3.1.3 has been released and now includes the move to evdev for controllers input under Linux.

This should allow the changes in #4744 to work with Linux as well.


tests/gdx-tests/src/com/badlogic/gdx/tests/FilesTest.java
```diff
@@ -208,11 +208,7 @@ private void testInternal () throws IOException {
 			fail();
 		} catch (Exception ignored) {
 		}
-		FileHandle dir;
-		if (Gdx.app.getType() == ApplicationType.Android || Gdx.app.getType() == ApplicationType.iOS)
-			dir = Gdx.files.internal("data");
-		else
-			dir = Gdx.files.internal("../gdx-tests-android/assets/data");
+		FileHandle dir = Gdx.files.internal("data");
```
code-disaster: Is this an intended change? Looks unrelated to the topic.

Zomby2D: That the test was broken and I ended up fixing it while checking if the update didn't introduce any new issue. Should I take this out?

Tom-Ski: Its preferred if you split up unrelated issues/pull requests. Or at very least if they are tiny changes document what you have changed and why.

Tom-Ski: We'll make an exception just for you this time  ;]

PR #4922, opened 2017-10-11
Made Cmotionmanager protected so that users can customize motion controls.

Leejjon:
This PR is a follow up from #4886 

Right now I have to init a separate CMotionManager to slightly alter motion controls. With this change that is no longer needed. I can have my IOSApplication use a subclass of IOSInput that override either the setupAccelerometer()/setupMagnetometer() methods or do even more custom stuff.

My forum post describing the problem:
http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=26824&p=105829#p105753


Leejjon:
Making new PR from new branch. Did this change in too many commits and didn't want to ruin the commit history of libgdx with my trial and errors.

Issue #4924, opened 2017-10-12
Tooltips broken on Android


Dakror:
#### Issue details
Tooltips broken on Android. They do not show but cause an exception instead.
Also referenced in #4921 

#### Reproduction steps/code
Run the UITest on android (i used a Huawei P8, Android 5)

#### Version of LibGDX and/or relevant dependencies
1.9.6

#### Stacktrace
```java
10-12 18:23:17.610 571-761/com.badlogic.gdx.tests.android W/System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'com.badlogic.gdx.scenes.scene2d.Stage com.badlogic.gdx.scenes.scene2d.Actor.getStage()' on a null object reference
10-12 18:23:17.610 571-761/com.badlogic.gdx.tests.android W/System.err:     at com.badlogic.gdx.scenes.scene2d.ui.TooltipManager$2.run(TooltipManager.java:69)
10-12 18:23:17.610 571-761/com.badlogic.gdx.tests.android W/System.err:     at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame(AndroidGraphics.java:488)
10-12 18:23:17.610 571-761/com.badlogic.gdx.tests.android W/System.err:     at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1531)
10-12 18:23:17.610 571-761/com.badlogic.gdx.tests.android W/System.err:     at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1248)
```

#### Please select the affected platforms
- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Tom-Ski:
I can't reproduce this on my device.  You are running the stock UITest?

Dakror:
I'm running the stock UITest and when tapping on the Button "Multi Line Toggle" there should be a tooltip popping up, but instead an exception gets thrown.

Tom-Ski:
Yeah, I can't reproduce this with your steps.  I'll try find another device to test on to see if i can it there. Otherwise, hopefully someone else can reproduce and test.

Zomby2D:
I tried various phones and tablets with Android versions ranging from 4.1.2 to 7.1, and I can't reproduce the issue either. The tooltip always pops up correctly for me. (I tried with both USB and Bluetooth mices)

[...]

PR #4925, opened 2017-10-12
Fix for tooltip not working on android

Dakror:
Fix for issue #4924

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Tooltip.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Tooltip.java
@@ -83,6 +83,7 @@ public class Tooltip<T extends Actor> extends InputListener {
 			container.toFront();
 			return false;
 		}
+		setContainerPosition(event.getListenerActor(), x, y);
 		manager.touchDown(this);
 		return false;
 	}


Tom-Ski:
You are using a USB mouse?

Dakror:
Yes I am, why does this matter though? :o

Tom-Ski:
For reproduction steps. Will have to order an adapter.

Dakror:
It's reproducible on an actual android phone, if you've got one of those. Not sure about emulator though.

Tom-Ski:
Sure, I just don't have a micro USB adapter to actually use a mouse, so I ordered one.

Zomby2D:
I can't reproduce the initial issue, but this fix actually causes some weird behavior if you press the multi-line button in the UITest without a mouse plugged in. The multi-line tooltip text is floating and it's background is not sized/positioned to match the the text.

Dakror:
The fix leads to the correct behaviour on both my desktop and my android phone... Strange

MrStahlfelge:
Closing this. Fixing an issue on android only in scene2d core doesn't make sense to me at short glance, and here is no explanation what's happening and why this should fix it. Additionaly, the PR is super old and no feedback from the community on importance given.
Feel free to reopen for some enlightenment.

PR #4937, opened 2017-10-21
Added support for distance field font and mipmap for all fonts to the skin json files

ProZhar:
Fixes http://badlogicgames.com/forum/viewtopic.php?f=11&t=12900

1. Added support to generate a bitmap font with mipmap
2. Added default support for Skin.class to generate distance field font
3. Added test BitmapFontDistanceFieldFromSkinTest that shows how it works


ProZhar:
That's all.

ProZhar:
@Tom-Ski @NathanSweet is this request interesting? Because I have some fixes for implementing stages shader if this request will be merged.

Tom-Ski:
Hi @ProZhar, thanks for the PR!

It is potentially interesting. I wouldn't ever recommend using this due to the performance implications, but if people understand that, and its not a problem for them (for example making a desktop only application with low intensity), then I can understand it would be useful.

I'll wait for more opinions before reviewing.

As an aside, please follow https://libgdx.badlogicgames.com/documentation/hacking/Contributing.html when contributing.

ProZhar:
Hi @Tom-Ski, thanks for the answer.

I understand that the use of mipmap has implications for performance. I only added this because mipmap is present in the BitmapFontDistanceFieldTest.java.
But the distance field font I use in all my games on android and along with the ExtendViewport it looks very good.

I understood the remarks about contributing :) 
So please remove this fork and I will split it and create one that solves single problems

Tom-Ski:
Its not the mip maps, but the nesting of distance field fonts in scene2d.

NathanSweet:
I'm not sure we need Skin to know about distance field fonts. Consider adding the font before loading the skin JSON:
```
Skin skin = new Skin();
skin.add("fontName", distanceFieldFont, BitmapFont.class);
skin.load(skinFile);
```
FWIW, lately I still use Skin of course, but most styles don't come from the JSON, as I find it a bit tedious. Instead I create a few utility methods. For example, to create an image text button with the given icon prefix, ,text, and button style. It starts by cloning the button style from the skin, then looks in the skin for all the icon drawables, eg `prefix + "-up"`, `prefix + "-over"`, `prefix + "-disabled"`, etc.

ProZhar:
@Tom-Ski @NathanSweet, thanks for the answer.

At the moment, I'll use it that way(without mipmaps), because for me it's convenient to use a skin.json and asset menager.

My steps:
1. Extension of the Skin class
https://gist.github.com/ProZhar/b5017123df4fd3eb967048994fbf33ec

2. Creation of the Skin2Loader class
https://gist.github.com/ProZhar/7c6cd0daf3535827b441558fb4ee4e63

3. Creation of the Assets class
https://gist.github.com/ProZhar/e88b2c97ae6d2c47993ee3d9ede4764c

4. after that I use
```java
skin = assets.get(Assets.SKIN);
```

5. In skin.json the distance field font using
```java
com.badlogic.gdx.graphics.g2d.BitmapFont: {
        default-font: {
            file: fonts/Roboto_DF_10sp_4spread.fnt,
            distanceFieldSmoothing: 2
        }
},
```
If is set distanceFieldSmoothing the skin generate DF-font, otherwise bitmap font. One font 30px size for me equals 10sp (material design like). Then I use only the scale for Labels (1,4 = 14sp, 1,6 = 16sp).

P.S. When i wrote PR, I found some bug for my projects. The Label uses own font scale, and the DF Font class in its method getSmoothingFactor () before draw() uses fonts scale. That's why my fonts are not perfect that I want.

This app is written using only one distance field font and this method.
https://play.google.com/store/apps/details?id=com.prozhar.ua.spelling&hl=en

PR #4945, opened 2017-10-25
Circular Motion dynamics modifier for 3d particles

ryanastout:
created a new DynamicsModifier and added it to the Flame tool and the GdxTests. The CircularMotion influencer adds a velocity to a particle in the activateParticles method and applies a force toward an axis specified by theta and phi, which is the axis of rotation for uniform circular motion.


ryanastout:
A couple of people have asked about Circular Motion for 3d particles in the irc channel, so this is probably useful to people.

PR #4950, opened 2017-10-26
Fix flush() to force synchronous write to disk

hevrard:
editor.apply() starts an asynchronous request to write to disk, but does not guarantee
the write to be completed when it returns:
https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()

On the other hand, editor.commit() does require a synchronous write, and thus is a better
candidate to implement flush():
https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#commit()

This fix comes from a real issue on a multithreaded app where calling flush() just before killing the
whole process does not actually write to disk before the process is killed.

### DIFF
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
@@ -195,6 +195,6 @@ public class IOSPreferences implements Preferences {
 				}
 			}
 		});
-		return success[0];.
+		return success[0];
 	}
 }
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
@@ -185,17 +185,16 @@ public class IOSPreferences implements Preferences {
 
 	@Override
 	public boolean flush () {
-		boolean success = false;
-		// FIXME: how to hoist the success or failure of write to return it?
+		final boolean[] success = {false};
 		ObjCRuntime.autoreleasepool(new Runnable() {
 			@Override
 			public void run() {
 				if (!nsDictionary.writeToFileAtomically(file.getAbsolutePath(), false)) {
 					Gdx.app.debug("IOSPreferences", "Failed to write NSDictionary to file " + file);
+					success[0] = true;
 				}
 			}
 		});
-		success = true; // FIXME: should reflect the success of NSDictionary write.
-		return success;
+		return success[0];.
 	}
 }
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
@@ -184,7 +184,9 @@ public class IOSPreferences implements Preferences {
 	}
 
 	@Override
-	public void flush () {
+	public boolean flush () {
+		boolean success = false;
+		// FIXME: how to hoist the success or failure of write to return it?
 		ObjCRuntime.autoreleasepool(new Runnable() {
 			@Override
 			public void run() {
@@ -193,5 +195,7 @@ public class IOSPreferences implements Preferences {
 				}
 			}
 		});
+		success = true; // FIXME: should reflect the success of NSDictionary write.
+		return success;
 	}
 }
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
@@ -189,9 +189,10 @@ public class IOSPreferences implements Preferences {
 		ObjCRuntime.autoreleasepool(new Runnable() {
 			@Override
 			public void run() {
-				if (!nsDictionary.writeToFileAtomically(file.getAbsolutePath(), false)) {
-					Gdx.app.debug("IOSPreferences", "Failed to write NSDictionary to file " + file);
+				if (nsDictionary.writeToFileAtomically(file.getAbsolutePath(), false)) {
 					success[0] = true;
+				} else {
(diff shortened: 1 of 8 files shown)


MobiDevelop:
I don't think that flush should always use commit, in many cases an asynchronous write is preferred. Perhaps it would be better to add a new version of flush that does it synchronously.

hevrard:
In most languages/libraries, `flush()` does perform a synchronous write, in the sense that once it returns we can assume that data has been written to persistent storage. As Preferences is the only persistent read/write storage over all platforms supported from Libgdx, I think its `flush()` method should indeed give the program the guarantee that the write has been performed. Also, with the current implementation, I can't see how it's possible to know (using libgdx API) whether the write has actually been completed.

It's also good to have a primitive for asynchronous writes, but for the sake of consistency with the general meaning of `flush()`, I still think here `flush()` should be synchronous, on all platforms.

Tom-Ski:
On all the other backends flush is synchronous, so it makes sense to normalize this behavior. I think we should add a new method as @MobiDevelop suggested so we don't lose the asynchronous persistence feature on android.

As an aside, commit doesn't guarantee you a successful write, so we should return a boolean state for success on flush.

hevrard:
On all backends, flush() now returns a boolean to indicate whether it has succeeded or failed.
This boolean is meaningless on the "moe" backend: I can't see how we can check the success of write for this backend. Maybe reading the log to grep the error string, but that sounds fragile? Could you guys have a look at this one?

backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
```diff
@@ -193,5 +195,7 @@ public void run() {
 				}
 			}
 		});
+		success = true; // FIXME: should reflect the success of NSDictionary write.
```
Darkyenus: You could change line 188 to `final boolean[] success = {false};`, success set on line 194 with `success[0] = true;` and `return success[0];`. This pattern is often used in similar situations. It leads to an allocation, but if the MOE is smart enough, it could optimize it out. And if not - flush is a slow operation anyway, so it doesn't matter that much.

hevrard:
Thanks @Darkyenus for the tip, much appreciated!
So now all backends' `flush()` implementation should return false if the write has failed (or, as per the already existing code, an exception may be thrown for some backends).

Do you feel like there anything more needed for this particular PR?

backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSPreferences.java
```diff
@@ -184,14 +184,17 @@ private NSString convertKey (String key) {
...
 	@Override
-	public void flush () {
+	public boolean flush () {
+		final boolean[] success = {false};
 		ObjCRuntime.autoreleasepool(new Runnable() {
 			@Override
 			public void run() {
 				if (!nsDictionary.writeToFileAtomically(file.getAbsolutePath(), false)) {
 					Gdx.app.debug("IOSPreferences", "Failed to write NSDictionary to file " + file);
+					success[0] = true;
```
Darkyenus: Please recheck the condition here, having `success=true` and `log("..failed..")` in the same block seems wrong.

hevrard:
My bad, the logic was wrong indeed. It should be fixed now.

Tom-Ski:
Looks good. Although as @MobiDevelop said, I think it would be best to expose a new version of flush that works synchronously.

hevrard:
You mean a version of flush that works *asynchronously*? (As this PR makes flush synchronous).

As a genuine question, I wonder when a program makes the effort to call `flush()`, but does not expect it to guarantee the write. Isn't it the whole point of this primitive? If a program does not need to make sure it is written, it just does not call flush() at all.

So I honestly ask, in which situations an asynchronous `flush()` can be useful?

Darkyenus:
> in which situations an asynchronous flush() can be useful?

For example when player completes a level and the game wants to save progress (but it is not expected that the app will close soon), so the correct way to save is to do it in an async background thread, as slow IO in a main loop may drop frames.

However I think that the purpose of this PR has been fulfilled and this should be a separate one. Especially when this functionality is trivially implementable probably only on Android and other platforms will need something more complex (or delegate to a blocking version, which is not ideal).

Tom-Ski:
Currently the version of flush is asynchronous where possible (only android atm afaik).  This PR currently changes this to synchronous.  I'm suggesting making an alternate method that doesn't break existing behavior on android, but also allows for a synchronous attempt at saving.

As for your question, a few things: 

- There is no guarantee even a synchronous flush will save your data.  
- Preferences should not be used for any important data. (Yes its the only way to save on all platforms including the web, but you are much safer saving to local in backends that support it)
- Asynchronous flush is handy if you don't want to interrupt the main thread which can cause your game to stutter.

MobiDevelop:
Calling flush is required to save the changes whether synchronous or asynchronous, so one could not simply not call flush... Otherwise there is no point putting anything in preferences.

I guess my other question is are you using System.exit or something? Under normal circumstances exiting the app will not immediately kill the process on Android.

obigu:
I agree it would be useful to have both versions, synchronous and asynchronous. both for consistency and to have a way avoid conflicts/race conditions in situations where persisting several times in a short amount of time is needed.

@hevrard It would be useful to know when you are persisting the state in your Android app. In theory, you don't have any guarantees that `onStop()` or `onDestroy()` will get called when a user/system kills the app but `onPause()` should always be called.

On the other hand the documentation states that
> You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states.

So, again in theory (we all know Android :), if you persist your preferences asynchronously on onPause(), you should be safe.

hevrard:
Thanks for the meaningful comment, we eventually resorted to use plain Android files and force flush, disk sync etc to make sure contents are written to real persistent storage.

Our use-case is not a video game, but stress-testing GPU drivers: we sometimes trigger crashes that leads to a whole device restart (see e.g. http://www.graphicsfuzz.com/#security , in the video), so we really need to be able to write to real persistent storage at some points.

I understand that typical apps do not have such a need, hence I'm dropping effort on this pull request. I still find it a bit odd that a primitive called `flush()` could be asynchronous, but you have exposed reasonable arguments for this behaviour. Thanks for the great work on libgdx, I let the team decide whether a synchronous variant of `flush()` is interesting.

PR #4963, opened 2017-11-04
Add modern shaders support

mgsx-dev:
This contribution brings geometry shader and tesslation shaders support.

This is a working branch, i need to cleanup things but i'd like to have your opinion on it before : 

* Even if it's not supported by WebGL, do you think it's resonable to merge it in libgdx ?
* Does the implementation/design is OK for you ?

Also I couldn't be able to test it fully cause i don't have all required hardware myself. I just tested it on Linux desktop and GWT to ensure it doesn't break shader program. It need some test with GLES3.2 android device and other OS (MacOSX and Windows).

There is simple examples provided you can test on : 

* SimpleGeometryShaderTest
* SimpleTesslationShaderTest
* SimplePNTriangleTesslation

Compatibility notes : 

* None of these shaders are supported by WebGL (even WebGL 2.0)
* Requires GLES 3.2 for android
* Requires OpenGL 4.2 for desktop

Note that i'm working on another branch for compute shader support but it involve more changes. You could take a look as well : https://github.com/mgsx-dev/libgdx/tree/modern-shaders/compute

Thanks to point me in the right direction if i'm wrong.


mgsx-dev:
If somebody could test the 3 provided examples with an android device supporting GLES 3.2 it could help !

To know if your device support it, you can search in the following list : http://opengles.gpuinfo.org/ with filter set to GLES 3.2.

Thanks!

Tom-Ski:
Nice, will check it out.  

With regards to support, I'd say its worth having even if its desktop only. At the moment we only support core GLES 3.0, so none of this is supported there.

cypherdare:
Just looked over it briefly, and I like the approach.

I was going to suggest making ShaderStage an enum, but after a closer look I saw that you can pass any arbitrary ShaderStage instance into a ShaderPart, so you are not locked into a fixed global configuration. So in that case, I think some Javadocs in ShaderStage should explain that those are provided for convenience and for default ShaderProgramLoader behavior. And it might be nice to mention in the ShaderPart constructor docs that some convenient preset ShaderStages are available.

mgsx-dev:
I bring back this PR with a full rebase and added documentation as suggested by @cypherdare 

This PR will provide a cleaner way to implement transform feedback buffers (see #5120) since it's typically using a vertex shader only program (and optional geometry shader).

gdx/src/com/badlogic/gdx/assets/loaders/ShaderProgramLoader.java
```diff
@@ -20,88 +20,117 @@
...
 import com.badlogic.gdx.utils.GdxRuntimeException;
 
-/** {@link AssetLoader} for {@link ShaderProgram} instances loaded from text files. If the file suffix is ".vert", it is assumed
- * to be a vertex shader, and a fragment shader is found using the same file name with a ".frag" suffix. And vice versa if the
- * file suffix is ".frag". These default suffixes can be changed in the ShaderProgramLoader constructor.
- * <p>
- * For all other file suffixes, the same file is used for both (and therefore should internally distinguish between the programs
- * using preprocessor directives and {@link ShaderProgram#prependVertexCode} and {@link ShaderProgram#prependFragmentCode}).
+/** {@link AssetLoader} for {@link ShaderProgram} instances loaded from text files. 
+ * Since asset loader only support one file reference, any of files can be used.
```
cypherdare, line 30: Could expand this explanation to say, "provided they use the same name prefix. If the names do not match, use ShaderProgramLoader to specify the individual names."

cypherdare:
I've looked over the new code (haven't tried using it yet). Overall I like how clean it is to use and the minimal code breakage. It think it would be good to list exactly what the breaking changes are. 

Here's what I see (please correct me if I'm wrong):

1. The static vertex and fragment prefixes are moved to ShaderStage.
2. The ShaderLoaderParameter uses ShaderFiles instead of directly using explicit shader file names.
3. It seems you will be forced to use consistent shader file suffixes across your project now. (Not a big deal, but potentially breaking since you could previously customize it per ShaderProgramParameter.)

It looks to me like if you were just loading without using a ShaderLoaderParameter, with a `.vert` or `.frag` suffix, it should continue to work without changes.

Comments:
 * Regarding 2: I think we should provide one more constructor that takes nothing but a file name and uses the suffix to determine the ShaderStage. It would make it much easier to instantiate any number of these when defining the loading parameters. I often have a mix of independent shaders where I just load a ".vert" file and let ShaderProgramLoader grab the corresponding ".frag", but I also have lots of shaders where a vertex shader is shared among multiple shaders using different fragment programs. So I will be using lots of these ShaderStage arrays.
* The five default ShaderStages should have Javadocs since their names don't indicate that they're just the defaults. But I agree with keeping the names short.
* ShaderFile needs Javadocs for its prepend code field, since its important to know that the static prepend code (usually used just for versioning) goes before the local prepend code.
* Should the ShaderStage constructor be public? It doesn't seem like they are intended to be replaced, since their respective fields are marked final. On a related note, maybe `stages` should be exposed as just an Iterator so it doesn't imply that you can change out the individual stages there. Probably unnecessary if the constructor is private, though.

I'll try to run this very soon, mostly to check that it doesn't break any behavior for traditional two-part shaders.

payne911:
@cypherdare any updates on your attempt to run this?

mgsx-dev:
Thank you @cypherdare for your feedback, i appreciate it. To be honest i'm not fully satisfied with this PR and few things missing right now (eg. transform feedback with geometry shader needs to declare vertex attribute target before program linking). I need to find time to work on all this, so maybe we could consider this PR as a long time work. So, this PR is definitely not ready to be merged soon.
By then, devs can always use it as is or a part of it in their own code/fork.

As a side note: For people interested in compute shaders (not included in this PR), i made some experiments and i think it doesn't need to be part of LibGDX ShaderProgram/Mesh API, it can be implemented independently using Lwjgl API.

cypherdare:
@payne911 I'd forgotten about it, sorry. I think I'll wait for the next commit. I don't have a lot of time for it right now.

PR #4966, opened 2017-11-05
Corrected array length test in Intersector.intersectRayTriangles

wytrem:
I just spotted a tiny mistake in the array length test.

(sorry for the invalid-email-address I forgot the configuration before commit...)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -620,7 +620,7 @@ public final class Intersector {
 		float min_dist = Float.MAX_VALUE;
 		boolean hit = false;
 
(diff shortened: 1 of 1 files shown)


mgsx-dev:
are you sure ? should be % 9 (3 vertex with the 3 XYZ components)

BTW the loop test `i < triangles.length - 6` could just be `i < triangles.length`

mgsx-dev:
my advice would be to : 

* remove the unnecessary `- 6` in the for loop
* change the message in the exception according to the test (multiple of 9)
* finally squash all your commits in one with a proper comment like "Fixed array length test in Intersector.intersectRayTriangles" to keep a clean git history.

wytrem:
Thanks for your advices, I'll apply this as soon as I can. Sorry for my noobness about PR, it's my first time doing it. 😊

Tom-Ski:
This is an already correct length check, no? This change is to avoid the extra operation?

Issue #4425, opened 2016-11-17
Labels don't wrap in the correct position with markup enabled


DigNZ:
#### Issue details
If markup is used to changed the colour of a word that is at a wrap boundary then the word will not wrap correctly and the label will spill out of the bounds of the table that it is contained in. This increases the width of the layout and causes text to be cut off or drawn in the wrong place.

I've tested combinations of opening and closing the colour markup as well as different colours and the consistently reproducible thing is that if the word after the colour change needed to wrap then it will not. (even with a space between the markup and the word).

#### Reproduction steps/code
Note in the example below if the [GREEN] tag is moved one word before or after its current position then all the text will wrap correctly and fit within the table as it should. 
```
public class TextWrapping extends ApplicationAdapter {

    Stage activeStage;

    @Override
    public void create() {
        activeStage = new Stage(new FillViewport(640, 480));
        BitmapFont font = new BitmapFont();
        font.getData().markupEnabled = true;
        Table stageContentTable = new Table();
        stageContentTable.setFillParent(true);
        Label l = new Label("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris ut lacus maximus, [GREEN]volutpat ex vel, tempor leo. Vivamus elit risus, iaculis at ante vel, commodo mattis tellus. Etiam ac tempor quam, quis aliquet nisl. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Duis nisl velit, scelerisque sed ultricies ornare, ultrices non tortor. Duis justo magna, dapibus quis accumsan a, tristique eu metus. Vestibulum pellentesque lobortis ante, vel maximus odio ultricies ut. Sed lobortis cursus sem, a facilisis ex volutpat in. Pellentesque id libero quis ipsum euismod ullamcorper sit amet at enim. Donec sed dictum ex."
                + "Donec dignissim felis et malesuada egestas. Vivamus vel fringilla lorem. Mauris accumsan eget turpis non volutpat. Nunc facilisis laoreet mauris, scelerisque aliquet lacus eleifend eu. Integer id dignissim purus. Nunc dignissim commodo venenatis. Nullam massa mauris, elementum at rutrum sit amet, lacinia ut arcu. Quisque non nibh eget justo ornare maximus. Vestibulum finibus mi augue, eu gravida purus pulvinar ut. In fringilla dignissim suscipit."
[...]

Issue #4192, opened 2016-07-15
Color markup causes text to be display in wrong position, and does not escape properly


Nickardson:
#### Issue details

When using color markup, each subsequent block after a colored block is offset to the left compared to unstyled text.

Additionally, the markup escape code `[[` does not correctly cause the markup to be ignored if a closing bracket follows after a word which happens to be a valid color.

![libgdxcolormarkup](https://cloud.githubusercontent.com/assets/1232124/16882909/68a9b026-4a8f-11e6-9150-241e3017b3b8.png)
#### Reproduction steps/code

``` java
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.g2d.BitmapFont;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;

public class ColorRenderingMain extends ApplicationAdapter {
    static final String TEXT_PLAIN = "AAA BBB CCC DDD EEE";
    static final String TEXT_COLOR = "[RED]AAA [BLUE]BBB [RED]CCC [BLUE]DDD [RED]EEE";
    static final String TEXT_ESCAPE = "[[ORANGE]Escaped, but is colored";

    SpriteBatch batch;
    BitmapFont font;

    public void create () {
        batch = new SpriteBatch();
        font = new BitmapFont();
        font.setColor(Color.BLACK);

        font.getData().markupEnabled = true;        
    }

    public void render () {
        Gdx.gl.glClearColor(1, 1, 1, 1);
        Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);

        batch.begin();
        font.draw(batch, TEXT_PLAIN, 5, Gdx.graphics.getHeight() - 5);
        font.draw(batch, TEXT_COLOR, 5, Gdx.graphics.getHeight() - 20);
        font.draw(batch, TEXT_ESCAPE, 5, Gdx.graphics.getHeight() - 35);
        batch.end();
    }

    public static void main (String[] args) throws Exception {
        new LwjglApplication(new ColorRenderingMain());
    }
}
```
#### Version of LibGDX and/or relevant dependencies

1.9.3
#### Please select the affected platforms

I've only tested this on windows.
- [x] Windows

NathanSweet:
[...]

PR #4970, opened 2017-11-08
Fixes Labels with color markup and wrap (#4192 and #4425)

DavidPDev:
This is my first PR...

We where having some issues with the color markup + wrapping (issue #4425). So after some debbugging I found the main problem was that the `run.width` wasn't set to 0 after a line break.

```
	if (wrapIndex == 0) {
		next = run; // No wrap index, move entire run to next line.
		run.width = 0;
	}
```
Also we found some minor glitches with offsets on the first glyph of every new line.
Not sure what was the intention of the code before. But I think the first character of a new line must be always set to offset 0 (as it works with the first line)

```
	//xAdvances2.set(0, -glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft);
	xAdvances2.set(0,0);	// Set first glyph starting at 0
```

Also the boolean variable ` colorRun` was breaking some color markup styles (issue #4192). Maybe it's some obsolete code. I removed the colorRun variable and send a false parameter when calling
`fontData.getGlyphs(run, str, runStart, runEnd, false);`

Now the text is wrapped and colored correctly (#4425):
glyphs1
And (#4192)
glyphs2

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -105,7 +105,7 @@ public class GlyphLayout implements Poolable {
 		while (true) {
 			// Each run is delimited by newline or left square bracket.
 			int runEnd = -1;
-			boolean newline = false;
+			boolean newline = false, colorRun = false;
 			if (start == end) {
 				if (runStart == end) break; // End of string with no run to process, we're done.
 				runEnd = end; // End of string, process last run.
@@ -124,6 +124,7 @@ public class GlyphLayout implements Poolable {
 							runEnd = start - 1;
 							start += length + 1;
 							nextColor = colorStack.peek();
+							colorRun = true;
 						} else if (length == -2) {
 							start++; // Skip first of "[[" escape sequence.
 							continue outer;
@@ -140,7 +141,7 @@ public class GlyphLayout implements Poolable {
 					run.color.set(color);
 					run.x = x;
 					run.y = y;
-					fontData.getGlyphs(run, str, runStart, runEnd, false);
+					fontData.getGlyphs(run, str, runStart, runEnd, colorRun);
 					if (run.glyphs.size == 0)
 						glyphRunPool.free(run);
 					else {


gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -824,6 +824,10 @@ public int getWrapIndex (Array<Glyph> glyphs, int start) {
 				char ch = (char)glyphs.get(i).id;
 				if (isWhitespace(ch) || isBreakChar(ch)) return i + 1;
 			}
+			for (i=0; i<start; i++) { // Move forward excluding whitespaces
+				char ch = (char)glyphs.get(i).id;
+				if (!isWhitespace(ch)) return i;
+			}
```
NathanSweet: This looks fishy. The code in this method:
1) Looks backward from `start - 1` to the first non-whitespace character. I don't remember exactly why.
2) From there, it continues looking backward to the first whitespace or break character. If found, it returns the index of the character after the whitespace or break character.
3) The new code starts looking from 0 and returns the first non-whitespace character. However, the new code is only encountered if there was no whitespace or break characters found in step 2.

When is the new code useful?

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -170,8 +169,10 @@ else if (targetWidth <= font.data.spaceWidth) //
 									wrapIndex = i - 1;
 								}
 								GlyphRun next;
-								if (wrapIndex == 0)
+								if (wrapIndex == 0) {
 									next = run; // No wrap index, move entire run to next line.
+									run.width = 0;
+								}
```
NathanSweet: The run width was fixed recently. Can you please merge the latest `master` into your branch, fix conflicts, commit, and push?

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -141,7 +140,7 @@ else if (targetWidth <= font.data.spaceWidth) //
 					run.color.set(color);
 					run.x = x;
 					run.y = y;
-					fontData.getGlyphs(run, str, runStart, runEnd, colorRun);
+					fontData.getGlyphs(run, str, runStart, runEnd, false);
```
NathanSweet: `colorRun` was added in 8ed9d9d928cee9008fb055dc67a3588d0f15be8c. Can you take BitmapFontTest.java from that commit and see how it behaves with your changes versus without? Also see comment below about `xoffset` and BitmapFont `tightBounds`.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -317,7 +318,8 @@ private GlyphRun wrap (BitmapFontData fontData, GlyphRun first, Pool<GlyphRun> g
 			FloatArray xAdvances2 = first.xAdvances; // Starts with all the xAdvances.
 			xAdvances1.addAll(xAdvances2, 0, wrapIndex + 1);
 			xAdvances2.removeRange(1, wrapIndex); // Leave first entry to be overwritten by next line.
-			xAdvances2.set(0, -glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft);
+			//xAdvances2.set(0, -glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft);
+			xAdvances2.set(0,0);	// Set first glyph starting at 0
```
NathanSweet: It's been a while, but I believe `xoffset` is the amount if whitespace pixels that were stripped off the left side of the glyph. A glyph drawn at position `x` is actually drawn at `x + xoffset` to render as if no whitespace pixels were stripped.

`padLeft` is extra padding added on the left side of the glyph, eg for a drop shadow or other effect.

The old code is moving the glyph so its actual left edge aligns to the drawing position. To see the difference, you'll need to have wrapping start a new line with a glyph that has `xoffset` (you could edit some glyphs in your font file to have a large xoffset) and/or use a font with `padLeft`.

> I think the first character of a new line must be always set to offset 0 (as it works with the first line)

The first line doesn't necessarily use a position of 0. When BitmapFont `getGlyphs` is passed `tightBounds=true`, it applies the same adjustment.

As explained in https://github.com/libgdx/libgdx/issues/4192#issuecomment-238059070, we could have separate `tightBoundsStart` and `tightBoundsEnd` parameters for BitmapFont `getGlyphs`. `tightBoundsStart` is only true when the run is the first one on a line. `tightBoundsEnd` is only true when a run os the last one on a line.

Note https://github.com/libgdx/libgdx/commit/0852f43d5cf8a0ffe35824ce9f7c1301d760747d#commitcomment-25344466 swapped the left/right font padding, apparently to match the AngelCode font file format. If your font was created before this swapping AND your font has different left and right padding, then that could cause incorrect rendering. The swapping snuck in an unrelated PR, which makes for an unfortunate gotcha.

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
DavidPDev: This is the case: Unwanted 1st space on the second line (I'll try to explain it):
glyphs3

When the line is cut-off and we try to find the index we use the function
```
/** Returns the first valid glyph index to use to wrap to the next line, starting at the specified start index and
 * (typically) moving toward the beginning of the glyphs array. */
public int getWrapIndex (Array<Glyph> glyphs, int start) {
 ...
 return 0;
}
```
In that case when cutting-off the second line the execution falls to the ``return 0;`` last sentence of the function, and the next line starts with a blank space (1st char of the second line). This is because the first char of the ``Array<Glyph> glyphs`` is la blank space.
The new block tries to avoid the ``return 0`` case, advancing the blanks.

You can reproduce the glitch with the sample at issue #4425. Maybe you have to move the "[GREEN]" markup until you catch it.
It occurs when the [GREEN] is before a space. In my sample:
```
String tex2= "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris ut lacus[GREEN] maximus, volutpat ex vel, tempor leo....."
```

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
DavidPDev: Ok, I saw it later :(. Thanks

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
DavidPDev: Ok, I will create again my font with the Hiero tool (last version available), and do some more tests.
Thanks a lot!

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
DavidPDev: 1) With the original code
``fontData.getGlyphs(run, str, runStart, runEnd, colorRun);``
glyphs2

2) With my code: always passing a false (dont using the tightBounds mode)
``fontData.getGlyphs(run, str, runStart, runEnd, false);``
glyphs1

Maybe I'm missing something but as you commented on the issue #4192 
>...  Ideally colored and not colored text would have exactly the same spacing, but the complexity of breaking the text into runs makes it hard, especially when wrapping and truncation are considered.

Or there should be a new parameter on the getGlyphs function for this cases. I'm not sure.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
DavidPDev: Hmmm... The last commits messed the Pull request with some unwanted commits with shity tests. I don't know how to remove them. Sorry for that.

xGnoSiSx:
Nice!!

Thanks for taking the time to look and submit something about GlyphLayout!

kalexmills:
(I need this PR for a project. All my <3 to whoever gets it reviewed. 👍 )

NathanSweet:
Something is very wrong with this PR, it has way too many changes:

kalexmills:
@DavidPDev you might try creating a new branch from the libgdx master branch and using "git cherry-pick" to cherry-pick the relevant commit(s) onto a new branch in the order they should be applied.

HTH

nelsonkyshin:
Awesome. This will enable proper animating of text labels letter by letter via a [CLEAR] color. @DavidPDev what is the status of this PR?

DavidPDev:
I close this PR..
I will do again some tests and create a new clean PR.

PR #4971, opened 2017-11-08
XmlReader wrong exception throws fix

Enigo:
Those methods never throws `IOException`.


NathanSweet:
Thanks. I will also change XmlReader.rl.

PR #4972, opened 2017-11-10
Added the ability to add new animations to an existing ModelInstance

dpproduction:
I faced with problem when there are many files with animations and need to add them to existed ModelInstance. 
I described the problem in detail here:
https://stackoverflow.com/questions/47092776/libgdx-cant-play-the-animation-on-3d-model

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/ModelInstance.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/ModelInstance.java
@@ -284,42 +284,66 @@ public class ModelInstance implements RenderableProvider {
 		}
 	}
 
-	private void copyAnimations (final Iterable<Animation> source, boolean shareKeyframes) {
+	/** Copy source animations to this ModelInstance
+	 * @param source Iterable collection of source animations {@link Animation}*/
+	public void copyAnimations (final Iterable<Animation> source) {
 		for (final Animation anim : source) {
-			Animation animation = new Animation();
-			animation.id = anim.id;
-			animation.duration = anim.duration;
-			for (final NodeAnimation nanim : anim.nodeAnimations) {
-				final Node node = getNode(nanim.node.id);
-				if (node == null) continue;
-				NodeAnimation nodeAnim = new NodeAnimation();
-				nodeAnim.node = node;
-				if (shareKeyframes) {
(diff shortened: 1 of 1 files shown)


mgsx-dev:
I think your need is reasonable but your solution has some drawbacks, mainly because the modelInstance modify the shared model : 

* existing model instance won't be affected
* new model instances will be affected

I faced the same need but i solved it differently : I just copy animations from a model to some modelInstance as it's done in model instance creation code like this : 

```java
			ModelInstance mi; // a model instance with animations or not
			Model model; // a model with some other animations
			for(Animation anim : model.animations)
			{
				Animation newAnim = new Animation();
				newAnim.duration = anim.duration;
				newAnim.id = anim.id;
				for(NodeAnimation nodeAnim : anim.nodeAnimations)
				{
					NodeAnimation newNodeAnim = new NodeAnimation();
					newNodeAnim.node = mi.getNode(nodeAnim.node.id);
					newNodeAnim.translation = nodeAnim.translation;
					newNodeAnim.rotation = nodeAnim.rotation;
					newNodeAnim.scaling = nodeAnim.scaling;
					
					newAnim.nodeAnimations.add(newNodeAnim);
				}
				
				mi.animations.add(newAnim);
			}
```

Maybe this approach could be used in your addAnimation method.
What do you think ?

dpproduction:
Yes, I agree with you. 
I made some changes, now it looks clear.

xoppa:
Looks good. A few things though:

`addXXX` feels not good here, because the animations aren't added as-is, but copied instead. So modifications made to the animation after the method call wont affect the animation within the instance. Just making the `copyAnimations` method `public` seems more intuitive to me.

I'd expect that a common use-case would be to add one single animation, which is not possible with this change. So perhaps better split the method into a `copyAnimation` and `copyAnimations` method, where the latter just iterates over the animations and calls the `copyAnimation` method.

Also, because it's part of the public API, this needs javadocs.

dpproduction:
Thank you for advice.
I made copyAnimation method public and split it to four methods, check the commit please.

GemmaLamont:
This is exactly what I need for my project, will it be added in soon? :)

catmanjan:
hey @mgsx-dev can you take a look at my issue #5643 and see what you think?

Issue #4974, opened 2017-11-11
AtlasTmxMapLoader not working as documented


heschlie:
#### Issue details
The javadocs for `AtlasTmxMapLoader` states the following (trimmed a bit)

> The atlas must have in it indexed regions named after the tilesets used in the map. The indexes shall be local to the tileset (not the global id).

But this is not the case, and when you do you get NPE if you have an object layer with stuff in it, or just a messed up map with the wrong tiles in the wrong places.

I feel, and I might be wrong on this, that you should be able to simply pack all your tilesets using the texturepacker indexed like stated above and point the map at it. It keeps the process simple for the end user and feels like different tools are working together. At a minimum the docs should be updated to reflect reality.

#### Reproduction steps/code
- Create a tmx file with the atlas property at the map level as indicated, include at least two tilesheets (with just one the local ID and Global ID are the same)
- Pack your tilesheets into an atlas using the libgdx texturepacker
- Try to load the map
- See NPE if you have an object layer with objects in it, or tiles in the wrong place

#### Version of LibGDX and/or relevant dependencies
libgdx version: 1.9.7

#### Stacktrace
```java
// NPE with an object layer
java.lang.NullPointerException
	at com.badlogic.gdx.maps.tiled.objects.TiledMapTileMapObject.<init>(TiledMapTileMapObject.java:42)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:278)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:222)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObjectGroup(BaseTmxMapLoader.java:160)
	at com.badlogic.gdx.maps.tiled.AtlasTmxMapLoader.loadMap(AtlasTmxMapLoader.java:276)
	at com.badlogic.gdx.maps.tiled.AtlasTmxMapLoader.load(AtlasTmxMapLoader.java:160)
	at com.badlogic.gdx.maps.tiled.AtlasTmxMapLoader.load(AtlasTmxMapLoader.java:112)
	at com.ifgames.rabbots.levels.Level.loadTileMap(Level.java:291)
	at com.ifgames.rabbots.levels.Level.init(Level.java:136)
	at com.ifgames.rabbots.screens.DebugLevelMenu$1.clicked(DebugLevelMenu.java:48)
[...]

PR #4975, opened 2017-11-11
fix AtlasTmxMapLoader tileset tile id offset

mgsx-dev:
There was an issue with tile indexing in tileset when using the AtlasTMXMapLoader.

It actually fix the libgdx test named "TideMapAssetManagerTest", now you can see the missing ground !

I created a SSCCE to illustrate the problem but I saw it fixes an existing libgdx test. I provide my SSCCE though, just in case : there is a static variable "useOriginal" allowing to switch from original (using TMXMapLoader) to the same generated with the TiledMapPacker tool (using AtlasTMXMapLoader).

[tmx-sccce.zip](https://github.com/libgdx/libgdx/files/1463628/tmx-sccce.zip)

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java
@@ -358,7 +358,7 @@ public class AtlasTmxMapLoader extends BaseTmxMapLoader<AtlasTmxMapLoader.AtlasT
 				for (AtlasRegion region : atlas.findRegions(regionsName)) {
 					// handle unused tile ids
 					if (region != null) {
-						int tileid = region.index + 1;
+						int tileid = region.index + firstgid;
 						if (tileid >= firstgid && tileid <= lastgid) {
 							StaticTiledMapTile tile = new StaticTiledMapTile(region);
 							tile.setId(tileid);


heschlie:
This fixes #4974 as well.

texoft:
This change breaks loading of Tiled maps with multiple tilesets for me.

E.g:
Have several tilesets defined in my map, but two is enough to demonstrate the problem.

First tileset has firstgid="1", second tileset has firstgid="106". I have texture atlas file generated with indices going from 0 to say 151. The tiled computation is fine for first tileset, but for the second tileset firstgid of 106 added to the index (starting with 105) therefore getting out of (firstgid, lastgid) bounds.

Attached is redacted set of files:

[libgdx_issue.zip](https://github.com/libgdx/libgdx/files/1853765/libgdx_issue.zip)

mgsx-dev:
@texoft 

I investigate your problem and i think your data is obsolete. Since about 3 years (see [this commit](https://github.com/libgdx/libgdx/commit/dba69d2c1af4394e24224321d8a4c79c6939e56e#diff-1c042a2fe2e7a0e16aa829b446410a3dR339) ), tiles are zero-based indexed. From the code : 

```
// AtlasTmxMapLoader expects every tileset's index to begin at zero for the first tile in every tileset.
// so the region's adjusted gid is (gid - layout.firstgid). firstgid will be added back in AtlasTmxMapLoader on load
```

I think my change actually just fix an old broken merge.

Maybe you could try to generate again your atlas tiled map from your original one (using latest version of TiledMapPacker) it should work.

PR #4978, opened 2017-11-12
fix android gradle tool 3.0 can not download

whitecostume:


Tom-Ski:
That repo is not valid for versions of gradle under 4.1.

PR #4989, opened 2017-11-19
Fixed input bug with Dialog letting click events getting to listeners with touchfocus

WinterAlexander:
Please consider this example 

```
TextButton testBtn = new TextButton("Test", skin);
testBtn.addListener(new ClickAdapter(() -> {

	if(testBtn.isDisabled())
		return;

	testBtn.setDisabled(true);

	new Thread(() -> {
		try
		{
			Thread.sleep(200); //simulate async call
		}
		catch(InterruptedException ignored) {}

		Gdx.app.postRunnable(() -> {
			Dialog diag1 = new Dialog("", skin);

			diag1.getContentTable().pad(20);
			diag1.getButtonTable().padBottom(10);

			Label label = new Label("test", diag1.getSkin());
			label.setWrap(true);
			diag1.getContentTable().add(label).prefWidth(500);

			Button ok = new Button(diag1.getSkin());
			ok.add(new Label("Ok", diag1.getSkin()));
			ok.padLeft(30).padRight(30);

			ok.addListener(new ClickAdapter(() -> {
				diag1.hide();
			}));

			diag1.getButtonTable().add(ok);
			diag1.show(stage);
			testBtn.setDisabled(false);
		});
	}).start();
}));
testBtn.setBounds(1420, 220, 160, 40);

stage.addActor(testBtn);
```

This is a button getting disabled during async work and undisabled when the work is done. The problem is that it is possible to click the button right after it being disabled but before the modal Dialog shows up.  All you have to do to produce this bug is to click the button once and then let it be pressed until the Dialog shows up and then release. The second release will trigger a second dialog no matter how long you waited.

ClickAdapter is simple class extending ClickListener and does nothing special but accepting a Runnable and running it in the clicked() method.

The problem is caused by the touchfocus array present in the stage. If this array is not cleared when the Dialog is shown, the existing touchfocus will match with the touchup events getting fired and call the listener.

The fix was simple: clearing the touchfocus array when showing the dialog.

Note: This should be applied to any modal Window for it to be really modal. However, it would require adding a show method to Window since Window doesn't have any.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Dialog.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Dialog.java
@@ -199,6 +199,7 @@ public class Dialog extends Window {
 
 		pack();
 		stage.addActor(this);
+		stage.cancelTouchFocus();
 		stage.setKeyboardFocus(this);
 		stage.setScrollFocus(this);
 		if (action != null) addAction(action);


NathanSweet:
Cheers!

PR #5006, opened 2017-12-04
perspective divide for shadow maps

ryanastout:
perspective divide is necessary for perspective cameras and is still correct for orthographic cameras since for ortho w is always 1. This change would allow users making shadow maps for point lights to use the depth.vertex.glsl and default.vertex.glsl vertex shaders if they want


Tom-Ski:
Thanks @ryanastout, implemented here https://github.com/libgdx/libgdx/commit/bc864a01de2e9ff37ce958ace597e31f944dd39a

PR #5010, opened 2017-12-07
Actor fixes

patrykcd:
Now functions:

myactor.setPosition(0, 0, Align.left); 
is equal to 
myactor.setX(0,Align.left);
myactor.setY(0,Align.center);
Move half actor to down and full to left (before that it moved only half actor down. All "aligns" were inverted - bottomLeft was topRight, left was right and so on)

And setOrigin is now relative to actor as is written in doc:
“Sets the origin position which is relative to the actor's bottom left corner."

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -589,7 +589,7 @@ public class Actor {
 	}
 
 	public void setOriginX (float originX) {
-		this.originX = originX;
+		this.originX = originX + x;
 	}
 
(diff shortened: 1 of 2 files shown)


Darkyenus:
I have a feeling that you misunderstood how the position with alignment works. What is aligned is the point to which the coordinates refer inside the Actor, not the actor around the point.

Also, the origin is independent on the Actor's position, as documented. This makes the origin relative to world coordinate system, which will probably break a lot of things inside the libGDX.

Please try to understand how the library works first, before concluding that it is broken.

patrykcd:
ok, thank you for explanation. I understand now.

PR #5021, opened 2017-12-14
Fix issue with polygon region out of bounds

michael-reznik:
This is a similar issue to [807c7c3](https://github.com/libgdx/libgdx/commit/807c7c323b740a0d6feddaf7d820124d39d6c230), where sometimes in some versions of Android 6 it throws ArrayIndexOutOfBoundsException when using PolygonRegion (appears that it has something to do with proguarding this file).
This issue has been also mentioned in the [community](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=23307).

After my changes it seems that the crash no longer happens.
Tested with problematic device and with devices that worked ok before the changes.

The functionality of the code has not changed.

Relevant stack trace part:
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=8; index=8
       at com.badlogic.gdx.graphics.g2d.PolygonRegion.(PolygonRegion.java)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonRegion.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonRegion.java
@@ -41,10 +41,10 @@ public class PolygonRegion {
 		float uvHeight = region.v2 - v;
 		int width = region.regionWidth;
 		int height = region.regionHeight;
-		for (int i = 0, n = vertices.length; i < n; i++) {
+		for (int i = 0; i < vertices.length;) {
 			textureCoords[i] = u + uvWidth * (vertices[i] / width);
-			i++;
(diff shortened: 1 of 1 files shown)


pavelzag:
Finally! This should be integrated ASAP.

NathanSweet:
If Proguard has a bug, it should be reproduced and submitted to the Proguard maintainers. A fix like this, which is just shuffling things and hoping it makes a difference, doesn't stop the same Proguard bug from affecting other parts of libgdx. We certainly aren't going to shuffle *all* situations like this.

This is just voodoo, but since it isn't a big change, we can try it this time.

theyonibomber:
When will a version with this fix is planned to be released? Many of our users are suffering from this voodoo and this fix solves the issue for them.

PR #5024, opened 2017-12-16
Make TextField's cursor blink with non continuous rendering

### LINKED ISSUE #5023: TextField's cursor is not blinking with non continuous rendering
**NCR** - non continuous rendering

With *NCR* `TextField`'s cursor is not blinking and I find it is little hard to spot the cursor position when it isn't animated at all. I get the idea that with *NCR* we should reduce render calls wherever it's possible, but cursor blinks only trice in a second and also it only happens when field is selected. So maybe it would be better to just leave blinking in a place (we request rendering when text is changed and every tick when any `Action` is present after all)?

Other option could be to provide user with an option to choose blinking mode, but I hardly believe that someone really prefers to save 3 frames per second to overall usability.

1.9.8

- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

metaphore:
Fixes #5023

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -418,14 +418,11 @@ public class TextField extends Widget implements Disableable {
 	}
 
 	private void blink () {
-		if (!Gdx.graphics.isContinuousRendering()) {
-			cursorOn = true;
-			return;
-		}
 		long time = TimeUtils.nanoTime();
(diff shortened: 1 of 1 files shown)


metaphore:
My bad, I've checked TextField's code once again and saw that blink() actually triggers from render() which makes my patch completely useless.

NathanSweet:
So far it has been OK to just show the cursor without blinking. I suppose the libgdx Timer class could be used to trigger a redraw when the cursor should be shown/hidden.

PR #5027, opened 2017-12-18
TextField uses timer task to perform cursor blinking

metaphore:
Related to #5023 

`TextField` now relies on `Timer` and update cursor visibility from its callback. This will allow `TextField` to have a blink animation even under non-continuous rendering state.

One drawback should be considered though. There are cases when blinking should be restarted (key input, cursor moved) and since `Timer` API doesn't allow to reset scheduled repeated task immediately, workaround is used and the task imitates reset by making cursor visible and skipping one blink cycle. But this should be fine due to blink ticks happen pretty often and effect is barely noticeable.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -36,13 +36,13 @@ import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener.ChangeEvent;
 import com.badlogic.gdx.scenes.scene2d.utils.ClickListener;
 import com.badlogic.gdx.scenes.scene2d.utils.Disableable;
 import com.badlogic.gdx.scenes.scene2d.utils.Drawable;
+import com.badlogic.gdx.scenes.scene2d.utils.FocusListener;
 import com.badlogic.gdx.scenes.scene2d.utils.UIUtils;
 import com.badlogic.gdx.utils.Align;
 import com.badlogic.gdx.utils.Array;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -773,6 +770,28 @@ public void run () {
...
+				skipNextBlink = false;
+				return;
+			}
+
+			cursorOn = !cursorOn;
+			Gdx.graphics.requestRendering();
+		}
+
+		/** Forces the task to skip next blink cycle update. Since {@link Timer} API doesn't provide a way to restart repeated task
+		 * immediately, this method is used to imitate blinking restart. */
```
NathanSweet: We should be able to cancel the task and schedule it again, then we don't need `skipNextBlink`.

metaphore: Nate, could you please implement `Timer#remove(Task)` or `Timer#restart(Task)` then it should be easy to change.

NathanSweet: I don't think we need that. Use:
```
Task task = ...
...
task.cancel();
Timer.instance().scheduleTask(task, delay, interval);
```
This works, though it will result in the task being in the timer task array 2+ times. The first one triggers the task, the rest are ignored, but it still isn't great. I have fixed that in a5fa372522445e23300ca8fe861dd384691b6932.

metaphore:
Oh, I'm a bit late. Anyway, thanks Nate!

PR #5033, opened 2017-12-23
Added addText to Label.java

kalexmills:
Should enable changing the color and subsequently adding new text (in the new color).


kalexmills:
If it turns out not to be this simple, I am very teachable... :-)

NathanSweet:
I don't think this does what you want. `invalidateHierarchy` will cause `layout` to get called which uses GlyphLayout `setText` and BitmapFontCache `setText`. These will reset the color of any text already in the label, making the new `addText` method obsolete -- you might as well just call Label `setText`.

For a label to have multiple colors, enable color markup for the font and use a string like `[white]text[red]more text`.

Issue #4890, opened 2017-09-15
MoveToAction should expose startX and startY?


ghost:
I'm looping through an Actor's actions, and if it's a MoveToAction, I need to extend the X position (because I need to preserve its interpolation logic instead of creating a new action). 

However, I need to know which direction MoveToAction is going, which means it would be really convenient to call moveToAction.getStartX().  My workaround is rather nasty...

Unfortunately, these are private fields. Any change these could be made public via getters/setters?

Thanks.

piotr-j:
Since you need custom logic, you should probably use custom action.

notsonotso:
@piotr-j Being able to get the starting position would be convenient.

gonzeD:
I've made a pull request to correct this issue https://github.com/libgdx/libgdx/pull/5043

MrStahlfelge:
Closing this, attached PR was closed, to. As suggested, you should use a custom action.

PR #5043, opened 2017-12-29
Adding  getter for startX and startY in moveToAction

gonzeD:
Correcting the error specified in the issue 4980 (https://github.com/libgdx/libgdx/issues/4890). Since the attributes are private, there is no risk in exposing them via getter.


cypherdare:
These should probably have javadocs that state that the values are only valid after the action is begun.

PR #5046, opened 2017-12-31
Added DebugModelInstance to easily change primitive type.

kalexmills:
Intended usage is to simply change out the class used whenever wireframes / vertices are needed on specific models. Tested locally.

Any feedback and ways to make this more a more useful wrapper would be welcome.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/DebugModelInstance.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/DebugModelInstance.java
@@ -1,0 +1,137 @@
+/*******************************************************************************
+ * Copyright 2017 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.graphics.g3d;
+
+import com.badlogic.gdx.graphics.g3d.model.Node;
+import com.badlogic.gdx.graphics.g3d.model.NodePart;
+import com.badlogic.gdx.math.Matrix4;
+import com.badlogic.gdx.math.Vector3;
+import com.badlogic.gdx.utils.Array;
+
+/**
+ * <p>Convenience wrapper for {@link ModelInstance} which can be used for displaying a 
+ * wireframe version of a model to aid in debugging.</p>
+ *
+ * <p>For example, to render a model in wireframe, use:</p>
+ *
+ * <pre>
+ * {@code
+ * Model model = ...;
+ * ModelInstance model = DebugModelInstance(model).setPrimitiveType(GL20.GL_TRIANGLE_STRIP);
+ * }
+ * </pre>
+ * @author kalexmills */
+public class DebugModelInstance extends ModelInstance {
+
+  private int glPrimitiveType;
+
+  public DebugModelInstance(Model model) {
+    super(model);
+  }
+
+  public DebugModelInstance(Model model, String nodeId, boolean mergeTransform) {
+    super(model, nodeId, mergeTransform);
+  }
+
+  public DebugModelInstance(Model model, Matrix4 transform, String nodeId, boolean mergeTransform) {
+    super(model, transform, nodeId, mergeTransform);
+  }
+
+  public DebugModelInstance(Model model, String nodeId, boolean parentTransform, boolean mergeTransform) {
+    super(model, nodeId, parentTransform, mergeTransform);
+  }
+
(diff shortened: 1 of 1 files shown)


Darkyenus:
I personally think that this particular functionality is better suited for a `Shader`, but I wouldn't be against it, if it did more things. Perhaps rendering additional information, like bounding box, center, etc., but implementing that would be probably also easier through a shader...

kalexmills:
Great ideas.

If a shader is preferred, it will probably need to be someone else's implementation. I am completely new to GLSL

Would it take less effort to switch out or set up a shader for a quick debug? Ideally I'd want this to be as easy as calling .debug() or .debugAll() in other areas of the API.

Darkyenus:
(This started as a short response, but evolved into a heap of ideas, sorry)

"`DebugShader`" would be probably best implemented as a [`Shader`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/Shader.java) wrapper, which would delegate all calls to the parent shader, but override `render(Renderable)` method to modify the argument `Renderable` before passing it to the parent (this is how wireframe would be implemented) and possibly calculating and rendering the bounding box, center, etc.

The implementation wouldn't need any GLSL, it would all be pretty straightforward Java code.
To ease the use of this, "`DebugShaderProvider`" could be created, as an implementation of `ShaderProvider`, again wrapping parent `ShaderProvider` and automatically replacing the assigned shader with a wrapped one.

It could be tricky to implement this correctly, because not all `Renderable`s must be pooled so we must check whether the shader isn't already wrapped. Also, since average program won't have too many Shaders, it will be beneficial to implement some sort of caching of wrapped shaders to reduce GC pressure (potential for memory leak here). But if done right, it should be pretty easy to use this.

What the `DebugShader` should do (wireframe?, draw bounding box?, ...) could be even stored in a separate object (`DebugShader.Attributes`?) that could be shared between multiple/all `DebugShader`s. So that there would be one `Attributes` object for all NPCs, one for items, one for effects, etc. and for example enabling wireframe for all NPC's (but not other objects) would be done through single `npcAttributes.setDrawMode(WIREFRAME)` call. Which `DebugShader` instance gets which attributes instance could be customized by overriding a method in `DebugShaderProvider`.

It may also be useful to be able to delay the drawing of the extra objects (bounding box) until later stages, but I currently can't think of any use case where it would be the best approach, so that should be probably left out.

Maybe I'm overthinking it. Nevertheless, thank Xoppa for the 3D API - it is so flexible, that it almost looks like it was written with this sort of thing in mind from the start.

kalexmills:
These are great thoughts. No apology needed. :-) 

> possibly calculating and rendering the bounding box, center, etc.

This should be easy enough using the existing `GeometryUtils` / `MathUtil`s implementation along with `createBox`.

> we must check whether the shader isn't already wrapped

So we should probably wrap using a (static?) method with signature like `Shader DebugShader.wrap(Shader)` method and use `instanceof` to check if the argument is already wrapped, in which case the wrapper can just return the argument unchanged. Something similar might need to be done for the Renderable wrapping, but I could be wrong.

> What the DebugShader should do (wireframe?, draw bounding box?, ...) could be even stored in a separate object (DebugShader.Attributes?) that could be shared between multiple/all DebugShaders.

If I understand you right, it sounds like you're suggesting using flyweight pattern, or doing some sort of resource pooling.  I'm unfamiliar with libgdx's pooling implementation, so if you think this would be the right approach, I'd appreciate a pointer to one of libgdx's existing pooling implementations which is worth reviewing to see how to do pooling right.

xoppa:
Please note that (mis)treating (usually triangle) vertices as line vertices, by setting a different primitive type, wont consistently result in a (good) wireframe. Although it might look like that in your tests, it heavily depends on the vertex layout, shader and render pipeline.

There is no simple (cheap) way to draw a good wireframe based on triangle vertices which is compatible with GLES2 (we've looked into that quite a few times). But if that limitation is acceptable and the goal is to mimic a wireframe by changing the primitive type anyways, then i agree that it should be done using a shader. There is no need to wrap any other shader though (or to pool anything), because you only need the position attributes of the vertices: the shader implementation will be the same for all models anyway.

Keep in mind though that you usually have multiple render passes and you probably dont want to affect all passes this way. Arguable debug rendering would be better fit as a separate render pass (ModelBatch) on the main render target (perhaps without depth testing). Filtering which objects should be rendered that way is probably a higher level decision (just like frustum culling is).

tl;dr Although I like the idea, I don't think that the PR as it is now (`DebugModelInstance`) should be merged.

Darkyenus:
@kalexmills The `Renderable` itself and all of its fields, except the shader field, would remain intact.

Automatic wrapping could leverage the fact, that `ShaderProvider. getShader(Renderable)` is called for all `Renderable`s rendered, and thus is a great place to swap the original shader with the wrapped copy - if the user determines that the given shader/renderable should be wrapped.

The idea of `DebugShader` was that it would, by default, delegate all its functions to the original shader, but it could be configured (directly or through shared configuration) to modify the rendered object (wireframe) or to render additional info (bounding box). The wireframe would still have the colors and shape of the original object (or at least the approximation of that, as noted). To be able to work like this, I believe that it will have to wrap the original shader - but I could be mistaken.

It would be also possible to create just a single-purpose shader, that renders whatever Renderable as a wireframe, and do the other functionality separately. Naive implementation of that separate mechanism would be probably faster, but possibly slightly more difficult to integrate into existing projects.

I'll try to provide some code sketches to show exactly what I mean.

kalexmills:
> xoppa: I don't think that the PR as it is now (DebugModelInstance) should be merged.

I agree. Thanks for all the insights / comments.

@Darkyenus Looking forward to the code sketches, but I will close this PR soon.

PR #5048, opened 2018-01-04
Fix Intersector#overlapConvexPolygons

schosin:
In some rare cases it's possible that Intersector#overlapConvexPolygons calculates an incorrect MTV. Currently the direction of the MTV is negated depending on the number of points on either side of the checked edge. 

The correct way is to negate the direction of the MTV depending on the direction of the centers of both polygons.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -949,17 +949,32 @@ public final class Intersector {
 		float overlap = Float.MAX_VALUE;
 		float smallestAxisX = 0;
 		float smallestAxisY = 0;
-		int numInNormalDir;
 
 		int end1 = offset1 + count1;
 		int end2 = offset2 + count2;
-
+		
+		float min1X = Float.MAX_VALUE, min1Y = Float.MAX_VALUE, min2X = Float.MAX_VALUE, min2Y = Float.MAX_VALUE;
+		float max1X = Float.MIN_VALUE, max1Y = Float.MIN_VALUE, max2X = Float.MIN_VALUE, max2Y = Float.MIN_VALUE;
+		
 		// Get polygon1 axes
 		for (int i = offset1; i < end1; i += 2) {
 			float x1 = verts1[i];
 			float y1 = verts1[i + 1];
 			float x2 = verts1[(i + 2) % count1];
 			float y2 = verts1[(i + 3) % count1];
+			
+			if (x1 < min1X) {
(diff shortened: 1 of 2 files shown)


obigu:
I think your fix doesn't work for other cases. Imagine a right triangle (1 angle is 90º) with 2 equal sides resting on one cathetus (short side). The center point of that triangle will be in the middle of the hypotenuse. Imagine now a small rectangle with the same center than the triangle. If I understand your fix correctly in that case MTV will resolve to 0. Now, imagine we move the rectangle diagonally up-right or down-left the MTV will invert but it actually should keep being positive x and y. I think it's a complex geometry problem with complex corner cases.

kdenzel:
So i had also issues with the function `overlapConvexPolygons` and i decided to understand the SAT Algorithm and how to handle the minimum translation Vector. I am making a game and completely did a rewrite on this function. If anyone is interested in seeing it [here is the link](https://github.com/kdenzel/BloodHunger/blob/master/core/src/de/kswmd/bloodhunger/math/Intersector.java). If you want, you can clone the repository and play with the Testclasses i made. Try changing my custom Intersector class with this of libgdx and you will see the issues with it. I made a Testclass where you can drag 2 triangles to overlap each other and when they collide, the minimum translation vector did return wrong results because the orientation of the x and y coordinates returned by the direction vector were wrong.

obigu:
> So i had also issues with the function `overlapConvexPolygons` and i decided to understand the SAT Algorithm and how to handle the minimum translation Vector. I am making a game and completely did a rewrite on this function. If anyone is interested in seeing it [here is the link](https://github.com/kdenzel/BloodHunger/blob/master/core/src/de/kswmd/bloodhunger/math/Intersector.java). If you want, you can clone the repository and play with the Testclasses i made. Try changing my custom Intersector class with this of libgdx and you will see the issues with it. I made a Testclass where you can drag 2 triangles to overlap each other and when they collide, the minimum translation vector did return wrong results because the orientation of the x and y coordinates returned by the direction vector were wrong.

On a first look it looks a good. I think we  can use your SAT algorithm implementation as a base to replace the current buggy implementation. I think there's a some room for improvement (such as using `Vector2` for some operations) but I'd rather discuss them on a PR. Would you like to create a PR with your implementation including some non visual/interactive tests that also cover corner cases (such as containment)?

MrStahlfelge:
Now that we have #5048, can we close this one?

obigu:
It will get closed when the PR is merged and closed.

PR #5049, opened 2018-01-05
Fix #5036 open URL on mobile device using HTML module

### LINKED ISSUE #5036: Gdx.net.openURI does not work when called from the default iOS browser on iPhone or Android phone
#### Issue details
The code to open an external URL:
e.g.:  `Gdx.net.openURI("http://bbc.com"); `
runs fine when executing the HTML module except when the browser you use for that is the default iPhone browser or Chrome on a real Android device.
On the Mac (using Safari or chrome) the instruction works, but somehow, The version of Safari running  on the iPhone differs and the code does not work (nothing happens) and on an Android phone a message shows that it has been blocked.
This also happens using the IPhone simulator.

#### Reproduction steps/code
Call Gdx.net.openURI("http://bbc.com"); in the  GDX standard application.
Access the HTML module through your browser:
if you access the HTML module from your computer browser it works fine.
If you access the same HTML module using your iPhone default browser  or from Chrome on an Android Device it does not.

1.9.8

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

julienvillegas:
Fix #5036 open URL on mobile device using HTML module
Current implementation did not work on Mobile device (iOS or Android)
as the call is treated as a blocked pop-up.
This implementation redirects the browser windows using JSNI code
(http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html)

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
@@ -64,21 +64,14 @@ public class Lwjgl3Net implements Net {
 
 	@Override
 	public boolean openURI (String URI) {
-		if(SharedLibraryLoader.isMac) {
-			try {
-				FileManager.openURL(URI);
-				return true;
-			} catch (IOException e) {
-				return false;
-			}
-		} else {
-			try {
-				Desktop.getDesktop().browse(new URI(URI));
-				return true;
-			} catch (Throwable t) {
-				return false;
-			}
+		class NativeWrapper {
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Why is this in the lwjgl backend?

julienvillegas:
OpenURL works fine on iOS, Core or Android Modules. On the HTML module however it works when you call it through the desktop but not through the mobile device. Sorry if incorrect but does it not means it should be addressed there?

julienvillegas:
ah!! No Sorry... silly me
ok, let me take it back.

Tom-Ski:
You need the html backend (gwt).

Also this doesn't compile even if it were in the right backend.

PR #5050, opened 2018-01-07
Gdx.net.openURI blocked as popup when using iOS browser on iPhone or Chrome on Android phone #5036

julienvillegas:
When calling Gdx.net.openURI on a GWT module using 
an iPhone or Android phone you may get a popup blocked (see #5036). This addresses it by changing the windows location.
Tested on Chrome/Safari on computer and mobile devices.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -210,7 +210,7 @@ public abstract class GwtApplication implements EntryPoint, Application {
 		Gdx.files = new GwtFiles(preloader);
 		this.input = new GwtInput(graphics.canvas);
 		Gdx.input = this.input;
-		this.net = new GwtNet();
+		this.net = new GwtNet(config);
 		Gdx.net = this.net;
 		this.clipboard = new GwtClipboard();
 		updateLogLabelSize();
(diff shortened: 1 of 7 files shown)


julienvillegas:
Also 

- Added OpenBrowser to list of GWT tests.
- Reordered test alphabetically

Tom-Ski:
test this please

Tom-Ski:
Great! Could you format the jsni code so it looks a little nicer?

What is the difference on desktop if any?

MobiDevelop:
I'm not sure the standard operation should be to take you away from the game in the same window/tab. The user is able to allow the pop-up after being blocked, right?

julienvillegas:
not on iPhone, I could not find a workaround. Happy is there is one.


Julien Villegas
+34 688 916 737
+34 681 135 970

Tom-Ski:
Yes, that's not ideal. If anything this should be an alternate method, or via additional parameter.

julienvillegas:
so two methods? existing one, without parameter which tries to open the new tab, and a new one with a boolean newTab parameter which let the user explicitly state it?

MobiDevelop:
It might be better to make it a configuration value so as to not require updates to all backends.

julienvillegas:
ok, then it would need to be a public static configuration value, right? otherwise I don't really see how we would access that property from within that class.

MobiDevelop:
Why not just pass the configuration down into GwtNet?

julienvillegas:
So in summary:

- Option 1) We could pass down the configuration down to GwtNet as just suggested, however that would mean an update to the backends. 

- Option 2) If we are going to apply an update to the backends, we might as well implement the initial suggestion to add a new OpenURI Method taking a "newTab" boolean flag. that option had been discarded because it required backend update. If we are going to do that anyway, we might as well do that one instead.

- Option 3) Alternatively  there is always the option to have it as a Static variable within the configuration but that's not very clean.

I am inclined for Option 2 but let me know.

MobiDevelop:
Option 1 requires no change to other backends. It is solely an implementation detail of the GWT backend.

julienvillegas:
ok, Done.
As instructed I:

- Added new openURLInNewWindow boolean property in GwtApplicationConfiguration
- Passed the config down to GwtNet
- use Window.open or Window.Location.assign depending on value of parameter.
Tested with existing OpenBrowser test

Issue #5025, opened 2017-12-16
Application's pause() method is called before initialization is done on GWT


MrStahlfelge:
#### Issue details
On GWT, libGDX preloads all assets before initializing the game.

Also on GWT, the game is paused when the web browser tab is hidden by changing to view another tab.

When the tab is changed while the libGDX GWT application is still preloading - users do this when they see that loading will take some time -, the `pause()` method on the game's main `Game` class is called *before* `Gdx.app` variable is initialized and before the `create()` method was called.  This should not happen according to the [documented lifecycle](https://github.com/libgdx/libgdx/wiki/The-life-cycle).

#### Reproduction steps/code
Place some huge assets in the asset directory, and add the following lines to your main game's class:

    public class Tralala extends Game {
	@Override
	public void create() {
                ....
		Gdx.app.log("TEST", "Created");
	}

	@Override
	public void pause()	{
		...
		Gdx.app.log("TEST", "Paused");
	}

Start html:superdev, open localhost:8080/html and while the application is preloading switch the browser tab.
When you switch back, you will see a js error caused by pause() method accessing Gdx.app which is null.

#### Version of LibGDX and/or relevant dependencies
All.

#### Stacktrace
The important lines are in `GwtApplication.java`:
Line [108](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L108) Application Listeners are added before preloading is done by calling `addEventListeners` .
Line [545](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L545) and line [550](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L550) also do not check if initialization is done and call pause()/resume() directly.

`Gdx.app` is assigned and `create()` is called when preloading was finished (line [172](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L172)). The call to `addEventListeners` should be placed here.

#### Please select the affected platforms
- [X] HTML/GWT

MrStahlfelge:
This is solved

PR #5060, opened 2018-01-20
Pause called before GdxApp initiated

julienvillegas:
Fixes Application's pause() method is called before initialization is
done on GWT #5025 by moving the call


julienvillegas:
Ran GWT tests to check it still works..

julienvillegas:
also includes commit for the OpenURI issue #5036

Tom-Ski:
@julienvillegas these fixes look good, but should be separate. As this is now conflicting, can you please send them as separate PRs and I'll merge.

PR #5062, opened 2018-01-23
Handle input pointer id lookup fail not to be fatal on robovm

obigu:
Changes robovm input not to throw exception and just log the error in case touch pointer lookup fails.

I receive from time to time a crash report in one of my games (that is pretty touch intensive) caused by a pointer not found ("Couldn't find pointer id for touch event!")

Even if this should never happen in theory it does in some cases (haven't been able to reproduce) but missing a touch event should not be fatal in my opinion.

It also makes it consistent with Android implementation.

**NOTE**
MOE should also be changed for consistency. I'm not familiar with MOE to be confident about making the changes but they should be pretty similar.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSInput.java
@@ -56,6 +56,7 @@ import com.badlogic.gdx.utils.Pool;
 
 public class IOSInput implements Input {
 	static final int MAX_TOUCHES = 20;
+	private static final int POINTER_NOT_FOUND = -1;
 
 	private static class NSObjectWrapper<T extends NSObject> {
 		private static final long HANDLE_OFFSET;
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSInput.java
```diff
@@ -639,7 +640,13 @@ private int findPointer (UITouch touch) {
 		for (int i = 0; i < touchDown.length; i++) {
 			if (touchDown[i] == ptr) return i;
 		}
-		throw new GdxRuntimeException("Couldn't find pointer id for touch event!");
+		// If pointer is not found
+		StringBuffer buf = new StringBuffer();
```
Darkyenus: Did you mean `StringBuilder`?

obigu: True, I just copied AndroidInput code :), will create a small PR for it and fix this

obigu: Ok, changes made and squashed, thanks!

PR #5063, opened 2018-01-23
Use StringBuilder instead of StringBuffer

obigu:
Just for correctness, no real impact


khorinis91:
@obigu Did you do that survey as well by any chance? 😄

PR #5065, opened 2018-01-23
added the constant GL_FRAMEBUFFER_SRGB which has value 0x8DB9 to GL30.java

ryanastout:
I added a constant to GL30.java


Tom-Ski:
GL_FRAMEBUFFER_SRGB is not part of the GLES 3.0 specification.

PR #5068, opened 2018-01-30
Use equals method of string literal to prevent NPE.

reudismam:


intrigus:
Do you actually have any problems with the current code, i.e. NPEs?

reudismam:
I myself do not get any NPE, but this is a kind of edit pattern that I have found in projects such as Apache Ant, Apache Hive, and Google ExoPlayer.

MobiDevelop:
It's generally considered a best practice to put the literal on the left. It's less important when the other value is known to be non-null (which I think is the case here).

The change is likely unnecessary but also can't hurt.

NathanSweet:
If a value can be null (which should be javadoc'ed), this pattern is appropriate. I don't feel it is necessary to apply this pattern when the value is known to be non-null. I personally don't like how the code reads and the changes don't actually fix anything.

PR #5071, opened 2018-02-01
Add Poolable#onObtain and Poolable#onFree

Ktar5:
Created default methods onObtain() and onFree() within the Poolable interface so that they may be used for state changes that a user may want. For example, the use case that gave me this idea was trying to check if an object was in active or in a pool and there was currently no plain & simple way to do it.

It's been implemented into Pool#freeAll() , Pool#free() , and Pool#obtain() and will work with this implementation due to the nature of Pools' methods using the Pool#free/all/obtain methods as opposed to their own implementation.

Any comments, concerns, suggestions, etcetera would be appreciated if there is anything keeping this from being accepted.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Pool.java
+++ b/gdx/src/com/badlogic/gdx/utils/Pool.java
@@ -48,7 +48,13 @@ abstract public class Pool<T> {
 	/** Returns an object from this pool. The object may be new (from {@link #newObject()}) or reused (previously
 	 * {@link #free(Object) freed}). */
 	public T obtain () {
-		return freeObjects.size == 0 ? newObject() : freeObjects.pop();
+		if (freeObjects.size == 0) {
+			return newObject();
+		} else {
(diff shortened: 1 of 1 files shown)


Ktar5:
Why are we building JDK 6 if minimum requirements (https://libgdx.badlogicgames.com/documentation/gettingstarted/Setting%20Up.html) are JDK 7

intrigus:
You can't use java 8 in the libgdx core

Darkyenus:
This seems to be a relatively big change for the `Pool` and I am worried about it's performance implications, but it could be added as an `Pool` subclass, to be used only when needed (this could also eliminate `instanceof` checks, as the new `PoolableWithOnObtainAndOnFree` interface (working name) could be required by the generic constraint).

Ktar5:
@Darkyenus I think I'll do what you said and work on a new Pool/able implementation where it doesn't use generic Object and instead uses an implements DescriptivePoolable or something of that sort.

PR #5072, opened 2018-02-02
Created "DescriptivePool" utility class

Ktar5:
This PR created the "DescriptivePool" utility class which is used for calling methods `onObtain` and `onFree` whenever an object is freed/obtained from a DescriptivePool instance. Provides extension interface of `Poolable` named `DescriptivePoolable`.

Any comments, concerns, suggestions, etcetera would be appreciated if there is anything keeping this from being accepted.

p.s.: thanks @Darkyenus for the tip. x2

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
+++ b/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
@@ -62,7 +62,7 @@ abstract class DescriptivePool<T extends DescriptivePool.DescriptivePoolable> ex
 
     	/** Objects implementing this interface will have {@link #reset()} called when passed to {@link Pool#free(Object)}. */
     	static public interface DescriptivePoolable extends Poolable {
-		/** Called ONLY when an object is added to the pool.
+		/** Called when an object is added to the pool or created.
       	   	* Note: Will call after {@link #reset()} in {@link Pool#free(Object)} related methods */
         	public void onFree ();
 
--- a/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
+++ b/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.utils;
+
+/** A pool of objects that can be reused to avoid allocation. However, this implements
+ * {@link #onFree(DescriptivePoolable) and {@link #onObtain(DescriptivePoolable)}}
+ * @see Pools
+ * @author Carter Gale */
+abstract class DescriptivePool<T extends DescriptivePool.DescriptivePoolable> extends Pool<T> {
+	/** Creates a pool with an initial capacity of 16 and no maximum. */
+	public DescriptivePool () {
+                super();
+        }
+
+        /** Creates a pool with the specified initial capacity and no maximum. */
+        public DescriptivePool (int initialCapacity) {
+                super(initialCapacity);
+        }
+
+        /** @param max The maximum number of free objects to store in this pool. */
+        public DescriptivePool (int initialCapacity, int max) {
+                super(initialCapacity, max);
+        }
+
+        @Override
+        protected void onObtain (T object) {
+                object.onObtain();
+        }
+
+    	@Override
+    	protected void onFree (T object){
+        	object.onFree();
+    	}
+
+
+    	/** Objects implementing this interface will have {@link #reset()} called when passed to {@link Pool#free(Object)}. */
+    	static public interface DescriptivePoolable extends Poolable {
+       		/** Called ONLY when an object is added to the pool.
+      	   	* Note: Will call after {@link #reset()} in {@link Pool#free(Object)} related methods */
+        	public void onFree ();
+
+        	/** Called whenever the object is removed from the pool. */
+        	public void onObtain ();
+    	}
+}
--- a/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
+++ b/gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
@@ -49,7 +49,7 @@ abstract class DescriptivePool<T extends DescriptivePool.DescriptivePoolable> ex
 
     	/** Objects implementing this interface will have {@link #reset()} called when passed to {@link Pool#free(Object)}. */
     	static public interface DescriptivePoolable extends Poolable {
-       		/** Called ONLY when an object is added to the pool.
+		/** Called ONLY when an object is added to the pool.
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/utils/Pool.java
```diff
@@ -48,19 +48,32 @@ public Pool (int initialCapacity, int max) {
...
+                } else {
+                        T object = freeObjects.pop();
+                        onObtain(object);
+                        return object;
+                }
 	}
-
+	
+	/** Called when an object is retrieved from the pool using {@link #obtain()} */
+	protected void onObtain (T object){ }
```
Darkyenus: The documentation is not 100% accurate: object may be returned by the pool, but without calling `onObtain` - if it is created by the Pool. I don't think that this is necessarily a problem, but it should be documented as it will lead to surprises. Same problem at `onFree`.

Looking at the whole PR, it feels that it would be most beneficial:
- if the `onObtain`/`onFree` were called deterministically (i.e. as documented, not as implemented),
- whole onObtain/onFree could be then implemented directly in `DescriptivePool`, without any changes to `Pool` necessary,
- and the addition of `onFree`/`onRelease` for overriding directly in the `Pool` subclass seems, in retrospect, much more useful than generic bounds.

All ideas are open to discussion, sorry for not realizing these sooner :/

Ktar5: The reason I made it that way for `onObtain`/`onFree` to be called only when actually retrieved or put back is because with `onObtain`, the object's constructor is simply called and I didn't want those two logics to be mixed, worse case scenario you can call `onObtain` from within the object's constructor if you want. So it allows for more configurability-- I guess. 

I'm not entirely sure what you're suggesting as far as your last bullet point. Are you saying remove `DescriptivePoolable` and just have the methods `onFree`/`onObtain` defined in `DescriptivePool` and have the implementation details open to subclassing `DescriptivePool`?

Darkyenus: Sorry for not being clearer. What I meant was, that when `Pool` implementation itself has `onObtain`/`onFree`, requiring the pooled objects to implement `DescriptivePoolable` is superfluous, because the implementation of whatever should happen `onObtain`/`onFree` can be implemented directly in the `Pool` subclass. (This would, in turn, mean, that default `onObtain`/`onFree` implementation in the `Pool` must do `instanceof` checks, which is a small price to pay for that flexibility I guess.)

And with the first part, I agree with you that not calling `onObtain` when the object is created may be beneficial, I think that this does not hold for `onFree`, because user may want to implement, for example, native resource disposal there. Then the game would work as long as the `Pool` operates within limits, but as soon as it starts throwing freed objects away, the game will leak resources, in a non-obvious way.

And if `onFree` is called always, it feels right (this is subjective, I know) to have it symmetrical with `onObtain`. (I suspect that many types of pooled objects won't need any explicit constructor at all in that case.) And when both `onFree` and `onObtain` are called always, it is possible to implement that functionality in `DescriptivePool` without changing `Pool` itself, which is a nice benefit, because then there is no need to worry about breaking it.

In the end, those are just my opinions and the final decision is on you and the owners.

Ktar5: Hmmmmmmm..... I see where you're coming from and I guess you're quite right. I agree now with onObtain and onFree. Lemme make those changes

Ktar5:
@Darkyenus Looks like, once again, you had a great idea and it ended up prettier :)

gdx/src/com/badlogic/gdx/utils/DescriptivePool.java
```diff
@@ -0,0 +1,72 @@
...
+			object.onFree();
+		}
+	}
+
+
+    	/** Objects implementing this interface will have {@link #reset()} called when passed to {@link Pool#free(Object)}. */
+    	static public interface DescriptivePoolable extends Poolable {
+		/** Called ONLY when an object is added to the pool.
+      	   	* Note: Will call after {@link #reset()} in {@link Pool#free(Object)} related methods */
+        	public void onFree ();
```
Darkyenus: Now the `onFree` is called always, so the documentation is wrong again, sorry for never-ending stream of comments  ☹️

(Also, now that I am looking at the code, isn't `onFree` just a duplicate of `reset`? It seems to be called in the same conditions... 😑)

Ktar5: Reset is a protected class method, so we don't really know if someone creates a subclass and calls it, and I don't really want those two mixed.

MobiDevelop:
Is it just me or is "DescriptivePool" not a very descriptive name for this? What makes it descriptive?

Ktar5:
@MobiDevelop I chose it because it gives more description to the Poolable object on whats happening (Free / Obtain)

Ktar5:
@MobiDevelop Yeah, you're definitely right now that I think about it some more. I've come up with two other possible names: CallbackPool and EventPool. How do you feel about those? As well, if you have any suggestions personally, I'm up to hear!

NathanSweet:
We already have `Poolable` which has `reset`, called when an object is freed. That is identical to this PR's `onFree`. The other thing this PR adds is `onObtain`. Do objects need to know when they were obtained? What would be a use case? Is it common enough to warrant a specialized type of pool? Does this make more sense as application code?

Ktar5:
@NathanSweet `Poolable#reset` is a protected method that can be overriden and thus its behavior cannot be reliably predicted. The main use case I found for this is for an object to know if it is active or inactive, which most definitely is useful.

NathanSweet:
@Ktar5 DescriptivePoolable `onFree` is public and can be overridden (it's an interface after all). I don't follow how that is any different from Poolable `reset`, except that it is public rather than protected.

It's very common to need to clean up state when an object is no longer used, hence Poolable `reset`. It's also very common to need to configure an object obtained from a pool for a particular use. However, such configuration usually always requires data from outside the object, eg:
```
Thing thing = pool.obtain();
thing.setPosition(x, y);
thing.setColor(...);
// Other application specific configuration for this particular thing instance.
```
For your use case, it is trivial to set the state, eg `thing.setActive(true);`. Even if the object wanted to do something like resource loading, it's easy enough to do `thing.loadResources();` after `obtain`.

I have a feeling that `onObtain` is not useful enough in most cases and is easily done with application code like the above rather than adding a new type of pool to libgdx. It might have made sense to add it to the Poolable interface, but doing that now would require adding the new method to a lot of existing code and isn't worth the trouble.

Ktar5:
@NathanSweet Oops, I meant to say `Pool#reset()` is protected and can be overriden. Because that is what's called, as ooposed to direct `Poolable#reset()`.

But I definitely see where you're coming from. It is quite an edge case. But I do feel a portion of people would find it useful.

The problem with writing it in application code is doing `thing.setActive(false);` currently I can't think of any way to do that without the `DescriptivePool` class.

obigu:
@Ktar5 I don't see the need for this either, why can't you just implement `Poolable#reset()` and call setActive(false) there?

Ktar5:
@obigu Because `Pool#reset()` (which calls `Poolable#reset()` ) is protected. So it could be randomly called by a subclass, or not call `Poolable#reset()` at all-- thus producing unexpected behavior.

obigu:
Pool subclasses will mostly always implement the only abstract method on `Pool `which is `newObject`. True, you might use a subclass you don't know the implementation details about and not have reset called if that implementation docs doesn't specify it. But following that logic you could say the same for all the `Pool `public methods which can be overriden aswell and cause unexpected behaviour. I'm sorry but is this a purely academic discussion or you have a real issue/use case that can't currenlty be implemented?

NathanSweet:
If someone wants to do something bad like call `reset` at odd times or never at all, that can't be avoided. It's not worth worry about, there are so many other terrible things someone can do, they deserve the consequences.

The rest I think can be implemented in application code, so I think we can close this issue.

PR #5074, opened 2018-02-02
Use StringBuilder instead of StringBuffer as it offers high performan…

reudismam:
…ce in single thread context as it is generally the case.

Other pull request in this project perform the same edit.
https://github.com/libgdx/libgdx/pull/5063

Other edits:
 - maintain the consistency of variable names whenever possible.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/SerializationException.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/SerializationException.java
@@ -19,7 +19,7 @@ package com.badlogic.gdx.utils;
 /** Indicates an error during serialization due to misconfiguration or during deserialization due to invalid input data.
  * @author Nathan Sweet */
 public class SerializationException extends RuntimeException {
-	private StringBuffer trace;
+	private StringBuilder trace;
 
 	public SerializationException () {
 		super();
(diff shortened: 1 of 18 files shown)


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/SerializationException.java
```diff
@@ -19,7 +19,7 @@
 /** Indicates an error during serialization due to misconfiguration or during deserialization due to invalid input data.
  * @author Nathan Sweet */
 public class SerializationException extends RuntimeException {
-	private StringBuffer trace;
+	private StringBuilder trace;
```
obigu: I can't think of a scenario with the Exception being shared across threads but I'd be cautious with this one just in case

MobiDevelop: Realistically, all of the GWT code is single-thread anyway (the implementations of [StringBuffer](https://gwt.googlesource.com/gwt/+/2.8.2/user/super/com/google/gwt/emul/java/lang/StringBuffer.java) and [StringBuilder](https://gwt.googlesource.com/gwt/+/2.8.2/user/super/com/google/gwt/emul/java/lang/StringBuilder.java) are exactly the same in GWT) so these types of changes in the GWT backend code would only be for consistency purposes.

obigu: Cool, thanks for the explanation. With this one reviewed, the rest of the changes are safe imo, do you agree @MobiDevelop?

MobiDevelop: Yeah, I don't see anything wrong with changing any of these to StringBuilder.

NathanSweet:
I wouldn't have changed `buffer` to `sb` (acronym variable names are not descriptive, eg `baos` for ByteArrayOutputStream, etc), otherwise looks OK.

PR #5077, opened 2018-02-03
Support emissive texture and color in default shader

eobermuhlner:
As proposed in the libgdx contributions forum I have implemented support for emissive texture and color in the default shader.

The change is trivial.

A test class was also added (I am not sure about the naming conventions of test classes, if there is any).


eobermuhlner:
Further testing has shown a missing change:
default.vertex.glsl should start with:

```
#if defined(diffuseTextureFlag) || defined(specularTextureFlag) || defined(emissiveTextureFlag)
#define textureFlag
```

Unfortunately I do not know whether I can modify this pull request (probably not) or whether I would need to create a new pull request.

obigu:
You can squash all changes into a single commit following this https://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

Tom-Ski:
Looks great, thanks!

PR #5081, opened 2018-02-07
Fix issue where Android wouldn't detect bluetooth

EtK2000:
keyboard connected before app launch.

I'm not sure why it's saying I changed the whole file...
All I did was remove the unused imports and add in lines 199-201.


mgsx-dev:
Your git problem seams related to CR/LF conflicts with original, see this diff https://github.com/libgdx/libgdx/pull/5081/commits/2a48be92508932c7be5396176e97fa484f161f0e?w=1 where "w=1" means ignore whitespace.

I suggest to check your local git configuration and/or your editor(s) settings to check if it doesn't automatically format files with CR+LF somewhere.

EtK2000:
I tried changing it to LF but it still doesn't seem to work...

EtK2000:
OK.
I changed it to LF and it seems to work.
Thanks for the tip!

mgsx-dev:
So now you need to squash your 2 commmits into one to keep a clean history as explained [here](https://gist.github.com/patik/b8a9dc5cd356f9f6f980)

Tom-Ski:
Github allows us to squash now as part of the merge process, so not required anymore

PR #5087, opened 2018-02-11
Fix PropertiesUtils space handling during output

toyknight:
Currently after PropertiesUtils.store, all spaces in values are getting removed.


Tom-Ski:
Thanks!

PR #5088, opened 2018-02-12
Updated a couple of mismatched version numbers

Zomby2D:


tests/gdx-tests-android/AndroidManifest.xml
```diff
@@ -6,8 +6,7 @@
 	android:versionCode="1"
 	android:versionName="1.0"
 	android:installLocation="preferExternal">
-	<uses-sdk android:minSdkVersion="9" android:targetSdkVersion="19"
-		tools:overrideLibrary="android.support.v4, android.support.compat, android.support.mediacompat, android.support.coreutils, android.support.coreui, android.support.fragment" />
```
Tom-Ski: Why have you removed these?

Zomby2D: The v26 support libraries that were automatically fetched by gradle requires an min SDK of 14. By limiting to the v25 libraries, which are actually compatible with SDK 9, we don't need to override them to force the app to compile.

PR #5092, opened 2018-02-12
Replaced Math calls by MathUtils calls in Vector2, Vector3 and Matrix3

WinterAlexander:
Replaced Math.atan2, Math.cos and Math.sin calls for corresponding MathUtils calls


Darkyenus:
IIRC `Math` instead of `MathUtils` is used here intentionally, for better precision.

WinterAlexander:
@Darkyenus Interesting. Why making math functions that are not precise enough for game development?

badlogic:
Different contexts require different precision/speed trade-offs.

PR #5098, opened 2018-02-19
Fixes #5070 - Fixed and replaced Intersector intersectSegmentCircleDisplace

### LINKED ISSUE #5070: intersectSegmentCircleDisplace does not work when the segment is partially inside the circle
#### Issue details
The [Intersector.intersectSegmentCircleDisplace()](https://github.com/libgdx/libgdx/blob/9b8f20b2d0e96e53f0f2b98dc8f6131c810aae71/gdx/src/com/badlogic/gdx/math/Intersector.java#L286
) function is wrong when one of the endpoint of the segment is inside the circle.
bitmap
The current algorithm calcultates the projection of the circle center on the infinite line (start, end), but returns no intersection if the projection point is not within the segment endpoints (as in the diagram above)

Also, the function is supposed to "returns by how much and in what direction the line has to move away from the circle to not intersect." but it returns by how much and in what direction the line has to move to cross the circle center.

Here's my correction: 

```
	public static float intersectSegmentCircleDisplace(Vector2 start, Vector2 end, Vector2 point, float radius, Vector2 displacement) {
		float u = (point.x - start.x) * (end.x - start.x) + (point.y - start.y) * (end.y - start.y);
		float d = start.dst(end);
		u /= d * d;
		if (u < 0)
			tmp2.set(start);
		else if (u > 1)
			tmp2.set(end);
		else {
			tmp.set(end.x, end.y).sub(start.x, start.y);
			tmp2.set(start.x, start.y).add(tmp.scl(u));
		}
		d = tmp2.dst(point.x, point.y);
		if (d < radius) {
			displacement.set(tmp2).sub(point).nor();
			return radius - d;
		} else
			return Float.POSITIVE_INFINITY;
	}
```

#### Reproduction steps/code
`System.out.println(Intersector.intersectSegmentCircleDisplace(new Vector2(0,0), new Vector2(0,1), new Vector2(0,2),1.5f, new Vector2()));
`
returns Infinity instead of 0.5

1.9.6

- [X] Android
- [X] iOS (robovm)
- [X] iOS (MOE)
- [X] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

obigu:
Fixes #5070

`intersectSegmentCircleDisplace` has been replaced by an overloaded method of `intersectSegmentCircle` that now accepts a `MinimumTranslationVector` as optional parameter.

Some design choices justifications
1. Use Vector2 instead of Vector3. This has required creating some static Vector2 instances for temporary use similar to Vector3 ones. To keep simple naming prefix v2 + letter seems the more compact/readable. These instances can be used by any other method in the class that needs temporary Vector2.
2. Use Vector2 operations when possible for readability
3. Use Circle as argument (the old one didn't use it as Circle class was created later in time)

**Even if the new `intersectSegmentCircle` that accepts an MTV parameter could also replace the old one, it has been left for backwards compatibility (up for discussion, in my opinion we should remove it)**

I will squash all the changes when I get confirmation on what to do with the old method + code review.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -301,8 +301,19 @@ public final class Intersector {
         v2a.set(v2c.x - circle.x, v2c.y - circle.y);
 
         if (mtv != null) {
-            mtv.normal.set(v2a).nor();
-            mtv.depth = circle.radius - v2a.len();
+			// Handle special case of segment containing circle center
+			if (v2a.equals(Vector2.Zero)) {
+				v2d.set(end.y - circle.y, circle.x - end.x);
+				// Handle special case of segment end point being circle center
+				if (v2d.equals(Vector2.Zero)) {
+					v2d.set(start.y - circle.y, circle.x - start.x );
+				}
+				mtv.normal.set(v2d).nor();
+				mtv.depth = circle.radius;
+			} else {
+				mtv.normal.set(v2a).nor();
+				mtv.depth = circle.radius - v2a.len();
+			}
         }
 
         return v2a.len2() <= circle.radius * circle.radius;
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -16,15 +16,14 @@
 
 package com.badlogic.gdx.math;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


FlorentMasson:
As you're using static fields as temp variables, I would write down in the function description that it is not thread-safe

obigu:
From libGDX docs https://github.com/libgdx/libgdx/wiki/Threading 
> No class in libgdx is thread-safe unless explicitly marked as thread-safe in the class documentation.

PR #5108, opened 2018-02-25
Fixed the missingGlyph of FreeTypeBitmapFontData doesn't initialized.

qinxiandiqi:
Fixed the missingGlyph of FreeTypeBitmapFontData doesn't initialized.
If the missingGlyph doesn't initialized, any glyph which not found will not rendered by the default glyph.


NathanSweet:
Cheers!

PR #5116, opened 2018-03-06
Update README.md

edthegamer:
Fix to NextPeer link.


intrigus:
Nextpeer seems to be dead, so I'd just completely remove it.

PR #5118, opened 2018-03-06
Vector2.java  rotateAround 

ahmetdenizyilmaz:
rotateAround / rotateAroundRad methods can be usefull


gdx/src/com/badlogic/gdx/math/Vector2.java
```diff
@@ -362,6 +362,14 @@ public Vector2 setAngleRad (float radians) {
...
 		return rotateRad(degrees * MathUtils.degreesToRadians);
 	}
+	/** Rotates the Vector2 by the given angle around reference vector, counter-clockwise assuming the y-axis points up.
+	 * @param degrees the angle in degrees
+	 * @param reference center Vector2
+	 */
+	public Vector2 rotateAround (Vector2 reference,float degrees) {
+		return this.sub(reference).rotate(degrees).add(reference);
+	}
+
```
yuripourre: This whitespace could be moved to line 365

PR #5123, opened 2018-03-08
Adds InstanceBufferObject and InstanceBufferObjectSubData to enable instanced rendering

mrdlink:
Adds InstanceBufferObject and InstanceBufferObjectSubData to enable instanced rendering.
Requires GLES3+ context. Must be used with VertexBufferObjectWithVAO.
Offers the same interface as VertexBufferObject and VertexBufferObjectSubData.

Both classes are tested for indexed and non-indexed rendering. There may be a lack of configuring the glVertexAttribDivisor, because it is set to 1 for each VertexAttribute and can not be changed.

### DIFF
--- a/gdx/build.gradle
+++ b/gdx/build.gradle
@@ -29,6 +29,7 @@ compileJava {
 }
 
 dependencies {
+    compileOnly "org.joml:joml:1.9.8"
     compileOnly project(":extensions:gdx-jnigen")
     testCompile libraries.junit
 }
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,8 @@
 - SkinLoader can be used on subclasses of Skin by overriding generateSkin(). 
 - API addition: Tree indentation can be customized.
 - Fixed GlyphLayout not respecting BitmapFontData#down.
+- Adds new InstanceBufferObject and InstanceBufferObjectSubData classes to enable instanced rendering.
+- Adds methods to enable instanced rendering and set per instance data in a Mesh.
 
 [1.9.8]
 - Add iPhoneX images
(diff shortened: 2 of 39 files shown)


intrigus:
Do you know if this would work with WebGL2?
(I know that libgdx currently doesn't have WebGL2 support iirc)

mrdlink:
If you can't create a WebGL2 context with libgdx, then it won't work. Instanced rendering is not available in WebGL1.0/GLES2.

intrigus:
But It would work if libgdx supported webgl2 right? 
Webgl2 is like GLES3 but there are always corner cases that's why I'm asking.

mrdlink:
Yes it would work with WebGL2. The [API](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext) clearly has all functions included for instanced rendering.

gdx/build.gradle
```diff
@@ -29,6 +29,7 @@ compileJava {
 }
 
 dependencies {
+    compileOnly "org.joml:joml:1.9.8"
```
Tom-Ski: Left in by mistake?

Tom-Ski:
Can you provide a test for this please?

gdx/build.gradle
mrdlink: Sorry, yes. I tried to experiment with JOML. I'll delete the dependency.

mrdlink:
Should I just create a JUnit test, or in what form should that test be?

Tomski <notifications@github.com> schrieb am Di., 27. März 2018, 22:11:

Tom-Ski:
Thanks for the contributions! This has been updated slightly and manually merged.

PR #5131, opened 2018-03-13
 Added Input.getMaxPointers(), fixes #4977 and update working from source link in README

### LINKED ISSUE #4977: There's no way to get the maximum touch count from Input
Current values:
 * AndroidInput: `public static final int NUM_TOUCHES = 20;`
 * IOSInput: `static final int MAX_TOUCHES = 20;`
 * LwjglInput: implicitly `1` via eg. `if (pointer > 0) return false;`
 * MockInput: implicitly `0` via eg. `return false`
 * Lwjgl3Input: implicitly `1` via eg. `return pointer == 0? isTouched(): false;`
 * ...

I would be great to bring this into the `Input` interface, something like `int getMaxTouches()`, so we can loop over all of the possible pointers and check whether they're touched down and what their coordinates are.

flaghacker:


EtK2000:
Something to keep in mind, some platforms using the desktop backend support more than one cursor.
I don't know if that's physically handled by LibGDX atm, but it's possible.

flaghacker:
I was careful to check that the returned values are always correct, they're pretty much hardwired into all of the Input implementations I encountered. Take a look at the examples given in #4977 to get an idea of what I mean.

EtK2000:
Everything else seems fine to me, I guess we wait for the merge now.
There must be a reason not to support multiple cursors I'm unaware of.

Tom-Ski:
Looks good. Can you please run the LibGDX formatter provider, on only the lines you have changed.

flaghacker:
I tried importing the `eclipse-formatter.xml` file into InteliJ, but now it wants to indent everything with a tab _followed by_  a space.

Are there any specific things wrong with the formatting? Installing and learning eclipse is a bit of a stretch to format 60 lines of code, I can go trough them manually.

Tom-Ski:
I use the formatter in intellij and its fine.  
You need a space between method name and parenthesis. 
`getMaxPointers()` should be `getMaxPointers ()`

PR #5144, opened 2018-03-25
Use UTF8 encoding for strings NetJavaImpl

PokeMMO:
Android uses UTF8 by default, and is a saner default than leaving it up to environment's default character set for desktop.

This could be fleshed out to add specifiable character set to requests, and to attempt to parse response character set from response headers.

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
+++ b/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
@@ -175,7 +175,7 @@ public class NetJavaImpl {
 							// we probably need to use the content as stream here instead of using it as a string.
 							String contentAsString = httpRequest.getContent();
 							if (contentAsString != null) {
-								OutputStreamWriter writer = new OutputStreamWriter(connection.getOutputStream());
+								OutputStreamWriter writer = new OutputStreamWriter(connection.getOutputStream(), "UTF8");
 								try {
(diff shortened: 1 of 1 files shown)


noblemaster:
+1 to have this added. Haven't had any problems as far as I know, but the default constructor uses a system's default charset and is platform-dependent if I was reading the documentation of the classes correctly? Makes sense to make this UTF-8 in my opinion!?

Others?

PokeMMO:
>but the default constructor uses a system's default charset and is platform-dependent if I was reading the documentation of the classes correctly?

Yeah. Unless you override the file.encoding property as a startup property, Desktop users will post/parse responses with an unpredictable character encoding. This is a pretty large pitfall that can only be worked around via the startup property or base64 encoding/decoding before interacting with NetJavaImpl.

If we wanted to prevent any issues from people relying on this, we could add an optional parameter in HttpRequest which could be used to specify an encoding.

Darkyenus:
I can't imagine anyone relying on sending unpredictably broken data. Having an optional argument to specify encoding explicitly would be nice, but I wouldn't add it unless someone explicitly asks for it.

MobiDevelop:
I'm sure someone will post an issue if any arise out of this change, better to have a consistent default.

Monarchis:
My Opinion:
* Consitent default: yes
* Option to cinfigure: yes

Issue #5137, opened 2018-03-19
isPeriphalAvailable HardwareKeyboard on GWT


MrStahlfelge:
#### Issue details
If isPeriphalAvailable for HardwareKeyboard is called on GWT backend, it just returns true.

#### Reproduction steps/code
https://github.com/libgdx/libgdx/blob/2cae3f5080bb00414fafb2ac3bd7035ab45427c5/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java#L286

Unfortunately, there is no way to detect if a hardware keyboard is connected with Javascript. However, I think we could come nearer to the truth if we detect if a mobile browser is running. [This can be done quite easy.](https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser)

If the maintainers like the idea, I would be happy to contribute.

#### Version of LibGDX and/or relevant dependencies
All

#### Please select the affected platforms
- [X] HTML/GWT

Tom-Ski:
Sure

EtK2000:
Please close this issue, this has been merged.

PR #5148, opened 2018-03-28
#5137 isPeriphalAvailable HardwareKeyboard on GWT

MrStahlfelge:
At this point of time, it is not possible to detect if a hardware keyboard is connected on GWT. So we assume that a mobile device has a soft keyboard while a desktop device has a hardware keyboard.


MobiDevelop:
Seems like a reasonable default.

PR #5149, opened 2018-03-28
GWT: AssetFilter changed to match Android build and file system restrictions

MrStahlfelge:
Fixed two issues with DefaultAssetFilter:

DefaultAssetFilter accepts files beginning with beginning full stops (for example .gitignore) and adds this files to the assets.txt. However, these files are (under Windows) not copied to war/assets leading to a JavaScript warning.

DefaultAssetFilter accepts files and folders beginning with an underscore in contrast to the Android build process were these files are not omitted when packaging the Android apk.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/DefaultAssetFilter.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/DefaultAssetFilter.java
@@ -26,6 +26,9 @@ public class DefaultAssetFilter implements AssetFilter {
 
 	@Override
 	public boolean accept (String file, boolean isDirectory) {
+		String normFile = file.replace('\\', '/');
+		if (normFile.contains("/.")) return false;
+		if (normFile.contains("/_")) return false;
 		if (isDirectory && file.endsWith(".svn")) return false;
 		return true;
 	}


intrigus:
Are you sure that windows doesn't copy **all** dot-files instead of only those listed [here](https://github.com/apache/ant/blob/ANT_193/src/main/org/apache/tools/ant/DirectoryScanner.java#L148)

Likewise there seems to be a workaround for android:
https://stackoverflow.com/questions/9206117/how-to-workaround-autoomitting-fiiles-folders-starting-with-underscore-in

And yet another question, but couldn't you do
````java
		String normFile = file;
		if (normFile.contains("\\.")) return false;
		if (normFile.contains("\\_")) return false;
````

instead of
````java
		String normFile = file.replace('\\', '/');
		if (normFile.contains("/.")) return false;
		if (normFile.contains("/_")) return false;
````

?

MrStahlfelge:
Interesting some see this as an issue on Android, for me it is a feature. I have seen several projects using folders in the assets directory starting with an underscore with the intention that it is not copied. I adapted that and use underscore folders for unpacked skins, drawables or marketing material in my projects. So I hope if they "fix" the Android build they make this configurable.

For Windows, your changed code would work, but I assume it would not work on Unix and Mac. That would lead to differing builds depending of the operating systems.

To make it clear, you can always override this AssetFilter if you don't like the behaviour. As a default filter, I am convinced one that resembles the Android build filter makes the most sense especially for new users.

PR #5151, opened 2018-03-29
GWT: Default localization depending on user language

MrStahlfelge:
This commit gives I18NBundles on GWT backend the ability to switch to the user language returned from the browser by default, matching the behaviour on other platforms.

For testing, use Firefox and change the preferred language settings or use Chrome on Android and change the system language. (Chrome on desktop always returns the language of the operating system)

I made also a cleanup of the class, deleting all the lines that are just commented out for four years.


intrigus:
Did you check whether gwt now supports the commented out code?

MrStahlfelge:
No. These parts are not necessary for libGDX GWT backend to work.

smelc:
This is a damn good PR, it should be merged. @Tom-Ski ?

PR #5153, opened 2018-03-31
Overload AssetManager#isLoaded and #finishLoadingAsset to take AssetDescriptor as parameter

ShipshapeDetour:
AssetManager#isLoaded and AssetManager#finishLoadingAsset normally only take a String for the filename of the asset.

If you want to check if an asset is loaded or you want the AssetManager to block for it to finish loading using an AssetDescriptor, you must use AssetDescriptor.fileName for these methods.

This change adds overloaded methods for both isLoaded and finishLoadingAsset to take an AssetDescriptor for convenience.

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -262,6 +262,12 @@ public class AssetManager implements Disposable {
 		return null;
 	}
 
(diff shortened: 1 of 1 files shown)


NathanSweet:
It's not much different (or more convenient, really), but the overloads are typed which does make it a little more clear about what is being passed (versus any old string).

PR #5155, opened 2018-03-31
 Distance between center points of circle and rectangle

giraybal:
I added minor and useful method for calculate distance between center points of circle and rectangle


obigu:
I don't think it's a good idea for 2 reasons:
1. It adds dependencies between classes
2. It makes the API larger for a not that useful to justify simple functionality 

I think best approach is to add this kind of simple methods to a separate helper class in your project.

PR #5156, opened 2018-04-01
Fix issues with FileHandle's handling of extensions

JerBear3:
When calling FileHandle.extension() on "multi-part" file extensions (such as .tar.gz, .tar.bz2, etc) currently libgdx only returns the very end (gz, bz2) and doesn't include the first part (tar). This was creating a mess when searching for .tar.gz files.

There was also a problem with pathWithoutExtension() - paths with a dot in one of the parent directories but no actual file extension got truncated. (For example, "/adirectorywitha.dot/myfile" was truncated to "/adirectorywitha" instead of returning the original "/adirectorywitha.dot/myfile")

This PR fixes both issues.

### DIFF
--- a/gdx/src/com/badlogic/gdx/files/FileHandle.java
+++ b/gdx/src/com/badlogic/gdx/files/FileHandle.java
@@ -94,7 +94,7 @@ public class FileHandle {
 
 	public String extension () {
 		String name = file.getName();
-		int dotIndex = name.lastIndexOf('.');
+		int dotIndex = name.indexOf('.');
 		if (dotIndex == -1) return "";
 		return name.substring(dotIndex + 1);
 	}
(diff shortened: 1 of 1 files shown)


Darkyenus:
I disagree with the first fix. While "file.tar.gz" technically has multiple extensions, `.` character is not used only for extensions and may appear in the file name arbitrarily. Changing the behavior as proposed will most likely break people's code in subtle ways. Also, the method is called `extension` not `extensions` so I think that it should not return multiple extensions but just the most important one.

Regarding the second fix, nice catch! Though it might be faster and more readable to implement it with a second `indexOf` check, instead of this approach, which ~doubles the amount of allocations.

JerBear3:
You make a good point about the extension() method, but I think something has to be done because it currently breaks my own code in a subtle way. Possible a new extensions() method as you mentioned?

Darkyenus:
I think that this is too niche for libGDX, games usually don't do anything fancy with multiple extensions. Why don't you just rewrite your code to something like `file.name().endsWith(".tar.gz")`?

Tom-Ski:
Agree with @Darkyenus here. Feel free to resubmit this PR with your second fix.

PR #5163, opened 2018-04-06
gdx-controllers: LWJGL3 remove a redundant controller scan

WickedShell:
Every call to `pollState()` will scan the list of controllers and allocate a new one if needed already.


code-disaster:
Technically that's true, but there's now a gap between initialization of the manager (done lazily whenever a function in `Controllers` is called) and execution of the first runnable. So during game initialization, some code like `... = Controllers.getControllers()` would not work.

I believe it should be fine to just call pollState() once more, inside the constructor.

WickedShell:
There is already a gap in finding the controller the first time and actually polling the state though (IE you start with a controller instance that has never polled it's buttons). If we wanted that to just directly call pollState() and allow that to create the runnable sequence that would make sense and would avoid the lag on state/controllers being present.

code-disaster:
I _Marty McFly'd_ this PR into `master` by cherry-picking.

Issue #5134, opened 2018-03-16
[HTML/GWT] Input.isKeyPressed Always TRUE If Release Occurs Outside of LibGDX Context


villain-bryan:
#### Issue details
While in a LibGDX environment if you press down a key and then switch to any window outside of LibGDX's scope and release that key, the LibGDX environment does not detect that release event. 

Upon re-entry to the LibGDX environment, there should probably be some re-polling evaluation of the keys LibGDX thinks are pressed keys to determine if they are still pressed or not.

It is worth noting that if you press down a key outside of the LibGDX environment and then go into the LibGDX environment, the keyDown is detected. So it appears to be just a bug on re-entry for keys that were pressed upon leaving the context since the release is happening outside of the environment.

#### Reproduction steps/code
If you ALT+TAB to a window outside of the browser, LibGDX thinks that ALT (and probably the TAB) is still pressed when you return to the LibGDX context. This really can make some keybinds feel sticky and broken.

**This also makes for some very weird behavior with TextField (and probably heaps of other components)**. For the CONTROL_LEFT pressed down case, TextField CONTROL+A and CONTROL+Z offers specialized functionality, so if you press down CONTROL_LEFT, click another window (lose focus on the LibGDX browser), release CONTROL_LEFT, return to LibGDX context, Gdx.input.isKeyPressed(Input.Keys.CONTROL_LEFT) will return true until you just press/release the bugged key in the LibGDX context.

**DEMO: https://arenaofkings.com/demo/keydownbug.mp4**
Relevant Code: https://hastebin.com/aqubejebuj.cpp

#### Version of LibGDX and/or relevant dependencies
1.9.8

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

smelc:
I'm affected on libgdx 1.9.6 too.

bardiharborow:
Fixed by #5167?

kingtestv587:
I'm affected on libgdx 1.9.6. It happens with chrome while it works fine with firefox. 华途

Sholanki:
I am affected on libgdx 1.9.6 on Chrome browser

mzpiggeeup1:
> #### Issue details
> While in a LibGDX environment if you press down a key and then switch to any window outside of LibGDX's scope and release that key, the LibGDX environment does not detect that release event.
[...]

PR #5167, opened 2018-04-06
Fix issue #5134

allen-chin:
Maintains an IntSet of pressed keys to clear when the application or browser loses focus.

This is my first (attempted) contribution and pull request so please let me know about any issues or things to improve. I have sent a signed CLA to contact@badlogicgames.com. Thanks!


Tom-Ski:
@allen-chin what browsers has this been tested on?

villain-bryan:
Can confirm #5167 fixes the bug for our Chrome users (I tested it on Version 67.0.3396.99 (Official Build) (64-bit)). Multiple people on our end will be testing with this fix over the next days - I'll report back if there is any issues for us. Thanks @allen-chin.

PR #5170, opened 2018-04-08
Set correct curve to "fastSlow" interpolation

obigu:
Fixes silly mistake in recent commit https://github.com/libgdx/libgdx/commit/e8cbe06c3cfa139ace078a5e97480c09aec4bfec


obigu:
@intrigus yep

NathanSweet:
My commit was so simple, what could possibly go wrong? I better not even read it before committing! ;)

PR #5171, opened 2018-04-09
Update Intersector.intersectPolygons()

anigart:
Update to intersectPolygons:
- Move initial zero length vertices check up (minor optimization)
- Prevent crash when overlap polygon is null
- Update Javadoc to reflect that vertices need to be clockwise and the polygons need to be convex

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -149,7 +149,7 @@ public final class Intersector {
 	private final static Vector2 s = new Vector2();
 	private final static Vector2 e = new Vector2();
 
-	/** Intersects two resulting polygons with the same winding and sets the overlap polygon resulting from the intersection.
+	/** Intersects two polygons with clockwise vertices and sets the overlap polygon resulting from the intersection.
 	 * Follows the Sutherland-Hodgman algorithm.
 	 *
 	 * @param p1 The polygon that is being clipped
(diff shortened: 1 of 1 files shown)


NathanSweet:
Are you sure it doesn't work if they are both counterclockwise? Can you please apply the libgdx source formatter? (or manually format it to match, since the changes are small)

anigart:
Yes, see this test:

// counter-clockwise
float[] vertices = new float[] {1, 1, 3, 1, 2, 2};
Polygon p1 = new Polygon(vertices);
Polygon p2 = new Polygon(vertices);
System.out.println("same triangles overlap: " + Intersector.intersectPolygons(p1, p2, new Polygon()));

float[] vertices2 = new float[] {1, 1, 3, 1, 2, 3};
Polygon p3 = new Polygon(vertices2);
System.out.println("different triangles overlap: " + Intersector.intersectPolygons(p1, p3, new Polygon()));

Output:
>> same triangles overlap: false
>> different triangles overlap: false

// clockwise
float[] vertices3 = new float[] {1, 1, 2, 2, 3, 1};
Polygon p4 = new Polygon(vertices3);
Polygon p5 = new Polygon(vertices3);
System.out.println("same triangles overlap: " + Intersector.intersectPolygons(p4, p5, new Polygon()));

float[] vertices4 = new float[] {1, 1, 2, 3, 3, 1};
Polygon p6 = new Polygon(vertices4);
System.out.println("different triangles overlap: " + Intersector.intersectPolygons(p4, p6, new Polygon()));

Output:
>> same triangles overlap: true
>> different triangles overlap: true

Will re-format.

NathanSweet:
Thanks for verifying. Odd someone spent the effort to specify the wrong info! But then `overlap` wasn't really optional either.

While we are here, maybe we could improve things. Passing in a container for an output variable is usually to avoid allocation. We could at least avoid allocation if the `overlap` vertices are already the right size. Also it would be nice if the `overlap` javadoc said explicitly that the parameter can be null, rather than just "optional". Would you care to make these changes?

Cheers!

NathanSweet:
No worries, I merged and made the change I had in mind. Cheers!

anigart:
Got it now - yeah, did not see that one :)

PR #5172, opened 2018-04-11
Stack data structure

georgwarden:
## Reasoning
A short pre-story: so I was writing something with LibGDX and suddenly felt an urgency to use a stack. There was no such thing in LibGDX API, and Stack implementation from Java library wasn't suitable for me for some reason. Thus, I had to write my own implementation, and only then, when it was too late, I got to understand that `com.badlogic.gdx.utils.Queue` provided deque functionality as well. It is not quite obvious that it should, along with that, it seems to be too complicated if you need *an old plain boring stack*.
## Description
There is the most canonical Stack in this PR, with only three most basic queries: to add, to remove, to look. I considered it as bad idea to overcomplicate it with additional features like `Iterator`, but it doesn't mean that I will not add it in the future. For now it is what it is - a data structure.
## ToDo
Whatever decision will be made about this particular PR, if I will get a positive feedback or will understand that it is a useful feature for the community (personally, for me it would be very handy to have it in gdx api) I would continue working on it. There should be implementations for primitives (ones that avoid boxing/unboxing) and it should be brought to a more Java-canonical look, with iterators and all.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Stack.java
+++ b/gdx/src/com/badlogic/gdx/utils/Stack.java
@@ -14,7 +14,7 @@ public class Stack<T> {
     public Stack(int initialSize) {
         this.items = (T[])new Object[initialSize];
         this.size = initialSize;
-        this.top = 0;
+        this.top = -1;
     }
 
     public void push(T element) throws IndexOutOfBoundsException {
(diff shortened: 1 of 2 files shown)


mgsx-dev:
Why not use LibGDX Array class ? It already provides add, pop, peek and clear features ...

georgwarden:
@mgsx-dev See, it is not obvious, too! Like, you don't expect that `Array` will provide you with stack features. Why not to tend to SRP?

MobiDevelop:
I'm inclined to agree that we don't need to add another way to do something we already have.

georgwarden:
Ok, then I'd tend to join the major and close PR then

PR #5188, opened 2018-04-25
GlyphLayout: wrap: remove whitespace at line end

TCreutzenberg:
Current behaviour:
GlyphLayout wrap does not remove spaces from line ends.
When the GlyphLayout wraps the text and halign is Align.right or Align.center the glyphs of all lines but the last line (which is fine as there are no space characters at the end) are not positioned where they should be:
Align.right: As the space character is still at the end of the line, the position is offset by xAdvanceOfSpaceCharacter from the very right where it should be
Align.center: As the space character is still at the end of the line, the position is offset by xAdvanceOfSpaceCharacter/2 from the very right where it should be

Fix:
Remove whitespace on wrap


TCreutzenberg:
Sorry, I just noticed that this does not fix the issue... I'll close this pull request and open a new one as soon as I fixed it.

PR #5189, opened 2018-04-25
Bugfix for GlyphLayout.wrap

TCreutzenberg:
GlyphLayout: wrap:
(Bug only visible with Align.left and wrapped lines)
On wrap the xAdvance of the seconds line should be 0.
The xAdvance of the second+ lines is aligned with that of the first line, so it should be 0.

In BitmapFontTest it can be observed: 

Adjust text to be Align.left, the top left text should look like this:

Sphinx of black
quartz, judge my
vow.

The second and third line are drawn 1 unit too far left because of:
GlyphLayout line 322:
xAdvances2.set(0, -glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft);

-glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft  = -(-19)*1 - 20 = -1

The second and third line start -1 units left of the first line.

It gets more obvious with bigger fontData.scaleX or not nearly equal glyphs2.first().xoffset and fontData.padLeft

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -319,7 +319,7 @@ public class GlyphLayout implements Poolable {
 			FloatArray xAdvances2 = first.xAdvances; // Starts with all the xAdvances.
 			xAdvances1.addAll(xAdvances2, 0, wrapIndex + 1);
 			xAdvances2.removeRange(1, wrapIndex); // Leave first entry to be overwritten by next line.
-			xAdvances2.set(0, -glyphs2.first().xoffset * fontData.scaleX - fontData.padLeft);
+			xAdvances2.set(0, 0);
 			first.xAdvances = xAdvances1;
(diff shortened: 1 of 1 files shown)


TCreutzenberg:
Updated initial pull request message to make it easier to reproduce the bug

TCreutzenberg:
xadvancebug
xadvancefixed

NathanSweet:
> Updated initial pull request message to make it easier to reproduce the bug

An [executable example](https://github.com/libgdx/libgdx/wiki/Getting-help#executable-example-code) would make it easy to reproduce. Since it was not provided, I had to make one. Here is a modified [BitmapFontTest](http://n4te.com/x/418-LSSr.txt). Move the mouse so the text wraps where you see the problem. I do not see a problem:



Consider this wrapping without your change:



And with your change:



Your change causes the `j` to be too far left.

The line of code you set to zero is aligning the glyph's pixel bounds so they start at the drawing position. The glyph pixels are not guaranteed to be to the right of the drawing position. Some fonts are more extreme with this than others.

PR #5192, opened 2018-04-25
 Put Batch calls in try-finally to allow recovery from thrown exceptions

EtK2000:
Basically this will reset the state of the Batch objects so that if the code handles the exception they can go back to rendering normally without the complications of trying to call .end() on both Batch objects.

Git seems to give me issues with squishing my fork...
Hopefully next time I submit a pull request I'll get that working,
in the meantime sorry for the inconvenience...

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
@@ -125,8 +125,11 @@ public class Stage extends InputAdapter implements Disposable {
 		Batch batch = this.batch;
 		batch.setProjectionMatrix(camera.combined);
 		batch.begin();
-		root.draw(batch, 1);
-		batch.end();
+		try {
+			root.draw(batch, 1);
+		} finally {
+			batch.end();
+		}
 
 		if (debug) drawDebug();
 	}
@@ -165,8 +168,11 @@ public class Stage extends InputAdapter implements Disposable {
 		Gdx.gl.glEnable(GL20.GL_BLEND);
 		debugShapes.setProjectionMatrix(viewport.getCamera().combined);
 		debugShapes.begin();
-		root.drawDebug(debugShapes);
-		debugShapes.end();
+		try {
+			root.drawDebug(debugShapes);
+		} finally {
+			debugShapes.end();
+		}
 	}
 
 	/** Disables debug on all actors recursively except the specified actor and any children. */


piotr-j:
Is that a good idea? If something throws an exception that you are not catching, its a bug.

EtK2000:
The exception will still be thrown.
All this does it allows the program to handle the exception and potentially recover.
For a developer nothing will change.

This is useful when some of the stage data is dynamic, if the exception is caught the faulty Actor can be removed and the program can continue as usual.
A prime example of this is if the colspan in Table is set incorrectly.

NathanSweet:
It seems unlikely that an exception not handled inside the stage is recoverable, but the change doesn't hurt.

EtK2000:
I can give you my exact situation if that sheds some light on an example.
Thanks for merging!

NathanSweet:
You can try. ;) Any situation that would throw an unhandled exception from Actor `draw` should handle the problem before an exception is thrown. Eg, if you configure a table (or any actor) incorrectly, you should avoid doing that instead of catching an exception. It may be that the program is still in a valid state most of the time, but this is not guaranteed and not something actors should have to worry about.

Tom-Ski:
This assumes that the stage's batch is begun all the time within the boundaries of draw, if this is not the case the exception is swalled up and you end up with batch end exception only, which isn't very nice.

EtK2000:
I guess end can be put into a try-catch.
Though I'm not sure why or how anyone would end the batch manually inside the stage.draw...

Tom-Ski:
Thats not the issue, the issue is that it swallows up the actual error that is unrelated to batch.begin/end.

Simply do this in some actor's draw:

endBatch
throw some exception
beginBatch //To maintain state of batch 

When exception is thrown, we get kicked out of stage draw, so batch is current ended, we see the batch.end error because we never got to beginBatch, and it swallows the exception we actually care about.

EtK2000:
Oh ok...
I'm a little confused why anyone would do that.
I'll try to find the best way to deal with that whilst still throwing the end exception if no exception was thrown.

Tom-Ski:
I gave you multiple use cases above. If you are asking about why someone would throw an uncaught exception, they wouldn't but it happens in development.

EtK2000:
Ya I meant the uncaught exception, I feel that's asking for trouble.

In any case, here's my idea:

Stage draw in try-catch-finally in finally we call end in try-catch.
We get the exception from catch or null otherwise.
If it isn't null we throw it (inside a gdx runtime exception) and if end throws an exception throw that too.

Tom-Ski:
Im talking about exceptions that aren't thrown intentionally, but that arise due to programming issues and mistakes. I'm not sure why that is asking for trouble, its an unavoidable part of development. 

However, it is certainly asking for trouble gobbling these exceptions up silently.

EtK2000:
Ok.

I'll submit a new pull request in the next few minutes.

EtK2000:
Should I have lastException a public field or should I make it private and add a better function?

NathanSweet:
Good point @Tom-Ski. I was mistaken that using finally had no side effects.

Since the change is not a small change with no side effects, I don't think we should merge either of the PRs. Any API call can throw RuntimeException and it is unreasonable to attempt to keep objects in a valid state if that happens. It doesn't make sense to attempt for just one case.

You can copy Stage#draw code and add try/finally, then call that code instead of Stage#draw. Or you can call Stage#draw and if an exception is thrown, check the state of the batch or anything else and try to restore it all to a usable state. Recovering from the exceptional state may not be possible in all situations though and so it is not feasible for it to be an officially supported feature.

PR #5198, opened 2018-04-26
GlyphLayout: Remove whitespace on line wrap

TCreutzenberg:
GlyphLayout:
On line wrap the trailing whitepsace is not deleted and so all lines except the last are shifted left (in case of Align.right or Align.center) by the whitespace xAdvance.
Expected: Automatic line wrapping gives the same result as "\n" newline. 

Screenshots from BitmapFontTest, changed color of rectangle in line 212 to GREEN.

Before (with Align.right):
beforeremovewhitespaceatend

With (one) whitespace removed at run end after wrap (with Align.right):
removewhitespaceatend

Why after wrap? Because then you can just use peek() and pop() on the arrays and not removeIndex() which has to move whole arrays around.

If you scale the font e.g. set font.getData().scaleX = 2; in line 162 of BitmapFontTest and use Align.left, then there is a bug which is fixed by my other pull request:
https://github.com/libgdx/libgdx/pull/5189

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -175,15 +175,8 @@ public class GlyphLayout implements Poolable {
 									run.width = 0;
 									width = Math.max(width, run.x);
 								} else {
-									next = wrap(fontData, run, glyphRunPool, wrapIndex, i);
+									next = wrap(fontData, run, glyphRunPool, halign, wrapIndex, i);
 									runs.add(next);
-
-									// Remove whitespace at the end of the run after wrap
-									if(fontData.isWhitespace((char)run.glyphs.peek().id)){
-										run.glyphs.pop();
-										float removedXAdvance = run.xAdvances.pop();
-										run.width -= removedXAdvance;
-									}
 
 									width = Math.max(width, run.x + run.width);
 								}
@@ -297,7 +290,7 @@ public class GlyphLayout implements Poolable {
 		glyphRunPool.free(truncateRun);
 	}
 
-	private GlyphRun wrap (BitmapFontData fontData, GlyphRun first, Pool<GlyphRun> glyphRunPool, int wrapIndex, int widthIndex) {
+	private GlyphRun wrap(BitmapFontData fontData, GlyphRun first, Pool<GlyphRun> glyphRunPool, int halign, int wrapIndex, int widthIndex) {
 		GlyphRun second = glyphRunPool.obtain();
 		second.color.set(first.color);
 		int glyphCount = first.glyphs.size;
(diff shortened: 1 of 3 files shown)


NathanSweet:
I agree there is a problem here. I also found a potentially related problem using left alignment which occurs even with your fix:



GlyphLayout used to have code to eat whitespace but it was removed in ee42693da067da7c5ddd747f051c1423d262cb96 for reasons.

TCreutzenberg:
As far as I can see TextArea's drawText has wrap set to false (i.e. it does not use GlyphLayout's wrapping). It uses a custom wrapping, so we are safe applying changes to GlyphLayout's wrapping without affecting TextArea.

I am trying out a version which also removes one whitespace at the front of the run. It works good, just in some edge cases it looks weird, but that has nothing to do with whitespaces but must be some other bug.

TCreutzenberg:
Check GlyphLayoutWrapTest to test fixes (see instructions at bottom of window).

The changes only remove a single whitespace at the beginning and/or the end.

At font scale 1 everything looks great, at font  scale 2 there are bugs. These are the bugs I tried to solve with https://github.com/libgdx/libgdx/pull/5189 but that didn't work. It has nothing to do with this pull request though...

TCreutzenberg:
Please check GlyphLayoutWrapTest. This looks correct to me.

When the centered text becomes really narrow "of" and "judge" sometimes jump a very little bit left/right depending on targetWidth. This also happens without my changes and seems to be another bug.

NathanSweet:
Ha, that's a great commit ID!

Ah, you're right about TextArea. It's not my class so I'm less familiar with it. It had [some issues](https://github.com/libgdx/libgdx/issues/5204) so I was digging around in it today. Upon closer look, #3234 was about someone writing their own textarea. I guess we can strip whitespace and they can avoid using wrapping, like TextArea.

Your changes in this PR are OK, but we can adjust the indices to skip whitespace before configuring the two runs in `wrap`. I did this and also found and fixed a myriad of other problems. It seems pretty solid now!

TCreutzenberg:
Haha, just woke up and had the exact same idea about the indices and just wanted to implement it myself 😄

NathanSweet:
:D It's good to have more eyes on this. It's a pretty hairy beast. I found a bunch of problems, like [this was wrong](https://github.com/libgdx/libgdx/commit/beeef537dd7f76c72e124a5be674cd0c3b2b76f6#diff-e5b62781008351f68aa52151db58f0d6R812) and [this was really wrong.](https://github.com/libgdx/libgdx/commit/beeef537dd7f76c72e124a5be674cd0c3b2b76f6#diff-e5b62781008351f68aa52151db58f0d6R880). Also did some clean up, used a wonderful labeled IF to avoid about 4 levels of indentation.

TCreutzenberg:
Ok, thanks for finding the other things, too, this should solve my other problems!
And the code is so much more readable and maintable  now. It took me ages to understand what was going wrong when I noticed there's something wrong and then it took me ages to find what was done where and how it all works together, much better now. It really is a hairy beast, but it is a very complex matter.

PR #5201, opened 2018-04-27
Support specifying which onscreen keyboard type to use

PokeMMO:
Is mainly useful for the ability to open the numeric/phone keyboards directly.

Current implementation adds Default, NumberPad, PhonePad, Email, Password, URI options which are mapped back to Android/iOS types.

Android Password/URI types do not visually change anything on my device, but I do not know if this is keyboard dependent. Password type may also prevent autocomplete data from learning passwords.

Instanceof checks for view in `AndroidInput` are ugly and might be able to be cleaned up with a shared interface for view implementations.

iOS code is untested. I am not sure if the current `setKeyboardType` call is sufficient enough to actually change the keyboard.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -674,7 +674,7 @@ public interface Input {
 	/** Sets the on-screen keyboard visible if available.
 	 * 
 	 * @param visible visible or not */
-	public void setOnscreenKeyboardVisible (boolean visible);
+	public void setOnscreenKeyboardVisible (boolean visible, OnscreenKeyboardType type);
 
 	/** Vibrates for the given amount of time. Note that you'll need the permission
 	 * <code> <uses-permission android:name="android.permission.VIBRATE" /></code> in your manifest file in order for this to work.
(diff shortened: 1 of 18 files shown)


gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -662,7 +662,7 @@ private static void initializeKeyNames () {
 	/** Sets the on-screen keyboard visible if available.
 	 * 
 	 * @param visible visible or not */
-	public void setOnscreenKeyboardVisible (boolean visible);
+	public void setOnscreenKeyboardVisible (boolean visible, OnscreenKeyboardType type);
```
intrigus: please also change the javadoc

PokeMMO:
Default methods in interfaces were java8, so would having two methods per Input implementation really be preferable?

MrStahlfelge:
You added this to `TextField`, do you think it is also possible to add this to `Gdx.input.getTextInput`? Would awesome!

EtK2000:
It's a shame this hasn't been merged yet 😭

PokeMMO:
There's some *rare* minor glitches with certain weird software keyboards that prevent me from suggesting this be merged still.

It works great 99.9% of the time, but occasionally requires users to restart.

MrStahlfelge:
Sad you closed it, looked like a good addition! What exactly was the problem?

PokeMMO:
I don't have the time to look into why, but sometimes the wrong keyboard type will get "stuck" open. I can't narrow down why or how, so it's not something I've been able to investigate.

PR #5202, opened 2018-04-27
Add game category to AndroidManifest

MrStahlfelge:
The category is used by some launchers to categorize the app icons.


obigu:
Interesting. Problem is that not all apps made with libGDX are games and having it as a default is not a good idea. I guess the proper solution would be to have a checkbox on the gdx-setup app but I'm afraid it may be a hard sell considering that, at the moment, these settings are for a very niche set of devices afaik.

MrStahlfelge:
I know not all applications made with libgdx are games, but after all, it still is a game development framework and I am sure developers making non-games are reviewing and changing the manifest anyway.

MobiDevelop:
I think more people are making games with libgdx than those who are not. Seems like a sensible default.

PR #5210, opened 2018-05-05
Removed unnecessary Align. in Align

TCreutzenberg:
Just something I stumbled upon.


TCreutzenberg:
Can somebody help me figure out why TravisCI is having an error?

NathanSweet:
Re: Travis, that's not my forte, but the [raw log](https://api.travis-ci.org/v3/job/375184447/log.txt) ends with:
```The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated```

PR #5218, opened 2018-05-11
Add focused to ButtonStyle

floralvikings:
This allows setting a background image for a button in focus.

Modeled it after the was the `focused` field is handled in `TextField`.

It's completely optional, shouldn't break any existing code.  `over` takes priority over `focused`.

`checkedFocused` also included.

Didn't add anything for the `Button` subclasses, but it would be reasonably simple to add equivalent fields for font color, checkbox, etc... in the appropriate subclasses.

The `if` statements I modified also have had curly braces added to match the recommendations in https://github.com/libgdx/libgdx/blob/master/.github/CONTRIBUTING.md

I've contributed in the past so I *believe* I've signed the CLA but if it's not on record I'd be happy to sign again.  I also didn't update `CONTRIBUTORS` or `CHANGES` since it's such a small change, but I'd be happy to do that as well if deemed appropriate.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Button.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Button.java
@@ -26,6 +26,7 @@ import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener.ChangeEvent;
 import com.badlogic.gdx.scenes.scene2d.utils.ClickListener;
 import com.badlogic.gdx.scenes.scene2d.utils.Disableable;
 import com.badlogic.gdx.scenes.scene2d.utils.Drawable;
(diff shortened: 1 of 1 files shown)


floralvikings:
@NathanSweet I'd appreciate any feedback on this; right now to indicate that a button is focused we're abusing the `checked` styles.  Having an appropriate style for focus would make our code significantly cleaner.

(Sorry if you're the wrong person to ping for this; just noticed you were the last committer in the `scene2d` package.)

NathanSweet:
The contributing guidelines have been corrected (only need curlies if the block spans more than one line). In general the style of the file being edited should be followed and changes to exist code minimized.

It looks OK. I'm a bit worried that people will want to double all the drawables in every widget to add focused styles, but I understand the need.

Cheers!

PR #5220, opened 2018-05-13
Optionally respect RGB of transparent pixels when aliasing (fix #5219)

### LINKED ISSUE #5219: TexturePacker assumes same-sized, all-zero-alpha images are identical
If aliasing is turned on, images that are the same size and have zero alpha in all pixels are treated as aliases of each other and only the first one is included in the pack.

This is a problem if you're using any kind of special effect in which the A channel of the image is not necessarily used for transparency. 

I'll take a stab at a PR. I realize that in most cases, the current behavior is desired, so I'll make it optional to turn it off.

cypherdare:
Rather than add a new TexturePacker setting, this looks at whether the `alphaThreshold` is set to less than zero for whitespace stripping. Presumably, if you don't want your zero alpha pixels ignored for aliasing, you also don't want them stripped off.

This addresses #5219.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/AlphaTest.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/AlphaTest.java
@@ -1,4 +1,67 @@
 package com.badlogic.gdx.tools;
 
-public class AlphaTest {
+import java.awt.AlphaComposite;
+import java.awt.Dimension;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+import java.awt.image.WritableRaster;
+
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+
+public class AlphaTest extends JPanel {
+
+    JFrame        frame = new JFrame();
+
+    BufferedImage image;
+    BufferedImage newImage, newImage2;
+
+    public static void main(String[] args) {
+        new AlphaTest().start();
+    }
+    public void start() {
+        frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
+        frame.add(this);
+        setPreferredSize(new Dimension(800, 500));
+        frame.pack();
+        frame.setLocationRelativeTo(null);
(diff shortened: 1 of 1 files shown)


NathanSweet:
The reason `alias` will discard two images with zero alpha but different RGB is probably that whitespace stripping reduces them to zero width/height and then `ignoreBlankImages` ignores them. Even if you turn `ignoreBlankImages` off, hashing is done after whitespace stripping so the stripped images are identical.

`alphaThreshold` is for whitespace stripping. If you set it to -1, you effectively disable whitespace stripping. In that case you might as well just set `stripWhitespaceX` and `stripWhitespaceY` to false.

cypherdare:
I checked whitespace stripping. Even with it off, it's ignoring the RGB
value of any pixel that has zero alpha.

I traced this to the drawImage() line in hash(), where it is converting
ABGR to ARGB. The Javadoc for this method says it does not transfer
transparent pixels. So all zero alpha pixels are effectively being
converted to black right before the hash takes place.

cypherdare:
My idea in this PR is if you are in a situation where RGB still carries
meaning even if A is zero, then you can't be using whitespace stripping
anyway and will want it turned off. Using the alpha threshold allowed me to
avoid adding yet another setting.

NathanSweet:
I don't think it makes sense to overload the `alphaThreshold` setting. It sounds like if converting the image copied the RGB when A is 0 then everything would work fine when you turn off whitespace stripping. Maybe you can try this?
```
Graphics2D g = (Graphics2D)newImage.getGraphics();
g.setComposite(AlphaComposite.Src);
g.drawImage(image, 0, 0, null);
```

cypherdare:
Sorry for the year-long response. My project using this was on hold during major life changes the past year.

I've tried updating the code to use AlphaComposite in every place we draw images, but it strangely still converts RGB to black if A is 0. Yet I tried using your code in a separate test and it worked as expected. 

Have I missed a location where we transfer the images?

cypherdare:
The AlphaTest isn't for pulling. It just shows an example where your method of preserving the RGB works.

PR #5231, opened 2018-05-23
GL20 texture filter types description

riki137:
Just thought it'd be nice if we had this.

Text is copied from the badlogicgames.com article


yuripourre:
Cool! Should be merged.

PR #5232, opened 2018-05-24
[Gradle] Android build.gradle - project.configurations.compile is deprecated

StrongJoshua:
The task `project.configurations.compile` is deprecated and likely to be removed by the end of 2018. `implementation` is what is supposed to be used now. There may be more that needs to be changed, but this seems to conform with new Gradle guidelines.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle
@@ -97,7 +97,7 @@ eclipse {
     }
 
     classpath {
-        plusConfigurations += [ project.configurations.compile ]        
+        plusConfigurations += [ project.configurations.implementation ]        
         containers 'com.android.ide.eclipse.adt.ANDROID_FRAMEWORK', 'com.android.ide.eclipse.adt.LIBRARIES'       
     }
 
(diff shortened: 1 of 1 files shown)


bernaferrari:
Hi, any progress on this?

TCreutzenberg:
When I change the lines in my project (1.9.9) then IntelliJ says "Resolving configuration 'implementation' directly is not allowed"?

itsabhiaryan:
changing compile to implementation in `build.gradle` of android module 

`project.configurations.compile` ->  `project.configurations.implementation`

giving me an Error 

> Resolving configuration 'implementation' directly is not allowed

Below version specified in my project :
LibGDX - 1.9.9
Gradle - 4.10.1
Gradle Android Plugin - 3.3.0

extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle
```diff
@@ -97,7 +97,7 @@ eclipse {
     }
 
     classpath {
-        plusConfigurations += [ project.configurations.compile ]        
+        plusConfigurations += [ project.configurations.implementation ]
```
eskatos: For IDE setup (here `eclipse` and `idea`) you'll want to use the `compileClasspath` configuration that is resolvable, it'll include dependencies from `implementation`. See https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations

MrStahlfelge:
Closing this, the lines in questions are long gone.

PR #5234, opened 2018-05-25
Shouldn't Ellipse's width & height be radius * 2f to match the original circle?

strongrex2001:
Hello! I happened to notice that when I use ellipse constructed with a circle, the real area of the shape became smaller.
I changed constructor Ellipse(Circle) and Ellipse.set(Circle). Shouldn't Ellipse's width & height be radius * 2f to match the original circle?


Tom-Ski:
Yup!

Darkyenus:
This should probably go into CHANGES as it will probably break someones code.

PR #5235, opened 2018-05-25
Added contains to binary heap

Shivang44:
Contains() method runs in linear time and does not interfere with any other parts of the binary heap, so I believe it would be a useful addition to the collection. I have personally found a use for it in my own game.


Shivang44:
@NathanSweet I believe a lot of the libgdx collections are your work, so I would love to get your opinion on whether you think this would be a useful addition.

NathanSweet:
Cheers!

PR #5242, opened 2018-06-01
fixed try-finally swallowing Exception after end

EtK2000:
Builds upon https://github.com/libgdx/libgdx/pull/5192

I have no clue why it's showing all my commits to LibGDX here...
Clearly IDK how to use git ;-;


obigu:
You are using master, it's good practice to create a separate branch per PR. Pull all changes from libgdx/master to origin/master then create a new branch and checkout, commit the changes and push to remote. You then can submit the PR from that branch.

Regarding the changes it all looks a bit forced/unnecessary, I also find it quite hard to imagine a scenario where you'd be able to recover from a non handled exception in Stage. And if it's a particular case, maybe it could be handled in that Actor itself.

EtK2000:
I guess it should be handled in the Actor too.
I'll look into it in Table, that's the source usually.
Setting colspan too high throwing an exception.

PR #5247, opened 2018-06-03
FreeType - Use zero copy memory mapping when possible

PokeMMO:
Should allow a lower memory footprint and faster startup times on Desktop/Android.

Android requires users to manually edit their gradle config to include certain aaptOptions to not compress font files. Would we want to include this in the default project setup?

Such as:
`android {
aaptOptions {
    noCompress 'ttf', 'ttc', 'otf'
}
}`

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -67,7 +67,8 @@ public class FreeType {
 		public void dispose () {
 			doneFreeType(address);
 			for(ByteBuffer buffer: fontData.values()) {
-				BufferUtils.disposeUnsafeByteBuffer(buffer);
+				if (BufferUtils.isUnsafeByteBuffer(buffer)) 
+					BufferUtils.disposeUnsafeByteBuffer(buffer);
 			}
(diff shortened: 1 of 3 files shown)


intrigus:
Just curious: Did you actually profile this?

PokeMMO:
Yes, and it's more extreme than I would expect.

Android Normal UnsafeByteBuffer:
freetypemap-normal

Android MappedByteBuffer:
freetypemap-map

I'm not sure if it's an accounting issue with the profiler, or if something silly like the whole apk is being mapped and this is just reusing the existing mapping. The difference is about the sum of my application's font files.

Time wise loading and generating (base characters only) a ~18MB ttc dropped from around 300-350ms to 150-225ms.

I would be really interested if anyone else could benchmark to compare.

NathanSweet:
It looks nice to me. We added FileHandle `map` without input from iOS guys. Is everything OK there?

PokeMMO:
[Code for FileChannel#map exists in RoboVM](https://github.com/MobiVM/robovm/blob/45b1de02cad5d857918d7604414404323c75f715/compiler/rt/libcore/luni/src/main/java/java/nio/FileChannelImpl.java#L216)

Everything I can see would indicate this should be fine. Even if it's not supported at runtime, compile compatibility should be enough given the fallback to unsafe ByteBuffer.

PR #5259, opened 2018-06-09
FileHandle#map compatibility for GWT

### LINKED ISSUE #5257: Building GWT project fails on nightlies due to unimplemented NIO classes
#### Issue details
I noticed this when @villain-bryan asked me to double-check a sudden build failure on the nightlies with GWT. I was able to reproduce by changing the dependency from `1.9.8` to `1.9.9-SNAPSHOT` in two existing projects with GWT targets and one project made today to check an unrelated issue (which also had a GWT target). Checking recent commits, the code that adds `Filehandle#map(FileChannel.MapMode)` seems to be to blame because `MappedByteBuffer` and `FileChannel.MapMode` are not implemented on GWT 2.8.0 . I haven't been able to get 2.8.1 to build any projects (haven't tried 2.8.2, but it sounds like @MobiDevelop is hard at work on that), and though I think it implements a tiny amount of NIO classes, they're only charset-related. It looks like the classes`MappedByteBuffer`, `FileChannel`, and `FileChannel.MapMode` would need to be given no-op super-source implementations on the GWT backend (matching the other NIO classes that are super-sourced in the `emu` folder); either that or the super-sourced FileHandle used on GWT needs to somehow omit map() or mark it as GwtIncompatible. I don't know how libGDX handles GWT-incompatible methods that have unimplemented classes as parameter and return types.

#### Reproduction steps/code

Make a default project, checking GWT in the setup. Update the libGDX version to the current nightly build as of June 8, 2018, using `1.9.9-SNAPSHOT`. Run `gradlew gwt:dist` . Using `gradlew gwt:superDev` can also reproduce.

LibGDX 1.9.9-SNAPSHOT nightly as of June 8, 2018, GWT backend, GWT 2.8.0.

#### Stacktrace
```
Executing tasks: [dist]

NDK is missing a "platforms" directory.
If you are using NDK, verify the ndk.dir is set to a valid NDK directory.  It is currently set to C:\Android\android\android-sdk\ndk-bundle.
If you are not using NDK, unset the NDK variable from ANDROID_NDK_HOME or local.properties to remove this warning.

:gwt:clean
:gwt:addSource
:core:compileJava UP-TO-DATE
:core:processResources UP-TO-DATE
:core:classes UP-TO-DATE
:core:jar UP-TO-DATE
warning: [options] bootstrap class path not set in conjunction with -source 1.7
1 warning
:gwt:compileJava
:gwt:processResources NO-SOURCE
:gwt:compileGwt
Compiling module color.guard.GdxDefinition
   Tracing compile failure path for type 'com.badlogic.gdx.backends.gwt.GwtFileHandle'
      [ERROR] Errors in 'jar:file:/C:/Users/Tommy/.gradle/caches/modules-2/files-2.1/com.badlogicgames.gdx/gdx-backend-gwt/1.9.9-SNAPSHOT/dae5286e9297c49b8db969fe6d29fb7ca4121ef2/gdx-backend-gwt-1.9.9-SNAPSHOT-sources.jar!/com/badlogic/gdx/backends/gwt/GwtFileHandle.java'
         [ERROR] Line 222: No source code is available for type java.nio.MappedByteBuffer; did you forget to inherit a required module?
         [ERROR] Line 222: No source code is available for type java.nio.channels.FileChannel.MapMode; did you forget to inherit a required module?
   [ERROR] Aborting compile due to errors in some input files
:gwt:compileGwt FAILED

* What went wrong:
Execution failed for task ':gwt:compileGwt'.
> Process 'command 'C:\Program Files\Zulu\zulu-8\bin\java.exe'' finished with non-zero exit value 1
```

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [X] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Thanks, and I hope this is a simple fix! Though the simplest fix would be to remove FileHandle.map(), that seems counterproductive. I'm also not sure how much work it would be to make no-op versions of the classes used by map() so GWT can't use them but will still compile. If you have any questions, I'll try to answer.

PokeMMO:
Fixes #5257

I don't think implementing a MappedByteBuffer class in GWT emu classes is worth the hassle. Android/Desktop specific code can always cast to MappedByteBuffer to invoke the 3 utility functions it implements.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFileHandle.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFileHandle.java
@@ -22,8 +22,8 @@ import java.io.FileInputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.nio.MappedByteBuffer;
 import java.nio.channels.FileChannel;
 
 import android.content.res.AssetFileDescriptor;
@@ -85,7 +85,7 @@ public class AndroidFileHandle extends FileHandle {
 		return super.read();
 	}
 
-	public MappedByteBuffer map (FileChannel.MapMode mode) {
+	public ByteBuffer map (FileChannel.MapMode mode) {
 		if (type == FileType.Internal) {
 			FileInputStream input = null;
 			try {
@@ -93,7 +93,7 @@ public class AndroidFileHandle extends FileHandle {
 				long startOffset = fd.getStartOffset();
 				long declaredLength = fd.getDeclaredLength();
 				input = new FileInputStream(fd.getFileDescriptor());
-				MappedByteBuffer map = input.getChannel().map(mode, startOffset, declaredLength);
+				ByteBuffer map = input.getChannel().map(mode, startOffset, declaredLength);
 				map.order(ByteOrder.nativeOrder());
 				return map;
 			} catch (Exception ex) {
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtFileHandle.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtFileHandle.java
@@ -28,7 +28,7 @@ import java.io.OutputStream;
 import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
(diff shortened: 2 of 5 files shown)


tommyettinger:
Works for me with `desktop:run` and `html:dist` when I depend on this version in a freshly-made project. I don't know how the FileChannel imports of non-existent classes worked before 418fd7f , but I suspect GWT could tell they were unused and so didn't try to import them. They also weren't ever built by a desktop JDK, and those may be more strict. Thanks for fixing this!

PR #5263, opened 2018-06-12
Use https for maven native fetching

PokeMMO:
http://libgdx.badlogicgames.com redirects to https (with HSTS) anyway


NathanSweet:
I don't think we can merge this based on this commit:
https://github.com/PokeMMO/libgdx/commit/2dd84e1abd658287cf67718b90b57770a16df590#diff-b37cae51e6d920f8036565a0c8af6f79

PR #5264, opened 2018-06-13
Textfield Focus Traversal Now Ignores Invisible Windows.

donte5252:
Fixed an issue that caused textfields to lose focus due to invisible windows when pressing tab. The findNextTextField method now searches up the parent hierarchy using the new isComponentHierarchyVisible method in the actor class to determine if all parent actors and the actor itself are visible. If any of the parent actors or the actor itself are invisible, the findNextTextField method will ignore them. Before the changes, pressing tab on a window which had other invisible windows in the stage would cause the cursor to lose focus and disappear in an attempt to get to the invisible window.


Darkyenus:
There is something terribly wrong with this PR. Also, please run the code formatter.

Btw. I don't mean to nitpick, but the implementation of `isComponentHierarchyVisible()` could be written more cleanly like this:
```java
public boolean isHierarchyVisible(){
 	Actor actor = this;
 	do {
 		if (!actor.isVisible()) {
 			return false;
 		}
 		actor = actor.getParent();
 	} while (actor != null);
 	return true;
 }
```

NathanSweet:
This brings up the question, "Should invisible actors respond to input events?" Actor `hit` javadoc says it respects touchable and visible, but this is a lie -- it doesn't check visible. Checking the history, it never has. Group `hit` does, which is why invisible actors don't get hit. 7d81d2b191545be6781d5bc6483c4dc474ce9b09 is where old me put the code into Group. Despite that commit's title, "hit() now must respect touchable and visible", it was jammed into Group. Good job!

I think since Group has been avoiding checking `hit` for invisible actors for 6 glorious years, that is the expected behavior. I think actors should be in control of that decision though, as was the original intended behavior, which means moving the visible check from Group `hit` to Actor `hit`. This could impact existing code, Actor subclasses who override `hit` but never had to check for visible before. Given that visible is not used terribly often, this is probably OK.

Back to this PR, since invisible actors shouldn't get input events, tab traversal should skip them. This PR deletes all the lines from TextField though, so we can't merge it.

Issue #5241, opened 2018-05-30
Touch pressure


piotr-j:
Seems apple likes when stuff uses their gimmicks and there is no support currently. There is some sort of support for that on android side as well. 

Simple api addition to input:
   public float getPressure(int pointerId);

One problem is scaling, on ios 1 is about normal touch and it can go up quite a bit, but on android its more like 0-1 according to docks, when 1 is the default if device is not capable. Any ideas?

Anyone working on this by any chance? If not i can give it a stab.

Tom-Ski:
Sounds good, normalize between 0-1 across platforms. On iOS you can query maximumPossibleForce to normalize.

piotr-j:
Crossed my mind, but maximum differs between touch sources, finger or pencil. It is already scaled so 1 is the average touch force. There was also a note about it going above 1.0 on android as well. 

Ive tested on 3 android devices, all reported different pressure values...
Nexus 4 A 5.0- 1 normal, 1.5 hard
Shield K1 A 7.0 - .3 normal, .5 hard
mediapad m3 A 7.0 - .7 normal, 1.0 hard
Small sample size, but not very encouraging

You can throw this into `AndroidLauncher` for quit and dirty test...

` graphics.getView().setOnTouchListener(new View.OnTouchListener() {
            @Override public boolean onTouch (View view, MotionEvent event) {
                final int action = event.getAction() & MotionEvent.ACTION_MASK;
                int pointerIndex = (event.getAction() & MotionEvent.ACTION_POINTER_INDEX_MASK) >> MotionEvent.ACTION_POINTER_INDEX_SHIFT;
                switch (action) {
                    case MotionEvent.ACTION_DOWN:
                    case MotionEvent.ACTION_POINTER_DOWN: {
                        Gdx.app.log("DOWN", pointerIndex + " -> " + event.getPressure(pointerIndex));
                    } break;
                    case MotionEvent.ACTION_UP:
                    case MotionEvent.ACTION_POINTER_UP: {
                        Gdx.app.log("UP", pointerIndex + " -> " + event.getPressure(pointerIndex));
                    } break;
                    case MotionEvent.ACTION_MOVE: {
                        for (int i = 0, n = event.getPointerCount(); i < n; i++) {
                            Gdx.app.log("MOVE", i + " -> " + event.getPressure(i));
                        }
                    } break;
                }
                return true;
            }
        });`

piotr-j:
[...]

PR #5270, opened 2018-06-15
Touch pressure support on ios and android

piotr-j:
Implemented touch pressure on android and ios, modified multitouch test to use it.
Could use some feedback!
#5241


piotr-j:
Any idea why does the pr check fails?

lukz:
Awesome! Really need this one!

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/input/MockInput.java
```diff
@@ -110,6 +110,14 @@ public boolean isTouched(int pointer) {
 		return false;
 	}
 
+	@Override public float getPressure () {
```
Tom-Ski: Can you format these overrides?

piotr-j:
Any other dumb things that i missed?

balepc:
This is very expected!!! Could you, please, merge it?

PR #5275, opened 2018-06-20
AnimatedTiledMapTile#updateAnimationBaseTime(...)

TCreutzenberg:
Problem to solve with this pull request:

lastTiledMapRenderTime is a private field in AnimatedTiledMapTile.
In AnimatedTiledMapTile there is no way of setting lastTiledMapRenderTime.
I.e. you cannot influence the speed of the animation as it is always set to: "lastTiledMapRenderTime = TimeUtils.millis() - initialTimeOffset;" when BatchTiledMapRenderer#beginRender() is called.

This pull request offers a method to set lastTiledMapRenderTime directly. So AnimatedTiledMapTile offers an option to influence the animation speed.

I was looking to solve this issue in a better way but I didn't find another good solution as this would mean changing the api of BatchTiledMapRenderer#beginRender() which might not be a good idea. But I'm more than happy to do so if wanted.


Tom-Ski:
The rates are determined by the frame intevals you supply for each tile. I'm guessing that you want to globally influence every AnimatedTile instance's frametime dynamically? Could you give the exact scenario? 

 If so we can include some global animation speed modifier variable with appropriate setter, as that is a little less hacky.

TCreutzenberg:
Hey Tom-Ski,

thanks for your reply!

The goal is to influence the speed at which the AnimatedTiledMapTiles are updated/animated, regardless of the rest of the game.

My scenario is as follows: I want to set the "speed" of AnimatedTiledMapTile animations (as a "special effect") and the "simulation" (this I can do) , but the UI should still animate at original speed (this I can do).

At the moment there is no way at all to influence the AnimatedTiledMapTiles animation timer, which is bad in any case, think of a global slowdown/speedup...

TCreutzenberg:
I change the method to be pretty much noninvasive and not hacky at all. Please have a look. People can now just overload BatchTiledMapRenderer.beginRender() to change tile animation speed.

TCreutzenberg:
@Tom-Ski Hey Tomski, how are you?
Did you see the changes? Is it better now?

TCreutzenberg:
@MobiDevelop Hey, it's me again, I just noticed you're also involved in this bit... If you have any suggestions on how to improve this, I'm more than happy to do so. Thanks in advance!

TCreutzenberg:
@MobiDevelop Sorry to bother you again, if you are not responsible for this file any more, please just tell me or maybe you can even tell me whom I could contact instead. Thank you very much in advance!

MobiDevelop:
What if we instead let the renderer have an "Animator" instance and always just set the animation time on AnimatedTiledMapTile? Something like:

```java
interface Animator {
	void updateAnimationTime();
}

public static class DefaultAnimator implements Animator {

	private final long initialTimeOffset = TimeUtils.millis();

	@Override
	public void updateAnimationTime() {
		AnimatedTiledMapTile.setAnimationTime(TimeUtils.millis() - initialTimeOffset);
	}

}
```

```java
public abstract class BatchTiledMapRenderer implements TiledMapRenderer, Disposable {
    ...
    protected void beginRender() {
        animator.updateAnimationTime();
        batch.begin();
    }
    ...
}
```

I guess this makes it a 'breaking' change since the automatic updating is removed, but it feels cleaner.

TCreutzenberg:
To be honest, I'm happy with any solution as long as it allows to control the animation speed / base time.
My implementation aimed at not being a breaking change as I thought this might increase the chances of this PR be be merged.

@MobiDevelop Just tell me if I should create a new PR with your solution and I'll do it in a few days when I have time do do it.

There are also other lightweight possibilities to achieve the same effect. I would already be very happy with just a AnimatedTiledMapTile.setAnimationTime(long animationBaseTime)

TCreutzenberg:
Closed in favour of https://github.com/libgdx/libgdx/pull/6523

PR #5283, opened 2018-06-27
Fix #5280 - Added Independent setting to Particle Editor Life properties

### LINKED ISSUE #5280: Life attribute on Particle Emitter is not applied to particles individually
#### Issue details
The behaviour of Life and Life Offset attributes of a Particle Emitter is different than the rest of attributes such as Rotation, Angle, Gravity, etc... in the sense that it is not evaluated by particle but only by time. What this means is that All particles emitted at one time will have the same Life. This limitation doesn't allow simple effects such an explosion in which not all particles live the same.

According to the Wiki:

>Life: How long a single particle will live.

>Life Offset: How much life is used up when a particle spawns. The particle is still moved/rotated/etc for the portion of its life that is used up. This allows particles to spawn, eg, halfway through their life.

It could be that the way the Life attribute behaves atm is useful for some type of effect but as a user I miss a Life attribute that affects each particle individually.

**Whether we need a new Life attriubute or fix the existing one I will happily work on a PR.**

#### Reproduction steps/code
Video showing the issue, observe that when life value is modified to be Ranged, all particles have the same Life.
https://youtu.be/IXOgCgAUmk8

1.9.8

obigu:
Fixes #5280.

Added a new setting ("Independent" checkbox) to Life and Life Offset properties that allows that when defining a ranged value it affects each particle independently instead of all particles. The change is backwards compatible with previously created effects which should behave exactly like before.

As a side note, it has been quite challenging to add the new setting in a clean way (and it's still not great) due to how the Particle Editor is designed and the rigid particle effects file descriptor format. The 2D Particle system definitely deserves a good refactoring :).

There's a link to the new 2D Particle Editor .jar in the issue comments if you want to try it.

I'll squash changes when I get the Ok.

**Do changes to the Wiki need to be part of the PR?**

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/particleeditor/ScaledNumericPanel.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/particleeditor/ScaledNumericPanel.java
@@ -31,6 +31,7 @@ import javax.swing.JPanel;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 
+import com.badlogic.gdx.graphics.g2d.ParticleEmitter;
 import com.badlogic.gdx.graphics.g2d.ParticleEmitter.ScaledNumericValue;
 
 class ScaledNumericPanel extends EditorPanel {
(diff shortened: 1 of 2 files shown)


NathanSweet:
GitHub has a squash button, makes it easy.

The wiki is stored in a separate repo and I believe can be edited without submitting PRs. (It occurs to me I've never tried to edit a GitHub wiki where I don't have repo write access.)

The particle stuff is very old. I used it only initially, then got stuck doing things besides making games. :/ For a long time now mostly the only parts of libgdx I work on are the parts I'm actually using in apps, which is mostly scene2d.

PR #5284, opened 2018-06-28
Add getter for the internal buffer in StringWriter

barkholt:
In contrast to the regular Java version of StringWriter, there is currently no way to access the internal buffer of the StringWriter used in libGDX. This patch adds a simple getter to allow this.

I realise that this will not be 100% source compatible with the Java StringWriter due to the use of StringBuffer and StringBuilder in the different implementations (but this is no different from the current situation). But it does make it possible to re-use StringWriter objects, and manipulate the buffer using the same function calls as in regular Java.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/java/io/StringWriter.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/java/io/StringWriter.java
@@ -28,6 +28,10 @@ public class StringWriter extends Writer {
 	public String toString () {
 		return out.toString();
 	}
+	
+	public StringBuilder getBuffer() {
+		return out;
+	}
 
 	public void flush () throws IOException {
 	}


barkholt:
It's really a case of no solution being perfect. Personally, I would vote for adding documentation to explain why it has been done this way (keep source compatability, if you refrain from storing the result of `getBuffer()`). 

Adding a dummy `StringBuffer` class seems nicer, but then we need to explain that it was introduced to solve this very specific case. And it might also create false expectations of its behaviour based on the regular JDK, that does not make sense in a JavaScript environment.

MobiDevelop:
Why don't we just change StringWriter to use StringBuffer internally? Since it isn't exposed in any way today, and StringBuffer is identical to StringBuilder on GWT, this seems like the best solution.

barkholt:
That is of course the right solution.

I don't know why I was under the impression that `StringBuffer` was not part of GWT. Unless anyone minds, I'll rework the patch.

PR #5288, opened 2018-06-30
Small improvements to the code complexity

asarlidou:
Testing on a static code analysis tool gave the some suggestions, decided to make a pr with them.
Mainly improvements on boolean expressions and some list size calls.


intrigus:
Please revert the ternary expression optimization. It doesn't express the intended meaning if you don't know these optimizations making the code hard to read. 

Additionally I'm pretty sure that the generated machine code (jit) would be the same.

obigu:
@intrigus I think that on maters related to code styling we should stick to general consensus. In the case of the ternary operator you will find most consensus on using ther ternary operator Only for simple expressions or assignements as most people consdiers it decreases code readibility.

NathanSweet:
I've never seen anyone suggest to use the ternary operator instead of `&&`. I would never do so in my code and I feel `&&` is perfectly clear.

PR #5311, opened 2018-07-17
Specify arrayType to avoid casting requirement in InputMultiplexer

PokeMMO:
Only downside is inconsequentially slower initialization time, but faster input processing should make up for that.


Darkyenus:
It is unlikely to have any measurable impact on performance (bytecode wise, the only difference is where compiler checks for correct type, and even if JIT doesn't optimize it out, this is hardly in anyone's hot path), but the code seems cleaner this way, IMHO.

MobiDevelop:
Did you check this on GWT? Since it will use reflection to create the array, we'd need to make sure InputProcessor is included in the reflection includes.

NathanSweet:
I'd say this isn't really necessary and GWT needing extra type info isn't great. This isn't a hotspot and the PR isn't faster in any meaningful way: calling Array `get` will do the same cast in the bytecode, even if the array has a generic type. It is slightly more readable, but I don't think anyone will be confused by such simple code and this is the standard iteration as shown in the Snapshot javadoc.

PR #5324, opened 2018-07-23
Added isButtonJustPressed method similar to isKeyJustPressed

### LINKED ISSUE #5266: Add isButtonJustPressed just like isKeyJustPressed
Simply put, this is a way to check if a mouse button was just pressed.
It shouldn't be too hard to add, but I'm unfamiliar with some of the backends so I doubt I can properly implement it myself.

I'd guess this would be added just like isKeyJustPressed as an array of booleans that are set and unset using events.

I can create the functions in the backends I know and supply them to whoever would like to implement the rest.

Pilzschaf:
This PR fixes #5266 by adding an isButtonJustPressed method to the Input interface. It works similar to isKeyJustPressed and only returns true if the specified button is pressed in this frame and was not in the previous frame.


mgsx-dev:
@Pilzschaf could you please resolve conflicts (CHANGES.md i think)

Pilzschaf:
Conflict is resolved now

PR #5329, opened 2018-08-02
Add proguard rule for AndroidVisibilityListener

thekeenant:
It is instantiated [via reflection](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java#L190) so proguard needs to know or else you get:

```
08-01 21:42:30.468 11993-11993/com.keenant.racer I/AndroidApplication: Failed to create AndroidVisibilityListener
    java.lang.NoSuchMethodException: createListener [interface com.badlogic.gdx.backends.android.AndroidApplicationBase]
        at java.lang.Class.getMethod(Class.java:2068)
        at java.lang.Class.getDeclaredMethod(Class.java:2047)
        at com.badlogic.gdx.backends.android.AndroidApplication.init(AndroidApplication.java:192)
        at com.badlogic.gdx.backends.android.AndroidApplication.initialize(AndroidApplication.java:100)
        at com.keenant.racer.AndroidLauncher.onCreate(AndroidLauncher.java:55)
        at android.app.Activity.performCreate(Activity.java:7009)
        at android.app.Activity.performCreate(Activity.java:7000)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1214)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2731)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2856)
        at android.app.ActivityThread.-wrap11(Unknown Source:0)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1589)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/proguard-rules.pro
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/proguard-rules.pro
@@ -29,6 +29,7 @@
 -dontwarn com.badlogic.gdx.graphics.g2d.freetype.FreetypeBuild
 
 -keep class com.badlogic.gdx.controllers.android.AndroidControllers
+-keep class com.badlogic.gdx.backends.android.AndroidVisibilityListener { *; }
 
 -keepclassmembers class com.badlogic.gdx.backends.android.AndroidInput* {
    <init>(com.badlogic.gdx.Application, android.content.Context, java.lang.Object, com.badlogic.gdx.backends.android.AndroidApplicationConfiguration);


obigu:
Good find. If this listener is never called in production builds and we haven't noticed any issue with Immersive Mode in those builds I believe we should understand why before merging this.

thekeenant:
My understanding was that proguard could understand this simple of reflection, but my device was throwing the exception above. Maybe something changed with proguard?

obigu:
I've just tested a ProGuarded production build and you are right, the Listener is never created. ProGuard is able to respect reflection under some specific circumstances but it doesn't seem to be the case here.

Having a look at the [Immersive Mode documentation] (https://developer.android.com/training/system-ui/immersive) I believe this may be related to the fact that we use Sticky Immersive and flags are never cleared:

`Note: When you use the SYSTEM_UI_FLAG_IMMERSIVE_STICKY flag, a swipe causes the system UI to temporarily appear in a semi-transparent state, but no flags are cleared and your system UI visibility change listeners are not triggered.`

If I understand correctly this means that there's no need to handle flags on visibility changes as flags are kept and that is the reason why nobody noticed the issue.

IMO, instead of adding this Proguard rule we should just get rid of the whole AndroidVisibilityListener. The way it's instantiated is unnecessarily convoluted imo and it has never worked in production anyway.

What do you think @Tom-Ski? Shall we create a new PR to get rid of it?

obigu:
Yes, when saying produciton build I was referring to using Proguard. The quote about the flags not being cleared when using STICKY is the best explanation I can give to the fact that no user that uses ProGuard has noticed any issue with Immersive Mode in their production apps. Users that don't use ProGuard would not be affected by this, the listener would be created but not be necessary, same as when running in Dev environment.
Best way to confirm would probably be to run without the SYSTEM_UI_FLAG_IMMERSIVE_STICKY flag and see if flags are cleared which would make the call to restore them on the system UI visibility listener required.

MobiDevelop:
What I was getting at is that the listener was added for a reason... To put it back in immersive mode when something screwed with the system ui visibility. Now, I can appreciate it may no longer be necessary but the simpler thing to do it's just add the Proguard rule than have to go figure out again should someone come across the scenario that caused it to be added in the first place.

obigu:
My point is that if nobody has ever noticed it did not work on production apps (when using ProGuard) I doubt it's required even if it may have been added for a reason in the past (I see comments on GitHub regarding Android CupCake support, who knows).
There's a small risk both ways, there's also a risk to break some live apps by adding the ProGuard rule. In my opinion, if we believe it's not needed it should be removed.

cypherdare:
@obigu This won' t break existing apps because it won't get added retroactively to existing projects. It is only for newly created projects from the Setup application.

One reason this possibly wasn't discovered before is that Android Studio now defaults to the R8 minifier instead of Proguard. It's theoretically backward compatible, but the android dev blog post says there may be some missed edge cases, still.

cypherdare:
Actually, do we even need reflection to instantiate this? I remember about 7-8 years ago this was necessary to access classes more recent than the minSDKVersion provides, but I thought they updated the Android build tools a number of years ago to handle it with a simple boolean check. [This is how they do it in their examples now.](https://developer.android.com/training/basics/supporting-devices/platforms#version-codes)

MobiDevelop:
This is no longer instantiated via reflection.

PR #5335, opened 2018-08-05
Removed unused minSDK and targetSDK on android manifest

obigu:
The minSDK and targetSDK are configured on build.gradle, this settings are unused


obigu:
I tought that may be the case but does that apply to new projects created
with Gdx Setup? And for people with such custom builds, shouldn't they know
how to configure those custom environments and add whatever is missing on
the manifest for them to work?

Tom-Ski:
It should be as complete as possible from the get go. And yes you can use the generator to create projects and still not build with Gradle.

obigu:
Ok, I thought Android could not be built without Gradle anymore. Can be
closed.

Tom-Ski:
Yeah, its possible. Painful and not recommended, but there are those out there that like that kind of thing. So I think it would be best to just leave it there, its not doing too much harm, and prevents an additional step for those wanting to go down that path.

PR #5355, opened 2018-08-26
Intersector.intersectSegmentRectangle()

riki137:
i couldn't believe this method was missing in the awesome Intersector.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -801,50 +801,37 @@ public final class Intersector {
 		return false;
 	}
 
+	/** Determines whether the given rectangle and segment intersect and, if they do, sets the supplied {@code intersection}
+	 * rectangle to the area of overlap.
+	 *
+	 * @param startX x-coordinate start of line segment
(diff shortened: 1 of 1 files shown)


Darkyenus:
Please run the [formatter](https://github.com/libgdx/libgdx/blob/master/.github/CONTRIBUTING.md#eclipse-formatter).

riki137:
@NathanSweet @Darkyenus done

NathanSweet:
Do we want to handle the segment being completely inside the rectangle? To do that, as a final check we could check if one end of the segment is inside the rectangle (no need to check both ends).

obigu:
There's also the special case of the segment being parallel to one of the rectangle edges and sharing at least one point in which case it would return false. I guess most times the Intersector class is used to check collsions/overlaps so in my opinion it should return yes. That would also make it consistent with intersectSegmentCircle (which has an open PR (https://github.com/libgdx/libgdx/pull/5098) that requires a review :P)

riki137:
@NathanSweet i was thinking about that possibility day after the commit, but didn't get myself to fix it. that is done.

riki137:
I also removed intersection vector argument since intersection point in this context makes no sense.

Issue #1375, opened 2014-02-20
List directory on desktop


saltares:
It'd be nice if we made `public FileHandle#FileHandle[] list ()` work on the desktop backend as well. Speaking from ignorance, could we use something like this? I'm happy to implement it myself.

http://stackoverflow.com/questions/1429172/how-do-i-list-the-files-inside-a-jar-file/1429275#1429275

Metaphore:
Any news on it guys? Seems game preloader that loads all assets automatically is still hard to maintain to achieve full compatibility. Actually I have problems not only with jar's classpath assets, but with GWT too. By some reason, FileHandle#list() can't grab my audio(.ogg/.wav) and particles(.p) files (other files loads fine). It looks even strange cuz all these files contain inside symlink file that generates by HTML build.

billy1380:
I found out on desktop (linux) doing a readString of a folder of type internal returns a line feed delimited string of the folder/filenames inside. If you really need to use a folder structure where you know whether the entries are going to be files or folders ahead of time or if you have left metadata/clues in the names, then you can easily use the combination to get to or copy file contents that way.

badlogic:
Sorry, there's simply not good solution to this. If you test on the desktop, start your app in the assets directory, don't package them in the classpath. When deploying your desktop app, don't package your assets, but keep them next to your jar.

Zvezdin:
How does one list and save the filenames in the directory during compile time as a workaround to the issue? I'm loading many resources from my assets directory and don't want to hardcode all their filenames.

PR #5366, opened 2018-09-09
Internal List Fix

abueide:
Fixes not being able to use FileHandle#list on internal files by creating an index at compile time and looking for that index when using the API. Related to (#1375)


gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -416,12 +400,30 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+	  * See for reference: * https://gist.github.com/abueide/e86ea0111025233fb6493dfbcc46501e
+	  *
+	  * @throws GdxRuntimeException if this file is an {@link FileType#Classpath} file.
+	  */
 	public FileHandle[] list () {
 		if (type == FileType.Classpath) throw new GdxRuntimeException("Cannot list a classpath directory: " + file);
+		if (type == FileType.Internal){
+			if(!Gdx.files.internal("assets.index").exists())
+				 //How is this supposed to be done? Is there a framework level logging system to log a warning?
+				System.out.println("Asset index not found, please update your gradle configuration " +
```
Tom-Ski: This should use Gdx.app.log

extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
```diff
@@ -81,6 +81,15 @@ public static void addAllProjects(BufferedWriter wr) throws IOException {
 		write(wr, "maven { url \"" + DependencyBank.libGDXSnapshotsUrl + "\" }");
 		write(wr, "maven { url \"" + DependencyBank.libGDXReleaseUrl + "\" }");
 		write(wr, "}");
+		space(wr);
```
Darkyenus: I suggest adding a comment here with explanation of what this does, for future reference. It would also be nice to add such comment to the generated source, so that users know whether this task can be safely deleted/modified/etc.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -416,25 +401,48 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
-	/** Returns the paths to the children of this directory that satisfy the specified filter. Returns an empty list if this file
-	 * handle represents a file and not a directory. On the desktop, an {@link FileType#Internal} handle to a directory on the
-	 * classpath will return a zero length array.
-	 * @param filter the {@link FileFilter} to filter files
-	 * @throws GdxRuntimeException if this file is an {@link FileType#Classpath} file. */
+	 /**
+	  * Returns the paths to the children of this directory. Returns an empty list if this file handle represents a file and not a
+	  * directory. An {@link FileType#Internal} handle to a directory on the classpath will search for
+	  * an asset.index file in the * root of the asset path. This index should be generated at compile time by a gradle task.
+	  * See for reference: * https://gist.github.com/abueide/e86ea0111025233fb6493dfbcc46501e
```
Darkyenus: Referencing a gist seems messy. It would be much better to document the feature fully here, including the full description of the `asset.index` file format (including where the file should be, what directory separators it should contain, if it should contain leading `/`, ...), so that somebody who would want to generate it in a different build system or through a shell script would not have to reverse engineer the source code.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -416,25 +401,48 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+	  *
+	  * @throws GdxRuntimeException if this file is an {@link FileType#Classpath} file.
+	  */
+	 public FileHandle[] list () {
+		  if (type == FileType.Classpath)
+				throw new GdxRuntimeException("Cannot list a classpath directory: " + file);
+		  if (type == FileType.Internal) {
+				if (!Gdx.files.internal("assets.index").exists())
+					 //How is this supposed to be done? Is there a framework level logging system to log a warning?
+					 Gdx.app.log("WARNING: ","Asset index not found, please update your gradle configuration "
```
Darkyenus: This should be something like: `Gdx.app.log("FileHandle", "Asset index not found, listing content of internal directories will not be possible. Check the documentation for more details.")`

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -416,25 +401,48 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+						 + "according to: https://gist.github.com/abueide/e86ea0111025233fb6493dfbcc46501e");
+				else {
+					 String[] filePaths = Gdx.files.internal("assets.index").readString().split("\n");
```
Darkyenus: It would probably be more efficient to use an explicit `BufferedReader` and read the file line by line. Also I am not sure if this shouldn't be cached, at least on the `FileHandle` instance, but some global cache would be even better. Cache in the `Gdx.files` would make the most sense.

However, I am not sure if I am reading this correctly, but it seems that `list()` on Internal files will always return list of all internal files, regardless of their relation to the path of used `FileHandle`. All other types of FileHandles only return list of files directly under the FileHandle's directory, so this would be extremely confusing.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -416,25 +401,48 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+				handles[i] = child(relativePaths[i]);
+		  return handles;
+	 }
+
+	 /**
+	  * Returns the paths to the children of this directory that satisfy the specified filter. Returns an empty list if this file
+	  * handle represents a file and not a directory. An {@link FileType#Internal} handle to a directory on the classpath will
+	  * search for an asset.index file in the root of the asset path. This index should be generated at compile time by a gradle task.
+	  * See for reference: * https://gist.github.com/abueide/e86ea0111025233fb6493dfbcc46501e
+	  *
```
Darkyenus: These `list` overloads don't need full documentation, I suggest just adding `@see #list() Internal file handling`, or something like that.

gdx/src/com/badlogic/gdx/files/FileHandle.java
abueide: Yeah the gist was just a stop-gap until I knew if this would be an accepted change or not. I was planning on editing the section on the libgdx wiki and then linking to that. Writing a spec in the doc instead of just a gradle task sounds like a good idea though.

gdx/src/com/badlogic/gdx/files/FileHandle.java
abueide: The second part is fixed. It will now no longer return a list of all internal files, just a list of files in the FileHandle it is called on. I'm not sure what you mean by putting it in the cache. I don't know how caching with libgdx works or what its used for.

abueide: I am also unsure how I would use a BufferedReader on the classpath to read assets, so for now I think I'll keep using Gdx.files.internal because I have already spent too much time on this.

abueide:
I think this PR is very near to being complete. I've caught up on finishing all of the suggestions, let me know if there is anything else I should change!

PokeMMO:
It looks like this will break the promise that it will check filesystem relative to the application's root directory first on desktop. Should we be merging the filesytem results with the index from classpath, or just accepting/documenting this limitation?

Should the index be cached to speed up subsequent calls?

[Could this be used for an optimized exists check for directories on android?](https://github.com/libgdx/libgdx/blob/a4805d6a017b80622d6bfdd3a791352257a3c539/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidFileHandle.java#L217)

abueide:
@PokeMMO I don't understand what you mean that it will break the promise of checking the filesystem relative to the application's root directory first. This change should only affect the api in cases where the assets.index file exists (e.g. you implemented a task in your build system to generate the assets.index). It works exactly the same as before if no assets.index file is found, and it only adds non-breaking functionality in places where it is found.

The index could be cached, but I don't really know how to do it (this is my first PR and I haven't actually used LibGDX extensively). I'll probably look into it later if it would be a decent performance benefit.

Wow yeah that android code is pretty... ugly. I think this would help a lot with that in cases where the asset index is being used.

PokeMMO:
>I don't understand what you mean that it will break the promise of checking the filesystem relative to the application's root directory first. 

[On the desktop, if the file is not found, then the classpath is checked.](https://github.com/libgdx/libgdx/blob/a4805d6a017b80622d6bfdd3a791352257a3c539/gdx/src/com/badlogic/gdx/Files.java#L35)

If the asset index exists, list will incorrectly neglect to list files on the main filesystem. Things like mods or other dynamic content which would not exist in the asset index.

This may be an acceptable consequence of using the asset index, but it should be documented if no solution to this is added.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -507,12 +528,23 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+	  * See {@link #list()} documentation for more information about using an Asset Index
+	  */
 	public boolean isDirectory () {
-		if (type == FileType.Classpath) return false;
-		return file().isDirectory();
+		 if (type == FileType.Classpath)
+			  return false;
+		 if (type == FileType.Internal) {
+			  for (String indexpath : assetIndex()) {
+					if (indexpath == path() + "/")
```
Darkyenus: This will probably [never return true](https://stackoverflow.com/questions/767372/string-equals-versus#767379) and generates a lot of garbage.

`indexpath.length() + 1 == path && indexpath.startsWith(path) && indexpath.endsWith("/")` will do the same comparison without any allocations, but caching `path() + "/"` and doing the comparison with `equals()` will also work.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -734,4 +766,8 @@ static private void copyDirectory (FileHandle sourceDir, FileHandle destDir) {
 				copyFile(srcFile, destFile);
 		}
 	}
+
+	static private String[] assetIndex(){
+		 return Gdx.files.internal("assets.index").readString().replace('\\', '/').split("\n");
```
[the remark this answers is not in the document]
abueide: doesn't this have more to do with readString itself than my code? it says it uses the platform's default charset which sounds better then forcing users to use a particular one.

/** Reads the entire file into a string using the platform's default charset.
	 * @throws GdxRuntimeException if the file handle represents a directory, doesn't exist, or could not be read. */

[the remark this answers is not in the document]
PokeMMO: Yes, but the problem is if you compile it using UTF8, but the client which runs it uses GBK (Common default character set in china) you will get a different result.

```java
byte[] b = "Tĕsţīng".getBytes("UTF-8");
String s = new String(b, "GBK");
System.out.println(s);
```

Will return `T臅s牛墨ng`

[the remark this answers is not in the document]
abueide: How should I fix it? I just specify UTF-8 as an argument to readString and then add it to the asset index spec and make sure the gradle task generates a UTF-8 file no matter what the default charset is?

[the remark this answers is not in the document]
PokeMMO: That should work.

build.gradle
```diff
@@ -9,7 +9,9 @@
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, eit
```
intrigus: That's probably not an intentional change.

extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
```diff
@@ -97,6 +98,35 @@ public static void addProject(Language language, ProjectType project, List<Depen
 		write(wr, "}");
 	}
 
+	public static void addAssetIndexTask(String assetsDir, BufferedWriter wr) throws IOException {
+		 /**
```
Tom-Ski: Please use the libgdx formatter on any and also only lines you change. Also please follow proper javadoc guidelines for documenting methods.

gdx/src/com/badlogic/gdx/files/FileHandle.java
```diff
@@ -507,12 +532,24 @@ public void writeBytes (byte[] bytes, int offset, int length, boolean append) {
...
+	  *
+	  * See {@link #list()} documentation for more information about using an Asset Index
+	  */
 	public boolean isDirectory () {
-		if (type == FileType.Classpath) return false;
-		return file().isDirectory();
+		 if (type == FileType.Classpath)
+			  return false;
+		 if (type == FileType.Internal) {
+			  for (String indexpath : assetIndex()) {
```
Tom-Ski: This should be cached. The internal dir is not going to change after first run, many calls to isDirectory will generate a lot of garbage.

abueide:
So I created a directory called test in my assets folder and running with the new FileTree update, it only took 394 ms to call list on a directory with 100,000 10 byte files! ... compared to the 145 ms it took before I wrote and used the FileTree class. I'm not sure how much caching the objects into the FileTree affects GC performance. Should I abandon this approach and just cache the strings into a variable like I probably should have done? In theory the FileTree should be better because when using the find method it won't have to iterate over every string in the index except in the worst case scenario where the file your looking for is the deepest one and every directory you check is the last one alphabetically but my find function does some weird stuff with appending strings to figure out what directory to go down combined with recursion that might be affecting performance. Anyways I would appreciate thoughts and/or advice to improve the code.

extensions/gdx-setup/src/com/badlogic/gdx/setup/BuildScriptHelper.java
```diff
@@ -97,6 +97,34 @@ public static void addProject(Language language, ProjectType project, List<Depen
...
+	public static void addAssetIndexTask (String assetsDir, BufferedWriter wr) throws IOException {
+		write(wr, "/**");
+		write(wr, "* This task generates an index of internal asset files so you can can call FileHandle#list() on them. This is "
+			+ "useful if you ");
+		write(wr, "* were going to load all assets in a certain folder or of a certain type without having to load them "
+			+ "explicitly 1 by 1 ");
+		write(wr, "* in your code. This task can safely be deleted if you don't need it.");
+		write(wr, "*/");
+
+		write(wr, "task indexAssets() {");
```
Tom-Ski: This is still the slow task?

MobiDevelop:
Long ago (probably the last time this topic was discussed), I made a Gradle plugin for listing internal assets and a simple helper class to read it. I'm not terribly convinced that the framework needs to have this as a built-in.

abueide:
@MobiDevelop The fact that I didn't know that after researching the problem and making this PR kind of says to me that it does, especially since its not anywhere in the documentation. Personally, I don't really see any reason to not implement it given that it has no breaking changes, and to revert to the old version you just delete the gradle task and assets.index file from your project.

MobiDevelop:
Maybe, or the fact that it has come up pretty infrequently over the years suggests that it isn't as big a problem as one might think. If others want it as part of the framework, that's fine. I just don't think the complexity is warranted.

abueide:
Personally the reason I wanted it was so can write my asset loader to load the assets for certain levels at specific times recursively without having write them out 1 by 1. I don't really think it adds too much complexity it only changes some code in FileHandle and a small task in gdx-setup.

soundasleep:
The fact that a `Gdx.files.internal()` can't be used to list directory contents bit me today, but I think I've found a solution that works on desktop.

It seems the [Reflections library](https://github.com/ronmamo/reflections) can be used in both Eclipse (during dev) and with JPackage (with .jars). As an example, this seems to load all `.properties` files across all .jars that provide an `org.*` class, and can return a `Gdx.files.classpath()` entry:

```java
Reflections reflections = new Reflections(new ConfigurationBuilder().forPackage("org").setScanners(Scanners.Resources));

Set<String> resources = reflections.get(Scanners.Resources.with(".*\\.properties"));
for (String s : resources) {
	FileHandle handle = Gdx.files.classpath(s);
	System.out.println(String.format(">> %s, exists = %s, isDirectory = %s, file = %s, path = %s", handle, handle.exists(), handle.isDirectory(), handle.file(), handle.file().getAbsolutePath()));
}
```

This [might not work on Android](https://github.com/ronmamo/reflections/issues/339), and NOT tested outside of Windows 10 + Java 8.

But perhaps this could be a nice workaround to introduce into libgdx?

tommyettinger:
The general solution that gets linked on the Discord when this question comes up is [this article by @lyze237](https://lyze.dev/2021/04/29/libGDX-Internal-Assets-List/). I believe Lyze's solution works on all platforms because it only adds a .txt asset (containing the list of files) at compile-time, and all platforms can load .txt assets from the internal folder.

Issue #3244, opened 2015-06-30
Increase source level in gdx-setup


nooone:
AFAIK the android build tools, as well as GWT support Java 7 these days. A project generated by gdx-setup is still set to 1.6 though. For example [here](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle#L76).

IMHO it should be safe to set it to 1.7. @MobiDevelop says that he already works with 1.7 for a while and didn't run into problems.

FyiurAmron:
ain't this a duplicate of https://github.com/libgdx/libgdx/issues/1662 ? Also, http://stackoverflow.com/questions/7153989/java-7-language-features-with-android & the fact that AFAIR libGDX supports (at least in theory) up to API Level 8 - would Java 7-styled code be runnable on API Level 8-era machines?

nooone:
Kind of a duplicate, yes. Times have changed though since this other issue. @MobiDevelop already pointed out that GWT 2.6.0 (which we are using now) supports Java 7. The Android Build Tools seem to do so since version 19.

AFAIK, all Java 7 language features, but try-with-resources, are working as well on older android versions. Try-with-resources works only on API 19+, because the AutoClosable interface was missing.

MobiDevelop:
I think we'd have to switch to GWT 2.7.0 to be able to actually use Java 7
language features there. Unfortunately, GWT 2.7.0 has some issues that make
release builds incredibly slow.

Not a show stopper, but something to note.

nooone:
Only for release builds though? That would probably not be a huge problem.

I haven't tried it myself yet, but I read everywhere that 1.6.0 supports Java 7 language features. What exactly do you mean by "actually use Java 7"? What's not working with GWT 2.6.0?

Tom-Ski:
Not just release builds from my tests.

Tom-Ski:
Going to leave this 'til we can update GWT.

PR #5370, opened 2018-09-12
Upgrade source compatibility to Java 7

obigu:
It should be safe now to upgrade source compatibility to Java 7 

Related https://github.com/libgdx/libgdx/issues/3244

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle
@@ -92,8 +92,8 @@ eclipse {
     }
 
     jdt {
-        sourceCompatibility = 1.6
-        targetCompatibility = 1.6
+        sourceCompatibility = 1.7
+        targetCompatibility = 1.7
     }
 
     classpath {
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/core/build.gradle
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/core/build.gradle
@@ -1,6 +1,6 @@
 apply plugin: "%LANG%"
 
-sourceCompatibility = 1.6
+sourceCompatibility = 1.7
 [compileJava, compileTestJava]*.options*.encoding = 'UTF-8'
 
 sourceSets.main.java.srcDirs = [ "src/" ]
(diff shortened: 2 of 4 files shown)


MrStahlfelge:
It is safe. Should get merged.

MobiDevelop:
It's only safe if minSdkVersion >= 19. Prior to 19, only language features are safe (as opposed to APIs).

obigu:
> It's only safe if minSdkVersion >= 19. Prior to 19, only language features are safe (as opposed to APIs).

@MobiDevelop Could you please be specific about which APIs are not safe to use that require minSDK >= 19 using latest Android build tools?

MobiDevelop:
@obigu Any classes or methods that were added in Java 7 are not safe. Only language features (e.g. diamond operators, string switch cases, multi-catch, and, more recently, try-with-resources) on Android versions lower than 4.4 (API 19).

That said, I think we may be nearing a time where we should stop supporting those old versions.

obigu:
@MobiDevelop Oh, I see what you mean, yes, but I don't think that's an issue, not the whole Java APIs are supported anyway, not even from Java 5 :). IDEs (at least IntelliJ and AS) also help avoiding their usage with nice warnings.

To be honest, I think that's not a reason not to merge this, those APIs may never be implemented on Android.

I do agree with increasing the minSDK of libGDX, API 15 would already help us avoid some of the current "hacks".

MrStahlfelge:
> That said, I think we may be nearing a time where we should stop supporting those old versions.

We should go to 15 or 16, as [Play Store is discontinuing support for older versions](https://android-developers.googleblog.com/2018/12/google-play-services-discontinuing.html). If you really have a game with a massive user base with older versions (tbh, I doubt that), than there is nothing wrong to just stay with an older libGDX version.

obigu:
> > That said, I think we may be nearing a time where we should stop supporting those old versions.
> 
> We should go to 15 or 16, as [Play Store is discontinuing support for older versions](https://android-developers.googleblog.com/2018/12/google-play-services-discontinuing.html). If you really have a game with a massive user base with older versions (tbh, I doubt that), than there is nothing wrong to just stay with an older libGDX version.

I'd say 16 would be the right choice (if we are taking the decision to upgrade it makes sense to go in line with Google (the API 15 userbase is tiny and as you say it's always an option to use libGDX 1.9.9 that supports down to API 9).

obigu:
Fixed in https://github.com/libgdx/libgdx/commit/1e82c12a4b5e2cfa552f5e806ed21eda56ff2243

PR #5372, opened 2018-09-14
Add overriden writePNG method with flipY as parameter

obigu:
PNG libGDX class has an extremely efficient way to flip the Pixmap. Currently the writePNG utility method assumes the Pixmap is already flipped which is quite a heavy task if you need to flip it manually before. For example taking a screenshot using libGDX classes does this https://github.com/libgdx/libgdx/wiki/Taking-a-Screenshot. 

The new overriden method makes taking a screenshot more efficient (30ms improvement on a Nexus 5X) and simplifies the code to 3 lines:

`Pixmap pixmap = ScreenUtils.getFrameBufferPixmap(0, 0, Gdx.graphics.getWidth(), Gdx.graphics.getHeight());`
`PixmapIO.writePNG(fh, pixmap, true);`
`pixmap.dispose();`

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/PixmapIO.java
+++ b/gdx/src/com/badlogic/gdx/graphics/PixmapIO.java
@@ -57,11 +57,11 @@ public class PixmapIO {
 
 	/** Writes the pixmap as a PNG with compression. See {@link PNG} to configure the compression level, more efficiently flip the
 	 * pixmap vertically, and to write out multiple PNGs with minimal allocation. */
-	static public void writePNG (FileHandle file, Pixmap pixmap) {
+	static public void writePNG (FileHandle file, Pixmap pixmap, boolean flipY) {
 		try {
 			PNG writer = new PNG((int)(pixmap.getWidth() * pixmap.getHeight() * 1.5f)); // Guess at deflated size.
 			try {
-				writer.setFlipY(false);
+				writer.setFlipY(flipY);
 				writer.write(file, pixmap);
 			} finally {
 				writer.dispose();
@@ -69,6 +69,10 @@ public class PixmapIO {
 		} catch (IOException ex) {
 			throw new GdxRuntimeException("Error writing PNG: " + file, ex);
 		}
+	}
+
+	static public void writePNG (FileHandle file, Pixmap pixmap) {
+		writePNG(file, pixmap, false);
 	}
 
 	/** @author mzechner */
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/PixmapIO.java
```diff
@@ -57,11 +57,11 @@ static public Pixmap readCIM (FileHandle file) {
 
 	/** Writes the pixmap as a PNG with compression. See {@link PNG} to configure the compression level, more efficiently flip the
 	 * pixmap vertically, and to write out multiple PNGs with minimal allocation. */
```
Darkyenus: This piece of javadoc is no longer relevant then.

obigu: Thank you. I've added new javadoc for the new method and kept the original javadoc in the old one.

NathanSweet:
Why do this for `setFlipY` and not `setCompression`?

obigu:
> Why do this for `setFlipY` and not `setCompression`?

I can add compression as a parameter, do you want me to add it? The reason for flipY is that it's a common use case.

NathanSweet:
I think we might as well have both flipY and compression (for a total of two `writePNG` methods).

obigu:
> I think we might as well have both flipY and compression (for a total of two `writePNG` methods).

Done. I was thinking, shouldn't this method throw IOException as a checked Exception instead of catching it to throw a RuntimeException? It would seem appropriate not to catch it and add it to the method's signature so that the caller handles it appropriately (it may not be critical such as taking a screenshot)

NathanSweet:
`writePNG` was a convenience to make the common case simple. The idea was to use PixmapIO.PNG directly if you want to do something more sophisticated. Having `writePNG` at all is a bit questionable since we don't really need 2 APIs to do the same thing. I'd say `writePNG` should not throw an exception, since you are really only using that to avoid typing and adding try/catch is more typing.

obigu:
@NathanSweet Fair enough. I've thought about a potential issue when adding `compression`. The writePNG implementation has a magic number (1.5) to estimate the deflated size to set buffer size. That number may have been calculated (or may have not) based on the compression in which case it would have been set for `DEFAULT`. Do you know if it's safe to let it as it is for any compression level?

NathanSweet:
@obigu Sorry I didn't see your question sooner. It's no problem if the estimate is wrong, the buffer will resize larger or be unused.

Issue #5373, opened 2018-09-15
ParticleEffect modification broke Particle Effect on GWT module #5280


julienvillegas:
Issue "Life attribute on Particle Emitter is not applied to particles individually" with ID #5280 broke Particle loader for the HTML/gwt module

#### Issue details
A new parameter (independent) has been added to the ParticleEffect configuration file ".p".
In order to ensure backward compatibility the lifeValue.loader in class ParticleEmitter
has been updated with a mark(100) followed by a reset (see below).
Problem is,

- On Desktop, Android, calling that works.
- On GWT this triggers an Exception as there is no implementation for the BuffereredReader.mark() method. This calls by default the method on Reader.java which by default raises an exception.

In class ParticleEmitter:
```
public void load (BufferedReader reader) throws IOException {
			super.load(reader);
			// For backwards compatibility, independent property may not be defined
			reader.mark(100);
			String line = reader.readLine();
			if (line == null) throw new IOException("Missing value: " + "independent");
			if (line.contains("independent"))
				independent = Boolean.parseBoolean(readString(line));
			else
				reader.reset();
		}
```

#### Reproduction steps/code

Run the Out of the box Tests on GWT, select the following test com.badlogic.gdx.tests.ParticleEmitterTest

You will get the Exception outlined in the stacktrace section below.

The issue does not happen with the android or Desktop module (which is probably why this went undetected).

#### Version of LibGDX and/or relevant dependencies

LibGDX Version: 1.9.9-SNAPSHOT

#### Stacktrace
```java
Uncaught Error: com.badlogic.gdx.utils.GdxRuntimeException: Error loading effect: data/test.p
    at rLd.JKd [as dyb] (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at rLd.SKd [as lyb] (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at rLd.EKd (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at rLd.eLd (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at rLd.lLd (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at new rLd (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:159)
    at ZX.jY [as zt] (com.badlogic.gdx.tests.gwt.GdxTestsGwt-0.js:120)
...
[...]

PR #5374, opened 2018-09-16
Fix #5373 - Particle effects using old format fail meaningfully on GWT

obigu:
Fix for #5373

In addition to providing info on how to update the particle effect files to GWT users, test .p descriptor files have been updated to the latest format. A line in CHANGES has been also added to warn users about the breaking change.

EDIT: Set "Hide whitespace changes" to true in Github "Files changed" tab to compare.

PR #5376, opened 2018-09-16
Ensure GWT backward compatibility for #5280 and #5373

julienvillegas:
This is to ensure the newly added `inpendendent` parameter (added here #5280) does not break the gwt module (as described here #5373)
Instead of using mark() and reset() which are not supported we read the line,
if it contains the new parameter we use it, if not we move on..

Tested on Desktop and as HTML module.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java
@@ -16,16 +16,16 @@
 
 package com.badlogic.gdx.graphics.g2d;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.Writer;
-import java.util.Arrays;
-
 import com.badlogic.gdx.graphics.GL20;
 import com.badlogic.gdx.math.MathUtils;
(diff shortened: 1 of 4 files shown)


julienvillegas:
Ready to merge

obigu:
Tests .p descriptor independent parameter should be set to false. by default. I also believe it's a good idea to update them to the latest version as in https://github.com/libgdx/libgdx/pull/5374 (just open and save again on latest Particle Editor).

julienvillegas:
updated p files (opened with editor, saved, ran test again..).
ready to merge.

obigu:
There still seems to be some `independent` paramenters set to `true`

julienvillegas:
That’s ok I would think. 
Any reason they should all be set to false?
Probably to have some set to true too to test.

obigu:
Those tests were made when as independent was set to `false`, the question is why should any test be changed and set to true, the point of the tests is to verify things keep working as usual. Unless there's a good reason they should be kept as they were.

julienvillegas:
as you wish... 
done

julienvillegas:
Not accepted

PR #5378, opened 2018-09-23
Fix pixmap packer concurrency issue in splits

binbitten:
Fixes an issue where using PixmapPacker in multiple threads would cause incorrect split values.


Darkyenus:
Just a general note: libGDX is not thread safe (often not even regarding multiple separate instances) unless specified explicitly. Make sure that the whole class is indeed safe to use to use instances on separate threads. And if it is, it may be worth documenting it so that others can rely on it and future versions won't break it.

PR #5390, opened 2018-10-06
[Scene2d] Fixed NPE for Cell.uniform

balepc:
uniformX, uniformY being Boolean and uninitialized causes many crashes on iOS.

No need for Boolean, primitive type is enough here.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Cell.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Cell.java
@@ -28,7 +28,7 @@ public class Cell<T extends Actor> implements Poolable {
 	Integer align;
 	Integer expandX, expandY;
 	Integer colspan;
-	Boolean uniformX, uniformY;
+	boolean uniformX, uniformY;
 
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Cell.java
```diff
@@ -971,8 +971,8 @@ void merge (Cell cell) {
 		if (cell.expandX != null) expandX = cell.expandX;
 		if (cell.expandY != null) expandY = cell.expandY;
 		if (cell.colspan != null) colspan = cell.colspan;
-		if (cell.uniformX != null) uniformX = cell.uniformX;
-		if (cell.uniformY != null) uniformY = cell.uniformY;
```
Darkyenus: This is not equivalent and is a reason why `Boolean` is used instead of `boolean`. What crashes are you experiencing? None of the changed lines could cause a crash, regardless of the value in `uniformX/Y`.

Tom-Ski:
Please provide crash report and reproduction steps that cause a crash on ios.

PR #5398, opened 2018-10-12
Fix support for FontAwesome

guichazan:
For roboto.ttf, missingGlyph loaded in line 413 is not set because width/height are 0. However, when loading FontAwesome, missingGlyph is set, and then data.hasGlyph(c) returns true for all characters, even before they are loaded.


vbousquet:
Just a +1 to this patch, since I faced the same bug (BlackChancery.ttf), but any ttf font with an outline defined for .notdef glyph should have the problem.

NathanSweet:
Sorry for the trouble guys! `hasGlyph` is tricky. I fixed the issue before seeing this PR. Cheers!

Issue #5399, opened 2018-10-13
Incorect color processing for SpriteDrawable


vbousquet:
#### Issue details
Transform `draw` method from SpriteDrawable modify the underlying sprite color, causing incorrect rendering.

#### Reproduction steps/code
SpriteDrawable will not process color the same way if called using TransformDrawable draw method or called using draw method from Drawable.

[Untransformed draw](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/SpriteDrawable.java#L45-L47) uses : 
```java
		Color spriteColor = sprite.getColor();
		temp.set(spriteColor);
		sprite.setColor(spriteColor.mul(batch.getColor()));
```

while [Transformed draw](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/SpriteDrawable.java#L60-L61) uses : 
```java
		temp.set(sprite.getColor());
		sprite.setColor(temp.mul(batch.getColor()));
```
hence, modifying `temp` value, which is then applied to sprite, loosing the sprite original color.

The fix should be to use the code from untransformed draw method for both methods.

#### Version of LibGDX and/or relevant dependencies
1.9.9-SNAPSHOT

#### Stacktrace
Not applicable

#### Please select the affected platforms
- [X] Android
- [X] iOS (robovm)
- [X] iOS (MOE)
- [X] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS
(all checked since they are logically concerned, but only tested on Android and Desktop)

PR #5401, opened 2018-10-13
Fix TransformDrawable's draw method modifying the sprite's color

vbousquet:
See issue #5399


NathanSweet:
I fixed the issue before seeing this PR. Thanks, and sorry for the trouble. Cheers!

Issue #5404, opened 2018-10-15
Feature request: Box2D: Collision manifolds between shapes


dlon:
I think these would occasionally be useful in addition to the existing AABB and raycasting querying, as neither of those allows you to test arbitrary polygons, for example.

Perhaps these could be static functions in a `Collision` class.

From b2Collision.h:
```C++
/// Compute the collision manifold between two circles.
void b2CollideCircles(b2Manifold* manifold,
					  const b2CircleShape* circleA, const b2Transform& xfA,
					  const b2CircleShape* circleB, const b2Transform& xfB);

/// Compute the collision manifold between a polygon and a circle.
void b2CollidePolygonAndCircle(b2Manifold* manifold,
							   const b2PolygonShape* polygonA, const b2Transform& xfA,
							   const b2CircleShape* circleB, const b2Transform& xfB);

/// Compute the collision manifold between two polygons.
void b2CollidePolygons(b2Manifold* manifold,
					   const b2PolygonShape* polygonA, const b2Transform& xfA,
					   const b2PolygonShape* polygonB, const b2Transform& xfB);

/// Compute the collision manifold between an edge and a circle.
void b2CollideEdgeAndCircle(b2Manifold* manifold,
							   const b2EdgeShape* polygonA, const b2Transform& xfA,
							   const b2CircleShape* circleB, const b2Transform& xfB);

/// Compute the collision manifold between an edge and a circle.
void b2CollideEdgeAndPolygon(b2Manifold* manifold,
							   const b2EdgeShape* edgeA, const b2Transform& xfA,
							   const b2PolygonShape* circleB, const b2Transform& xfB);

/// Clipping for contact manifolds.
int32 b2ClipSegmentToLine(b2ClipVertex vOut[2], const b2ClipVertex vIn[2],
							const b2Vec2& normal, float32 offset, int32 vertexIndexA);

/// Determine if two generic shapes overlap.
bool b2TestOverlap(	const b2Shape* shapeA, int32 indexA,
					const b2Shape* shapeB, int32 indexB,
					const b2Transform& xfA, const b2Transform& xfB);
```

dlon:
Closing due to PR

PR #5407, opened 2018-10-16
Box2D: Collision manifolds

dlon:
#5404 

I don't know if you're interested in this, but I wanted it anyhow so I'll check.

If so, I guess I'll add a patch for GWT.


pixartist:
Can this be merged? It's been open 2 years and seems like a huge oversight

Frosty-J:
Only if it gets tested on more platforms. If need be, I can check Linux and Android without much hassle, but not iOS (or macOS, but I don't see much reason for it to go wrong there if all else is okay). I don't know a great deal about Box2D, but hopefully the linked test is good enough.

Tom-Ski:
, ***@***.****,On 21 Aug 2021 10:50, Frosty-J ***@***.***> wrote:
Only if it gets tested on more platforms. If need be, I can check Linux and Android without much hassle, but not iOS (or macOS, but I don't see much reason for it to go wrong there if all else is okay). I don't know a great deal about Box2D, but hopefully the linked test is good enough.

Frosty-J:
@Tom-Ski Did you mean to send that? Looks to be nothing more than the first revision of my comment in email form.

ice1000 (approved):

PR #5408, opened 2018-10-16
Minor JNI/C fixes

dlon:
Discovered these while compiling with VC:
* gdx2d.c: use char* instead of void* for pointer arithmetic.
* gdx-box2d: fixed return type for jniSetMaxLength (should be void).


PokeMMO:
Looks good to me.

PokeMMO (commented):
Looks good to me.

PR #5413, opened 2018-10-18
flag "prefersHomeIndicatorAutoHidden" is now configurable

Tienisto:
On new iPhones, all modern games hide the home indicator. In LibGDX there should be this option too.


Tienisto:
This should be done

obigu:
@Tienisto Have you tested this? It doesn't even compile...

Tom-Ski:
My bad if it doesn't, PR builder should check this

ryanastout:
I ran the current version of gdx-setup.jar and included Ios-moe (which is unselected by default).  
moe-core.jar and moe-ios.jar were both downloaded locally. I copied these files into libgdx/backends/gdx-backend-moe/libs overwriting the old versions which were last updated in January 2017 (see https://github.com/libgdx/libgdx/tree/master/backends/gdx-backend-moe/libs  )
and then I was able to successfully do mvn install with my local changes to libgdx. 

So it seems like newer moe-core.jar and moe-ios.jar should be committed into the libgdx/backends/gdx-backend-moe/libs/ directory in order for this change not to break things?

PR #5415, opened 2018-10-19
removed @Override from prefersHomeIndicatorAutoHidden

ryanastout:
I got a compilation error when making a local maven package. The error was 
[ERROR] bootstrap class path not set in conjunction with -source 1.6
[ERROR] libgdx/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSUIViewController.java:[111,1] error: method does not override or implement a method from a supertype

### DIFF
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSUIViewController.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSUIViewController.java
@@ -108,7 +108,6 @@ class IOSUIViewController extends GLKViewController {
 		return !app.config.statusBarVisible;
 	}
 	
-	@Override
 	public boolean prefersHomeIndicatorAutoHidden() {
 		return app.config.hideHomeIndicator;
 	}


obigu:
The problem is sue to https://github.com/libgdx/libgdx/pull/5413, that method is apparently not bound on Moe

Tienisto:
It seems that your MOE library is outdated.
When I build a new project with the gdx-setup.jar the method is actually there:

`@Generated
	@Selector("prefersHomeIndicatorAutoHidden")
	public native boolean prefersHomeIndicatorAutoHidden();`

The last commit of this class was on Sep 27, 2017:
https://github.com/multi-os-engine/moe-core/blob/moe-master/moe.apple/moe.platform.ios/src/main/java/apple/uikit/UIViewController.java

ryanastout:
I ran ./gradlew fetchNatives and then did mvn install and that's where I saw this compilation error. Is there another step necessary in order to build the local maven package?

ryanastout:
abandoning in favor of https://github.com/libgdx/libgdx/pull/5416

PR #5416, opened 2018-10-19
updated the moe-core.jar and moe-ios.jar files

ryanastout:
https://github.com/libgdx/libgdx/pull/5413 overrides a function which was added in a later version than the current version of these jar files. This update allows to build a maven package locally from source. 

To create these files, I ran the current version of gdx-setup.jar and included Ios-moe. The two jar files included here were downloaded to my local machine by that set up process. I copied the files to  backends/gdx-backend-moe/libs and created this pull request. 

I tested building a local maven package and the compilation error was fixed.


Tienisto:
This is actually a long overdue, (As you said) the gdx-setup.jar already download those jar files, so it should not break anything.

PR #5420, opened 2018-10-23
Added check for nullValue in JsonValue for primitive types

Kromzem:
Getting primitive type from JsonValue leads to exceptions if the given value is 'null'. 
Getting a String has a check for nullType and returning a default value instead. This check is missing for the primitive types (boolean, int, ...)
With this addition all request for values with default values should behave same.


EtK2000:
As much as I agree with not throwing an exception, a null value for a primitive is incorrect.
I haven't actually looking at the JSON standard, but I'd assume it isn't allowed.

That being said, I still think this should be accepted.

Darkyenus:
JSON does not specify any schema/typing, so "a null value for a primitive is incorrect" doesn't make sense. The "schema and types" are enforced only at deserialization time, so this change seems sensible, especially when precedented by `getString`.

PR #5427, opened 2018-10-26
Added debug mode to I18NBundle

wdavies973:
When developing a project, it's common to hardcode strings during testing rather than localizing them. Usually the developer will miss lots of unlocalized strings. A common method to spot unlocalized strings is to create another localization file with some placeholder value like "X" for every single value. This makes it easy to see what strings haven't been localized in game. To avoid having to maintain another localization file, the debug() method in I18NBundle will set all strings to placeholder to simplify this process.


costimuraru:
Useful! 👍

Issue #5385, opened 2018-09-29
Vector2.angleRad(Vector2) returns angle toward -y (instead of toward +y)


jcornaz:
#### Issue details
Documentation of `Vector2.angleRad(Vector2)` states:
> return the angle in radians of this vector (point) relative to the given vector. Angles are towards the positive y-axis. (typically counter-clockwise.)

So based on this statement `Vector2.Y.angleRad(Vector2.X)` should return `Math.PI / 2`.
But it returns `-Math.PI / 2`

#### Reproduction steps/code
```java
System.out.println(Vector2.Y.angleRad(Vector2.X)) // prints "-1.5707964" (which is wrong)

// not passing the argument works as expected
System.out.println(Vector2.Y.angleRad()) // prints "1.5707964" (which is correct)
```

#### Version of LibGDX and/or relevant dependencies
1.9.8

#### Please select the affected platforms
Should not be platform related. This is only math. But I observed the problem on Windows.

obigu:
It is actually correct, maybe the wording of the docs could be improved.

Basically the key point here is that order matters.
With an Y positive coordinate system, this method will return the counter clock-wise angle between the current Vector and the reference (argument) Vector. 

If you calculate the counter-clock angle between (0, 1) and (1, 0) you should get 270º which is what
`System.out.println(Vector2.Y.angleRad(Vector2.X))` returns.

On the other hand
`System.out.println(Vector2.X.angleRad(Vector2.Y))` returns 90º (in radians) as expected.

jcornaz:
>It is actually correct, maybe the wording of the docs could be improved.

Sorry I don't understand. How can `Vector2.Y.angleRad(Vector2.X))` returns a different value than `Vector2.Y.angleRad()`?

jcornaz:
By the way, as a side question, may I ask what is the rationale behind the fact that `angle` returns value between `0` and `360` while `angleRad` returns values between `-Math.PI` and `Math.PI`?

Why not return values between `-180` and `180` for `angle` for instance? It would be more consistent and simpler if `angle(x)` would always be equal to `angleRad(x) * MathUtils.radiansToDegrees`.

obigu:
I do agree that the docs can be improved because "relative to" may be misleading. Using your definitions, `Vector2.angleRad(Vector2 argument)`  returns the angle considering the current Vector is the reference. 
So, in your specific example, 
[...]

PR #5428, opened 2018-10-26
Fix #5388. Made Vector2 angle methods consistent

### LINKED ISSUE #5388: Make Vector2.angle and Vector2.angleRad consistent
### Current situation
* `Vector2.angle()` returns values between `[0, 360]`
* `Vector2.angle(Vector2)` returns values between `[-180, 180]` 
* `Vector2.angleRad()` returns values between `[-Math.PI, Math.PI]`
* `Vector2.angleRad(Vector2)` returns values between `[-Math.PI, Math.PI]`

There's two inconsistencies here.
1. Result range: Why only one of these function returns always positive values, while the 3 others don't?
2. Naming: Why some function are named with the angle unit, and not the others?

### Proposition:
1. Add `angleDeg()` and `angleDeg(Vetor2)` functions returning *both* values between `[-180, +180]`
2. Deprecate `angle()` and `angle(Vector2)` functions

This would make usage of angles in radians and degrees more consistent. Conversions would also be easier, since multiplying by `MathUtils.degreeToRadians` or `MathUtils.radiansToDegree` would always be enough.

### Reproduction steps/code

This:
```java
System.out.println(new Vector.2(0f, -1f).angleDeg()); // -90
System.out.println(new Vector.2(0f, -1f).angleRad()); // -Math.PI / 2
```

Would be preferable to:
```java
System.out.println(new Vector.2(0f, -1f).angle()); // 270
System.out.println(new Vector.2(0f, -1f).angleRad()); // -Math.PI / 2
```

### Version of LibGDX and/or relevant dependencies
1.9.8

### Please select the affected platforms
Platform is irrelevant here. This is math.

obigu:
Fix #5388 and #5385.

This PR contains two different changes to Vector2 angle methods:
1. Fixed `angle(Vector2)` and `angleRad(Vector2)` to return counter-clockwise angles as specified in their javadocs.
2. Made `angle(Vector2)` consistent to `angle()` to return values in the [0, 360) range.


obigu:
Well, there are 2 different points on the PR.

Point **1.** is a matter of fixing a wrong implementation which I think comes as top priority over backwards compatibility (otherwise the only alternative solution is to change the javadocs to reflect what it actually does, mark it as deprecated and create a new one that is consistent with the rest of methods which are counter-clockwise). Or do we just ignore it's wrong?

Regarding **2.** it does break backwards compatibility and it may be argued that for not a good enough reason, but not "heavily". There will be some cases in which it does break something but it won't in the majority of cases as angles are usually dealt with no matter the range. 

**If the PR is not being merged due to point **2.** I'm happy to remove it and keep only **1.****

ShibaBandit:
I would recommend just clarifying JavaDocs to the actual behavior.

obigu:
> I would recommend just clarifying JavaDocs to the actual behavior.

I assume you've read https://github.com/libgdx/libgdx/issues/5388. 

If we **just** change the javadocs (and do not deprecate tham and create new consistent ones) we would be validating/embracing a bad API and giving up on improving it. 

Both approaches (breaking backwards compatibility or deprecation) are valid and acceptable depending on the case, taking shortcuts isn't.

crykn:
Hm, my two cents on this issue: I thinks, in general, it makes sense to fix to fix these inconsistencies. However, to make it more obvious that something changed with the angle methods, I vote for renaming them to angleDeg(), as suggested in one of the initial issues. This way, the overall naming is more consistent and only `angleRad(Vector2)` returns different values than before.

obigu:
I think it's a good suggestion @crykn. By renaming them to `angleDeg`, not only we "force" the users to look for the new method in CHANGES but we also make the name more meaningful (I've found myself checking in docs if `angle()` returned `deg` or `rad` more than once). I'll update the PR.

Darkyenus:
How about keeping the old methods and deprecating them for a few versions (then delete)? The price is very low and it would be a good opportunity to update the javadocs of the old methods to mention that the replacements have a different behavior.

yuripourre:
@Darkyenus now that libgdx is going to have more frequent releases sounds a very good idea, indeed.

obigu:
@crykn @Darkyenus I've added new methods with "deg" suffix, deprecated the old ones and added javadocs were required. To make it consistent I've done it for all methods (not only `angle()` and `angle(Vector2)`. The only backwards compatibility change affects `angleRad(Vector2)` (which should not be a very commonly used method) and CHANGES have been updated to mention it especifically.

crykn (approved):
I don't think deprecating `rotate90 (int dir)` is needed, but I don't mind.

obigu:
> I don't think deprecating `rotate90 (int dir)` is needed, but I don't mind.

It's just an unnecessary method

Darkyenus:
I am also against deprecating `rotate90`. While it is a functional subset, it is computationally more efficient, and, more importantly, 100% precise.

obigu:
I don't agree with your point, I don't think it makes sense to accept this extra computational cost and lack of precision for any other rotation except for 90º (if that's critical for an app it should implement higher precision logic on their end). But from the thumbs up votes I guess most of you are in favout of keeping it so I go with the majority.

My question before making the change is, if this method is useful for the API, shall we create the Radians version (that rotates PI/2) for consistency?

Darkyenus:
Why break existing code over something that works well? I would keep the name as it is, 90 in context of rotations is self explanatory.

obigu:
Requested change made

gdx/src/com/badlogic/gdx/math/Vector2.java
```diff
@@ -385,6 +415,22 @@ public Vector2 rotateRad (float radians) {
 		return this;
 	}
 
+	/** Rotates the Vector2 by the given angle around reference vector, counter-clockwise assuming the y-axis points up.
```
mgsx-dev, line 418: could you please don't change methods order, it messing the diff and history, it looks like you added a new method rotateAround instead of modifying it.

gdx/src/com/badlogic/gdx/math/Vector2.java
```diff
@@ -316,19 +316,40 @@ public float crs (float x, float y) {
 		return this.x * y - this.y * x;
 	}
 
+
```
mgsx-dev, line 319: you can remove empty line (or use formater on modified blocs)

mgsx-dev (changes requested):

gdx/src/com/badlogic/gdx/math/Vector2.java
obigu: thanks, missed that one. Unfortunately using formatter is still not an option, last time I ran it on modified code I had to change it all over again due to format differences between IDEs/formatters.

gdx/src/com/badlogic/gdx/math/Vector2.java
obigu: the change of order on `rotateAround` was deliberate to try to keep the 3 versions of each method together for clarity but that's ok, I'll revert.

mgsx-dev (approved):

PR #5430, opened 2018-10-27
Made ViewController call super.viewDidAppear to allow hooks for screen tracking

almilli:
Using the Firebase Analytics api for iOS (https://firebase.google.com/) and using the MOE backend, the analytics will not generate "screen_view" events either automatically or manually due to the fact the IOSUIViewController overrides the `viewDidAppear(boolean)` method, but does not call `super.viewDidAppear(boolean)`.  So firebase thinks there are no screens showing and does not track the screen view.  In addition, it also does not let you manually set the screen information (to hook it up to scene2d screen views) because it requires at least one view controller appears before doing so.

This issue fixes the issue with Firebase Analytics not detecting that a view controller has appeared so even manual screen tracking gets ignored and you would see this log line:
```
2018-10-26 17:51:11.407667-0700 app-analytics-example[694:1165898] 5.11.0 - [Firebase/Analytics][I-ACS031003] setScreenName:screenClass: must be called after a view controller has appeared
```

### DIFF
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSUIViewController.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSUIViewController.java
@@ -59,6 +59,7 @@ class IOSUIViewController extends GLKViewController {
 
 	@Override
 	public void viewDidAppear (boolean animated) {
+		super.viewDidAppear(animated);
 		if (app.viewControllerListener != null)
 			app.viewControllerListener.viewDidAppear(animated);
 	}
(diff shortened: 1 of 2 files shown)


obigu:
Good find! Can you please add it to the RoboVM backend as well?

Just to complete, Apple docs warn super must be called if viewDidAppear is overridden https://developer.apple.com/documentation/uikit/uiviewcontroller/1621423-viewdidappear
> You can override this method to perform additional tasks associated with presenting the view. If you override this method, you must call super at some point in your implementation.

almilli:
@obigu I've added the one for RoboVM too.

PR #5436, opened 2018-11-01
Update TextureUnpacker.java (use original region size when unpacking atlas)

bigstupidx:
use original region size when unpacking atlas


mgsx-dev:
could you explain the fix ?

bigstupidx:
When packing small pics, texturePacker could erase the blank alpha part, this cause the small pic in the atlas is not the same size with the original size, but the altas txt file recorded this change.

example below 
pic cd_01 original size is 241x240, after packing, the size in the atlas changed to 34x34, (the other part  is blank alpha)


caidai.png
size: 1024,1024
format: RGBA8888
filter: Linear,Linear
repeat: none
cd_01
  rotate: false
  xy: 643, 645
  size: 34, 34
  orig: 241, 240
  offset: 108, 122
  index: -1

bigstupidx:
Yes, when unpacking the kind of atlas, the original image size is changed.
 
Example above when cd_01 is unpacked, the size changed to 34x34, not the original 241x240

mgsx-dev:
Oh sorry i just realized you fixed an issue with **Un**packing, I misread it.

maybe you could change your PR title to something like "use original region size when unpacking atlas"

PR #5443, opened 2018-11-05
Fixing scroll event rate

jpauty:
Opening a new PR with a proper branch, some improvements and squashed commits.

OSX generates many events when scrolling with the touch pad. This is
also the case for mouses that have a continuous wheel. Lwjgl3 reports
small movements with a scroll value lower than 1. The Lwjgl3 backend
rounds these values to 1 or -1. If there are many such "small" events
the scrolling becomes ultra fast.

The current fix accumulates the scrolling values reported by Lwjgl3
and sends a Gdx scroll event when the summed value is greater than
1. This is done until the accumulated value is lower than 1. This
approach also supports accelerations which are reported as scrolling
value greater than 1. These values become a stream of 1 or -1 scroll
events.

There is also an event which is triggered when the user starts acting
on the wheel or when the user changes the scrolling direction. This
gives a better reactivity, you don't have to scroll until you reach 1
for the scrolling to start.

Cheers,

Julien

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Input.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Input.java
@@ -92,7 +92,6 @@ public class Lwjgl3Input implements Input, Disposable {
 		@Override
 		public void invoke(long window, double scrollX, double scrollY) {
 			Lwjgl3Input.this.window.getGraphics().requestRendering();
-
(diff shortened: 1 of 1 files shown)


carlislefox:
Why are scrollYRemainder and lastScrollEventTime class level fields when they are only used by GLFWScrollCallback?

jpauty:
For no reason. I'll move them inside GLFWScrollCallback.

Thanks.

Julien

code-disaster:
We are currently testing your PR with our editor (on OSX and Windows), and it improves mouse wheel scroll behavior quite well - so I think we can merge it.

jpauty:
Great ! 👍

PR #5444, opened 2018-11-06
Fixed isKeyJustPressed not returning true on Android

braids:
### Overview
This fixes an issue on Android where Gdx.input.isKeyJustPressed() would not return a true value when pressing a key on a USB or Bluetooth keyboard.

#### AndroidInput.java
- When an _ACTION_DOWN_ Android event is caught by the _OnKey()_ method and the key does not exist in the _keys_ array, the key's index in the _justPressedKeys_ array is set to _true_. Otherwise, the same index in _justPressedKeys_ is set to _false_ while pressed.
- The flag _keyJustPressed_ is now set to _false_ at the start of the _processEvents()_ call regardless of the previous state. If a processed KeyEvent's index in the _justPressedKeys_ array is set to _true_, the _keyJustPressed_ flag is set to _true_.

### Builds Tested (w/ Test Platform)
- Android (Samsung Note 5)
- Desktop (Windows 10)
- HTML (Firefox)


EtK2000:
If you only modified Android code, there's no reason to test Desktop and HTML backends.
Tht being said, after looking into your code, it seems good.
I'll get around to confirming with my Bluetooth keyboard soon.

Tom-Ski:
Don't have ability to test, will need to wait til confirmation from someone else who can test this.

MrStahlfelge:
Should be same behaviour on emulator with hardware keyboard. I'll look into it this evening.

MrStahlfelge:
Emulator with hardware keyboard is not viable to test this because its behaviour is somehow a bit different from a hardware keyboard on the Android device. But you can test this by using the volume buttons @Tom-Ski - they have the same behaviour than the hardware keys on a connected keyboard.

I discovered the following while testing:
`isKeyJustPressed` indeed does not work like on other platforms. Unlike on Desktop and GWT, it returns true on every call while the key is hold and not just on first call. This is caused by Android sending multiple key down events while the key is hold. This is even clearly stated in Android's documentation:

> If the key is held sufficiently long that it repeats, then the initial down is followed additional key events with ACTION_DOWN and a non-zero value for getRepeatCount().

While this PR might fix the problem for `isKeyJustPressed`, there's another problem with this left: libGDX' `InputProcessor` gets multiple `keyDown`events. To fix both problems, the easiest way is to ignore Android's ACTION_DOWN events with a non-zero value for getRepeatCount(). I'll make a PR for this.

By the way, there's still another minor bug with `isKeyJustPressed` (and `isJustTouched` as well) left: It only works if an `InputProcessor` is set because the responsible code is guarded by a null-check:

https://github.com/libgdx/libgdx/blob/37e7bd72609309f5b449554cd4746fa24eaab16c/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java#L377

PR #5445, opened 2018-11-07
Added adaptive icon for API 26+

obigu:
API 26 introduced Adaptive Icons on Android (https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive). Many devices (such as Samsung devices) fallback to default icons when no adaptive icon is present. Google has started to push for their adoption by displaying a framed ugly version of the icon when no adaptive icon is present (https://www.androidpolice.com/2018/10/09/pixel-launcher-pixel-3-brings-assistant-button-back-forces-adaptive-icons-apps-apk-download/) and it's a matter of time until other manufacturers adopt similar policies.

With this change the app will use the old icons when running on < API 26 and the adaptive icon when running on >= API 26.

The size of the libGDX text has been set to 64px width to ensure is never cropped.


EtK2000:
Why did you externalize the white color and not the black or transparent colors?

obigu:
The white color is not externalized. libgGDX SVG asset doesn't have a background/background color. That gives the option not to create a vector drawable resource for the background and use a solid color resource instead which is what I've done here. 
The colors from the libGDX text are extracted automatically by Vector Asset when it generates the vector descriptor.

EtK2000:
I'm still trying to understand why the color.xml file exists, and if it already does why it doesn't contain all the colors from the svg.

obigu:
Because there's no point on extracting colors from an autogenerated file.

obigu:
> There is no need to create color.xml, use `@android:color/white` if you need white.

Can't be sure it will always be 0xFFFFFF, you know how Android device OEM love to mess with all they can  :)
https://twitter.com/mandybess/status/910556038461407232?lang=en

extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/res/values/color.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="white">#FFFFFFFF</color>
```
MrStahlfelge: Do not name it "white", but "ic_background_color" to prevent misunderstandings

MrStahlfelge:
@EtK2000 The reason for the extracted colors is the adaptive icon itself. Adaptive icons need to be seperated into foreground and background. Foreground drawable should be transparent while the background color defines how the circle (or whatever shape is used by the launcher) is filled. See [here ("adaptive icons are defined by 2 layers and a mask")](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive)

extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/res/values/color.xml
obigu: Thank you, good point, I'll make the change.

obigu:
@MrStahlfelge 

> Generally, the launcher icons should go into mipmap folder. I can understand you did not move that, because the current version's launcher icon is also placed in drawable, but perhaps this should be done now.

Yes, I saw that when reading about it but I wanted to increase the chances of this being merged :)

>Another suggestions is not to use an xml file, but pngs as ic_launcher_foreground. It is possible to use png as foreground and color as background. This way it is easier to understand and replace for beginners.

I'm not sure I agree on this one. I do agree using PNGs is a more beginner friendly as it doesn't require understanding of how Android works with vector drawables. But it's an inferior solution that exists as a fallback for those cases where it's not possible (or too heavy) to transform the icon to vector format (and requires new set of pngs for each resolution). Not defining it as a vector is, in my opinion, a missed opportunity to show beginners how it should be done.

EtK2000:
> @EtK2000 The reason for the extracted colors is the adaptive icon itself. Adaptive icons need to be seperated into foreground and background. Foreground drawable should be transparent while the background color defines how the circle (or whatever shape is used by the launcher) is filled. See [here ("adaptive icons are defined by 2 layers and a mask")](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive)

My question was regarding some but not all colors being in a separate file.
Basically, I didn't understand why colora didn't contain the colors from the foreground.
But it's unimportant to me, so I don't care much.

MobiDevelop:
Android Studio uses a vector when generating a new project (with PNG fallbacks) so I think this is probably fine.

PR #5450, opened 2018-11-15
Fixed nullpointer due to Vibrator instance being null on some devices

obigu:
User _hoangtunho2611_ has reported a crash when using `Gdx.input.vibrate()` on some Samsung devices using Android OS version 8.1 (https://www.badlogicgames.com/forum/viewtopic.php?f=11&t=28507&p=112037#p112037)

Even if this is probably a bug on those devices which should instead return an instance of Vibrator that returns `false` on `hasVibrator`, a cheap null check fixes the issue on the libGDX side.

Otherwise users who want to use vibration in a safe manner need to make the following check first `Gdx.input.isPeripheralAvailable(Peripheral.Vibrator)` which is not obvious in the way the API is designed.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -617,17 +617,20 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 
 	@Override
 	public void vibrate (int milliseconds) {
-		vibrator.vibrate(milliseconds);
+		if (vibrator != null)
+			vibrator.vibrate(milliseconds);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


hoangtunho2611:
Hi friends,
I see fix to Vibrator instance null with line code: if (vibrator != null) . But maybe it's not exactly.
I read guide here: https://developer.android.com/reference/android/os/Vibrator
I see method was deprecated in API level 26 : 

vibrate(long milliseconds)
This method was deprecated in API level 26. 
Use vibrate(VibrationEffect) instead.

obigu:
Grrr... you're right, I assumed the null pointer was on the libGDX side, read over it too fast. I'll close this PR.

Darkyenus:
But the underlying problem still remains, no?
LibGDX could detect platform level and translate the calls to the new Android API, to prevent similar crashes.

PR #5456, opened 2018-11-19
Update dtd in gwt.xml files

SimonIT:
According to [this](https://stackoverflow.com/questions/22865648/gwt-module-gwt-xml-xml-validation-warning) Stackoverlfow question these are the new correct urls to the dtd's for the gwt xml files.


SimonIT:
That's makes sense. IntelliJ says all files I changed have now CRLF. Should I really change back some files to LF? I think different line separators are not very beautiful

EtK2000:
You should use whatever LibGDX uses everywhere.
It's the standard thing to do, and I'm fairly sure part of the formatting guide (I'll check that soon).
It's just an advantage for getting the pull accepted.

SimonIT:
Ok, I will do it. It's a bit weird because on my PC the files got CRLF but if I download one of these files from the commit, it got only LF

EtK2000:
When I have to change the line ends I usually run it through a program like Notepad++ (on Windows) that allows you to change it easily.

EtK2000:
Now it's showing a lot of Java files that I'm fairly sure you didn't modify...

SimonIT:
Should I change _trunk_ to the version, in this case 2.8.0?

Issue #5471, opened 2018-12-12
Queue RemoveValue() Crash


Anuken:
#### Issue details
Calling `Queue#removeValue()` causes a crash when `removeLast()` is called beforehand.

#### Reproduction steps/code
```java
Queue<Integer> queue = new Queue<>();
queue.addFirst(1);
queue.addFirst(2);
queue.removeLast();
queue.removeValue(2, false);
```

This may also cause crashes in other scenarios. I believe the error is related to `indexOf()` returning the wrong value, e.g. 14 when it should be 0 in this case.

#### Version of LibGDX and/or relevant dependencies
1.9.9

#### Stacktrace
```java
java.lang.IndexOutOfBoundsException: index can't be >= size: 14 >= 1
	at com.badlogic.gdx.utils.Queue.removeIndex(Queue.java:220)
	at com.badlogic.gdx.utils.Queue.removeValue(Queue.java:213)
```

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [X] Linux
- [ ] MacOS

negativeview:
Confirmed issue on MacOS as well.

slebed:
I believe that this is the expected result.  If you add two items to the queue and perform removeLast(), your left with only one item in the queue.  RemoveValue(2, false) would be removing an item that doesn't exist.

Anuken:
@slebed That is not the case. Printing the queue's contents before removal reveals that 2 is the only item in the queue. 
After calling addFirst with 1 and 2, I get the queue `[1 2]`. Calling removeLast removes the tail, or `1`.

Even if `2` *wasn't* in the queue, documentation states that `@return true if value was found and removed, false otherwise`. A crash should not occur in any scenario.

NathanSweet:
Closed by #5472.

PR #5472, opened 2018-12-12
Change how we compute indexOf to account for a moving head value.

negativeview:
Fix for https://github.com/libgdx/libgdx/issues/5471

If head < tail, we weren't accounting for head when computing the index, resulting in the returned index being far too large.


intrigus:
Pinging @NathanSweet

NathanSweet:
Thanks!

PR #5475, opened 2018-12-13
setting playback rate of music and sound

SimonIT:
I have already implemented the option to set the playback speed for android and GWT.
I have problems implementing:
- IOS, because I do not have any option to run the IOS SDK.
- Desktop, because I have problems understanding the OpenAL.

I will try to make the feature work on Desktop.
Help with implementing the backends would be appreciated.


MrStahlfelge:
That's a great enhancement. Is the pitch affected by the speed?

mgsx-dev:
Cool ! a great enhancement would be to be able to change pitch as well for GWT. I know it's not related to this PR but since you updated sound-manager maybe it could be possible to implement it ?

SimonIT:
After some research, I read quite often that changing the pitch so i went with this solution on desktop. I found also a master thesis about this topic [here](https://www.audiolabs-erlangen.de/content/05-fau/professor/00-mueller/01-students/2011_DriedgerJonathan_TSM_MasterThesis.pdf), but that's too high for me.

Now only the ios backend is missing. I can't implement this because I don't own an apple device. It would be nice if someone can help me writing the code and test it for me.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/soundmanager2/SMSound.java
```diff
@@ -116,7 +116,15 @@ public native final void setPan (int pan) /*-{
 	public native final int getPan () /*-{
 		return this.@com.badlogic.gdx.backends.gwt.soundmanager2.SMSound::jsSound.pan;
 	}-*/;
-	
+
+	public native final void setPlaybackRate (float playbackRate) /*-{
+        this.@com.badlogic.gdx.backends.gwt.soundmanager2.SMSound::jsSound.setPlaybackRate(playbackRate)
```
intrigus: Indentation is wrong.

gdx/src/com/badlogic/gdx/audio/Music.java
```diff
@@ -84,6 +84,13 @@
 	/** Returns the playback position in seconds. */
 	public float getPosition ();
 
+	/** Sets the playback rate
+	 * @param playbackRate rate: 1 => normal */
```
intrigus: I guess you want to express that `playbackRate` should be greater than equal 1?
The wording could be better in my opinion.

Additionally you should document, that some backends only support `playbackRate`s that are smaller than 4 (e.g GwtMusic).

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtMusic.java
```diff
@@ -117,6 +117,18 @@ public float getPosition () {
 		return sound.getPosition() / 1000f;
 	}
 
+	@Override
+	public void setPlaybackRate(float playbackRate) {
+		if (playbackRate >= 0.5f && playbackRate <= 4.0f) {
```
intrigus: The `Music` interface says that `playbackRate`s should be greater equal than 1, so why allow lower values here?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/soundmanager2/SMSound.java
SimonIT:
```suggestion
		this.@com.badlogic.gdx.backends.gwt.soundmanager2.SMSound::jsSound.setPlaybackRate(playbackRate)
```

gdx/src/com/badlogic/gdx/audio/Music.java
SimonIT: I wanted to say that a playback rate of 1 is the normal speed
```suggestion
	 * range: Android + Desktop 0.5 to 2, GWT 0.5 to 4
	 * @param playbackRate < 1 slower, = 1 normal, > 1 faster */
```

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtMusic.java
SimonIT: That's a misunderstanding of my doc string (see the other review)

obigu:
Isn't changing the `pitch` the same as changing the `playback rate`? This is a quote from Android `SoundPool` docs (https://developer.android.com/reference/android/media/SoundPool)
> The playback rate allows the application to vary the **playback rate (pitch)** of the sound.

Considering the `Sound` interface already allows changing the `pitch`, I understand the aim of this PR should be to add it to `Music` and for consistency should keep the same terminology on both `Sound` and `Music`.

SimonIT:
Not always, soundmanager on gwt supports changing the playback rate without affecting the pitch (for my surprise).  The MediaPlayer on Android supports it too. But for desktop and the soundpool on android I couldn't find any solution without affecting the pitch

mgsx-dev:
@SimonIT did you considered using combination of OpenAL velocity (Doppler effect) and pitch to emulate it ?

for the javadoc you can be more verbose like "a rate of 1 play at normal speed, 2 at twice speed and 0.5 at half speed" it could avoid misunderstanding :-)

gdx/src/com/badlogic/gdx/audio/Music.java
SimonIT:
```suggestion
	 * range: Android and Desktop 0.5 to 2, GWT 0.5 to 4
	 * a rate of 1 play at normal speed, 2 at twice speed and 0.5 at half speed
	 * @param playbackRate new playback rate */
```

SimonIT:
better? I don't know exactly which type of warning do you mean?

SimonIT:
@mgsx-dev I tried setting `alSourcef(sourceID, AL_VELOCITY, playbackRate);`, but I didn't recognized any change and did not find any examples how to use. [Here](https://stackoverflow.com/questions/34664022/increasing-the-speed-of-speech-without-changing-the-pitch) is a Stackoverflow answer which says that's not possible to change the speed without changing the pitch, only with an own implementation of an WSOLA algorithm. So I searched for an WSOLA algorithm and found one [here](https://de.mathworks.com/matlabcentral/fileexchange/45451-waveform-similarity-and-overlap-add-wsola-for-speech-and-audio) for MathLab. But I never used MathLab so I have to the language if I want implement the algorithm here.

MrStahlfelge:
Probably you need to update this :)

SimonIT:
I'm closing this PR because in my opinion it became pretty pointless. On almost every backend is changing the playback rate impossible without changing the pitch or vice versa. So to get the same effect just change the pitch

PR #5477, opened 2018-12-14
Extra customization options for ImageProcessor and TexturePacker

piotr-j:
This simple pr allows for a bit easier customization of processing of each image. 
Due to external reason, *cough* spine *cough*, i want to skip whitespace stripping on some of the images.  

Also, different encoding in each file? :(


obigu:
I understand the approach is to make it as simple as possible but inheritance here is not ideal imo. I may be wrong but it seems to me that in this case we could benefit from having an ImageProcessor interface and add the option to set a different implementation to TexturePacker (current one would be default). If having an interface is not needed as changes on implementations are minimal, at least have a setter instead of the protected builder method `newImageProcessor`.

piotr-j:
Well, to truly change how any of this stuff works you need to copy pasted all of it more or less (and i did just that previously). The reason why i went with `newImageProcessor` instead of a setter is to be consistent with `newTexturePacker` in `TexturePackerFileProcessor`.

piotr-j:
Besides the inability to enable/disable whitespace stripping on per folder basis im quite happy with current implementation. Last few commits already made it quite a bit easier to change. From 3 copy pasted classes to one extended one in the package and a bit of reflection.

obigu:
> to one extended one in the package and a bit of reflection.

Better but definitely very improvable :)


> If you know of any other painpoints im listening.

I don't use TexturePacker but I can see it could definitely benefit from some decoupling.

NathanSweet:
I agree TexturePacker could use some clean up. That's a bit of work, would break people's current usage, and it works pretty OK even being cobbled together as it is, so I'm fine with adding slightly to the mess!

I've been meaning to do more for packing Spine meshes for ages. Maybe it's time...

NathanSweet:
@piotr-j What was that about encodings?

piotr-j:
I misspoke, meant line endings. Vast majority of the files seem to use crlf, but ImageProcessor uses lf.

NathanSweet:
Ah. I see CRLF for ImageProcessor (on Windows). Usually Git converts from LF in the repo to CRLF for Windows or LF elsewhere for you locally, then back to LF when pushing.

NathanSweet:
> Due to external reason, cough spine cough, i want to skip whitespace stripping on some of the images.
BTW, @piotr-j Spine 3.7.77-beta now supports whitespace stripping for mesh regions.
https://github.com/EsotericSoftware/spine-runtimes/issues/1232

Issue #5020, opened 2017-12-13
Feature Request Android: Possibility to define keys that are catched other than menu/back


MrStahlfelge:
Hello,

we have the possibility to catch the Android back and menu buttons by calling `setCatchBackKey`/`setCatchMenuKey`. However, all other keys are **not** catched on Android and there is no possibility to do so. The responsible code in `AndroidInput.onKey`:

		if (keyCode == Keys.BUTTON_CIRCLE) return true;
		if (catchBack && keyCode == android.view.KeyEvent.KEYCODE_BACK) return true;
		if (catchMenu && keyCode == android.view.KeyEvent.KEYCODE_MENU) return true;
		return false;

([Go to AndroidInput](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java))

The `return false` statement tells Android that the game did not process the event - this is not correct in many cases. However, it also doesn't matter in most cases, so we normally have no problem here.

But besides the menu or back key, there are some other keys that need to be catched when used: Volume Up/Down, Fast Foward buttons are some examples. When targetting for FireTV devices and using the buttons available on the FireTV Remote Controller, Amazon rejects the game submission if the game does not return that the key event was handled. (They give the reason that a media player in background will react on the key events, which is true.) There is no easy way to solve this problem at the moment.

I suggest to add an array with keycodes that are catc hed, change the code above to something like

		if (keyCode == Keys.BUTTON_CIRCLE) return true;
		return keysToCatch.contains(keyCode);

and add a method `setCatchKey(keyCode, boolean)` to `AndroidInput` and the core interface for adding and removing the catched keycodes. With this approach, all buttons that need to be catched can be set by the developer.

I would be happy to add the change, but before doing so I wanted to discuss this change here.

kalexmills:
<thinking-aloud>
What, if any, integration between Android and iOS should be considered? Could the same button have the same meaning on both platforms? Should we care?
</thinking-aloud>

MrStahlfelge:
Closing this due to the open PR

PR #5478, opened 2018-12-14
#5020 Possibility to define caught keys on Android

MrStahlfelge:
See #5020:

Besides the menu or back key, there are some other keys that need to be catched when used: Volume Up/Down, Fast Foward buttons are some examples. When targetting for FireTV devices and using the buttons available on the FireTV Remote Controller, Amazon rejects the game submission if the game does not return that the key event was handled. (They give the reason that a media player in background will react on the key events, which is true.) There is no easy way to solve this problem at the moment.

This PR adds the ability to set other catched keys than just back/menu.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -49,6 +49,7 @@ import com.badlogic.gdx.Input;
 import com.badlogic.gdx.Input.TextInputListener;
 import com.badlogic.gdx.InputProcessor;
 import com.badlogic.gdx.backends.android.AndroidLiveWallpaperService.AndroidWallpaperEngine;
+import com.badlogic.gdx.utils.IntArray;
 import com.badlogic.gdx.utils.IntSet;
 import com.badlogic.gdx.utils.Pool;
 
@@ -127,8 +128,7 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 	final Context context;
 	protected final AndroidTouchHandler touchHandler;
 	private int sleepTime = 0;
-	private boolean catchBack = false;
-	private boolean catchMenu = false;
+	private IntArray catchedKeys = new IntArray();
 	protected final Vibrator vibrator;
 	private boolean compassAvailable = false;
 	private boolean rotationVectorAvailable = false;
@@ -184,6 +184,10 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 		} else {
 			nativeOrientation = Orientation.Portrait;
 		}
+
+		// this is for backward compatibility: libGDX always catched the circle button, original comment:
+		// circle button on Xperia Play shouldn't need catchBack == true
+		catchedKeys.add(Keys.BUTTON_CIRCLE);
 	}
 
 	@Override
@@ -571,11 +575,7 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 			app.getGraphics().requestRendering();
 		}
 
-		// circle button on Xperia Play shouldn't need catchBack == true
-		if (keyCode == Keys.BUTTON_CIRCLE) return true;
-		if (catchBack && keyCode == android.view.KeyEvent.KEYCODE_BACK) return true;
-		if (catchMenu && keyCode == android.view.KeyEvent.KEYCODE_MENU) return true;
-		return false;
+		return catchedKeys.contains(keyCode);
 	}
 
 	@Override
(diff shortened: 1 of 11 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
```diff
@@ -127,8 +128,7 @@ protected TouchEvent newObject () {
 	final Context context;
 	protected final AndroidTouchHandler touchHandler;
 	private int sleepTime = 0;
-	private boolean catchBack = false;
-	private boolean catchMenu = false;
+	private IntArray catchedKeys = new IntArray();
```
intrigus: Is there any reason why you didn't use an IntSet instead?
It should be faster (`onKey()` may be called many times) than an IntArray and also easier to use in my opinion.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
```diff
@@ -571,11 +575,7 @@ public boolean onKey (View v, int keyCode, android.view.KeyEvent e) {
 			app.getGraphics().requestRendering();
 		}
 
-		// circle button on Xperia Play shouldn't need catchBack == true
-		if (keyCode == Keys.BUTTON_CIRCLE) return true;
```
intrigus: You could now in theory do this:
````java
setCatchKey(Keys.BUTTON_CIRCLE, false);
````
which would break this workaround.
See https://badlogicgames.com/forum/viewtopic.php?f=11&t=1317
and https://github.com/dwaite/imame4all/issues/203

Therefore I think this line should stay unchanged.

File: gdx/src/com/badlogic/gdx/Input.java [hunk 3]
```diff
@@ -742,6 +742,18 @@ private static void initializeKeyNames () {
 	/** @return whether the menu button is currently being caught */
 	public boolean isCatchMenuKey ();
 
+	/**
+	 * Sets whether the given key on Android should be caught. No effect on other platforms.
+	 * All keys that are not catched may handled by other apps or background processes. For example, media or volume
```
intrigus: All keys that are not catched may **be** handled by other apps or background processes. For example, media or volume

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -742,6 +742,18 @@ private static void initializeKeyNames () {
...
 
+	/**
+	 * Sets whether the given key on Android should be caught. No effect on other platforms.
+	 * All keys that are not catched may handled by other apps or background processes. For example, media or volume
+	 * buttons are handled by background media players if present. If you use these keys to control your game, they
+	 * must be catched to prevent unintended behaviour.
+	 *
+	 * @param keycode  keycode to catch
+	 * @param catchKey whether to catch the given keycode
+	 * @return
```
intrigus: There is no return value.

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -742,6 +742,18 @@ private static void initializeKeyNames () {
 	/** @return whether the menu button is currently being caught */
 	public boolean isCatchMenuKey ();
 
+	/**
```
intrigus: I'm not a native speaker, but is it correct to say "catched"?
My browser is underlining it in red.

tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
```diff
@@ -418,6 +418,11 @@ public boolean isCatchMenuKey () {
 			return input.isCatchMenuKey();
 		}
 
+		@Override
+		public void setCatchKey(int keycode, boolean catchKey) {
+
```
intrigus: For consistency's sake I'd change this to
````java
input.setCatchKey(keycode, catchKey);
````
Since the other methods in `GwtTestWrapper` are doing the same.

intrigus:
Clicked the wrong button, I wanted to request changes.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
MrStahlfelge: Just because I was not aware of it, will change it.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
MrStahlfelge: That's by intention. If someone needs to, the default behaviour can be changed by the developer.

gdx/src/com/badlogic/gdx/Input.java
MrStahlfelge: I'm also not native, caught in the act. Will change all this incidences.

MrStahlfelge:
@EtK2000 
I thought about adding it, and it is easy to implement. 
I skipped it because it is another change of the interface with "empty" `return false` implementations in all other backends and I don't see much benefit: I never used any of the `isCatch..` methods - either a key should be catched or not, depending on the game's state. 

What do the audience think? Do we need `isCatchKey(...)`?

MobiDevelop:
This looks okay to me. To be honest, I'd almost rather the specific methods for back and menu get removed in favor of calling the new methods.

EtK2000:
Maybe they can be depricated to keep it backwards compatible.

MrStahlfelge:
I did not remove the old methods because I was afraid that this won't get merged when backwards compatiblity is broken and I suspect many projects out there are using them. But of course, adding a deprecation tag or removing the old method is all fine for me so I'll do what the maintainers prefer.

Tom-Ski:
Merged manually. Thanks @MrStahlfelge

PR #5480, opened 2018-12-15
Call set root in constructor, extension of #5476

EtK2000:
Due to discussion of calling a function in a constructor I split this into 2 pulls.


NathanSweet:
I think it's OK for the default root provided by Stage to not use Stage `setRoot`. There's not much benefit and it's less annoying in case someone overrides `setRoot` (it would be called before a subclass constructor). Overriding `setRoot` is probably rare anyway but FWIW, we do try to avoid calling methods that are somewhat commonly overridden from constructors, eg Layout `pack` javadocs mention it and widget constructors instead use `setSize(getPrefWidth(), getPrefHeight());`.

PR #5482, opened 2018-12-16
implement geolocation

SimonIT:
All backends are implemented (except desktop). I tested GWT and Android and they work well. I can't test the ios backends (I got no apple device to compile) so it would be great if someone could test it. (And in case it doesn't work help me fixing.) I implemented ios with [this](https://github.com/robovm/robovm-samples/tree/master/ios/locateme) samples so i hope it works.

Since Android Marshmallow, the user has to enable the permissions for apps by hand. This can be a bit annoying to go into the system settings and turn them on. I wanted to add a method `requestPermission()` into `AndroidApplication.java` which ask the user for the permission (for example location) like [this ](https://developer.android.com/training/permissions/requesting). The problem is, libgdx is currently using _support-v4-19.0.1.jar_ as support library which doesn't offer these methods. Can I upgrade this lib or is there any reason for keep the old version? Should I add this change to this pr or make another one?


obigu:
> libgdx is currently using _support-v4-19.0.1.jar_ as support library which doesn't offer these methods. Can I upgrade this lib or is there any reason for keep the old version? Should I add this change to this pr or make another one?

Upgrading the support version on libGDX project is harder than it looks at first sight:
1. We build libGDX with a mix of Maven, Gradle and even Ant and that's the reason there are 2 jars (android and support libraries inside libs) used for compilation instead of a dependency in a Gradle config file.
2. Android support libraries were split into modules (around v24 iirc) which means it's not just replacing 1 .jar but a whole bunch of them.
3. Current Android minSDK is 9 and current minSDK for Android Support libraries (since 26 included) is API 14.
4. It looks like the only reason libGDX depends on Android Support v4 is for `android.support.v4.app.Fragment` on `AndroidFragmentApplication`. Ideally we should "only" depend on  `support-fragment` which (on v25.4) has the folowing dependencies by having a quick look at the pom (https://maven.google.com/com/android/support/support-fragment/25.4.0/support-fragment-25.4.0.pom).

IMO fully building libGDX with Gradle would be the best approach but it's a big task. If this is not viable and no solution is presented we can always copy all the required jars from android-supprt-fragment to the libs folder for the time being.

MrStahlfelge:
You should release this as a seperate extension lib in an own repo like gdx-facebook and other extensions do. It is really an interesting feature, but not needed in the core project as most projects will not make use of it. Use [this as an template](https://github.com/TomGrill/my-gdx-extension) and if you have further questions how to do so, you can contact me via the libGDX Discord server.

SimonIT:
@MrStahlfelge Gyroscope and so on are also in the core project and I think not every game is using them. It's also not that big addition, so I think an own extension isn't worth. And with an own extension I would have the problem that I can't keep up the ios code because I don't own a mac.

tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
```diff
@@ -623,6 +644,14 @@ public GdxTest instance () {
 			public GdxTest instance () {
 				return new IndexBufferObjectShaderTest();
 			}
+		}, new Instancer() {
+			public GdxTest instance() {
+				return new InputTest();
+			}
+		}, new Instancer() {
+			public GdxTest instance() {
```
intrigus: The gwt test list should be sorted alphabetically

PR #5489, opened 2018-12-22
adding usage of the new vibration api on android since sdk 26

SimonIT:
Fixes #5450


MobiDevelop:
We build the Android backend against a copy of the android.jar for API 25, which would have to be updated to at least the API 26 version for this to compile.

SimonIT:
I know^^ I thought libgdx would upgrade the api level in the future. Almost 40% of Android devices got a higher api version. https://developer.android.com/about/dashboards

MobiDevelop:
Updated the backend to build against Android 9.0, this will now compile.

Issue #5256, opened 2018-06-08
Libgdx is not compatible with GWT 2.8.2


patton73:
Please ensure you have given all the following requested information in your report.

#### Issue details
Compilation gives this error:

Exception in thread "main" java.lang.NoSuchMethodError: com.google.gwt.util.regexfilter.RegexFilter: method <init>()V not found
	at com.google.gwt.util.regexfilter.WhitelistRegexFilter.<init>(WhitelistRegexFilter.java:21)
	at com.google.gwt.dev.jjs.JJSOptionsImpl.<init>(JJSOptionsImpl.java:54)
	at com.google.gwt.dev.PrecompileTaskOptionsImpl.<init>(PrecompileTaskOptionsImpl.java:39)
	at com.google.gwt.dev.CompilerOptionsImpl.<init>(CompilerOptionsImpl.java:30)
	at com.google.gwt.dev.Compiler.main(Compiler.java:108)

#### Reproduction steps/code

Just use GWT 2.8.2 and try to recompile

#### Version of LibGDX and/or relevant dependencies
1.9.8

#### Please select the affected platforms
- [X ] HTML/GWT

MobiDevelop:
I'm working on it.

tommyettinger:
This also affects 2.8.1, by the way, but not 2.8.0.

niloc132:
This error happens a lot in normal GWT development as well - it means that somehow you have both GWT 2.8.0 and 2.8.1 (or 2.8.2) on the classpath at the same time, so some classes are conflicting.

Unless libgdx has a custom build of GWT internally, this is probably just a matter of fixing all dependencies - `gwt-user`, `gwt-dev` are the two main ones, but `gwt-servlet` might be required as well.

Gustl22:
Any workarounds for this? Is it possible to [update](https://github.com/libgdx/libgdx/commit/7ed40698018bb789916ffea6ca8b2a4f64961f92) libGDX to 2.8.2?

niloc132:
In normal GWT development, you must simply ensure that only one copy of each GWT jar is present on the classpath, and that each is exactly the same version - it is very likely the case here too, suggesting that you have some conflicting dependencies or some second copy of gwt (but in another artifact, etc). If you're using maven, try `mvn dependency:tree -Dverbose`, possibly share the result if it isn't clear right away. If using gradle, I believe the analog is something like `:dependencyInsight --dependency gwt-dev`, but I don't know gradle very well.

[...]

PR #5494, opened 2018-12-28
Setting the "real" clipboard in gwt

SimonIT:
Here's an implementation for setting the clipboard in gwt using the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API). For getting the clipboard I see no possibility because it is an asynchronous event which returns the content of the clipboard asynchronously. Or do you got an idea how to do it?

> Access to the contents of the clipboard is gated behind the Permissions API; without user permission, reading or altering the clipboard contents is not permitted.

For this I created a separate class for querying the permission. It could be also useful in the future for accessing for example the [Accelerometer ](https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer) which needs also permissions.

I added also a test for the clipboard which makes it easier to test it.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Clipboard.java
+++ b/gdx/src/com/badlogic/gdx/utils/Clipboard.java
@@ -20,7 +20,7 @@ package com.badlogic.gdx.utils;
  * @author mzechner */
 public interface Clipboard {
 	/** gets the current content of the clipboard if it contains text
-	 * for WebGL app, setting the system clipboard is currently not supported. It works only inside the app
+	 * for WebGL app, getting the system clipboard is currently not supported. It works only inside the app
 	 * @return the clipboard content or null */
 	public String getContents ();
 
--- a/gdx/src/com/badlogic/gdx/utils/Clipboard.java
+++ b/gdx/src/com/badlogic/gdx/utils/Clipboard.java
@@ -20,10 +20,13 @@ package com.badlogic.gdx.utils;
  * @author mzechner */
 public interface Clipboard {
 	/** gets the current content of the clipboard if it contains text
+	 * for WebGL app, setting the system clipboard is currently not supported. It works only inside the app
 	 * @return the clipboard content or null */
 	public String getContents ();
 
 	/** Sets the content of the system clipboard.
+	 * for WebGL app, clipboard content might not be set if user denied permission, setting clipboard is not synchronous
+	 * so you can't rely on getting same content just after setting it
 	 * @param content the content */
 	public void setContents (String content);
 }
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/ClipboardTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/ClipboardTest.java
@@ -1,3 +1,19 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
 package com.badlogic.gdx.tests;
 
 import com.badlogic.gdx.Gdx;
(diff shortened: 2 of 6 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/ClipboardTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Override
+	public void create() {
+		stage = new Stage();
+		Skin skin = new Skin(Gdx.files.internal("data/uiskin.json"));
+		textArea = new TextArea("", skin);
+		buttonCopy = new TextButton("Copy", skin);
+		buttonPaste = new TextButton("Paste", skin);
+
+		textArea.setSize(Gdx.graphics.getWidth() / 3f, Gdx.graphics.getHeight() / 3f);
```
lodenrogue: Can we put magic numbers into appropriately named constants?

SimonIT: Yes we could, but in my opinion it's unnecessary because it's only for the size of the textarea in the test

lodenrogue: For tests it makes sense. Thanks

SimonIT:
I found a possible solution for synchronising `navigator.clipboard.readText()`. The [await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) operator is IMO the easiest solution. But sadly gwt 2.8.0 seem not to support this operator. 
I get this error by running the test:
```
   [ERROR] Errors in 'file:libgdx/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtClipboard.java'
      [ERROR] Line 37: missing ; before statement
>       return await $wnd.navigator.clipboard.readText();
> ----------------------^
   Tracing compile failure path for type 'com.badlogic.gdx.backends.gwt.GwtClipboard'
      [ERROR] Errors in 'file:libgdx/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtClipboard.java'
         [ERROR] Line 37: missing ; before statement
>       return await $wnd.navigator.clipboard.readText();
> ----------------------^
   [ERROR] Hint: Your source appears not to live underneath a subpackage called 'client'; no problem, but you'll need to use the <source> directive in your module to make it accessible
```

Maybe it's possible to use it if we upgrade gwt.

tommyettinger:
Updating to a more recent GWT is apparently a large task, see https://github.com/libgdx/libgdx/issues/5256 . I don't know which, if any, GWT versions support `await`, or which browser versions it works reliably on if GWT just sends it over to the browser. There's probably a way to do the same thing (probably writing more verbose code) that works in earlier ECMAScript versions.

MobiDevelop:
Updating GWT to 2.8.2 isn't a large task, it was done shortly after that comment was made. But it will 'break' the setup if we don't follow with a release after applying the change.

SimonIT:
@MobiDevelop I will try it with your gwt-2.8.2 branch if this PR is merged and I will make a seperate PR for getting the clipboard if it works

gdx/src/com/badlogic/gdx/utils/Clipboard.java
```diff
@@ -20,10 +20,13 @@
  * @author mzechner */
 public interface Clipboard {
 	/** gets the current content of the clipboard if it contains text
+	 * for WebGL app, setting the system clipboard is currently not supported. It works only inside the app
```
intrigus: This seems mixed up with the setContents Javadoc.

gdx/src/com/badlogic/gdx/utils/Clipboard.java
```diff
@@ -20,10 +20,13 @@
...
 	 * @return the clipboard content or null */
 	public String getContents ();
 
 	/** Sets the content of the system clipboard.
+	 * for WebGL app, clipboard content might not be set if user denied permission, setting clipboard is not synchronous
```
intrigus: Same but vice versa.

tests/gdx-tests/src/com/badlogic/gdx/tests/ClipboardTest.java
```diff
@@ -0,0 +1,61 @@
+package com.badlogic.gdx.tests;
```
intrigus: Missing license header.

intrigus:
Aren't some apps faking clipboard access by adding and removing a html text field dynamically when the user presses CTRL+C and CTRL+V i.e. they redirect the event to an ordinary text field and then copy the result?
This is will of course not allow to always set/get the clipboard anytime but it will allow correct copy and pasting which is currently broken if I remember correctly.

So Prev:
Copy/Paste works in libgdx app only
Now:
Copy/Paste works globally, but clipboard can't be changed without user pressing "Copy/Paste".

SimonIT:
@intrigus done
That's also a good idea for getting the clipboard content. I will implement it if this PR is merged. I don't wanna put more effort in it if even this PR doesn't get merged

PR #5497, opened 2019-01-02
Added support for file properties in Tiled maps

tylerhasman:
I need to have file properties in my tiled map to link map objects to script files.
I'm sure that this isn't the best way to do it but I am open for feedback.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+[1.12.1]
+
 [1.12.0]
 - [BREAKING CHANGE] Added #touchCancelled to InputProcessor interface, see #6871.
 - [BREAKING CHANGE] Android: Immersive mode is now true by default. Set `useImmersiveMode` config to `false` to disable it.
--- a/gdx/src/com/badlogic/gdx/Version.java
+++ b/gdx/src/com/badlogic/gdx/Version.java
@@ -23,7 +23,7 @@ import com.badlogic.gdx.utils.GdxRuntimeException;
  * @author mzechner */
 public class Version {
 	/** The current version of libGDX as a String in the major.minor.revision format **/
-	public static final String VERSION = "1.12.0";
+	public static final String VERSION = "1.12.1";
 
 	/** The current major version of libGDX **/
 	public static final int MAJOR;
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ org.gradle.configureondemand=false
 android.useAndroidX=true
 
 group=com.badlogicgames.gdx
-version=1.12.0
+version=1.12.1
 POM_NAME=libGDX
 POM_DESCRIPTION=Android/Desktop/iOS/HTML5 game development framework
 POM_URL=https://github.com/libgdx/libgdx


EtK2000:
I'm not really sure the value of this, but if it's already implemented you might want to override toString() to return the file path.

tylerhasman:
It's useful because my game scripts are in Javascript and the tiled editor has no built-in formatter. The scripts get quite messy because of this so I have chosen to migrate the longer ones to files.

tylerhasman:
> I'm not really sure the value of this, but if it's already implemented you might want to override toString() to return the file path.

To add one to my above reply:
When you try to load a tiled map with a file attribute, it throws an exception so this is a necessary change.

EtK2000:
I don't really use the classes that you changed so I don't understand the addition and if/why it's necessary.
I'll step away from this and let someone more knowledgeable look at it.

tylerhasman:
The new Tiled update supports file properties so it only makes sense for LibGDX to handle them.

MobiDevelop:
I'm sure the maps stuff is so far behind the Tiled releases that there's plenty more that needs to be done besides handling file properties.

CHANGES
```diff
@@ -1,3 +1,5 @@
+[1.9.10]
+- Added support for 'file' properties when loading Tiled maps.
```
MobiDevelop: Maybe add a space to separate from the 1.9.9 list.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -351,6 +354,8 @@ private Object castProperty (String name, String value, String type) {
 			String opaqueColor = value.substring(3);
 			String alpha = value.substring(1, 3);
 			return Color.valueOf(opaqueColor + alpha);
+		}else if(type.equals("file")){
```
MobiDevelop: Spacing is off on this.

gdx/src/com/badlogic/gdx/maps/tiled/TmxMapLoader.java
```diff
@@ -77,6 +77,7 @@ public TiledMap load (String fileName, TmxMapLoader.Parameters parameters) {
 			this.convertObjectToTileSpace = parameters.convertObjectToTileSpace;
 			this.flipY = parameters.flipY;
 			FileHandle tmxFile = resolve(fileName);
+			this.tmxFile = tmxFile;
```
MobiDevelop: Should probably just set the field in the previous line.

gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java
```diff
@@ -133,6 +133,7 @@ public TiledMap load (String fileName, AtlasTiledMapLoaderParameters parameter)
 			}
 
 			FileHandle tmxFile = resolve(fileName);
+			this.tmxFile = tmxFile;
```
MobiDevelop: Should probably just set the field in the previous line.

MobiDevelop:
After refactor, this is no longer able to be merged due to conflicts.

MrStahlfelge:
@tylerhasman are you interested in updating the PR? otherwise we should close it due to the mentioned problem.

tylerhasman:
@MrStahlfelge i'll see if I have time to work on it

PR #5501, opened 2019-01-06
[pixmap packer io] added index based on file name, so can be used for…

wheelergames:
… animations

I lifted the same matcher patern from ImageProcessor.java so it should be working in the same way.

I also added useIndexes to SaveParameters so this can be turned off if required

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPackerIO.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPackerIO.java
@@ -2,6 +2,8 @@ package com.badlogic.gdx.graphics.g2d;
 
 import java.io.IOException;
 import java.io.Writer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


wheelergames:
Oh, I don't know if a change should be added to the CHANGES file, but from what I could tell, there is no 1.9.10 row in there, and I didn't want to be the first to start it! This is only my 2nd contribution to libgdx so I'm not senior enough to start that section!

wheelergames:
Yeah I thought about that, but I think (in my personal opinion) that useIndexes SHOULD be the default behaviour. I'm basing that on it being default behaviour in ImageProcessor, and in the docs when discussing how Animation class works.

Also, should there be a section in CHANGES file for things in 1.9.10? Is this enough of a change to qualify, or is it more of a bug fix?

wheelergames:
Oh, thanks for merging btw!!

wheelergames:
The docs for animations mention TexturePacker

> TexturePacker will automatically use these numbers as frame numbers (so long as the packing parameter useIndexes is left true).

If *left* true...maybe true should be the default?

PR #5506, opened 2019-01-11
fix TMX map image layer offset parsing

mgsx-dev:
layer offsets are floats, they are parsed as floats in loadBasicLayerInfo method but parsed as integer in loadImageLayer method which crash on loading for maps with non integer offsets.

I checked tiledmap editor, it allows float value.

I updated test as well to validate the right rendering.


MobiDevelop:
Good find!

PR #5508, opened 2019-01-12
optimize 3D animation key frame search

mgsx-dev:
this is a replacement for #4377

It's mostly inspired by the original PR with overall improvements and fixes to not break previous behavior.

It performs a kind of binary search with an heuristic based on time range.

Junit tests included. I had to change method visibility in order to test it.

Don't hesitate to tell me if you think it should be more documentated or if you think i missed some test cases.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
@@ -166,7 +166,7 @@ public class BaseAnimationController {
 		final int lastIndex = arr.size - 1;
 
 		// edges cases : time out of range always return first index
-		if (lastIndex < 0 || time < arr.get(0).keytime || time > arr.get(lastIndex).keytime) {
+		if (lastIndex <= 0 || time < arr.get(0).keytime || time > arr.get(lastIndex).keytime) {
 			return 0;
 		}
 
@@ -174,32 +174,17 @@ public class BaseAnimationController {
 		int minIndex = 0;
 		int maxIndex = lastIndex;
 
-		for (;;) {
-			final float minTime = arr.get(minIndex).keytime;
-			final float maxTime = arr.get(maxIndex).keytime;
-
-			if (time < minTime) {
-				return Math.max(0, minIndex - 1);
-			}
-			if (time > maxTime) {
-				return Math.min(lastIndex - 1, maxIndex);
-			}
-			if (minTime == maxTime) {
-				return minIndex;
-			}
-
-			// best guess index based on time range
-			float t = (time - minTime) / (maxTime - minTime);
-			int index = (int)(minIndex + t * (maxIndex - minIndex));
-
-			if (time < arr.get(index).keytime) {
-				maxIndex = index - 1;
-			} else if (time > arr.get(index).keytime) {
-				minIndex = index + 1;
+		while (minIndex < maxIndex) {
+			int i = (minIndex + maxIndex) / 2;
+			if (time > arr.get(i + 1).keytime) {
+				minIndex = i + 1;
+			} else if (time < arr.get(i).keytime) {
+				maxIndex = i - 1;
 			} else {
-				return Math.min(lastIndex - 1, index);
+				return i;
 			}
 		}
+		return minIndex;
 	}
 
 	private final static Vector3 getTranslationAtTime (final NodeAnimation nodeAnim, final float time, final Vector3 out) {
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
@@ -158,14 +158,48 @@ public class BaseAnimationController {
 
 	private final static Transform tmpT = new Transform();
 
-	private final static <T> int getFirstKeyframeIndexAtTime (final Array<NodeKeyframe<T>> arr, final float time) {
-		final int n = arr.size - 1;
-		for (int i = 0; i < n; i++) {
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
```diff
@@ -158,14 +158,48 @@ else if (applying)
...
+		// edges cases : time out of range always return first index
+		if (lastIndex < 0 || time < arr.get(0).keytime || time > arr.get(lastIndex).keytime) {
+			return 0;
+		}
+
+		// binary search
+		int minIndex = 0;
+		int maxIndex = lastIndex;
+
+		for (;;) {
```
SimonIT: Why don't you use `while(true)`?

Darkyenus:
Looks nice. Out of curiosity: have you tried to measure how it compares to the old linear version? (or how much does the midpoint heuristic help? )

mgsx-dev:
@Darkyenus to be honest i didn't benchmark anything. Out of curiosity i made some benchmarks and it seams that heuristic is not usefull, worse it is slower! nice catch!

Here is some tests i made : the first one is with minimum entropy (one key frame every seconds), the second one is with more entropy (one key frame every randomly 1s to 100s). Results are is milliseconds.

| frames | samples | With Heuristic | Without heuristic | Linear |
|--------|---------|----------------|-------------------|--------|
| 4 | 60000000 | 871.0 | 262.0 | 321.0 |
| 8 | 30000000 | 631.0 | 220.0 | 256.0 |
| 16 | 15000000 | 327.0 | 151.0 | 299.0 |
| 32 | 7500000 | 180.0 | 95.0 | 316.0 |
| 64 | 3750000 | 96.0 | 59.0 | 342.0 |
| 128 | 1875000 | 50.0 | 34.0 | 341.0 |
| 256 | 937500 | 25.0 | 20.0 | 339.0 |
| 512 | 468750 | 14.0 | 12.0 | 343.0 |
| 1024 | 234375 | 6.0 | 6.0 | 338.0 |
| 2048 | 117187 | 3.0 | 4.0 | 340.0 |
| 4096 | 58593 | 2.0 | 2.0 | 339.0 |
| 8192 | 29296 | 1.0 | 2.0 | 339.0 |
| 16384 | 14648 | 1.0 | 1.0 | 344.0 |
| 32768 | 7324 | 0.0 | 0.0 | 344.0 |
| 65536 | 3662 | 1.0 | 0.0 | 344.0 |

| frames | samples | With Heuristic | Without heuristic | Linear |
|--------|---------|----------------|-------------------|--------|
| 4 | 60000000 | 1080.0 | 304.0 | 345.0 |
| 8 | 30000000 | 695.0 | 242.0 | 146.0 |
| 16 | 15000000 | 421.0 | 162.0 | 338.0 |
| 32 | 7500000 | 190.0 | 98.0 | 327.0 |
| 64 | 3750000 | 106.0 | 59.0 | 350.0 |
| 128 | 1875000 | 62.0 | 37.0 | 332.0 |
| 256 | 937500 | 31.0 | 20.0 | 342.0 |
| 512 | 468750 | 17.0 | 11.0 | 330.0 |
| 1024 | 234375 | 9.0 | 6.0 | 337.0 |
| 2048 | 117187 | 5.0 | 4.0 | 336.0 |
| 4096 | 58593 | 3.0 | 2.0 | 352.0 |
| 8192 | 29296 | 2.0 | 2.0 | 341.0 |
| 16384 | 14648 | 1.0 | 1.0 | 345.0 |
| 32768 | 7324 | 0.0 | 1.0 | 359.0 |
| 65536 | 3662 | 0.0 | 1.0 | 428.0 |

Benchmark code available here : https://gist.github.com/mgsx-dev/290d06ef3d5e966725a6600000f5411a

I'll update my PR since it's faster and less code.

mgsx-dev:
Side note : Due to binSearch algorithm, it changes previous behavior in case of duplicates by not returning the first one. However, duplicates would raise error in interpolation code (zero division) anyway, so i decided to just remove this kind of tests.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/BaseAnimationController.java
mgsx-dev: tbh, i don't know but this code is obsolete now due to my PR update.

Darkyenus:
:+1: for benchmarking it. I think that there are some minor problems with that benchmark code, like keyframes.size actually being around 2x the expected size because it is not cleared, but nevertheless the result seems valid and pretty interesting.

I didn't know what the result would be, but it makes sense in retrospect. When it misses, the search space is not cut in half, but, in the worst case, made only very slightly smaller. Though it is still surprising that it would flop so badly, especially in the low randomness case.

By the way, I have been searching around and there is an search algorithm which does what the heuristic attempted to do. It is called [interpolation search](https://en.wikipedia.org/wiki/Interpolation_search) and has O(log(log(n))) best case and O(n) worst case, depending on the uniformity of the array.

mgsx-dev:
@Darkyenus oh, sorry about the benchmark code, i just read it again and you're right, i forget to clear keyFrame list :grimacing: but yeah, it doesn't change the conclusion.

About interpolation search, thanks for the link, i didn't know it and i think my implementation wasn't right at all. Anyway, since we can't predicte time distribution of the keyframes, classical binary search is the best choice IMO.

Thanks again to put me on the right track!

intrigus:
I can't reproduce your findings with my benchmark(https://github.com/intrigus/gdxbenchtest/tree/master).
Please try it.
Note that getXXXLinear may be extremely slow.

mgsx-dev:
I fixed benchmark code by clearing keyframes before each run. 
I updated result in my previous post as well.

PokeMMO:
Initial testing looks good here, long animations now run smoothly.

PR #5510, opened 2019-01-13
added index usage in PixmapPacker for updateTextureAtlas (currently missing so can't create animations)

wheelergames:
In the other classes (PixmapPackerIO) there is a useIndexes flag, but I can't see anything that this could be added to. In PixmapPackerIO there is parameters that it can be added to, but I'm not seeing an obvious comparison in PixmapPacker.

Alternative idea would be to overload the method, so that the original signature is still available to people, but now it could include the option to pass through useIndexes, rather than having it set to true as default.

(Personally I think it should be true as default, as this doesn't break anything, unless you intentionally call your image files _01.png, _02.png on purpose and don't do that for animations.


tommyettinger:
Make sure to test this on GWT, since it uses the `java.util.regex` package, which is partly emulated by libGDX but GWT itself does not implement. I can't tell by looking at it if it will have issues on GWT. PixmapPackerIO isn't really a concern, but PixmapPacker can be used on GWT.

tommyettinger:
@EtK2000 I don't think `_01.png` could be considered an indexed file, since the filename section would be blank, and I don't know if that can be put into an atlas. It certainly seems like it would be confusing if someone had a file named `_01.png` and expected it to work as a file with an index.

I do think that getting the indexed files working would be really good, so I'm hoping this can get tested soon on GWT and other platforms. I could maybe help, but I don't have any current code using PixmapPacker to try this on.

wheelergames:
@EtK2000 - I'm not disagreeing that the regex maybe not be the best, the only thing I'll say is it's the same regex used throughout the codebase to do indexing, so at least it's consistently broken ;)

I'll try to test on GWT, though I don't really know anything about it!

wheelergames:
I've tried testing on GWT but I literally have no idea how or what I'm doing!
My only comments to add in is that both BitmapFont.java and GLVersion.java use Matcher and Pattern. If they are already covered in the tests, or are known to work on GWT then maybe this isn't an issue?

I really wish I could help to test, but I've never done a thing on GWT, and after running ./gradlew tests:gdx-tests-gwt:superDev I was faced with a web page which said I hadn't compiled a module (do what now?) and no links that looked like they would be tests, especially not PixmapPacker tests

Sorry

tommyettinger:
@wheelergames PixmapPackerIO definitely does not work on GWT; I was pretty sure that @BenMcLean found PixmapPacker was the only way he could pack atlases at runtime (in-memory) on GWT, but that was a while ago. I'll try to get your PR building if my machine lets me.

wheelergames:
so I've removed PixmapPackerIO from the PixmapPacker test (moved it into a duplicate test called PixmapPackerIOTest).

I think the issue now, with a lot of help from Mobidevelop, is that AtlasRegion is not in the reflection cache. It is used in findRegions() in a new part of PixmapPackerTest, that proves indexes work, by making an animation

wheelergames:
Additional commit, now providing overloaded method to give option of turning off new useIndexes code.

Old method defaults to useIndexes true

BenMcLean:
I'm on my phone at the moment so I can't dive into the code, but I thought I'd mention that another problem related to repacking atlases at runtime was that (when I was working on the problem) patch9 information was not preserved. That's definitely an important feature. Please don't forget it. Thanks. :)

BenMcLean:
Oh, that's probably already been fixed then. Never mind! :)

wheelergames:
@BenMcLean can confirm in the method I updated (updateTextureAtlas) it already had the splits and pads information. It was added in aug and sept '18. So would've gone out in 1.9.9

gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPacker.java
```diff
@@ -342,9 +346,18 @@ public synchronized TextureAtlas generateTextureAtlas (TextureFilter minFilter,
 	/** Updates the {@link TextureAtlas}, adding any new {@link Pixmap} instances packed since the last call to this method. This
 	 * can be used to insert Pixmap instances on a separate thread via {@link #pack(String, Pixmap)} and update the TextureAtlas on
 	 * the rendering thread. This method must be called on the rendering thread. After calling this method, disposing the packer
-	 * will no longer dispose the page pixmaps. */
+	 * will no longer dispose the page pixmaps. Has useIndexes on by default so as to keep backwards compatibility*/
```
intrigus: Why is the default `true`?
How does this preserve backwards compatibility? 
The code that uses useIndex has been newly added.
So shouldn't it be off?

wheelergames: Because it's imitating the TexturePacker tool, and the tool has it on by default.
Also, the only reason it would break would be if you have been naming your assets with _01.png, _02.png, which could be argued to be a mistake in the first place, because this is the convention used by TexturePacker and PixmapPackerIO in order to create animations.
Ultimately, although it IS a new way of the class working, it's more bringing it inline with how all other texture packing in the libgdx world works, so I see it more as a bug fix, than a new feature.
But I can understand your view. If others agree, it can be swapped out to being false.
My thought process would be, it's better to fix the bug, than to keep it in to preserve backwards compatibility. Though that's only my own personal thoughts.

PR #5514, opened 2019-01-14
Array returned from AndroidZipFileHandle.list(...) included directory itself.

dlon:
For instance, `new AndroidZipFileHandle("music").list();` would return `[music/x0.mp3, music, music/x1.mp3]` instead of `[music/x0.mp3, music/x1.mp3]`.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
@@ -108,25 +108,31 @@ public class AndroidZipFileHandle extends AndroidFileHandle {
 	@Override
 	public FileHandle[] list() {
 		ZipEntryRO[] zipEntries = expansionFile.getEntriesAt(getPath());
-		FileHandle[] handles = new FileHandle[zipEntries.length];
-		for (int i = 0, n = handles.length; i < n; i++)
-			handles[i] = new AndroidZipFileHandle(zipEntries[i].mFileName);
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidZipFileHandle.java
```diff
@@ -108,25 +108,31 @@ public FileHandle parent() {
 	@Override
 	public FileHandle[] list() {
 		ZipEntryRO[] zipEntries = expansionFile.getEntriesAt(getPath());
-		FileHandle[] handles = new FileHandle[zipEntries.length];
-		for (int i = 0, n = handles.length; i < n; i++)
-			handles[i] = new AndroidZipFileHandle(zipEntries[i].mFileName);
+		FileHandle[] handles = new FileHandle[zipEntries.length - 1];
+		int count = 0;
+		for (int i = 0, n = zipEntries.length; i < n; i++) {
+			if (zipEntries[i].mFileName.length() == getPath().length())
```
Tom-Ski: Why comparing the lengths of the file names?

dlon: I believe it's a comparison with the directory path. It should always be shorter than the path of any file inside the directory, and I figured comparing the lengths was faster than a string or hash comparison.

Tom-Ski: Ah, missed that it was mFileName is including the full path also. In that case should be good.

PR #5517, opened 2019-01-16
Implement double to-and-from long on GWT

tommyettinger:
This has been marked NYI and FIXME for a long, long time, and it's an easy fix, following the behavior of the float and int conversion. This also fixes a compatibility bug introduced in 1.9.9; `NumberUtils.floatToIntColor()` was changed in the main project but not in GWT. There was also an unused import of `Color` in `NumberUtils`. I believe at least one test on GWT fails without this PR, calling `Numbers.doubleToLongBits()`, `Numbers.doubleToRawLongBits()` or `Numbers.longBitsToDouble()` and getting a NYI exception, but I don't know which one it was.


tommyettinger:
Hooray, thanks @MobiDevelop !

Issue #4561, opened 2017-01-24
"Auto Scrolling" TextArea removes first line instead of just growing


LennartH:
#### Issue details
I tried to build an auto scrolling TextArea (for an in-game log) based on [this snippet](https://github.com/libgdx/libgdx/issues/3796#issuecomment-202729107). But when text is added consecutively, the TextArea removes the first line instead of just growing, so that the outer ScrollPane can do its magic.

A simple workaround, was to just add 1 to the value returned by `getLines()`, but I don't like magic numbers and I think there's something wrong in the way the backing `lineBreaks` array is maintained.
I tried calling `calculateOffsets()` before the call of `getLines()`, but that didn't help.

#### SSCCE
```java
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
import com.badlogic.gdx.backends.lwjgl.LwjglApplicationConfiguration;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.scenes.scene2d.Actor;
import com.badlogic.gdx.scenes.scene2d.Stage;
import com.badlogic.gdx.scenes.scene2d.ui.Button;
import com.badlogic.gdx.scenes.scene2d.ui.ScrollPane;
import com.badlogic.gdx.scenes.scene2d.ui.Skin;
import com.badlogic.gdx.scenes.scene2d.ui.Table;
import com.badlogic.gdx.scenes.scene2d.ui.TextArea;
import com.badlogic.gdx.scenes.scene2d.ui.TextButton;
import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener;
import com.badlogic.gdx.utils.viewport.ScreenViewport;

public class AutoScrollingTextAreaPrototype extends ApplicationAdapter {
    Stage stage;
    int lineCounter;

    public void create () {
        stage = new Stage(new ScreenViewport());
        Skin skin = new Skin(Gdx.files.internal("skin/default/uiskin.json"));
        Gdx.input.setInputProcessor(stage);

        Table container = new Table();
        stage.addActor(container);
        container.setFillParent(true);
        container.pad(10).defaults().expandX().fillX().space(4);

        final OpenScrollPane scrollPane = new OpenScrollPane(null, skin);
        scrollPane.setFadeScrollBars(false);
        scrollPane.setFlickScroll(false);
        scrollPane.setScrollingDisabled(true, false);

[...]

PR #5520, opened 2019-01-18
fix libgdx/libgdx#5488 ScrollPane scrollTo

### LINKED ISSUE #5488: Issue with ScrollPane.scrollTo
#### Issue details
If a large amount of content is loaded inside a ScrollPane, and you wish to scroll to the end of the content, multiple calls to ScrollPane.layout() are required. (You can find other examples of this issue on the web.)

#### Reproduction steps/code
ScrollPane created with the following code:

Table tblOuter = new Table();
Table tblInner = new Table();
lblChat = new Label("", skin, "small");
lblChat.setWrap(true);

tblInner.add(lblChat).width(chatWidth - 40f).padLeft(5f).top();
tblInner.add(new Label("", skin, "small")).fillX().right();
tblInner.row();

pneChat = new ScrollBarPane(tblInner, skin, "default");
tblOuter.add(pneChat).width(chatWidth - 20f).height(controlHeight - 30f).expand().row();

Then, if after a large amount of text is loaded in lblChat, layout() has to be called twice prior to scrollTo working:

pneChat.layout();
pneChat.layout();
pneChat.scrollTo(0, 0, 0, 0);

1.9.9

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [X ] Windows
- [X ] Linux
- [ ] MacOS

mgsx-dev:
Calling ScrollPane scrollTo just after changing children size didn't work.

A another way (workaround) was to manually validate scrollPane before calling scrollTo but i think it's better to do it automatically in ScrollPane class. @NathanSweet What do you think?

It fixes #5488 and in some way #4561 as well : I'll comment on the latest issue.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneWithDynamicScrolling.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneWithDynamicScrolling.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2011 See AUTHORS file.
+ * Copyright 2019 See AUTHORS file.
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneWithDynamicScrolling.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneWithDynamicScrolling.java
@@ -1,0 +1,97 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ * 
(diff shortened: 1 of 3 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneWithDynamicScrolling.java
```diff
@@ -0,0 +1,97 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
```
Lonami: Are you supposed to still use 2011 in new copyright notices?

NathanSweet:
I think it makes sense to validate inside scrollTo. Thanks!

PR #5161, opened 2018-04-05
PooledLinkedList: Support some common doubly linked list operations

WickedShell:
PooledLinkedLists look quite useful for me, but it was missing a couple of common doubly linked list operations. (Mainly the ability to add to the head of the list, and to be able to remove from the head/tail arbitrarily (without having to shift the iter to it)). I haven't added the head removal yet because that one isn't hurting me nearly as much (and is achievable via an ugly series of calls to `iter()`, `next()`, `remove()` ).

I'm happy to add that if people are interested in it. (I'll come back tomorrow and fix the tab/spaces issues I just noticed here in the review).

PR #5521, opened 2019-01-18
Don't refer to freed element. Fix by order change.

intrigus:
tail.next is still referring to the old (now freed) tail.
So first assign the new value to tail, then change the tail.next value.


intrigus:
CC: #5161

MrStahlfelge:
It is a good change because it fixes a bug. Here you have my review (which changes nothing in effect).

MobiDevelop:
Good find, thanks for the fix.

PR #5532, opened 2019-01-23
Lwjgl.Graphics getDeltaTime() to actually return smoothed out delta

minjaesong:
Couldn't test it for LWJGL3 but it'll work on there too. This PR only concerns LWJGL2.

It takes raw delta as an "sensor read" and applies Kalman filter to get smoothed out delta. This kalman filter implementation is by no means optimal, but with some dirty heuristics it works well.


NathanSweet:
IIRC, lwjgl2 already does some delta smoothing. I don't think the backend should do smoothing at all, but especially not by default with no way to disable it. You can easily do this in application code.

PR #5541, opened 2019-01-26
Fixes #5540, improved Align.toString

### LINKED ISSUE #5540: Align.isCenterVertical and Horizontal return wrong results
[`Align`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/Align.java)'s `isCenterVertical` and `isCenterHorizontal` return wrong results.

```java
int align = Align.top | Align.bottom | Align.left
System.out.println(Align.isCenterVertical(align))
>> false
```
```java
int align = Align.center
System.out.println(Align.isCenterVertical(align))
System.out.println(Align.isCenterHorizontal(align))
>> false
>> false
```
I think all of these should return true.

maltaisn:
- Improved the toString method to give meaningful results for alignments centered only on one axis, e.g.: `top center` and `center right`.
- Added a few more constants for alignements centered only on one axis.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Align.java
+++ b/gdx/src/com/badlogic/gdx/utils/Align.java
@@ -19,55 +19,68 @@ package com.badlogic.gdx.utils;
 /** Provides bit flag constants for alignment.
  * @author Nathan Sweet */
 public class Align {
-	static public final int center = 1 << 0;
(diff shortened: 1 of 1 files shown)


intrigus:
Please only use the libgdx formatter for parts that you changed.

Issue #5491, opened 2018-12-23
Implementation tags in build.gradle file causes issues in desktop:dist


HeinousGames:
#### Issue details
desktop:dist exports a jar that has no class defined for ApplicationListener when not using "compile" in the build.gradle file.

#### Reproduction steps/code
- Create a new project from the gdx-setup.jar tool.
- Change "compile" tags to "implementation" tags inside the build.gradle file.
- Attempt desktop:dist from command line.

A jar will be exported but will not run. This happens on a project that is only made with the desktop module. If attempting to make a project with the Android module, changing the tags from implementation to compile is not enough, you must also ensure the android:build:gradle plugin is 3.1.3. **I am alarmed by this as Android Studio says "compile" will be obsolete at the end of 2018.**

#### Version of LibGDX and/or relevant dependencies
LibGDX v1.9.8 & 1.9.9
Android Studio v3.1.4
Mac OS Mojave v10.14.2

#### Stacktrace when attempting to run the jar via command line
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.NoClassDefFoundError: com/badlogic/gdx/ApplicationListener
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.privateGetMethodRecursive(Class.java:3048)
	at java.lang.Class.getMethod0(Class.java:3018)
	at java.lang.Class.getMethod(Class.java:1784)
	at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:544)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:526)
Caused by: java.lang.ClassNotFoundException: com.badlogic.gdx.ApplicationListener
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 7 more

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [Not sure but assuming] Windows
- [Not sure but assuming] Linux
- [X] MacOS

intrigus:
Please provide a `diff` of the changes you did.

[...]

PR #5544, opened 2019-01-28
change 'compile' to 'implementation' in gdx-setup.jar

ryanastout:
This fixes the "WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'. " warning from Gradle. 
Rather than using "compile" for dependencies, uses "implementation" instead.
See 
https://developer.android.com/studio/build/dependencies?utm_source=android-studio#dependency_configurations

To test this, I built the gdx-setup.jar by running "ant" in the libgdx/extensions/gdx-setup directory, then I created a new gdx project with the gdx-setup.jar file it created. I also verified that occurrences of "compile" in the generated project's build.gradle file had all be replaced with "implementation"


mgsx-dev:
There are some side effects, please read #5491

Why replace by implementation instead of api scope?

ryanastout:
Ah, I hadn't tested with desktop:dist; I'll close this pull request for now.

PR #5551, opened 2019-02-03
compareTo realization for Vector2 and Vector3

nikitamiroshnichenko:
Comparable java interface implementation for Vector2 and Vector3 classes. May be useful when you need to sort array of vectors.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -23,7 +23,7 @@ import com.badlogic.gdx.utils.NumberUtils;
 
 /** Encapsulates a 2D vector. Allows chaining methods by returning a reference to itself
  * @author badlogicgames@gmail.com */
-public class Vector2 implements Serializable, Vector<Vector2> {
+public class Vector2 implements Serializable, Comparable, Vector<Vector2> {
 	private static final long serialVersionUID = 913902788239530931L;
 
 	public final static Vector2 X = new Vector2(1, 0);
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/math/Vector2.java
```diff
@@ -551,4 +551,24 @@ public Vector2 setZero () {
 		this.y = 0;
 		return this;
 	}
+
+	/** Compares this vector with the other {@link Vector2} based on x and y variables.
+     * Useful for sorting.
```
SimonIT: Please fix the indentation

nikitamiroshnichenko: done

mgsx-dev:
Vector2 and Vector3 doesn't have natural ordering. It doesn't make sense to implement this. See https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html contract.

You can always use comparators based on specific ordering logic in your own code.

Please reconsider these changes.

Darkyenus:
I agree with @mgsx-dev. While it sometimes might be useful to sort vectors, the required ordering is usually wildly different from case to case. Implementing `Comparable` would probably just be confusing and bug prone.

Additionally, if the goal of this PR is to provide *any* total ordering (which is probably the most unbiased/useful way to go about this, and it is what I would from javadoc expect), this PR doesn't provide that (logic is more complicated than it needs to be and `Vector3` doesn't consider all components).

Nevertheless, thanks for the PR, don't let this discourage you.

PR #5552, opened 2019-02-04
Create dynamically the ExtendedChars

guichazan:
The original code does not support creation of fonts like FontAwesome, that starts at F000. Also, it does not contemple all characters that a font may have. So i'm proposing a change that the Extended char set will use all available characters in the font from range 33 to 65535.


NathanSweet:
ExtendedChars is not meant to be all chars. I don't think we should change it. We could add a new setting for all chars instead.

Tom-Ski:
Agree. PR for additional setting for all chars would be accepted. Feel free to update/reopen if done.

PR #5575, opened 2019-02-22
make keyboardAvailable protected so subclasses of AndroidInput can as…

guichazan:
…sign to it.

See title


MrStahlfelge:
While I think it would be an advantage to get more things protected in overall libGDX (thinking of other backends, and Scene2d as well), this PR alone does not make sense IMO. There's no way to instantiate an own subclass of AndroidInput. It would be good to have a possibility to do so, but I doubt any PRs changing this won't get pulled.

NathanSweet:
MrStahlfelge seems to make sense.

Exposing internals comes up regularly. We are unlikely to expose everything, but may expose some things or make them configurable where it makes sense.

PR #5577, opened 2019-03-01
Remove support for OSX 32-bit builds.

PokeMMO:
All OSX releases after [Snow Leopard](https://en.wikipedia.org/wiki/Mac_OS_X_Snow_Leopard) (July 20, 2011) are 64-bit only. 32-bit OSX is deprecated.

Oracle and Openjdk (as far as I can see) do not provide any currently supported 32bit JRE builds.

It seems like including 32-bit osx builds is no longer necessary.


PokeMMO:
Neither [lwjgl2](http://central.maven.org/maven2/org/lwjgl/lwjgl/lwjgl-platform/2.9.3/lwjgl-platform-2.9.3-natives-osx.jar) nor [lwjgl3](http://central.maven.org/maven2/org/lwjgl/lwjgl-opengl/3.2.1/lwjgl-opengl-3.2.1-natives-macos.jar) include support for OSX 32-bit libraries.

NathanSweet:
Cheers!

PR #5579, opened 2019-03-03
Update MultipleRenderTargetTest.java

PriNova:
Fixed addDepthTextureAttachment internalFormat- and type-parameter to work on Android 6.0.1. Don't know if this fix is only for Android or influences other platforms too.


NathanSweet:
Can @PriNova or anyone else verify this is fine for non-Android platforms?

mgsx-dev:
I just tested with both Lwjgl and Lwjgl3 desktop backends. It seams to be OK.

LWJGL before
lwjgl-without

LWJGL after
lwjgl-with

LWJGL3 before
lwjgl3-without

LWJGL3 after
lwjgl3-with

NathanSweet:
Cheers!

PR #5583, opened 2019-03-05
Add SpriteCache.getCustomShader() and SpriteCache.isDrawing()

dlon:


EtK2000:
Might I suggest renaming to "getCustomShader" as it returns null if using the default.

dlon:
Or would it be better to return the default shader in that case, like SpriteBatch does? (I find this less useful.)

MobiDevelop:
I'd call it getCustomShader of it will only return the custom shader, otherwise I think it should return the default if there is no custom

RUSshy:
setShader() 
getShader()

Seems like it follow what was used for the setter, maybe should rename this one too to avoid confusion ?

PR #5611, opened 2019-04-16
Enhanced Animation.java.

Blezyn:
Spares the user from keeping track of the stateTime, which is also not safe. Instead, it keeps track of its own stateTime, just by calling its update(float deltaTime) method, for every frame, opposed to the previous approach, where the user had to keep track and update the stateTime, on their own. 
Ofcourse, if the user wants to keep track of one or multiple stateTime, they can, just by calling the previous methods.

A few other methods were added, to manipulate the stateTime.


mgsx-dev:
This change makes Animation stateful which totally change its design. Animations can be shared across multiple sprite instances and such which is why it's stateless. Please reconsider this change.

Tom-Ski:
I agree. If you would like this behavior it seems sensible enough to just create your own tiny wrapper.

Issue #5599, opened 2019-03-22
@NonNull and @Nullable annotations on parameters and return types


maltaisn:
I think adding `@NonNull` and `@Nullable` annotations to the source code would be very helpful documentation. It would also allow better Kotlin interoperability, since the Kotlin compiler can [recognize these annotations](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations). There are already a few places where the javadoc indicates that a parameter may be null but it's far from being widespread.

For example the Batch in draw method and the events in event listeners are never null. These are obvious, but there are many cases where all the usages of a method but be looked at to know whether the parameter type is nullable or not. Kotlin also marks the `Color` and `Interpolation` constants as potentially nullable.

I understand that adding these annotations to every public method would be a very big undertaking, but I think it would prove to be very useful.

Darkyenus:
I have recently added them to [Spigot API](https://hub.spigotmc.org/jira/browse/SPIGOT-2540) (for Minecaft server mods). It took about 30 hours, but I did found some bugs and I think that their API is now better for it. I think that this would be a *good-first-issue* (especially if it is split up into multiple smaller PRs), but someone (=core contributor) must first decide, which annotations should be used, as there is a lot of options, and all have their pros and cons. I outlined some of the options in the issue linked above, and ended up using the Jetbrains flavored ones, but libGDX may have a different needs and decide differently.

NathanSweet:
The way it _should_ be currently is all parameters and return values that can be null should have javadoc stating that. If the javadoc doesn't say it can be null, you can assume it can never be null. Of course no checking is enforced.

I've not personally added annotations for null to a large project. I'd want to try that on a smaller, personal project to get an idea of what I like and don't like about it before tackling libgdx.

maltaisn:
Already having javadoc for all nullable parameters will make it easier to add the nullability annotations.

Another advantage of using the annotations is that an IDE like IntelliJ will issue warnings when dereferencing a nullable object or when making null checks on a non-null object for example.

If you want to try on a smaller project, maybe start by adding them to a single part of libgdx, like only scene2d? I think it would also be the part that benefits the most. I can make a pull request, I just need to know what annotations you want to use. I would recommend using JetBrains' like @Darkyenus said.

maltaisn:
I added `@NotNull` and `@Nullable` annotations to base scene2d classes. I chose `org.jetbrains:annotations` but it can easily be changed.

[...]

PR #5615, opened 2019-04-23
Nullability annotations for base scene2d classes

maltaisn:
I added `@NotNull` and `@Nullable` annotations to base scene2d classes as discussed in #5599. I chose `org.jetbrains:annotations` but it can easily be changed. 

Parameters that I wasn't sure about have a `/* FIXME not sure */` comment next to them.

This pull request is for review, don't merge it.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
@@ -588,8 +588,9 @@ public class GwtInput implements Input {
 			hasFocus = true;
 			this.justTouched = true;
 			this.touched[0] = true;
-			this.pressedButtons.add(getButton(e.getButton()));
-			justPressedButtons[e.getButton()] = true;
+			final int button = getButton(e.getButton());
+			this.pressedButtons.add(button);
+			justPressedButtons[button] = true;
 			this.deltaX[0] = 0;
 			this.deltaY[0] = 0;
 			if (isCursorCatched()) {
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -653,7 +653,7 @@ public interface Input {
 	public boolean isButtonPressed (int button);
 
 	/** Returns whether a given button has just been pressed. Button constants can be found in {@link Buttons}. On Android only the Buttons#LEFT
-	 * constant is meaningful before version 4.0.
+	 * constant is meaningful before version 4.0. On WebGL (GWT), only LEFT, RIGHT and MIDDLE buttons are supported.
 	 *
 	 * @param button the button to check.
 	 * @return true or false. */
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtInputTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtInputTest.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.tests.gwt;
 import com.badlogic.gdx.Application;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Graphics;
+import com.badlogic.gdx.Input;
 import com.badlogic.gdx.InputProcessor;
 import com.badlogic.gdx.Input.Keys;
 import com.badlogic.gdx.graphics.Color;
@@ -73,6 +74,15 @@ public class GwtInputTest extends GdxTest {
 
 		if (Gdx.input.isKeyPressed(Keys.DOWN)) {
 			y -= 1;
+		}
+		if (Gdx.input.isButtonJustPressed(Input.Buttons.LEFT)){
+			Gdx.app.log("GwtInputTest", "button pressed: LEFT");
+		}
+		if (Gdx.input.isButtonJustPressed(Input.Buttons.MIDDLE)){
+			Gdx.app.log("GwtInputTest", "button pressed: MIDDLE");
+		}
+		if (Gdx.input.isButtonJustPressed(Input.Buttons.RIGHT)){
+			Gdx.app.log("GwtInputTest", "button pressed: RIGHT");
 		}
 	}
 
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
@@ -359,7 +359,7 @@ public class GwtTestWrapper extends GdxTest {
 
 		@Override
 		public boolean isButtonJustPressed (int button) {
-			return false;
+			return input.isButtonJustPressed(button);
 		}
 
 		@Override


MrStahlfelge:
Did you check GWT compatiblity?

Darkyenus:
These annotations are source only, so unless GWT has a really bad parser, it should work.

MrStahlfelge:
The GWT parser needs to see the sources. I can't see that you added the sources, and I can't see that you added the GWT inherits flag. In my experiences, this is often needed for such changes and I had problems with GWT myself when I added Nullability annotations in my projects. 

So I am not sure if you checked the compatiblity but just guessed that there should be no problems.

maltaisn:
I just added `compile 'org.jetbrains:annotations:16.0.2'` to base gradle script (and forgot to commit it). Do I also need to add the annotation classes to [gdx.gwt.xml](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx.gwt.xml)? I didn't manage to run the gwt test suite on IntelliJ so I couldn't check.

SimonIT:
For my surprise run the tests without any additional addition in gwt

maltaisn:
Does that mean it works with GWT? Should I continue adding the annotations?

SimonIT:
Yes, it works (if I did nothing wrong)

maltaisn:
I finished adding nullability annotations for `gdx.scenes.scene2d` package. 
I think Scene2d was probably the hardest part to add annotations for, the rest should
be pretty straightforward. Some notes:

- Late initialized fields were not annotated as nullable. 
  
  For example, `Actor.setStage`, `Event.setTarget`, `Action.setTarget` are
  not annotated. In the same way, style classes fields set by reflection are not nullable
  if they are required: `LabelStyle.font` is not annotated because it's required,
  but `LabelStyle.fontColor` is nullable.
  
  These fields are null when uninitialized but most of the time they aren't,
  so being nullable would only be a nuisance.
  
- All non-final and non-null fields were annotated as not null. Final fields
  initialized outside the constructor were not annotated since their nullibility is obvious.
  
- Private, package private and public methods were all annotated the same.
  
- When possible, fields were annotated. For example:
  ```java
  @NotNull private Touchable touchable = Touchable.enabled;
  ```
  Since the field is initialized directly and not in constructor, it can
  be marked as not null without creating a warning "`@NotNull` fields must be initialized".
  However:
  ```java
  private LabelStyle style;
  ```
  Cannot be annotated without creating a warning, since it's initialized in constructor.
  Getters and setters are still annotated as usual since field is really not null.
   
- Methods returning `this` were never annotated as `@NotNull` since it's obvious.

- `assert variable != null` was used at some places to prevent warnings.

- Null checks were always kept, since `@NotNull` does no runtime checking.

- To prevent some warnings, methods like `Table.requireCell` 
  or `ArraySelection.requireSelected` could be added to throw an exception 
  if result is null. (which is never supposed to happen) This is how Android does it.

I found some code issues on my way:
- Some return values and method parameters had a "May be null" message
  in javadoc, although they weren't. For example primitive values or
  values for which a null value would immediately cause a NPE. (fixed)
- `ListStyle`, `TextTooltipStyle` and `TreeStyle` 
  copy constructors don't copy all fields. (not fixed)
- Unnecessary instanceof in `Slider.setStyle`, `Skin.newDrawable` and
  `TextTooltip.setStyle`. (not fixed)
- Very inconsistent use of `@Override`, not sure if this is intended or not. (not fixed)
- Some toString methods can return null, like `BaseDrawable.toString`. (not fixed)

maltaisn:
@NathanSweet Any update on this?

RUSshy:
On top of that, i think adding ``@Override`` attribute everywhere method is overwritten will help too, blind overwritting is bad, it is even more bad for public APIs

NathanSweet:
@maltaisn Sorry for not being able to look at your PR sooner.

I'm not excited about adding an new dependency. Is it possible to host our own @NotNull and still have IDE integration?

I would still like to apply the annotations to a personal project before deciding on scene2d. I haven't gotten to do that yet. From only looking at your PR, I don't hate `@Nullable`.

I think that fields that are assigned in constructors should have the annotation. If that causes a warning, just turn the warning off. The annotation is a tool for application code, we can assume library code works correctly.

What was `@NotNull` added? Everything should be considered as not being null unless it has `@Nullable`.

`Skin.newDrawable` and `BaseDrawable.toString` look OK. The other items have been fixed, thanks!

@RUSshy This has been discussed before, and to death. PRs to add `@Override` won't be merged. It makes sense for application code. It isn't required for library code, where it just adds noise and the benefits for refactoring are insignificant.

Darkyenus:
> What was @NotNull added? Everything should be considered as not being null unless it has @Nullable.

IIRC that can be done by annotating the whole class as `@NotNull`, which would be a nice compromise, since otherwise IntelliJ would consider those fields as "unknown nullability" and not provide any warnings on e.g. unnecessary null checks.

NathanSweet:
`@NotNull` seems nasty. Annotations make sense for something that is abnormal. I would not want extraneous annotations all over the place for the common case where things are not null.

maltaisn:
> I'm not excited about adding an new dependency. Is it possible to host our own @NotNull and still have IDE integration?

Yes this is possible with both Eclipse and IntelliJ, although I recommend sticking with the `org.jetbrains.annotations` dependency since they are set by default in IntelliJ which I believe is what most people use. It's a compile time only dependency so there shouldn't be any problem.

> I think that fields that are assigned in constructors should have the annotation. If that causes a warning, just turn the warning off. The annotation is a tool for application code, we can assume library code works correctly.

Sure that makes sense, I can make the changes.

> What was `@NotNull` added? Everything should be considered as not being null unless it has `@Nullable`.

> `@NotNull` seems nasty. Annotations make sense for something that is abnormal. I would not want extraneous annotations all over the place for the common case where things are not null.

Each annotation is as useful as the other and each provide their own warnings. Without the `@NotNull` annotation, you don't get any warning when passing a null or a nullable value to an argument that shouldn't be null.

Yes, it does make the code more verbose. But it's also more maintainable, eg: you won't have to look all over the place every time you wonder if a value is nullable or not. The same goes with `@Override`, but that's not the question here.

NathanSweet:
> this is possible with both Eclipse and IntelliJ, although I recommend sticking with the org.jetbrains.annotations dependency since they are set by default in IntelliJ which I believe is what most people use.

I could go either way. FWIW, I don't use IntelliJ.

> Each annotation is as useful as the other and each provide their own warnings. Without the @NotNull annotation, you don't get any warning when passing a null or a nullable value to an argument that shouldn't be null.

Is there no IDE setting to assume not null when there is no annotation? If not, why? I could see it causing a lot of warnings where @NotNull has not been fully embraced, but otherwise would be better.

> Yes, it does make the code more verbose. But it's also more maintainable, eg: you won't have to look all over the place every time you wonder if a value is nullable or not. The same goes with @Override, but that's not the question here.

There is no wondering: if something has `@Nullable` then it can be null, if it doesn't then it can't be null. The most common case is not null. Using an annotation for the most common case would add a lot of clutter.

maltaisn:
See #5812 for the updated pull-request.

PR #5619, opened 2019-04-27
Maximised display positioning

thechubbypanda:
The added code ensures that the window is maximised on the correct screen (the primary one). 

I code on my secondary monitor (portrait) and since my editor is on that monitor, the game opens on that monitor when set to maximised.

Note: I might not be accounting for some config settings.


Tom-Ski:
Maybe we should have an additional setting for a maximise target to be the primary monitor? Otherwise every new window we spawn, or "sub" window is always forced to open on the primary here.

thechubbypanda:
Good idea. In the display config, right?

Tom-Ski:
Lwjgl3WindowConfiguration, yeah. Please also keep the current behavior. (Opening on whatever monitor it thinks is good).

thechubbypanda:
Ok, that does the trick I think.

thechubbypanda:
I realised that was really bad so I changed it to conform with existing methods.

thechubbypanda:
Do the commits made by other people get counted as this merge???

MrStahlfelge:
It's because you opened this PR on your master branch and you push these commits on this branch. Always create a new branch for a PR and open the PR for this branch.

Tom-Ski:
Yeah, can't merge this.

thechubbypanda:
Ah, I'll make a new one. Sorry about that! Still kinda new to github!

PR #5623, opened 2019-04-28
Add resources folder path

slawtul:
During starting the DesktopLauncher without this fix I get: Exception in thread "LWJGL Application" com.badlogic.gdx.utils.GdxRuntimeException: Couldn't load file: badlogic.jpg

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/core/build.gradle
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/core/build.gradle
@@ -4,6 +4,7 @@ sourceCompatibility = 1.6
 [compileJava, compileTestJava]*.options*.encoding = 'UTF-8'
 
 sourceSets.main.java.srcDirs = [ "src/" ]
+sourceSets.main.resources.srcDirs = [ "assets/" ]
 
 
 eclipse.project {


mustii82:
This should fix the annyoying part of declaring the Assets Folder as Working Directory on Run Properties

SimonIT:
The problem is that if you create an android project, this will not work because the assets folder is inside the android project in this case.

slawtul:
Hi Simon,
thanks for explanation. I was not aware of it. I am not Android programmer.
Any idea how to fix it? It's not good for new libgdx programmer to get this error just after launching empty project. It's daunting.

SimonIT:
I totally agree with you. Maybe ``` sourceSets.main.resources.srcDirs = [ "../%ASSET_PATH%" ] ``` works. But you have to test it

slawtul:
Checked. Does not work.

piotr-j:
You just need to set the working directory to one with assets. You dont need to do that when you run with gradle.

Tom-Ski:
Yup, this is in the official documentation too. Eclipse shouldn't have this issue though, because we are able to link the asset folder up (unlike IDEA). Will take a look at a solution, as well as this proposed solution. I use this in some of my projects, but there are some caveats which aren't so nice.

payne911:
May I ask why it was chosen that Android's `assets` folder would be used as the default unless it isn't an Android project, which then means the `core` earns the privilege to hold the assets.

It seems counter-intuitive to not simply always have things in the `core` package.

Tom-Ski:
Its legacy from eclipse @payne911 .

MrStahlfelge:
@Tom-Ski you said you will take a look at this at a time, but can we close this PR to clean up the open PRs?

PR #5635, opened 2019-05-25
Experimental travis builds

PokeMMO:
Jenkins hasn't reliably run builds in a while, so using a proper CI might be more ideal. This will also allow more easy changes for things like adding arm compilers for future arm builds.

This will allow users to more easily test their own changes by using their own travis instances.

Doesn't actually deploy anything anywhere atm, would need a maintainer to configure maven deploy to actually upload the resulting binaries.

Also replaced https://search.maven.org/ with direct links, since https://search.maven.org/ was failing during testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 sudo: true
 git:
-  depth: 1
-branches:
-  only:
-    - master
+  depth: 3
+#branches:
+#  only:
+#    - master
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


Tom-Ski:
Our official build and deployments will continue to be handled by our jenkins instance.

I'm more than happy to continue using this for testing, and we can expand upon it for other users if they wish to run their own travis instances.

Ping when updated for merge conflicts.

PokeMMO:
I've only had a little more time to look into fixing this up and it seems something has changed on travis/ubuntu's end making the cross compiler installation fail.

I also looked into running mac first, and then handing off the mac natives to the linux build, but travis offers no built in support to do this. Which means each user would have to setup an s3 bucket or other script in order to transfer between stages. This makes this a lot less useful, but might still be worthwhile.

PokeMMO:
I haven't had time to look into this more, and given Github Actions is out. I'd probably try to use Github Actions for this instead.

PR #5640, opened 2019-05-29
Add missing copyAndroidNatives task in gdx-tests-android

PokeMMO:
Without this `./gradlew tests:gdx-tests-android:assembleDebug` on a clean project will not include natives in the generated apk.

Copied verbatim from https://github.com/libgdx/libgdx/blob/2a429a82463ed6dc2e24219f9ac67328789c8f43/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle#L35-L68

### DIFF
--- a/tests/gdx-tests-android/build.gradle
+++ b/tests/gdx-tests-android/build.gradle
@@ -23,6 +23,10 @@ buildscript {
 apply plugin: "com.android.application"
 apply from: "obb.gradle"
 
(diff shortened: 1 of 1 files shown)


Tom-Ski:
Natives are brought in by the fetchNatives task.

PokeMMO:
So they are, thanks for the quick reply.

Issue #5485, opened 2018-12-18
AndroidGL20 JNI Removal


Anuken:
While snooping around the libGDX Android backend's source, I noticed something strange.
[AndroidGL20](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL20.java) uses JNI to access openGL 2.0 methods, while [AndroidGL30](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL30.java) uses a standard Android library class to do so: [GLES30](https://developer.android.com/reference/android/opengl/GLES30). However, standard Android also has a [GLES20](https://developer.android.com/reference/android/opengl/GLES20) class.

Why does AndroidGL20 use JNI when it could be using standard Android libraries? I've re-implemented AndroidGL20 without any JNI and tested it in one of my games, and there doesn't seem to be any performance difference. After asking around in the Discord, someone suggested that the JNI-based GL20 implementation could have been made before the standard GLES20 implementation was released.

If that is the case, would it be a good idea for me to submit a pull request removing [these C++ files](https://github.com/libgdx/libgdx/tree/master/gdx/jni/android) and replacing the current AndroidGL20 implementation with [my own](https://gist.github.com/Anuken/38e3e6c0db299c7ff22d5cfd500ab070), to reduce the amount of unnecessary C++ code in the repository?

MrStahlfelge:
The problem is, most people don't like to bother with an own build to try things. I had no problem to run my debug and test builds on a new Android backend while keeping 1.9.8 for release builds, but I will not throw in the effort to check if there are changes, then pull, build and publish locally every time I build a new test version.  And I think I am not alone with this.

What libGDX really lacks the last time (the last 18 month), is continous releases. There's an easy schema to implement new, experimental features while constantly having a stable version: Running two versions at a time. The version numbers have three numerals, by intention. They are named major, minor, incremental. All fixes go into a new version where the first two numerals stay the same and just the third one, the incremental is incremented. Experimental features or code-breaking changes go into a version where at least the minor version number is incremented.

Two important points: It must be stated on the release page if a release is experimental, beta or stable. And: A new incremental release for all running versions should be done every time there are enough changes that should go out to the public. For the experimental versions, that means when it should be tested by all devs. For stable versions, it is when there are important fixes.

noblemaster:
[...]

PR #5641, opened 2019-05-29
AndroidGL20 JNI Removal

PokeMMO:
See https://github.com/libgdx/libgdx/issues/5485

Removes dependency on [greylisted](https://developer.android.com/distribute/best-practices/develop/restrictions-non-sdk-interfaces) [NIOAccess api's](https://android.googlesource.com/platform/frameworks/base/+/pie-release/config/hiddenapi-light-greylist.txt#8982)

https://developer.android.com/reference/android/opengl/GLES20 was added in api level 8, libgdx's minimum is already 9.

Several functions rely on an unchecked cast from Buffer to (IntBuffer). It might be better to update the base GL20 interface to restrict buffer type.

Initial runthrough of gdx-tests-android with newly generated natives seems to work without issue.

As usual needs jnigen code regenerated manually:
```
mvn install

cd gdx
mvn clean compile exec:java -Dexec.mainClass="com.badlogic.gdx.utils.GdxBuild"
```


noblemaster:
Given the deprecation warning by Google, it makes sense to have `AndroidGL20` call `GLES20` directly without JNI.

PokeMMO:
The jnigen artifacts still need regenerated for this.

noblemaster:
Not so sure how this has to be done? Can you take care of it?

PokeMMO:
Sure, will open a new PR in a few here.

noblemaster:
Thanks for the hard work! I have merged it. :)

PR #5644, opened 2019-06-01
LWJGL3 Maximized Window Positioning

thechubbypanda:
Allows the user to specify which window the app should maximise to.


thechubbypanda:
Good point!

PR #5648, opened 2019-06-02
GWT: Log to JS console if no log text area is specified

MrStahlfelge:
GWT logs the output to an own text area. I don't know why this was made this way, but I think logging to the console like it is done in the other backends is preferable. This is done here.

Due to backwards compatibility the old behaviour is still in place in case a logging text area is given via config.


MobiDevelop:
Looks good to me, thanks!

PR #5650, opened 2019-06-03
iOS RoboVM: Fix crash if getCurrentEventTime is called before first t…

MrStahlfelge:
…ouch was made

This can happen if the game is controlled without touches.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidInput.java
@@ -144,7 +144,7 @@ public class AndroidInput implements Input, OnKeyListener, OnTouchListener {
 	private InputProcessor processor;
 	private final AndroidApplicationConfiguration config;
 	protected final Orientation nativeOrientation;
-	private long currentEventTimeStamp = System.nanoTime();
+	private long currentEventTimeStamp = 0;
 	private final AndroidOnscreenKeyboard onscreenKeyboard;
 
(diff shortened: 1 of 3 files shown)


Tom-Ski:
Docs state that this is the time of the event currently reported by the InputProcessor. When are you calling this?

MrStahlfelge:
It's called by a GestureDetector if it is notified by a faked touchDown or touchUp event. My extension to make Scene2d.ui useable by button and key fires these events on Actors to simulate user input, as touchDown/touchUp events are built-in.

https://github.com/MrStahlfelge/gdx-controllerutils/blob/4e62c1b59b1796d63c02140477fa57c7db2d619d/core-scene2d/src/de/golfgl/gdx/controllers/ControllerMenuStage.java#L256

Before you judge that this is not intended usage, please take into account that all other backends return 0 if this method is called before the first touch is done. The problem just exists on iOS. Thanks :)

Tom-Ski:
This is not the case on android, which is what we follow mostly for the ios backends. All the other backends seem to do this more appropriately anyway. Instead of doing this null check, why not store the time as a field like the other backends are doing. Default value of 0 (maybe we should also change android to have this default too to be consistent)

MrStahlfelge:
That's a good way, too. I will change it like that and make a new PR if you prefer it.

Yes, Android returns the startup time - that's weird. I will change it to 0, too.

MrStahlfelge:
Done for MOE, too.

Tom-Ski:
Awesome, thanks!

PR #5652, opened 2019-06-04
Fix - transforming the string into a float to avoid an exception when…

thoced:
… a floating point position is introduced in the TiledMap software

When loading an image layer from the Tiled software. An exception arises when the displacement values ​​x and y are not integers.

Tiled software allows positioning with float value.

The loadImageLayer method of the BaseTmxMapLoader class generates an exception "java.lang.NumberFormatException" because it tries to transform a string into an integer from possibly a string that can give a float value


TCreutzenberg:
I like it 😉

PR #5661, opened 2019-06-07
GWT HttpResponse throw exceptions on unavailable operations

MrStahlfelge:
Binary operations on HttpResponse are not available on GWT, unfortunately the implementation just returns null which can lead to confusion. This small change makes the implementation throw an Exception with a hint for the developer how to solve the problem.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java
@@ -58,7 +58,8 @@ public class GwtNet implements Net {
 
 		@Override
 		public byte[] getResult () {
-			return null;
+			throw new GdxRuntimeException("HttpResponse.getResult() not available on GWT. " +
+					"Use getResultAsString() if possible, or make use of AssetDownloader class.");
 		}
 
 		@Override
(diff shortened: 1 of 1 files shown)


intrigus:
This could be fixed if GWT implements a getResult() byte[] method here https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/http/client/ResponseImpl.java#L102 using https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/xhr/client/XMLHttpRequest.java#L179
But chances are slim that GWT will change this I guess.

So LGTM.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java
```diff
@@ -68,7 +69,8 @@ public String getResultAsString () {
 
 		@Override
 		public InputStream getResultAsStream () {
-			return null;
+			throw new GdxRuntimeException("HttpResponse.getResultAsStream not available on GWT. " +
```
intrigus: Grammar `is not available`

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java
```diff
@@ -58,7 +58,8 @@ public HttpClientResponse (Response response) {
 
 		@Override
 		public byte[] getResult () {
-			return null;
+			throw new GdxRuntimeException("HttpResponse.getResult() not available on GWT. " +
```
intrigus: Dito

PR #5676, opened 2019-06-19
iOS: Add possibility to add iOS device information in game projects

MrStahlfelge:
New iOS devices pop up regularly, but existing games can't always be updated to the latest libGDX version. This PR makes the device list accessible to the game code. You can add new iOS device information via `IOSApplicationConfiguration.addIosDevice()`.


Tom-Ski:
It's not ideal as we discussed, but the enum is too restrictive for when new devices pop up and we aren't quick enough. Think its the best solution for now.

PR #4265, opened 2016-08-22
I wanted to improve the loading process of an HTML5 game with libGDX.…

MonsterOfCookie:
… As currently the process is.
1. Load index.html
2. Synchronously load the javascript files
3. Preload assets, get them cached in the browser (This is currently, where a player would see the libGDX logo and a progress bar).
4. (optional) Loading assets into memory with a custom loading screen via an AssetManager

In smaller games this is can be acceptable loads times. However as a game grows I have generally found that whilst the code size (The JS files are around 1-3MB) the art (Artists demanding HD level quality) the process of getting a player into the game becomes longer and drawn out and offputting. I have also been involved in projects where a HTML5 gaming platform would require a single loading experience for all binary and asset loading, providing their own loading screen etc.

I have a working example of a game where the JS is split and a boot strap occurs first then the main game JS is loaded separately. Then any assets are then loaded on demand, e.g. the AssetManager wishes to load a texture, this request is intercepted and the browser instructed to download the asset, then the loading into memory can occur.

Before I unleash this all on the world I wanted to see if I could clean up a couple of things in the GWT backend to make what I did simplier and cleaner looking. So this commit is more about setting the stage ready for the next phase of the loading.

When going through the code, I discovered that the canvasId that can be specified in the configuration is never used. Thus never letting you create and specify your own canvas, this commit fixes that.
1. I moved the preloading code into separate classes and I created an interface to allow other people to create their own preloaders and reuse the existing download code.
2. I have removed the config.root property as it was no longer needed, the code now only checks for the presence of the canvasId field.
3. If the canvasId does exist, the default preloader will still create the progress bar and libGDX logo except upon completion it no longers creates the canvas

This is backwards compatible and I have tested it the project created via the project setup tool.

I hope this makes sense to people, if I have done something I shouldn't have please let me know and I will fix it.

PR #5677, opened 2019-06-19
GWT possibility to lazy load assets

MrStahlfelge:
Based on the idea of #4265, I implemented a solution to skip the GWT preloader if it is not needed.

The preloader prefetches all assets before the game is started. That's useful for all assets that are accessed directly from the render thread and can't get fetched asynchronously, but assets that are loaded via AssetManager don't need to be preloaded.

This PR extends the AssetFilter with a new method that can be overriden to filter assets that don't need to be preloaded, and changes the Preloader to respect this setting and the AssetManager (via AssetLoadingTask) to download the asset if not already done.

To test this, specify an own AssetFilter in your game (as described [here](https://www.badlogicgames.com/wordpress/?p=2372)) and override the `preload()` method. Only return `false` for assets that are only loaded via AssetManager and not with `AssetManager.finishLoading[Asset]`. :)

Runnable demo game: [Lazy loading version](https://www.golfgl.de/sccplf/experimental/) - open the developer network tab, you'll see that the game load its assets on level start.

For those wanting to try or to use this in existing games: you can find Jitpack dependencies to use with libGDX 1.9.8 (working with 1.9.9, too), 1.9.10 and 1.9.11 [here](https://github.com/MrStahlfelge/gdx-backends).


MrStahlfelge:
We don't need to wait for other reviews, but I need to update CHANGES, make a comment on the main Asset manager to note that there is an emu (and compare them for changes) and will transfer the information from the first comment to the wiki. So it's good you waited. :-)

PR #5680, opened 2019-06-23
[spritebatch] Always activate texture slot 0

RUSshy:


NathanSweet:
Can you please explain why this change is better?

RUSshy:
I had one issue on my C# port, where it would messup with textures (render previous texture instead of current one) when using color texture of Framebuffer

Maybe because i updated shader syntax to 3.3?

I'll try to come up with a project to reproduce the issue with Java

Here is my simplified "port" of libgdx https://github.com/ryuukk/arc_cs/

Tom-Ski:
If you mess with glActiveTexture before SpriteBatch gets to flush and you don't restore it, you will have invalid texture.  This only happens when you don't restore gl state correctly after doing things externally from SpriteBatch.

RUSshy:
That would explain my issue then, i'm not sure i restored the gl state correctly

I wish there was some sort of automatic gl state management like in monogame, maybe https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderContext.java ? but it is missing some stuff

PR #5556, opened 2019-02-09
libgdx/jnigen armhf/aarch64 native build support

PokeMMO:
- Added arm support to BuildTarget/AntScriptGenerator.
- Created default BuildTarget's for arm32(debian's armhf)/arm64
- Fixed aarch64 (arm64) not being detected correctly in SharedLibaryLoader.
- JniGenSharedLibraryLoader is now more consistent with SharedLibaryLoader.
- ~~SharedLibraryLoader#extractFile and SharedLibraryLoader#extractFileTo now have optional resultFileName parameter to support renaming during extraction.~~
- ~~Renamed private SharedLibraryLoader#extractFile to SharedLibraryLoader#extractFileIfRequired to differentiate method signature. (I'm open to better name suggestions)~~
- ~~LwjglNativesLoader will utilize renaming support to always use liblwjgl.so on disk to match expected library name.~~
- Kept as much backwards compatible as possible.

# Things to note: #
- If compiler is not on path, jnigen build scripts simply skip builds. See https://github.com/libgdx/libgdx/blob/a4805d6a017b80622d6bfdd3a791352257a3c539/extensions/gdx-jnigen/src/com/badlogic/gdx/jnigen/resources/scripts/build-target.xml.template#L48-L67
- Ubuntu's multiarch env for ports ~~seems broken~~ is finicky but can be made to work. 
- armhf (gnueabihf) is not compatible with rpi 1/zero. Debian's armhf is armv7+ while the pi 1/zero is armv6+hf.
- "GL (Fake KMS) Desktop Driver" mode was sometimes required on raspbian to get decent performance.

# Rasperrypi custom arm-linux-gnueabihf-gXX: #
- Raspberry pi provides a custom toolchain that supports [armv6 hf](https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf)
- Using this toolchain is required for rpi 1/zero support.
- Requires you to add custom arguments and or copy dependencies manually.
- `cp -r /usr/include/X11 $PATH_TOOL_TOOLS/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/arm-linux-gnueabihf/include/X11/`
[...]

PR #5681, opened 2019-06-23
jnigen: Fix double quote issue in BuildExecutor.

PokeMMO:
* Added missing working directory change in BuildExecutor#startProcess

Original changes were originally in https://github.com/libgdx/libgdx/pull/5556 but includes additional changes from https://github.com/MobiDevelop/gdx-jnigen


PokeMMO:
Commented out calls to executeAnt with old space delimited string parameters aren't updated here. I'd rather remove/fix them after #5556 gets merged to avoid conflicts.

PR #5682, opened 2019-06-23
Adding Gyroscope implementation for gwt

SimonIT:
I added a output of the values to the AccelerometerTest to validate the correctness of the values.
You must use https for using the Gyroscope API because it's only available in Secure Context.
The updated version of the tests are running [here](https://libgdx.v22018086706171227.nicesrv.de/).


MrStahlfelge (approved):
Looks good, it is light-weight, non-breaking and a good enhancement for the gwt backend.
My only concern is if you tested it under feature policy restrictions?

SimonIT:
Good catch! If you wanna try it yourself:
Inside iframe: https://libgdx2.v22018086706171227.nicesrv.de/ 
Without iframe: https://libgdx.v22018086706171227.nicesrv.de/

PR #5683, opened 2019-06-24
Getting the screen orientation in gwt

SimonIT:
The implemetation is based on https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation


MrStahlfelge:
Interesting! What does it return on desktops and did you test on iOS, too?

SimonIT:
It returns Landscape on desktop, I have no iOS device to test. But if the API is not available, it should return the previous default of Landscape.

SimonIT:
Should work now. I assume now that the native mode of mobile devices portrait mode is

MrStahlfelge (changes requested):
There's a way you can do this slightly better. Web browser can return angle and orientation. With the The "screen orientation values table" from here: https://www.w3.org/TR/screen-orientation/ , the native orientation can be calculated. 
By using the angle, you probably can implement getRotation a bit better.

MrStahlfelge (approved):
Looks good now. I tested it and it didn't work on Safari and Firefox though. Problem was the old firefox fallback didn't support angle, and Safari doesn't support this at all. 

It now works on:
Chrome/Chrome mobile
Firefox/Firefox mobile
Safari/Safari mobile (returning defaults)

PR #5685, opened 2019-06-26
Moved allocateIterators config to new Collections class

obigu:
Created a new Collections class to hold common configuration/utility methods for all collection classes. Moved allocationIterators from Array class to Collections.

I've made a getter/setter for the property, can change to public as it's not that great anyway.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -33,7 +33,7 @@
 - API Addition: new InstanceBufferObject and InstanceBufferObjectSubData classes to enable instanced rendering.
 - API Addition: Support for InstancedRendering via Mesh
 - API Change: Cell#setLayout renamed to setTable.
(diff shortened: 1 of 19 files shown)


NathanSweet:
I agree putting it on Array is a bit ugly. I wish there was a good place to put it other than to have a whole class for one boolean, but I don't have any better ideas. `isAllocateIterators` quite is an awkward getter name, `getAllocateIterators` would be better (being boolean doesn't require `is`!). However, I think I'd prefer it as just a public field named `allocateIterators`.

obigu:
Fixed, replaced by getter/setter by public field. Hopefully there's some common functionality we can add to the new Collections class so that is has a better reason to exist

PR #5689, opened 2019-06-27
removed a line disabling autoBind when passing Renderable to ModelBat…

ryanastout:
before this change, ModelBatch.render behaves differently depending on whether a Renderable or a RenderableProvider is passed as the argument. If a RenderableProvider is passed, autoBind is not disabled. However, if a Renderable is passed, autoBind is disabled, which causes an exception unless the custom Shader takes steps to force autoBind to be used. 

I tested this by creating https://gist.github.com/ryanastout/ec842b8ce63342834909c07ff37c66f8 and verifying that I don't see the exception with this change, but I do without the change from this commit

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/ModelBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/ModelBatch.java
@@ -225,7 +225,6 @@ public class ModelBatch implements Disposable {
 	 * @param renderable The {@link Renderable} to be added. */
 	public void render (final Renderable renderable) {
 		renderable.shader = shaderProvider.getShader(renderable);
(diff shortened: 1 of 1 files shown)


ryanastout:
@xoppa I thought you might be in the best position to evaluate if this is a good change to make or if it will break something I'm not aware of.

mgsx-dev:
I dig into this topic with @ryanastout. This line of code is very old, before MeshPart was introduced. I think it was a quirk long time ago but it's irrelevent now since BaseShader don't use this variable (performing manual binding).

RenderableProvider (mainly ModelInstance) contains Mesh with default autoBind to true anyway. It's totally safe to remove this quirk. 

More, this code was causing issues with custom shader implementations relying on autoBinding and was prone of side effects. So, thanks for the fix!

PR #5696, opened 2019-06-29
try fix quick pause/resume

zaijie1213:
in some case,It works


MrStahlfelge:
Can you put in some information here on the problem you are trying to fix here?

obigu:
May be related https://github.com/libgdx/libgdx/pull/4326

Tom-Ski:
Yes it would be addressing those issues, but it doesn't explain the reasoning behind this code change, and the description isn't that comforting. Not mergable until @zaijie1213 provides more info.

zaijie1213:
when Fragment resume and quickly paused, onDrawFrame method may  have No chance to execute,
then deadlock will happen.in this case, we can skip wait.

PR #5697, opened 2019-06-29
Android: keyDown events fire multiple times, isKeyJustPressed returns…

MrStahlfelge:
… true as long as the key is held

Fixes #5444


Tom-Ski:
Seems sensible, thanks!

PR #5698, opened 2019-06-29
Improve code formatting

SimonIT:
I noticed by contributing that almost every file has a different code formatting. Some files use line feed, some carriage return line feed. I saw a lot of PR's which had problem with this (me including).

I worked on an integration of the plugin [spotless](https://github.com/diffplug/spotless). Spotless checks at build time if all files are correct formatted. It uses the specified rules from the gradle configuration. I imported the [eclipse-formatter file](https://github.com/libgdx/libgdx/blob/master/eclipse-formatter.xml) for the formatting rules. If the build fails because of wrong formatting, you can format the sources by typing `gradlew spotlessApply`. This makes it more comfortable for contributors.

**TODOs**:
- [ ] Maybe add some rules (also for other file types)
- [x] Wait for diffplug/spotless#412 because of wrong formatting of JSNI of GWT ([eclipse formatter bug tracker](https://bugs.eclipse.org/bugs/show_bug.cgi?id=222083))


Tom-Ski:
Can be useful for a PR builder for new changes, but I would not like to run this on the entire project.

SimonIT:
That will be a problem. But if the formatter would be run on bad formatted file with the wrong file ending, also this PR will have a lot changes which will lead to not getting merged. 
Also `gradlew build` will check every time if all files a right formatted. The build will fail if not. So I don't think there's a way to avoid a big commit for formatting.

Tom-Ski:
I don't think line endings are worth having as a strict rule.

SimonIT:
I think that's the greatest benefit. I think almost every contributor had in a PR changed line endings because of the default setting of git to convert to line feed. It would really help to avoid messi PRs with a lot of changes because the line endings were converted by the os, idea or git. And it's always exhausting to revert the line endings

Tom-Ski:
I agree. This tool wouldn't help with that though? It would just let them know why they are not following formatting guide.  Why don't we use .gitattributes and set a per-repo setting for unix style endings.  Then we wouldn't have this two stage process of someone pushing and only then seeing the problem. 
This also would involve quite a big change as git will want to enforce the normalized line endings on all files that aren't subscribing to that, but probably less than if we were to run this plugin.

SimonIT:
Off course does the plugin help. `gradlew spotlessApply` will change the line endings to specified ones. I set `lineEndings 'UNIX'` to use line feed as line ending. For sure we can also use .gitattributes. I would suggest only as addition and not as replacement for the spotless line ending configuration. Spotless warns the user and git does not, so without spotless it would be possible that someone pushes still the wrong line ending.

Tom-Ski:
What I'm saying is that includes the additional step, which is not a huge amount different to just using your IDE to do it.

With git attributes, correct me if I am wrong, but git will always convert the files to the supplied line ending on checkout.  The only way someone would be able to push something with incorrect line ending would be to actively bypass this by manipulating the projects' gitattributes before checkout.

SimonIT:
There is also a maven version of spotless. It's an additional step, yes. I think typing one command is not to much effort.
Yes, I think git does it like you described.

tommyettinger:
I strongly agree that the formatting should all be standardized, once, over every file, then automatically for future contributions. Most existing (older) code doesn't appear to have ever had the formatter run on it, which is mind-boggling if PRs are getting held up or rejected for functionality-unrelated formatting reasons. So either the formatting matters, and the project should hold itself up to its own standards, or the formatting doesn't matter, and no PRs should be blocked because they aren't pretty enough. The middle ground we're in where "it was written in 2015 so we can't change it, even though it isn't the formatting we would allow in a PR" but also "it was partially written in 2020 so you can't change some lines with the formatter, but you need to run the formatter to get merged" is absurd.

tl;dr If you're going to have formatting rules they should apply to libGDX itself, otherwise what's the point?

NathanSweet:
I would be willing to accept the one time pain of formatting all the source if we thereafter do it regularly and we have an IDE independent way of formatting the source. I have applied formatting to the entire source in the past, but it wasn't kept up.

I'm not sure about failing the build for not having the right format. I like the enforcement, but I worry it proves annoying in practice. The alternative is to regularly make commits to fix formatting, which adds some noise (which may not matter much, as our repo is relatively quiet anyway).

I tried applying this PR and ran `gradlew spotlessApply`. The initial formatting is pretty intense at 3722 files, but that's what we get for leaving it this long! It does appears to match the Eclipse formatter exactly. This makes me _super happy_ and I would love to see it merged, especially since [recently](https://github.com/libgdx/libgdx/pull/5900#pullrequestreview-358651387) it was discovered that IDEA does not match the Eclipse formatting when importing the Eclipse formatter file. Spotless even applies missing copyright headers. The only issue I saw were comments like this one from AndroidAudio:
```
				soundPool = new SoundPool(config.maxSimultaneousSounds, AudioManager.STREAM_MUSIC, 0);// srcQuality: the sample-rate converter quality. Currently has no effect. Use 0 for the default.
```
Gets turned into this monstrosity:
```
				soundPool = new SoundPool(config.maxSimultaneousSounds, AudioManager.STREAM_MUSIC, 0);// srcQuality: the sample-rate
																																	// converter quality. Currently
																																	// has no effect. Use 0 for the
																																	// default.
```
There may be other situations that were not intended to be formatted and format poorly. These situations are probably rare and can be fixed up manually over time.

SimonIT:
We can disable the enforcement with this line (see [here](https://github.com/diffplug/spotless/tree/master/plugin-gradle#disabling-warnings-and-error-messages)):
```groovy
spotless {
  enforceCheck false
}
```

tommyettinger:
*sees 3722 lines* <-- Oh, that's not so bad.
*re-reads*
*sees 3722 files* <-- Jaw drops.

Are they all Java files, none generated/JNI-related? Is there a way we can make sure JSNI for GWT is still working OK, as well as jnigen comments? I'm wondering if the AndroidAudio "monstrosity line" can be adjusted by changing the tab width; with tab ==  2-4 spaces it should be much closer to the area it should occupy. It would be amazing if this plugin can do the code-style changes correctly across-the-board, and it seems like it's within reach!

SimonIT:
I read on the eclipse bugtacker https://bugs.eclipse.org/bugs/show_bug.cgi?id=222083 that the formatting is a problem which exists since 2008 and seem to get fixed. Mateusz Matela suggested there on [comment eight](https://bugs.eclipse.org/bugs/show_bug.cgi?id=222083#c8) to use custom on off tags for the formatter. _/*-{_ as off tag and _}-*/_ as on tag. The problem is, that we got already custom on and off tags:
https://github.com/libgdx/libgdx/blob/d765b94eed4503d8389ef97a770316bd4f4561f1/eclipse-formatter.xml#L5
https://github.com/libgdx/libgdx/blob/d765b94eed4503d8389ef97a770316bd4f4561f1/eclipse-formatter.xml#L87
I don't know, if it's possible to use multiple custom on off tags and if so, how to add the to the xml file.

I got no other ideas how we could fix this problem.

By the way, spotless can also format for example xml, gradle and  cpp files

NathanSweet:
It's some XML and CPP files, but I think just fixing spaces->tabs there (which is good).

The on/off tags are OK for the rare block where it's really needed, whether our current tags or something else. The current tags are maybe less cryptic that `/*-{`. It doesn't help the AndroidAudio problem. For that the comment should just be on its own line(s), not at the end of a long line. We could run the formatter for the first time, then before committing, search for "tabtabtabtabtabtabtabtabtabtab..." and manually fix things up.

SimonIT:
Note:
As stated [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=222083#c10), no multiple custom formatter tags are supported, we have to add `// @off` and `// @on` to almost every jsni block, else things get dirty.
Should I disable the forced formatting?
Should I add these tags to this PR? If yes, I would add it also to the mentioned SoundPool line.

NathanSweet:
Would it help if we change to the `/*-{` and `}-*/` tags?

SimonIT:
Yes, but the current on/off tags are used for the jni-gen code (If I'm not wrong) and for some Shader declarations. I think replacing them with the jsni comments will let the IDEs freak out and make the whole developing a pain.
I don't see any other option to add off tags to every jsni block, but I'm happy to hear some ideas for easier solutions

payne911:
My personal opinion: **this is a priority**.

This repository has seen an increase in activity and there are already problems with the formatting for a few PRs.

crykn:
I think a much more pressing issue is getting a proper IntelliJ formatter setting. The _idea-formatter.xml_ file in the repo produces differente results in Android Studio than _eclipse-formatter.xml_ does in Eclipse, just see here:

**Eclipse:**
eclipse

**Android Studio:**
intellij

And importing the eclipse formatter file in Android Studio (as is recommended on the libGDX homepage) doesn' work, because: "The file doesn't seem to be a valid Intellij IDEA XML scheme, it should contain 'code_scheme root' element, found 'profiles' instead."

payne911:
There is an IntelliJ plugin for "eclipse-formatter", btw.

Anyways, I'm quite certain that I've seen something in IntelliJ formatting that allowed better customization of the comments.

crykn:
Hey Payne, in the screenshots I used the `idea-formatter.xml`. 

I couldn't get Android Studio to import the eclipse formatter files, even though that should [be possible](https://blog.jetbrains.com/idea/2014/01/intellij-idea-13-importing-code-formatter-settings-from-eclipse/).

payne911:
On the default capacities : https://stackoverflow.com/a/27035850/9768291

But anyways, I would suggest using the Eclipse-Code-Formatter plugin : https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter

abueide:
you can make the "classes" task depend on the "spotlessApply" task so any time you run build, jar, or run tasks it automatically applies the formatter, and therefore we can leave it so it fails build if the style is incorrect for ci/cd purposes.

nedtwigg:
I'm a Spotless maintainer, I just noticed this issue link and wanted to share some thoughts.  If you want to avoid a "format everything" commit, you can use [ratchetFrom](https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet) so that only files which have changed since `origin/master` will get formatted.  It's the same no-bikeshedding `apply / check` workflow, but just more fine-grained (and faster!).  I think a "format everything" commit is still a good idea, but to each their own 👍 

I see you have places where you want to disable formatting inline.  We're about [to merge](https://github.com/diffplug/spotless/pull/691) a piece of infrastructure which makes that possible for all formatters, not just eclipse, so if you want to ignore sections inside of XML or CPP, you can do that.  Once that gets merged, it becomes pretty easy to implement the [javascript inside of java](https://github.com/diffplug/spotless/issues/412) issue, ala

```
native static public void consoleLog(String message) /*-{
	console.log( message );
}-*/;
```

Lastly, Gradle's caching makes Spotless really fast.  So I've personally never felt much need for IDE support.  But, we do have an [IDE hook](https://github.com/diffplug/spotless/blob/main/plugin-gradle/IDE_HOOK.md), although so far only VS Code has a plugin for it.

SimonIT:
Okay guys, here we are. I configured spotless so that only files which have changed since origin/master will get formatted like @nedtwigg suggested. I think that's the best currently to avoid a big formatting commit.
Another idea I had is a pre-commit hook but that's a different story....

.github/workflows/build-pullrequest.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          fetch-depth: 0
```
tommyettinger, line 14: Setting fetch-depth to 0 seems to be the only major change to the GH Actions .yml files; what does it do in this case? I don't have the slightest clue about GH Actions, but it just might be good to have the rationale noted in this PR.

build.gradle
```diff
@@ -16,6 +16,16 @@
 
 apply from: "gradle/dependencies.gradle"
 
+buildscript {
+    repositories {
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+    dependencies {
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:5.14.3"
```
tommyettinger, line 24: Should we track the spotless version in dependencies.gradle , alongside most other versions? I imagine it gets updated more frequently than most other dependencies we have.

tests/gdx-tests-gwt/webapp/index.html
```diff
@@ -1,25 +1,25 @@
...
+<html>
+  <head>
+    <meta http-equiv="content-type" content="text/html; charset=UTF-8">
+    <title>libGDX GWT Test Runner</title>
+  </head>
+  <body>
+    <h1>libGDX GWT Test Runner</h1>
+    UI should load here shortly:<br>
+    <div id="embed-com.badlogic.gdx.tests.gwt.GdxTestsGwt"></div>
+    <small>This page is from <code>tests/gdx-tests-gwt/war/index.html</code>. Enter superdev here:
```
tommyettinger, line 11: Is this path incorrect now? I know the webapp folder gets copied to war, but if people are looking for this file to edit it, they should go to webapp.

SimonIT:
The gwt formatting problem is also resolved

tommyettinger (approved):
Mostly, this looks great! The only concerns I have are very minor (maybe relocating the version for spotless, and maybe double-checking the path in the GWT test index.html), and if this were merged as-is I would still be happy as a clam. Approving.

.github/workflows/build-pullrequest.yml
SimonIT: GitHub does a bare checkout by default, which seem to have no enough information to make the difference check (If I understood correctly what @nedtwigg said diffplug/spotless#710)

nedtwigg: Correct. If you want to only format the files which have changed in the PR, then you need a git clone with enough history to determine which files changed. A bare checkout (the default on most CI platforms) doesn't have that info, but you get all the info you need by switching to `fetch-depth: 0`.

tommyettinger: Good to know! That explains it.

tommyettinger:
Great! It's merged! Are there any instructions to contributors about how to use this (so people don't have to go through this whole thread)?

SimonIT:
Spotless is very self-explaining.

For example:
I changed something in `DefaultGwtInput`.
The following error appears because of a none matching format:

```

* What went wrong:
Execution failed for task ':backends:gdx-backends-gwt:spotlessJavaCheck'.
> The following files had format violations:
      backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```
With this instruction at the end:
````
Run './gradlew :backends:gdx-backends-gwt:spotlessApply' to fix these violations.
````

After running the command is the file correct formatted and the project build successful.

It's very easy to use

obigu:
Being the format now enforced on the build server any new commits will contain lots of format changes unrelated to the PR just for the build not to fail.
To avoid that, either we run it ASAP on the whole project as a one time thing (and break most open PRs) or we remove the enforcement check on the build server.
@NathanSweet How do you want to proceed?

tommyettinger:
I think many open PRs, at least the larger ones, are probably broken to some extent by the spotless requirement, unless the author already ran the spotless task. If we have the capability to edit PRs, any contributor interested in merging a PR can run the spotless task on it to allow it to merge cleanly, as long as spotless doesn't risk changing many unrelated files. It shouldn't run that risk if we've formatted the whole project with spotless.

`.git-blame-ignore-revs` was mentioned as a possible solution for hiding the changes spotless makes in one commit, but I don't know the specifics.

PokeMMO:
> `.git-blame-ignore-revs` was mentioned as a possible solution for hiding the changes spotless makes in one commit, but I don't know the specifics.

https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame goes over it a bit, unfortunately github still doesn't support it.

nedtwigg:
> any new commits will contain lots of format changes unrelated to the PR

It should be that only files which the PR has changed will get formatted. https://github.com/diffplug/spotless/tree/main/plugin-gradle#how-can-i-enforce-formatting-gradually-aka-ratchet

https://github.com/libgdx/libgdx/blob/0be3ba94aeb3fecd1c4c09deb3e95c154e8f2457/build.gradle#L66-L67

obigu:
Yes, it only changes the touched files but it reformats the Whole file, not
just the changed lines. In many cases, just for changing a line in a PR it
will  reformat hundreds of lines on that same class.

NathanSweet:
I'm OK with a one time format of everything. It does make the history a little annoying and cause some pain with PRs, but at least it's only once. We should first make sure the format is what we want. If it's the Eclipse formatter than that should be OK. The commit would also benefit from manual review, looking for areas that have formatted poorly and would be improved with `// @off` and `// @on` comments.

PR #5701, opened 2019-06-30
jnigen/gradle: Add new task `jnigen` which regenerates jnigen artifacts.

PokeMMO:
* One line command `./gradlew jnigen` will regenerate all projects.
* Each BuildTarget gets a new task named `jnigenBuild${os}${64}`
*  `./gradlew jnigenBuild` task will build all builds that do not require a mac host.

**TODO**:
* [x] Potentially make `requireMacOSToBuild` a generic `requireOSToBuild` or an even more generic condition.
* [x] Support providing `APP_STL` and other extra lines to `Application.mk` `Android.mk`
* [x] Add better jar tasks for Desktop/Android/IOS.
* [ ] Add clean task.


PokeMMO:
It might actually be simpler to just specify all possible options eg `bin:build/classes/java/main:target/classes` which would be legacy, gradle, and maven. I think javah will simply pick the first it finds, but will have to verify.

Tom-Ski:
Maybe a custom plugin via buildSrc might be nice?

PokeMMO:
> Maybe a custom gradle task that can be configured via buildSrc might be nice?

Initial PoC buldSrc gradle task created. Should be moved to it's own `gdx-jnigen-gradle` project and published as a proper plugin.

The `buildSrc/build.gradle` and `buildSrc/settings.gradle` could probably be done better, I only spent as much time on them as required to get it to work.

@Tom-Ski any feedback would be appreciated.

buildSrc/.classpath
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Tom-Ski: Does eclipse not have support for buildSrc via buildship?

Tom-Ski:
Perhaps we should think abut having gdx-jnigen extracted from the repo also with the plugin.

MobiDevelop:
I did some initial work to move gdx-jnigen out - https://github.com/MobiDevelop/gdx-jnigen, needs the maven publish piece but I could move this repo into the libgdx org if we want to start that process.

buildSrc/.classpath
PokeMMO: I don't typically use buildship in eclipse, so I'm not sure. I can always remove the eclipse artifacts.

PokeMMO:
I'll start by making a PR to MobiDevelop's gdx-jnigen project and setup a proper gradle plugin subproject in it.

PokeMMO:
Initial gradle plugin PR created https://github.com/MobiDevelop/gdx-jnigen/pull/1

This PR has been updated to no longer modify any gdx-jnigen project files, but simply omit them from other project's *gradle* dependencies. No modifications to maven or included eclipse project files have been done.

Nothing breaking changes have been made, so maven/traditional Java based Build classes can continue to be used.

MrStahlfelge:
Is this PR still needed/up to date?

PokeMMO:
Not up to date, but should be updated and still merged.

PokeMMO:
This PR has been rebased and updated. 

Should I also be removing `extensions/gdx-jnigen` with this or do that separately?
Should I be removing legacy build scripts (And ignoring them from git?) now or later?

Note: This is incompatible with full travis builds because travis can't run gradle on macOS due to missing Android SDK. We need to disable macOS builds until we migrate to Github Actions.

build.gradle
```diff
@@ -25,14 +24,19 @@ allprojects {
         repositories {
             google()
             jcenter()
+            mavenLocal()
             mavenCentral()
             maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
         }
+        dependencies {
+            classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.0.0-SNAPSHOT"
```
Tom-Ski, line 32: Lets do a release before we move to this?

Tom-Ski (approved):
Looks great, I'd just say lets release jnigen before using it here in core. 

Regarding killing old build scripts, I say yes.

build.gradle
PokeMMO: Probably a good idea, there's nothing I'm aware of that needs done before release. There's a few small tweaks/features which might be good, but they're not worth holding it back for.

Tom-Ski:
Also @PokeMMO what has this been tested on? 

If you haven't been able to test on all platforms, makes sense to get a checklist here and get some community help in testing anything you don't have device ability to test.

PokeMMO:
This should create exact matches of the output ant scripts, so it should be fully compatible out of the box. Except I broke something when moving to `javac -h` from `javah`. Only noticed because none of the other test projects have relative project dependencies on classpath.

Github Actions has tested everything else on all platforms, except maybe iOS. I'll finish testing what I can first and then ask for help to verify the rest.

PokeMMO:
Issues fixed, was InnerClasses being broken with `javac -h` and some classes having `compileOnly` preventing classpath from being populated correctly.

Linux/Windows/Android all compile correctly, will have to test macOS on GHA unless someone else can.

PokeMMO:
I got a little carried away making this PR delete as many lines as possible, and removed the old gdx-jnigen extension too. If you'd rather me split that into it's own PR after this one let me know.

PokeMMO:
@Tom-Ski
One last question before merge, is it acceptable that the build process require `./gradle jnigen` before running per-existing ant/maven based build for now?

Tom-Ski:
Yeah thats ok. Its a step in the right direction to consolidate imo.

Tom-Ski:
As a note to everyone, this is going to be merged on the weekend, so if you have any reservations, now is the time!

PR #5702, opened 2019-06-30
AndroidControllers: possibility to not ignore some buttons

MrStahlfelge:
ffe15f0d introduced that AndroidControllers extension ignore all controller buttons that are not considered a Gamepad button. However, some users could be irritated that some buttons on their controller can't be used. With this commit, the developer can switch to the old behaviour and the game will get events for all buttons the controller provides.


MrStahlfelge:
Added this to be backwards compatible.

PR #5705, opened 2019-07-03
ignore temporary folders of roboVM

guichazan:
these folders are generated by roboVM and should not go to Git


Tom-Ski:
You are patching the gitignore of the main source folder. THis has nothing to do with generated projects

PR #5714, opened 2019-07-18
Android/iOS: Added methods to get safe area insets

MrStahlfelge:
We have display cutouts on iPhone X and now after updating on Android API 28 we can also have some on Android (although Android's default behaviour avoids that you have to deal with them).

This PR adds support for iOS RoboVM and Android backend to return the safe area insets. The changed `GdxTestActivity` demonstrates how to expand the game under the Notch on Android. A further change to apply this automatically if immersive mode is set might be useful, but this can be discussed.

This is tested on Android 4.2 and Android 9 and iOS 12.1 with RoboVM. The FullScreenTest tints the badlogic logo red if the logo is outside of the safe display area.


noblemaster:
I think that's a great idea!

PR #5729, opened 2019-07-27
Add anisotropic filtering parameters for textures

cypherdare:
I've been using the same basic code in live wallpapers for a few years. It doesn't do anything if you never set it.


gdx/src/com/badlogic/gdx/graphics/GLTexture.java
```diff
@@ -164,6 +170,62 @@ public void setFilter (TextureFilter minFilter, TextureFilter magFilter) {
...
+	 * @return The actual level set, which may be lower than the provided value due to device limitations. */
+	public float unsafeSetAnisotropicFilter (float level){
+		return unsafeSetAnisotropicFilter(level, false);
+	}
+
+	/** Sets the anisotropic filter level for the texture. Assumes the texture is bound and active!
+	 * @param level The desired level of filtering. The maximum level supported by the device up to this value will be used.
+	 * @param force True to always set the value, even if it is the same as the current values.
+	 * @return The actual level set, which may be lower than the provided value due to device limitations. */
+	public float unsafeSetAnisotropicFilter (float level, boolean force){
```
Tom-Ski: Inconsitent method formatting. Should have space after method name, and after the final param parenthesis.

Tom-Ski:
Looks good other than formatting inconsistencies.

gdx/src/com/badlogic/gdx/graphics/GLTexture.java
```diff
@@ -164,6 +170,75 @@ public void setFilter (TextureFilter minFilter, TextureFilter magFilter) {
...
+	 * @param level The desired level of filtering. The maximum level supported by the device up to this value will be used.
+	 * @param force True to always set the value, even if it is the same as the current values.
+	 * @return The actual level set, which may be lower than the provided value due to device limitations.
+	 */
+	public float unsafeSetAnisotropicFilter (float level, boolean force) {
+		float max = getMaxAnisotropicFilterLevel();
+		if (max == 1f)
+			return 1f;
+		level = Math.min(level, max);
+		if (!force && MathUtils.isEqual(level, anisotropicFilterLevel, 0.1f))
```
intrigus: Isn't 0.1f a pretty bad value for an epsilon?

cypherdare:
@intrigus AF levels are only ever whole number powers of two, so really 0.5 would be adequate. Unfortunately the GLES 2.0 spec defines them as floats or we'd be using ints instead. I discussed it with Tomski. We could use ints, but we'd have to pass floats to GL, and I was concerned about some implementations rounding differently when getting the current level for comparison.

intrigus:
Didn't know that they are only ever power of twos.
Only saw the suspicious comparison :)

PR #5733, opened 2019-07-29
jnigen: Support specifying `APP_ABI` in `Application.mk`

PokeMMO:
* Default changed to "all". 
* "all" will allow building on more recent NDK's, omitting armeabi builds.
* Updated generated Application.mk included for all libgdx projects.
* See https://developer.android.com/ndk/guides/application_mk#app_abi for more information.

This is step 1 on removing support for armeabi.


PokeMMO:
Closed in favor of #5701

PR #5734, opened 2019-07-29
SoundManager2: Changed mp3 required to false

GrantOjanen:
My game was blocked from running on an older version of Chromium, because the browser didn't support mp3. Changing required:!0 to required:!1 for mp3 inside SoundManager2 fixed my game and it runs now. The game uses ogg by the way.

When I say the game was blocked I mean the loading bar doesn't appear. I got this in the console.

SoundManager 2: No Flash detected. Trying HTML5-only mode.
SoundManager V2.97a.20150601 (AS3/Flash 9) + HTML5 audio
SoundManager 2: soundManager: Fatal error: Flash is needed to play some required formats, but is not available.
SoundManager 2 HTML5 support: mp3 = false (using flash), mp4 = false (using flash), ogg = true, opus = true, wav = true

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/soundmanager2/SoundManager.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/soundmanager2/SoundManager.java
@@ -16,8 +16,6 @@
 
 package com.badlogic.gdx.backends.gwt.soundmanager2;
 
-import com.google.gwt.core.client.JavaScriptObject;
-
 public class SoundManager {
 	
 	public interface SoundManagerCallback {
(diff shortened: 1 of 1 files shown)


mgsx-dev:
As per SM2 documentation, this is the right way to do it : http://www.schillmania.com/projects/soundmanager2/doc/

MP3 is required by default and it shouldn't.

MobiDevelop:
Did you try adding `$wnd.soundManager.audioFormats.mp3.required = false;` to [`SoundManager.init(...)`](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/soundmanager2/SoundManager.java#L76)?

That feels safer than modifying the minified javascript file of a library that could be updated.

GrantOjanen:
@MobiDevelop Yes, that works as well. I replaced my old commit. I'm unsure weather or not force pushing the commit like I did is OK. I'm new to using GitHub.

Darkyenus:
Force pushing into pull request branches are IMHO okay, unless you know (or can reasonably assume) that somebody else has checked out that branch and has done some modifications. It can still be okay, but it needs some additional communication.

But in most cases, force pushing keeps the PR clean and easy to review (with the exception, that if you are doing a small changes on top of already reviewed code, it may be easier to do them in a regular commit, so that people who have already reviewed past work can easily see what has changed).

PR #5735, opened 2019-07-29
Made isFullscreen() work on Android

GrantOjanen:
Tested on Android 8 and Android 4.4. I haven't tested it with live wallpapers or daydreams, but it should always return true on those. It only returns false when in multi-window mode in activities or fragments.

Also I made it so FullscreenTest now displays true/false for fullscreen.


intrigus:
Isn't "android fullscreen" quite different from desktop/web fullscreen?

GrantOjanen:
@intrigus What do you mean by "different"? Android 7 has a split window mode as well as free-form window mode. The free-form window mode must be activated on tablets/phones. I've never actually used an Android 7+ laptop. I don't think developers can force their activity to windowed/fullscreen mode.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationBase.java
```diff
@@ -91,4 +88,8 @@
 	/** Returns the Handler object created by the application
 	 * @return The {@link Handler} object created by the application */
 	Handler getHandler ();
+
+	/** Returns the current Android Activity
+	 * @return The current Android {@link Activity} */
+	Activity getActivity ();
```
obigu: You probably don't need this, use `getContext()` and check if it returns instance of `Activity`

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
```diff
@@ -718,10 +716,10 @@ public void requestRendering () {
 
 	@Override
 	public boolean isFullscreen () {
-		return true;
+		Activity activity = app.getActivity();
+		return activity == null || Build.VERSION.SDK_INT < 24 || !activity.isInMultiWindowMode();
```
obigu: Not very readable in my opinion but I accept it's a matter of personal taste

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationBase.java
noblemaster: I don't think ``Context`` supports ``isInMultiWindowMode()``?

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
noblemaster: I would probably put brackets ( ... ) around the sub-conditions, but maybe I am little bit strict there too :)

noblemaster:
Let's wait for the checks to complete (still pending) and then merge.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationBase.java
obigu: `Activity` is a subclass of `Context` and in standard Android apps getContext() returns an Activity being LiveWallpapers an exception. I'm suggesting to check if `getContext()`returns an instance of `Activity` to then be able to call `isInMultiWindowMode()`.

noblemaster: I see your point, there indeed is a ``getContext()`` method. I agree with obigu, let's remove the ``getActivity()`` and use the ``getContext()``! :)

MrStahlfelge:
It is a good addition in my opinion, but has conflicts and unresolved requests for changes.
Please reopen the PR if you solved that.

PR #5738, opened 2019-07-30
Android controller vibration

GrantOjanen:
Tested on Android 4.4 and 8.1. The only controllers I have that support vibration on the Android devices are Xbox 360 and Xbox original (with USB adapter.) Those controllers only had vibration on the Android 8.1 device, but the Android APIs for it go back to Android 4.1. The amplitude control wasn't supported on either of the controllers, so the vibration was always at max. The Android APIs do support special patterns, but I didn't incorporate that.

It seems LWJGL doesn't support rumble.

JavaScript doesn't have rumble support yet:
https://github.com/w3c/gamepad/issues/19
https://www.chromestatus.com/feature/5705158763741184

I'm not sure if you want to add this to the APIs since it's so limited, but I figured I'd share it anyways.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+[1.9.11]
+- API Addition: added controller vibration for Android
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
 - API Addition: Accelerometer support on GWT
--- a/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidController.java
+++ b/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidController.java
@@ -16,6 +16,9 @@
 
 package com.badlogic.gdx.controllers.android;
 
+import android.os.Build;
+import android.os.VibrationEffect;
+import android.os.Vibrator;
 import android.view.InputDevice;
 import android.view.InputDevice.MotionRange;
 import android.view.MotionEvent;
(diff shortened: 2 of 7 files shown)


MrStahlfelge:
Thanks, added it to [gdx controller utilities](https://github.com/MrStahlfelge/gdx-controllerutils)

MrStahlfelge:
While the code for Android is good, there is no implementation given for the other backends. After testing, I discovered that it does not work on Android in most cases. I got it to work with a controller on Android 8, but on my current Android 9 device it doesn't work even with my Xbox controllers.

Hence, the practical improvement here is very small but comes with a major problem: Pulling this change in will break third party libs like gdx-controllerutils and sdl2gdx, both extending the current `Controller` interface. These two libs are currently in use when controller vibration is needed by a game, and we should only break them when the official implementations are reach feature equality. This PR does not do it.

I close this PR, but with a thanks for the contribution. As I said, the code is adopted in controllerutils and I hope to revive the official controller implementations with the experience made there in the near future.

F0RIS:
@MrStahlfelge Have you found any information why android gamepads vibration isn't solid across different versions of android? 
I connect DualShock4 to Galaxy S7 with android 8 and ```inputDevice.getVibrator().hasVibrator()``` returns false. But why? I have tried both usb and BT connection

MrStahlfelge:
There's an issue on AOSP stating that vibration is not supported at all on Android. It is accepted by Google, but it is not clear if or when the functionality is added

PR #5739, opened 2019-07-31
Get game controller id/index assigned by backend

GrantOjanen:
This is a new method that returns an id or index number generated for a game controller by the backend. They are assigned differently by each backend when the controller is connected.

Javascript returns the value of Gamepad.index
Android returns the value of InputDevice.getId()
Lwjgl2 & Lwjgl3 return an index

This is mostly useful if you need to interface with code on a specific backend. This would let you expand the controller functions without having to edit libGDX.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+[1.9.11]
+- API Addition: get ID for game controllers assigned by backend
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
 - API Addition: Accelerometer support on GWT
--- a/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidController.java
+++ b/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidController.java
@@ -157,4 +157,9 @@ public class AndroidController implements Controller {
 	public String getName () {
 		return name;
 	}
+
+	@Override
+	public long getSystemId () {
+		return deviceId;
+	}
 }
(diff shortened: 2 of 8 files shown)


CHANGES
```diff
@@ -1,4 +1,6 @@
 [1.9.10]
+- API Addition: get ID for game controllers assigned by backend
```
intrigus: This should go under 1.9.11 and not 1.9.10 I guess.

GrantOjanen: That was a mistake. I moved it to 1.9.11.

MrStahlfelge:
IMO having this method in the core interface does not make any sense. There's nothing platform-agnostic you can do with this information.

GrantOjanen:
@MrStahlfelge It's mostly for if someone needs to expand/fix something in the controller support for Android/Javascript without having to edit/update libGDX. The game engine I'm switching away from is closed source, but it's not really necessary when you know how to edit libGDX. Now that I have a good setup for editing libGDX I'm no longer in need of this. I'll be fine with it whether it gets accepted or not.

noblemaster:
I think this make sense to have an unique ID to identify a controller especially to identify if a controller changed? Any reason this shouldn't be added?

obigu:
I don't know much about controllers but I agree with @MrStahlfelge this doesn't seem right on the core interface. @noblemaster Isn't there a proper mechanism to notify when controllers are connected/disconnected? 
OP answers he doesn't needs this anymore but the original use case for it was to "fix/expand" a specific implementation. If a controller implementation needs a "fix" we want it done in libGDX project. Regarding the "expand" part I think we need a specific example to justify adding this.

MrStahlfelge:
Added feature `getUniqueId()` to [controllerutils](https://github.com/MrStahlfelge/gdx-controllerutils/wiki/Advanced-Controller-interface) in the meantime

MrStahlfelge:
>  Now that I have a good setup for editing libGDX I'm no longer in need of this.
Closing this because has conflicts, won't get updated and there's no reason to return the internal assigned id back to the core.

PR #5742, opened 2019-08-02
 jnigen: Allow changing the c/cpp compiler from gcc to clang.

PokeMMO:
Also includes two compiler issues to allow compiling with clang.

Backported https://github.com/bulletphysics/bullet3/pull/2232
And a fix for a fix for a "binds to a temporary object whose lifetime would be shorter than the lifetime of the constructed object" error.


PokeMMO:
Closed in favor of #5701

PR #5618, opened 2019-04-27
BaseTmxMapLoader refactoring Mk2

TCreutzenberg:
This is a replacement pull request for https://github.com/libgdx/libgdx/pull/5495

Refactored BaseTmxMapLoader, TmxMapLoader and AtlasTmxMapLoader.

Shared functionality was moved to BaseTmxMapLoader, duplicate and incomplete code was removed.

Basic load/loadAsync functionality remains the same and all TmxMapLoader-tests run normally.

The old codebase had a lot of duplicate code. This is fixed now but required a lot of methods to be changed completely.

Now TmxMapLoader and AtlasTmxMapLoader work the same way, and they are capable of the same things (this was not the case before).
New features can now be implemented and both TmxMapLoader and AtlasTmxMapLoader will automatically receive the benefits.

@MobiDevelop @mgsx-dev @Ktar5 This is a proper refactor. Please have a look at this pull request if you have the time, thank you!

PS: TiledMapGroupLayerTest was changed because the .tmx was missing from the repository so I just reused the test as it was not working anyway. Please check TiledMapGroupLayerTest  for groups and nested groups.

PR #5745, opened 2019-08-04
fix TiledMap animated tiles loading. fixes #5744

### LINKED ISSUE #5744: crash when loading an animated map
#### Issue details
GDX attempts to cast from AnimatedTiledMapTile to StaticTiledMapTile which causes an exception.

#### Reproduction steps/code
Load a map with animated tiles.

1.9.10

#### Stacktrace
```java
Exception in thread "main" java.lang.ClassCastException: com.badlogic.gdx.maps.tiled.tiles.AnimatedTiledMapTile cannot be cast to com.badlogic.gdx.maps.tiled.tiles.StaticTiledMapTile
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.addAnimatedTile(BaseTmxMapLoader.java:656)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadTileSet(BaseTmxMapLoader.java:614)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadTiledMap(BaseTmxMapLoader.java:156)
	at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:81)
```

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] MacOS

mgsx-dev:
There was a regression due to TMXMapLoader refactoring (#5618).

Before refactoring, animated tiles was loaded in 2 passes : the first pass retrieve frames from tileset as static tiles. The second pass replace original static tiles by animated tiles.
This algorithm was removed during refactoring by replacing static tiles on the fly and lead to type cast exceptions.

There might be some more elegant solutions but i think it was safer to bring back previous algorithm. I had to adapt it a little but it's very close.

This change also fix another issue : tile properties (terrain, properties, etc.) was copied over static tiles and lost when replaced by animated tiles.

original code (before refactoring) is here :  <https://github.com/libgdx/libgdx/commit/7f692fc9a53829b44a6aec2f59420babd348fd67#diff-bd72904ff32971cc8f4ce6186e59a869L424>

I also added a test case because the error only happens with 2 animations : when the second one references the first one as an animation frame.


gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -605,15 +605,26 @@ protected void loadTileSet (Element element, FileHandle tmxFile, ImageResolver i
...
 				margin, source, offsetX, offsetY, imageSource, imageWidth, imageHeight, image);
 
+			Array<AnimatedTiledMapTile> animatedTiles = new Array<AnimatedTiledMapTile>();
+			
 			for (Element tileElement : tileElements) {
 				int localtid = tileElement.getIntAttribute("id", 0);
 				TiledMapTile tile = tileSet.getTile(firstgid + localtid);
 				if (tile != null) {
+					AnimatedTiledMapTile animatedTile = createAnimatedTile(tileSet, tile, tileElement, firstgid);
+					if(animatedTile != null){
```
Tom-Ski: Minor formatting issues in the code in BaseTmxMapLoader

Tom-Ski:
Looks good, thanks for the test and comments 👍 Will merge when formatting updated.

mgsx-dev:
yeah, sorry, i forget to run formatter, it should be ok now.

PR #5749, opened 2019-08-07
[scene2d.ui] If flickScroll disabled, only show scrollbar if hover it or if scrolling using mouse

RUSshy:


RUSshy:
Produce better result

pr

NathanSweet:
If flick scroll is disabled, likely the scroll pane is used for a desktop application where it is common to want to always see the scrollbars. If you need this behavior it would need to be a setting, but we should not change the default behavior.

RUSshy:
leaving it all the time visible is fine if it was separate and not hide the content of the table

Maybe an option would be nice, but add another field would bloat the already huge class

Maybe make use of flags like in ImGui would be a nice idea

NathanSweet:
This is Java, bloat is life!

NathanSweet:
Configuring scrollpane using flags is a bigger change than a setting. If you still need this, please add a setting, since as mentioned desktop apps would not like to hide the scrollbars when flick scrolling is disabled. Also note that you can support both clicking on children and flick scrolling. It may be more comfortable to increase the tap square size, so scrolling doesn't start as easily. You can also cancel touch focus for the scrollpane when children are touchDowned.

PR #5750, opened 2019-08-08
[scene2d.ui] Table rounds position and size values when they are set if round is true.

kylemclean:
Currently, calling any method on `Table` that changes its position or size does not round its values even if `round` is true.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Table.java
@@ -956,6 +956,66 @@ public class Table extends WidgetGroup {
 		tablePrefHeight = Math.max(tablePrefHeight + vpadding, tableMinHeight);
 	}
 
+	public void setX (float x) {
+		if (round) {
+			x = Math.round(x);
+		}
+		super.setX(x);
+	}
+
+	public void setY (float y) {
+		if (round) {
+			y = Math.round(y);
+		}
+		super.setY(y);
+	}
+
+	/** Sets the position of the Table's bottom left corner. */
+	public void setPosition (float x, float y) {
+		if (round) {
+			x = Math.round(x);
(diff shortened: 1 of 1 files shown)


NathanSweet:
The `round` setting is intended to be applied when sizing and positioning the table's child actors (the javadoc is poor, I'll improve it). I think it should be the responsibility of whatever is laying out the table to round its size and position, if needed. It seems odd to specify the size and position of an actor but have the actor ignore that and use different values.

I'll close this issue but we can continue discussion here if needed.

kylemclean:
Thanks for clarifying; the behaviour you've described does make sense. However, it presents an issue for actors that do not exist within a Table. 

This behaviour became an issue for me because I have Window objects that are immediate children of the Stage. I attempt to center a Window in a Stage directly by calling
`window.setPosition(stage.getWidth() / 2, stage.getHeight() / 2, Align.center);` Even if you round the x and y arguments in that call, the procedure of aligning by the center will cause the Window to have non-integer x and y values. This results in the Window and its contents appearing blurry. From what I can tell there is no way to round the position without calling `setPosition` again.

I would propose giving the Stage itself the `round` property, which would be useful when it is used solely for UI where floating point positions and sizes are almost always undesirable. For the Stage it would probably be set to false by default so that game worlds that use it and rely on floating point values don't break. There are probably a multitude of other solutions to consider as well.

NathanSweet:
In my apps I am careful to round values to stay on pixels. I don't use the `setPosition` that takes alignment.

It seems like a lot of things would need to take into considering a stage-wide `round` setting. That would make animations less smooth, for animations it's OK to go off pixels since the things are moving. As soon as a group is rotated or scaled, the round setting would no longer make sense. I realize that doesn't usually happen for UIs (except for animations). I'm not sure such a setting would be better than just being careful when sizing and positioning actors.

PR #5753, opened 2019-08-08
Added method to prefill 2d ParticleEmitter particles pool

obigu:
`ParticleEmitter` instantiates `Particle` instances with a lazy approach. When needed it creates a `Particle` instance and stores it in the "pool" (it's an array of max count particles size) and will reuse this instance after that. The problem is that there is currently no mechanism to initialize the pool potentially causing frame drops when lots of particles are displayed for the first time (noticeable even in high end devices). 

This issue has been worked around by some users by setting `minCount` to a high number and executing `update(0)` on the effect but this is only useful with "explosive" types of effects that emit lots of particles on first update.


gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java
```diff
@@ -703,6 +703,20 @@ public void setSpriteMode (SpriteMode spriteMode) {
 		this.spriteMode = spriteMode;
 	}
 
+	/** Fills the pool of particles with new placeholder {@link Particle} instances. Useful to control when allocation occurs and avoid them mid game.
+	 * {@link #setSprites(Array)} must have been set before calling this method */
+	public void preFillParticlesPool () {
+		if (sprites.isEmpty()) return;
```
mgsx-dev: i would throw a GdxRuntimeException as Sprite(null) does.

mgsx-dev:
once merged, it would be nice to update the wiki somewhere around here: https://github.com/libgdx/libgdx/wiki/2D-ParticleEffects#efficiency

Devs might not be aware of this optimization possibility.

obigu:
@NathanSweet Can you please have a look to this one as well? It's in the same spirit as https://github.com/libgdx/libgdx/pull/5773.

obigu:
@NathanSweet Friendly ping :)

NathanSweet:
Very sorry it's taken so long!

PR #5758, opened 2019-08-14
GWT Accelerometer: report availability only if sensor is present

MrStahlfelge:
GWT Accelerometer is reported as present if the web browser supports it even if there is no accelerometer. This is not correct, using Chrome on Desktop reports an available accelerometer.

This commit changes getPeriphalAvailable in a way that accelerometer availability is only returned if real sensor data is present, this matches the behaviour of the other platforms.

Note: At first I wanted to use the onError callback of the sensor, but it was called on my mobile device too. I was not able to find out why, because the callback provides no error message.

Sorry for the diff. GwtInput had CRLF line endings which I was not able to commit. I think this is okay because (most) libGDX sources have LF line endings in general.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
@@ -120,6 +120,10 @@ public class GwtInput implements Input {
 		return this.accelerometer != null ? (float) this.accelerometer.z() : 0;
 	}
 
+	private boolean isAccelerometerPresent() {
+		return getAccelerometerX() != 0 || getAccelerometerY() != 0 || getAccelerometerZ() != 0;
+	}
+
 	@Override
 	public float getGyroscopeX () {
 		// TODO Auto-generated method stub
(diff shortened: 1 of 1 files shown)


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java
```diff
@@ -1,1087 +1,1091 @@
...
+		return this.accelerometer != null ? (float) this.accelerometer.y() : 0;
+	}
+
+	@Override
+	public float getAccelerometerZ () {
+		return this.accelerometer != null ? (float) this.accelerometer.z() : 0;
+	}
+
+	private boolean isAccelerometerPresent() {
+		return getAccelerometerX() != 0 || getAccelerometerY() != 0 || getAccelerometerZ() != 0;
```
MobiDevelop: Isn't there a better way to determine this? What about checking whether the accelerometer "activated" or "hasReading" returns true?

MrStahlfelge: It's indeed what is done here although it doesn't look like it. See GwtAccelerometer:

https://github.com/libgdx/libgdx/blob/24a6c9fb03eb7c8f8de17dcb71238a41a04cf4dc/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtAccelerometer.java#L44

This line returns 0 exactly if the sensor data is not present (then it is null). There's no chance that a real accelerometer returns exactly 0, so we won't have false positives on this.

MobiDevelop: I see, hmm, would be nice if it was more explicit but this is probably fine.

PR #5760, opened 2019-08-17
Label should take the background in LabelStyle to compute the prefSize

### LINKED ISSUE #5736: Label should take the background in LabelStyle to compute the prefSize
As above title, currently it's ok if the LabelStyle.background minSize smaller than the Label size. But when the LabelStyle.background minSize larger than the Label size, it will stretch the background. It is inconvenient if you want to create a Label is inside a large background, you have to create a layout like Table, setBackground to it and add Label.

bachtrong43:
Fix #5736. If the label smaller than the background, the background will stretch down do the label size.


NathanSweet:
This can't be merged, as it has tilemap stuff in it. Will do it manually (and in the other places it is needed).

PR #5766, opened 2019-08-20
Use Pools in Table instead of custom Pool

Dgzt:
Hi!

I don't know Table class why uses custom Pool instead of Pools, but I've update the Table class.

Best regards,
Dgzt


MobiDevelop:
I think that having a custom pool is fine - it's likely more efficient and doesn't require reflection.

Dgzt:
I see. If you think this merge request is useless, just close it.

NathanSweet:
Aye, custom pool is OK.

PR #5767, opened 2019-08-20
Fix FileHandle#map compatibility with GWT

intrigus:
https://github.com/libgdx/libgdx/pull/5259 doesn't work for me.
When compiling gdx-freetype-gwt I had to comment the part that used FileHandle.map() (https://github.com/intrigus/gdx-freetype-gwt/commit/93221646a8f56c37b6a3362558a8392189dec5ad#diff-11ed83df84f87aedd79b6b30b2c85b2aR93)


PokeMMO:
I think you are missing the ByteBuffer/FileChannel imports.

MrStahlfelge:
Yes, they are missing. Apart from that, I am surprised that this was not noticed before.

MrStahlfelge (changes requested):
Add imports so the build succeeds.

intrigus:
@MobiDevelop thanks for fixing it!

PR #5772, opened 2019-08-27
Added cache packed color for later uses.

AndEditor7:
Added cache packed color by storing last color Intbits and compare the current color Intbits and if equals, then use previous cached FloatBits to avoid uses of Float.intBitsToFloat(value) inside col.toFloatBits().

The rundown:
If the color is null, then use "packed float white float bits" in vertex[cpOffset], else convert the color into Int bits (to avoid uses of Float.intBitsToFloat(value) inside NumberUtils) and store in local variable; colIntBits, than compare colIntBits to lastColIntBits, if equals, use cashed float bits ("tmpColFloatBits") from previous runs and put in vertex[cpOffset], else store colIntBits in lastColIntBits and convert colIntBits to cacheFloatBits (for later uses) and put in vertex[cpOffset].

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
@@ -564,13 +564,11 @@ public class MeshBuilder implements MeshPartBuilder {
 				vertex[cpOffset] = Color.WHITE_FLOAT_BITS;
 			} else {
 				final int colIntBits = col.toIntBits();
-				if (colIntBits == lastColIntBits) {
-					vertex[cpOffset] = cacheFloatBits;
-				} else {
+				if (colIntBits != lastColIntBits) {
 					lastColIntBits = colIntBits;
 					cacheFloatBits = NumberUtils.intToFloatColor(colIntBits);
-					vertex[cpOffset] = cacheFloatBits;
 				}
+				vertex[cpOffset] = cacheFloatBits;
 			}
 		}
 
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
@@ -534,6 +534,9 @@ public class MeshBuilder implements MeshPartBuilder {
 	}
 
 	private final Vector3 tmpNormal = new Vector3();
+	
+	private float cacheFloatBits = Color.WHITE_FLOAT_BITS;
+	private int lastColIntBits = Color.WHITE.toIntBits();
 
 	@Override
 	public short vertex (Vector3 pos, Vector3 nor, Color col, Vector2 uv) {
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java
```diff
@@ -557,8 +560,18 @@ public short vertex (Vector3 pos, Vector3 nor, Color col, Vector2 uv) {
 			vertex[colOffset + 2] = col.b;
 			if (colSize > 3) vertex[colOffset + 3] = col.a;
 		} else if (cpOffset > 0) {
-			if (col == null) col = Color.WHITE;
-			vertex[cpOffset] = col.toFloatBits(); // FIXME cache packed color?
+			if (col == null) {
+				vertex[cpOffset] = Color.WHITE_FLOAT_BITS;
+			} else {
+				final int colIntBits = col.toIntBits();
```
mgsx-dev: Maybe you could simplify this if/else block a bit : 
```java
if(colIntBits != lastColIntBits) {
    lastColIntBits = colIntBits;
    cacheFloatBits = NumberUtils.intToFloatColor(colIntBits);
}
vertex[cpOffset] = cacheFloatBits;
```

AndEditor7: Yaa, I noticed it possible after I created pull requests, I just so tired after hours of figure it out most optimize ways to cache packed color after I saw FIXME comment. Really, it looks like they left it for me! after you point it out, I'll simplify the code. 👍

Darkyenus:
Does this solve any particular performance problem? (i.e. do you have any benchmarks that this improves performance?) I'm not sure whether the added branching and memory access will outweigh the benefit of omitting an `intoToFloatColor` call, which will most likely be JIT inlined and optimized into a few simple instructions.

A more optimal version of the `vertex` method would probably just allow to pass the color directly as a `float`.

AndEditor7:
> Does this solve any particular performance problem? (i.e. do you have any benchmarks that this improves performance?) I'm not sure whether the added branching and memory access will outweigh the benefit of omitting an `intoToFloatColor` call, which will most likely be JIT inlined and optimized into a few simple instructions.
> 
> A more optimal version of the `vertex` method would probably just allow to pass the color directly as a `float`.

I don't know much about hardware like memory access speed, how to do a benchmark within LibGDX and the source code inside Float.intBitsToFloat(value) because it uses JIT, but let say; The Meshbuilder used by ModelBuilder class if you build a shape sphere with a lot of vertexes with the same color constantly converting intoToFloatColor but cache it can reduce it. but I'm still not sure what memory access speed is and how to benchmark within LibGDX for an accurate result.

AndEditor7:
I just did multiple benchmarks (I figure out) after benchmarks, after benchmarks, after benchmarks. result of approximate with cache: 223ms and without cache: 212ms. instead of performance what I expect but it's deterioration. *disappointed*

PR #5773, opened 2019-08-27
Add option to pre-fill Pool with objects

obigu:
Pool works on-demand and will be empty until objects are obtained and then fred.
In many cases it's preferable to pre-fill the Pool for performance reasons, specially to avoid allocation to occur during gameplay.

PR adds a new constructor to `Pool` to allow pre-filling it with the initial capacity.

Even if it's possible to do it manually by obtaining and freeing objects after the creation of the Pool I think it's a useful and common enough use case to be added to Pool. Also, doing it in the constructor is more efficient as it avoids a call to `reset()` on all pre-filled objects.


cypherdare:
In my opinion, this would be more useful as a method you can call on a pool to seed it up to a certain number. Then you could call it on pools that you get from Pools. Or if you use lots of the actions from Actions, you could seed the pools you know it's going to use. Or if you instantiate pools before OpenGL has been spun up, you can safely seed them after `create()` has been called (the reason to do this comes up more often in Kotlin, where you want to avoid nullable variables).

obigu:
@cypherdare Definitely adding a method would allow having this feature in cases where you don't have control over the Pool instantiation. In any case, even if we add it as a method, I think it's really useful having it in the constructor as well because I think it's a really common use case to want to prefill the pool with the intial capacity (in mobile apps where this is critical). @NathanSweet What do you think?

NathanSweet:
I've not had a need for this, but it looks OK. `Pool` javadocs aren't formatted -- can you run the formatter? Would you like to add a `fill` method?

obigu:
@NathanSweet Cool, I'll add a fill method and reformat

obigu:
@NathanSweet Should be good to merge. In addition to javadoc there was some other stuff that wasn't properly formatted

NathanSweet:
Sorry for the delay, had a little vacation!

Seems the javadocs are formatted differently (eg `/**` and `*/` are no longer on the same line), otherwise looks good.

obigu:
@NathanSweet Yes, I applied the official libGDX formatter to the whole class and some javadocs were changed as they were not formatted following the code style (eg /** and */ are not on the same line). Do you want me to reformat them as they were before applying the formatter?

obigu:
@NathanSweet Please let me know what to do with the formatting so we can merge this. The current changes (including javadocs) are result of applying the official libGDX formatter.

NathanSweet:
Sorry again for being ridiculously late.

It seems IDEA is not able to apply the Eclipse formatting rules like Eclipse does. I don't use IDEA, so I'd prefer to leave figuring out how to get IDEA to match Eclipse to someone familiar with IDEA.

PR #5775, opened 2019-08-28
Better error message when fetching Uniform from uncompiled shader.

AndEditor7:
Better error message in ShaderProgam class.

When programming shaders, I use visual code studio to program shaders. The issue with visual code studio it doesn't have OpenGL ES error checking or "Code language not supported or defined". When saving the shader file to asset folder with a typo or syntax error, then run the game and a misleading exception "no uniform with the name" show on console log and close the game - this creates confusing from programmer checking the variable is correct without realizing there is an error when compiling the shader, it happens to me once and with a new exception would prevent from happening who using a text editor like visual code studio.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
@@ -305,7 +305,10 @@ public class ShaderProgram implements Disposable {
 		int location;
 		if ((location = uniforms.get(name, -2)) == -2) {
 			location = gl.glGetUniformLocation(program, name);
-			if (location == -1 && pedantic) throw new IllegalArgumentException("no uniform with name '" + name + "' in shader");
+			if (location == -1 && pedantic) {
+				if (isCompiled) throw new IllegalArgumentException("no uniform with name '" + name + "' in shader");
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
```diff
@@ -305,7 +306,10 @@ public int fetchUniformLocation (String name, boolean pedantic) {
 		int location;
 		if ((location = uniforms.get(name, -2)) == -2) {
 			location = gl.glGetUniformLocation(program, name);
-			if (location == -1 && pedantic) throw new IllegalArgumentException("no uniform with name '" + name + "' in shader");
+			if (location == -1 && pedantic) {
+				if (isCompiled) throw new IllegalArgumentException("no uniform with name '" + name + "' in shader");
+				else throw new GdxRuntimeException("An attempted fetch uniform from uncompiled shader \n" + getLog());
```
mgsx-dev: maybe IllegalStateException would fit better.

Tom-Ski:
You should use a real shader editor & compiler. PVRShaderEditor is good

AndEditor7:
> You should use a real shader editor & compiler. PVRShaderEditor is good

You are a life saver

PR #5777, opened 2019-08-29
Minor changes to MeshBuilder and Color and interfaces.

AndEditor7:
Added extended Disposable in ImmediateModeRenderer and Screen interface; for reason. - I found out how useful the Disposable interface is. 

Added missing Z-Axis in the comment. And removed 2 unused imports and FIXME comment, for FIXME comment "cache packed color?" - no, these are no efficient way to cache packed color in MeshBuilder, according to from my closed pull request (the second and last reply): #5772. And also a minor change in the vertex method.

Minor changes to Color class for readability.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Screen.java
+++ b/gdx/src/com/badlogic/gdx/Screen.java
@@ -16,8 +16,6 @@
 
 package com.badlogic.gdx;
 
-import com.badlogic.gdx.utils.Disposable;
-
 /** <p>
  * Represents one of many application screens, such as a main menu, a settings menu, the game screen and so on.
  * </p>
@@ -25,7 +23,7 @@ import com.badlogic.gdx.utils.Disposable;
  * Note that {@link #dispose()} is not called automatically.
  * </p>
  * @see Game */
-public interface Screen extends Disposable {
+public interface Screen {
 	
 	/** Called when this screen becomes the current screen for a {@link Game}. */
 	public void show ();
@@ -47,6 +45,5 @@ public interface Screen extends Disposable {
 	public void hide ();
 
 	/** Called when this screen should release all resources. */
-	@Override
 	public void dispose ();
 }
--- a/gdx/src/com/badlogic/gdx/Screen.java
+++ b/gdx/src/com/badlogic/gdx/Screen.java
@@ -16,6 +16,8 @@
 
 package com.badlogic.gdx;
 
+import com.badlogic.gdx.utils.Disposable;
+
 /** <p>
  * Represents one of many application screens, such as a main menu, a settings menu, the game screen and so on.
  * </p>
@@ -23,7 +25,7 @@ package com.badlogic.gdx;
  * Note that {@link #dispose()} is not called automatically.
  * </p>
  * @see Game */
-public interface Screen {
+public interface Screen extends Disposable {
 	
 	/** Called when this screen becomes the current screen for a {@link Game}. */
 	public void show ();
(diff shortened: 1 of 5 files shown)


cypherdare:
In my opinion, interfaces should not extend Disposable. Disposable indicates that something **must** be disposed because it holds native memory. That's never a guarantee with an interface.

cypherdare:
The Color change can't be done because alpha can only use 7 bits due to the negative bit.

AndEditor7:
The first of, the reason I put Disposable in both ImmediateModeRenderer and Screen interfaces because if have Disposable method/function, should be Disposable; simply put, I carefully make the decision and check LibGDX's interfaces witch does have some extended Disposable interface, so I included in both ImmediateModeRenderer and Screen interfaces. But, the **real** reason I added extend Disposable so I can able to use my utility function here: 
```
public void dispose () {
	Util.disposes(batch, stage, asset);
}

/** Creates an array list of disposable objects and release all resources - null safe. */
public static void disposes(Disposable... obj) {
	for (int i = 0, s = obj.length; i < s; i++)
		if (obj[i] != null) obj[i].dispose();
}
```
I know it's greedy, but lOOK how many linES I rEdUsED, IT'S BEAUTIFUL!!!!

The color. I'm not sure what you mean by "the alpha can only use 7 bits due to the negative bit" - I simply added toIntBits() method instead create a local variable - could you give me more explanation and what lines of code you meant?

gdx/src/com/badlogic/gdx/Screen.java
```diff
@@ -16,14 +16,16 @@
...
+
 /** <p>
  * Represents one of many application screens, such as a main menu, a settings menu, the game screen and so on.
  * </p>
  * <p>
  * Note that {@link #dispose()} is not called automatically.
  * </p>
  * @see Game */
-public interface Screen {
+public interface Screen extends Disposable {
```
NathanSweet: I think not all Screens must be Disposable. Also this will break every existing code which uses Screen and doesn't implement Disposable.

gdx/src/com/badlogic/gdx/Screen.java
```diff
@@ -45,5 +47,6 @@
 	public void hide ();
 
 	/** Called when this screen should release all resources. */
+	@Override
```
NathanSweet: Please don't add `@Override` to files that don't already use it.

gdx/src/com/badlogic/gdx/graphics/glutils/ImmediateModeRenderer.java
```diff
@@ -18,8 +18,9 @@
 
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.math.Matrix4;
+import com.badlogic.gdx.utils.Disposable;
 
-public interface ImmediateModeRenderer {
+public interface ImmediateModeRenderer extends Disposable {
```
NathanSweet: Unlike Screen, this is probably OK. I doubt much existing code implements this interface.

gdx/src/com/badlogic/gdx/Screen.java
AndEditor7: It has been restored back it was. 👍

gdx/src/com/badlogic/gdx/graphics/glutils/ImmediateModeRenderer.java
AndEditor7: Even it was, programmer (That I use) should able to fix if even a case.

cypherdare:
Never mind about the Color. I misread a line of your code.

PR #5780, opened 2019-08-30
Have Flame store 3d particle texture atlas into save file

ryanastout:
make it so that in the 3d particle effect editor (Flame), if a texture region

is used, the name of the sub-image (region) is stored. This makes it so that if a texture atlas is updated, the uv-coordinates of the region can be read from the atlas rather than having the uv coordinates stored in the particle effect save file.


ryanastout:
(sorry, didn't mean to close the request)

PR #5784, opened 2019-09-03
Feature policy implementation for gwt

SimonIT:
With the feature policy, developers can allow and disallow some web browser features in iframes. If the game wants to access a disallowed feature, an error is thrown. This happens for example on itch.io. See for example the game of @MrStahlfelge for the game jam https://mrstahlfelge.itch.io/gdxjam1908. Make sure to use chrome. There will be an error in the console ```Uncaught (in promise) DOMException: Failed to construct 'Accelerometer': Access to sensor features is disallowed by feature policy```
This PR adds a check whether the feature, in this case accelerometer, is allowed. If not isPeripheralAvailable returns false and the setupAccelerometer method wont get called.

https://developers.google.com/web/updates/2018/06/feature-policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy


MrStahlfelge:
Hi, can you put the html container site you tested with here? Makes it easier for me and others to test. Thanks. :)

SimonIT:
Here's a live version https://libgdx2.v22018086706171227.nicesrv.de/
```<iframe src="https://libgdx2.v22018086706171227.nicesrv.de/"></iframe>```
You can add also some allow attributes for example ```<iframe src="https://libgdx2.v22018086706171227.nicesrv.de/" allow="accelerometer;"></iframe>```
For a version without the fix remove the 2 behind libgdx

Issue #5755, opened 2019-08-11
Drawing example image crashes on Android 2.3.3


yerAWeeGit:
Please ensure you have given all the following requested information in your report.

#### Issue details
Pretty much the title. With the example script `gdx-setup.jar` produces, if `batch.draw(img, 0, 0);` is commented out, the app does not crash.

#### Reproduction steps/code
Import the example produced by `gdx-setup.jar` into Android Studio and run the app on a device with Android Froyo or Gingerbread. The issue is specifically occurring on the GT-S5830 with the latest stock ROM (rooted but otherwise unmodified). This issue does not occur using the same device with Android 4.2.2 (custom ROM).

#### Version of LibGDX and/or relevant dependencies
- LibGDX 1.9.10
- Android Studio 3.4.2
- Gradle 5.4.1
- SDK 29.0.2

#### Stacktrace
```java
09-09 15:19:55.649 2301-2309/my.app E/AndroidRuntime: FATAL EXCEPTION: GLThread 10
    java.lang.UnsupportedOperationException: glBlendFuncSeparate
        at android.opengl.GLES20.glBlendFuncSeparate(Native Method)
        at com.badlogic.gdx.backends.android.AndroidGL20.glBlendFuncSeparate(AndroidGL20.java:76)
        at com.badlogic.gdx.graphics.g2d.SpriteBatch.flush(SpriteBatch.java:970)
        at com.badlogic.gdx.graphics.g2d.SpriteBatch.end(SpriteBatch.java:185)
        at my.app.MyApp.render(MyApp.java:27)
        at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame(AndroidGraphics.java:494)
        at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1388)
        at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1138)
09-09 15:19:59.849 2301-2301/my.app E/AndroidGraphics: waiting for pause synchronization took too long; assuming deadlock and killing
```

#### Please select the affected platforms
- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

intrigus:
Is this a new issue, i.e. did this work in a prior gdx version?

Anuken:
[...]

Issue #3335, opened 2015-08-08
Immersive Mode not working after onWindowFocusChanged()


ConteZero:
In AndroidApplication.java:

``` java
    protected void hideStatusBar (boolean hide) {
        if (!hide || getVersion() < 11) return;

        View rootView = getWindow().getDecorView();

        try {
            Method m = View.class.getMethod("setSystemUiVisibility", int.class);
            if (getVersion() <= 13) m.invoke(rootView, 0x0);
            m.invoke(rootView, 0x1);
        } catch (Exception e) {
            log("AndroidApplication", "Can't hide status bar", e);
        }
    }

    @Override
    public void onWindowFocusChanged (boolean hasFocus) {
        super.onWindowFocusChanged(hasFocus);
        useImmersiveMode(this.useImmersiveMode);
        hideStatusBar(this.hideStatusBar);
        if (hasFocus) {
            this.wasFocusChanged = 1;
            if (this.isWaitingForAudio) {
                this.audio.resume();
                this.isWaitingForAudio = false;
            }
        } else {
            this.wasFocusChanged = 0;
        }
    }

    @TargetApi(19)
    @Override
    public void useImmersiveMode (boolean use) {
        if (!use || getVersion() < Build.VERSION_CODES.KITKAT) return;

        View view = getWindow().getDecorView();
        try {
            Method m = View.class.getMethod("setSystemUiVisibility", int.class);
            int code = View.SYSTEM_UI_FLAG_LAYOUT_STABLE | View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
                | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN | View.SYSTEM_UI_FLAG_HIDE_NAVIGATION | View.SYSTEM_UI_FLAG_FULLSCREEN
                | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY;
            m.invoke(view, code);
        } catch (Exception e) {
            log("AndroidApplication", "Can't set immersive mode", e);
        }
    }
```

Method onWindowFocusChanged() contains a bug, useImmersiveMode() should be executed after hideStatusBar() and not before.
As it is now, hideStatusBar() overwrites the changes to SystemUiVisibility done by useImmersiveMode(), and this makes Immersive Mode not working after a call to onWindowFocusChanged()

obigu:
[...]

PR #5787, opened 2019-09-07
Remove unnecessary reflection usage in android backend

### LINKED ISSUE #5785: New R8 obfuscator with android
I tried releasing my game on android with enabled minifying. It did work with proguard, but the new R8 obfuscator did not work. I used the default proguard config file. 
I tried this https://badlogicgames.com/forum/viewtopic.php?t=29056 solution but it didn't work

#### Reproduction steps/code
Set `minifyEnabled true` in android/build.gradle

Android Gradle plugin 3.4.0 and higher, in my case 3.4.2

#### Stacktrace
```
Couldn't construct AndroidInput, this should never happen
```

- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

SimonIT:
This fixes #5785 because the new R8 obfuscator removes AndroidInput classes which were instantiated with reflection.


MobiDevelop:
Just to confirm, you're testing this running on API 9?

SimonIT:
The problem is that I don't got any old device with api level 9. I tried using the emulator, but it doesn't start. I also tried an api level 10 emulator, but it also doesn't start. So I'm not sure how I can test it. All emulators until api level 14 are deprecated

MobiDevelop:
We can't accept this until it has been tested on API 9, or we increase the minimum sdk version. We can't just assume it is safe.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
```diff
@@ -214,18 +192,14 @@ protected void createWakeLock (boolean use) {
...
 		View rootView = getWindow().getDecorView();
 
-		try {
-			Method m = View.class.getMethod("setSystemUiVisibility", int.class);
-			if (getVersion() <= 13) m.invoke(rootView, 0x0);
-			m.invoke(rootView, 0x1);
-		} catch (Exception e) {
-			log("AndroidApplication", "Can't hide status bar", e);
-		}
+		if (getVersion() <= 13) rootView.setSystemUiVisibility(0x0);
```
intrigus: Does anybody know why it is this way?
Isn't the uiVisibility always changed to 0x1 the line below?

Anuken:
I just want to point out that [libGDX doesn't work on API 9 anyway](https://github.com/libgdx/libgdx/issues/5755), so the best solution would be to change the minimum API level to 14 (as discussed).

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
obigu: No idea, I think many things could be cleaned up because the whole immersive mode / status bar code is a bit of a mess right now. Have a look at https://github.com/libgdx/libgdx/issues/3335 and https://github.com/libgdx/libgdx/pull/5329.

noblemaster:
I think we agree to move the API level to 14. Let's do that with this pull request also.

SimonIT:
Is there any other location to increase the minSdk? I can't find more

noblemaster:
Thanks! I think you got all the minSDK locations. I merged it for better or worse. :)

This also addresses https://github.com/libgdx/libgdx/issues/5755

SimonIT:
I think me can also remove some proguard rules

PR #5792, opened 2019-09-13
fix #5791 Fixing `AtlasRegion.name` documentation

### LINKED ISSUE #5791: Incorrect JavaDoc for `TextureAtlas.name`
The documentation currently says `The name of the original image file, up to the first underscore. Underscores denote special instructions to the texture packer.`.

However, for example, if you pack an image called `bob_walking.png`, the returned String will be `bob_walking`.

payne911:
The documentation currently says `The name of the original image file, up to the first underscore. Underscores denote special instructions to the texture packer.`.

However, for example, if you pack an image called `bob_walking.png`, the returned String will be `bob_walking`.

The JavaDoc is now accurate as of the content of that variable.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/TextureAtlas.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/TextureAtlas.java
@@ -473,8 +473,9 @@ public class TextureAtlas implements Disposable {
 		 * @see TextureAtlas#findRegions(String) */
 		public int index;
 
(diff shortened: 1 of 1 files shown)


payne911:
I'm, however, unsure if I did everything I was supposed to to get this PR accepted.
Where do I log in the CHANGES (https://github.com/libgdx/libgdx/blob/master/CHANGES) file?
And what about the license-related signature file: am I actually supposed to sign that to submit a PR?

PR #5812, opened 2019-10-06
@Nullable annotations for scene2d classes

maltaisn:
This pull request fixes what was discussed in #5615.
- `@NotNull` annotations were completely removed. While I think this is wrong, it's not my project. IntelliJ can infer `@NotNull` annotations in most cases so drawbacks are limited.
- `org.jetbrains.annotations.Nullable` was replaced with `com.badlogic.gdx.annotation.Nullable` to avoid the dependency.
- Squashed and rebased past commits onto master.

To configure nullibility annotations on your IDE and get warnings you'll have to add `com.badlogic.gdx.annotation.Nullable` [in IntelliJ IDEA](https://www.jetbrains.com/help/idea/nullable-notnull-configuration.html) or [in Eclipse](https://help.eclipse.org/oxygen/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm).

If this is merged I can continue adding them for the rest of LibGDX.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/FlameMain.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/FlameMain.java
@@ -112,6 +112,8 @@ import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.StreamUtils;
 import com.badlogic.gdx.utils.StringBuilder;
 
+import org.jetbrains.annotations.NotNull;
+
 /** @author Inferno */
 public class FlameMain extends JFrame implements AssetErrorListener {
 	public static final String 	DEFAULT_FONT = "default.fnt",
@@ -718,7 +720,7 @@ public class FlameMain extends JFrame implements AssetErrorListener {
 			playPauseButton = new TextButton("Pause", skin);
 			playPauseButton.addListener(new ClickListener(){
 				@Override
-				public void clicked (InputEvent event, float x, float y) {
+				public void clicked (@NotNull InputEvent event, float x, float y) {
 					isUpdate = !isUpdate;
 					playPauseButton.setText(isUpdate ? "Pause" : "Play");
 				}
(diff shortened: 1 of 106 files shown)


NathanSweet:
I finally had some time to look at this. As mentioned previously, I like the idea of using annotations instead of javadoc, but it is important that it is not annoying. Enabling null analysis with use of only the Nullable annotation works great within libgdx's projects. For projects that depend on libgdx, there are some definite annoyances.

It is relatively common for application code to assume something is not null. For example, if I need to grab a TextButtonStyle and set the padding on the up, down, over, focused, checked, checkedOver, checkedFocused, and disabled drawables I have to 1) null check all those, or 2) suffer a warning, or 3) disable the warning, which defeats the whole purpose. Here are more examples from the hundreds of warnings across my many projects:
```
scrollPaneStyle.background = skin.newDrawable("bg-panel-dark");
scrollPaneStyle.background.setTopHeight(5); // background could be null

TextButtonStyle style = new TextButtonStyle(skin.get(TextButtonStyle.class));
style.up.setTopHeight(2); // up could be null

// getActor() could be null
menu.add("Label").getActor().setAlignment(Align.center);
buttons.add(new TextButton("Win", skin)).getActor().addListener(new ClickListener() {

label.getParent().toFront(); // getParent could be null

Condition condition = (Condition)button.getUserObject();
box.flashIcon(condition.name()); // condition could be null because getUserObject could be

Dialog dialog = new Dialog("Title", skin) {
	protected void result (Object object) {
		if (!(Boolean)object) return; // object could be null
```
In all these cases the application knows the value cannot be null. Needing to check it solely for the compiler is a pain and gets in the way quite a lot. Suffering the warnings is out because I'm a believer in having an empty problems view -- anything that shows up there is something that should be fixed, otherwise the warning should be disabled.

The only way I found to make it not annoying is to set the problem level to `info` for null analysis, then have my IDE only show problems that are warning or higher. This way the potential nulls get underlined in my code, but I never have to jump through stupid hoops. With this approach I don't think asserts or warning suppression should be used.

I think it would be better to name the annotation `Null`, just so it is shorter. IMHO that doesn't look bad and I'm not at all concerned that it differs from other codebases which use `Nullable`.

I'll go ahead and add some changes on top of your branch.

NathanSweet:
It seems no amount of @NonNullByDefault and @NonNull will get Eclipse to show a warning/info when passing a null parameter. I hope to use @NonNullByDefault on the package (in a package-info.java), but it seems like it doesn't work.

NathanSweet:
Cheers!

PR #5815, opened 2019-10-14
LWJGL3: Window positioning fixes.

PokeMMO:
* Fixes #5845 on Windows.
* Fix position for maximized window when specifying maximizedMonitor.
* Center default window on primary monitor's working area.

PR #5816, opened 2019-10-14
LWJGL3: Center default window on primary monitor.

PokeMMO:
We check the resolution of primary monitor, but do not account for primary monitor's relative position.

If primary is to the right or below another monitor the position would have been incorrect.


thechubbypanda:
I was told to keep the original functionality by @Tom-Ski [here](https://github.com/libgdx/libgdx/pull/5619#issuecomment-496000657).

PokeMMO:
It looks like the current behavior is actually broken. It's checking primary monitor's video mode to center the window, but not placing it on primary monitor.

PokeMMO:
Closed in favor of #5815

PR #5817, opened 2019-10-14
Adding a getter.

payne911:
Hard to justify having a setter, and no getter. Open to discussion.


intrigus:
Did anybody wish for such a getter when developing games with libGDX?
In all the time I've been using libGDX I (personally) never needed such a getter

Tom-Ski:
Nope

payne911:
I was requiring it to go on with the way I was designing my custom widget for libgdx, but I realized it's just that I had the wrong approach to the construction of that widget.
In discord, the general feedback seemed to be that a getter shouldn't be required, but I thought I'd bring the discussion out here as well. Some did mention they thought it wouldn't hurt.

Tom-Ski:
As per comments, and OPs further statement, don't see need for this.

Issue #5818, opened 2019-10-21
MockGraphics.getFramesPerSecond should return fps


SebastianSchlegel:
#### Issue details
* MockGraphics.getFramesPerSecond always returns 0 even though fps is calculated correctly

#### Reproduction steps/code
* Create a HeadlessApplication
* Run it
* call Gdx.graphics.getFramesPerSecond()
* Expected result: current fps
* Actual result: 0

#### Version of LibGDX and/or relevant dependencies
gdxVersion: 1.9.9 (but I checked the latest sources)

#### Stacktrace
N/A

#### Please select the affected platforms
* Headless

EtK2000:
If you created a pull request close this.

intrigus:
Please close as your PR has been accepted.

PR #5819, opened 2019-10-21
MockGraphics.getFramesPerSecond should return fps #5818

SebastianSchlegel:
I've just read that you don't need an issue for pull requests. Too late, sorry for that. Anyway I hope this code change fixes the problem described in #5818


intrigus:
Curious, when do you need `getFramesPerSecond` to not return zero?

SebastianSchlegel:
> 
> 
> Curious, when do you need `getFramesPerSecond` to not return zero?

To see how my server is doing.  Actually the `Application.mainloop` is responsible for the update rate but afaik only `Graphics `keeps track of how fast it's really going. Furthermore, it would be possible to use the same code on the client and server side the check the performance.

PR #5821, opened 2019-10-24
Access to  BitmapFont for extends It on RTL language support

scorpion16:
For supporting another languages (like RTL language) we need to customize the BitmapFont... So,for doing this approach I need to change (override) some methods on BitmapFont that generated glyphs using FreeTypeFontGenerator.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -166,8 +166,9 @@ public class FreeTypeFontGenerator implements Disposable {
 		return font;
 	}
 	
+	/** Create a new {@link BitmapFont} instance for inject customized {@link BitmapFont} (ex. support RTL fonts). */
 	protected BitmapFont newBitmapFont(BitmapFont.BitmapFontData data, Array<TextureRegion> pageRegions, boolean integer) {
-		return new BitmapFont(data, pageRegions, true);
+		return new BitmapFont(data, pageRegions, integer);
 	}
 	
 	/** Uses ascender and descender of font to calculate real height that makes all glyphs to fit in given pixel size. Source:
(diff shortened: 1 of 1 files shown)


extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -161,11 +161,15 @@ public BitmapFont generateFont (FreeTypeFontParameter parameter, FreeTypeBitmapF
...
 		if (data.regions.isEmpty()) throw new GdxRuntimeException("Unable to create a font with no texture regions.");
-		BitmapFont font = new BitmapFont(data, data.regions, true);
+		BitmapFont font = newBitmapFont(data, data.regions, true);
 		font.setOwnsTexture(parameter.packer == null);
 		return font;
 	}
-
+	
+	protected BitmapFont newBitmapFont(BitmapFont.BitmapFontData data, Array<TextureRegion> pageRegions, boolean integer) {
+		return new BitmapFont(data, pageRegions, true);
```
mgsx-dev: it should use integer parameter instead of true i think

shapcode: Also, how about a better name or at least javadoc on that `integer` parameter.

MobiDevelop: Also, how about a better name or at least javadoc on that `integer` parameter.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -161,11 +161,16 @@ public BitmapFont generateFont (FreeTypeFontParameter parameter, FreeTypeBitmapF
...
 		if (data.regions.isEmpty()) throw new GdxRuntimeException("Unable to create a font with no texture regions.");
-		BitmapFont font = new BitmapFont(data, data.regions, true);
+		BitmapFont font = newBitmapFont(data, data.regions, true);
 		font.setOwnsTexture(parameter.packer == null);
 		return font;
 	}
-
+	
+	/** Create a new {@link BitmapFont} instance for inject customized {@link BitmapFont} (ex. support RTL fonts). */
+	protected BitmapFont newBitmapFont(BitmapFont.BitmapFontData data, Array<TextureRegion> pageRegions, boolean integer) {
```
intrigus: What is the boolean `integer` for?
Maybe add a reference to https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java#L151 ?

scorpion16: It's for to use integer charachter like unicode or utf-8 encoding. i think.

noblemaster: Yes, just copy the description from the constructor, so others know what it all means! :)

Issue #5487, opened 2018-12-19
[Dicussion] An argument for upgrading to Java 8


Anuken:
Now, I know there's been a few issues and PRs requesting Java 7 support posted here already. Nearly all of them have been closed for various reasons, mostly related to Android and GWT compatibility. Regardless, I would like to discuss the possibility of upgrading source compatibility to Java 8 across the entire codebase.

### Why?
Java 7 and 8 come with some language features that, in my opinion, could greatly improve the codebase:
- Diamond operators could be used to reduce generic declaration redundancy
- Default methods could be used in classes like `ApplicationListener` to automatically provide `ApplicationAdapter` functionality out of the box without creating a separate abstract class
- Static interface methods could improve code organization
- Listener utility interfaces could be created and used in Scene2D in conjunction with lambdas to provide much cleaner listener support, e.g. `button.clicked(e -> System.out.println("Clicked!"))`
- Lambdas could be used instead of anonymous classes across the codebase, e.g. when posting `Runnable`s to an application or using `gdx.utils.Predicate` in `Array`s

### But isn't Java 8 unsupported?
No, Java 8 language features are currently supported across all libGDX backends. This includes [GWT](https://stackoverflow.com/questions/15693169/java-8-support-in-gwt), [Android](https://developer.android.com/studio/write/java8-support), Desktop and [RoboVM](https://github.com/MobiVM/robovm/issues/259).

### What about minimum Android API levels? Will they need to be increased?
No. According to the [Android documentation](https://developer.android.com/studio/write/java8-support), default interfaces, static interface methods and lambdas work at _any minimum API_ level, even across third party libraries.

### But the new Java 7/8 API additions don't work on Android/GWT/RoboVM!
This is mostly true. `java.util.stream`, `java.util.function` and other subpackages and classes are not supported at the current libGDX minimum Android API level. _However_, they don't need to be. Many, many packages and classes from Java 6 are _already_ unsupported in Android, and these new 7/8 API additions are no different. The solution is to not use any new API from Java 7 or 8. I am proposing this upgrade for the sake of the **new language features**, not new API.

What this means is that if you want your game to run on older phones, you don't use any new Java 7/8 API, just like you don't use `java.lang.reflection.*` on GWT or `java.awt.geom.*` on Android. This is not something that should be a problem. Most modern IDEs warn users when their code uses classes that are not compatible with their minimum API level.

### Has Java 8 been tested on these backends to make sure it works?
[...]

PR #5829, opened 2019-10-31
Java 8

thechubbypanda:
Moves the Maven and Gradle build process and source version to 1.8.
Fixes the no longer working ```mvn install``` since Maven no longer supports Java version 6 and below.

### DIFF
--- a/backends/build.gradle
+++ b/backends/build.gradle
@@ -17,8 +17,8 @@
 subprojects {
     apply plugin: "java"
 
-    sourceCompatibility = 1.6
-    targetCompatibility = 1.6
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
     sourceSets.main.java.srcDirs = ["src"]
     sourceSets.main.resources.srcDirs = ["src"]
 
--- a/extensions/build.gradle
+++ b/extensions/build.gradle
@@ -17,8 +17,8 @@
 subprojects {
     apply plugin: "java"
 
-    sourceCompatibility = 1.6
-    targetCompatibility = 1.6
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
 
     compileJava.options.encoding = 'UTF-8'
 
--- a/gdx/build.gradle
+++ b/gdx/build.gradle
@@ -16,8 +16,8 @@
 
 apply plugin: "java"
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
 
 sourceSets.main.java.srcDirs = ["src"]
 sourceSets.main.resources.srcDirs = ["src"]
--- a/pom.xml
+++ b/pom.xml
@@ -78,8 +78,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
+          <target>1.8</target>
           <fork>true</fork>
           <showWarnings>true</showWarnings>
         </configuration>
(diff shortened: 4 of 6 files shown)


PokeMMO:
The biggest problem with enabling java 8 is android. Only API 24+ supports most java 8 features, but some are supported on lower API levels. Simply raising the java version would allow people to accidentally use unsupported features without proper IDE warnings.

See https://developer.android.com/studio/write/java8-support for full details.

Anuken:
> Simply raising the java version would allow people to accidentally use unsupported features without proper IDE warnings.

This is not true. IntelliJ (and Android Studio, presumably) do not care about your supported language level, they care about your minimum API level. Here's what this looks like in my Java 8 project @ min API level 14:

image

PokeMMO:
Last time I checked shared projects (the non-android specific parts) are not aware of SDK levels.

Anuken:
This screenshot was taken from code in the `core` project.

thechubbypanda:
Maven no longer supports java 6, so this kinda has to happen.
At worst we can make it java 7, I just assumed it would be a safer bet to go with the LTS version.

noblemaster:
We shouldn't do that. 

> Android Studio 3.0 and later supports all Java 7 language features and a subset of Java 8 language features that vary by platform version.

We could probably go to Java 7. 

> Fixes the no longer working mvn install since Maven no longer supports Java version 6 and below.

Isn't that just he JDK maven is running on? There shouldn't be a problem keeping 1.6 as the source version and running JDK 8 for Maven? Maybe I am wrong there?

intrigus:
What exactly does this fix?
I just cloned the latest master and ran `mvn install` and it worked fine.

````
mvn --version
Apache Maven 3.3.9
Maven home: /usr/share/maven
Java version: 11.0.2, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-11-oracle
Default locale: de_DE, platform encoding: UTF-8
OS name: "linux", version: "4.9.0-8-amd64", arch: "amd64", family: "unix"

**Comment by SimonIT** [comment]
Gradle says currently (for example with `gradlew tests:gdx-tests-gwt:superDev`):
```
warning: [options] source value 6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
```

PokeMMO:
Is there a way to test SDK compatibility from Gradle/CLI? Gradle's lint task doesn't seem to apply to core project.

thechubbypanda:
I used maven 3.6.2 and java version 13...
I apologise since there is no major problem at the moment

> Isn't that just he JDK maven is running on? There shouldn't be a problem keeping 1.6 as the source version and running JDK 8 for Maven? Maybe I am wrong there?

You are correct, running with java 11 incurs no issues.

SimonIT:
https://maven.apache.org/docs/history.html Maybe the information helps, that maven requires since the 2015-03-18 / 3.3.1 java 7

PokeMMO:
Android Gradle plugin 4.0.0 and higher now supports desugaring most of the java 8 core library. With this I believe there's no real reason to put off migration to java 8.

https://developer.android.com/studio/write/java8-support

noblemaster:
> **most** 

It should be 100% everything without fail. 

What about RoboVM and Java 8. Is that relevant:
https://github.com/MobiVM/robovm/issues/259

PokeMMO:
>It should be 100% everything without fail.

I only said most because I noticed it said `A subset of java.time`. I'm not actually sure what is missing. I also posted this in the wrong ticket. I meant to post this in #5487 

>What about RoboVM and Java 8.

I was under the impression RoboVM fully supported java8, but it appears only the features. This makes my argument pointless.

noblemaster:
Given the possible problems, Let's defer this for now. There is still a request for Java 7 which might make sense to do first if needed. See: https://github.com/libgdx/libgdx/pull/5839

PR #5832, opened 2019-11-02
Add a LongQueue variant

WickedShell:
This adds a LongQueue that works much the same way as the existing Queue class but is specialized to the primitive type. This was done by copying and editing down from the existing Queue class and test.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/LongQueue.java
+++ b/gdx/src/com/badlogic/gdx/utils/LongQueue.java
@@ -1,0 +1,356 @@
+/*******************************************************************************
+ * Copyright 2015 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


intrigus:
Did you try this with GWT?

SimonIT:
The travis build doesn't fail, so it should work with gwt

WickedShell:
I haven't tried it with GWT as I've never done anything with GWT. I did actually provide a test (which definetly works when I run the build task in gradle, but I'm not sure if CI runs those tests within GWT at all.

SimonIT:
The CI runs your test and compiles the test project. The test project contains the libgdx core (obviously), so if there would be a problem with gwt, it would fail

Tom-Ski:
Was just missing cache entry

WickedShell:
@Tom-Ski Thanks! For my future reference what is the point of that xml you added it to, and what should go in?

Tom-Ski:
It defines the classes that are available to gwt for the core libgdx module. Every class that you want to be able to use in gwt backend needs to be put there, or excluded if its not gwt compatible.

PR #5836, opened 2019-11-06
Android music looping fix

wheelergames:
In an attempt to make a fix to the looping issue found often on Android devices when playing Music, I have lifted the answer from https://stackoverflow.com/questions/21247670/android-mediaplayer-loop-has-gaps-even-with-ogg-format/33274466#33274466 which I have used in a previous game of mine and works well, and it is now working fine on my current game (tested with an .ogg and a .wav file)


wheelergames:
Sorry I jumped the gun a bit on making this PR. Closed until it's cleaned up etc

PR #5837, opened 2019-11-08
work to fix an issue in Android when looping music

wheelergames:
(there is a tiny gap between the end and start of the song)

- the fix uses multiple (3) MediaPlayer objects which play one after each other, instead of having one that tries to loop
- potential refactoring to test with 2 MediaPlayer might reduce the footprint


wheelergames:
some edge cases exist, after fully testing

PR #5838, opened 2019-11-10
Log the exception inside flush and fix NullPointerException if LocalStorage is not supported

SimonIT:
We experienced some "Couldn't flush preferences" erros. The cause seem to be that `getLocalStorageIfSupported` can return null if the browser doesn't support it or "block third-party cookies" is enabled. Then in `flush`, GwtPreferences tries to access to LocalStorage but it's null.
Logging the original exception is helpful to get causes of problems.
Also removed the unnecessary creation of objects from the same type in toObject (the constructors are deprecated since java 9)

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtFiles.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtFiles.java
@@ -24,7 +24,7 @@ import com.google.gwt.storage.client.Storage;
 
 public class GwtFiles implements Files {
 	
-	public static final Storage LocalStorage = Storage.getLocalStorageIfSupported();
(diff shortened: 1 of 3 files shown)


intrigus:
Why should `flush` *not* throw an exception when local storage is not supported?

SimonIT:
It makes the whole game unplayable if the exception occurs. Maybe an IOException would be better? So the developer know and can handle the case of not saveable?

MrStahlfelge:
Because of backwards compatibility (in the past, possibility to save was granted) and to keep it conform with the other platforms. There's also nothing to gain here. Handling the exception yourself has no value.

Additionally, it is completely unexpected that the execption is thrown on flush. The moment I would expect it is when the object is loaded with Gdx.app.getPreferences.

SimonIT:
Thanks for testing and approving

MobiDevelop:
Looks okay to me.

PR #5839, opened 2019-11-10
Java 7

thechubbypanda:
In reference to #5829 this is the alternate option.


EtK2000:
Setting target to Java 7 shouldn't really break anything, as its bytecode is compatible with Java 6 AFAIK.

Darkyenus:
Java 6 JVM cannot run Java 7 bytecode, as the bytecode version is different. Although the code itself is probably compatible, as the `invokedynamic` instruction which was added in 7 was not emitted by `javac` yet (I think).

noblemaster:
What @Darkyenus said: it stops working for people that use Java 6.

Users for libGDX can use any Java version they prefer, be it 7, 8, 10, 13 etc. Switching libGDX to 7 doesn't change anything. Methods are still being the same. Keeping libGDX at 6 however, increases compatibility for users that can't go up to higher versions :)

EtK2000:
We should see how many people use LibGDX for a java 6 audience.
If it isn't too many we can upgrade to 7 and explain how to lower it to 6 (with a retro-compiler which many exist).

Darkyenus:
Will we get any benefits by upgrading to 7? For reference, these are the new features present in Java 7 (from [wiki](https://en.wikipedia.org/wiki/Java_version_history#Java_7), italic mine):
- Strings in switch
- Automatic resource management in try-statement *(from Android API level 19! Our min is [currently 9](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/android/build.gradle#L21))*
- Improved type inference for generic instance creation, aka the diamond operator <>
- Simplified varargs method declaration *(just warnings)*
- Binary integer literals
- Allowing underscores in numeric literals
- Catching multiple exception types and rethrowing exceptions with improved type checking *([apparently also dangerous for our API levels](https://gyulajuhasz.com/blog/dangers-of-multi-catch-in-android/))*

From the features which remain (basically better numeric literals and String switch), I am not sure if anything is good enough to break the compatibility, even for one Java 6 user.

Tom-Ski:
I agree with @Darkyenus.

The only problems arise when we start losing forward compatibility. At the moment we can still build using tools, and we don't have issues with people on later jdks using the library. Until then I don't see us _upgrading_ purely because it does limit backwards compatibility.

thechubbypanda:
That's the problem though @tomcashman , when using JDK 13 you can't `mvn install`.
Not can you use the Android SDK manager to be fair but still. It is a major pain to have multiple JDKs installed just so that I can switch and use things that refuse to update such as Android and Libgdx.

noblemaster:
I agree with @Tom-Ski. 

@Darkyenus: nothing prevents you from using Java 7 or 13 in your projects. You get all the list points you mentioned. Doesn't matter if libGDX is compiled to Java 6.

@thechubbypanda: what problem with ``mvn install``? Maybe I am wrong on that(?), but can't you just use the latest JDK and target lower Java versions?

thechubbypanda:
@eobermuhlner If you try to `mvn install` while running jdk 13, it fails saying that java 1.6 is no longer supported. This is to do with the JDK version, not maven since installing with a lower jdk yields no errors. So my understanding is that jdk 13 no longer supports java 1.6 compilation. I may be wrong but the proof is in the maven install.

noblemaster:
You are correct. Based on the comment in the Java 8 pull request https://github.com/libgdx/libgdx/pull/5829, I found it:

> The first and most important thing is that Maven 3.3.1 needs JDK 1.7.
> https://maven.apache.org/docs/3.3.1/release-notes.html

Darkyenus:
@noblemaster I think you misunderstood my comment, I was saying that there is almost no benefit to upgarding to Java 7. I didn't say that I want it.

While Maven does require 1.7 to run, it can still compile for 1.6 just fine - the warning about 1.6 being removed comes from `javac`, not from Maven. `javac` 12 is [still able](https://docs.oracle.com/en/java/javase/12/tools/javac.html) to compile 1.6 sources, while `javac` 13 [documentation](https://docs.oracle.com/en/java/javase/13/docs/specs/man/javac.html) does not specify supported versions.

So if I am not mistaken, there are two possible outcomes of this PR:
- It gets merged, libGDX developers can still use the latest Java on their system, but we lose any users which are still stuck with Java 6. Later down someone might accidentally use a Java 7 feature which isn't available on Android and cause a crash.
- It does not get merged, libGDX developers will have to use at least Java 12 or older to compile libGDX, no change for users.

I am not on either side, I just wanted to make both sides of argument clear.

EtK2000:
Truthfully, if a JRE is bundled, it doesn't affect the end user much except for the file size.

obigu:
I'm seeing the following error message when trying to build with Maven 3.6.3 on a Mac using Adopt JDK 8.

> error: Source option 6 is no longer supported. Use 7 or later.

Setting it to 7 on the `maven-compiler` plugin settings fixes it.

@Tom-Ski I think we are already losing forward compatibility and should reconsider upgrading to Java 7.

Tom-Ski:
I think we are good yes

obigu:
The `build.xml` files need to be updated too. I'll try to push changes on the PR.

obigu (approved):
Should be good to merge

thechubbypanda (commented):
Finally :)

NathanSweet:
I haven't been following this closely so someone else can feel free to press the scary button. Good luck!

MrStahlfelge:
No problem, I'll take all bashes

PR #5840, opened 2019-11-10
Merge difference and delete stale branch

thechubbypanda:
Is there a reason we keep these old branches that are long outdated?


NathanSweet:
I don't know. They aren't mine, so I don't delete them.

thechubbypanda:
@Tom-Ski Do we still need this branch?

PR #5844, opened 2019-11-14
Add assertions to prove that scene2d code is on the GL thread

WickedShell:
This was something I've been wanting to do for a bit now. A lot of code in LibGDX is unsafe to be called anywhere other then the GL thread, which generally is quite manageable, but when you have enough callbacks/helpers it's quite easy to start calling things from threads other then the GL thread, which can lead to race conditions causing crashing within the GL thread. This adds assertions (that you would have to enable when launching the JVM), that would catch the call to manipulate an object from somewhere other then the GL thread. Because they are completely ignored if they aren't enabled on JVM launch there should be no performance penalty during normal usage, but would make it much easier to track down this class of error when needed.

As an example of what of these assertions would look like when tripped:
``` java
Exception in thread "Thread-31" java.lang.AssertionError: Not on the GL thread
        at com.badlogic.gdx.scenes.scene2d.ui.Label.setText(Label.java:108)
        at com.badlogic.gdx.tests.LabelTest$1.run(LabelTest.java:141)
```

To do this I added a `isGLThread()` helper on the Graphics object that the asserts can use. I've only been able to test the Lwjgl and Lwjgl3 backends thus far, and will need help with the iOS and GWT backends. (I can test the Android one, but haven't gotten to it yet).

I'd like to extend this to a lot more of the scene2d code at a minimum, but before going further wanted to send in this prototype to get people's opinion.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -105,6 +105,7 @@ public class AndroidGraphics implements Graphics, Renderer {
 	protected final AndroidApplicationConfiguration config;
 	private BufferFormat bufferFormat = new BufferFormat(5, 6, 5, 0, 16, 0, 0, false);
 	private boolean isContinuous = true;
+	private Thread glThread;
 
 	public AndroidGraphics (AndroidApplicationBase application, AndroidApplicationConfiguration config,
 		ResolutionStrategy resolutionStrategy) {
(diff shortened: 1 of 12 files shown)


Darkyenus:
Creating an static method (`assertOnGlThread`?) which contains the assert might be better for less code duplication. And the performance impact should be non-existent, since it will be very easy to inline an empty static method.

PokeMMO:
It would probably be better to implement this functionality in an external java agent.

I've added a quick implementation that matches the functionality of this PR (Except a warning instead of a fatal error) to https://github.com/PokeMMO/gdx-dbgagent

Tom-Ski:
@PokeMMO great idea!

PokeMMO:
Even if we decide this belongs in an external tool, there is some benefit to be had to adding a custom annotation to methods in libgdx's source code. Maintaining a list of methods externally is much more fragile.

WickedShell:
Yeah, I hadn't really considered a java agent for this either, I still need to try out what @PokeMMO linked, but on first glance it's probably a lot better :)

Tom-Ski:
I think its much better to have this externally.  Where some higher level annotation system might be nice, its also very subjective as to what should get this treatment, and I think best being able to be configured on a per dev basis in some external agent.

PR #5846, opened 2019-11-15
Android backend as android library

SimonIT:
This pr makes an android library of the android backend. This has the advantage, that if someone increases their libgdx version in their build.gradle file and the minsdk is lower than the libgdx one, it throws an error. The next advantage is a better IDE support for the android backend. 
The renaming of the modules fixes also this problems:
image


tommyettinger:
Is an "android library" an AAR file, as opposed to the current JAR? Does Sonatype Maven Central distribute AARs (it looks like this does handle Travis for nightly builds)? Do Eclipse, Netbeans, and command-line Maven have the capability to build libGDX with this PR, or does this restrict builds to IDEA/AS/command-line Gradle? It's clear that Gradle wasn't used for recent nightlies because this PR changed the version from 1.9.9-SNAPSHOT to the current one, unless the version was changed elsewhere.

SimonIT:
The release can be done with maven like before. The goal is to get to the point where we have only gradle to publish. Sonatype maven central supports android libraries. [Here](http://wittchen.io/publishing-jar-aar-to-maven-central/) is for example a tutorial how to do it. If an IDE does not support the "new" (regular) build system for building android libs, you can use the old way. I did not remove this option

SimonIT:
Hey @Tom-Ski, some guys told me on discord told me that this PR got to much changes. It fixes the error from the screenshot above, makes it possible to deploy aar's (Old method is still possible) and migrates to androidx. Are you agree with them? If so, I would make a PR which fixes only the unique path problem with the android gradle plugin. This means still a huge PR, because I have to rename some modules. Will you accept the PR then? I don't want put effort in it if it doesn't get pulled (like some of my other PR's)

Tom-Ski:
Mainly I can't reproduce the problems you are having here, I don't have issues importing and building from IDE, other than a resource problem with IDEA marking directories as non-source, but resource.

I like the goal, of only having Gradle to publish, but we are pretty far away from this with the native builds. And last time I checked, aar had pretty poor support in other IDEs.

mgsx-dev:
about folder renaming, i can confirm there is an issue:
* i'm using eclipse with the old import method (based on git tracked eclipse files) without issues.
* i tried the other method (via gradle) using a fresh Eclipse 2019.12 (latest), a fresh libgdx repo and follow official setup instructions. I can confirm i have a similar issue : 
    * both box2d and controllers projects are correctly imported but are weirdly renamed (eg. gdx-controllers-gdx-controllers).
    * however, dependant projects (eg. gdx-test-lwjgl) reference it incorrectly (eg. gdx-controllers) and reported as missing dependencies.

note that git tracked eclipse files is an issue when importing via gradle: all .project and .classpath files are modified so it's annoying to work with. By fixing folder names, i think we could also remove all eclipse files from repo and remove the old eclipse import method.

obigu:
I also see the issue affecting the 4 modules described in the PR on IDEA 2019.3.3 but the error plays a bit of seek and hide and sometimes doesn't appear even if it's there and pops up at times. 

Importing a totally clean checked out libGDX project displays the project. Also, it ususally shows running a "Reimport All Gradle Projects" from IDEA Gradle menu.

SimonIT:
@Tom-Ski As you can see, more users have the same problem. Sadly, I can't really tell you how to reproduce the error because it appears directly at the start of IntelliJ, like obigu said. Maybe you have to do something with android? Because it's problem with the android gradle plugin. I had a similar error on travis, while making the android backend an android library. Maybe that helps? https://travis-ci.com/github/SimonIT/libgdx/jobs/256907200

Tom-Ski:
Yes I've spoken to @obigu about this, but with unrelated fixes he is able to import ok? (just some other problems to fix also)

SimonIT:
Yes, somehow it runs.
In my opinion, a functional build system is a must have for a library/framework and the current gradle setup is only more or less working. Like Nathan said, libgdx is counting on PR contributions because "it's in maintenance mode". I saw on discord several times persons who ask how to fix these errors, because they think it's their setup which causes the problem. So it scares developers away from contributing which not very great for libgdx.
Another point is, as you said, to have full deployable gradle setup. If we want the android backend as normal android library (I think it's the way to go), we must do this change, because the build will fail as you can see in the previous linked travis build.

obigu:
Yes, at the moment there are 2 separate issues affecting IDE integration, one is related to resources (no issue number created for it) and makes it impossible to work with IDEA and the other is the one this PR fixes which is not that critical but we definitely also want to fix.

IMO this PR goes on the right direction but I'd come back to discuss about the option of splitting it on 3 different PRs when we get the resources issue fixed as it will also probably require a large refactor.

Issue #5111, opened 2018-02-28
AtomicInteger in HTML5 backend


sonicfran86:
Please ensure you have given all the following requested information in your report.

#### Issue details
HTML5/GWT does not have an implementation of AtomicInteger from java.concurrent.

The Guava version can be used to fulfill this.
https://github.com/google/guava/blob/master/guava-gwt/src-super/java/util/super/java/util/concurrent/atomic/AtomicInteger.java

#### Version of LibGDX and/or relevant dependencies
1.9.8

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Darkyenus:
We seem to use AtomicInteger in [`AtomicQueue`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/AtomicQueue.java), but that is [not supported on gwt](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx.gwt.xml#L392). So I am not sure what the point of this issue is.

badlogic:
Spine uses AtomicInteger

MobiDevelop:
Looks like Google moved Guava's `concurrent` emulation into GWT 2.8.2. I'll look into updating to that this weekend.

MobiDevelop:
GWT was updated to 2.8.2, AtomicInteger is now available.

PR #5847, opened 2019-11-20
Give network threads a unique name

WickedShell:
This just allows all the threads a unique name, this is helpful when using a profiler, or when looking at log files that include thread names.

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
+++ b/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
@@ -24,6 +24,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.SynchronousQueue;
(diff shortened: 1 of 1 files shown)


EtK2000:
Wouldn't it be better to have a counter and have the names increment up?

WickedShell:
It's definitely more humane, the only real attraction to the current time was it involved less state to keep in memory and still gives fairly unique IDs. Having said that I've just realized that I then promptly burn more memory on the string to store the timestamp as it's a much longer number, so I should just swap to the counter anyways. Will push an update up for that.

MobiDevelop:
For the same reason that the timestamps could have been the same, it is technically possible to get the same id via the counter. I suppose you could use an AtomicInteger if you wanted to be sure.

WickedShell:
@MobiDevelop Yeah, I'm not to concerned about that, but if we're going to do it, might as well do it right, swapped to (and tested) using `AtomicInteger` instead.

WickedShell:
@MobiDevelop tests pass, but is #5111 going to be a problem? (I'm really not familiar with what is/isn't included with GWT)

MobiDevelop:
@WickedShell Nope, I already updated the GWT backend to GWT 2.8.2 and AtomicInteger is now available (I forgot to close that issue).

Issue #5858, opened 2019-12-11
LibGdx Class Ray : setting a new direction does not normalize the vector


Laurent1967:
When creating a ray, the direction vector is normalized as it should 
```
public Ray (Vector3 origin, Vector3 direction) {
	this.origin.set(origin);
	this.direction.set(direction).nor();
} 
```
but when a new direction is set, the vector is NOT normalized, which is an error

```
public Ray set (Vector3 origin, Vector3 direction) {
	this.origin.set(origin);
	this.direction.set(direction);
	return this;
}
```

these codes are extracted from LibGdx 1.9.11-SNAPSHOT but applies probably to all previous versions.
An obvious consequence is that Ray#getEndPoint(final Vector3 out, final float distance) may return a wrong distance because of a direction vector not normalized.

Laurent Crenier (gquark)

noblemaster:
This needs further discussion. See discord.

Laurent1967:
Happy to see that my little inconsistency bug report generates so much debate ;-)

yuripourre:
It shows how much LibGDX is active.

noblemaster:
PR integrated.

PR #5859, opened 2019-12-11
#5858 Ray normalise origin vector after set methods

fabiitch:
issue #5858

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
+++ b/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
@@ -24,6 +24,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


mgsx-dev:
Since direction field is public, wouldn't be better to let user code ensure providing unit vector and avoid normalization instead?

Javadoc could be improved in that way.

fabiitch:
My java developer part said "if constructor normalise direction vector why set methods dont".

But you are right, if you use set method for re-use same object. its dump to add calcul with nor.

i change commit , i remove nor() on public Ray set (Vector3 origin, Vector3 direction) and precise in javadocs.
I remove it on public Ray set (Ray ray) too cause it was dumb , if you got a Ray object, direction vector is already normalised.

I keep on methods public Ray set (float x, float y, float z, float dx, float dy, float dz) cause i think its necessary with 3 float param.

Your opinion ?

fabiitch:
i push force on my fork and i close this pr.
i reopen new at https://github.com/libgdx/libgdx/pull/5860

sry for miss

mgsx-dev:
there was no need to create a new PR, force push on your branch automatically update this PR.

PR #5860, opened 2019-12-12
#5858 Ray normalise origin vector after set methods

fabiitch:
create a new pr after closing https://github.com/libgdx/libgdx/pull/5859 with push force on my fork.

So :
- Add javadocs for direction vector must be normalized
	public Ray set (Vector3 origin, Vector3 direction)

- Add nor on direction vector3 in method public Ray set (float x, float y, float z, float dx, float dy, float dz)

thx for reading.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/collision/Ray.java
+++ b/gdx/src/com/badlogic/gdx/math/collision/Ray.java
@@ -34,10 +34,15 @@ public class Ray implements Serializable {
 	/** Constructor, sets the starting position of the ray and the direction.
 	 * 
 	 * @param origin The starting position
-	 * @param direction The direction */
+	 * @param direction The direction(must be normalized before) */
 	public Ray (Vector3 origin, Vector3 direction) {
 		this.origin.set(origin);
-		this.direction.set(direction).nor();
+		this.direction.set(direction);
+	}
+
+	/** Constructor, sets the starting position of the ray and the direction. */
+	public Ray (Vector3 origin, Vector3 direction, boolean normalize) {
+		this(origin, normalize ? direction.nor() : direction);
 	}
 
 	/** @return a copy of this ray. */
@@ -73,9 +78,9 @@ public class Ray implements Serializable {
 	}
 
 	/** Sets the starting position and the direction of this ray.
-	 * 
 	 * @param origin The starting position
-	 * @param direction The direction
+	 * @param direction The direction (Vector must be normalized before, use vector.nor() or
+	 *           {@link Ray#setAndNormalizeDirection(Vector3, Vector3)})
 	 * @return this ray for chaining */
 	public Ray set (Vector3 origin, Vector3 direction) {
 		this.origin.set(origin);
@@ -83,8 +88,19 @@ public class Ray implements Serializable {
 		return this;
 	}
 
-	/** Sets this ray from the given starting position and direction.
+	/** Sets the starting position and the direction of this ray.
 	 * 
+	 * @param origin The starting position
+	 * @param direction The direction
+	 * @return this ray for chaining */
+	public Ray setAndNormalizeDirection (Vector3 origin, Vector3 direction) {
+		set(origin, direction.nor());
+		return this;
+	}
+
+	/** Sets this ray from the given starting position and direction.
+	 * dx, dy, dz must give a Vector3 normalized (instead use
+	 * {@link Ray#setAndNormalizeDirection(float, float, float, float, float, float)})
 	 * @param x The x-component of the starting position
 	 * @param y The y-component of the starting position
 	 * @param z The z-component of the starting position
@@ -95,6 +111,20 @@ public class Ray implements Serializable {
 	public Ray set (float x, float y, float z, float dx, float dy, float dz) {
 		this.origin.set(x, y, z);
 		this.direction.set(dx, dy, dz);
+		return this;
+	}
+
+	/** Sets this ray from the given starting position and direction.
+	 * @param x The x-component of the starting position
+	 * @param y The y-component of the starting position
+	 * @param z The z-component of the starting position
+	 * @param dx The x-component of the direction
+	 * @param dy The y-component of the direction
+	 * @param dz The z-component of the direction
+	 * @return this ray for chaining */
+	public Ray setAndNormalizeDirection (float x, float y, float z, float dx, float dy, float dz) {
+		set(x, y, z, dx, dy, dz);
+		direction.set(direction.nor());
 		return this;
 	}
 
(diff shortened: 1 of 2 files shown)


mgsx-dev:
what i suggested was to remove normalization and update javadoc at all place where user code can set direction. 

note that it's a breaking change, CHANGES.md should be updated as well.

It was a suggestion for discussion, there are pros and cons with this change.

fabiitch:
I agree with your suggestion except for set (float x, float y, float z, float dx, float dy, float dz)
cause dx, dy ,dz can easly give an bad direction vector.
changes.md updated too.

The only cons i see to don't normalize vector direction at each set is the error risk for novice user.
gdx users aren't automatically familiar with math and vector normalization.
 
Normalize each time include a loss of performance.
Its possibly to duplicate method with nor() or not , but it is really usefull  ?   

Thanks for your feedback

mgsx-dev:
about CHANGES.md, breaking change is normalization removal in constructor, not added normalization .
the only cons i see is the breaking change actually, but i'm in favor to it.

i'm not fan of half normalization : either normalize everywhere or normalize nowhere. I think we can assume libgdx users knows what are unit vectors and normalization. 

I let maintainers state on this subject now.

Thanks for your contribution BTW :-)

noblemaster:
I think better than changing the existing method which might result in performance loss for developers that already normalized a vector, how about adding 1-2 additional methods `setAndNormalize(...)` instead? Maybe there is a better method name than what I suggested.

PokeMMO:
Imo, constructor equivalent methods should produce the same values as the constructor. Is there a specific reason why constructor is normalized and set is not?

fabiitch:
@PokeMMO is for performance, if you re-use same direction vector on `public Ray set (Vector3 origin, Vector3 direction)` its useless to normalise again the vector.
I think its needed on set(float x,y....) cause its hard to be sure that 3 float give a normalised vector.

fabiitch:
@noblemaster why not create more method, but its really efficient ?

I  think if nobody got same vision, this pr is bad.

Darkyenus:
Having the normalization on `set(Vector2)` but not on `set(float, float)` would probably lead to a lot of confusion. (This issue has been opened because constructor and setter have different behavior!)

I personally think that either all methods should normalize or that none should normalize. Both approaches have their merits:
1. Not normalizing anywhere - we treat `Ray` as a dumb container for vectors. A simple `struct` if you will, with no behavior.
2. Normalizing everywhere - we treat `Ray` as a representation of a mathematical ray. If someone wants to set the direction without normalization, the `direction` variable is public. This leads to no confusion as it is obvious that setting a random `Vector3` will not normalize it.

I initially preferred the first variant, because I like to keep things simple. But after writing this out, I like the second option better, as it has a smaller chance to break existing code and confuse people. It also follows the libGDX pattern of "methods are high level API which does what you expect, fields are low level API, don't touch unless you know what you are doing" (for example `Array` behaves this way).

noblemaster:
I think we all agree that the current implementation is inconsistent. Constructor and set(...) methods all normalizing or not.

I think @Darkyenus's solution (1) makes probably the most sense. We do **not** normalization by default. In addition we should add another constructor and additional methods e.g. `setNormalize(...)` or so that do the work for people that don't want to calculate it themselves. Also, obviously the documentation should be updated.

fabiitch:
re ! 
I apply your suggestion .

So all base method and constructor dont normalize the direction vector.
And one new contructor public Ray (Vector3 origin, Vector3 direction, boolean normalizeDirection) 
And new methods setAndNormalizeDirection.

Changeslog files updated too.

Thanks for your feedback. and check it too 👍

noblemaster:
Thanks for the update. That looks fine! However, I would definitely change `this.direction.set(direction.nor())` into `direction.nor()`. Calling `set(...)` isn't needed.

Other than that, the only thing is probably the method name `setAndNormalizeDirection(...)`. I somewhat suggested it and it's rather long. We can commit it this way. Otherwise, maybe someone has a better suggestion before we finalize things?

fabiitch:
ok done ! 
for direction.nor()) into direction.nor() my bad, i dont go on final method to see my error !

thx for review

gdx/src/com/badlogic/gdx/math/collision/Ray.java
```diff
@@ -34,10 +35,15 @@ public Ray () { }
...
+	 * @param direction The direction(must be normalized before) */
 	public Ray (Vector3 origin, Vector3 direction) {
 		this.origin.set(origin);
-		this.direction.set(direction).nor();
+		this.direction.set(direction);
+	}
+
+	/** Constructor, sets the starting position of the ray and the direction. */
+	public Ray (Vector3 origin, Vector3 direction, boolean normalizeDirection) {
+		this(origin, normalizeDirection ? direction.nor() : direction);
```
Darkyenus: This will change the `direction` vector that was passed in, please don't do that. Simple
```java
this(origin, direction);
if (normalizeDirection) {
    this.direction.nor();
}
```
should suffice. Same for `setNormalize`.

fabiitch: done !

noblemaster:
That should settle it. Thanks for the hard works :)

obigu:
Sorry to comment too late but shouldn't the API enforce the direction normalization? The class is supposed to encapsulate a unit length direction ray (and is treated as such in other classes such as `Intersector`). 

I agree both constructor and setters should normalize, but why is it an option now? Performance wise, if an already normalized `Vector3` is provided, it won't evaluate the `sqrt`, what do you mean by "_Normalize each time include a loss of performance_." @fabiitch? Do you mean if you instantiate it and then set it again with same direction passed in the constructor?

mgsx-dev:
i think this change both over complexify API and breaks backward compatibility. I have some doubts about javadoc wording and method names as well.

I'm in favor to Darkyenus "tradeoff" solution : normalize in all methods and let advanced users setting public field to bypass unnecessary normalization.

@noblemaster : commits wasn't squashed and makes history hard to read (is it too late to revert?)

fabiitch:
@obigu 
> Do you mean if you instantiate it and then set it again with same direction passed in the constructor?
> 
yes, the nor() operation was applied each time, if you give a normalized vector on set() .its a little loss of performance.
The original version don't apply nor() on set() methods. But it was the issues #5858.

I totally agree with you, Ray class should be autonom. Let advanced user set by public field is a good solution. (its a rollback to first version  of this pr^^).

at least as @mgsx-dev said, revert it !

obigu:
@fabiitch If you pass a normalized version to `Vector3#nor()` it won't execute the `sqrt` (please check implementation https://libgdx.badlogicgames.com/ci/nightlies/docs/api/com/badlogic/gdx/math/Vector3.html) so performance wise it won't make a difference. 

I agree we should revert it to just make the setters to apply normalization.

fabiitch:
yes but it calcule lenght of vector and check size. Its not a lot, but useless if v3 is already normalized.

obigu:
> yes but it calcule lenght of vector and check size. Its not a lot, but useless if v3 is already normalized.

Useless optimization, `len2` is immediate.

PR #5871, opened 2019-12-26
Added invert method to vectors.

AndEditor7:
Useful to invert normals or direction if necessary.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector.java
@@ -186,6 +186,10 @@ public interface Vector<T extends Vector<T>> {
 	 * @param v addition vector
 	 * @param mulVec vector by whose values the addition vector will be scaled */
 	T mulAdd (T v, T mulVec);
+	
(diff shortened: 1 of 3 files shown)


MobiDevelop:
In the Matrix classes, there's inv() to get the inverse. Does it make sense to stick with the same method name here?

AndEditor7:
I never knew you can multiply -1 to get an inverted value. Now I'm unsure should I close this pull request since it can multiply -1 to invert the vector. I'll let members decide what to do with this pull request.

I'm petty sure it dose since most vector's method has shortened names.

ryanastout:
I'd vote to close this as not useful. One could call `scl(-1)` as mentioned above to get the same result. If it does get accepted for some reason, I'd vote to name it `negate()`or something; "inv" for matrices means multiplicative inverse, which is well-understood. People rarely refer to the additive inverse of vectors, which is what this would be.

AndEditor7:
I agree - thank you for your response. I hardly remember this pull request, It wasn't long ago. Anyway - it closed now.

PR #5873, opened 2019-12-29
[lwjgl3 backend] fix an empty title and NullPointerException.

AndEditor7:
In the code Lwjgl3Application.java, when constructing a new lwjgl3 application, it clones config instance, then it checks the **cloned** config's title if it is null than replace it with the class's name. The problem within construct initialization, it uses the local config (non-clone) which hasn't check if the title is null, hence when creating a new application with the title set to null will crash when initialization.

With these changes, it will fixes when the title is null, preventing the crash. And will behave similarly to lwjgl2 backend (if the title is null, then replace it with the listener's class name).


Darkyenus:
It seems to me that the real problem is that non-cloned `config` is used after the clone is made, as the clone seems to be there to prevent changing user provided object, and to prevent user from changing the config after it has been used.

Something like:
```java
this.config = config = Lwjgl3ApplicationConfiguration.copy(config);
```
would probably be better and less error prone than the current approach.

PR #5877, opened 2020-01-01
#5858 Ray.java normalize direction vector in set() methods

fabiitch:
finnally ! 

nor() direction vector everytime.
Cause Math.sqrt avoid if vector is already normalized in Vector2#nor().
If you want to avoid the call of nor() , use the public field direction.

thx for reading.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - Update to LWJGL 3.2.3
 - API Change: Table#getRow now returns -1 when over the table but not over a row (used to return the last row).
 - API Change: Tree#addToTree and #removeFromTree now have an "int actorIndex" parameter.
+- API Change: Ray#set methods and Ray#mul(Matrix4) normalize direction vector.Use public field to set and avoid nor()
 
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
--- a/gdx/src/com/badlogic/gdx/math/collision/Ray.java
+++ b/gdx/src/com/badlogic/gdx/math/collision/Ray.java
@@ -63,7 +63,7 @@ public class Ray implements Serializable {
 		tmp.set(origin).add(direction);
 		tmp.mul(matrix);
 		origin.mul(matrix);
-		direction.set(tmp.sub(origin));
+		direction.set(tmp.sub(origin)).nor();
 		return this;
 	}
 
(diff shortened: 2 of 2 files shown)


gdx/src/com/badlogic/gdx/math/collision/Ray.java
```diff
@@ -104,7 +104,7 @@ public Ray set (float x, float y, float z, float dx, float dy, float dz) {
 	 * @return This ray for chaining */
 	public Ray set (Ray ray) {
 		this.origin.set(ray.origin);
-		this.direction.set(ray.direction);
+		this.direction.set(ray.direction).nor();
```
obigu: It's minor but there's no need to normalize as the passed ray will have its direction already normalized

obigu:
To make sure Ray instances have their direction always normalized (except if manipulating via the public field) I think we should also add normalize to `Ray mul (Matrix4 matrix)` in case the matrix scales.

gdx/src/com/badlogic/gdx/math/collision/Ray.java
fabiitch: ye, i got this interogation of the utility in this case. 
I add the nor() for be similar with other method.

fabiitch:
effectively ,
Ray mul (Matrix4 matrix) should included in this pr.

will ammend commit soon.

noblemaster:
We should also add the change to the "CHANGES" text file! :)

fabiitch:
Changelog edited !

CHANGES
```diff
@@ -3,6 +3,7 @@
 - Update to LWJGL 3.2.3
 - API Change: Table#getRow now returns -1 when over the table but not over a row (used to return the last row).
 - API Change: Tree#addToTree and #removeFromTree now have an "int actorIndex" parameter.
+- API Change: Ray#set methods and Ray#mul(Matrix4) normalize direction vector.Use public field to set and avoid nor()
```
intrigus: Nitpick: Missing space between "vector.Use public field".

noblemaster:
Looks fine to me now. Any comments, reservations or is this ready to be merged?

CHANGES
noblemaster: It's now fixed.

intrigus:
LGTM

PR #5878, opened 2020-01-02
Update ExtendViewport.java

shatterblast:
Lines 60 through 65 fix the Viewport re-sizing issue from 1.9.10 and before, concerning ExtendViewport and OrthographicCamera.  In certain situations where an ExtendViewport object was quickly removed and created anew, it would double the values of the OrthographicCamera reacting to the parent Viewport class variables of worldWidth and worldHeight.  The OrthographicCamera would effectively act as if zoomed in by 200%.

Lines 60 through 65 verify if the Camera type is truly an OrthographicCamera and reacts accordingly with the fix.  This may carry a benefit for future 3D endeavors with non-Orthographic Camera types, such as PerspectiveCamera.

I have tested this with my own code, relying upon my own version of ExtendViewport only.  Thank you for reading.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/viewport/ExtendViewport.java
+++ b/gdx/src/com/badlogic/gdx/utils/viewport/ExtendViewport.java
@@ -54,6 +54,15 @@ public class ExtendViewport extends Viewport {
 		this.maxWorldWidth = maxWorldWidth;
 		this.maxWorldHeight = maxWorldHeight;
 		setCamera(camera);
+		
+		//This fixes the Viewport re-sizing issue from 1.9.10 and before, concerning ExtendViewport 
+		//      and OrthographicCamera.
+        	if ( super.getCamera() instanceof OrthographicCamera ) {
+            		( ( OrthographicCamera ) super.getCamera() ).setToOrtho(
+                    		false,
+                    		super.getWorldWidth(),
+                    		super.getWorldHeight() );
+        	}
 	}
 
 	public void update (int screenWidth, int screenHeight, boolean centerCamera) {


mgsx-dev:
I think you forgot to call viewport.update after viewport recreation. It's required to call update at least once before using it (typically done when window is resized). Another reason could be that you changed camera zoom value without resetting it.

If you think it's not the case, please provide a runnable example (SSCCE) that illustrate the issue you explained.

shatterblast:
No, I am correct.  Viewport.update() gets called when OrthographicCamera.update() gets called.  This issue affects only Android to my awareness.  I do not use scaling.  The update is in the Constructor of ExtendViewport.

MobiDevelop:
You must have something you used to reproduce this in order to fix it?

shatterblast:
Yeah, my own source code for my project.  I can break it down maybe in a month or so into a simpler form.  I plan on releasing Open ID Connect stuff merged with gdx-pay and a working GUI in the future into my gpgs-libgdx repository.

I'm sort of blitzing through work right now, and I want to get that done at nearly the same time I finish my other stuff.  Otherwise, I would have to remove myself from work for a good week or more maybe to provide a clean working example to test.

mgsx-dev:
@shatterblast what i meant is : code you added is unnecessary because camera viewport will be updated in viewport.update. And viewport.update should be call before using camera. So your code should be something like this : 

```java
viewport = new ExtendViewport(w, h, camera);
viewport.update(screenWidth, screenHeight);
// then you can use your camera matrix to draw things..
```

note that's the same thing for any viewport subclass.

> Others complain about this issue on the support forums

Could you provide links to issues please?

shatterblast:
@mgsx-dev I will need to produce a list of threads from the support forums.  I will come back with that relatively soon.  Basically, I have been telling folks to use `OrthographicCamera.setToOrtho( boolean, width, height)`, and it frequently fixes their issues.  The width and height values pull from the developer's chosen Viewport at the time, which frequently includes ExtendViewport.  Gdx-pay uses ExtendViewport in the default given example after all.  I previously recommended it, but now, I just say it's an option.

My issue with `viewport.update(screenWidth, screenHeight)` relates to scaling, though some way probably exists that I just fail to notice.  While I think that would very likely work, the developer would still need to place that into the render() method.  That would act as a band-aid of sorts, assuming it would work.

While your suggested code would work in the source code of the update() method from ExtendViewport.java, the potential effects on scaling might cause un-foreseen difficulty in my perspective.  I could certainly test it.  Since my fix goes into the Constructor, it would not have the potential negative consequences with scaling.  Thus, I believe a fix should stay in the Constructor for now.

mgsx-dev:
@shatterblast there are several cases : 
* when using multiple viewports, you have to apply each in render method (for each rendering parts)
* when using a single viewport, viewport is updated in the first resize callback and then when windows is resized by user.
* when changing viewport (your case), you should update it just after creating it and you update it when window is resized.

In both 3 cases, there is no issue and no fix required.

My guess is that you forgot to update it just after recreation (no need to update it at each frame).

shatterblast:
@mgsx-dev Regardless of that, the issue affects me and others.  As stated before by me in my first message...

> In certain situations where an ExtendViewport object was quickly removed and created anew, it would double the values of the OrthographicCamera reacting to the parent Viewport class variables of worldWidth and worldHeight.

This has been an un-avoidable and repeatable issue.  I'm not sure how to reproduce it, so that you can examine the same exact problem on your end without a Stack Trace.  I will produce a working example later.

NathanSweet:
I'm hesitant to merge this change without understanding why it is needed (feels like voodoo!).

shatterblast:
Then stick to your voodoo.

NathanSweet:
@shatterblast it's not my voodoo, it's yours: you suggest a fix without being able to show why it is needed. It's nothing personal, we just can't put in fixes we don't understand. Plus, why should creating an ExtendViewport modify the camera (but only if it's an ortho camera)? That's an odd and unexpected side effect to creating a viewport. It may fix the problem people are encountering, but it seems unlikely to be the right fix -- I can't say for sure until we understand the problem.

shatterblast:
NathanSweet:  > @shatterblast it's not my voodoo, it's yours: you suggest a fix without being able to show why it is needed. It's nothing personal, we just can't put in fixes we don't understand. Plus, why should creating an ExtendViewport modify the camera (but only if it's an ortho camera)? That's an odd and unexpected side effect to creating a viewport. It may fix the problem people are encountering, but it seems unlikely to be the right fix -- I can't say for sure until we understand the problem.

I'm not responding to a bot that says the same exact time every fix.

shatterblast:
@mgsx-dev I'm not sure if I will a working example ready by the time Google prepares their next update in a month or so.

mgsx-dev:
OK then. I think this PR can be closed now.

NathanSweet:
@shatterblast Asking us to make mystery changes is simply unreasonable. It is unfortunate you have chosen such a poor attitude.

badlogic:
WTF?

PR #5883, opened 2020-01-09
performance improvements for Matrix4 operations

mgsx-dev:
Some Matrix4 native methods give significant improvements when replacing them by java code (probably because of JNI overhead). .

I made few tests on a real scenario, gain is around 16% FPS : 320 FPS before, 373 after. 
Context: 1000 model instance, 4000 Matrix4.mul.

Matrix4.mul is heavily used, especially for 3D node tree 
Matrix4.inv is only used by cameras.
others are not used by framework code.

Thanks @Anuken to brought this up and @tommyettinger for the [Matrix4 Comparative Benchmark](https://github.com/tommyettinger/assorted-benchmarks/blob/master/src/main/java/net/adoptopenjdk/bumblebench/examples/Matrix4Benchmarks.md)

Notes: 
* java code was directly copied from GWT emulation implementation.
* my second commit might be questionnable, i made choice of mouving methods instead of adding a bunch of On/Off formatting tags. Tell me if i should reconsider. 
* few C functions become unused, so i removed it. I thing it's fine.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Matrix4.java
+++ b/gdx/src/com/badlogic/gdx/math/Matrix4.java
@@ -1197,75 +1197,6 @@ public class Matrix4 implements Serializable {
 		memcpy(mata, tmp, sizeof(float) *  16);
 	}
 	
-	static inline float matrix4_det(float* val) {
-		return val[M30] * val[M21] * val[M12] * val[M03] - val[M20] * val[M31] * val[M12] * val[M03] - val[M30] * val[M11]
-				* val[M22] * val[M03] + val[M10] * val[M31] * val[M22] * val[M03] + val[M20] * val[M11] * val[M32] * val[M03] - val[M10]
-				* val[M21] * val[M32] * val[M03] - val[M30] * val[M21] * val[M02] * val[M13] + val[M20] * val[M31] * val[M02] * val[M13]
-				+ val[M30] * val[M01] * val[M22] * val[M13] - val[M00] * val[M31] * val[M22] * val[M13] - val[M20] * val[M01] * val[M32]
-				* val[M13] + val[M00] * val[M21] * val[M32] * val[M13] + val[M30] * val[M11] * val[M02] * val[M23] - val[M10] * val[M31]
-				* val[M02] * val[M23] - val[M30] * val[M01] * val[M12] * val[M23] + val[M00] * val[M31] * val[M12] * val[M23] + val[M10]
-				* val[M01] * val[M32] * val[M23] - val[M00] * val[M11] * val[M32] * val[M23] - val[M20] * val[M11] * val[M02] * val[M33]
-				+ val[M10] * val[M21] * val[M02] * val[M33] + val[M20] * val[M01] * val[M12] * val[M33] - val[M00] * val[M21] * val[M12]
-				* val[M33] - val[M10] * val[M01] * val[M22] * val[M33] + val[M00] * val[M11] * val[M22] * val[M33];
-	}
-	
-	static inline bool matrix4_inv(float* val) {
-		float tmp[16];
-		float l_det = matrix4_det(val);
-		if (l_det == 0) return false;
-		tmp[M00] = val[M12] * val[M23] * val[M31] - val[M13] * val[M22] * val[M31] + val[M13] * val[M21] * val[M32] - val[M11]
-			* val[M23] * val[M32] - val[M12] * val[M21] * val[M33] + val[M11] * val[M22] * val[M33];
-		tmp[M01] = val[M03] * val[M22] * val[M31] - val[M02] * val[M23] * val[M31] - val[M03] * val[M21] * val[M32] + val[M01]
-			* val[M23] * val[M32] + val[M02] * val[M21] * val[M33] - val[M01] * val[M22] * val[M33];
-		tmp[M02] = val[M02] * val[M13] * val[M31] - val[M03] * val[M12] * val[M31] + val[M03] * val[M11] * val[M32] - val[M01]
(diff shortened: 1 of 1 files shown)


tommyettinger:
I was pretty surprised at how significant the overhead is for JNI. I wonder if making the changes here would affect GWT, which probably has a custom Matrix4 to avoid JNI and wouldn't need that any more after this PR.

As @Anuken pointed out on Discord, the `float[]` allocation in `inv()` could be avoided, since I think in C it doesn't require GC cleanup but does in Java. There's nothing requiring the temporary 16 floats to be in an array, as far as I can tell, and they could just be 16 separate local float variables. That would avoid all allocation in `inv()`.

PokeMMO:
Before or after this PR is merged, jnigen artifacts need regenerated.

```
mvn install

cd gdx
mvn clean compile exec:java -Dexec.mainClass="com.badlogic.gdx.utils.GdxBuild"
````

If we can manage to remove all native code from Matrix4.java the following two files will probably need to be manually deleted

```
gdx/jni/com.badlogic.gdx.math.Matrix4.cpp
gdx/jni/com.badlogic.gdx.math.Matrix4.h
```

intrigus:
Did anyone try these changes on Android/iOS?

Anuken:
Yes, I ran a test on Android (7.1.1); the performance benefit of using Java vs JNI was even greater there than on desktop.

intrigus:
What about Android 4.4?
E.g. Dalvik instead of ART?

PokeMMO:
It's worth pointing out that `GetPrimitiveArrayCritical` is implementation dependent and some implementations require an allocation and copying back.

https://github.com/libgdx/libgdx/blob/master/gdx/jni/com.badlogic.gdx.math.Matrix4.cpp#L157
https://developer.android.com/training/articles/perf-jni#faq:-how-do-i-share-raw-data-with-native-code

Peter-Warlock:
Problems of many libgdx.math especially Matrix4, Vector3, Quaternion - it using `static tmp variables`. It make it unusable for multithreading. I wrote and use my own inheritor classes Matrix4TS, Vector3TS, QuaternionTS with overload for all methods that uses a static tmp in origanal code without using this tmp in my code.

yuripourre:
@Peter-Warlock, the implementation you mentioned is opensource? If so do you mind sharing?

payne911:
So if I understood properly, this PR offers an improvement which has been confirmed by @Anuken , and approved by @tommyettinger as not being a regression in terms of multithreaded usage. Anything preventing this from being merged?

@intrigus seems concerned about Android 4.4

I do have an old Android laying around (API 16, I believe?). @Anuken what exactly did you do to test out the improvements?

Anuken:
@payne911 I don't have the code around anymore, but essentially what I did was run both operations being benchmarked in a loop many times to warm up, then ran them as separate loops with many iterations and took note of the time taken. It didn't use any special benchmark framework, so the results could have inaccurate, but @tommyettinger 's BumbleBench benchmarks seem to support the conclusion that JNI has worse performance.

tommyettinger:
I don't know of any particularly good Android microbenchmarking libraries/frameworks, so Anuken's approach seems to be the right one. BumbleBench is by the OpenJ9 team (now part of the Eclipse Foundation), and was initially made at IBM for their oddly-named J9 JVM. BumbleBench allows benchmarking with non-HotSpot JVMs; JMH is an alternative by the OpenJDK team that only works on HotSpot OpenJDK-based JVMs. BumbleBench is also quite convenient when all you need is a small comparative benchmark. [The benchmark results I had are here](https://github.com/tommyettinger/assorted-benchmarks/blob/master/src/main/java/net/adoptopenjdk/bumblebench/examples/Matrix4Benchmarks.md), and the benchmarks themselves are in the same folder.

obigu:
JNI overhead greatly exceeds potential performance improvements over Java on small methods (as opposed of heavier stuff such as running the whole time step in Box2D). 

I think the benchmarks and explanations provided on this PR are solid enough and should be merged but, if it may help, I've experienced first hand JNI overhead performance issues on libGDX on a real game both on Android and iOS due to a Box2D JNI call. The culprit in my case is a JNI call inside Box2D `contactFilter` (ends up in `jniGetFilterData`) that can be called hundreds of times per second on scenarios with lots of potential collisions. Avoiding this call and running the same thing on Java makes a huge difference (like allowing 3x the number of bodies on the same World (it obliterates any benchmark on any platform)). I will work on a PR for it but I'd love to see this one merged.

NathanSweet:
Seems like there's enough evidence to merge this.

Can someone handle what PokeMMO mentioned? https://github.com/libgdx/libgdx/pull/5883#issuecomment-572671782

NathanSweet:
Ah, Matrix4 `translate` uses `tmp` and calls `mul`, so `mul` cannot use `tmp`. Will fix.

Peter-Warlock:
I already said before - Matrix, Vector, Quaternion unusable for multithreading, because it using static tmp object variables for static tmp vector, matrix, quaternion. I wrote and use my own inheritor classes Matrix4TS, Vector3TS, QuaternionTS with overload for all methods that uses a static tmp in origanal code without using this tmp in my code.
If somebody need it for use or to fix libGDX classes here is it:
[ThreadSafe.zip](https://github.com/libgdx/libgdx/files/4954364/ThreadSafe.zip)

NathanSweet:
@Peter-Warlock thanks, that looks useful. I'm hesitant to replace libgdx's classes with it, as I'm not sure I'd have time for any follow up maintenance. Also I guess some merging would be needed, as libgdx has some native stuff mixed in.

yuripourre:
@Peter-Warlock Thank you very much for providing it. I will definitely use it in personal projects.

PR #5884, opened 2020-01-09
Polyline contains implemented and mini opti on Polygon/Polyline

fabiitch:
Hello, i see Polyline contains method (from Shape interface) wasnt implemented.

So the contains method check if the point is on one segment of polyline.
And another method to check if a point is nearest of one segment.

I do a mini "opti" on getTransformedVertices , cos and sin arent used if rotation =0.

i know a polyline is not really a shape, but i think contains method can be util.

thanks for reading.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,8 @@
 - Update to LWJGL 3.2.3
 - API Change: Table#getRow now returns -1 when over the table but not over a row (used to return the last row).
 - API Change: Tree#addToTree and #removeFromTree now have an "int actorIndex" parameter.
+- API Change: Polyline#contains implemented (return true if point on one segment)
+- API Addition: Polyline#contains with approximation (return true if point is nearnest of approximation on one segment
 
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
--- a/gdx/src/com/badlogic/gdx/math/Polygon.java
+++ b/gdx/src/com/badlogic/gdx/math/Polygon.java
@@ -68,8 +68,8 @@ public class Polygon implements Shape2D {
 		final float scaleY = this.scaleY;
 		final boolean scale = scaleX != 1 || scaleY != 1;
 		final float rotation = this.rotation;
-		final float cos = MathUtils.cosDeg(rotation);
-		final float sin = MathUtils.sinDeg(rotation);
+		final float cos = (rotation != 0) ? MathUtils.cosDeg(rotation) : 0;
+		final float sin = (rotation != 0) ? MathUtils.sinDeg(rotation) : 0;
 
 		for (int i = 0, n = localVertices.length; i < n; i += 2) {
 			float x = localVertices[i] - originX;
--- a/gdx/src/com/badlogic/gdx/math/Polyline.java
+++ b/gdx/src/com/badlogic/gdx/math/Polyline.java
@@ -204,9 +204,10 @@ public class Polyline implements Shape2D {
 		dirty = true;
 	}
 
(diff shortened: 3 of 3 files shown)


Darkyenus:
I don't think that operation "contains" is meaningful for polylines. Also the first implementation (without tolerances) will work only on very simple cases and will break down very quickly with any "interesting" numbers, due to floating point precision. (Have you tested it?)

The second implementation with tolerances could be ok, but something more general, like distance-polyline-point could be even more useful (if it is not implemented anywhere yet, I haven't checked).

fabiitch:
ah yest, i dont think the case of  floating number , i gonna change and add test.

For second impl, i can add method like get shortest distance to polyline.

If contributors find this pull request useless, there is no problem to close it. I just keep methods on my project. 

So i continue or close pr ?

mgsx-dev:
cos and sin use lookup table under the hood, so i'm not sure it worth it. It could be even worse because of branching. Benchmark would be required at least to ensure it's not worse but i wouldn't spend time to do it.

Contains concept only make sense with solid shapes (shapes with an area), which is not the case here. It brings more confusion IMO.

However, contains with approximation could be useful but i would implement it in Intersector class for consistency and to avoidi confusion : something like Intersector.isPointCloseToPolyline().

fabiitch:
i do a little benchmark 
`public class TesterMain {

	public static void main(String args[]) {
		float max = 1_000_000_000;
		
		long currentTimeMillis = System.currentTimeMillis();
		for (int i = 0; i < max; i++) {
			old(0);
		}
		System.out.println("old 0 : " + (System.currentTimeMillis() - currentTimeMillis));
		
		currentTimeMillis = System.currentTimeMillis();
		for (int i = 0; i < max; i++) {
			new1(0);
		}
		System.out.println("new 0 : " + (System.currentTimeMillis() - currentTimeMillis));

		currentTimeMillis = System.currentTimeMillis();
		for (int i = 0; i < max; i++) {
			old(i);
		}
		System.out.println("old i : " + (System.currentTimeMillis() - currentTimeMillis));
		
		currentTimeMillis = System.currentTimeMillis();
		for (int i = 0; i < max; i++) {
			new1(i);
		}
		System.out.println("new i : " + (System.currentTimeMillis() - currentTimeMillis));
	}

	static void old(float rotation) {
		final float cos = MathUtils.cosDeg(rotation);
	}

	static void new1(float rotation) {
		final float cos = (rotation != 0) ? MathUtils.cosDeg(rotation) : 0;
	}
}`
output : 
old 0 : 1164
new 0 : 1148
old i : 1161
new i : 1175

effectivly there is a loss cause branching on rotation != 0. 

i don't know if its a good benchmark but i see the problem.

I close PR. 
Thanks for feedback

obigu:
`contains` should throw an `UnsupportedOpperationException` to avoid any possible confusion imo.

Agree with @mgsx-dev , the "optimization" is probably not worth it.

I agree with @Darkyenus with adding a more useful `distanceToPolyline()`. I wouldn't clutter `Intersector` any more though.

MobiDevelop:
I don't know about an exception - always returning  false seems fine for indicating that this shape cannot contain points.

obigu:
Well, a user that calls that method expects to know if a point is contained
on a Polyline and return true if that point is part of one of its segments.
Returning always false may be confusing as happened to PR author (until
checking the actual implementation). Throwing an
UnsupportedOperationException conveys exactly what we want, that the
operation doesn't apply to this shape.

Alternatively the javadocs should clarify it as currently the ones from the
interface apply.

Issue #5822, opened 2019-10-25
Sound doesn't loop on Android


bachtrong43:
```java
Sound sound = Gdx.audio.newSound("Sounds/sound.wav");
long soundId = sound.play();
sound.setLooping(soundId, true);
```
Above code will not work on android because a bug of SoundPool. The sound will not loop if you set loop after play. So If you want to loop a sound on Android, you need to use ```java Sound.loop``` method. I think we should add some javadoc to warning this issue on setLooping method.

yuripourre:
Have you tried other sound formats? Background music (with loops) is working for me with .mp3. I didn't try wav.

MrStahlfelge:
He didn't say it does not work, he said it does not work if the sound is already playing.

yuripourre:
@MrStahlfelge right, I misread that.

Monarchis:
@MobiDevelop I'd recommend adding a isPlaying check if possible instead of blind calling pause / resume.

https://github.com/MobiDevelop/libgdx/blob/805476b6b0be5060667d550707169f6b0cda17ef/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidSound.java#L113

MobiDevelop:
Unfortunately there is no such check, but pause and resume are no-ops in cases where it was not playing.

PR #5888, opened 2020-01-13
Add warning about setLooping method in Sound interface

jaeheonshim:
As mentioned in issue #5822 , sounds on Android do not loop when the setLooping method is called after the play method because of a SoundPool bug. This commit adds a notice to the java documentation on the setLooping method of interface Sound.

### DIFF
--- a/gdx/src/com/badlogic/gdx/audio/Sound.java
+++ b/gdx/src/com/badlogic/gdx/audio/Sound.java
@@ -102,7 +102,7 @@ public interface Sound extends Disposable {
 	public void resume (long soundId);
 
 	/** Sets the sound instance with the given id to be looping. If the sound is no longer playing this has no effect.
-	 * <p><b>Note: </b>Due to a bug in SoundPool, the sound will not loop if this method is called after play(). The sound can be still looped using the loop() method.</p>
+	 * <p><b>Note: </b>Due to a bug in SoundPool, the sound will not loop if this method is called after {@link #play()}. The sound can be still looped using the {@link #loop()} method.</p>
 	 * @param soundId the sound id
 	 * @param looping whether to loop or not. */
 	public void setLooping (long soundId, boolean looping);
--- a/gdx/src/com/badlogic/gdx/audio/Sound.java
+++ b/gdx/src/com/badlogic/gdx/audio/Sound.java
@@ -102,7 +102,7 @@ public interface Sound extends Disposable {
 	public void resume (long soundId);
 
 	/** Sets the sound instance with the given id to be looping. If the sound is no longer playing this has no effect.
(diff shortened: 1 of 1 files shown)


mgsx-dev:
maybe specify it's for Android only, since other backends work fine.

also, better use taglets when referencing other methods (eg. `{@link #play()}` )

MobiDevelop:
Having trouble finding a bug report on Android tracker indicating there is a known bug in SoundPool. Got a link?

jaeheonshim:
@mgsx-dev 
I've added a note which specifies the bug relates only to the Android backend.
I also added the @link taglets for the methods in my documentation. Forgot about those!

jaeheonshim:
> 
> 
> Having trouble finding a bug report on Android tracker indicating there is a known bug in SoundPool. Got a link?

Unfortunately, I can't seem to find a bug report for the issue. I just know that the bug is there from the issue I mentioned above, and I added documentation to support it.

obigu:
Being specific to Android I think this java doc should be added to AndroidSound implementation instead.

mgsx-dev:
@obigu i think it's better to have documentation in core because this is where users read documentation. There are already a bunch of this kind of specification for other methods/interfaces in core (eg. Preferences, Files, Clipboard, Input, ...)

MobiDevelop:
We could make it work as opposed to documenting a flaw.

obigu:
@mgsx-dev you're right, I've realised the problem of documenting the implentation later on. It's kind of against best practices to make the interface aware of its implementations (even if it's only javadocs)  but it's actually the best option.

jaeheonshim:
@obigu Just to be clear, you're suggesting to document the problem for now, right?

MobiDevelop:
This commit seems to resolve the issue - https://github.com/MobiDevelop/libgdx/commit/805476b6b0be5060667d550707169f6b0cda17ef if someone wants to give it a try and confirm, I'll push the fix to the main repo.

Monarchis:
> This commit seems to resolve the issue - MobiDevelop@805476b if someone wants to give it a try and confirm, I'll push the fix to the main repo.

Am I wrong or does this pr only changes the docs? @MobiDevelop

MobiDevelop:
That's correct @Monarchis. This PR changes the docs, I was suggesting to address the issue instead.

noblemaster:
@MobiDevelop: I haven't tested it but it looks harmless enough. :) It can't be worse than what we have now. Don't think updating only the docu makes sense.

How about you push it to the main repo (please also update the CHANGES file, so we know what we have done)!

wheelergames:
Does anyone else have issue with music not looping on Android? I've seen it mentioned a lot of times across the internet, for Android but also for LibGDX. I have a workaround that I use, but do others have it too, or is it only me? I don't think that https://github.com/MobiDevelop/libgdx/commit/805476b would fix it, only the soundPool issue. Would it be worth opening another PR for that one?

PR #5893, opened 2020-01-25
Workaround for macOS OpenGL performance issues.

PokeMMO:
Some macOS users experience a large performance degradation rendering 3D with default ModelBatch settings. This workaround seems to greatly alleviate this issue. This issue may be specific to AMD graphics drivers, and if so we may want to limit this workaround to AMD GL renderer strings.

See:
- https://stackoverflow.com/questions/42238124/gldrawelements-on-osx-has-high-cpu-usage#
- https://badlogicgames.com/forum/viewtopic.php?f=11&t=26738


gdx/src/com/badlogic/gdx/graphics/g3d/utils/DefaultTextureBinder.java
```diff
@@ -66,8 +70,13 @@ public DefaultTextureBinder (final int method, final int offset, final int count
 	}
 
 	public DefaultTextureBinder (final int method, final int offset, int count, final int reuseWeight) {
-		final int max = Math.min(getMaxTextureUnits(), MAX_GLES_UNITS);
-		if (count < 0) count = max - offset;
+		int max = Math.min(getMaxTextureUnits(), MAX_GLES_UNITS);
```
fabiitch: Should let final no?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/DefaultTextureBinder.java
```diff
@@ -66,8 +70,13 @@ public DefaultTextureBinder (final int method, final int offset, final int count
 	}
 
 	public DefaultTextureBinder (final int method, final int offset, int count, final int reuseWeight) {
-		final int max = Math.min(getMaxTextureUnits(), MAX_GLES_UNITS);
-		if (count < 0) count = max - offset;
+		int max = Math.min(getMaxTextureUnits(), MAX_GLES_UNITS);
+		if (count < 0) {
+			if (enableMacOSWorkaround && UIUtils.isMac && Gdx.app.getType() == ApplicationType.Desktop)
```
fabiitch: There is no other method to know if the application run on macOs? It include a class from scene2d

gdx/src/com/badlogic/gdx/graphics/g3d/utils/DefaultTextureBinder.java
PokeMMO: Yeah, but it's probably not worth the effort to change back.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/DefaultTextureBinder.java
PokeMMO: Not as far as I could see. At least none compatible with gwt.

Tom-Ski:
This would just disable any mutli-texturing, no? Also the first and last link (the second being undetermined) seem to be talking about two different things. They might be related, but that would be need to be confirmed.

PokeMMO:
>This would just disable any mutli-texturing, no?

Effectively yeah.

>Also the first and last link (the second being undetermined) seem to be talking about two different things. They might be related, but that would be need to be confirmed.

Looks like it yeah, I was linked to it after I made the PR and it looked like the same issue, but seems it's only similar. The underlying cause might be the same, but you're right we should verify this.

mgsx-dev:
not sure it disable multi-texturing, i think it forces to use texture unit 1 (not zero), see [ModelBatch default texture binder offset](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/ModelBatch.java#L82).

I agree, it needs more tests to confirm what's the issue exactly. I'd try several things: ROUND_ROBIN method, force to texture unit 0, other units, force count to 2, more. Also, take a look at GLProfiler stats : how many shader switches you have, what behavior with less switches. I don't have Mac hardware so i can't help.

If issue is confirmed, i would add a new binding strategy (method) "DISABLED" instead of forcing count/offset. Also add some documentation to warn users that parameters could not be used. It's a breaking change if user code rely on texture slots.

Tom-Ski:
Well then it might work for 2 textures, but for any more it won't. 

Were you experiencing this issue yourself @PokeMMO ?

PokeMMO:
>Were you experiencing this issue yourself @PokeMMO ?

Not personally, but another member of my team has recreated it.

Setting it to round-robin, offset to 0, count to 2 still has the issue. The only time we've gotten decent results seems to be with count 1.

It seems to me the issue is more in the *changing* of the texture index instead of what it actually is. So offset 1 vs offset 0 has a minimal difference.

PokeMMO:
I've deployed this workaround and have confirmation that this resolves the issue from several users.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/DefaultTextureBinder.java
Monarchis: @PokeMMO 
`UIUutils#isMac` does not check if it's mac os and is not compatible with GWT, why?
I think it does, also in GWT.
So if the check isMac works, i'd not recommend to add  `DefaultTextureBinder#enableMacOSWorkaround`

btw: #5910 extends UIUtils by `UIUtils#isAndroid` and `UIUtils#isIos`, the pr is currently in progress to test, may if you request changes to this constants, i could look for a solution that ensures that it's mac, if it's not doing right now.

PokeMMO: `UIUutils#isMac` works fine, we also check if it's a desktop app to avoid interfering with GWT.

`DefaultTextureBinder#enableMacOSWorkaround` exists to allow users to disable this workaround in their own apps more easily.

Monarchis: > DefaultTextureBinder#enableMacOSWorkaround exists to allow users to disable this workaround in their own apps more easily.

OK, so my 2nd question: If we have a `DefaultTextureBinder#enableMacOSWorkaround` flag that is only used for some few or only(?) an architecture of mac, why it does not work for all? And how to handle this while deploying, shouldn't do this libgdx automatically, depending on the cpu / system architecture you are using / the app is deployed?

and my third question is:
https://stackoverflow.com/a/42320505 is the answer of you first link, and is the solution, this guy used more than MAX_GLES_UNITS as count, so isn't the solution to reduce the count to the correct value instead of to 1?

mgsx-dev:
This PR can't be merged as is because it disable multi texturing. In order to prevent texture unit changes, a clean solution would be :
- to track already set units in BaseShader
- call texture binder #bind with a "preferred" unit (API change required in texture binder)
- and have a new DefaultTextureBinder method "no recycling" which always bind to the preferred unit.

However, since it seams related to some weird macos driver, i'm not sure it worth to make that change since it could cause more texture switches with good mac drivers.
As per @PokeMMO :
* Intel GPUs are probably not affected
* some Nvidia are (unconfirmed), some Nvidia aren't
* some AMD are (confirmed)

So, we don't have enough information to choose at runtime between Minimize texture switches VS minimize shader unit changes.

PokeMMO:
https://github.com/servo/pathfinder/pull/337 seems to be resolving the same/a similar issue on another project. Their solution seems like it might be a better one.

PR #5896, opened 2020-01-25
add anim from array

OliPerraul:
It would be useful to create an animation with java array instead.
````
        animations.put(
                IDLE_SIDE_ANIM,
                new Animation(
                        FRAME_LENGTH,
                        new TextureRegion[]{
                                sprites[0][4]
                        },
                        Animation.PlayMode.LOOP
                ));
````

instead of 

````
        animations.put(
                IDLE_SIDE_ANIM,
                new Animation(
                        FRAME_LENGTH,
                        new Array(new TextureRegion[]{
                                sprites[0][4]
                        }),
                        Animation.PlayMode.LOOP
                ));

````

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -76,6 +76,15 @@ public class Animation<T> {
 	}
 
 	/** Constructor, storing the frame duration and key frames.
+	 *
+	 * @param frameDuration the time between frames in seconds.
+	 * @param keyFrames the objects representing the frames.
+	 * return the correct type of array. Otherwise, it returns an Object[].*/
+	public Animation (float frameDuration, T[] keyFrames, PlayMode playMode) {
+		this(frameDuration, new Array<T>(keyFrames), playMode);
+	}
+
+	/** Constructor, storing the frame duration and key frames.
 	 * 
 	 * @param frameDuration the time between frames in seconds.
 	 * @param keyFrames the objects representing the frames. */
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -79,6 +79,7 @@ public class Animation<T> {
 	 *
 	 * @param frameDuration the time between frames in seconds.
 	 * @param keyFrames the objects representing the frames.
+	 * @param playMode indicates playback mode
 	 * Create an animation with the given frame duration, keyframes and playmode
 	 */
 	public Animation (float frameDuration, T[] keyFrames, PlayMode playMode) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java [hunk 1]
```diff
@@ -75,6 +75,15 @@ public Animation (float frameDuration, Array<? extends T> keyFrames, PlayMode pl
 		setPlayMode(playMode);
 	}
 
+	/** Constructor, storing the frame duration and key frames.
+	 *
+	 * @param frameDuration the time between frames in seconds.
+	 * @param keyFrames the objects representing the frames.
+	 * return the correct type of array. Otherwise, it returns an Object[].*/
```
Tom-Ski: This javadoc doesn't make sense.

SimonIT: I think he copied the return from the constructor above with the same weird return text
https://github.com/libgdx/libgdx/blob/7c2c9b01e9e76379dea9c2720f7a566ae3b959da/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java#L68-L76

OliPerraul: I did

Darkyenus:
It would be inconsistent with the `Array` constructor, but if the order of the last two parameters was swiched (`PlayMode`, then key frames), we could have vararg there.

Since I guess that the usecase of this new constructor is a creation of animation from hardcoded keyframes, having a vararg there would help even more.

What everybody thinks?

cypherdare:
I vote for making the signature `(float duration, T... keyFrames)` so we don't have mismatched parameter orders. It's easy to set the loop mode after construction.

As is, it will not cooperate with `getFrames()` unless you make the backing GDX Array "type aware". See the first constructor in the existing class. Doing it like this should work, and avoids the extra Array instantiation.

```
	public Animation (float frameDuration, T... keyFrames) {
		this.frameDuration = frameDuration;
		setKeyFrames(frames);
	}
```

cypherdare:
Ignore what I just said. That constructor already exists. But when you call through to another constructor, it should be the existing varargs constructor that cooperates with `getFrames()` and doesn't instantiate a throw-away GDX Array.

cypherdare:
Ignore my deleted comments. 

We already have a `(float duration, T... keyFrames)` constructor. Adding another overload just to save a call to `setPlayMode()` is not worth the cost of having inconsistent parameter orders, in my opinion.

tommyettinger:
I should probably note that anywhere that takes a `T...` varargs can also already take a Java array, just in case anyone wasn't aware. I'd rather not mess with parameter order after seeing BufferUtils... That thing can't be unseen.

MrStahlfelge:
So overall comments here think the change is not good as it is because of the parameter order, and because it is not difficult to achieve the same behaviour with two lines of code instead of one. Closing this.

PR #5897, opened 2020-01-28
Improved isAndroid check in SharedLibraryLoader Should fix #5863

### LINKED ISSUE #5863: Crashes not loading libgdx.so on certain devices

#### Issue details
We see some amount of crash reports from certain devices when loading libgdx games.

#### Reproduction steps/code
Seeing crash reports from firebase.

1.9.10

#### Stacktrace
```java
Caused by com.badlogic.gdx.utils.GdxRuntimeException
Couldn't load shared library 'gdx' for target: Linux, 32-bit
com.badlogic.gdx.utils.SharedLibraryLoader.load (SharedLibraryLoader.java:125)
com.badlogic.gdx.utils.GdxNativesLoader.load (GdxNativesLoader.java:33)
com.badlogic.gdx.backends.android.AndroidApplication.<clinit> (AndroidApplication.java:60)
java.lang.Class.newInstance (Class.java)
android.app.Instrumentation.newActivity (Instrumentation.java:1078)
android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2558)
android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2727)
android.app.ActivityThread.-wrap12 (ActivityThread.java)
android.app.ActivityThread$H.handleMessage (ActivityThread.java:1478)
android.os.Handler.dispatchMessage (Handler.java:102)
android.os.Looper.loop (Looper.java:154)
android.app.ActivityThread.main (ActivityThread.java:6121)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:889)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:779)
de.robv.android.xposed.XposedBridge.main (XposedBridge.java:107)
Caused by java.lang.UnsatisfiedLinkError
dlopen failed: library "/system/lib64/libgdx.so" needed or dlopened by "/system/lib64/libnativeloader.so" is not accessible for the namespace "classloader-namespace"
```

- [ X] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Monarchis:
Improved Android check for newer Versions. Should fix #5863

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java
+++ b/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java
@@ -46,10 +46,22 @@ public class SharedLibraryLoader {
 	// JDK 8 only.
 	static public String abi = (System.getProperty("sun.arch.abi") != null ? System.getProperty("sun.arch.abi") : "");
 
+	//Since Android 5 multiple checks are necessary
+	private static boolean performIsAndroidCheck() {
+		String property = System.getProperty("java.runtime.name");
+		boolean isAndroid = (property != null && property.contains("Android Runtime"));
+
+		property = System.getProperty("java.vm.vendor");
+		isAndroid = isAndroid || (property != null && property.contains("The Android Project"));
+
+		property = System.getProperty("java.vendor");
+		isAndroid = isAndroid || (property != null && property.contains("The Android Project"));
+		return isAndroid;
+	}
+
 	static {
 		boolean isMOEiOS = "iOS".equals(System.getProperty("moe.platform.name"));
-		String vm = System.getProperty("java.runtime.name");
-		if (vm != null && vm.contains("Android Runtime")) {
+		if (performIsAndroidCheck()) {
 			isAndroid = true;
 			isWindows = false;
 			isLinux = false;


PokeMMO:
Detecting Android is not a problem. The only "Problem" was that certain debug strings were printing incorrect data.

https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L124-L125

This change is unnecessary.

Monarchis:
> What if you checked just for "Android".
> Should still be correct, and might catch weird custom ROMs that change it, all without changing much.
> 
> IDK about case sensitivity though.

@EtK2000 you might be right, checking some properties with case sensitivity would make it possible to catch custom roms and might fix just some library loading issues depending on some special devices.

> Detecting Android is not a problem. The only "Problem" was that certain debug strings were printing incorrect data.
> 
> https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L124-L125
> 
> This change is unnecessary.

@PokeMMO Yes, that's correct. I've seen to late that the Exception is thrown on Line 119, like descriped in https://github.com/libgdx/libgdx/issues/5863#issuecomment-566647199:
https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L119

**So what do you recommend closing pr or change the check of the Properties (and rename the pr) to support custom roms, described by @EtK2000 's answer (https://github.com/libgdx/libgdx/pull/5897#pullrequestreview-349873762)?**

PokeMMO:
I am not aware of any custom rom not being properly detected, so I would recommend closing.

Issue #2903, opened 2015-03-04
OutOfMemoryError in ObjectSet<String>


nikialeksey:
```
public class Test extends ApplicationAdapter {
	
	final static int cntWord = 2 * 100 * 100;
	final static int lenWord = 2;
	final static int cntDistinctLetter = 100;
	Random random = new Random();
	
	ObjectSet<String> objSet = new ObjectSet<String>();
	
	String generateRandomString(int lenWord, int cntDistinctLetter) {
		StringBuilder stringBuilder = new StringBuilder();
		for (int i = 0; i < lenWord; i++)
			stringBuilder.append((char)(random.nextInt() % cntDistinctLetter + 32));
		return stringBuilder.toString();
	}
	
	double tms(Consumer<String[]> consumer, String[] items) {
		long start = System.nanoTime();
		
		consumer.accept(items);
		
		return (System.nanoTime() - start) / 1000000.0;
	}
	
	void objectSetAdd(String[] items) {
		for (int i = 0; i < cntWord; i++)
			objSet.add(items[i]);
	}
	
	@Override
	public void create () {
		
		String[] items = new String[cntWord];
		for (int i = 0; i < cntWord; i++)
			items[i] = generateRandomString(lenWord, cntDistinctLetter);
		
		double t = tms(this::objectSetAdd, items);
		
		System.err.println("=========Testing========");
		
		System.err.println("ObjectSet: Adding: " + String.format("%.2f", t));
		
		System.err.println("=========Testing========");
	}

	@Override
	public void render () {
	}
}
```

After run this code i got OutOfMemoryError.

NathanSweet:
Please post code that is runnable.
https://github.com/libgdx/libgdx/wiki/Getting-help#executable-example-code

nikialeksey:
Is this right?
```
package testObjectSet;

import java.util.Random;
import java.util.function.Consumer;

import com.badlogic.gdx.utils.ObjectSet;

public class Example {
	
	final static int cntWord = 2 * 100 * 100;
	final static int lenWord = 2;
	final static int cntDistinctLetter = 100;
	Random random = new Random();
	
	ObjectSet<String> objSet = new ObjectSet<String>();
	
	String generateRandomString(int lenWord, int cntDistinctLetter) {
		StringBuilder stringBuilder = new StringBuilder();
		for (int i = 0; i < lenWord; i++)
			stringBuilder.append((char)(random.nextInt() % cntDistinctLetter + 32));
[...]

PR #5900, opened 2020-02-03
Fix for OutOfMemoryError, slow remove(), high RAM usage in most maps and sets

tommyettinger:
# The Problem:

Almost 5 years ago in #2903 , a seemingly-obscure condition was found that causes ObjectSet, ObjectMap, and related data structures to throw an OutOfMemoryError long before heap should run out. That issue doesn't really convey how serious the bug is; OutOfMemoryErrors are thrown not just by tens of thousands of random keys, but by thousands of typical GridPoint2 keys, hundreds of regular Vector2 keys, or less than 50 maliciously-chosen Strings (!). To make matters worse, memory usage balloons before an OutOfMemoryError would be thrown, and the low-memory-usage goal of libGDX collections becomes missed entirely even if there is no crash. The high memory usage is caused by the capacity doubling in size whenever the stash increases in size by 2 items. While partial fixes are possible by allowing the stash to increase to much larger sizes without changing capacity, this slows the data structure down to a crawl as the stash gets large.

To verify, try to insert the following group of 49 String keys into an ObjectSet:
```java
mySet.addAll("21oo 0oq1 0opP 0ooo 0pPo 21pP 21q1 1Poo 1Pq1 1PpP 0q31 0pR1 0q2P 0q1o 232P 231o 2331 0pQP 22QP 22Po 22R1 1QQP 1R1o 1QR1 1R2P 1R31 1QPo 1Qup 1S7p 0r8Q 0r7p 0r92 23X2 2492 248Q 247p 22vQ 22up 1S92 1S8Q 23WQ 23Vp 22w2 1QvQ 1Qw2 1RVp 1RWQ 1RX2 0qX2".split(" "));
```

These Strings have two possible hashCode() results across 49 non-equal keys, which exhausts more than 8GB of heap because of this bug. Malicious users could use these 49 keys (or many others) as usernames or other potential keys in online games and cause havoc.

# The Fix

I have reimplemented all hash-table-based maps and sets in libGDX to use linear probing and Fibonacci hashing, while keeping their current API (adding some protected fields and methods to make it easier for users to extend some classes). Linear probing is mentioned almost everywhere new hash table algorithms are. [The Wikipedia page on cuckoo hashing](https://en.wikipedia.org/wiki/Cuckoo_hashing#Practice) mentions that linear probing tends to be faster, while cuckoo hashing can offer better worst-case algorithmic complexity (libGDX's current approach does not have this worst-case guarantee, because its stash can resize). Fibonacci hashing helps a lot with practical hashCode()s, which could be biased toward changing only upper or lower bits of their result, by using a relatively simple step to disperse most kinds of hashCode() evenly across the current capacity.

I have also made sure that Json acts the same with these collections as the earlier ones, while also widening support to all hash-based data structures (including ObjectFloatMap, LongMap, etc.). I have not changed UBJson or XML readers or writers. since I don't use either and don't really know what UBJson is. The code in Json for ObjectSet and ObjectMap reading/writing didn't change.

I have run extensive tests on speed, memory usage, and stability against problem hashCodes()s. These data structures, which are called `MERRY` in the test data, routinely are at the best or within a few bytes of the best memory usage of all libraries tested, have the best total memory consumption across the tested sizes, and don't use more memory when hashCode()s are frequently nearing collisions. The speed of these collections tends to be close to libGDX (usually a few percent slower), except that the versions in this PR are twice as fast as the libGDX maps and sets at removal (the current libGDX sets have unusually slow removal when compared across libraries). [Speed graphs for unordered sets on Java 13](https://tommyettinger.github.io/assorted-benchmarks/). [Speed graphs for ordered sets on Java 13](https://tommyettinger.github.io/assorted-benchmarks/ordered/).  [Speed graphs for unordered sets on Java 8](https://tommyettinger.github.io/assorted-benchmarks/java8/). [Speed graphs for ordered sets on Java 8](https://tommyettinger.github.io/assorted-benchmarks/ordered8/). [Memory results on well-distributed Float hashCode()s](https://github.com/tommyettinger/assorted-benchmarks/blob/cfc096c139745e673e181e76808785233cb8cd81/jmh/memory_results_raw.txt#L190-L513), as an analogue to realistic hashCode()s. [Memory results on problematic but non-colliding Integer keys](https://github.com/tommyettinger/assorted-benchmarks/blob/cfc096c139745e673e181e76808785233cb8cd81/jmh/memory_results_raw.txt#L515-L922). [Code that shows ObjectSet crashing with various key types](https://github.com/tommyettinger/assorted-benchmarks/tree/master/src/test/java/com/badlogic/gdx/utils). [JMH benchmarks for libGDX, the PR collections, and many more](https://github.com/tommyettinger/assorted-benchmarks/tree/master/jmh), if you want to spend hours running the speed benchmarks yourself. [Code for the memory benchmark](https://github.com/tommyettinger/assorted-benchmarks/blob/master/jmh/src/test/java/de/heidelberg/pvs/container_bench/MemoryMBeanCheck.java).

## NOTE

This is a complex PR, but it should be API-compatible and not affect the functionality of other code in libGDX. I think that the security bug here is actually quite significant, and if it won't be addressed by updating the maps and sets to some other implementation, it can only be resolved by replacing their usage in front-facing code with JDK Collections.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
+++ b/gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,79 +16,118 @@
 
 package com.badlogic.gdx.utils;
 
+import com.badlogic.gdx.math.MathUtils;
+
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
-import com.badlogic.gdx.math.MathUtils;
-
-/** An unordered map where the keys are ints and values are floats. This implementation is a cuckoo hash map using 3 hashes,
- * random walking, and a small stash for problematic keys. Null keys are not allowed. No allocation is done except when growing
- * the table size. <br>
+/**
+ * An unordered map where the keys are unboxed ints and values are unboxed floats. This implementation uses linear probing with
+ * the backward-shift algorithm for removal, and finds space for keys using Fibonacci hashing instead of the more-common
+ * power-of-two mask. Null keys are not allowed. No allocation is done except when growing the table size.
  * <br>
- * This map performs very fast get, containsKey, and remove (typically O(1), worst case O(log(n))). Put may be a bit slower,
- * depending on hash collisions. Load factors greater than 0.91 greatly increase the chances the map will have to rehash to the
- * next higher POT size.
- * @author Nathan Sweet */
-public class IntFloatMap implements Iterable<IntFloatMap.Entry> {
-	private static final int PRIME1 = 0xbe1f14b1;
-	private static final int PRIME2 = 0xb4b82e39;
-	private static final int PRIME3 = 0xced1c241;
-	private static final int EMPTY = 0;
-
+ * This map uses Fibonacci hashing to help distribute what may be very bad hashCode() results across the
+ * whole capacity. See <a href="https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/">Malte Skarupke's blog post</a>
+ * for more information on Fibonacci hashing. It uses linear probing to resolve collisions, which is far from the academically
+ * optimal algorithm, but performs considerably better in practice than most alternatives, and combined with Fibonacci hashing, it
+ * can handle "normal" generated hashCode() implementations, and not just theoretically optimal hashing functions. Even if all
+ * hashCode()s this is given collide, it will still work, just slowly; the older libGDX implementation using cuckoo hashing would
+ * crash with an OutOfMemoryError with under 50 collisions.
+ * <br>
+ * This map performs very fast contains and remove (typically O(1), worst case O(n) due to occasional probing, but still very
+ * fast). Add may be a bit slower, depending on hash collisions, but this data structure is somewhat collision-resistant.
+ * Load factors greater than 0.91 greatly increase the chances the map will have to rehash to the next higher POT size.
+ * Memory usage is excellent, and the aforementioned collision-resistance helps avoid too much capacity resizing.
+ * <br>
+ * The <a href="http://codecapsule.com/2013/11/17/robin-hood-hashing-backward-shift-deletion/">backward-shift algorithm</a>
+ * used during removal apparently is key to the good performance of this implementation, even though this doesn't use Robin Hood
+ * hashing; the performance of {@link #remove(int, float)} has improved considerably over the previous libGDX version.
+ * <br>
+ * Iteration won't be as fast here as with OrderedSet and OrderedMap.
+ *
+ * @author Tommy Ettinger
+ * @author Nathan Sweet
+ */
+public class IntFloatMap implements Json.Serializable, Iterable<IntFloatMap.Entry> {
 	public int size;
 
-	int[] keyTable;
-	float[] valueTable;
-	int capacity, stashSize;
-	float zeroValue;
-	boolean hasZeroValue;
+	private int[] keyTable;
+	private float[] valueTable;
+
+	private float zeroValue;
+	private boolean hasZeroValue;
 
 	private float loadFactor;
-	private int hashShift, mask, threshold;
-	private int stashCapacity;
-	private int pushIterations;
+	private int threshold;
+	/**
+	 * Used by {@link #place(int)} to bit-shift the upper bits of a {@code long} into a usable range (less than or
+	 * equal to {@link #mask}, greater than or equal to 0). If you're setting it in a subclass, this shift can be
+	 * negative, which is a convenient way to match the number of bits in mask; if mask is a 7-bit number, then a shift
+	 * of -7 will correctly shift the upper 7 bits into the lowest 7 positions. If using what this class sets, shift
+	 * will be greater than 32 and less than 64; if you use this shift with an int, it will still correctly move the
+	 * upper bits of an int to the lower bits, thanks to Java's implicit modulus on shifts.
+	 * <br>
+	 * You can also use {@link #mask} to mask the low bits of a number, which may be faster for some hashCode()s, if you
+	 * reimplement {@link #place(int)}.
+	 */
+	private int shift;
+	/**
+	 * The bitmask used to contain hashCode()s to the indices that can be fit into the key array this uses. This should
+	 * always be all-1-bits in its low positions; that is, it must be a power of two minus 1. If you subclass and change
+	 * {@link #place(int)}, you may want to use this instead of {@link #shift} to isolate usable bits of a hash.
+	 */
+	private int mask;
 
 	private Entries entries1, entries2;
 	private Values values1, values2;
 	private Keys keys1, keys2;
 
-	/** Creates a new map with an initial capacity of 51 and a load factor of 0.8. */
+	/**
+	 * Creates a new map with an initial capacity of 51 and a load factor of 0.8.
+	 */
 	public IntFloatMap () {
 		this(51, 0.8f);
 	}
 
-	/** Creates a new map with a load factor of 0.8.
-	 * @param initialCapacity If not a power of two, it is increased to the next nearest power of two. */
+	/**
+	 * Creates a new map with a load factor of 0.8.
+	 *
+	 * @param initialCapacity If not a power of two, it is increased to the next nearest power of two.
+	 */
 	public IntFloatMap (int initialCapacity) {
 		this(initialCapacity, 0.8f);
 	}
 
-	/** Creates a new map with the specified initial capacity and load factor. This map will hold initialCapacity items before
+	/**
+	 * Creates a new map with the specified initial capacity and load factor. This map will hold initialCapacity items before
 	 * growing the backing table.
-	 * @param initialCapacity If not a power of two, it is increased to the next nearest power of two. */
+	 *
+	 * @param initialCapacity If not a power of two, it is increased to the next nearest power of two.
+	 */
 	public IntFloatMap (int initialCapacity, float loadFactor) {
-		if (loadFactor <= 0) throw new IllegalArgumentException("loadFactor must be > 0: " + loadFactor);
+		if (initialCapacity < 0)
+			throw new IllegalArgumentException("initialCapacity must be >= 0: " + initialCapacity);
+		if (loadFactor <= 0f || loadFactor >= 1f)
+			throw new IllegalArgumentException("loadFactor must be > 0 and < 1: " + loadFactor);
+		initialCapacity = MathUtils.nextPowerOfTwo((int)Math.ceil(initialCapacity / loadFactor));
+		if (initialCapacity > 1 << 30)
+			throw new IllegalArgumentException("initialCapacity is too large: " + initialCapacity);
+
 		this.loadFactor = loadFactor;
 
-		if (initialCapacity < 0) throw new IllegalArgumentException("initialCapacity must be >= 0: " + initialCapacity);
-		initialCapacity = MathUtils.nextPowerOfTwo((int)Math.ceil(initialCapacity / loadFactor));
-		if (initialCapacity > 1 << 30) throw new IllegalArgumentException("initialCapacity is too large: " + initialCapacity);
-		capacity = initialCapacity;
+		threshold = (int)(initialCapacity * loadFactor);
+		mask = initialCapacity - 1;
+		shift = Long.numberOfLeadingZeros(mask);
 
-		threshold = (int)(capacity * loadFactor);
-		mask = capacity - 1;
-		hashShift = 31 - Integer.numberOfTrailingZeros(capacity);
-		stashCapacity = Math.max(3, (int)Math.ceil(Math.log(capacity)) * 2);
-		pushIterations = Math.max(Math.min(capacity, 8), (int)Math.sqrt(capacity) / 8);
-
-		keyTable = new int[capacity + stashCapacity];
-		valueTable = new float[keyTable.length];
+		keyTable = new int[initialCapacity];
+		valueTable = new float[initialCapacity];
 	}
 
-	/** Creates a new map identical to the specified map. */
+	/**
+	 * Creates a new map identical to the specified map.
+	 */
 	public IntFloatMap (IntFloatMap map) {
-		this((int)Math.floor(map.capacity * map.loadFactor), map.loadFactor);
-		stashSize = map.stashSize;
+		this((int)(map.keyTable.length * map.loadFactor), map.loadFactor);
 		System.arraycopy(map.keyTable, 0, keyTable, 0, map.keyTable.length);
 		System.arraycopy(map.valueTable, 0, valueTable, 0, map.valueTable.length);
 		size = map.size;
@@ -96,6 +135,61 @@ public class IntFloatMap implements Iterable<IntFloatMap.Entry> {
 		hasZeroValue = map.hasZeroValue;
 	}
 
+	/**
+	 * Finds an array index between 0 and {@link #mask}, both inclusive, corresponding to the hash code of {@code item}.
+	 * By default, this uses "Fibonacci Hashing" on the int {@code item} directly; this multiplies
+	 * {@code item} by a long constant (2 to the 64, divided by the golden ratio) and shifts the high-quality
+	 * uppermost bits into the lowest positions so they can be used as array indices. The multiplication by a long may
+	 * be somewhat slow on GWT, but it will be correct across all platforms and won't lose precision. Using Fibonacci
+	 * Hashing allows even very poor hashCode() implementations, such as those that only differ in their upper bits, to
+	 * work in a hash table without heavy collision rates. It has known problems when all or most hashCode()s are
+	 * multiples of larger Fibonacci numbers; see <a href="https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/">this blog post by Malte Skarupke</a>
+	 * for more details. In the unlikely event that most of your hashCode()s are Fibonacci numbers, you can subclass
+	 * this to change this method, which is a one-liner in this form:
+	 * {@code return (int) (item * 0x9E3779B97F4A7C15L >>> shift);}
+	 * <br>
+	 * This can be overridden by subclasses, which you may want to do if your key type needs special consideration for
+	 * its hash (such as if you use arrays as keys, which still requires that the arrays are not modified). Subclasses
+	 * that don't need the collision decrease of Fibonacci Hashing (assuming the keys are well-distributed) may do
+	 * fine with a simple implementation:
+	 * {@code return (item & mask);}
+	 *
+	 * @param item a key that this method will use to get a hashed position
+	 * @return an int between 0 and {@link #mask}, both inclusive
+	 */
+	private int place (final int item) {
+		// shift is always greater than 32, less than 64
+		return (int)(item * 0x9E3779B97F4A7C15L >>> shift);
+	}
+
+	private int locateKey (final int key) {
+		return locateKey(key, place(key));
+	}
+
+	/**
+	 * Given a key and its initial placement to try in an array, this finds the actual location of the key in the array
+	 * if it is present, or -1 if the key is not present. This can be overridden if a subclass needs to compare for
+	 * equality differently than just by using == with int keys, but only within the same package.
+	 *
+	 * @param key       a K key that will be checked for equality if a similar-seeming key is found
+	 * @param placement as calculated by {@link #place(int)}, almost always with {@code place(key)}
+	 * @return the location in the key array of key, if found, or -1 if it was not found.
+	 */
+	private int locateKey (final int key, final int placement) {
+		for (int i = placement; ; i = i + 1 & mask) {
+			// empty space is available
+			if (keyTable[i] == 0) {
+				return -1;
+			}
+			if (key == (keyTable[i])) {
+				return i;
+			}
+		}
+	}
+
+	/**
+	 * Doesn't return a value, unlike other maps.
+	 */
 	public void put (int key, float value) {
 		if (key == 0) {
 			zeroValue = value;
@@ -106,336 +200,172 @@ public class IntFloatMap implements Iterable<IntFloatMap.Entry> {
 			return;
 		}
 
-		int[] keyTable = this.keyTable;
-
-		// Check for existing keys.
-		int index1 = key & mask;
-		int key1 = keyTable[index1];
-		if (key == key1) {
-			valueTable[index1] = value;
+		int b = place(key);
+		int loc = locateKey(key, b);
+		// an identical key already exists
+		if (loc != -1) {
+			valueTable[loc] = value;
 			return;
 		}
+		final int[] keyTable = this.keyTable;
+		final float[] valueTable = this.valueTable;
 
-		int index2 = hash2(key);
-		int key2 = keyTable[index2];
-		if (key == key2) {
-			valueTable[index2] = value;
-			return;
-		}
-
-		int index3 = hash3(key);
-		int key3 = keyTable[index3];
-		if (key == key3) {
-			valueTable[index3] = value;
-			return;
-		}
-
-		// Update key in the stash.
-		for (int i = capacity, n = i + stashSize; i < n; i++) {
-			if (key == keyTable[i]) {
+		for (int i = b; ; i = (i + 1) & mask) {
+			// space is available so we insert and break
+			if (keyTable[i] == 0) {
+				keyTable[i] = key;
 				valueTable[i] = value;
+
+				if (++size >= threshold) {
+					resize(keyTable.length << 1);
+				}
 				return;
 			}
 		}
-
-		// Check for empty buckets.
-		if (key1 == EMPTY) {
-			keyTable[index1] = key;
-			valueTable[index1] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		if (key2 == EMPTY) {
-			keyTable[index2] = key;
-			valueTable[index2] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		if (key3 == EMPTY) {
-			keyTable[index3] = key;
-			valueTable[index3] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		push(key, value, index1, key1, index2, key2, index3, key3);
+		// never reached
 	}
 
 	public void putAll (IntFloatMap map) {
-		for (Entry entry : map.entries())
-			put(entry.key, entry.value);
+		ensureCapacity(map.size);
+		if (map.hasZeroValue)
+			put(0, map.zeroValue);
+		final int[] keyTable = map.keyTable;
+		final float[] valueTable = map.valueTable;
+		int k;
+		for (int i = 0, n = keyTable.length; i < n; i++) {
+			if ((k = keyTable[i]) != 0)
+				put(k, valueTable[i]);
+		}
 	}
+	// the old version; I think the new way avoids a little work
+//	   ensureCapacity(map.size);
+//		for (Entry entry : map.entries())
+//			put(entry.key, entry.value);
 
-	/** Skips checks for existing keys. */
+	/**
+	 * Skips checks for existing keys.
+	 */
 	private void putResize (int key, float value) {
 		if (key == 0) {
 			zeroValue = value;
-			hasZeroValue = true;
-			return;
-		}
-
-		// Check for empty buckets.
-		int index1 = key & mask;
-		int key1 = keyTable[index1];
-		if (key1 == EMPTY) {
-			keyTable[index1] = key;
-			valueTable[index1] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		int index2 = hash2(key);
-		int key2 = keyTable[index2];
-		if (key2 == EMPTY) {
-			keyTable[index2] = key;
-			valueTable[index2] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		int index3 = hash3(key);
-		int key3 = keyTable[index3];
-		if (key3 == EMPTY) {
-			keyTable[index3] = key;
-			valueTable[index3] = value;
-			if (size++ >= threshold) resize(capacity << 1);
-			return;
-		}
-
-		push(key, value, index1, key1, index2, key2, index3, key3);
-	}
-
-	private void push (int insertKey, float insertValue, int index1, int key1, int index2, int key2, int index3, int key3) {
-		int[] keyTable = this.keyTable;
-		float[] valueTable = this.valueTable;
-		int mask = this.mask;
-
-		// Push keys until an empty bucket is found.
-		int evictedKey;
-		float evictedValue;
-		int i = 0, pushIterations = this.pushIterations;
-		do {
-			// Replace the key and value for one of the hashes.
-			switch (MathUtils.random(2)) {
-			case 0:
-				evictedKey = key1;
-				evictedValue = valueTable[index1];
-				keyTable[index1] = insertKey;
-				valueTable[index1] = insertValue;
-				break;
-			case 1:
-				evictedKey = key2;
-				evictedValue = valueTable[index2];
-				keyTable[index2] = insertKey;
-				valueTable[index2] = insertValue;
-				break;
-			default:
-				evictedKey = key3;
-				evictedValue = valueTable[index3];
-				keyTable[index3] = insertKey;
-				valueTable[index3] = insertValue;
-				break;
+			if (!hasZeroValue) {
+				hasZeroValue = true;
+				size++;
 			}
-
-			// If the evicted key hashes to an empty bucket, put it there and stop.
-			index1 = evictedKey & mask;
-			key1 = keyTable[index1];
-			if (key1 == EMPTY) {
-				keyTable[index1] = evictedKey;
-				valueTable[index1] = evictedValue;
-				if (size++ >= threshold) resize(capacity << 1);
+			return;
+		}
+		final int[] keyTable = this.keyTable;
+		final float[] valueTable = this.valueTable;
+		
+		int b = place(key);
+		for (int i = b; ; i = (i + 1) & mask) {
+			// space is available so we insert and break (resize is later)
+			if (keyTable[i] == 0) {
+				keyTable[i] = key;
+				valueTable[i] = value;
+				
+				if (++size >= threshold) {
+					resize(keyTable.length << 1);
+				}
 				return;
 			}
-
-			index2 = hash2(evictedKey);
-			key2 = keyTable[index2];
-			if (key2 == EMPTY) {
-				keyTable[index2] = evictedKey;
-				valueTable[index2] = evictedValue;
-				if (size++ >= threshold) resize(capacity << 1);
-				return;
-			}
-
-			index3 = hash3(evictedKey);
-			key3 = keyTable[index3];
-			if (key3 == EMPTY) {
-				keyTable[index3] = evictedKey;
-				valueTable[index3] = evictedValue;
-				if (size++ >= threshold) resize(capacity << 1);
-				return;
-			}
-
-			if (++i == pushIterations) break;
-
-			insertKey = evictedKey;
-			insertValue = evictedValue;
-		} while (true);
-
-		putStash(evictedKey, evictedValue);
-	}
-
-	private void putStash (int key, float value) {
-		if (stashSize == stashCapacity) {
-			// Too many pushes occurred and the stash is full, increase the table size.
-			resize(capacity << 1);
-			putResize(key, value);
-			return;
 		}
-		// Store key in the stash.
-		int index = capacity + stashSize;
-		keyTable[index] = key;
-		valueTable[index] = value;
-		stashSize++;
-		size++;
 	}
 
-	/** @param defaultValue Returned if the key was not associated with a value. */
 	public float get (int key, float defaultValue) {
 		if (key == 0) {
-			if (!hasZeroValue) return defaultValue;
+			if (!hasZeroValue)
+				return defaultValue;
 			return zeroValue;
 		}
-		int index = key & mask;
-		if (keyTable[index] != key) {
-			index = hash2(key);
-			if (keyTable[index] != key) {
-				index = hash3(key);
-				if (keyTable[index] != key) return getStash(key, defaultValue);
-			}
-		}
-		return valueTable[index];
-	}
-
-	private float getStash (int key, float defaultValue) {
-		int[] keyTable = this.keyTable;
-		for (int i = capacity, n = i + stashSize; i < n; i++)
-			if (key == keyTable[i]) return valueTable[i];
-		return defaultValue;
-	}
-
-	/** Returns the key's current value and increments the stored value. If the key is not in the map, defaultValue + increment is
-	 * put into the map. */
-	public float getAndIncrement (int key, float defaultValue, float increment) {
-		if (key == 0) {
-			if (hasZeroValue) {
-				float value = zeroValue;
-				zeroValue += increment;
-				return value;
-			} else {
-				hasZeroValue = true;
-				zeroValue = defaultValue + increment;
-				++size;
+		final int placement = place(key);
+		for (int i = placement; ; i = i + 1 & mask) {
+			// empty space is available
+			if (keyTable[i] == 0) {
 				return defaultValue;
 			}
-		}
-		int index = key & mask;
-		if (key != keyTable[index]) {
-			index = hash2(key);
-			if (key != keyTable[index]) {
-				index = hash3(key);
-				if (key != keyTable[index]) return getAndIncrementStash(key, defaultValue, increment);
+			if (key == (keyTable[i])) {
+				return valueTable[i];
 			}
 		}
-		float value = valueTable[index];
-		valueTable[index] = value + increment;
-		return value;
 	}
 
-	private float getAndIncrementStash (int key, float defaultValue, float increment) {
-		int[] keyTable = this.keyTable;
-		for (int i = capacity, n = i + stashSize; i < n; i++)
-			if (key == keyTable[i]) {
-				float value = valueTable[i];
-				valueTable[i] = value + increment;
-				return value;
-			}
-		put(key, defaultValue + increment);
-		return defaultValue;
+	/**
+	 * Returns the key's current value and increments the stored value. If the key is not in the map, defaultValue + increment is
+	 * put into the map.
+	 */
+	public float getAndIncrement (int key, int defaultValue, int increment) {
+		final int loc = locateKey(key);
+		// key was not found
+		if (loc == -1) {
+			// because we know there's no existing duplicate key, we can use putResize().
+			putResize(key, defaultValue + increment);
+			return defaultValue;
+		}
+		final float oldValue = valueTable[loc];
+		valueTable[loc] += increment;
+		return oldValue;
 	}
 
 	public float remove (int key, float defaultValue) {
 		if (key == 0) {
-			if (!hasZeroValue) return defaultValue;
+			if (!hasZeroValue)
+				return defaultValue;
+			float oldValue = zeroValue;
 			hasZeroValue = false;
 			size--;
-			return zeroValue;
-		}
-
-		int index = key & mask;
-		if (key == keyTable[index]) {
-			keyTable[index] = EMPTY;
-			float oldValue = valueTable[index];
-			size--;
 			return oldValue;
 		}
 
-		index = hash2(key);
-		if (key == keyTable[index]) {
-			keyTable[index] = EMPTY;
-			float oldValue = valueTable[index];
-			size--;
-			return oldValue;
+		int loc = locateKey(key);
+		if (loc == -1) {
+			return defaultValue;
 		}
-
-		index = hash3(key);
-		if (key == keyTable[index]) {
-			keyTable[index] = EMPTY;
-			float oldValue = valueTable[index];
-			size--;
-			return oldValue;
+		final int[] keyTable = this.keyTable;
+		final float[] valueTable = this.valueTable;
+		final float oldValue = valueTable[loc];
+		while ((key = keyTable[loc + 1 & mask]) != 0 && (loc + 1 & mask) != place(key)) {
+			keyTable[loc] = key;
+			valueTable[loc] = valueTable[++loc & mask];
 		}
-
-		return removeStash(key, defaultValue);
+		keyTable[loc] = 0;
+		--size;
+		return oldValue;
 	}
 
-	float removeStash (int key, float defaultValue) {
-		int[] keyTable = this.keyTable;
-		for (int i = capacity, n = i + stashSize; i < n; i++) {
-			if (key == keyTable[i]) {
-				float oldValue = valueTable[i];
-				removeStashIndex(i);
-				size--;
-				return oldValue;
-			}
-		}
-		return defaultValue;
-	}
-
-	void removeStashIndex (int index) {
-		// If the removed location was not last, move the last tuple to the removed location.
-		stashSize--;
-		int lastIndex = capacity + stashSize;
-		if (index < lastIndex) {
-			keyTable[index] = keyTable[lastIndex];
-			valueTable[index] = valueTable[lastIndex];
-		}
-	}
-
-	/** Returns true if the map has one or more items. */
+	/**
+	 * Returns true if the map has one or more items.
+	 */
 	public boolean notEmpty () {
 		return size > 0;
 	}
 
-	/** Returns true if the map is empty. */
+	/**
+	 * Returns true if the map is empty.
+	 */
 	public boolean isEmpty () {
 		return size == 0;
 	}
 
-	/** Reduces the size of the backing arrays to be the specified capacity or less. If the capacity is already less, nothing is
-	 * done. If the map contains more items than the specified capacity, the next highest power of two capacity is used instead. */
+	/**
+	 * Reduces the size of the backing arrays to be the specified capacity or less. If the capacity is already less, nothing is
+	 * done. If the map contains more items than the specified capacity, the next highest power of two capacity is used instead.
+	 */
 	public void shrink (int maximumCapacity) {
-		if (maximumCapacity < 0) throw new IllegalArgumentException("maximumCapacity must be >= 0: " + maximumCapacity);
-		if (size > maximumCapacity) maximumCapacity = size;
-		if (capacity <= maximumCapacity) return;
-		maximumCapacity = MathUtils.nextPowerOfTwo(maximumCapacity);
-		resize(maximumCapacity);
+		if (maximumCapacity < 0)
+			throw new IllegalArgumentException("maximumCapacity must be >= 0: " + maximumCapacity);
+		if (size > maximumCapacity)
+			maximumCapacity = size;
+		if (keyTable.length <= maximumCapacity)
+			return;
+		resize(MathUtils.nextPowerOfTwo(maximumCapacity));
 	}
 
-	/** Clears the map and reduces the size of the backing arrays to be the specified capacity if they are larger. */
+	/**
+	 * Clears the map and reduces the size of the backing arrays to be the specified capacity if they are larger.
+	 */
 	public void clear (int maximumCapacity) {
-		if (capacity <= maximumCapacity) {
+		if (keyTable.length <= maximumCapacity) {
 			clear();
 			return;
 		}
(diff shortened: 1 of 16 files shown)


noblemaster:
+1 from me.

@tommyettinger put a lot of effort into those classes. He also created a large number of tests. The old collections can consume a lot of memory or under some circumstances crash. In my opinion, there is little reason (?) not to use them as a replacement. That said, please comment if there are concerns.

tommyettinger:
It's possible that we can get these collections faster, but not without facing other tradeoffs (most likely, more complex code). The cuckoo-hashed collections were narrowly ahead of just about every other library out there in speed on contains() and add() performance, and it's possible that some contemporaries of cuckoo hashing could get similar results. I initially implemented these collections with Robin Hood hashing, but that used more memory in the setup I used (4 bytes per slot of capacity) and actually turned out slower than linear probing. The linear probing code is very similar, but smaller, and both benefit from a "backward-shift removal" optimization. The PR's data structures are pretty consistently near the best ranking for speed, but rarely at the best position except for memory. They don't have clear weak points in performance.

I'm 100% a fan of Fibonacci hashing now; I've seen from the last memory test (with problematic Integers) that most other ways of mixing can result in excessive capacities when the hashCode() isn't what some other collections would like. I intentionally used a `long` multiplier for the `place()` method. even though it's slower on GWT, because it's guaranteed to be the same on GWT as on desktop, Android, and iOS. I also made a quick test with one of SquidLib's data structures, swapping it out to use a similar linear-probing algorithm, and testing on GWT, but I didn't see FPS dip below 60 any more than usual (and I think the data structure was being edited every frame).

I think unless bugs or incompatibilities are found, any improvements can be submitted in a later PR. I think the security concerns with the current cuckoo-hashed collections are enough to encourage using just about any other collections library for user-submitted keys in online games, and fixing this sooner rather than later would allow such games to just use the latest nightly build of libGDX instead of changing their ObjectMaps and ObjectSets to something else.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+	 * @param placement as calculated by {@link #place(Object)}, almost always with {@code place(key)}
+	 * @return the location in the key array of key, if found, or -1 if it was not found.
+	 */
+	int locateKey (final K key, final int placement) {
+		for (int i = placement; ; i = i + 1 & mask) {
+			// empty space is available
+			if (keyTable[i] == null) {
+				return -1;
+			}
+			if (key.equals(keyTable[i])) {
```
Monarchis: This check in the loop is very slow.
Try to just use:
key.equals(keyTable[i]) => null ckeck is included in the .equals method => double nullckeck;
also 2 if-blocks without an else are also slow => better use if else or use keyTable[i] == null && key.equals(keyTable[i]) => Will be a performance Imrovement.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		}
+	}
+
+	/**
+	 * Returns the old value associated with the specified key, or null.
+	 */
+	public V put (K key, V value) {
+		if (key == null)
+			throw new IllegalArgumentException("key cannot be null.");
+		K[] keyTable = this.keyTable;
```
Monarchis: Avoid the creation of unused variables, even if its just a reference the GarbageCollector has to clean them up, use the field of the class this.keyTable instead during the loop and in later acess.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		V[] valueTable = this.valueTable;
+		int b = place(key);
+		int loc = locateKey(key, b);
+		// an identical key already exists
+		if (loc != -1) {
```
Monarchis: Avoid x != -1 better use x > -1 => also faster

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		}
+	}
+
+	/**
+	 * Skips checks for existing keys.
+	 */
+	private void putResize (K key, V value) {
+		K[] keyTable = this.keyTable;
+		V[] valueTable = this.valueTable;
+		int b = place(key);
```
Monarchis: b is only used once, use for(i = place(key) ) => GC will thank you ;)

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		return loc == -1 ? defaultValue : valueTable[loc];
+	}
+
+	public V remove (K key) {
+		int loc = locateKey(key);
+		if (loc == -1) {
+			return null;
+		}
+		V oldValue = valueTable[loc];
+		while ((key = keyTable[loc + 1 & mask]) != null && (loc + 1 & mask) != place(key)) {
```
Monarchis: avoid doubled calculations (loc + 1 & mask) => use a variable to calc the index

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		resize(maximumCapacity);
+	}
+
+	public void clear () {
+		if (size == 0)
+			return;
+		K[] keyTable = this.keyTable;
+		V[] valueTable = this.valueTable;
+		for (int i = keyTable.length; i > 0; ) {
+			keyTable[--i] = null;
```
Monarchis: for (int i = keyTable.length; i > 0; --i) instead of decreasing in loop, not needed here to decrease in loop only or use a while instead bute here => for (int i = keyTable.length; i > 0; --i) or for (int i = keyTable.length; i-- > 0; ) should be faster

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		ObjectMap other = (ObjectMap)obj;
+		if (other.size != size)
+			return false;
+		K[] keyTable = this.keyTable;
+		V[] valueTable = this.valueTable;
+		for (int i = 0, n = keyTable.length; i < n; i++) {
+			K key = keyTable[i];
+			if (key != null) {
+				V value = valueTable[i];
+				if (value == null) {
```
Monarchis:
```
if ( ( value == null && other.get(key, dummy) != null ) || !value.equals(other.get(key)) )
	return false;
}
```
is the same

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		if (obj == this)
+			return true;
+		if (!(obj instanceof IdentityMap))
+			return false;
+		IdentityMap other = (IdentityMap)obj;
+		if (other.size != size)
+			return false;
+		K[] keyTable = this.keyTable;
+		V[] valueTable = this.valueTable;
+		for (int i = 0, n = keyTable.length; i < n; i++) {
```
Monarchis: ++i instead of i++

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+	public int hashCode () {
+		int h = size;
+		K[] keyTable = this.keyTable;
+		V[] valueTable = this.valueTable;
+		for (int i = 0, n = keyTable.length; i < n; i++) {
+			K key = keyTable[i];
+			if (key != null) {
+				h ^= key.hashCode();
+
+				V value = valueTable[i];
```
Monarchis: use reference instead of variable creation

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,779 @@
...
+		}
+
+		public void remove () {
+			if (currentIndex < 0)
+				throw new IllegalStateException("next must be called before remove.");
+			K[] keyTable = map.keyTable;
+			V[] valueTable = map.valueTable;
+			int loc = currentIndex;
+			final int mask = map.mask;
+			K key;
```
Monarchis: Same here, doubled calc

mquickmann (changes requested):
This is what i'd do in another way, just reviewed the ObjectMap (looked for performance issues)

Tell me if i'm wrong.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: An `int` isn't GC'ed, it's primitive... I can benchmark with all these changes and without, at least.

Monarchis: About that case i nor very sure, but i think in case of an int you are right.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Source? That seems counter-intuitive.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: I'm only doing it this way because the old ObjectMap, and a lot of places in libGDX, do this. I haven't seen much explanation as to why it's done.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Source? Last I heard, this was irrelevant.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
PokeMMO: Local variables are slightly more performant for multiple accesses. And garbage collection does not have an overhead on local variables (Unless doing an allocation via new).

https://stackoverflow.com/questions/21613098/java-local-vs-instance-variable-access-speed

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
PokeMMO: If there is even a benefit that can't be JIT'd, the difference should be so low to not matter. Imo, use whichever is cleaner and easier to understand.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Nope, not the same. Your code will throw an NPE if value is null and other does not contain key.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
PokeMMO: Using the local variable to do a null check is safer and better practice.

tommyettinger: So here you're saying to use this?
```java
if(valueTable[i] != null) h+= valueTable[i].hashCode();
```

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Indeed, in other places `b` is used again, but not here.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
PokeMMO: This doesn't seem like it gives the same result.  `keyTable[i] == null` should return -1, but your suggestion does not.

A short circuiting && probably performed similarly to two if statements anyway.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
noblemaster: That doesn't seem to be the same: ++i will start variable i at '1' (not '0')!? Unless that's intended!?

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Maybe look again at what this method does. Its most common case when the map isn't heavily loaded is to have a null slot immediately where we have the initial placement, so avoiding a call to equals() is advantageous since we can just return -1. `keyTable[i] == null && key.equals(keyTable[i])` => will be an infinite performance degradation, since that's an infinite loop. It needs to return -1 if a null (empty) slot is found or a non-negative slot index if the same key is found.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Thanks for having me look at this code again, I may have spotted a bug.

tommyettinger:
Don't pull yet, I think I spotted a (rare) bug in removal code.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: Chosed the wromng Operator i mentioned || ... It's early in the morning here.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
noblemaster: Maybe we should add some additional tests to make sure they are no other bugs?

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: Tha's true and it's processor / compiler dempending, i just observed it in code i wrote in my rendering loop.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: Welcome.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: Correct, we'd benchmark it,
I'm not very sure if this is faster, but with only few accesses it should be faster

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: `(value == null && other.get(key, dummy) != null) || other.get(key).equals(value)`
could work sorry for that. (swapping the equals)

Monarchis: `(value == null && other.get(key, dummy) != null) || other.get(key).equals(value)`
COULD work sorry for that. (swapping the equals)

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: > Using the local variable to do a null check is safer and better practice.

Why it's saver?

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: > Source? Last I heard, this was irrelevant.

Internally it increments the value and needs to return the old value, like:

```
function inc(int i) {
int retVal = i;
i += 1;
return retVal;
}
```

PokeMMO: These both are functionally identical in a for loop, see https://stackoverflow.com/a/2315795

Monarchis: Correct "functionally" identical, not performance identical!

PokeMMO: Also performance, they compile to the same bytecode.

Monarchis: > Also performance, they compile to the same bytecode.

That's compiler dependent.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: That's part of the problem, this passed Apache Harmony's relevant tests and the bug never surfaced once, not even for a million removals. It would require very specific conditions to occur, mostly relating to hash collisions in the highest slots that wrap around to the earliest, and I'll double-check for other cases this could occur. Most other code used a different technique to look up across that border, because it wasn't removing; only remove() (the class method and also remove() in a nested Iterator class) removes.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: > Local variables are slightly more performant for multiple accesses. And garbage collection does not have an overhead on local variables (Unless doing an allocation via new).
> 
> https://stackoverflow.com/questions/21613098/java-local-vs-instance-variable-access-speed

Thanks showing me that source. Seems like I programmed that the wrong way in the past.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: By that logic I could say "Avoid x != -1 better use (x <= -2 || x >= 0)", because for some magical processor/compiler combination it's faster. Unless you removed all extraneous factors from your inherently-interconnected rendering loop, you really can't make this recommendation as gospel.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: The current way works and I'm doubtful now about whether the replacements you're suggesting are going to work. I'm trying to take my time and fix legitimate bugs, and I don't feel like adding more bugs thanks to because of hastily-written code meant to be a fraction of a nanosecond faster. Think about whether there really might be a performance gain, and read your code before you post it.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: x!=-1 and x>-1 is compiler / processor dependent.
But "(x <= -2 || x >= 0)" really not not even in any case and especially it's wrong in this use case (x > -1 is required for the array) and it's wrong in case of logical processing.

gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
```diff
@@ -320,9 +320,12 @@ public float remove (int key, float defaultValue) {
...
 		final float[] valueTable = this.valueTable;
 		final float oldValue = valueTable[loc];
-		while ((key = keyTable[loc + 1 & mask]) != 0 && (loc + 1 & mask) != place(key)) {
+		int nl = (loc + 1 & mask);
+		while ((key = keyTable[nl]) != 0 && nl != place(key)) {
 			keyTable[loc] = key;
-			valueTable[loc] = valueTable[++loc & mask];
+			valueTable[loc] = valueTable[nl];
+			loc = nl;
+			nl = loc + 1 & mask;
```
Monarchis: You don't need to increment loc anymore?

tommyettinger: This is weird code yeah, but nl starts with the value of (loc + 1 & mask), so we can assign that to loc to have it increment *and* wrap with mask. Then we need to increase nl again, also wrapping, so it still refers to the next slot.

tommyettinger: The bug before had to do with `++loc` incrementing but not wrapping, so when `valueTable[loc]` was accessed, it would be out of bounds. I'm wondering if some condition I didn't see was blocking that out of bounds case from occurring, because it never happened once over many tests and benchmarks. Still, it's better to have certainty that the bug can't happen or silently degrade behavior.

Monarchis: @tommyettinger 

All, right, sorry for my review, next time I'll try to be more precise, especially reading my code better.

gdx/src/com/badlogic/gdx/utils/Json.java
```diff
@@ -572,6 +572,26 @@ public void writeValue (Object value, Class knownType, Class elementType) {
 				writeObjectEnd();
 				return;
 			}
+			if (value instanceof ObjectIntMap) {
```
Monarchis: Why there are so much if blocks and no else if?
it always checks every if block if it matches.

mquickmann (changes requested):

gdx/src/com/badlogic/gdx/utils/Json.java
tommyettinger: It does not, because each returns early. I didn't write this style, I was matching the existing code.

Monarchis: @tommyettinger Yes, but i was wondering in major thinking i was wondering about that style. All right could leave that.

Darkyenus:
Have you checked how it compares to e.g. [hppc](https://github.com/carrotsearch/hppc), [Koloboke](https://github.com/OpenHFT/Koloboke) or other high performance collection libraries? Just curious if it is even worth debugging and maintaining custom ones.

One reason why it would make sense is GWT support, I don't know if there are possible workarounds / if they would be needed. There is also the issue of source compatibility, which could be solved by borrowing already functional implementation from these libraries and "rebranding" it. Both linked libraries are under Apache 2.0.

noblemaster:
For compatibility's sake, I don't think it's feasible to replace the existing collection libraries with something else. Also, considering libGDX/mostly games, the existing libraries makes probably the most sense as opposed to general 3rd party libraries. They are optimized for libGDX, while others aren't. 

Also, there is an upper limit on how far we can go with performance from the physical point of view (so far there are no papers on the ultimate collection library). Using a 3rd partly library might give a +5% speed bonus a the moment but there are that many, who guarantees that they will be maintained in the future? As long as the collections are in libGDX, we have control, possibly incorporate quantum computing. Also, both hppc & Koloboke haven't been updated in years.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
MobiDevelop: I generally assume that the compiler is better at deciding than I am about how to optimize.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
noblemaster: @tommyettinger can you come up with a test that covers "everything"? I think that's the main thing, it should work solidly.

tommyettinger: I don't believe an all-covering test for every usage of a map can exist, that's an infinitely-deep pitfall during TDD. I'll link the  tests I have run against; Harmony's tests are supposed to cover "everything" but obviously that's unreasonable. The adapted Harmony [tests for ObjectMap are here](https://github.com/tommyettinger/merry-ds/blob/master/src/test/java/ds/merry/ObjectMapTest.java#L444-L494) (highlighted range is for remove()). This is already a much higher bar than was set for the current ObjectMap, which doesn't seem to have any tests for most methods, and when it does, the tests are skeletal.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: This has been fixed/changed in all putResize() methods.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
Monarchis: @tommyettinger May basing on the hashing it would be nice to know how high is the chance for collisions is and maybe a few different read write delete combinations would be interesting + performance

tommyettinger:
@Darkyenus actually yes, I have investigated them; see the speed and memory benchmarks in the original post. Koloboke is tested, HPPC is not because it's a pain to test. I think Koloboke had some minor momentum recently, but HPPC seems dead. The PR collections use less memory than Koloboke about half the time, often using half the memory, and sometimes using a few bytes more. Performance is close between most of these optimized collections, except on remove() where almost everything is faster than the current libGDX collections. These PR collections are GWT-compatible, Koloboke probably isn't (I think it uses sun.misc.Unsafe).

tommyettinger:
Also, I should note that Guava may be GWT-compatible, but it seems to use a tremendous amount of memory: usage of memory across libraries

JCF is Java's built-in Collections, Gu is Guava, and Ko is Koloboke, among others. Guava is the only library to use more memory than the Java collections, and uses *a lot* more.
Other competitors...

  - FastUtil does some things out-of-the box that aren't good for GWT, like using very large int multipliers, plus it lacks a .gwt.xml file, the library itself is massive (I think the JAR is larger than all of libGDX), and the current version needs Java 8.
  - Koloboke seems pretty dead in terms of development and at least at one point required a closed-source annotation processor to be run on code to generate optimal set and map types; I think I tested the version before that change. It would need IdentityMap, OrderedMap, and OrderedSet to be added to it. I believe it also needs Java 8.
  - Eclipse Collections are nice, again, Java 8 only and dubious GWT support.
  - Trove is at the bottom of rankings for speed, the only library tested with even worse remove() performance than libGDX; it's often so bad it distorts the graph.

You can write your own too! I'd like to see someone else try! I'd love to see other people testing real-world performance! I'd love to see actual tests for collections in libGDX! No matter what those tests are, I guarantee they will not be sufficient for all cases! Some backend will have or will develop quirks, and this will affect the tests and code run on that backend! Your tests will not be proven to function in the case of cosmic ray bombardment, as NASA requires for some space projects! I started this project over a month ago, and had been thinking about it since May 2019! My code is probably not rock-solid, but it's a lot more solid than the current maps and sets! It uses a proven algorithm! Cuckoo hashing with a resizable stash is not! This is software, it can be updated!

*GASP*

If you find bugs, I will fix them.

I doubt it will be easy to find bugs.

noblemaster:
It appears all concerns have been addressed? The request has a lot of upvote emojis, so I think we should merge it given it fixed memory issues & crashes. Are there any concerns left why this shouldn't be added?

@tommyettinger (1) Is this considered complete & fully tested? Are you still planning to update it? 
@tommyettinger (2) Can you also update the CHANGES.txt indicating that the collection classes have been updated?

tommyettinger:
@noblemaster I consider it complete. Now that @code-disaster tried it in a large codebase, found a bug with initial capacity 0, and I fixed that bug, I would normally have stopped, but I also updated the somewhat-lacking libGDX tests,and I found+fixed another bug (int and long-keyed versions could have had size not update during a resize(); Object keyed maps and sets were unaffected). No one else has found an issue so far, given a week; it doesn't mean there are none, but it certainly means that even with people looking, the potential bugs aren't easy to find. I'll update CHANGES.txt now.

NathanSweet:
I'm a little late to the party, but I'm here! I'd like to review before this is merged. I can find some time tomorrow. While cuckoo is fun, I'm not against replacing it with linear probing. Also I have a large codebase (Spine) that can serve as a real world test.

>  The JDK collections have the best performance on remove and generally good performance all-around, but as the libGDX developers discovered, they tend to be memory-hungry. 

The JRE HashMap `put` allocates an entry for its linked list, which is the main reason it's not usable for games.

I am curious why ObjectMap `remove` would be slow? That seems odd because cuckoo should be fast for get, containsKey, and remove. It should only be slow for put.

I've only taken a brief look at your new ObjectMap so far. It looks pretty good. Can you point me to exactly how you ran the benchmarks? How are the graphs generated? FWIW, I once [put together](https://github.com/EsotericSoftware/kryo/tree/master/benchmarks) some JMH benchmarks with graphs generated using R (looks [like this](http://n4te.com/x/7766-chart.png)). It's relatively easy to reuse and I've since used it as a template for other projects.

gdx/src/com/badlogic/gdx/utils/Json.java
```diff
@@ -572,6 +572,26 @@ public void writeValue (Object value, Class knownType, Class elementType) {
...
+				if (knownType == null) knownType = ObjectIntMap.class;
+				writeObjectStart(actualType, knownType);
+				for (ObjectIntMap.Entry entry : ((ObjectIntMap<?>)value).entries()) {
+					writer.name(convertToString(entry.key));
```
NathanSweet: Is there an actual need to serialize maps with numeric keys to JSON? Keys are always strings in JSON and I've not seen the need for this. Note serialization can be customized via the Serializer interface.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+	 */
+	protected int mask;
+
+	Entries entries1, entries2;
+	Values values1, values2;
+	Keys keys1, keys2;
+
+	/**
+	 * Creates a new map with an initial capacity of 51 and a load factor of 0.8.
+	 */
```
NathanSweet: Should run the libgdx source code formatter so this matches the style of the rest of the source.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+//		for (Entry<K, V> entry : map)
+//			put(entry.key, entry.value);
+//	}
+	public void putAll (ObjectMap<K, V> map) {
+		ensureCapacity(map.size);
+		final K[] keyTable = map.keyTable;
+		final V[] valueTable = map.valueTable;
+		K k;
+		for (int i = 0, n = keyTable.length; i < n; i++) {
+			if ((k = keyTable[i]) != null)
```
NathanSweet: For readability, I'd prefer assignment on a separate line, then checking `k != null`.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+	 * that don't need the collision decrease of Fibonacci Hashing (assuming the key class has a good hashCode()) may do
+	 * fine with a simple implementation:
+	 * {@code return (item.hashCode() & mask);}
+	 *
+	 * @param item a key that this method will hash, by default by calling {@link Object#hashCode()} on it; non-null
+	 * @return an int between 0 and {@link #mask}, both inclusive
+	 */
+	protected int place (final K item) {
+		// shift is always greater than 32, less than 64
+		return (int)(item.hashCode() * 0x9E3779B97F4A7C15L >>> shift);
```
NathanSweet: HashMap also uses a supplemental hash function (and special cases string keys). How does that compare to this?

I wonder how cuckoo ObjectMap behavior changes if this is used for one of its hash functions? It will still have the cache miss disadvantage of course.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+		int loc = locateKey(key, b);
+		// an identical key already exists
+		if (loc != -1) {
+			V tv = valueTable[loc];
+			valueTable[loc] = value;
+			return tv;
+		}
+		for (int i = b; ; i = (i + 1) & mask) {
+			// space is available so we insert and break (resize is later)
+			if (keyTable[i] == null) {
```
NathanSweet: `locateKey` has already done the work to find the next empty slot, it's wasteful to find it again.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+			resize(keyTable.length << 1);
+		}
+		return null;
+	}
+
+	//	public void putAll (ObjectMap<K, V> map) {
+//		ensureCapacity(map.size);
+//		for (Entry<K, V> entry : map)
+//			put(entry.key, entry.value);
+//	}
```
NathanSweet: Should delete.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+		V[] valueTable = this.valueTable;
+		for (int i = place(key); ; i = (i + 1) & mask) {
+			// space is available so we insert and break (resize is later)
+			if (keyTable[i] == null) {
+				keyTable[i] = key;
+				valueTable[i] = value;
+				break;
+			}
+		}
+		if (++size >= threshold) {
```
NathanSweet: I think we don't need to check this, as we know that the resize will not result in the threshold being exceeded.

gdx/src/com/badlogic/gdx/utils/Json.java
Monarchis: @NathanSweet `convertToString(entry.key)` aren't they converted to string?

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+	}
+
+	public V remove (K key) {
+		int loc = locateKey(key);
+		if (loc == -1) {
+			return null;
+		}
+		V oldValue = valueTable[loc];
+		int nl = (loc + 1 & mask);
+		while ((key = keyTable[nl]) != null && nl != place(key)) {
```
NathanSweet: I wonder about finding the indices of the elements to shift, then shifting them with arraycopy. Dealing with wrapping around is a little annoying, could use two loops. Maybe only worthwhile for large maps near the threshold. Would not want to hurt performance for smaller maps.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
```diff
@@ -1,846 +1,782 @@
...
+		return true;
+	}
+
+	/**
+	 * Uses == for comparison of each value.
+	 */
+	public boolean equalsIdentity (Object obj) {
+		if (obj == this)
+			return true;
+		if (!(obj instanceof IdentityMap))
```
NathanSweet: This (and below) should be ObjectMap, not IdentityMap. Blame shows this was my fault! Probably copy/paste from IdentityMap and no one ever noticed.

NathanSweet:
BTW, could you please sign and email the libgdx CLA? contact@badlogicgames.com
https://github.com/libgdx/libgdx/blob/master/CLA.txt
Cheers!

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Agreed, on it.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: The main advantage here comes from that it actually does linear probing instead of just failing when the stash is full of collisions. Using this in the cuckoo-hashing framework wouldn't help much without significant hacking on its response to failure states, and in all the attempts I made I couldn't balance the need to resize capacity (to enlarge the mask) vs. resizing the stash (which gets quite slow if it can grow without check).

Also, remove() is twice as fast now.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: I did, that's the only reason why it looks like this. Either the source code formatter as exactly posted in the libGDX repo is producing different results on my machine than yours, or perhaps you never ran it on the original ObjectMap over 5 years ago. I point to #5698 and gesture.

gdx/src/com/badlogic/gdx/utils/Json.java
tommyettinger: Indeed, @Monarchis is right, they are converted to and from String to store in JSON format. Just because you haven't seen the need yet doesn't mean you won't encounter some such need in the future. Passing around Json objects with custom serializers set is a hassle that the library can't do, while IIRC using Json.Serializable required an inner map inside each map, and adapting Json seemed to match the other special treatment that libGDX data structures get in that class.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: This shows up in a lot of places, I think. I'll try a mass find/replace and see how it goes.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: The highlighted line 200 only runs if `loc == -1`, meaning the given key was not found already present in the map. `locateKey()` could be changed, or a variant added, that somehow can give either the position of the first empty space or the position of the found key. I have some idea of how I could do it.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: I'll try this out. It looks like the resizing shouldn't happen again, and even if the first resize() didn't give all the keys unique locations, probing should ensure that they'll settle somewhere in the extra capacity.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: It's an option, I guess. I think the bulk of the time spent here is used by finding those indices of what to shift, and the arraycopy would require some complex wraparound code (since it copies a linear span of memory, and when wrapping there are two spans) that would probably counteract any performance gain. If you want to write it and you think this might benefit, feel free. I don't feel terribly confident that I would write that wrapping code correctly the first time.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Fixing now. The only other places I see `equalsIdentity()` in this PR are IntMap and LongMap, and neither of those mentions IdentityMap.

tommyettinger:
RE: Benchmarks, code is here https://github.com/tommyettinger/assorted-benchmarks/tree/master/jmh along with the .fods spreadsheets (OpenOffice/LibreOffice format) that have all the chart/graph info. I can get the exact command lines I used to run JMH, but I'll need to boot up a different computer. I ran on Java 8, but before that I ran the tests on Java 13, all on a recent Manjaro Linux laptop with an 8th-gen i7 hexacore processor.

RE: CLA, the process involves placing a liquid more expensive than human blood onto a dead bleached tree, correct? That is, printing a sheet of paper with that exorbitant ink, signing it with a pen, and scanning it to email again? If lawyers are involved at all, are the libGDX forums GDPR-compliant? If lawyers aren't involved, are Voodoo priests involved? For that matter, if lawyers are involved, you may need some Voodoo priests to morally cleanse the place. I'll make sure to provide detailed stick-figure illustrations in my CLA.

gdx/src/com/badlogic/gdx/utils/Json.java
NathanSweet: Sorry, I meant to post my comment below, where IntMap and LongMap are handled. There's two issues: 1) non-string keys are not a great fit for JSON data, and 2) the instanceof checks were never intended to be expanded to a long list. It would be better to write various `Json.Serializer`s for each libgdx class we want to support, chosen based on expected need, and those can be registered normally, via `setSerializer` (whether automatically, or by the user one-by-one, or by the user via a new method `registerDefaultSerializers`).

> Passing around Json objects with custom serializers set is a hassle that the library can't do

What does this mean? Configuring the Json instance is standard API usage.

> while IIRC using Json.Serializable required an inner map inside each map

It would be done like this, there is no extra map:
```
Json json = new Json();
json.setOutputType(OutputType.json);
json.setSerializer(IntMap.class, new Serializer<IntMap>() {
	public void write (Json json, IntMap map, Class knownType) {
		json.writeObjectStart();
		for (IntMap.Entry entry : ((IntMap<?>)map).entries())
			json.writeValue(Integer.toString(entry.key), entry.value);
		json.writeObjectEnd();
	}

	public IntMap read (Json json, JsonValue jsonData, Class type) {
		IntMap map = new IntMap(jsonData.size);
		for (JsonValue child = jsonData.child; child != null; child = child.next)
			map.put(Integer.parseInt(child.name), json.readValue(null, child));
		return map;
	}
});
IntMap map = new IntMap();
map.put(1, "one");
map.put(2, "two");
map.put(3, "three");
System.out.println(map); // [3=three, 2=two, 1=one]
String string = json.toJson(map);
System.out.println(string); // {"1":"one","2":"two","3":"three"}
IntMap map2 = json.fromJson(IntMap.class, string);
System.out.println(map2); // [3=three, 2=two, 1=one]
```

Anyway, changes to Json can be separate from this PR, which is already large.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
NathanSweet: For some reason your source does not have the right format. Which formatter did you use? [This](https://github.com/libgdx/libgdx/blob/master/eclipse-formatter.xml) is the one. I don't know if the IDEA formatter is equivalent. If it is not, it should be fixed.

All of the libgdx classes I maintain use the formatter. Other contributors tend to be less strict. You can test your formatter by choosing a class I maintain, eg `FloatArray`, run the formatter, and the result should be no changes.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
NathanSweet: The problem with our cuckoo map is likely to do with the choice of hash functions. With 3 hash functions and a stash, cuckoo is not supposed to fail as it does.

You didn't address how your rehashing compares to HashMap's. It would be helpful to have a good understanding of which is better in theory.

As previously mentioned, cuckoo remove should not be slow. The work needed is the same as get, with the addition of nulling the value. The algorithms are different, cuckoo has cache misses and linear probing has variable worst case time, but I would still expect linear probing remove to be slower on average than cuckoo remove, certainly not twice as fast, because values need to be shifted. Cuckoo should be slower on put and faster for remove, get, contains. That we don't see that in the benchmarks indicates something is wrong, somewhere.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
NathanSweet: `locateKey` could be inlined for put, so the code can handle an existing key and still make use of the location of the next null. Java's lack of tuples make using a method for this awkward.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
NathanSweet: One advantage to two loops (from `loc` to the end and from 0 to `loc`) is we don't need to apply the mask each iteration. Just a thought, but if we used two loops everywhere, then we wouldn't need the mask at all and growing the tables would not be limited to powers of two.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
PokeMMO: The main issue seems to be how it handles collisions and resizing the stash.

When collisions occur, it doubles the entire capacity to add a small amount of stash slots. Better hashing can't fix this, especially if you are adding user provided strings which can all have the exact same hashcode.

As for trying to balance resizing the stash or resizing the capacity, it's hard to know which should be done. You want to resize capacity if the hashing is the problem, but stash if rehashing provides minimal benefit.

NathanSweet: The cuckoo issue is failure for the pathological case where all the keys have the same hash value. The stash reduces how often resizes are needed but isn't intended as a solution for bad keys. I think replacing it with linear probing is reasonable, though I still find cuckoo interesting. It's kind of funny it has constant worst case time but then doesn't have a solution for bad keys.

Other maps do poorly with bad keys but don't completely fail: chaining becomes a linked list (ouch!), linear probing does a whooole lot of comparisons.

NathanSweet:
I created a simple benchmark by copying what I linked above and putting in ObjectSet, your ObjectSet (as of 56a80e4) which I named MerrySet for now, and HashSet. The benchmarks are very simple, without being convoluted by framework stuff, but the main reason was this setup generates nice charts. I put up the code here:
https://github.com/NathanSweet/mapbench
The charts resulting from running `run.sh` are:



Hmm, I see the 1,000,000 chart has the wrong label, but the data is legit. Not worth another 35 minute run to fix!

As in your charts, performance for linear probing looks pretty acceptable already, nice work! There may be some ways to improve it slightly, eg inlining `locateKey` in `add`, described above. I see the cuckoo remove is a bit slow, as you also found, though not twice as slow on my hardware (Java 13, 5ghz 8 core beast). I guess the 3 look ups end up more expensive on average, and probably linear probing's shifting is not always a big hit.

It would be interesting to see performance on Android, though I'm not personally up for that level of suffering.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
tommyettinger: Yep, that's the one I used, and installed as the default in IntelliJ using the instructions on the official site. I'll download it again and make sure the formatter is exactly as-downloaded. If the linked Eclipse formatter doesn't produce the same results across different installations (especially since **the official site strongly recommends IDEA**), then I think a Gradle plugin is required if formatting somehow matters.

I also want to emphasize that libGDX's code is not a beauty pageant, and this is a bug fix.

tommyettinger: Yep, reinstalled eclipse-formatter.xml , it brings all sorts of new unpleasantness to `FloatArray`: https://gist.github.com/tommyettinger/9c43514f5b4a9d98675829f16862524c <-- here, Before is FloatArray from the repo, After is with eclipse-formatter run (check the indents, it mixes tab and spaces), and After_IDEA uses idea-formatter (which fixes the mixing, but changes comments and single-line if statements). I'm pretty sure I had to manually fix the spaces and tabs once, and I vaguely remember doing so in a past contribution, then I must have gotten it to match the repo behavior and use just tabs.

Spending this much time and effort on non-functionality-related code-style is silly. If eclipse-formatter.xml from *this repo* and idea-formatter.xml from *this repo* don't replicate the style you want, **if the style matters to you**, then add instructions or tools to *this repo or its associated documentation* to allow contributors to produce that style. If I can't produce this style automatically, I simply won't do it; the effort to hand-tweak every one of the many lines in this PR, to match a code style that the IDE will fight me on at every step, is absurd. If you can do it automatically, then do it. You, @NathanSweet , have the ability to edit this PR; only you, Mario Zechner, and I can do that to my knowledge. Of those three, I can't seem to get any IDE to reproduce the code style used in libGDX given the XML files in libGDX itself, Mario I don't think is still involved in libGDX (at least for PRs), and that leaves you.

obigu: @tommyettinger is making a very valid point here. The current Eclipse formatter file produces different outputs depending on the IDE. If PRs are going to be rejected based on code format (this is not the first PR that is on hold for this reason) and there's no way to ensure a standard format across IDEs, formatting should be made automatic. Alternatively some flexibility should apply on the merging policy. Wasting time on manually editing lines to please a very specific non automatable code style test is quite demoralizing for contributors.

NathanSweet: IMO code style is important to normalize contributions. No one suggested manually matching the style. If the Eclipse formatter doesn't work as expected in IDEA and the IDEA formatter doesn't match Eclipse, then it would be helpful to find a solution. Maybe #5698 is that, I haven't had time to do more than read the issue.

@Tom-Ski added the IDEA formatter file, maybe he has some insight on why it doesn't match Eclipse or if it can be made to match?

NathanSweet: For this PR I'll format and merge.

Tom-Ski: I pretty much only format method code on rare occasions, the rest is just the style that I am used to that matches. It _should_ in theory be ok because this comes directly from IDEA conversion. If it doesn't we should address it in a different issue and try to get these matched up. I agree with the general sentiments regarding this issue though, and only really have problems when code styles in the code don't match. Spaces around methods, wrapping, and indentation values.  These are the only critical things for me at least.

@tommyettinger just a side, the reason why that is recommended because it is the most robust way to get setup with the project and developing quickly for all platforms.

NathanSweet:
Merged! Great research and work @tommyettinger! Open to changes if there are improvements to be made or if you hate something that I did to your baby. Cheers!

NathanSweet:
FWIW, I ran benchmarks overnight (7.5hrs!) with the latest:



Data: http://n4te.com/x/7778-results.zip
The run used this commit of the benchmarks (the JMH args are in `run.sh`):
https://github.com/NathanSweet/mapbench/tree/d3b5feae7d1af099f36808c27cd96fa6a7fad4d8
I think the batch size might be a little low for the < 10k sizes, but I'm no JMH expert.

tommyettinger:
Times all seem quite close; that's good I hope. For every size change, the second-best seems to also change, which suggests either background noise (which seems unlikely because there's quite a bit of relative change; there would need to be a lot of noise somehow) or some "sweet spot" being hit by a particular power-of-ten size. If you want a quick check of a problematic case for the previous GDX IntSet, you can enter multiples of 256, which shouldn't crash IntSet but may cause it to use stash sooner than expected. I think that type of input may also be an issue for HashSet, though not an especially serious one. Even though HashSet has good times in these benchmarks, it seemed to use quite a lot of memory for me; libGDX (Cuckoo) was about equal to best or second-best as long as it didn't use stash, and the merged collections were best (often a few bytes better than Cuckoo libGDX) or a few bytes behind Koloboke when Cuckoo had trouble and Koloboke didn't. Koloboke is very narrowly focused and doesn't have, for instance, IdentityMap; Koloboke sometimes had its own problematic resizing that made it use twice the memory, but doubling memory usage really hurts for million-item sets. HashSet was way in back, using easily 3x to 5x the memory, depending on the data structure, and sometimes more.

gdx/src/com/badlogic/gdx/utils/ObjectMap.java
NathanSweet: To add another thought after all this time: the mask requires POT table sizes, making 1073741824 the largest possible table size. Doubling that gives -2147483648 and failure. Needing such a huge map is likely rare, so we'd not like to affect performance for much more common smaller maps, but maybe there's some special case trickery that would allow us to use a 2147483647 table size?

tommyettinger: I've thought about it a bit; there's actually a maximum array size in Java that's slightly less than 2147483647, depending on VM: https://stackoverflow.com/a/3039805/786740 It's a bit of a challenge to test, for me, because the responses mention needing 13GB of RAM available...

NathanSweet:
@tommyettinger An internet denizen gave some input on our hashing here, if you're interested:
- https://www.reddit.com/r/programming/comments/1v1lqdd/comment/oypmlh5/
- More detail: https://www.reddit.com/r/programming/comments/1v1lqdd/comment/oyskdzg/

NathanSweet:
Blog analysis: https://sooniln.github.io/posts/hashmap-benchmarks-2026/#libgdx-naivecopy-performance

tommyettinger:
Both of them independently got it wrong! We use `Long.numberOfLeadingZeros(mask);` and they tested `Integer.numberOfLeadingZeros(mask);`, where the Long one is equivalent to the `+ 32` version so recommended at Reddit. I'll investigate the rest, though!

tommyettinger:
Also, Soonil's post isn't using valid Java code, and I'm not able to reproduce his basic math in any kind of tests, so... take it with a grain of salt. Just try to run `return (int)(key * 0x9E3779B97F4A7C15L >>> Integer.countLeadingZeros(size - 1));` with appropriate key and size from the blog post and you'll immediately see Soonil didn't run that code. Integer.countLeadingZeros is't a function, for one thing, it's numberOfLeadingZeros, and we also use the one in Long, not Integer.

tommyettinger:
The criticism about collisions remaining collisions in smaller hash tables is valid, assuming you are copying into a map with a smaller capacity (our copy constructor doesn't do that). I have a way to get around it if you want to try... If you XOR the current `mask` with the `hashCode()` or `item` before multiplying by 0x9E3779B97F4A7C15L, then collisions at a larger size won't still be collisions at a smaller size. The `mask` changes on its highest set bits only when putting keys into a smaller hash table, so it will affect the high-order bits of the result (the ones we use).

EDIT: I benchmarked this in my [assorted-benchmarks](https://github.com/tommyettinger/assorted-benchmarks/blob/e0913597a9cf4bba66f674b8d5e9b7dd6ada64a8/jmh/GDX_JDGGDXDS_IntIntMaps.txt) repo, and the "CHANGING" map using a XOR with `mask` performs nearly identically to libGDX currently and jdkgdxds' last stable release. There's essentially no performance penalty, it looks like, and it may help in some rare adversarial conditions. It only would matter when put() is used sequentially, not putAll(). That's because putAll() calls ensureCapacity(), so it would never place keys from a large table into a smaller table. The XOR with `mask` is also a lot cheaper than the other mixes int-based maps tried in the blog post, many of which used multiple steps.

The reason I prefer using the high-order bits to the low-order bits wasn't mentioned on Reddit or the blog post, but Vector2.hashCode() collides extremely often on its low-order bits if the Vector2 items contain mostly integers or "nice power of two fractions" like 3f/4f and 9f/8f. It does not collide on its high-order bits because of how Float.hashCode() works. The actual case that got me interested in the Cuckoo hashing problem was the Vector2 case, which was reproducible in games.

tommyettinger:
If we're revisiting this, I'd encourage adding a super-sourced `imul` to NumberUtils (I have one in my library `digital`, which copies and expands on NumberUtils in its BitConversion class). Then we wouldn't need a long multiply, which is slower on both GWT and to a lesser extent desktop, and we could XOR with mask before multiplying by 0x9E3779B9 with imul. The advantage to using mask instead of a random value is that the iteration order will be predictable between two equal-sized (and equal-capacity) maps, rather than completely unpredictable, which is harder to debug. The iteration order is already allowed to change whenever an item is added or removed, but shouldn't change during iteration, of course.

sooniln:
Leaving a drive-by comment since Nate linked this from the reddit thread:

Putting my professional hat on, what is the core value proposition of these maps? Is it avoiding GC pauses? Saving memory? Get/put performance? How important is it to resist adversarial attacks? Is there actually a problem worth solving here, given that the naiveCopy performance doesn't appear to have caused any real-world issues so far (and it's generally avoidable in most situations)?

If GC is the most important, is it worth exploring extendible hashing, which makes rehashing an incremental operation rather than a stop the world large re-alloc? Roman Leventov (author of Koloboke) wrote SmoothieMap as an extendible hashing  follow-up for his HFT needs, so that could be an interesting starting point. That will likely reduce CPU performance however. How important is performance actually - is this map used anywhere where a couple extra nanoseconds of put/get time will actually make a realistic difference? It's quite rare for things to be that CPU bound these days and I might bet it'd be unnoticeable (then again, I don't work in the web space at all). If adversarial attacks are an problem (this issue mentions them as part of the original rationale for moving away from the cuckoo hashing implementation, but it's unclear to me if that actually happened or was just theoretical), then a per-map random seed is the safest way to go. A sufficiently dedicated adversary could probably figure out the random seed with enough time and dedicated effort, but it's a pretty larger barrier, and I don't know that the market for targeting games that use libgdx is that large.

Taking off my professional hat, and wanting to just design a cool, fast hashmap, which is much more fun, there are a couple interesting areas to explore:

@tommyettinger is completely correct that it's preferable to use the high-order bits vs the low-order bits. After the phi-multiplication, the high order bits correspond the start of the fractional result, where there is much more entropy than in the lower order bits. Multiplication generally tends to push entropy into higher bits in my understanding. If we check out Koloboke (and that whole family of hashtables, including fastutil, etc), they use a finalizer that does an xor-right-shift after the phi multiplication to try and fold entropy from the high bits down into the low bits. This uses a constant right shift (naiveCopy doesn't degrade as pathologically), but it doesn't perform as well as the size based shift you're using, which accesses better entropy.

My initial solution in the FastCollect library was bit reversal. If you simply reverse the bit order after the phi-multiplication, now all the entropy is located in the lower bits, exactly where you want it, and no shift is necessary at all. The problem is that while bit-reversal is a single cycle instruction (RBIT) on arm, there is no instruction for it on x86-64. Usually it's implemented with a divide-and-conquer that finishes with a single byte reversal instruction (10s of cycles IIRC). This still actually performs reasonably well in benchmarking. Since I wanted FastCollect to look better in benchmarking lol, and since I wasn't benchmarking on arm, I went with a slightly different alternative, xor-right-shift to premix bits, and then byte reversal (single instruction on all platforms). With that in mind, what are the most important targets for you? I would guess that performance is perhaps more important on arm (where you're likely targeting lower power Android devices?) vs desktop x64 machines, but you guys would know far better than me. I'm also biased, I'd love if you implemented a bit-reversal hash on arm so that I could validate how well it worked without having to do all the benchmarking myself haha.

All that said, since it sounds like you're targeting JS via GWT as well - I have absolutely no clue on what the performance implications of any of this are in JS, since I've managed to completely avoid JS thus far.

Re the idea about xor'ing with mask above - it's unclear to me how that would change the hash? The mask is always set to all 1's for every bit that is used to calculate the slot (since you're later ANDing with the mask anyways), right? In which case, this effectively wouldn't result in changing the slot at different sizes if I'm understanding the proposal correctly?

tommyettinger:
I'll double-check my math, I've been testing in jshell and it's entirely possible I missed something.

```
// Here, I'm creating a mapping of slots in a 64-capacity map to lists of the potentially many keys out of 1024 that hash to those slots.
// This only actually needs a HashMap with capacity 64, but I wanted to be able to inverse the mapping possibly.
jshell> HashMap<Integer, ArrayList<Integer>> large = new HashMap<>(1024)
large ==> {}

// We start with mask = 64 - 1 and fill up `large` with the slots as keys and the hashCode() results (or int items for an int-int map) as the values in ArrayLists per slot.
// I'm using 0xFAB9E45B as the multiplier here because it performed well in a pretty-exhaustive empirical test, but I haven't tested it against other multipliers in this case yet.
jshell> int mask = 64 - 1; for(int hash = 0; hash < 1024; hash++) large.computeIfAbsent((hash ^ mask) * 0xFAB9E45B >>> Long.numberOfLeadingZeros(mask), k -> new ArrayList<Integer>()).add(hash);
mask ==> 63

// I wanted to find the list of int items that hit the most collisions in one slot. That's result $143 .
jshell> ArrayList<ArrayList<Integer>> sorted = new ArrayList<>(large.values()); sorted.sort(Comparator.comparing(ArrayList::size)); sorted.get(sorted.size() - 1)
sorted ==> [[15, 63, 94, 174, 205, 253, 284, 443, 474, 584,  ... 822, 853, 932, 963, 1012]]
$143 ==> [21, 100, 131, 180, 211, 290, 321, 370, 480, 639, 670, 749, 780, 829, 860, 939, 970, 1019]

// I store the full group of colliding items as `problems`.
jshell> Integer[] problems = new Integer[]{21, 100, 131, 180, 211, 290, 321, 370, 480, 639, 670, 749, 780, 829, 860, 939, 970, 1019}
problems ==> Integer[18] { 21, 100, 131, 180, 211, 290, 321, 3 ... 829, 860, 939, 970, 1019 }

// As if we have a smaller table, I set mask to 16 - 1.
// I then run the same mix with the new, smaller mask, which also entails a different shift.
// There's only 4 separate slots out of 16 this can mix to, but we also shrank by 2 bits, so this isn't the worst outcome (or best, really).
jshell> mask = 16 - 1; for(Integer hash : problems) System.out.println(hash + ": " + ((hash ^ mask) * 0xFAB9E45B >>> Long.numberOfLeadingZeros(mask)));
mask ==> 15
21: 7
100: 12
131: 1
180: 2
211: 7
290: 12
321: 1
370: 2
480: 12
639: 2
670: 7
749: 12
780: 1
829: 2
860: 7
939: 12
970: 1
1019: 2
```

So yes, the slot changes, though not as frequently as I'd like, when the map resizes, which is permitted since resizing capacity is allowed to change iteration order. Really, anything that can't happen during iteration can change iteration order for an unordered map, since the order is undefined.

Several reasons we had to do this how we did are related to GWT limitations, which since your FastCollect uses Kotlin, it could thankfully never encounter (GWT translates pure-Java source code to JS). We use `long` math because GWT turns int multiplications into non-wrapping, precision-losing math on JS Number types, and even though `long` math is slower, it behaves the same on all platforms. More recently, `imul` has become widely available in browsers, and we could be using that to speed up map and set hashing. `imul` acts like normal overflowing int multiplication on 32-bit ints, but accepts Java Numbers and coerces them to 32-bit ints first. I did this in my [Delongator repo](https://github.com/tommyettinger/delongator/blob/9f5fcaa611349ccbee7bcd447df93eca138e9158/src/main/java/com/github/tommyettinger/delongator/emu/com/badlogic/gdx/utils/IntIntMap.java#L109-L111) at one point. Delongator currently uses bitwise rotations and XORs between them, masking the low bits, which is an OK solution but could be better.

Whether we're concerned about adversarial insertion is... well, the Cuckoo hashed collections barely needed any level of adversary to use up all available RAM, since 40-something colliding keys would make it need multiple GB of RAM and keep trying to use more. That level of ease of exploitation, and how bad the result was, could be accomplished without needing any kind of expertise or distributed effort. Quadratic insertion time is bad on a shared server getting a DOS attack, sure, but libGDX collections are all not thread-safe and there's probably different priorities you'd have on a server anyway. Linear probing gives us more resistance to collisions that we had before, which was basically no resistance with Cuckoo hashing.

I'd personally rather have a proven and well-understood hash table design that I can understand, than Cuckoo hashing, which as far as I can tell is documented by a bunch of academic papers that gloss over the failure condition as some kind of minor detail. That said, Robin Hood hashing is interesting and I've looked into it but haven't found significant benefits in my implementation, which was pretty simplistic.

I have a benchmark here but haven't really looked into improving my Robin Hood hashing after I found how slow the populate benchmark was for it. https://github.com/tommyettinger/assorted-benchmarks/blob/5536e1c1a28cfc14e4462ceb42e6f1759209ad96/jmh/MapInt2IntAll.txt JDKGDXDS_I2I_CHANGING is the one I proposed using the mask XOR, GDX_I2I_HASH is the libGDX 1.14.2 IntIntMap, and there's my attempt at a RobinHood I2I hash at the bottom. There's some curious results here; the jdkgdxds Ordered map has over an order of magnitude faster key iteration than fastutil's LinkedHashMap, which really surprised me. jdkgdxds uses an extra IntList (primitive-specialized array-backed list) to store keys, which is much faster to iterate than the linked-list-based approach in fastutil. Deletion is much slower in an array-backed list, at O(n), though, so it gives up some generality. You can sort or shuffle an OrderedMap though, which is nice, and I don't think fastutil permits that. I'd like to add fastcollect to the benchmark if I can run it using comparable Java APIs to what I'm currently using. It would certainly be good to compare a well-optimized Robin Hood hashed map to my not-well-optimized one to see where it went wrong!

sooniln:
Ah I see, you're applying the xor-mask before the phi-multiplication, that makes sense - I had wrongly assumed you meant to apply it after the multiplication. Given the low entropy nature of the mask, I'm not surprised it only has a small effect on the final slot, though maybe that will have a larger effect than I'm expecting on the final timings.

Thanks for explaining the JS constraints, I'm very unfamiliar with all the problems there and it sounds very annoying to deal with 😅. Kotlin is compatible with Java so if you take a dep on 'io.github.sooniln:fastcollect-kotlin-jvm:2.0.2' and invoke the APIs from java it should look pretty much just as you'd expect if it were java code instead. Occasionally Kotlin will mangle API shapes that can't be easily expressed in Java, but I don't think that should be much of an issue here as everything should be pretty straightforward. Feel free to ping me if anything causes problems.

tommyettinger:
I think I have a better approach to using `mask` now; I can still change the XORed constant to a different value at each size, but I don't have to just change one bit per capacity increase.

```
jshell> int[] HASH_MULTIPLIERS = new int[]{   0xF0CFFC71, 0x1F2BD44D, 0xC143F257, 0x115583CF, 0xA4613217, 0x9645256D, 0xF0EE34C7, 0xF6C4F429,   0x854066D1, 0x9104FADF, 0x93A04925, 0x0AEF4329, 0xF58C4C35, 0x31A6C2EB, 0xE5182F73, 0xB26FABE5,   0xB520960B, 0x570B3F85, 0x83657DE7, 0x9980FAB9, 0x2F299C91, 0xB423727D, 0xAA333A2D, 0x8AE04AAD,   0x288FAD91, 0xD90AC247, 0xC5F768E7, 0x92317571, 0xD5FA5B15, 0xDB6B35F7, 0xCC8965C7, 0xE0503E4F,  };
HASH_MULTIPLIERS ==> int[32] { -254804879, 522966093, -1052511657, 290 ... , -863410745, -531612081 }

jshell> HashMap<Integer, ArrayList<Integer>> large = new HashMap<>(1024)
large ==> {}

jshell> int mask = 64 - 1, shift = 6; for(int hash = 0; hash < 1024; hash++) large.computeIfAbsent((hash ^ HASH_MULTIPLIERS[shift]) * 0xFAB9E45B >>> Long.numberOfLeadingZeros(mask), k -> new ArrayList<Integer>()).add(hash);
mask ==> 63
shift ==> 6

jshell> ArrayList<ArrayList<Integer>> sorted = new ArrayList<>(large.values()); sorted.sort(Comparator.comparing(ArrayList::size)); sorted.get(sorted.size() - 1)
sorted ==> [[19, 100, 133, 180, 213, 304, 337, 451, 498, 557 ...  751, 794, 874, 955, 988]]
$155 ==> [11, 58, 91, 172, 205, 329, 376, 409, 490, 549, 582, 629, 743, 786, 867, 900, 947, 980]

jshell> Integer[] problems = new Integer[]{11, 58, 91, 172, 205, 329, 376, 409, 490, 549, 582, 629, 743, 786, 867, 900, 947, 980}
problems ==> Integer[18] { 11, 58, 91, 172, 205, 329, 376, 409 ...  786, 867, 900, 947, 980 }

jshell> mask = 16 - 1; shift = 4; for(Integer hash : problems) System.out.println(hash + ": " + ((hash ^ HASH_MULTIPLIERS[shift]) * 0xFAB9E45B >>> Long.numberOfLeadingZeros(mask)));
mask ==> 15
shift ==> 4
11: 8
58: 3
91: 9
172: 4
205: 10
329: 14
376: 9
409: 14
490: 10
549: 10
582: 0
629: 10
743: 11
786: 4
867: 0
900: 5
947: 0
980: 6
```

The last step looks better-distributed than with `mask`, right?

I already have HASH_MULTIPLIERS in jdkgdxds, it's just 32 ints in an array so it's almost nothing to fit in cache or in a JAR, and it only needs to get accessed on resize anyway. I didn't store `HASH_MULTIPLIERS[shift]` here for clarity, but I would in a map implementation, since `shift` only changes in `resize()` or at construction, and `HASH_MULTIPLIERS` never changes. The multipliers there were chosen because they have Hamming weights (AKA population counts or number of set bits) between 15 and 17, and that remains true for their multiplicative inverses modulo 2 to the 32, their Gray codes (calculated by `x ^ x >>> 1`), and the Gray codes of their multiplicative inverses. The logic here is to eliminate certain known-bad multipliers like 0xAAAAAAAB, which has a Hamming weight of 17 but a modular multiplicative inverse of 0x00000003 and a gray code of 0xFFFFFFFE. I'm not actually using these 32 constants as multipliers here, though I could. I did use them as multipliers an earlier version of jdkgdxds and there was a small speed penalty for multiplying by a variable, instead of a constant, where one XOR by either seems to cost nearly nothing.

I'm hoping the variable XOR is cheaper to compute than the more-involved, multiple-step mixers with only constants used by libraries like fastutil. It seems to resist the naive reinsertion case much better than before, without requiring bit or byte reversal. The `bswap` operation is definitely fast when available, though I don't think JS has easy access to it except via endianness changes in a TypedArray.

There's also the possibility of using a random value per capacity per map, which could start as a (secure?) random value and would only need a new one, of potentially lower security/higher generation speed, on each resize. However, that would make iteration order different between otherwise identical maps, which I think makes debugging painful.

PR #5901, opened 2020-02-03
Documented HTTP-Methods and added HEAD-Method

Monarchis:
Added and documented some more HTTP-Methods

Just to update the existing ones.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Net.java
+++ b/gdx/src/com/badlogic/gdx/Net.java
@@ -93,16 +93,38 @@ public interface Net {
 
 	/** Provides common HTTP methods to use when creating a {@link HttpRequest}.
 	 * <ul>
-	 * <li>HEAD</li>
-	 * <li>GET</li>
-	 * <li>POST</li>
-	 * <li>PUT</li>
-	 * <li>PATCH</li>
-	 * <li>DELETE</li>
-	 * <li>OPTIONS</li>
-	 * <li>CONNECT</li>
-	 * <li>TRACE</li>
-	 * </ul> */
+	 * <li>
+	 * <b>HEAD</b> Asks for a response identical to that of a GET request but without the response body.
+	 * </li>
+	 * <li>
+	 * <b>GET</b> requests a representation of the specified resource.
+	 * Requests using GET should only retrieve data.
+	 * </li>
+	 * <li>
+	 * <b>POST</b> is used to submit an entity to the specified resource,
+	 * often causing a change in state or side effects on the server.
+	 * </li>
+	 * <li>
+	 * <b>PUT</b> replaces all current representations of the target resource with the request payload.
+	 * </li>
+	 * <li>
+	 * <b>PATCH</b> method is used to apply partial modifications to a resource.
+	 * </li>
+	 * <li>
+	 * <b>DELETE</b> deletes the specified resource.
+	 * </li>
+	 * <li>
+	 * <b>OPTIONS</b> is used to describe the communication options for the target resource.
+	 * </li>
+	 * <li>
+	 * <b>CONNECT</b> establishes a tunnel to the server identified by the target resource.
+	 * </li>
+	 * <li>
+	 * <b>TRACE</b> performs a message loop-back test along the path to the target resource.
+	 * </li>
+	 * </ul>
+	 * @author Monarchis
+	 * */
 	public static interface HttpMethods {
 		/**
 		 * The HEAD method asks for a response identical to that of a GET request,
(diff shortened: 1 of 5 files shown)


fabiitch:
nice !
just need to edit changelog i think.

intrigus:
Do these work on gwt at all?

Monarchis:
@intrigus sur ethis is just adding new fields to a static class that had not all HTTP-Methods as constants before, (all checks passed)

noblemaster:
There is more to it than just adding a few constants. Please look at the code:
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java#L175

SimonIT:
To make it work on gwt, there has to be made some changes for the new methods, if they are supported. Else, it should be documented that they are not supported on gwt
https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java#L129-L145

MobiDevelop:
On GWT, we use RequestBuilder, which supports `DELETE`, `GET`, `HEAD`, `POST`, `PUT`. Any others should be documented as unavailable on GWT.

noblemaster:
Unless GWT can support the other methods (sorry not too familiar?), it probably makes sense to drop this pull request for compatiblity's sake?

Monarchis:
@SimonIT
I just extended the constants, but you are free to implement it.
This Line should tell that it's not supported:
https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtNet.java#L144

Monarchis:
@noblemaster Compatibility does not break

gdx/src/com/badlogic/gdx/Net.java
```diff
@@ -93,20 +93,92 @@
...
+		 * The CONNECT method establishes a tunnel to the server
+		 * identified by the target resource.
+		 * Not Supported by libgdx yet.
+		 **/
+		public static final String CONNECT = "CONNECT";
+
+		/**
+		 * The TRACE method performs a message loop-back test along
+		 * the path to the target resource.
+		 * Not Supported by libgdx yet.
```
MobiDevelop: Adding things only to say they aren't supported isn't super nice.

Monarchis: @MobiDevelop well, then reject this pr

noblemaster: I agree. I think better would be to update Net.java to accept any method string also allowing to specify if output is desired or not. After all, GET or POST could both be used with our without stream output, even though not commonly used for GET as opposed to POST.

Monarchis: @noblemaster I'm abled to add some of that methods easy, but connection and trace need more efford.

Monarchis: Getting Head to work with GWT, but the others are not supported right now, I'll mark it.

noblemaster:
@Monarchis If you want this pull request to continue, please (1) make sure those methods work on all backends. Also, (2) given we add a lot of methods, let's also allow custom HTTP methods in addition to the ones you proposed supporting both ones that do output (ditto POST) and those that don't (ditto GET). I think that make the most sense.

Also, don't forget to (3) write JUnit test for your methods and the custom methods testing them on all backends.

Monarchis:
@noblemaster The NetGwt is using google lib that does not support PATCH; OPTIONS, TRACE and CONNECT so there will be a problem mapping them to the other 5 existing ones, HEAD is possible to implement properly.

Monarchis:
One thing that i can do is marking the one that do not work on GWT as not supported by gtw, and removing the rest that does not work right now.

noblemaster:
I see. In that sense it probably makes little sense to add them besides probably HEAD which is of little use if you ask me. libGDX shouldn't offer methods that are not supported on all backends.

Monarchis:
@noblemaster 
They can be implemented on all Platforms, only gwt has problems.

Monarchis:
ok only HEAD is left as change.

noblemaster:
There is some duplicate code you added in `GwtNet`. Also, what about the unit tests for all the platform to make sure it actually works?

Monarchis:
@noblemaster the only thing i added was the HEAD-Method, and setting the UriConnection.setDoInput() to false if it's head. (GwtNet).
The redundancies, idk maybe i'll remove them.

But currently I'm still not sure if head also needs an Input (the headers that are responded to the client).

May i really have to test this Method, otherwise, nothing else has changed.

MrStahlfelge (approved):
Last change is a good change. It is non-breaking, an enhancement and should be sqash merged.

PR #5902, opened 2020-02-03
Implemented 2nd loop for better HeadlessApplication-Support ...

Monarchis:
..., added `final float delta` to loops

While implementing Libgdx I wondered why there isn't an update loop to keep code spitted in client server applications.

The Major reason I want to announce this is that Headless-Applications need a spitted cycle, one that is updating the game and the other one that is rendering the game on client side.

I added the `final float delta` parameter, because it's needed in most cycle implementations.

update -> updating world positions, etc.
render -> draw them

Pro:
- call of `Gdx.graphics.getDeltaTime()` in different places of the Application is not needed anymore.
- makes it possible to modify the delta for sub calls.
- allows Headless-Applications only use the update()-cycle
- performance increase for headless apps (ensures that no rendering calcs are done)
- splits update code from rendering code.
- reduces redundancy in implementations for Client and Server.

Con:
- old apps upgrading to new gdx version need to implement the new loop, more or less changes needed, depending on implementation.

What do you think about that change?

Some minor Code changes will follow entirely marking all values as final.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -475,7 +475,8 @@ public class AndroidGraphics implements Graphics, Renderer {
 			}
 			app.getInput().processEvents();
 			frameId++;
-			app.getApplicationListener().render();
+			app.getApplicationListener().update(Gdx.graphics.getDeltaTime());
+			app.getApplicationListener().render(Gdx.graphics.getDeltaTime());
 		}
 
 		if (lpause) {
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java
@@ -214,7 +214,8 @@ public final class AndroidGraphicsLiveWallpaper extends AndroidGraphics {
 
 			app.getInput().processEvents();
 			frameId++;
-			app.getApplicationListener().render();
+			app.getApplicationListener().update(Gdx.graphics.getDeltaTime());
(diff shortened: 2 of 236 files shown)


Darkyenus:
:-1: from me. This point of view is too idealized - most games/apps I have worked on either did update and render at the same time (simple case, libGDX's default) or needed special update logic, which is not clearly generalizable and does not fit this PR at all.

This is a huge and breaking change and I don't see any benefit. In the rare case when `update` and `render` is separated and always called right after each other, doing that separation yourself is not a problem. And in each other case, this just introduces a split that can't be honored, which will lead to confusing code.

Monarchis:
@Darkyenus well Stages are already designed by that (splitted rendering and updating). The point is the implementation of the ApplicationListener. It's used for HeadlessApplications and all other.

intrigus:
Breaking every app that is currently existing is not so nice.

Monarchis:
@intrigus and what do you think about just extending the ApllicationListener by this 2 methods with a float value? -> update(float delta) + render(float delta) and leaving render()?

noblemaster:
There are 236 changed files: this is not going to make it as it breaks compatibility for everyone that develops for libGDX (although an easy fix).

It's not that I don't disagree with you, it possibly makes sense to call those methods with the delta time. Another option is to call with the timestamp since epoch or local time since the computer started up in milliseconds. Another option is nanoseconds. There are other values that probably should be passed in. If anything, this should have been decided at the start of libGDX, not now.

Monarchis:
@noblemaster round about 200 of them are test classes ...

Tom-Ski:
If you want this behaviour, it takes two seconds it to add it to your app. Render is just the entry point for a render call, its up to you what you want to do with it. Splitting these up doesn't improve performance, and leads to ambiguity about the pipeline.

Monarchis:
@Tom-Ski, i know that, it takes not that long. I'll show you what I mentioned in the next pr. It wont break changes or and won't lead to ambiguity.

I'll implement it configurable, for those who need to have those 2 loops spitted.

PR #5903, opened 2020-02-03
WIP: Loops the 2nd, this won't break any Changes

Monarchis:
I know that my old pr #5902  was an overkill. Especially in breaking changes.

- Now it's implemented in that way, that all platforms will keep their current behaving and ensure a better server-client architecture for games need to be run on both sides.
- This reduces the strong difference in implementing the Screen, Game, ApplicationListener classes.
- They are now applicable for the use with HealessApplication and <Other>Application.
- This leads to simpler implementation with synchronous running instances of the game, communication with each other. There is not much afford needed (if implemented as intended) to run the game in Headless-Mode as a Server.

Another discussion would be:
> When adding new loop methods, i could improve this pr to use the update loop multi threaded with additionally an `int threadIndex` param.
> The number of cores would be configurable via Application Configuration. So there is a possibility to use the update-loop for updating game objects multi threaded, for client and server it would be a hard improvement in Object updating and would increase the possibilities of libgdx from ground.

So give it a try.
Regards.

PS: if yes i can put in some extra afford, implementing the multi-threaded variant, before a normal update loop is getting default for most apps (Reason for WIP).

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -475,6 +475,8 @@ public class AndroidGraphics implements Graphics, Renderer {
 			}
 			app.getInput().processEvents();
 			frameId++;
+			app.getApplicationListener().update(Gdx.graphics.getDeltaTime());
+			app.getApplicationListener().render(Gdx.graphics.getDeltaTime());
 			app.getApplicationListener().render();
 		}
 
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphicsLiveWallpaper.java
@@ -214,6 +214,8 @@ public final class AndroidGraphicsLiveWallpaper extends AndroidGraphics {
 
 			app.getInput().processEvents();
 			frameId++;
+			app.getApplicationListener().update(Gdx.graphics.getDeltaTime());
+			app.getApplicationListener().render(Gdx.graphics.getDeltaTime());
 			app.getApplicationListener().render();
 		}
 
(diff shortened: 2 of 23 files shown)


MobiDevelop:
This still requires everyone who updates libgdx to add additional methods they may not need or want and (in my opinion) is confusing with there being two render methods.

Monarchis:
@MobiDevelop the ApplicationListenerAdapter should to it's job, if extended.

MobiDevelop:
Not everyone uses ApplicationListenerAdapter...

Monarchis:
To avoid confusion I can improve the docs more than now.
the old render()-loop is marked as deprecated, and documented why.

mgsx-dev:
i think that most comments not in favor to this kind of changes in #5902 still apply here.

Also, marking method as Deprecated means it'll be removed later, which won't happens: it simply differs breaking change.

I don't understand the benefits of this kind of changes. for me it breaks separation of concerns : 
* application user code decide which deltaTime to use (some prefer static deltaTime, some prefer clamp it, some prefer calculate it differently, etc). 
* user code decide if split is preferable and how to split their render process : could be update(delta)/render, could be update(delta)/render(delta), etc. Some may implement multi render pass and/or multi update pass.

I'm not sure what you want to do regarding multi threading, but most libgdx is not thread safe and multi-threading is not supported for Web build. User code should decide to implement or not multi threading and how (separation of concerns again).

it might be better to discuss this kind of changes via irc/discord prior to put a lot of efforts in such PR.

Monarchis:
@mgsx-dev Thanks for that answer it told me what the reason for rejecting is, the current code is more or less sampled.
I know that most classes in gdx are not multi save, especially rendering is definitive not, but that does not mean that you cant use multi-threading.

The fact of self implemented delta time by user is new to me. So there is also the possibility  to remove those float values.

Maybe a suggestion about a decorator pattern for different types of ApplicationListeners could be a possibility to enlarge flexibility of ApplicationSetup / Loop design.

mgsx-dev:
@Monarchis see https://libgdx.badlogicgames.com/community.html :-)

Monarchis:
Closed for now, will show an example.

PR #5906, opened 2020-02-06
Remove unnecessary code for android min sdk <14.

PokeMMO:
The min sdk level was changed to 14 in #5787
This PR removes now unnecessary code for api levels lower than the current min of 14.

There are three main parts split into 3 commits.
- Miscellaneous api level checks.
- Removing now unnecessary GLSurfaceViewAPI18/GLSurfaceView20API18 classes.
- Merging AndroidInputThreePlus into AndroidInput.


noblemaster:
+1 to have this added!

PokeMMO:
I wont have time to fully test for a few days, but otherwise it should be good.

PokeMMO:
That was a long few days. Tested as much as I could and nothing seems to have broken.

PR #5909, opened 2020-02-08
Add watch face implementation for wear os in android backend

liangchenhe55:
[Wear OS by Google](https://wearos.google.com/#hands-free-help) is an operation system designed for smart watches based on Android. Developers can build a watch face application which drive from `WallpaperService` but has some special functions and APIs.

So `AndroidLiveWallpaperService` already in libGDX not suitable for watch face, I create a new implementation  for that. Also I added a test to make sure it works well.

Tip: watch face only available on Wear OS devices, if you want to test it without real device, create an virtual device in Android Studio and select `Wear OS`.

Once install the test apk, long press the current watch face, Click `See more watch faces` and select `GDX Test`.

I have signed the CLAs.


MobiDevelop:
Hmm, not sure about this. Seems like it would be pretty rarely used.

liangchenhe55:
> Hmm, not sure about this. Seems like it would be pretty rarely used.

Sorry for the late reply. In fact, the watch face is a scene that is very suitable for using libgdx. At the same time, the watch face is also a very important part of the smart watch with high user activity.

The CPU of a watch device is very limited, using libgdx to manipulate GPU can greatly improve performance. Developers who are not familiar with android development can make beautiful watch faces.

And the idea is not alone, see [this post](https://stackoverflow.com/questions/37976297/how-can-i-convert-an-android-live-wallpaper-to-a-watch-face-for-wear) on stackoverflow. Since the watch face is an extension of the wallpaper(LWP) which is already a stablepart of libgdx, this new implementation is natural and does not change the original architecture.

Indeed, smart watches are not as widely used as mobile phones, as mentioned above this feature has no additional cost and I think this can promote the development of the community to a certain extent.

benoitletondor:
Actually had this exact need for a project. Copied your file while hoping this gets merged. Thank you for your work!

SimonIT:
You have to add backends/gdx-backend-android/libs/wearable-2.6.0.jar also in backends/gdx-backend-android/pom.xml. The past releases were done with maven.

I really like the idea of having the possibility to create watch faces with libgdx.
I tried your added example in an emulator and it looks really cool. Sadly I wasn't able to test it on my real watch because it supports only bluetooth debugging and the installation process took to long.

MrStahlfelge:
I also like the idea, but I don't like another dependency needed for libgdx that's not needed for 99.9% of the projects. 

I would prefer a way where all the real watch face specifics - that is, if I have seen correctly, only `WearOsWatchFaceService` anyway - are in an own lib as an example project with appropriate documentation, and we adpot all changes needed to the Android backend for it to work (that is, if I have seen correctly, mainly `getSurfaceHolder`)

This way, we can have Watch Face Support without pulling in another dependency.

liangchenhe55:
@MrStahlfelge Maybe I didn’t explain it clearly. To create a watch face, we must create a class to extend `WatchFaceService` which was provided by `wearable-2.6.0.jar` by Google. Unlike Android, Wear OS is not open sourced, so we'd better use official jar instead of extracting the watch specific api.

Of course, we can use `compileOnly` so that this jar will not be packaged into the artifact. Developers need to explicitly depend on it if necessary. In this way, it has no effect on the other 99.9% of the projects.

liangchenhe55:
@SimonIT Thanks. I have added it to `pom.xml` and fixed the conflicts. For simplicity, I compressed commits and forced push.

The watch face test is packaged with the whole android test project so the final apk file will be a little big. You'd better use a watch with a usb interface or wifi adb. The bluetooth is too slow.

MrStahlfelge:
Closed because of the given reasons
* will be used rarely
* we don't want to pull in new dependencies for this

Feel free to open a PR again with needed changes for a lib/example project to get this to work.

Issue #5908, opened 2020-02-07
[Bug] TextField Tab-Focus works different on Unix systems, reacts on ENTER-Button


Monarchis:
These 2 lines define not correctly the ENTER button of Desktop/Android systems:
https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L67-L68
Unix: \n
Windows: \r\n (so \r is ok for windows)
Mac: \n (old Macs \r)
Android: \n

This Line needs an additional check if it's an android system:
https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L1027-L1028
otherwise, the ENTER button on Unix-Desktop systems will have the same behavior like the TAB button.
Something like:
`if ( (character == TAB || (isAndroid && character == ENTER_ANDROID)) ) && focusTraversal)`
would solve the problem.

cpbeto:
This also affects `TextArea` which extends `TextField`.

On the Linux system I'm working on, pressing the ENTER key fails to add a new line (a hotfix I'm using is doing `setFocusTraversal(false)`). I'm guessing the same is true for all Unix systems, even Android.

Perhaps we could split up the conditional and distinguish between TAB and ENTER.`TextArea` constructor could disable focus traversal for the ENTER key.

Monarchis:
#5910 continues this. /closed

Monarchis:
Approved by #5910

PR #5910, opened 2020-02-08
[Fix/Addition] #5908, implemented correct TAB/ENTER behavior of TextField and TextArea / Added OS constants in UIUtils.

Monarchis:
~~I've not tested this commit, but it should work, so please review.~~
see #5908 

The changes are:
* Corrected naming of the key-char-constants in TextField and changed their  visibility to `protected`.
* Added all platforms checks to UIUtils, adopted from SharedLibraryLoader.
* TextField: All mobile versions use ENTER or TAB for focus traversal now.
* TextField: All desktop versions use TAB for focus traversal now.
* -~~TextArea (multi-line-TextField) initializes by default `focusTraversal = false;`.~~
* +TextArea uses ENTER for a new Line on all platforms now and TAB for traversal.
* Adds the tests TextAreaTest and TextAreaTest2 to the GWT-test module. Thanks to @tommyettinger 

Now this should be the correct behavior.

Affected platforms
- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

PS: I don't know what's happening with the ENTER on Desktop, normally it should fire a submit-event, but seems like it does nothing now.

PPS: Test in progress, feel free to pic an os out of https://github.com/libgdx/libgdx/pull/5910#issuecomment-584571284 and test please.


SimonIT:
You have to change also the emulated files for gwt, like https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/scenes/scene2d/utils/UIUtils.java

Monarchis:
@SimonIT yep, WIP

Monarchis:
Wow ... what a mess ...
https://stackoverflow.com/a/19883965
Some possible values to detect the OS in GWT.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextArea.java
```diff
@@ -73,6 +72,7 @@ protected void initialize () {
 		firstLineShowing = 0;
 		moveOffset = -1;
 		linesShowing = 0;
+		focusTraversal = false;
```
petoncle: Any reason for disabling focus traversal by default in TextArea?

Darkyenus:
> Some possible values to detect the OS in GWT.

Do you need to detect OS? Isn't the fact that it is GWT (and therefore web) enough? Since web is mostly homogeneous across OSes, it could be treated as its own platform, at least from the UI perspective.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextArea.java
cpbeto: I'm thinking TextArea could provide focus traversal with TAB while ENTER creates a new line. That's how it appears to be on HTML / web pages.
I wouldn't know what is the default behaviour on mobile devices (specially regarding the ENTER key).

Monarchis:
@Darkyenus yes, read https://github.com/libgdx/libgdx/pull/5910#issue-372768626 ... There is a different handling of inputs on mobile devices

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextArea.java
Monarchis: Ok seems like i did not remind that.
Need to improve this, just implemented stupid the focus traversal suggestion in Issue #5908 

The default behaving of a TextArea on mobile is ENTER = new Line, TAB does not really exist.
On Desktops ENTER = new Line, TAB = focus traversal.
Need to check if it's mobile or desktop.

Monarchis:
Tested, works now below Linux like intended.
Other platforms not Tested (i currently have device to achieve this)
But depending on the changes and the Key-Character, all platforms should work now.

See PR main post (edited).

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -64,9 +63,11 @@
...
-	static protected final char ENTER_DESKTOP = '\r';
-	static protected final char ENTER_ANDROID = '\n';
-	static private final char TAB = '\t';
+	//Windows and Mac before OS X as Enter
+	static protected final char CARRIAGE_RETURN = '\r';
+	//Unix based systems as Enter
+	static protected final char NEWLINE = '\n';
+	static protected final char TAB = '\t';
 	static private final char DELETE = 127;
 	static private final char BULLET = 149;
```
tommyettinger: It's a little odd that these mix `protected` and `private`. Even though I can see why some are protected, I don't know why all of these chars aren't protected; could they be used in subclasses? Particularly backspace and delete.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/UIUtils.java
```diff
@@ -6,9 +6,12 @@
...
 public class UIUtils {
-	static public boolean isMac = System.getProperty("os.name").contains("OS X");
-	static public boolean isWindows = System.getProperty("os.name").contains("Windows");
-	static public boolean isLinux = System.getProperty("os.name").contains("Linux");
+	//Adapted system checks from com.badlogic.gdx.utils.SharedLibraryLoader
+	static public boolean isAndroid = System.getProperty("java.runtime.name").contains("Android");
+	static public boolean isMac = !isAndroid && System.getProperty("os.name").contains("OS X");
+	static public boolean isWindows = !isAndroid && System.getProperty("os.name").contains("Windows");
+	static public boolean isLinux = !isAndroid && System.getProperty("os.name").contains("Linux");
+	static public boolean isIos = !isAndroid && ("iOS".equals(System.getProperty("moe.platform.name")) || !(isWindows || isLinux || isMac));
```
tommyettinger: If this works, which I don't have much reason to doubt, this should be a valuable addition to UIUtils.

Monarchis: @tommyettinger 

https://github.com/libgdx/libgdx/blob/11a53e2d9d573e2a0a94457b309dbf7044623497/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L36-L67

Yes it works.

Monarchis: Ok got something! "Mac" and  "OS X" damn.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
Monarchis: @tommyettinger Yes, they could be reused in subclasses, the3 protected are used in TextArea, but not DELETE AND BULLET.
this was implemented already, i did not change them.

Monarchis: @tommyettinger well, just did it now.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextArea.java
Monarchis: Removed `focusTraversal = false;` again, solved it in a better way.

Monarchis:
All right, followings testings are open':
* ~~Linux -> @Monarchis~~
* Windows
* Mac
* Android
* IOs
* ~~GWT -> @tommyettinger~~

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -1006,15 +1007,21 @@ public boolean keyUp (InputEvent event, int keycode) {
 			return true;
 		}
 
+		protected boolean checkFocusTraverse(char character) {
+			return focusTraversal && (character == TAB || (
+							(character == CARRIAGE_RETURN || character == NEWLINE)
+							&& (UIUtils.isAndroid || UIUtils.isIos) ));
+		}
+
```
petoncle: I'm not a big fan of this method, especially since it uses the field TextField.focusTraversal, and allows an overriding subclass to ignore the value of this field?

Monarchis: @petoncle that is why i implemented it, see TextArea -> other focus traversal behaving.

petoncle: Just sharing an idea, not sure if it's any good: could we maybe have a field `TextField.focusTraversalCharacters`, and populate it in the constructor of TextField like:
```java
focusTraversalCharacters.add(TAB);
if (UIUtils.isAndroid || UIUtils.isIos) {
	focusTraversalCharacters.add(CARRIAGE_RETURN);
	focusTraversalCharacters.add(NEWLINE);
}
```

Monarchis: @petoncle, yes that an idea, but the thing is: normally focus traversal depends only those few chars, i think it doesn't make much sense to overextend this.

This would lead to using smileys or something else for focus traverse by adding chars to that list. But i think nobody will implement that :D

petoncle: Maybe `isFocusTraversalCharacter()`?

Monarchis: @petoncle, yes that's good, but then i have to remove the focusTraversal-Flag from the method.

Darkyenus: How about `shouldTraverseFocusOn(char)`?

Monarchis: Ok I think I just let it just checkFocusTraverse(char) or checkFocusTraversal(char) that's the reason for my original question. Currently it's called checkFocusTraverse(char).

But thx for the suggestions ;)

tommyettinger:
Whoops, I accidentally merged the current libGDX upstream into this PR, I hope I didn't break anything; feel free to revert #fb5d83d or #91f8928 if there are issues. GitHub Desktop doesn't tell you which `libgdx` fork you're on when you start using it...

Monarchis:
@tommyettinger well seems like nothing changed, but i'm wondering why....

There are some changes in the merge, but the diffs to libgdx/master don't show anything else then the previous changes.

This is the reason, why i'm only using the command-line tool (requires more attendance basically) of git and no gui to do my changes ;)

NathanSweet:
FWIW, SmartGit is a great UI for using Git. The command line has it's place, but SmartGit makes needing it very rare and IMO is much better for diffs.

tommyettinger:
Heh, I wonder if this is related to the big Github error rate issue today and yesterday, but the merge apparently marked a commit on my commit graph (unexpected but woohoo!), and the commit is invisible ( https://github.com/libgdx/libgdx/commits?author=tommyettinger&since=2020-05-22&until=2020-05-23 ). I don't really understand what affects Git history, still...

The OS constants should be a really nice added feature here, on top of a bugfix. :tada: Thanks for merging!

Monarchis:
@tommyettinger if you blame the lines then it's commited by me and you together, don't know how this is possible.

Issue #5911, opened 2020-02-11
Lack of information about obtaining nightly builds via gradle


thechubbypanda:
Please ensure you have given all the following requested information in your report.

#### Issue details
Lack of information about obtaining nightly builds via gradle.
I cannot find the sonatype nightly repo anywhere to put into a maven repository for gradle.
Perhaps we need to be building to a repository every day if we're not?

#### Version of LibGDX and/or relevant dependencies
Latest

#### Please select the affected platforms
- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

SimonIT:
Read here https://github.com/libgdx/libgdx/wiki/Updating-LibGDX#switching-libgdx-versions

thechubbypanda:
I have read that page, it does not give any URLs of snapshot repositories or anything relating.
Attempting to obtain `1.9.11-SNAPSHOT` from maven central returns an error.

Darkyenus:
I agree with this, the libgdx maven coordinates (and repo url for snapshots) should be more visible. I know that it is somewhere on the wiki (or at least it was), but it should be more visible. We could even add a maven badge to the readme or to the wiki.

Also it seems that the link in the README for "Working with the Source" is half dead - it points to https://github.com/libgdx/libgdx/wiki/Running-demos-%26-tests which is supposedly deprecated and links to https://libgdx.badlogicgames.com/documentation/hacking/Working%20from%20source.html. I didn't even know that there is documentation on badlogicgames.com.

Monarchis:
You also could use (example from https://blog.gradle.org/introducing-source-dependencies):
```
dependencies {
    implementation 'org.gradle.cpp-samples:utilities:1.0'
}
```
and
```
sourceControl {
    gitRepository("https://github.com/gradle/native-samples-cpp-library.git") {
        producesModule("org.gradle.cpp-samples:utilities:1.0")
    }
}
```

MobiDevelop:
Doesn't the project setup include the snapshot repo?

Monarchis:
@MobiDevelop not anymore.

![gdx-setup](https://user-images.githubusercontent.com/2159235/74277232-71f9bc00-4d17-11ea-9f25-3276a058f11f.jpg)

mgsx-dev:
[...]

PR #5912, opened 2020-02-11
Add version badges to readme

SimonIT:
see #5911

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 [![Jenkins build status](https://libgdx.badlogicgames.com/jenkins/buildStatus/icon?job=libgdx&.png)](https://libgdx.badlogicgames.com/jenkins/job/libgdx/) (Jenkins)
 [![Travis build status](https://travis-ci.org/libgdx/libgdx.svg?branch=master)](https://travis-ci.org/libgdx/libgdx) (Travis)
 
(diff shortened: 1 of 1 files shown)


mgsx-dev:
why not make badges as links, pointing to https://oss.sonatype.org/ ?

SimonIT:
Is there a detail view for a dependency? Linking the start page is in my opinion useless

mgsx-dev:
I don't think so, but it's better than linking to badge image (what i get whne i click badges). 
#5911 was about "I cannot find the sonatype nightly repo", i thought a link to sonatype oss repos could be hepful.

SimonIT:
Okay, will add thaf

Darkyenus:
The release badge can point to `https://search.maven.org/search?q=com.badlogicgames.gdx` or more specifically to `https://search.maven.org/search?q=g:com.badlogicgames.gdx+a:gdx`.
Maven.org has a nice ["artifact homepage"](https://search.maven.org/artifact/com.badlogicgames.gdx/gdx/1.9.10/jar), but it is specific for a version, which is not so nice.

The snapshot badge could point to `https://oss.sonatype.org/#nexus-search;gav~com.badlogicgames.gdx~gdx~~~~kw,versionexpand`. It is more useful than nothing, but it incorrectly lists 1.9.9-SNAPSHOT as the latest version and mixes in non-snapshot releases as well. I did not manage to find anything better yet, their UI is surprisingly terrible.

PR #5923, opened 2020-02-27
fix some camera/viewport project/unproject

### LINKED ISSUE #5854: Stage.stageToScreenCoordinates returns incorrect value with y offset in viewport
#### Issue details
Stage.stageToScreenCoordinates returns incorrect value with y offset in viewport

#### Reproduction steps/code

Create a 100x100 window.
Setup a stage with a viewport in the top quarter of the window i.e. width 100, height 25, y=75
stage.stageToScreenCoordinates returns incorrect values.

   ```
 public void create() {
        Viewport viewport = new Viewport() {
            @Override
            public void update(int screenWidth, int screenHeight, boolean centerCamera) {
                assert screenWidth == 100 && screenHeight == 100;
                setScreenBounds(0, 75, 100, 25);
                setWorldSize(100, 25);
                apply(centerCamera);
            }
        };
        viewport.setCamera(new OrthographicCamera());
        Stage stage = new Stage(viewport);

        Vector2 tmpvec = new Vector2();

        tmpvec.set(0, 0);
        stage.getViewport().project(tmpvec);
        System.out.println(tmpvec); // prints (0, 75), correct so far

        tmpvec.set(0, 0);
        stage.stageToScreenCoordinates(tmpvec);
        System.out.println(tmpvec); // prints (0, -50), should be (0, 25)
    }
```

Due to the following line in Stage:

https://github.com/libgdx/libgdx/blob/c0c1f515df54092cb47717f43f022db02cf7e328/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L738

After getting the result of camera.project (in absolute screen coordinates with origin at bottom left), it converts to top left origin using viewport.screenHeight instead of actual window height.

1.9.10

### PR DESCRIPTION
**Pull request by mgsx-dev**
i was trying to understand #5854 issue, so i write a test.

It took me a while to understand that project is not the inverse of unproject, there are some Y flips required in between. I thought it was a bug but reading javadoc carefully i understood it was intentional.

Anyway, during my tests i noticed and fixed few little bugs : 
* camera#unproject had a 1 pixel error (propbably a mistake)
* viewport#toScreenCoordinates is only used in GroupTest but this test doesn't show the issue since it use a ScreenViewport (virtual screen bounds are the same as real screen bounds). My test illustrate the issue.

My PR doesn't fix #5854 (if there is an issue) but i think the test class i added could be a good base to track it down.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Camera.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Camera.java
@@ -194,7 +194,7 @@ public abstract class Camera {
 	public Vector3 unproject (Vector3 screenCoords, float viewportX, float viewportY, float viewportWidth, float viewportHeight) {
 		float x = screenCoords.x, y = screenCoords.y;
 		x = x - viewportX;
-		y = Gdx.graphics.getHeight() - y - 1;
+		y = Gdx.graphics.getHeight() - y;
 		y = y - viewportY;
 		screenCoords.x = (2 * x) / viewportWidth - 1;
 		screenCoords.y = (2 * y) / viewportHeight - 1;
--- a/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java
+++ b/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java
@@ -117,7 +117,7 @@ public abstract class Viewport {
 	public Vector2 toScreenCoordinates (Vector2 worldCoords, Matrix4 transformMatrix) {
 		tmp.set(worldCoords.x, worldCoords.y, 0);
 		tmp.mul(transformMatrix);
-		camera.project(tmp);
+		camera.project(tmp, screenX, screenY, screenWidth, screenHeight);
 		tmp.y = Gdx.graphics.getHeight() - tmp.y;
 		worldCoords.x = tmp.x;
 		worldCoords.y = tmp.y;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by NathanSweet** [comment]
Looks like you are right about Camera `unproject`. Good find! Especially since this code has been in service almost 10 years! [Original code](https://github.com/libgdx/libgdx/commit/2f3f22689477e3e42bcfbc5d1194f5c34e9d0e73#diff-741deba8d8eec185cc2d4b6c01c292a6R101) has the `- 1`, found the root cause in the commit message:

> Works like a charme! So little code for such a nice thing... drunk 5 minutes coding is awesome...

Viewport fix also looks good. Cheers!

**Comment by jusw85** [comment]
Unfortunately, it does not fix #5854 and I don't think the issue should be closed. I can reproduce the issue using your test case.

Instead of 
```
stageViewport = new FitViewport(500, 500);
```
use a custom viewport which has a y-offset
```
stageViewport = new Viewport() {
    @Override
    public void update(int screenWidth, int screenHeight, boolean centerCamera) {
        setScreenBounds(20, 20, screenWidth - 20, screenHeight - 20);
        setWorldSize(500, 500);
        apply(centerCamera);
    }
};
stageViewport.setCamera(new OrthographicCamera());
```

Screenshot from 2020-02-29 03-24-39

Stage reproject gives the wrong value. Notice that only the y value is wrong, the x value is correct.

This is because of the following line in stageToScreenCoordinates, which uses the *viewport* screen height, instead of the *actual* screenheight i.e. `Gdx.graphics.height`

https://github.com/libgdx/libgdx/blob/c0c1f515df54092cb47717f43f022db02cf7e328/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L738

NathanSweet:
Thanks for the explanation and reproduction code, @jusw85 and @mgsx-dev!

PR #5925, opened 2020-02-27
Allow 64k vertices Meshes

mgsx-dev:
This is a **work in progress** PR that allow 64k vertices mesh instead of 32k.

There might be some other impacts like with Bullet, i have to test a bit more. Note that i tested it against desktop and GWT.
I didn't wanted to spread this change to 2D domain (sprite batch and such) for now.
I wanted to have a feedback about this change before going further.

I now that fbx-conv doesn't allow 64k vertices yet but other loaders does (eg. OBJ loader). I needed this change for my [GLTF library](https://github.com/mgsx-dev/gdx-gltf) so i monkey patched Mesh class and it works well so far, so i wanted to share this improvement.

Note that i also fixed an issue in ModelCache using indices count instead of vertices count. I'm not fully satisfied with the fix: we don't know how many vertices are needed without counting indices in mesh part to add. I could isolate this fix in another PR if you think it's prefereable.

I quickly discussed about it with @xoppa recently, i wasn't sure if there was other reasons than java unsigned short for this limitation.

Tell me what do you think about it.


fabiitch:
Its compatible ios, android ? If this limitation is break, Its a really good news

mgsx-dev:
@fabiitch i think so, i can't see any reason why it wouldn't work but i didn't test it on android yet and i can't test on iOS.

waverider404:
Screen Shot 2020-03-28 at 6 31 15 AM
works perfectly fine on android emulator and old android devices

noblemaster:
I think this makes sense to merge unless there are objections? Could someone test on iOS as well?

obigu:
I'm happy to test on iOS, is running the new BigMeshTest enough as a test?

mgsx-dev:
@obigu yes, i think it's enough since it tests both model cache and non cache.

obigu:
I've run BigMeshTest on an iPad Mini 5 and the Simulator (both iOS 13.3) and I get a totally black screen with the following logs https://pastebin.com/tCPk47b3. Just to confirm is not a problem if the tests on iOS instead of your changes, is there an equivalent test I can run that should work without your PR?

mgsx-dev:
i fixed screen clearing in my test (sorry about that), result should looks like this : 
Screenshot 2020-04-21 15:41:46

obigu:
Test also runs without issues on iOS (iPad Mini 5 iOS 13.4.1).

noblemaster:
Great! If this is ready, can you also update the VERSION file with the change and let's get this in! :)

mgsx-dev:
@noblemaster done! feel free to squash and reword if necessary. I checked, there are no breaking changes except that now indices can be negative short.

mgsx-dev:
oops, sorry, i didn't correctly pulled before modifying this file xD

noblemaster:
Great :-D

Issue #5929, opened 2020-03-01
AssetManager: NullPointerException vs isLoaded => true ~ unload() called twice


j3nda:
Please ensure you have given all the following requested information in your report.

#### Issue details
when AssetManager try to unload same resource twice it will fail on NullPointerException, here:
```java
		// get the asset and its type
		Class type = assetTypes.get(fileName);
		if (type == null) throw new GdxRuntimeException("Asset not loaded: " + fileName);

		RefCountedContainer assetRef = assets.get(type).get(fileName); // <<< NullPointerException

		// if it is reference counted, decrement ref count and check if we can really get rid of it.
		assetRef.decRefCount();
```

- on 1st attempt, when resource is loaded: assetManager.isLoaded(resourceName) will return true; and then assetManager.unload(resourceName); is OK!

but

- on 2nd attemt, resource is already unloaded and: somehow assetManager.isLoaded(resourceName) returns TRUE instead of false! so mine code will call again: 
assetManager.unload(resourceName) ~ because it says its loaded, so: next pass is failing on NullPointerException: "RefCountedContainer assetRef = assets.get(type).get(fileName);"
where first get(type) is positive - but second one get(fileName) will return null!

#### Reproduction steps/code
v1.9.10:
1) assetManager.load(any TextureAtlas);
2) ask for isLoaded(textureAtlasName) => true;
3) so unload(TextureAtlasName); => OK
4) ask again for isLoaded(textureAtlasName) => false; <<< OK
5) so unload(TextureAtlasName); => GdxRuntimeException("Asset not loaded: " + fileName); <<< OK

v1.9.11-SNAPSHOT ~ 2020-03-01
1) assetManager.load(any TextureAtlas);
2) ask for isLoaded(textureAtlasName) => true;
3) so unload(TextureAtlasName); => OK
4) ask again for isLoaded(textureAtlasName) => true; <<< BUG
5) so unload(TextureAtlasName); => NullPointerException <<< BUG

#### Version of LibGDX and/or relevant dependencies
1.9.11-SNAPSHOT (v.1.9.10 is ok)

#### Stacktrace
```java
| 2020/03/01 21:09:46.103 |  ERROR  | Current screen: com.sjet.gula.graphics.screens.singleplayer.SingleplayerGameScreen                                                                               [ WorldOfClonesGame.java:343 ]
[...]

Issue #5931, opened 2020-03-03
ObjectMap is broken


TCreutzenberg:
#### Issue details
To start: Everything works fine with LibGDX 1.9.10 (just tested) and it worked fine with 1.9.11-SNAPSHOT until a few days ago, this definately is a "new" 1.9.11-SNAPSHOT problem.

My IDE just updated to the latest LibGDX 1.9.11-SNAPSHOT, and now my AssetLoader doesn't load some assets any more.
The tasks are empty, but the asset is nowhere to be found (checked via breakpoint-debugging before unloading the asset, after AssetManager.finishLoading()).

#### Reproduction steps/code
I would have to upload my whole repo which I don't want to actually, but I get the problem only with the (new) snapshot, everything worked well a few days ago.

#### Version of LibGDX and/or relevant dependencies
1.9.11-SNAPSHOT

#### Stacktrace
My stacktrace will not help but here it is anyway
```java
Exception in thread "LWJGL Application" com.badlogic.gdx.utils.GdxRuntimeException: Asset not loaded: config/worldMap.data
	at com.badlogic.gdx.assets.AssetManager.unload(AssetManager.java:232)
	at de.creutzenberg.towerdefence.framework.assets.TCAssetManager.unload(TCAssetManager.java:528)
	at de.creutzenberg.towerdefence.framework.Global.initializeAndUnloadGameFiles(Global.java:146)
	at de.creutzenberg.towerdefence.TCGame.initializeAndUnloadBasicStuff(TCGame.java:159)
	at de.creutzenberg.towerdefence.ScreenManager.switchToNextScreen(ScreenManager.java:204)
	at de.creutzenberg.towerdefence.framework.ATCScreen.render(ATCScreen.java:117)
	at de.creutzenberg.towerdefence.TCGame.render(TCGame.java:198)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:233)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:128)
```

#### Please select the affected platforms
Only tested and veryfied on:
- [X] Windows

TCreutzenberg:
![AssetManagerBug](https://user-images.githubusercontent.com/17275393/75779044-3b80f100-5d59-11ea-9f86-2368c6174b95.PNG)

mgsx-dev:
it might be the same issue as #5929 
[...]

PR #5932, opened 2020-03-03
Fix for ObjectMap/LongMap/others when mixing put()/remove()

### LINKED ISSUE #5930: LongMap is buggy
#### Description
The current implementation of LongMap definitely has a bug. 
After switching back to commit e8fd323c485db7a292c848ad11d6f4c92a386ca7 the problem does not occur anymore. Unfortunately I have a relatively complex construct in my project, which I can't convert into a simple example at the moment.

I add entries and delete them relatively often. After a while an entry is not found anymore, although it is still visible in the debugger in the variable "keyTable".

In any case, I can see that in the following routine the condition "`if(other == 0)`" causes subsequent entries in keyTable to be ignored (called by containsKey):

```java
	private int locateKey (long key) {
		long[] keyTable = this.keyTable;
		for (int i = place(key);; i = i + 1 & mask) {
			long other = keyTable[i];
			if (other == 0) return -(i + 1); // Empty space is available.
			if (other == key) return i; // Same key was found.
		}
	}
```

#### Version of LibGDX
1.9.11-SNAPSHOT

propably all?

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [x] Linux
- [ ] MacOS

tommyettinger:
This should fix #5930 and #5931 , it may also help with #5929 .

This is a relatively small set of changes; the backwards-shift algorithm for deletion was (if I remember it right) correct when the maps and sets used Robin Hood hashing, but not for linear probing. The changes needed to fix all the maps and sets are the same; it just needs to avoid backwards-shifting any keys to before where their place() slot would be without probing. The code is somewhat hard to read because it needs to compare two differences between slots while considering wrap-around from the mask; I don't see any way around this. Each file for a hash-based map or set should have changed in two places (remove() and the iterator's remove()), except IdentityMap, OrderedMap, and OrderedSet, which use a parent implementation. There is a JUnit test; without the changes in the first commit, the test should fail, but with the changes it should pass.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
+++ b/gdx/src/com/badlogic/gdx/utils/IntFloatMap.java
@@ -16,11 +16,11 @@
 
 package com.badlogic.gdx.utils;
 
-import static com.badlogic.gdx.utils.ObjectSet.*;
-
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+
+import static com.badlogic.gdx.utils.ObjectSet.tableSize;
 
 /** An unordered map where the keys are unboxed ints and values are unboxed floats. No allocation is done except when growing the
  * table size.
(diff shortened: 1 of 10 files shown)


NathanSweet:
Looks good, thanks!

The logic for the "between" cyclic check is:
```if (i <= next ? i >= placement || placement > next : i >= placement && placement > next) {```
I don't think we can do better than your math though.

tommyettinger:
Thanks, agh, my brain and that check... Once nightlies build we'll be able to see if this fixes #5929 as well as the more obviously-related #5930 and #5931 .

NathanSweet:
Mac wasn't up so the build stalled: https://libgdx.badlogicgames.com/jenkins/job/libgdx-mac/

>  (pending—All nodes of label ‘anymac’ are offline) 

Will try to find someone to fire up a Mac!

Issue #4446, opened 2016-11-30
Slider/ProgressBar knobAfter incorrect length


raeleus:
#### Issue details
Recent changes in the 1.9.5-SNAPSHOT have negatively affected how sliders and progress bars render. If you include a "knob" and a "knobAfter" in the ProgressBarStyle or SliderStyle, the knobAfter will extend outside of the bar to the right if oriented horizontally. It is inconsistent to behavior in previous versions. See below for the expected behavior (1.9.4):
![preview1](https://cloud.githubusercontent.com/assets/12948924/20738426/e4396e72-b667-11e6-9a36-4627f7f8a4a7.png)

And the result from updating to 1.9.5-SNAPSHOT:
![preview2](https://cloud.githubusercontent.com/assets/12948924/20738433/f225b4a0-b667-11e6-8f61-0b5fa8e19221.png)

Notice how the knobAfter extends well past the edge of the background drawable.

#### Reproduction steps/code
Using these images:

![background](https://cloud.githubusercontent.com/assets/12948924/20738471/3952ae32-b668-11e6-9c16-5717af99e5f8.png) ![knob](https://cloud.githubusercontent.com/assets/12948924/20738477/40cb88f0-b668-11e6-9484-53d41bf731a4.png) ![knob-before](https://cloud.githubusercontent.com/assets/12948924/20738484/451a148a-b668-11e6-92eb-c131e0e78877.png) ![knob-after](https://cloud.githubusercontent.com/assets/12948924/20738489/48e77a44-b668-11e6-95cf-400d2a6943ce.png)

I wrote the following class to reproduce the issue:

```java
package com.ray3k.slidertest;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.graphics.GL20;
import com.badlogic.gdx.graphics.Pixmap;
import com.badlogic.gdx.graphics.Texture;
import com.badlogic.gdx.graphics.g2d.PixmapPacker;
import com.badlogic.gdx.graphics.g2d.TextureAtlas;
import com.badlogic.gdx.scenes.scene2d.Actor;
import com.badlogic.gdx.scenes.scene2d.Stage;
import com.badlogic.gdx.scenes.scene2d.ui.ProgressBar;
import com.badlogic.gdx.scenes.scene2d.ui.ProgressBar.ProgressBarStyle;
import com.badlogic.gdx.scenes.scene2d.ui.Skin;
[...]

Issue #1419, opened 2014-03-03
ProgressBar knobBefore & knobAfter ignore background NinePatch paddning


Swiftwork:
Found a small issue where the ProgressBar renders both knobBefore and knobAfter outside the specified NinePatch backgrounds padding. I believe a simple change to the draw should fix this.
![capture](https://f.cloud.github.com/assets/455178/2305808/76acd8d6-a26e-11e3-9aa5-c0bafd89cf67.PNG)

Origional:

``` java
if (knobBefore != null) {
    knobBefore.draw(batch, x, y + (int)((height - knobBefore.getMinHeight()) * 0.5f), (int)(position + knobWidthHalf), knobBefore.getMinHeight());
}
if (knobAfter != null) {
    knobAfter.draw(batch, x + (int)(position + knobWidthHalf), y + (int)((height - knobAfter.getMinHeight()) * 0.5f), width - (int)(position + knobWidthHalf), knobAfter.getMinHeight());
}
```

Change:

``` java
if (knobBefore != null) {
    knobBefore.draw(batch, x + bg.getLeftWidth(), y + (int)((height - knobBefore.getMinHeight()) * 0.5f), (int)(position + knobWidthHalf), knobBefore.getMinHeight());
}
if (knobAfter != null) {
    knobAfter.draw(batch, x + (int)(position + knobWidthHalf), y + (int)((height - knobAfter.getMinHeight()) * 0.5f), width - (int)(position + knobWidthHalf) - bg.getRightWidth(), knobAfter.getMinHeight());
}
```

Add bg.getLeftWidth() to x on knobBefore and subtract bg.getRightWidth() from width on knobAfter.

**_Might be worth checking vertical bar aswell.**_

Kind Regards
Swiftwork

WinterAlexander:
That resolved the knobBefore issue but the knobAfter is not correct. Similar change to what was done in the PR should be made to support proper padding in knobAfter.

PR #5937, opened 2020-03-05
Fixed incorrect padding of knobAfter

WinterAlexander:
In this issue #1419 it was found that the rendering knobBefore taking in consideration the padding of the background was incorrect. The same was true for knobAfter yet nothing was done about it (fix was only applied to knobBefore). This results in knobBefore getting drawn up to the end of the bar without caring about the padding of the background:

image

Fixed version:

image

Note that this only affects: Uses of ProgressBar with knobAfter that have a background with non-zero padding. It is possible that nobody noticed this bug for many years given how uncommon the use of knobAfter is. (Even myself don't really need it in a vanilla type of way)

Another issue with progress bar was later fixed after the initial one I mentionned: #4446  But this issue only address the typo which caused the bar to have an extra length of the size of the knob. The test example for this has no padding and doesn't mention the padding problem I am fixing here in anyway.

To test this bug fix: make a progress bar with a ninepatch background with padding, and add both knobBefore and knobAfter to be something.


NathanSweet:
Cheers!

PR #5941, opened 2020-03-07
Add constructor to create a rectangle from center point

abueide:
Simple change, would be useful to me and maybe others.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -49,6 +49,15 @@ public class Rectangle implements Serializable, Shape2D {
 		this.height = height;
 	}
 
+    /**
+     * Constructs a new rectangle with the given center point.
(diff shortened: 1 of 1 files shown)


mgsx-dev:
i think it adds confusion with existing constructors and may require to create a Vector2.
Possibly something like `setCentered(Vector2,float,float)` and `setCentered(float,float,float,float)` instead but i'm not sure it worth it since you can already do it like this: `new Rectangle().setSize(...).setCenter(...)`

NathanSweet:
Agreed, I think it's important not to have too many constructors (or overloads). They don't have descriptive names, so choosing the one you want can require reading all the javadocs (see the mess that is Batch `draw(...)`!). In this case `setCenter` seems sufficient. You could use utility methods with a static import in your app code if you do it a lot.

abueide:
Ah okay I failed to notice that setCenter() returns itself after mutating the state. Makes sense, I will just do that instead.

PR #5942, opened 2020-03-07
Fix DefaultTextureBinder replacing WEIGHTED strategy by LRU

mgsx-dev:
DefaultTextureBinder WEIGHTED method doesn't work. Consider this example : 
- a DefaultTextureBinder with count=4
- a model A with 3 textures (textures 1, 2 and 3)
- a model B with 3 other textures (textures 4, 5, 6)
- render model A several times (textures 1, 2, and 3 get high weights)
- render model B : 
    - texture 4 is bounds to the last available slot with a light weight
    - texture 5 takes texture 4 slot because of its lighter weight, then model B cannot be rendered correctly.

I split my changes in several commits for readability and so you can test it yourself this bad behavior with first commit. (please squash them all when merging)

the new LRU strategy becomes the default which may be a little better than round robin method for most use cases.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@
 - API Change: scene2d: In TextField, only revert the text if the change event was cancelled. This allows the text to be manipulated in the change listener.
 - API Change: scene2d: Tree.Node#removeAll renamed to clearChildren.
 - API Addition: scene2d: Added SelectBox#setSelectedPrefWidth to make the pref width based on the selected item and SelectBoxStyle#overFontColor.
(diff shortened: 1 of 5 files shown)


NathanSweet:
Is someone else maintaining this part of libgdx? I haven't found a need to use it myself yet. If there is no one then I can merge and hopefully @mgsx-dev can follow up if any problems occur.

xoppa:
Looks good, thanks!

RUSshy:
Xoppa is back :D

PR #5944, opened 2020-03-10
remove unnecessary glUseProgram(0)

mgsx-dev:
Unbinding a shader program is not necessary, it was useful to enable the old fixed-function pipeline which libgdx don't use.

ShaderProgram begin and end method are deprecated for a smooth transition.

I made a quick dumb test to see how my GPU was impacted (if my GPU driver optimize it or not) : 

here is the code:
```java
    private ShaderProgram sp;
    public void render () {
        int N = 1000;
        for(int i=0 ; i<N ; i++){
            sp.begin();
            sp.end();
        }
        logger.log();
    }
```
by commenting begin/end in all combinations i get:
- calling none of them (baseline): ~ 10 500 FPS
- calling only end: ~ 9500 FPS
- calling only begin: ~ 2000 FPS
- calling both begin and end: ~ 1000 FPS

GPU : Nvidia GTX 950m

I'm not calling this a benchmark, i just wanted to highlight this small optimization and ensure there was no drawbacks. Most of games try to minimize shader switches to only a few anyway, so i don't think it would improve perf. noticeably. But it still better to not call it at all.

Note that my original issue was GLProfiler counting a shader switch as twice because of it. This issue is then fixed by this PR as well.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@
 - API Change: scene2d: Tree.Node#removeAll renamed to clearChildren.
 - API Addition: scene2d: Added SelectBox#setSelectedPrefWidth to make the pref width based on the selected item and SelectBoxStyle#overFontColor.
 - API Change: DefaultTextureBinder WEIGHTED strategy replaced by LRU strategy.
+- API Change: ShaderProgram begin and end methods are deprecated in favor to bind method.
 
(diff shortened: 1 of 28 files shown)


yuripourre:
Nice finding!

PR #5949, opened 2020-03-13
Add AndroidLiveWallpaper.notifyColorsChanged

cypherdare:
This exposes the live wallpaper engine's ability to report "visually significant" colors for the launcher and lock screen to use. I don't have a phone that noticeably uses this feature, but one of my users confirmed after I updated my app with this code that their lock screen now correctly shows bright text for the clock overlay when my wallpaper is dark, and vice versa.

I didn't add the method to the `Application` interface, since it would just be clutter for anything that isn't a Live Wallpaper. If it needs to be called from the `core` module, a platform resolver interface can be used.

I thought it would be most convenient to use Gdx's `Color` class for reporting the colors. The Colors are copied rather than directly referenced so this method can be thread-safe. The documentation of `WallpaperService.Engine` doesn't specify whether `onComputeColors()` has to be thread-safe, so I just went ahead and did it. This isn't a method that will be called frequently.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -59,6 +59,7 @@
 	- 32-bit: ARMv7/armhf
 	- 64-bit: ARMv8/AArch64
 - API Change: Removed arm abi from SharedLibraryLoader
+- API Addition: Added AndroidLiveWallpaper.notifyColorsChanged() to communicate visually significant colors back to the wallpaper engine.
 
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,7 @@
 - Gdx.files.external on Android now uses app external storage - see wiki article File handling for more information
 - API Addition: Added setProgrammaticChangeEvents to ProgressBar/Slider.
 - Keyboard events working on RoboVM on iOS 13.5 and up, uses same API as on other platforms
+- API Addition: Added AndroidLiveWallpaper.notifyColorsChanged() to communicate visually significant colors back to the wallpaper engine.
 
 [1.9.11]
 - Update to MobiVM 2.3.8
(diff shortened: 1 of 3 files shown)


MrStahlfelge (changes requested):
This PR has to changed because CHANGES will list the addition under 1.9.11

That said, I am not sure if this should be added to the framework. It looks like a good addition, but it is not useable from the core and therefore not ready to use. It has to be called from the Wallpaper code - so on Android code, which can do all the things anyway.

cypherdare:
There's already a precedent for confining live wallpaper specific stuff to the `android` backend in order to avoid polluting the API of all games on all platforms for wallpaper-specific stuff. See [AndroidWallpaperListener](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidWallpaperListener.java), for example.

I'm not opposed to adding a `notifyWallpaperColorsChanged()` method to the Gdx `Application` interface, but that would be kind of selfish since it's a breaking change and most people are not making live wallpapers. I'm guessing if we polled libGDX users, most would be opposed.

If this is not put in the libGDX code, it can only be implemented by copy-pasting and making edits to AndroidLiveWallpaperService, AndroidLiveWallpaper, and AndroidGraphicsLiveWallpaper. And these clones have to be in the badlogic package. It would be a big help to have this in the backend code so we aren't having to maintain the clones to keep up with other changes that are made to LibGDX.

MobiDevelop:
I don't see any problem with keeping this Android specific thing in the Android backends only.

MrStahlfelge (approved):

MrStahlfelge:
Yes, I completely agree that wallpaper specific stuff shouldn't be added to the main interface. I understand your reasons wanting this to have it in the wallpaper classes, too. Im okay with adding this.

MrStahlfelge:
Thanks for the contribution. Wallpapers lack a bit of documentation on the wiki in my opinion. If you are willing to, you could add an article on the topic @cypherdare

cypherdare:
@MrStahlfelge Will do.

PR #5953, opened 2020-03-15
Update gitignore for Idea projects

raeleus:
Updated gitignore to exclude "out" folders which contain temporary build files and copies of assets in Intellij Idea.


extensions/gdx-setup/src/com/badlogic/gdx/setup/resources/gitignore
```diff
@@ -30,6 +30,9 @@ www-test/
 *.iml
 /android/out/
 com_crashlytics_export_strings.xml
+desktop/out/
```
mgsx-dev: please use a slash prefix to avoid unwanted ignore (eg. /desktop/out/) like other ignore in this file. Same for the 2 other lines.

mgsx-dev: also, maybe ios and ios-moe projects needs this?

raeleus: I can't test those backends. Can someone comment on this? I would be happy to add.

raeleus: Nevermind, I have included the changes. Thanks MGSX.

obigu: I would test but I haven't seen /out folders generated by IDEA on any backend. How do build to generate them?

raeleus: I don't know if it's some strange setting on my IDE, but I generated a new project and ran the DesktopLauncher directly. First time I opened the project, it didn't generate the out folders. After I closed and reopened, it does it every time now. I don't know if this started since a recent update to Idea or from the GDX setup update, but it wasn't always like this.  I'll test on a fresh PC in a little while.

obigu:
As discussed on Discord this looks like a specific environment/configuration issue on OP's machine. After some testing on different newly generated projects, `/out` folders are never created by default on imported libGDX projects on IDEA. 
Even if it's a minor thing imo the PR should be closed unless new information on how to reproduce is provided.

PR #5956, opened 2020-03-16
Move resources from java src directory into own resource directory

SimonIT:
This fixes the problem in IntelliJ, that the src directory is marked as resource directory and no code editing is possible


Tom-Ski:
Have you tested the maven build with the changes?

SimonIT:
Yes, I executed `compiler:compile` and `jar:jar` in IntelliJ and both times I got a _BUILD SUCCESS_. 
Here for example the "core"
[gdx-1.9.11-SNAPSHOT.jar.zip](https://github.com/libgdx/libgdx/files/4339133/gdx-1.9.11-SNAPSHOT.jar.zip)

intrigus:
What about gwt?

SimonIT:
What do you mean exactly? I can run `.\gradlew tests:gdx-tests-gwt:superDev` and everything is fine if I go thru the tests. The music is playing, the scene2d skin is loaded

obigu:
I'm seeing an issue when running tests using LwjglTestStarter like UISimpleTest where it fails finding resources:
```
Exception in thread "LWJGL Application" com.badlogic.gdx.utils.GdxRuntimeException: File not found: com\badlogic\gdx\utils\arial-15.fnt (Classpath)
	at com.badlogic.gdx.files.FileHandle.read(FileHandle.java:142)
	at com.badlogic.gdx.graphics.g2d.BitmapFont$BitmapFontData.load(BitmapFont.java:493)
	at com.badlogic.gdx.graphics.g2d.BitmapFont$BitmapFontData.<init>(BitmapFont.java:485)
	at com.badlogic.gdx.graphics.g2d.BitmapFont.<init>(BitmapFont.java:131)
	at com.badlogic.gdx.graphics.g2d.BitmapFont.<init>(BitmapFont.java:72)
	at com.badlogic.gdx.tests.UISimpleTest.create(UISimpleTest.java:59)
```

obigu:
Previous issue only occurs when running from IDEA (`Run as` on `LwjglTestStarter`). When running from Gradle the classpath is properly resolved.

This may be an IDEA limitation (https://intellij-support.jetbrains.com/hc/en-us/community/posts/206234729-Resources-and-classpath-when-running-test) that populates classpath by only taking into consideration the output directory of the modules (in this case `gdx\build\classes\java\main`) instead of the proper Gradle classpath.

I would say it's not a bug, the downside is that Gradle is required to run the tests (not sure how Eclipse deals with this).

obigu:
After some more investigation and discussion with @SimonIT  we've just realised that this issue is already occuring on master branch. We have a fix for it but will be independent from this PR.

Tom-Ski:
Usually a build should fix this. I have the same problem in a non-libgdx, but similarly managed project.

SimonIT:
It seems, that the automatic builded gdx setup is broken because of this split. The reason that ant is used for building.
We have two options to fix this:
1. Use the gradle script I created in a recent PR to create the gdx-setup.jar (IMO preferred)
2. Fix the ant scripts (I got no clue about ant)

PR #5966, opened 2020-03-22
Fix Ant script

obigu:
This should fix a couple of issues caused by #5956 until we migrate to Gradle.


OliPerraul:
Where can I download the fixed jar?

obigu:
From the libGDX site link once the build server completes the new build (minutes probably).

OliPerraul:
Are you sure these are not just uploaded at night?

image

Issue #5957, opened 2020-03-17
difficulties in updating old projects [enhancement]


Rami-Majdoub:
It is very hard to tell what to update.
It would be very cool if there was a command to update the old projects to the newest version.
I was trying to update a project generated by libGDX v 1.9.9 (Gradle distribution 4.10.1) and I had many Gradle issues.

Tom-Ski:
https://libgdx.badlogicgames.com/versions.html shows the versions that are used in the current snapshots and release verisons. These are known compatible versions, so you can use these with the Gradle version that is provided by the setup tool.

Rami-Majdoub:
I have changed the versions and sync the project but it didn't compile.
there have been some changes in the Gradle files (deprecated methods and code that has been deleted)

tommyettinger:
It's much, much easier to generate a new empty project and copy your old code/assets into that, making any changes to build files that your project needs. LibGDX is using an older version of Gradle than the current latest, to get better compatibility with Android build tools, but there are still necessary changes to the Gradle files that should be re-generated by setup. Because people generally customize their build.gradle scripts, I don't see any reasonable way to automatically update a Gradle project.

Rami-Majdoub:
that's what I do, but I have older projects and it is frustrating to generate a new project, copy-paste all the code, to remember every modification and to do it again.

RUSshy:
gradle sucks, but that is what Java world and Jetbrains choosed.. sadly

https://docs.gradle.org/current/userguide/upgrading_version_4.html#for_all_users

check this official guide for upgrading from old to newest version of gradle

shatterblast:
Gradle 4.8 should be safe enough, and it support Java 11.

tommyettinger:
Gradle 4.8 is pretty old, though if it works, it works. Most libGDX developers are using JDK 8 anyway unless they target desktop only. The concern is with Java 14 becoming the new LTS version (has this already happened?); Gradle 6.3 or higher is required to support Java 14, and there are actual new useful features in this Java version. Last I checked, Gradle 6.3 was still in a release candidate version.

IntelliJ Community does allow building GWT applications just fine; it doesn't debug them especially well (you can use superDev and step through in a browser though) and it doesn't recognize some GWT conventions out of the box (like the `super` folder for super-sourced code, which libGDX uses), but you can run the Gradle GWT tasks like you can from anywhere.

shatterblast:
[...]

PR #5976, opened 2020-03-23
Android backends as android library

SimonIT:
As part of #5846, a goal was to have the android backends as library. With this change, we get a better IDE integration and more possibilities for building the backends with gradle (which is the goal for the future).
With this configuration, we can set some android requirements for libgdx. For example the minSdk gives developers a hint while update their libgdx version. (May help with #5957)
An option which we may use in the future is the deploying of native via aar. This is normal way for android libraries to deploy their natives. Currently, it isn't configured.
Another great value has now the linter which checks the android code for errors with different api versions etc.

Someone may say, that these changes will drop the support of other IDEs than Android Studio / IntelliJ to develop android application. That's not true. This PR only adds more options to develop and deploy libgdx. It doesn't remove the current ones. Only if we use only the aar deployment, it will break probably the support for other IDEs. In my opinion, this wouldn't hurt either because it's a big effort to use the other IDEs with android anyway, what most developers know.


shatterblast:
Until GWT possibly becomes free on IntelliJ Community, which it might not, Eclipse is still useful for LibGDX web applications.

MrStahlfelge:
The important question here: is this a change breaking something, or is it non-breaking?

SimonIT:
I don't see it as breaking because no one has to change something after merging this. At first because maven is used as deployment and secondary is the logic for downloading the natives is still present inside the build gradle files and should also still work

MrStahlfelge:
@mgsx-dev can you test if this works for you - except Android - in Eclipse?

mgsx-dev:
@MrStahlfelge it seams to be working, even for android. Actually Eclipse setup doesn't use Gradle at all, it using eclipse files and jars in libs folder. So i don't think there are any impacts. I'll removed me from reviewers because i can't say if changes or fine, i'm not an Android guy anymore :-)

mgsx-dev:
however, the question is for libgdx projects, does android backend still a jar?

SimonIT:
What if we add a task which extracts the jar from the aar and adds it as dependency? @PokeMMO talked about something like this (If I remember correctly)
IntelliJ / Android Studio users would have the good support which they now from other android projects and eclipse user can still edit their files

PokeMMO:
It's possible to do, but every user will ahve to edit their gradle config to do it.

What benefit does this add, since natives aren't supplied from this aar?

obigu:
I think the discussion about AAR or Eclipse support is independent from this PR. As @SimonIT explains, this PR only changes the Gradle build of libGDX project itself to make it play nice with IDEs that do integrate Android Gradle tool chain. But Gradle is not used to build the libGDX published artifacts afaik (we use Maven and Ant) which are still JARs.

Building Android modules the Android way makes IDEs integrations that support it better and more solid.

I've merged it locally and run into some issues when running Gradle Build with the modules `tests:gdx-tests-android` first and then `extensions:gdx-controllers_gdx-controllers-android`. 
The error is the following: `Program type already present: com.badlogic.gdx.Application$ApplicationType`

Unfortunately, fixing this errors though is a bit like a waterfall.

First issue is we are using an old version of the Android plugin (`3.1.0`) which doesn't play well with Android X (see https://github.com/libgdx/libgdx/issues/6038). By listing gdx-test-android current dependencies we can see there's a mix of android.support and Android X dependencies which is bad. Updating it to `3.5.4` (which also requires updating the Gradle wrapper to `5.4.1`) fixes the problem on `tests:gdx-tests-android` but then a similar error (but not exactly the same) appears on `extensions:gdx-controllers_gdx-controllers-android`.

To fix that one I've had to remove a workaround we needed to run tests on Idea which seems not to be needed anymore after making everything Gradle friendly. 

The last one is a minor Lint error complaining about not having Vibration permission on the manifest (it will probably worth disabling this lint check as we don't want to add vibration permission to libGDX manifest).

@SimonIT I've made these changes on this branch https://github.com/obigu/libgdx/tree/5976 (see last commit https://github.com/obigu/libgdx/commit/f05d6ac2fedb5807251a854b1592e6ec85db90d3). It would be great if you could have a look and consider merging those changes.

PokeMMO:
>But Gradle is not used to build the libGDX published artifacts afaik (we use Maven and Ant) which are still JARs.

Yet...

obigu:
Wow, small update, the errors when building with Gradle happen in Master, they are not directly related to this PR even if merging this is probably necessary to fix them.

obigu (approved):
I've pushed a few changes to fix the Gradle issues after having a chat with @SimonIT who is a bit busy atm. It should be good to merge now.

PokeMMO:
>It should be good to merge now.

I don't think it is, afaik the plan is to seriously consider moving the build to gradle sometime soon™. Unless it's also still possible to build a jar from an android library project?

obigu:
> > It should be good to merge now.
> 
> I don't think it is, afaik the plan is to seriously consider moving the build to gradle sometime soon™. Unless it's also still possible to build a jar from an android library project?

I understand the dilema but, so far, this is the only way we've managed to fix the build which has been broken for a long time. If the plan is to migrate to Gradle, we need to get the Gradle build working first.

PokeMMO:
Now that we build with gradle, I think it's finally time to revive this PR. @SimonIT would you like to clean it up vs master?

SimonIT:
I rebased it on master. Two things have to be done before merging:

- [ ] Add publishing. The current publishing config doesn't work with the android plugin
- [ ] Add a `uberJar` task for android (I think @badlogic want's that) because the current one also doesn't work

badlogic:
No need for uberJar for Android.

PokeMMO:
https://github.com/PokeMMO/libgdx/commit/c09e269de56976f74277b532bdb0c1c577195d7f should implement aar publishing support.

See https://stackoverflow.com/a/42160584

If we ever move to AGP 7.1 we can use a simpler syntax.

obigu (approved):
Tested building and running tests. So good to see the Android /libs gone!!

Berstanio:
Is there anything left that prevents the merge of this?
I rebased it on master and simplified publishing logic as suggested (because #6941 is merged) and it worked perfectly well for me: 
https://github.com/Berstanio/libgdx/tree/android-backends-as-android-library

PR #5982, opened 2020-03-27
Remove armeabi native libs

obigu:
Armeabi v5 and v6 ("armeabi") architecture is very old and the last devices that required it were Gingerbread (API 9) devices.

Since we upgraded minSDK to 14 we don't need armeabi libs anymore. I've confirmed on Google Play Console simulating the release of an app with min SDK 14 with and without "armeabi" architecture and there are 0 devices that stop being supported after removing it. With this change APK are smaller and build/download times will improve a bit.

Also API 14 has been set on NDK Application.mk files (it was still set to 9).


PokeMMO:
The[ WIP standalone jnigen projec](https://github.com/libgdx/gdx-jnigen)t already has better support for doing this in a more future proof manner. (See https://github.com/libgdx/libgdx/pull/5733)

PR #5989, opened 2020-04-04
Fix android gap in music when looping

wheelergames:
Fixes longstanding issue with android looping problems in music.

Currently there is often a small quiet gap between the end of music finishing and starting again on Android.

This fix uses 3 Mediaplayers that queue up the music to be played one after the other instead of looping on the one Mediaplayer. 

This fix has come from someone else on the internet fixing it for non LibGDX app, but it works here (I've used it in two of my games so far).

It's possible that it could work with just 2 Mediaplayers, but as it works as it is, I've never felt the need to refactor it to try. I guess having 2 would be less memory intensive.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudio.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


noblemaster:
+1 to have this added. Haven't tested it. Comments?

obigu:
I'm not sure it's a good idea to merge this without a lot of testing. Audio on Android is very delicate, some bugs are device or SDK version specific. We should also measure the impact on performance because MediaPlayer is a heavy object (https://developer.android.com/guide/topics/media/mediaplayer). 

If we end up creating interfaces for our "modules" (see discussion on https://github.com/libgdx/libgdx/pull/5985 ) I see this more being an optional replacement for the current AndroidAudio that could be distributed as a library.

PR #5998, opened 2020-04-15
add ContactListenerTest to Box2DTestCollection

fabiitch:
I build gdx and try test classes for discover.
And i see ContactListenerTest was missing in Box2DTestCollection.

I see also a number of Box2D class Test wasnt named with Test at end. like Pyramid or VecticalStack
Should i rename this classes ?

Thanks for reading


fabiitch:
ah i see why he wasnt in box2Dcollection, gwt incompatble, i close this pr !

PR #5999, opened 2020-04-15
box2D World.java contact listener method, dont affect addr if no listener

fabiitch:
hello, i explore gdx source, and on World.java , i see useless code if contactListener is null.
Not a big gain for sure ^^.
Thanks for reading

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
@@ -983,25 +983,33 @@ b2ContactFilter defaultFilter;
 	private final ContactImpulse impulse = new ContactImpulse(this, 0);
 
 	private void beginContact (long contactAddr) {
-		contact.addr = contactAddr;
-		if (contactListener != null) contactListener.beginContact(contact);
+		if (contactListener != null){
+			contact.addr = contactAddr;
+			contactListener.beginContact(contact);
+		}
 	}
 
 	private void endContact (long contactAddr) {
-		contact.addr = contactAddr;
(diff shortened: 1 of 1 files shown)


noblemaster:
I am going to reject this: just because there is no listener, the changes still have to be applied! Unless I am missing something?

obigu:
@noblemaster Well, those objects are just used in the listeners and its purpose is to avoid creating new instances every time a listener is called so they can be reused. It is safe to merge the PR but, as OP says, it will not make much of a difference.

fabiitch:
:(

noblemaster:
I re-opened the request. I see the gist of it now. I think, given that all those objects are used by the listener only, they should be moved into the listener. It's makes no sense to keep those objects in the ```World``` object if they are for the listener only.

Can you move them into the listener? Unless they are any objections, that makes more sense to approve than the changes you provided. :)

obigu:
ContactListener is an interface. I think they are ok where they are, we do this in many places to avoid instantiation. What i'd say is, if this is going to be merged, also change the code in the GWT World for consistency.

fabiitch:
as say Obigu, ContactListener is an interface, so no private fields
Is possible to change ContactListener to an abstract class, but its an hard breaking changes.

@obigu  i look for World.java in GWT, i think its already done 
com/badlogic/gdx/physics/box2d/gwt/emu/com/badlogic/gdx/physics/box2d/World.java:90

`	public void setContactListener (final ContactListener listener) {
		if (listener != null) {`
This is the part you are talking about ?

obigu:
@fabiitch yeah, perfect if it's already done there, just wanted to ensure consistency :)

Blunderchips:
Any progress on this issue?

PR #6002, opened 2020-04-17
TextureView option for android backend.

asimaruk:
It gives an ability to have renderring TextureView in a regular android-view layouts, allowing you to render on part of a screen and over other views in a layout.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/textureview/GLTextureView.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/textureview/GLTextureView.java
@@ -1,0 +1,1705 @@
+package com.badlogic.gdx.backends.android.textureview;
+
+import android.annotation.TargetApi;
+import android.content.Context;
+import android.graphics.SurfaceTexture;
+import android.opengl.GLDebugHelper;
+import android.opengl.GLSurfaceView;
+import android.os.SystemClock;
+import android.util.Log;
+import android.view.KeyCharacterMap;
+import android.view.KeyEvent;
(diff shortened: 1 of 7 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/textureview/GLTextureView.java
```diff
@@ -0,0 +1,1705 @@
...
+import javax.microedition.khronos.egl.EGL11;
+import javax.microedition.khronos.egl.EGLConfig;
+import javax.microedition.khronos.egl.EGLContext;
+import javax.microedition.khronos.egl.EGLDisplay;
+import javax.microedition.khronos.egl.EGLSurface;
+import javax.microedition.khronos.opengles.GL;
+import javax.microedition.khronos.opengles.GL10;
+
+/*
+ * Copyright (C) 2018 Wasabeef
```
Darkyenus, line 36: This seems to be a 3rd party library. (Not necessarily a problem, but just to make it clear, since the PR text does not.)

asimaruk: Yes, I took this class from other project and modified a little. Wasn't sure about copyright and how to deal with it, so left it as it was. Should it be replaced or modified somehow?

MobiDevelop:
I like giving the option, in spirit. It would be nice to have some of this be more common with GLSurfaceView20, if possible. Is there any way to share code between them? I haven't looked too closely at the GLTextureView to see exactly what it is doing so if they are necessarily totally different, let me know.

PR #6003, opened 2020-04-17
Iskeyjustpressed index out of bounds on Lwjgl3

BambooBandit:
Fixed an index out of bounds error that occurs when you pass 256 into isKeyJustPressed() on Lwjgl3 backend


Darkyenus:
Replacing this with a constant or `justPressedKeys.length` (as it should have been originally anyway) would probably be a better fix and more obviously correct.

BambooBandit:
Alright, I made that change

PR #6004, opened 2020-04-19
Add boilerplate for screen switching

thechubbypanda:
Everywhere I've found tutorials, this is the way they do it. I figured it would be nice to have it built in.


thechubbypanda:
Realized workarounds

Issue #5904, opened 2020-02-04
Starting April 2020, apps submitted to the App Store must use an Xcode storyboard.


itsabhiaryan:
#### Issue details
Starting April 2020, apps submitted to the App Store must use an Xcode storyboard to provide the app’s launch screen and must have an interface that supports any display size but currently by default we don't have any storyboard in iOS module.

Thread link : https://developer.apple.com/news/?id=01132020b

#### Version of LibGDX and/or relevant dependencies
LibGDX : 1.9.10
roboVM : 2.3.8

#### Please select the affected platforms
- [X] iOS (robovm)

obigu:
Storyboard based launch screens work on MobiVM so it's just a matter of creating one for default project.

PR #6005, opened 2020-04-19
Replaced static launch screen images by StoryBoard

obigu:
The AppStore now requires launch screen to be a StoryBoard (https://github.com/libgdx/libgdx/issues/5904) which were introduced on iOS 8.

The new launch screen design is more minimal with the flat libGDX logo over a white background.

Screenshot-2020-04-19-at-15 52


noblemaster:
Thanks 👍 🥇

PR #6007, opened 2020-04-20
[Label] Fixed the setText() that won't update the size.

AndEditor7:
After Label created with short string than use setText() with a long string, The background I have using LabelStyle won't update with new string length.

The image shows Label with debug enabled, and the dark background using LabelStyle:
image

With this fix. It will always update the size after the text changed.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
@@ -112,6 +112,7 @@ public class Label extends Widget {
 			text.setLength(0);
 			text.append(newText);
 		}
+		if (newText.length() > 0) setSize(getPrefWidth(), getPrefHeight());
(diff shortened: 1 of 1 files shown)


Darkyenus:
This should be handled by whatever is doing the layout no? And the layout is invalidated two lines down.

The special handling of empty strings is also weird - why should it behave differently to non-empty strings?

NathanSweet:
This PR is unlikely to be merged because widgets are not intended to set their own size/position. Please read:
https://github.com/libgdx/libgdx/wiki/Scene2d.ui#layout

Changing the label text invalidates the hierarchy (the parents to the root-most widget group are marked as needing to layout). The parent should validate its layout (the parent sizes and positions its child widgets, usually using the min/pref/max size of each child) before drawing next, which means the label will be sized for the new text. If you are doing layout manually, then you should call `pack` after `setText` to set the label to its pref size.

PR #6009, opened 2020-04-21
correct bug in Matrix3 setToRotation

ryanastout:
Prior to this change, the following code:
```
    Matrix3 rot3 = new Matrix3();
    rot3.setToRotation(Vector3.Z, 90f);
    System.out.println("setToRotation with axis   \n"+ rot3);
    rot3.setToRotation(90);
    System.out.println("setToRotation without axis\n"+ rot3);
```
has this output:
setToRotation with axis   
[-8.742278E-8|1.0|0.0]
[-1.0|-8.742278E-8|0.0]
[0.0|0.0|1.0]
setToRotation without axis
[-4.371139E-8|-1.0|0.0]
[1.0|-4.371139E-8|0.0]
[0.0|0.0|1.0]
So the outputs when rotating by 90 degrees about the Z axis from the two methods give the following matrices:
0,1,0
-1,0,0
0,0,1 (which is incorrect) 
and 
0, -1 , 0
1, 0, 0
0, 0, 1 (which is correct)

After this change, the output is

setToRotation with axis   
[-8.742278E-8|-1.0|0.0]
[1.0|-8.742278E-8|0.0]
[0.0|0.0|1.0]
setToRotation without axis
[-4.371139E-8|-1.0|0.0]
[1.0|-4.371139E-8|0.0]
[0.0|0.0|1.0]

So the two outputs are both correctly 
0, -1 , 0
1, 0, 0
0, 0, 1 

https://en.wikipedia.org/wiki/Rotation_matrix#Rotation_matrix_from_axis_and_angle shows that the prior code for setToRotation(Vector3, float) was computing the transpose of what it could have computed.


mgsx-dev:
It looks great, however, this method is used by BillboardParticleBatch, so i guess some changes are required there as well.

ryanastout:
In BillboardParticleBatch.java, 
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/BillboardParticleBatch.java#L506
and
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/particles/batches/BillboardParticleBatch.java#L578
are the two places `setToRotation` is called. In each of those cases, the axis, `look` is the vector from the pixel to the camera or the negative of the camera direction, respectively. 
Unless I'm mistaken, the correct behavior in both cases is for the particle to rotate counterclockwise about the `look` vector. The behavior had been incorrect and this change would fix each of those.

noblemaster:
Can someone confirm it's indeed wrong ```BillboardParticleBatch``` and the new code fixes it?

ryanastout:
You could pull these changes yourself and test if you'd like. 
An easy way to test the BillboardParticleBatch code (the way I did it) is to build Flame (the 3d particle editor) before these changes, use a texture that's not radially symmetric and add an "Angular Velocity 2D" dynamics modifier. 
Give an angular velocity of 360 degrees per second. Before this change, you'll see the rotation is clockwise. After the change, the rotation of the particles with respect to the axis from the particle to the camera becomes clockwise. It's easier to see if you change the emission count to only be a small number of particles.

noblemaster:
OK, let's add this. Would you mind adding a line to the `CHANGES` file also just in case some people get surprised by rotation being the other way.

PR #6014, opened 2020-04-22
Create interfaces for Audio and Input on Lwjgl backend

obigu:
#5985 Lwjgl backend refactor.

Not many changes required. The only choice I've made is to create a new `LwjglApplicationBase` interface similar to what exists on Android that is implemented by all `LwjglApplication` implementations.


obigu:
Concerns about the naming conventions used in these PR have raised on Discord channel. I agree with those concerns that implementations should not be named with "Impl" suffix but with a descriptive prefix. So, for instance, instead of naming the `AndroidAudio` implementation `AndroidAudioImpl` we should name it `XXXXAndroidAudio` being "XXXX" descriptive or "Default" if no better name.

This is a good time to make this change as only the Android PR has been merged. 

Please give your thumbs up or thumbs down.

NathanSweet:
Seems OK to me. I'm OK with merging if no one has objections.

obigu:
@NathanSweet I think now that we've waited for a while for objections, it'd be great if we could start merging the whole set of refactoring PRs. Due to all the renaming changes it's easy to have conflicts with other PRs and I think it'd be positive to let them mature for a while in Snapshot.

PR #6016, opened 2020-04-24
Makes ios-moe generated by gdx-setup.jar work

ark100:
This PR makes ios-moe generated by gdx-setup.jar work.
libGDX with Multi-OS Engine(community) is available.

About org.multi-os-engine.community
https://discuss.multi-os-engine.org/t/moe-sdk-gradle-plugin-community-edition-release-1-5-0/2769

Generate project sample (using gdx-setup.jar)
https://github.com/ark100/multi-os-engine-libgdx-generateproject-sample

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
@@ -26,7 +26,7 @@ public class DependencyBank {
 	//Temporary snapshot version, we need a more dynamic solution for pointing to the latest nightly
 	static String libgdxNightlyVersion = "1.9.11-SNAPSHOT";
 	static String roboVMVersion = "2.3.8";
-	static String moeVersion = "1.4.0";
+	static String moeVersion = "1.5.1";
 	static String buildToolsVersion = "29.0.2";
 	static String androidAPILevel = "29";
 	static String gwtVersion = "2.8.2";
@@ -44,7 +44,7 @@ public class DependencyBank {
 	static String gwtPluginImport = "org.wisepersist:gwt-gradle-plugin:1.0.9";
 	static String androidPluginImport = "com.android.tools.build:gradle:3.4.1";
 	static String roboVMPluginImport = "com.mobidevelop.robovm:robovm-gradle-plugin:" + roboVMVersion;
-	static String moePluginImport = "org.multi-os-engine:moe-gradle:" + moeVersion;
+	static String moePluginImport = "org.multi-os-engine.community:moe-gradle:" + moeVersion;
 	
 	//Extension versions
 	static String box2DLightsVersion = "1.4";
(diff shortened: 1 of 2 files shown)


noblemaster:
The changes seem fine. However, is `org.multi-os-engine.community` now the official version?

obigu:
I think we should consider removing the Moe backend from libGDX (maybe we can move it somewhere else). The official version has been dead for a long time now and, even if it's good news for legacy projects that still use ut that somebody has forked it to make it work with the latest iOS versions, nobody is mantaining it properly on libGDX side. The reason Moe was added originally as an iOS backend was that RoboVM went closed source and when it was acquired by Microsoft and it looked like the best alternative. But years later the other iOS backend added at that time (MobiVM) is the only one that has been actively mantained and well tested by libGDX users.

Tom-Ski:
I agree; I don't think that it is currently suited to be here since we have a robust solution currently.  

Multiple options are good, but I don't think that is worthwhile for us to have the overhead of maintaining, testing, and ensuring compatibility of two complex backends and build infrastructures.

noblemaster:
I agree too, unless someone is willing to properly maintain it?

tommyettinger:
The current MOE community version looks like 1.6.0, https://discuss.multi-os-engine.org/t/moe-community-edition-gradle-plugin-sdk-1-6-0/2791 . Unfortunately, it seems like a no-go with libGDX because so many classes omit `@Override`, which is supposed to only matter as a compile-time check, not something that is ever required... But MOE 1.6.0 requires it. So if we merged this, we'd be stuck again on an old version of MOE, unless libGDX changed its coding style across the whole codebase (also all libraries would need the annotation; that would be *delightful* to fix...). I'm also in favor of both sticking with RoboVM and releasing 1.9.11, or 1.10.0, or whatever you want to call it. Release 1.fuzzy.puppy if you want; we haven't been using semantic versioning for a while.

MobiDevelop:
So, what's the plan here? Are we going to drop MOE? If so, who will pull the plug?

Should we move it out to a separate repo then delete from libgdx proper?

tommyettinger:
Does MOE (Community or the old version) even work on iOS 14? As I understand it, iOS 14 makes a lot of breaking changes and adds requirements that weren't present when MOE Community 1.5.0 was released, and I don't think it's feasible to update to MOE Community 1.6.0 or higher unless they backtrack on their Override requirement. We have RoboVM and it works, I don't know why we would need half-support for MOE.

I'd like to hear from anyone using MOE with libGDX and ask them why they prefer MOE to RoboVM. If we can't find anyone who prefers MOE in practice, I think it's safe to close this and remove the vestiges of MOE from the libGDX repo. Old stable releases will still exist, and if anyone can figure out how to get an app made with out-of-date tools on the notoriously stringent App Store, they can use MOE 1.4.0.

noblemaster:
Sorry, I meant to close this (not merge). I reverted it.

I think, let's consider removing MOE. Not much has happened with MOE in the past year. It seems to be going nowhere. I think it's probably best to remove for the time being unless someone steps up and takes responsibility?

MrStahlfelge:
When I tried two years ago, moe backend was behind robovm in terms of features. Nothing happened since then. I vote for removing it, Mario added it because RoboVMs future was not clear at that time. If we ever need it again, it is not much work to readd it from the history. For now, it is just confusing for new users.

noblemaster:
+1 to remove it. That would also mean we can remove a bunch of issues that all reference MOE.

PR #6018, opened 2020-04-24
Create interfaces for Audio and Input on Lwjgl3 backend

obigu:
#5985 Lwjgl3 backend refactor.

Lwjgl3 backend has a particularity where a `Lwjgl3Application` can have multiple `Lwjgl3Window` and each `Lwjgl3Window` has its own `Input` instance. This means that `Lwjgl3Window` needs either a mechanism to create `Input` implementations or the `Input` implementations are injected/provided externally. To minimize the amount of code changes I've opted for the first solution.

In order to keep consistency across backends and keep `createAudio()` and `createInput()` methods in `Lwjgl3Application`, `Lwjgl3Window` requires a reference to the `Lwjgl3Application` to invoke `createInput()`.

Not to fully couple `Lwjgl3Window` and `Lwjgl3Application` (and following other backends structure),  an interface called `Lwjgl3ApplicationBase` has been created to act as factory and is passed to the `Lwjgl3Window` constructor.


fgnm:
Is it possible to have Lwjgl3 backend as a `SNAPSHOT`?

PR #6024, opened 2020-04-27
Universal Logging Tool for LibGDX

shatterblast:
This will enable you to use:

```
Log.debug("Such and such.");
Log.error("Such and such.");
Log.info("Such and such.");
```
This will work in any build variant for LibGDX.  With the TAG constant, Android builds will have IntelliJ's Logcat type out:  LibGDX:  Such and such.

Other wise, normal output goes to the Console, where you can view your desired results.


Darkyenus:
How does this improve the current API? The point of the logging `TAG` is to show where a message comes from. Just printing `"LibGDX:  "` everywhere is worse than nothing, as it does not carry any additional information in any competently implemented system. It may help to differentiate libGDX output from random stdout/stderr output that could be confused with logging output, but that shouldn't exist in the first place.

The method signatures themselves aren't very useful either - for example of a good low-overhead design (IMO) check [slf4j](http://www.slf4j.org/apidocs/org/slf4j/Logger.html) (it has a way greater scope than what we actually need, but the core is there).

Even though the API is simple, it would still require documentation - for example it is not obvious that `format` refers to `String.format`.

shatterblast:
It's simplicity for the sake of simplicity.

MobiDevelop:
I don't think GWT supports `String.format(...)` so I think you'd have to find an alternative for that platform.

shatterblast:
I don't use GWT.  It's better to have some thing and build upon it later then not include it.

NathanSweet:
libgdx doesn't log much or at all itself. It provides a way for apps to do the logging. I think it's fine that app code does the logging, ie we don't really need a class like this in libgdx. We provide a cross platform way to log and we don't try to make a super great, multipurpose logging API, as many others have done that (over and over).

It would make more sense to make something for a library like slf4j that puts it's logging output through the libgdx methods. I think that doesn't belong in libgdx either though, it would be better as a separate project.

noblemaster:
I agree with Nathan. I think the way logging is done should be left to the user. There are tons of logging frameworks that you can use. Or just use the class you provided yourself.

PR #6026, opened 2020-04-28
Lazy loading DependenciesInjector

IlyaGazman:
Added a more efficient and easy way to inject dependencies that work in all the platforms.

Features

- Lazy loading. An instance of the library will be created only when you need it(by calling to inject)
- This is a static class so you can call it from anywhere in the code and no longer have to provide your library as part of the application constructor
- In case you need it you can call the inject multiple times to create multiple instances of your library.

# Usage 

The first step is to map your library in all the platforms that you implement it, like this:

Java 7:
```Java
        DependenciesInjector.map(MySuperClass.class, new LazyLoader<MySuperClass>() {
            @Override
            public MySuperClass load() {
                return new MyImplantation();
            }
        });
```
Java 8:
```Java
        DependenciesInjector.map(MySuperClass.class, MyImplantation::new);
```

Then when you need to use it in the main project call:
```Java
        MySuperClass mySuperClass = DependenciesInjector.inject(MySuperClass.class);
```

### DIFF
--- a/gdx/src/com/badlogic/gdx/DependenciesInjector.java
+++ b/gdx/src/com/badlogic/gdx/DependenciesInjector.java
@@ -1,0 +1,79 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx;
+
+import java.util.HashMap;
+
+/**
+ * A lazy load dependencies manager, the dependency will be created only when the inject method is called.
+ * Very efficient in saving memory and managing the loading time
+ *
+ * <br><br>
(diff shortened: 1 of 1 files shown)


yuripourre:
I am not sure if I get it. How it is currently used in Libgdx? It may be more useful as a library

IlyaGazman:
@yuripourre  Today Libgdx is using this guide: https://github.com/libgdx/libgdx/wiki/Interfacing-with-platform-specific-code

It is more complex and less efficient as it does not provide lazy load and require to add params to the main app constructor.

mgsx-dev:
There are several ways to interface with backends, some want to do like in the wiki, some want to use some static fields, some want something more advanced. TBH, i really think it's a client code concern.

As a side note, "injection" is maybe a bit misleading here, typical DI inject automatically via annotations or bytecode. Also, lazy instantiation is something you may not want in many case to avoid some side effects.

Since it's not something that will be used by the framework, maybe you can make it a library, a gist, or a wiki addition?

IlyaGazman:
I see what you mean. Do you have a better name suggestion for me?

Is there any special meaning for libraries in libgdx or you just suggesting
to make a repo for it?

How does libgdx is handling those kind of things? I want to make an image
loader utility for libgdx that can dynamically load images. It has a
special code for HTML and Java, my original idea was to use this
DependencyInjector solution but as you said it might not be the best fit.

mgsx-dev:
Since it's one simple class, i think a gist would be enough. However, i'm not sure it's the best place to talk about all this, maybe you can join [libgdx discord](https://libgdx.badlogicgames.com/community.html), we have a dedicated channel for contribution topics. :-)

NathanSweet:
It looks OK but I think doesn't need to be in libgdx. libgdx has quite a lot of code and more non-core functionality makes it harder to maintain. It would maybe make sense as a separate project.

PR #6032, opened 2020-05-01
Use better implementation to get iOS version

obigu:
iOS 8 added a more efficient and robust mechanism to get the iOS version that doesn't require parsing and splitting a String through `NSProcessInfo` (https://stackoverflow.com/questions/3339722/how-to-check-ios-version).

The `getVersion()` implementation on `IOSApplication` has been modified and the unnecessary/duplicated `getIosVersion()` has been deleted.


noblemaster:
Makes sense +1. Any reservations by anyone?

PR #6049, opened 2020-05-13
added GL_TEXTURE_CUBE_MAP_SEAMLESS

mgsx-dev:
By default Cubemaps are not seamless, it requires either GLES 3.2, OpenGL 3.2 or GL_ARB_seamless_cube_map in order to enable it.

```java
if(Gdx.graphics.getGLVersion().isVersionEqualToOrHigher(3, 2) || 
		Gdx.graphics.supportsExtension("GL_ARB_seamless_cube_map")){
	Gdx.gl.glEnable(GL30.GL_TEXTURE_CUBE_MAP_SEAMLESS);
}
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
 - Input Keycodes added: CAPS_LOCK, PAUSE (aka Break), PRINT_SCREEN, SCROLL_LOCK, F13 to F24, NUMPAD_DIVIDE, NUMPAD_MULTIPLY, NUMPAD_SUBTRACT, NUMPAD_ADD, NUMPAD_DOT, NUMPAD_COMMA, NUMPAD_ENTER, NUMPAD_EQUALS, NUMPAD_LEFT_PAREN, NUMPAD_RIGHT_PAREN, NUM_LOCK. 
   Following changes might be done depending on platform: Keys.STAR to Keys.NUMPAD_MULTIPLY, Keys.SLASH to Keys.NUMPAD_DIVIDE, Keys.NUM to Keys.NUM_LOCK, Keys.COMMA to Keys.NUMPAD_COMMA, Keys.PERIOD to Keys.NUMPAD_DOT, Keys.COMMA to Keys.NUMPAD_COMMA, Keys.ENTER to Keys.NUMPAD_ENTER, Keys.PLUS to Keys.NUMPAD_ADD, Keys.MINUS to Keys.NUMPAD_SUBTRACT
 - Added a QuadFloatTree class.
+- Desktop: Cubemap Seamless feature is now enabled by default when supported, can be changed via backend specific methods, see supportsCubeMapSeamless and enableCubeMapSeamless in both LwjglGraphics and Lwjgl3Graphics.
 
 [1.9.12]
 - [BREAKING CHANGE] iOS: Changed how Retina/hdpi handled on iOS. See #3709.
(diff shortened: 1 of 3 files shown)


noblemaster:
Should this really go into `GL30` (rather than let's say a `GL32` or `GL30Extension` or similar)? I am not very familiar with OpenGL 3.0/`GL30`. Are there other variables and functions that are not always supported but still available in `GL30`? If this this is a first, we might want to think about adding it differently?

Also, that said, maybe we should comment the variables that are only partly supported?

tommyettinger:
I'm pretty sure I brought the same thing up to @mgsx-dev and he had a great response; IIRC OpenGL 3.2 features are already used in GL30, plus some from even later, but I'd have to check the GL docs to be sure. I haven't seen mgsx online in a few days, so I hope he's doing OK; he might be busy.

mgsx-dev:
I had a second look and If i understand it correctly, libgdx GL30 means "GLES 3.0 like" which means it supports all GLES 3.0 features. As you can see in [desktop implementation](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL30.java#L38) some of these features requires OpenGL 3 or 4, for instance `glInvalidateSubFramebuffer` method is [part of GLES 3.0](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glInvalidateSubFramebuffer.xhtml) but [requires OpenGL 4.3](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glInvalidateSubFramebuffer.xhtml).

About `GL_TEXTURE_CUBE_MAP_SEAMLESS` feature, i was a bit wrong and this feature is a bit special, it seems to be automatic (driver requirement) since [GLES 3.0 page 152](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf) but is an option for OpenGL (desktop), available since OpenGL 3.2 or via extension.

To summarize things up (from what i read) : 
* GLES 2, WebGL 1 : no seamless cubemap at all.
* GLES 3, WebGL 2 : always seamless cubemap.
* OpenGL : optionnal flag to enable/disable it since OpenGL 3.2 or extension (it's disabled by default).

So it appears to be a desktop only thing and i'm not sure how to address that correctly. What i'm thinking is to enable it by default in desktop backends when possible (using the code i provided at first), i can't see a reason to have it disabled. what do you think?

noblemaster:
Yes, how about we add a Javadoc description to `GL_TEXTURE_CUBE_MAP_SEAMLESS` shortly explaining the limitations. We probably should comment all other ones that are only partially supported (but I guess this would go over this pull request).

Maybe we don't need to enable it automatically, but explain how to do it, i.e. add your code sample? One still would have to check if it's enabled anyway.

mgsx-dev:
Considering it's a desktop thing only, i preferred to handle this in desktop backends code directly since this toggle is not part of any GLES specifications.
User code can change it if needed but i doubt there are use cases for this. It's more like a bug fix in OpenGL which was implemented "de facto" for modern GL version (GLES 3.0, WebGL 2.0).
Impacts can only be noticeable with small cubemaps, i did a quick compairison by messing with FrameBufferCubemapTest. 
Disabled on the left, enabled on the right.
seamless

PR #6057, opened 2020-05-23
[Label] Set text to int value opt

MaanooAk:
I removed the `Integer.toString` and replaced it with the `StringBuilder.append` to avoid some object creations. Side effect, some redundant checks in `setText(CharSequence)` are no longer performed.

I also used the `StringBuilder.clear()` instead of the `StringBuilder.setLength(0)` as is more descriptive and more efficient.

I changed slightly the null check (which I now realize is irrelevant) to avoid some redundant checks.

I would also suggest that `setText(CharSequence)` also return a boolean like `setText(int)`.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Label.java
@@ -95,21 +95,25 @@ public class Label extends Widget {
 	 * @return true if the text was changed. */
 	public boolean setText (int value) {
 		if (this.intValue == value) return false;
-		setText(Integer.toString(value));
+		text.clear();
+		text.append(value);
 		intValue = value;
+		invalidateHierarchy();
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


noblemaster:
Does ```StringBuilder.clear()``` actually exist in Java? That said, shouldn't the checks fail?

MaanooAk:
@noblemaster This is the `com.badlogic.gdx.utils.StringBuilder`. Can you please explain in which checks are you referring.

noblemaster:
Never mind me then. I thought this was the default Java StringBuilder!

MobiDevelop:
The StringBuilder clear method just sets the length to 0, so not really an efficiency thing... But, it is more clear (heh) than setLength(0).

kangsatekhmer:
https://joker123gaming.co this game so fun
let play and win

kangsatekhmer:
lo guys, i just want you to know there is a game called [https://joker123gaming.co]Joker123(url) You can play this game where ever you are, and so simple to play.

PR #6061, opened 2020-05-31
Specify encoding for Base64Coder.encodeString(), as the rest of GDX does

tommyettinger:
Without specifying a String encoding for String.getBytes(), it uses a platform-dependent encoding that can break Base64 data transmitted between computers. Almost all of libGDX specifies "UTF-8" as the encoding; only a few tests don't (and currently Base64Coder.encodeString()).

This commit also adds Javadocs where they were missing for one encodeString() and makes the spelling of "Base64" consistently "Base64" and not sometimes "Base 64".

EDIT FOR FUTURE REFERENCE:

This PR should have no effect if you only use Base64Encoder's byte array encoding, or if you use String encoding with ASCII strings only. If you previously wrote Base64 data using `Base64Coder.encodeString()` and used non-ASCII characters, then that data was written with the platform's encoding, which can vary at least on Windows and Linux. If two different encodings are used for encoding and decoding, then the retrieved data won't be the same. This PR changes the default to use UTF-8 encoding for `encodeString()`, which is what Android and iOS already use, and will make it so the output of `encodeString()` is consistent across all platforms. If you want to continue to use the platform encoding, you can use this instead of calling `Base64Coder.encodeString(s, useUrlsafeEncoding)`:
```java
new String(encode(s.getBytes(), useUrlsafeEncoding ? Base64Coder.urlsafeMap : Base64Coder.regularMap));
```

Using the platform encoding may be a good idea if you need to save files onto one device only and need to maintain compatibility with earlier saved files. If the Base64-encoded files go between devices on different OSes, you should use UTF-8 or handle the conversion to bytes yourself.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -31,6 +31,7 @@
 - API Addition: TextField and TextArea are providing the protected method TextField#checkFocusTraverse(char) to handle the focus traversal.
 - API Addition: UIUtils provides the constants UIUtils#isAndroid and UIUtils#isIos now.
 - Fixed: The behaving of TextFields and TextAreas new line and focus traversal works like intended on all platforms now.
+- API Change: Changed Base64Coder#encodeString() to use UTF-8 instead of the platform default encoding. See #6061
 
 [1.9.10]
 - API Addition: Allow target display for maximization LWJGL3 backend
--- a/gdx/src/com/badlogic/gdx/utils/Base64Coder.java
+++ b/gdx/src/com/badlogic/gdx/utils/Base64Coder.java
@@ -29,6 +29,8 @@
 
 package com.badlogic.gdx.utils;
 
+import java.io.UnsupportedEncodingException;
+
 public class Base64Coder {
 	public static class CharMap {
 		protected final char[] encodingMap = new char[64];
@@ -76,11 +78,21 @@ public class Base64Coder {
 		return encodeString(s, false);
 	}
 
+	/** Encodes a string into Base64 format, optionally using URL-safe encoding instead of the "regular" Base64 encoding.
+	 * No blanks or line breaks are inserted.
+	 * @param s A String to be encoded.
+	 * @param useUrlsafeEncoding If true, this encodes the result with an alternate URL-safe set of characters.
+	 * @return A String containing the Base64 encoded data. */
 	public static String encodeString (String s, boolean useUrlsafeEncoding) {
-		return new String(encode(s.getBytes(), useUrlsafeEncoding ? urlsafeMap.encodingMap : regularMap.encodingMap));
+		try {
+			return new String(encode(s.getBytes("UTF-8"), useUrlsafeEncoding ? urlsafeMap.encodingMap : regularMap.encodingMap));
+		} catch (UnsupportedEncodingException e) {
+			// shouldn't ever happen; only needed because we specify an encoding with a String
+			return "";
+		}
 	}
 
-	/** Encodes a byte array into Base 64 format and breaks the output into lines of 76 characters. This method is compatible with
+	/** Encodes a byte array into Base64 format and breaks the output into lines of 76 characters. This method is compatible with
 	 * <code>sun.misc.BASE64Encoder.encodeBuffer(byte[])</code>.
 	 * @param in An array containing the data bytes to be encoded.
 	 * @return A String containing the Base64 encoded data, broken into lines. */
@@ -92,7 +104,7 @@ public class Base64Coder {
 		return encodeLines(in, iOff, iLen, lineLen, lineSeparator, charMap.encodingMap);
 	}
 
-	/** Encodes a byte array into Base 64 format and breaks the output into lines.
+	/** Encodes a byte array into Base64 format and breaks the output into lines.
 	 * @param in An array containing the data bytes to be encoded.
 	 * @param iOff Offset of the first byte in <code>in</code> to be processed.
 	 * @param iLen Number of bytes to be processed in <code>in</code>, starting at <code>iOff</code>.


NathanSweet:
Sounds OK, but we'd need to update CHANGES as it could break someone. It's preferable to use `StandardCharsets.UTF_8`, I think that's OK for our source level.

tommyettinger:
I tried `StandardCharsets`; the source level that the project was at didn't support it when I wrote this yesterday. The rest of libGDX doesn't use StandardCharsets, [Find in Files shows no results](https://i.imgur.com/iCmQYBm.png). StandardCharsets should be usable if we can use Java 7 ([Javadocs](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html)). Oracle hasn't provided any download of Java 7 [for 5 years now](https://en.wikipedia.org/wiki/Java_version_history#Java_7), but at least the libGDX project configuration I imported uses Java 6.

I'm not sure I follow your logic on CHANGES. If consistently using UTF-8 breaks anyone, they were already depending on platform-specific behavior (they would get different Base64 results on a MacOS machine and on a Windows machine, not to mention anyone in a non-English-language locale), so it doesn't break anything unless it was already broken. This PR would not change results from existing code for any encoding of ASCII text, nor would it change results from encodeString() if they were run on a machine that uses UTF-8 encoding. If you want to put something in CHANGES, go ahead, but I don't know what you have in mind.

obigu:
I use `Base64Coder` to encode to base64 the game state and preferences of my games so the encoded files are on many different devices and locales. I'm not sure this is the case but if changing this means that an update to the game that uses the new `Base64Coder` may cause progress loss on some locales/devices we definitely need, at least, to make a big warning in CHANGES.

Anuken:
> the source level that the project was at didn't support it when I wrote this yesterday

Also note that the minimum Android API level used by libGDX doesn't support `StandardCharsets` either - `API 19` is required.

> I use Base64Coder to encode to base64 the game state and preferences of my games so the encoded files are on many different devices and locales.

Unless you're using base64-encoded data for file*names* or URLs, I don't see `encodeString` being useful for game data. Perhaps a warning is still needed, but this particular method doesn't seem like it would be used very often (if at all).

mgsx-dev:
>  I don't see encodeString being useful for game data.

I use it to encode/decode binary player game data (compressed and/or crypted). I think it's a common use case to encrypt your game save data in order to prevent players to modify it.

obigu:
> Unless you're using base64-encoded data for file_names_ or URLs, I don't see `encodeString` being useful for game data. Perhaps a warning is still needed, but this particular method doesn't seem like it would be used very often (if at all).

I use it as a cheap obfuscation against script kiddies. I think your assumption is risky and the warning a must.

tommyettinger:
OK, then what about an encoding option? If encoding is really so critically important here, why isn't it configurable?

tommyettinger:
Also, @mgsx-dev , are you sure you use `encodeString()` to encode
> binary player game data (compressed and/or crypted)
? Why encodeString() instead of the encoders that take a byte array? The issue here only affects encodeString(), since it uses the OS' encoding and so can't really be used in a networked situation.

obigu:
@tommyettinger We're not opposing merging this PR (at least not me), I think it makes sense to be consistent, specially considering the PR has been motivated by a real issue. I'm just bringing attention to the importance of adding the CHANGES line like @NathanSweet suggested because the change may impact some users.

mgsx-dev:
@tommyettinger my bad i'm using encode(byte[]). Sorry guys.

tommyettinger:
Thanks @obigu ; do you have any suggestions for how the CHANGES should be worded? We know the conditions where this can't break anything (when the encoded String is ASCII-only, or when the previously used encoding was UTF-8), but AFAICT this only breaks compatibility if a previously-encoded String was always loaded on a machine without UTF-8 platform encoding (always using the same encoding) and used non-ASCII chars, then that same Base64-encoded String was decoded with a version of libGDX containing this PR. This isn't an empty set of users, so should there be an option to specify encoding directly, as a String? There's probably some quick way to get the current platform encoding, and that could be used to duplicate the old behavior as a fallback.

obigu:
@tommyettinger On CHANGES I would simply add the name of the method that changes and refer/link to this issue. Here in this issue I would add in the description which cases (what you've just said in the las comment about non ASCII characters) and ideally which platforms (Android for instance is UTF-8 only) this change could break backwards compatibility on.

NathanSweet:
> I tried `StandardCharsets`; the source level that the project was at didn't support it when I wrote this yesterday. 

Ah that's right, I guess we'll live with the charset lookup by name a while longer.

> I'm not sure I follow your logic on CHANGES.

Any change that can break existing code should be put in CHANGES. This gives people a first place to look for why their stuff broke when they updated libgdx.

> If consistently using UTF-8 breaks anyone, they were already depending on platform-specific behavior, so it doesn't break anything unless it was already broken. 

I don't follow. If code was relying on the platform encoding and we change it to UTF-8, that code may break, as unlikely as that may be. The platform encoding can be anything, eg GB2312. Relying on the platform encoding may not be a great idea, but doing so is not broken. Most often it is probably unintentional but can still be working for an app.

> do you have any suggestions for how the CHANGES should be worded?

Nothing elaborate is needed, eg: Changed Base64Coder.encodeString to use UTF-8 instead of the platform default encoding. #6061

> should there be an option to specify encoding directly, as a String?

I'd say yes only if we thought it was needed very often, eg some methods in FileHandle have a charset name string parameter. For this class I doubt it's needed very often and someone could do the conversion to bytes themselves and encode those, so I don't think we need it here.

tommyettinger:
CHANGES have been added to, almost exactly how you worded it (just edited the method reference syntax to match the rest of the file). It points to here, and I edited the first post so people will get a summary right away.

PR #6068, opened 2020-06-04
fix SphereShapeBuilder poles are now merged

mgsx-dev:
Previous algorithm didn't merge sphere poles and used quads to join them to the rings which produced lighting artifacts (because of normals) as you can see here (before fix on the left, after fix on the right) : 
Screenshot 2020-06-04 10:20:27

Manually moving vertices poles help to see the flat triangles : 
Screenshot 2020-06-04 10:21:54

You can also see how normals on poles was weird (before the fix on the right) : 
Screenshot 2020-06-04 10:23:58

UVs was also a bit off because of flat edge on pole. Here is UVs before the fix: 
Screenshot 2020-06-04 10:24:23

UVs are now like this (similar to default Blender UV sphere) : 
Screenshot 2020-06-04 10:24:32

Note that this change doesn't impact opened spheres (eg. when V angle range from 30 to 150).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,6 +32,7 @@
 - API Addition: UIUtils provides the constants UIUtils#isAndroid and UIUtils#isIos now.
 - Fixed: The behaving of TextFields and TextAreas new line and focus traversal works like intended on all platforms now.
 - API Change: Changed Base64Coder#encodeString() to use UTF-8 instead of the platform default encoding. See #6061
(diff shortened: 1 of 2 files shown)


mgsx-dev:
Note that i tried to fix also the other "TODO" about lines and points mode optimization. I get something working but it overcomplexify the code too much and doesn't worth it IMO (points and lines are probably not used a lot).

yuripourre:
This is interesting, having the images really helps to understand it.

PR #6073, opened 2020-06-08
Fix an error that occurs when loading models by AssetManager when the…

pkruczynski:
… program is run from a JAR file and the relative reference is used in the loaded file. For example, the G3DJ model refers to the texture file: ../..textures/texture.png. InternalFileHandleResolver encounters the path: assets/models/../../textures/texture.png that it can't handle when the resource being loaded from inside the JAR file. This problem does not occur when the file being read is available directly on FS. It can be bypassed in two ways: 1. always keep the textures in the same directory as the model (this solution is most often recommended on forums), 2. By creating AssetManager with the indication InternalFileHandleResolver, while overwriting method: FileHandle resolve (String fileName)

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/loaders/resolvers/InternalFileHandleResolver.java
+++ b/gdx/src/com/badlogic/gdx/assets/loaders/resolvers/InternalFileHandleResolver.java
@@ -20,9 +20,12 @@ import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.assets.loaders.FileHandleResolver;
 import com.badlogic.gdx.files.FileHandle;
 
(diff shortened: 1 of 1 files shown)


Darkyenus:
It is not possible to use `java.nio.file` because it is [not supported by GWT](http://www.gwtproject.org/doc/latest/RefJreEmulation.html) nor [by Android](https://developer.android.com/reference/java/nio/file/Paths) (at libGDX's min support level).

MobiDevelop:
There's likely a way to do what you want without nio. It sounds like a similar (same?) problem we solved in BaseTmxMapLoader, where we added a getRelativeFilePath method to handle `../` in paths.

PR #6074, opened 2020-06-09
Fix Android fast pause/resume bug

ufoq:
Bug is there for like 8 years...
I can easily reproduce it by double clicking screen-lock button on Xiaomi mi4 and mi8.

For some reason when you instantly resume activity that hasn't fully paused then internally android stops trigerring `onDrawFrame` in GL thread what causes deadlock, cos `synch.wait()` needs `notify` from `onDrawFrame`.
It can be solved by executing what we need in GL Thread with `GLSurfaceView#queueEvent`. It doesn't have requirements like usual `onDrawFrame` calls from `GLSurfaceView` (f.e. frame cap, some synchronization etc).

There were multiple PRs that tried to fix this but were dirty hacks that use executors (that may cause pause to be ran from non-gl thread) or used another locking techniques on top of Java synchronized/monitors, that could reason in other deadlock.
Related PRs: #4267 #5696 and more

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -355,14 +355,25 @@ public class AndroidGraphics implements Graphics, Renderer {
 			if (!running) return;
 			running = false;
 			pause = true;
+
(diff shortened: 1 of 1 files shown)


yuripourre:
I didn't check in a real device but the logic seems good (and clean) to me.

Issue #5159, opened 2018-04-04
In Android, constantly setting the lookAt method will affect the camera rotation.


jianwenzhou:
Issue details

Through the orientation sensor, set the LookAt method in the PerspectiveCamera to continuously set the X axis and Y axis. The Z axis does not change. After turning the mobile phone many times, the model will be found to be offset, similar to the camera rotation angle.

Reproduction steps/code

import com.badlogic.gdx.graphics.PerspectiveCamera;
import com.badlogic.gdx.math.Quaternion;
import com.richinfo.libgdx.listener.HeadRotationListener;

public class SensorCamera extends PerspectiveCamera implements HeadRotationListener {

    private volatile float[] lookAtVector = new float[3];
    private volatile float[] oldLookAtVector = new float[3];
    private float[] tempLookAtVector = new float[3];
    private OnSensorChangeListener mListener;

    public SensorCamera(float fov, int width, int height) {
        super(fov, width, height);
        this.near = 1f;
        this.far = 100f;
        this.position.set(0f, 0f, 12f);
        this.lookAt(0f, 0f, 0f);
        this.update();
    }

    public void render() { //Constantly calling
        this.lookAt(lookAtVector[0], -lookAtVector[1], lookAtVector[2]);
        this.update();
    }

    public void setLookAtVector(float x, float y, float z) {
        lookAtVector[0] = x;
        lookAtVector[1] = y;
        lookAtVector[2] = z;
    }

    public float[] getLookAtVector() {
        return lookAtVector;
    }

    @Override
    public void update() {
        super.update();
    }

    @Override
    public void update(boolean updateFrustum) {
        super.update(updateFrustum);
    }

    public void dispose() {

    }

    @Override
    public void onRotationChanged(float[] orientation) {
        mListener.onSensorChange(orientation[0], orientation[1]);
        if (orientation[0] > 120 && orientation[0] < 280) {
            lookAtVector[0] = 12;
        } else {
            double cosX = Math.cos(0.5f * Math.PI + (orientation[1] / 180f * Math.PI));
            double sinX = Math.sin(orientation[0] / 180f * Math.PI);
            lookAtVector[0] = (float) (8f * sinX * cosX);
        }

        double cosY = Math.sin(0.5f * Math.PI + (orientation[0] / 180f * Math.PI));
        double sinY = Math.cos(orientation[1] / 180f * Math.PI);
        lookAtVector[1] = (float) (8f * sinY * cosY);
[...]

PR #6078, opened 2020-06-14
Remove unnecessary normalization

yuripourre:
This PR will not break anything, it just removed unnecessary normalizations to reduce the number of operations (and errors) when lookAt method is called multiple times. It will avoid 1 square root calculation (2 if you call the update method).

This is an attempt to fix: #5159

### How to test
A good way to test it is to run the [SimpleDecalTest](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/SimpleDecalTest.java) (or any other test that calls lookAt) and check if the behavior is the same.

### Why it works?
The first line will produce a non-unit vector but the second line will calculate the perpendicular vector and normalize it. (as @obigu explained, the length of the vector will not affect the cross product).

```
tmpVec.set(direction).crs(up);
up.set(tmpVec).crs(direction).nor();
```
If you are not convinced yet, the same thing is done at: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1522

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Camera.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Camera.java
@@ -98,7 +98,7 @@ public abstract class Camera {
 	/** Normalizes the up vector by first calculating the right vector via a cross product between direction and up, and then
 	 * recalculating the up vector via a cross product between right and direction. */
 	public void normalizeUp () {
-		tmpVec.set(direction).crs(up).nor();
+		tmpVec.set(direction).crs(up);
 		up.set(tmpVec).crs(direction).nor();
 	}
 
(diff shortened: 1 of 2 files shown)


intrigus:
Why is this unnecessary?

obigu:
@intrigus On a cross product, the vectors magnitudes don't affect the direction of the resulting vector. I think there's no need to normalize the vector resulting from the first cross product because we only use it to find a perpendicular vector to it (and to `direction`).

yuripourre:
@obigu that's right, unless I can't see the real reason for that, the first normalization is unnecessary.

@intrigus thank you for asking, I updated the description a bit.

yuripourre:
@noblemaster, @mgsx-dev Can you test it, please?

yuripourre:
@noblemaster Thanks!

PR #6080, opened 2020-06-16
Fixed a few JavaDoc twerks

payne911:
The first two commits are irrelevant. (The second one reverts the first one.)

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Interpolation.java
+++ b/gdx/src/com/badlogic/gdx/math/Interpolation.java
@@ -19,19 +19,19 @@ package com.badlogic.gdx.math;
 /** Takes a linear value in the range of 0-1 and outputs a (usually) non-linear, interpolated value.
  * @author Nathan Sweet */
 public abstract class Interpolation {
-	/** @param a Normalized value (between 0 and 1). */
-	abstract public float apply (float a);
+	/** @param n Normalized value (between 0 and 1). */
+	abstract public float apply (float n);
 
-	/** @param a Normalized value (between 0 and 1). */
-	public float apply (float start, float end, float a) {
-		return start + (end - start) * apply(a);
+	/** @param n Normalized value (between 0 and 1). */
+	public float apply (float start, float end, float n) {
+		return start + (end - start) * apply(n);
 	}
 
 	//
 
 	static public final Interpolation linear = new Interpolation() {
-		public float apply (float a) {
-			return a;
+		public float apply (float n) {
+			return n;
 		}
 	};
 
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/math/Interpolation.java
```diff
@@ -294,14 +294,14 @@ public Bounce (int bounces) {
 		}
 
 		private float out (float a) {
```
crykn, line 296: ^You forgot to change this parameter.

Darkyenus:
I'm not a fan of these changes. I understand the motivation, but a *normalized value* is not really a thing (as far as I know) and I have never encountered this naming in the context of interpolation. `progress` (as used in `lerp()`) seems to fit better.

gdx/src/com/badlogic/gdx/math/Interpolation.java
payne911: Thanks, it's fixed!

I was meant to forget one since I've done those edits in GitHub directly.

payne911:
> I'm not a fan of these changes. I understand the motivation, but a _normalized value_ is not really a thing (as far as I know) and I have never encountered this naming in the context of interpolation. `progress` (as used in `lerp()`) seems to fit better.

This is where I took the inspiration from:

image

`progress` does not convey the idea that the input value should be clamped between 0 and 1, as is a requirement mentioned by the documentation:

image

The word `normalized`, however, does have a mathematical definition which recalls a lot what the input requires to be. No one there, for example, seems confused by the use of the word: https://stats.stackexchange.com/questions/70801/how-to-normalize-data-to-0-1-range

Many other suggestions over there: https://english.stackexchange.com/questions/275734/a-word-for-a-value-between-0-and-1-inclusive

NathanSweet:
Normalized may be slightly more descriptive (though the range to which you normalize something does not need to be 0-1) but I don't think it's very important here, since before and after the PR the javadoc still needs to say the range is 0 to 1 (funny enough the value can actually be outside the 0-1 range, if the results are what you want). Changing the variable names from `a` to `n` doesn't really make anything more clear, need better reasons to make changes. I'll merge the javadoc typo separately.

payne911:
> Normalized may be slightly more descriptive (though the range to which you normalize something does not need to be 0-1) but I don't think it's very important here, since before and after the PR the javadoc still needs to say the range is 0 to 1 (funny enough the value can actually be outside the 0-1 range, if the results are what you want). Changing the variable names from `a` to `n` doesn't really make anything more clear, need better reasons to make changes. I'll merge the javadoc typo separately.

Sure, but `alpha` is just plainly **wrong** as a parameter name. `norm` at least reuses a word employed elsewhere in the libGDX API (namely: `MathUtils#norm`) with a clear definition of the goal. `alpha` is generally understood by *any* programmer as relating to transparency, and that is also how that word is used throughout the libGDX API, *except* here.

The `n` thing is purely to reproduce what was being done with the reduce version one-letter-version of the primary parameter name (`alpha` -> `a`, thus `norm` -> `n`). It's a secondary aspect of the real goal of this PR.

Darkyenus:
Well, alpha can't be plainly wrong, because it does not actually mean anything, it is just a letter. It could just as well be called `x` or something. Equating alpha to transparency is so niche, that it does not even appear in [Alpha (disambiguation)](https://en.wikipedia.org/wiki/Alpha_(disambiguation)) page on Wikipedia.

PR #6087, opened 2020-06-24
Utility private constructor

payne911:
The `Intersector` class is a utility class: it only has `static` methods. Thus, it is misleading to provide the library users with a possibility to instantiate it. Moreover, I went ahead and "uniformalized" the JavaDoc. Thus:

1) **I replaced the implicit no-arg constructor by a private constructor.**
2) I made sure the format was the same for all the present JavaDoc, not adding or removing any extra details. (Separating description from params and returns by an empty space.)
3) I added `{@code ...}` where it made sense.
4) The method you might think I changed simply was formatted to use 'tabs', like the rest of this whole class.

**The main goal for this PR was basically: why allow users to do `Intersector myIntersector = new Intersector();` if having such an instance is completely useless?** That's why I'm adding a private constructor, as is usually recommended for Utility classes.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -36,6 +36,10 @@ public final class Intersector {
 	private final static Vector3 v2 = new Vector3();
 	private final static FloatArray floatArray = new FloatArray();
 	private final static FloatArray floatArray2 = new FloatArray();
+	
+	private Intersector() {
+		// utility classes shouldn't be instantiated
+	}
 
 	/** Returns whether the given point is inside the triangle. This assumes that the point is on the plane of the triangle. No
 	 * check is performed that this is the case.
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -65,7 +65,7 @@ public final class Intersector {
 		return true;
 	}
 
-	/** Returns true if the given point is inside the triangle. */
+	/** Returns {@code true} if the given point is inside the triangle. */
 	public static boolean isPointInTriangle (Vector2 p, Vector2 a, Vector2 b, Vector2 c) {
 		float px1 = p.x - a.x;
 		float py1 = p.y - a.y;
@@ -75,7 +75,7 @@ public final class Intersector {
 		return true;
 	}
 
-	/** Returns true if the given point is inside the triangle. */
+	/** Returns {@code true} if the given point is inside the triangle. */
 	public static boolean isPointInTriangle (float px, float py, float ax, float ay, float bx, float by, float cx, float cy) {
 		float px1 = px - ax;
 		float py1 = py - ay;
@@ -96,14 +96,25 @@ public final class Intersector {
 		return true;
 	}
 
-	/** Determines on which side of the given line the point is. Returns -1 if the point is on the left side of the line, 0 if the
-	 * point is on the line and 1 if the point is on the right side of the line. Left and right are relative to the lines direction
-	 * which is linePoint1 to linePoint2. */
+	/**
+	 * Determines on which side of the given line the point is.<p>
+	 * Left and right are relative to the line's direction which is linePoint1 to linePoint2.
+	 * 
+	 * @return {@code -1} if the point is on the left side of the line.<p>
+	 * {@code 0} if the point is on the line.<p>
+	 * {@code 1} if the point is on the right side of the line. */
 	public static int pointLineSide (Vector2 linePoint1, Vector2 linePoint2, Vector2 point) {
 		return (int)Math.signum(
 			(linePoint2.x - linePoint1.x) * (point.y - linePoint1.y) - (linePoint2.y - linePoint1.y) * (point.x - linePoint1.x));
 	}
 
+	/**
+	 * Determines on which side of the given line the point is.<p>
+	 * Left and right are relative to the line's direction which is linePoint1 to linePoint2.
+	 * 
+	 * @return {@code -1} if the point is on the left side of the line.<p>
+	 * {@code 0} if the point is on the line.<p>
+	 * {@code 1} if the point is on the right side of the line. */
 	public static int pointLineSide (float linePoint1X, float linePoint1Y, float linePoint2X, float linePoint2Y, float pointX,
 		float pointY) {
 		return (int)Math
@@ -111,9 +122,10 @@ public final class Intersector {
 	}
 
 	/** Checks whether the given point is in the polygon.
+	 * 
 	 * @param polygon The polygon vertices passed as an array
 	 * @param point The point
-	 * @return true if the point is in the polygon */
+	 * @return {@code true} if the point is in the polygon */
 	public static boolean isPointInPolygon (Array<Vector2> polygon, Vector2 point) {
 		Vector2 last = polygon.peek();
 		float x = point.x, y = point.y;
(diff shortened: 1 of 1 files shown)


payne911:
I also noticed some methods are package-private: I'm assuming that's how they were meant to be?

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -36,6 +36,10 @@
 	private final static Vector3 v2 = new Vector3();
 	private final static FloatArray floatArray = new FloatArray();
 	private final static FloatArray floatArray2 = new FloatArray();
+	
+	private Intersector() {
+		// utility classes shouldn't be instantiated
```
NathanSweet, line 41: Don't need a comment here.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -36,6 +36,10 @@
 	private final static Vector3 v2 = new Vector3();
 	private final static FloatArray floatArray = new FloatArray();
 	private final static FloatArray floatArray2 = new FloatArray();
+	
+	private Intersector() {
```
NathanSweet, line 40: You can run the libgdx formatter to fix up spaces vs tabs, etc automatically (I can tell you haven't because there's no space before `()`). However, running it manually on each file is not a good use of time. We have an issue to improve this: #5698

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -61,7 +65,7 @@ public static boolean isPointInTriangle (Vector3 point, Vector3 t1, Vector3 t2,
 		return true;
 	}
 
-	/** Returns true if the given point is inside the triangle. */
+	/** Returns {@code true} if the given point is inside the triangle. */
```
NathanSweet, line 68: I don't see this as being important. It is clear what it means either way. `true` is slightly prettier in javadocs but uglier in the code. I think it is much more common to read the code than to read the javadocs, so I would vote against using @code in most places.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -92,24 +96,36 @@ public static boolean intersectSegmentPlane (Vector3 start, Vector3 end, Plane p
 		return true;
 	}
 
-	/** Determines on which side of the given line the point is. Returns -1 if the point is on the left side of the line, 0 if the
-	 * point is on the line and 1 if the point is on the right side of the line. Left and right are relative to the lines direction
-	 * which is linePoint1 to linePoint2. */
+	/**
```
NathanSweet, line 99: IMO it is wasteful to use a whole line for this. The libgdx formatter will remove this line break. Also it's unreasonable to try to manually maintain code styling with a codebase as large as libgdx's.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -92,24 +96,36 @@ public static boolean intersectSegmentPlane (Vector3 start, Vector3 end, Plane p
...
+	 * Determines on which side of the given line the point is.<p>
+	 * Left and right are relative to the line's direction which is linePoint1 to linePoint2.
+	 *
```
NathanSweet, line 102: I am not usually a fan of blank lines in javadocs. When they are really needed I use a `<p>` on the line by itself (which is what you will get if you ran the formatter).

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -285,7 +304,8 @@ public static Vector2 nearestSegmentPoint (float startX, float startY, float end
 		return nearest.set(startX + t * (endX - startX), startY + t * (endY - startY));
 	}
 
-	/** Returns whether the given line segment intersects the given circle.
+	/** Checks whether the given line segment intersects the given circle.
```
NathanSweet, line 307: Starting javadoc with "Returns ..." is a common pattern and I think more descriptive than "Checks ..." (which is see is ued elsewhere in the file). Also by using returns in the text, an @return is not needed at all.

gdx/src/com/badlogic/gdx/math/Intersector.java
Monarchis: Yes i also recommend this command, it saves a lot of time.

payne911:
@NathanSweet I fixed the constructor according to your 2 related comments.

The rest of this PR can just be dropped. If it's just a matter of running a formatter on the file, I guess it'll be done at some point eventually. I personally couldn't quickly find that formatter in the Issue you've linked.

And I don't feel like arguing about `{@code ...}` and `Returns`.

gdx/src/com/badlogic/gdx/math/Intersector.java
NathanSweet: </p> is not needed at all.

gdx/src/com/badlogic/gdx/math/Intersector.java
NathanSweet: No, it is not required.

gdx/src/com/badlogic/gdx/math/Intersector.java
NathanSweet: These words don't make sense. Line breaks don't "get closed". No idea when ensuring completeness or a doc generator is.

gdx/src/com/badlogic/gdx/math/Intersector.java
payne911: In fact, I've had trouble in the past trying to use `</br>` in JavaDoc in the past, so I now avoid using closed tags (such as `</p>`).

Monarchis:
Sorry for deleting my comments and answering that late on a Sunday, but:
Until friday evening (MET) I had to figure out some huge bugs I noticed in my Matrix4 calculations I did in past.
At the evening I had a new look into this file, and notices some litte missing thing in the JavaDocs, so i started to test it.

Yesterday there was nice weather in southern germany so i used the time to enjoy the nature and I thought about the bugs I have, that I really have to fix promptly.

I felt free to do some additional commits, but in my opinion there should be some work for others left, so i dropped them.
Yesterday I also thought about the triangle's vectors and it's different meanings. This is the only commit left: be03bad0c2fa3db99168ae6539b0cc01d4462aec
Seems like the formatter works like intended.

And I did another commit: 84feb0db6a11982e3fc1a6d6883202645508cedb
Feel free to pick one of these.

About the CLA -> I'm not sure anymore if I already sent one.

Enjoy the rest of the weekend.

Monarchis:
Sorry for deleting my comments and answering that late on a Sunday, but:
Until friday evening (MET) I had to figure out some huge bugs I noticed in my Matrix4 calculations I did in past.
At the evening I had a new look into this file, and noticed some litte missing thing in the JavaDocs, so i started to test it.

Yesterday there was nice weather in southern germany so i used the time to enjoy the nature and I thought about the bugs I have, that I really have to fix promptly.

I felt free to do some additional commits, but in my opinion there should be some work left for others, so i dropped them.
Yesterday I also thought about some different meanings, and this is the only left: 2131c8726c6c492adb38fe0e50fa4b1035287907
Seems like the formatter works like intended.

And I did another commit: 0e94bf5af6d5b210b49297174c7b97619d7e30b7
Feel free to pick one of these.

About the CLA -> I'm not sure anymore if I already sent one.

Enjoy the rest of the weekend.

Monarchis:
Recently I found another thing in Intersector.class:

I saw that the `intersectRectangle`-method did not allowed the param intersections as beeing optional, changed this, see:
601879a077fcd1c4f409e582d8794e3bed1128a1

This is only the first, i saw some other things to change, please give me some time, i'll let you know if have some other changes or feel free to checkout [here](https://github.com/Monarchis/libgdx/tree/meaningful-fix)

Monarchis:
Ok i'll give it another try, yesterday i started fix my system I think i have some security issues, so need to do this immediatly, but i'll do it another day, when i have more time.

**WIP: [here](https://github.com/libgdx/libgdx/compare/master...Monarchis:meaningful-fix)

payne911:
#6110 has been created to focus on the main idea presented here, without being staggered by JavaDoc discussions which were unrelated to the actual goal of the PR.

This current PR can be closed: the JavaDoc modifications seemed rather unwanted.

payne911:
@Monarchis I hadn't realized you had been working from this current branch. Maybe you should move your work into a new branch?

Monarchis:
@payne911 Sure, that would make sense, I think I'll run into some merge conflicts, but probably not very complicated.

PR #6102, opened 2020-07-14
fix save method not numbering pages correctly if a page within only contains fonts

wheelergames:
When using `FreeTypeFontGenerator` to create fonts and then pack them in a pixmap, the following method is used 

`public synchronized Rectangle pack (Pixmap image) {
	return pack(null, image);
}`

Later when the rect is put onto rects with 

`
if (name != null) {
	page.rects.put(name, rect);
	page.addedRects.add(name);
}`

because there is no `name` it's not added to the rects. This is fine it still works etc, but in the PixmapPackerIO save method the file names are currently incremented and saved when `if (page.rects.size > 0) {` and therefore the pages are not incremented correctly, and the wrong files are used for images/fonts.

This only happens if the entire Pixmap is made up of fonts, which happens to me quite often. 

The fix increments and saves the Pixmap regardless of the size of rects.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPackerIO.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PixmapPackerIO.java
@@ -64,18 +64,9 @@ public class PixmapPackerIO {
 		Writer writer = file.writer(false);
 		int index = 0;
 		for (Page page : packer.pages) {
+			index++;
+			FileHandle pageFile = writeImageFile(file, index, parameters, page);
 			if (page.rects.size > 0) {
-				FileHandle pageFile = file.sibling(file.nameWithoutExtension() + "_" + (++index) + parameters.format.getExtension());
-				switch (parameters.format) {
-					case CIM:{
-						PixmapIO.writeCIM(pageFile, page.image);
-						break;
-					}
-					case PNG: {
-						PixmapIO.writePNG(pageFile, page.image);
-						break;
-					}
-				}
 				writer.write("\n");
 				writer.write(pageFile.name() + "\n");
 				writer.write("size: " + page.image.getWidth() + "," + page.image.getHeight() + "\n");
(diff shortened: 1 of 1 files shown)


wheelergames:
@tommyettinger so I haven't tested this locally per se, but I am using the fix live in my own app, and it fixes an otherwise crashing bug.

I pack fonts AND images into multiple pages, some pages contain just fonts others contain just images and others still contain a mixture of both.

To give a little more context as you seem to have grasped it but just to confirm.

Currently, the file names of each page are incremented ONLY if the page has rects (rects i think are seen as each individual sprite, so the letters in fonts and the images in, well images). As when you use FreetypeFontGenerator there are no rects created, because it doesn't take in a name String, this means the numbering of the pages gets out of sync and the texture atlas starts to look for things on the wrong page/file.

Does that make more sense?

Tom-Ski:
Can you give a sample that causes this? I have same use cases in production and do not see what you have reported.

wheelergames:
@Tom-Ski I'll look and try. Can you 100% confirm that none of your pages that are generated contain ONLY fonts. This is the case when this breaks.

wheelergames:
```
        generatedFonts = new HashMap<>();
        for (String fontName : FONTNAMES) {
            FreeTypeFontGenerator generator = new FreeTypeFontGenerator(Gdx.files.internal("fonts/ttf/" + getFontFilename(fontName) + ".ttf"));

            FreetypeFontLoader.FreeTypeFontLoaderParameter parameter = getFreeTypeParameter(fontName);
            parameter.fontFileName = "fonts/ttf/" + getFontFilename(fontName) + ".ttf";
            parameter.fontParameters.packer = packer;
            parameter.fontParameters.characters = "= _-,:!?/.'\"()[]{#@£$€*\\&%<>+0123456789aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ";
            if (fontName.equals(COUNT_DOWN)) {
                parameter.fontParameters.characters = "123";
            } else if (fontName.equals(COUNT_DOWN_ADVERT)) {
                parameter.fontParameters.characters = "12345";
            }
            generatedFonts.put(fontName, generator.generateFont(parameter.fontParameters));
        }
```
This is the code I use to generate my fonts in the packer. 

Following the code flow:
```
generateFont() -> generateData() -> createGlyph() -> packer.pack(mainPixmap)
```
this last command is the pack command that doesn't take a name field, therefore the page index will never be incremented.

Issue #3896, opened 2016-03-03
LibGDX's Android Keyboard input not working with the "Samsung Keyboard"


AshTNA:
#### Issue details

On Samsung's keyboard, there is a glitch where it does not process the command InputType.NULL sent by the Android App and therefore shows predictive text above the keyboard, which breaks LibGDX's keyboard input. 

I think this may be temporarily fixed if we can modify the InputType of the on screen keyboard. This would allow us to set the keyboard's input type as a visible password, which would prevent predictive text appearing.

I know this is not necessarily an issue with LibGDX, it is more Samsung's issue. It was reported to [Samsung a while ago ](http://developer.samsung.com/forum/board/thread/view.do?boardName=General&messageId=281092&frm=7&tagValue=keyboardbug&curPage=1)but I don't think they will be fixing it anytime soon.
#### Reproduction steps/code

Create a game that uses an InputProcessor to receive keyboard input.

Load the game on a Samsung Galaxy S5, S6, S6 Edge or Note 5 that uses the Samsung Keyboard app. Make sure predictive text is enabled on the keyboard settings.

Try to type something and no input will be registered.
#### Version of LibGDX and/or relevant dependencies

_Please provide the version(s) affected._
I have only tested with 1.6 and 1.9.3
#### Please select the affected platforms
- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

bazola:
I've just run into this issue on the S7.  If you tap on the checkmark or one of the other options in the predictive text, the text will be input into the TextField, but otherwise nothing shows up.

danielbse:
I have this issue aswell. Text appears when using the keyboard on Android Studios emulator but on my Phone, a S6 Edge you have to hit enter for it to show up.

EmmettGGI:
I also have this issue on my note 5. Please lmk if you guys find a work around or a fix is released. I think this is very important because a large amount of android users are on Samsung devices.

Ash98:
Yesterday I tried to contact Samsung via twitter about it but I think I confused them...we might get a better response if more people enquire them about it: 

https://twitter.com/AshleyABurnett/status/716392983986315264

As for a fix through LibGDX:I'm not sure I did this properly but I tried to edit the AndroidInput.java in gdx-backend-android,jar and append this code to the EditText variable called "input":

[...]

PR #6104, opened 2020-07-15
Samsung keyboard TextField fix

wheelergames:
On samsung keyboards onKey is not always fired for all key presses, in the current codebase. Because of this, when entering text in TextField, the typed letters don't appear until the keyboard is closed. The user has no idea that anything is being entered until after closing which is very poor UI.

The https://developer.android.com/reference/android/view/View.OnKeyListener#onKey(android.view.View,%20int,%20android.view.KeyEvent) is used inside AndroidInput but as the web page states

`Called when a hardware key is dispatched to a view. This allows listeners to get a chance to respond before the target view.

Key presses in software keyboards will generally NOT trigger this method, although some may elect to do so in some situations. Do not assume a software input method has to be key-based; even if it is, it may use key presses in a different way than you expect, so there is no way to reliably catch soft input key presses.`

It seems that this works for a lot of keyboards but not on the official Samsung one. The fix has been suggested on an Issue raised many years ago, but has come to my attention (and some others) recently.

#3896 (comment)

One person commented that this fix reverts if someone changes the language, though I have tried that and can't replicate it myself so I'm not sure if that's still an issue.

One thing I think that might be worth adding, that is not in this PR (I wouldn't know how to do it) is to put a check in to see if the device is Samsung or even if the keyboard is Samsung, that way it wouldn't affect everyone. Though when I've tested this fix against GBoard app it doesn't have any adverse affects.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/surfaceview/GLSurfaceView20.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/surfaceview/GLSurfaceView20.java
@@ -22,6 +22,7 @@ import android.content.Context;
 import android.graphics.PixelFormat;
 import android.opengl.GLSurfaceView;
 import android.os.SystemClock;
+import android.text.InputType;
 import android.util.Log;
 import android.view.KeyCharacterMap;
 import android.view.KeyEvent;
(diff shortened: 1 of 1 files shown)


bitdream:
I have tested on Samsung S10 device and it fixed my previous problem.
Besides, I tested on other no issue phones HW/KINDLE FIRE 8, it also worked.
If this issue is not fixed, the TextFiled is actully unusable.

NathanSweet:
It looks like voodoo to me, but if you guys say it's needed, OK!

PokeMMO:
This breaks non-latin text input.
I'm not sure the correct fix, TYPE_NULL (0) is seemingly the *correct* nput here, but Samsung is screwing it up.

We could detect samsung brand keyboards and apply this fix only to them. Via `Settings.Secure.getString(getContentResolver(), Settings.Secure.DEFAULT_INPUT_METHOD)` but this would still result in non-latin input not working on Samsung keyboards.

It seems like some applications have resorted to using a user provided configuration to determine which behavior to use.

https://github.com/termux/termux-app/commit/382da7e8f7374d9c073abe0a6ad94c0174830074
https://gist.github.com/lmller/dd143275f88cad8a21f96843cfca0d58

PR #6107, opened 2020-07-16
Lwjgl3GL20's glGetActiveUniform and glGetActiveAttrib report incorrect sizes and types

Haspamelodica:
#### Issue details
In the LWJGL3 backend, `Lwjgl3GL20#glGetActiveUniform` and `...ActiveAttrib` report incorrect sizes and types. This causes the size and type of uniforms and attributes reported by `ShaderProgram` to be incorrect when using the LWJGL3 backend.
From here on I will only describe the case of uniforms.

<details>
  <summary>Lwjgl3GL20#glGetActiveUniform implementation</summary>
  
  ```java
  public String glGetActiveUniform (int program, int index, IntBuffer size, Buffer type) {
    IntBuffer typeTmp = BufferUtils.createIntBuffer(2);
    String name = GL20.glGetActiveUniform(program, index, 256, size, typeTmp);
    size.put(typeTmp.get(0));
    if (type instanceof IntBuffer) ((IntBuffer)type).put(typeTmp.get(1));
    return name;
  }
  ```
</details>

For some reason, the call to `GL20#glGetActiveUniform` is not given the `type` argument from the caller, but instead a new buffer allocated in the method body. This buffer's first element is then put into `size`, overriding the `size` value obtained from the `GL20#glGetActiveUniform` call; and the buffer's second element is put into `type` of the caller. This seems to be incorrect, maybe this is an artifact from an older version.

Workaround: Manually call `org.lwjgl.opengl.GL20#glGetActiveUniform`. This is obviously not a very good solution because it breaks compability with other backends.

#### Reproduction steps/code
<details>
  <summary>Example code showing the bug and a workaround (for uniforms)</summary>
  
  ```java
  package net.haspamelodica.gdx;
  
  import java.lang.reflect.Field;
  import java.nio.IntBuffer;
  import java.util.Arrays;
  
  import org.lwjgl.BufferUtils;
  
  import com.badlogic.gdx.ApplicationListener;
  import com.badlogic.gdx.Gdx;
  import com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application;
  import com.badlogic.gdx.backends.lwjgl3.Lwjgl3ApplicationConfiguration;
  import com.badlogic.gdx.graphics.glutils.ShaderProgram;
  
  public class ShowLwjgl3GL20Bug implements ApplicationListener
  {
    @Override
    public void create()
    {
      ShaderProgram shader = new ShaderProgram(
          "void main() {gl_Position = vec4(0.0);}",
          "uniform float[5] exampleUniform; void main() {gl_FragColor = vec4(exampleUniform[0]);}");
      String uniformName = "exampleUniform[0]";
  
      if(!shader.isCompiled())
      {
        System.err.println("Error compiling shader:");
        System.err.println(shader.getLog());
      } else
      {
        System.out.println("GL_FLOAT: " + GL20.GL_FLOAT);
        System.out.println("shader.getUniforms(): " + Arrays.toString(shader.getUniforms()));
  
        //print size and type reported by the shader
        System.out.println("--- ShaderProgram");
        System.out.println("shader.getUniformSize(...): " + shader.getUniformSize(uniformName)); //5126 (GL_FLOAT) - should be 5
        System.out.println("shader.getUniformType(...): " + shader.getUniformType(uniformName)); //0 - should be 5126 (GL_FLOAT)
  
        int program = getProgramNumber(shader);
        int exampleUniformLocation = shader.getUniformLocation(uniformName);
        IntBuffer sizeBuf = BufferUtils.createIntBuffer(2);
        IntBuffer typeBuf = BufferUtils.createIntBuffer(2);
  
        //print size and type reported by Gdx.gl
        System.out.println("--- Gdx.gl");
        String uniformNameGdx = Gdx.gl.glGetActiveUniform(program, exampleUniformLocation, sizeBuf, typeBuf);
        System.out.println("Gdx.gl.glGetActiveUniform(...) returns: \"" + uniformNameGdx + "\"");
        System.out.println("size[0]: " + sizeBuf.get(0)); //5126 (GL_FLOAT) - should be 5
        System.out.println("size[1]: " + sizeBuf.get(1)); //0
        System.out.println("type[0]: " + typeBuf.get(0)); //0 - should be 5126 (GL_FLOAT)
        System.out.println("type[1]: " + typeBuf.get(1)); //0
  
        sizeBuf.position(0);
        typeBuf.position(0);
  
        //print size and type reported by LWJGL's GL20
        System.out.println("--- LWJGL's GL20");
        String uniformNameLWJGL = org.lwjgl.opengl.GL20.glGetActiveUniform(program, exampleUniformLocation, sizeBuf, typeBuf);
        System.out.println("org.lwjgl.opengl.GL20.glGetActiveUniform(...) returns: \"" + uniformNameLWJGL + "\"");
        System.out.println("size[0]: " + sizeBuf.get(0)); //5 - correct
        System.out.println("size[1]: " + sizeBuf.get(1)); //0
        System.out.println("type[0]: " + typeBuf.get(0)); //5126 (GL_FLOAT) - correct
        System.out.println("type[1]: " + typeBuf.get(1)); //0
      }
  
      shader.dispose();
      Gdx.app.exit();//don't keep the annoying black window
    }
  
    private int getProgramNumber(ShaderProgram shader)
    {
      //TODO is there a better way?
      try
      {
        Field programF = ShaderProgram.class.getDeclaredField("program");
        programF.setAccessible(true);
        return (int) programF.get(shader);
      } catch(Exception e)
      {
        throw new RuntimeException(e);
      }
    }
  
    @Override
    public void resize(int width, int height)
    {}
    @Override
    public void render()
    {}
    @Override
    public void pause()
    {}
    @Override
    public void resume()
    {}
    @Override
    public void dispose()
    {}
  
    public static void main(String[] args)
    {
      new Lwjgl3Application(new ShowLwjgl3GL20Bug(), new Lwjgl3ApplicationConfiguration());
    }
  }
  ```
</details>

libGDX version 1.9.10
LWJGL version 3.2.3

#### Stacktrace
_n/a_

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] MacOS
_Note: I only tested this on Windows, but the bug probably occurs on every platform when using the LWJGL3 backend._

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL20.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL20.java
@@ -288,31 +288,27 @@ public class AndroidGL20 implements GL20 {
 		return ints[0];
 	}
 
-	public String glGetActiveAttrib(int program, int index, IntBuffer size, Buffer type) {
-		//it is assumed that size and type are both int buffers of length 1 with a single integer at position 0
-		
-		//length
-		ints[0] = 0;
-		//size
-		ints2[0] = size.get(0);
-		//type
-		ints3[0] = ((IntBuffer)type).get(0);
-
+	public String glGetActiveAttrib(int program, int index, IntBuffer size, IntBuffer type) {
 		GLES20.glGetActiveAttrib(program, index, buffer.length, ints, 0, ints2, 0, ints3, 0, buffer, 0);
+
+		//size
+		size.put(ints2[0]);
+		//type
+		type.put(ints3[0]);
+
 		return new String(buffer, 0, ints[0]);
 	}
 
-    	public String glGetActiveUniform(int program, int index, IntBuffer size, Buffer type) {
-        	//length
-        	ints[0] = 0;
-        	//size
-        	ints2[0] = size.get(0);
-        	//type
-        	ints3[0] = ((IntBuffer)type).get(0);
+	public String glGetActiveUniform(int program, int index, IntBuffer size, IntBuffer type) {
+		GLES20.glGetActiveUniform(program, index, buffer.length, ints, 0, ints2, 0, ints3, 0, buffer, 0);
 
-        	GLES20.glGetActiveUniform(program, index, buffer.length, ints, 0, ints2, 0, ints3, 0, buffer, 0);
-        	return new String(buffer, 0, ints[0]);
-    	}
+		//size
+		size.put(ints2[0]);
+		//type
+		type.put(ints3[0]);
+
+		return new String(buffer, 0, ints[0]);
+	}
 
 	public void glGetAttachedShaders (int program, int maxcount, Buffer count, IntBuffer shaders) {
 		GLES20.glGetAttachedShaders (program, maxcount, (IntBuffer)count, shaders);
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java
@@ -352,21 +352,21 @@ class LwjglGL20 implements com.badlogic.gdx.graphics.GL20 {
 		EXTFramebufferObject.glGenerateMipmapEXT(target);
 	}
 
-	public String glGetActiveAttrib (int program, int index, IntBuffer size, Buffer type) {
+	public String glGetActiveAttrib (int program, int index, IntBuffer size, IntBuffer type) {
 		// FIXME this is less than ideal of course...
 		IntBuffer typeTmp = BufferUtils.createIntBuffer(2);
 		String name = GL20.glGetActiveAttrib(program, index, 256, typeTmp);
 		size.put(typeTmp.get(0));
-		if (type instanceof IntBuffer) ((IntBuffer)type).put(typeTmp.get(1));
+		type.put(typeTmp.get(1));
 		return name;
 	}
 
-	public String glGetActiveUniform (int program, int index, IntBuffer size, Buffer type) {
+	public String glGetActiveUniform (int program, int index, IntBuffer size, IntBuffer type) {
 		// FIXME this is less than ideal of course...
 		IntBuffer typeTmp = BufferUtils.createIntBuffer(2);
 		String name = GL20.glGetActiveUniform(program, index, 256, typeTmp);
 		size.put(typeTmp.get(0));
-		if (type instanceof IntBuffer) ((IntBuffer)type).put(typeTmp.get(1));
+		type.put(typeTmp.get(1));
 		return name;
 	}
 
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3GL20.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3GL20.java
@@ -352,20 +352,12 @@ class Lwjgl3GL20 implements com.badlogic.gdx.graphics.GL20 {
 		EXTFramebufferObject.glGenerateMipmapEXT(target);
 	}
 
-	public String glGetActiveAttrib (int program, int index, IntBuffer size, Buffer type) {
-		IntBuffer typeTmp = BufferUtils.createIntBuffer(2);
-		String name = GL20.glGetActiveAttrib(program, index, 256, size, typeTmp);
-		size.put(typeTmp.get(0));
-		if (type instanceof IntBuffer) ((IntBuffer)type).put(typeTmp.get(1));
-		return name;
+	public String glGetActiveAttrib (int program, int index, IntBuffer size, IntBuffer type) {
+		return GL20.glGetActiveAttrib(program, index, 256, size, type);
 	}
 
-	public String glGetActiveUniform (int program, int index, IntBuffer size, Buffer type) {
-		IntBuffer typeTmp = BufferUtils.createIntBuffer(2);
-		String name = GL20.glGetActiveUniform(program, index, 256, size, typeTmp);
-		size.put(typeTmp.get(0));
-		if (type instanceof IntBuffer) ((IntBuffer)type).put(typeTmp.get(1));
-		return name;
+	public String glGetActiveUniform (int program, int index, IntBuffer size, IntBuffer type) {
+		return GL20.glGetActiveUniform(program, index, 256, size, type);
 	}
 
 	public void glGetAttachedShaders (int program, int maxcount, Buffer count, IntBuffer shaders) {
--- a/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
+++ b/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
@@ -159,9 +159,9 @@ public class IOSGLES20 implements GL20
 	
 	public native int glGenTexture();
 
-	public native String glGetActiveAttrib ( int program, int index, IntBuffer size, Buffer type );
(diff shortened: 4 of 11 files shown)


Darkyenus:
You are right, the current implementation seems to be very strange. Since your fix mandates the use of `IntBuffer` anyway and this function is not used that often, maybe it would be a good idea to change the signature from `Buffer` to `IntBuffer`, to prevent any surprises. (But I don't know how will it affect other backends.)

Haspamelodica:
I am not sure if it's a good idea to change the interface of `com.badlogic.gdx.graphics.GL20#glGetActiveUniform`, because this 
could break application code. (Other backends won't be affected as far as I can see.) On the other hand, code handing a non-`IntBuffer` to `glGetActiveUniform` is broken either way, so it maybe would even help the developer to get a compile error with newer versions of Gdx, instead of exceptions at runtime.

Another possibility: Introduce a second variant of `glGetActiveUniform`, which takes an `IntBuffer` as the `type` parameter; and mark the variant expecting a generic `Buffer` as deprecated. (To avoid breaking other backends, this method would of course have to have a default implementation, which could just delegate to the `Buffer` variant.) I am not sure if this is a good idea, though; because then two methods with almost the same signature would exist; mix-ups are bound to happen.

Dealing with non-`IntBuffer`s by throwing a r untime exception is of course not a very good solution either, but `Lwjgl3GL20` does this in some other methods as well, for example in `glCompressedTexImage2D`. This is why I chose this solution to create this PR.

What do you think is the best way to deal with this problem?

By the way: `LwjglGL20`'s implementation (not the LWJGL3 one) contains exactly the same code. LWJGL2, in contrast to LWJGL3, seems to expect just one buffer of length 2. For this case the code does exactly the right thing. This sounds like the bug got introduced when migrating the LWJGL backend to LWJGL3. This seems to be what actually happened: See commit 1cb815170a2fd69558f737d92f5665c9d64cebd1.

Darkyenus:
IMO: Changing the signature is the most reliable and clean way to fix this and will give us the best result. I don't think that this particular function is used often, so I don't think that it will break many projects, especially when people should have been passing `IntBuffer`s in the first place and their code is probably wrong if they are not.

How is this handled in other backends:
- GWT: [casts to `IntBuffer` without check](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java#L560)
- LWJGL2: [casts to `IntBuffer` with a check](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java#L364) though the underlying GL function signature is different, this is probably what got us into this mess
- LWJGL3: [like LWJGL2, but, as this PR notices, it is incorrect](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3GL20.java#L363)
- Android: [unchecked cast to `IntBuffer`](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL20.java#L305)
- iOS MOE and RoboVM: Both are [native](https://github.com/libgdx/libgdx/blob/master/gdx/jni/iosgl/iosgl20.cpp#L857), so they don't care, but the underlying type is `GLenum`, but that is [32 bit](https://www.khronos.org/opengl/wiki/OpenGL_Type), so it assumes IntBuffer anyway. (It can also lead to memory corruption, because there are no size checks, but that is not something related to this PR.)

So it seems that a change to `IntBuffer` is the right thing to do.

Haspamelodica:
Should I include that change into this PR or should I open a new PR? And should I change other functions casting `Buffer`s as well? What about functions where one backend requires a specific `Buffer` type, but other backends do not?

Darkyenus:
I'd say change this PR so that all discussion is at the same place, and change only `glGetActiveUniform`, so that it is easier to review. Other functions may have a different tradeoffs and considerations, so let's keep it simple. Additionally, the other functions are probably correct (while this one is buggy on LWJGL3), so this is a bugfix PR and the rest will be just an enhancement.

Haspamelodica:
While updating the Android backend (specifically, `AndroidGL20`), I noticed this method also seems incorrect: It passes the old values in `size` and `type` to the GL function, instead of updating `size` and `type`.) Should I fix this in this PR, too?

Darkyenus:
You are right, both `glGetActiveUniform` and `glGetActiveAttrib` updates `size` and `type` as if it was an input parameter, not an output. I think it would fit into this PR nicely.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL20.java
```diff
@@ -288,31 +288,27 @@ public int glGenTexture () {
...
-		//it is assumed that size and type are both int buffers of length 1 with a single integer at position 0
-		
-		//length
-		ints[0] = 0;
+	public String glGetActiveAttrib(int program, int index, IntBuffer size, IntBuffer type) {
+		GLES20.glGetActiveAttrib(program, index, buffer.length, ints, 0, ints2, 0, ints3, 0, buffer, 0);
+
 		//size
-		ints2[0] = size.get(0);
+		size.put(ints2[0]);
```
Darkyenus, line 295: I think you should use absolute puts (with index 0) rather than relative. I think that is the convention? (The gets that this replaced were absolute.)

Haspamelodica: I am not sure if it is Java convention to use `Buffer`s absolutely. Many methods in the standard library use `Buffer`s relatively: for example `ByteChannel.read` or `.write`.
Other backends also do relative puts:
 * GWT [does relative puts](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java#L551)
 * RoboVM/Moe seem to do realtive puts: the JNI binding for glGetActiveUniform [calls getDirectBufferPointer](https://github.com/libgdx/libgdx/blob/master/gdx/jni/iosgl/iosgl20.cpp#L855), which [returns the address of the current position in the buffer](https://github.com/libgdx/libgdx/blob/master/gdx/jni/iosgl/iosgl20.cpp#L89), as far as I understand.
 * LWJGL [does relative (but incorrect) puts](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java#L359)
 * LWJGL3 does relative puts: `GL20C` [calls `MemoryUtils.memAddress`](https://github.com/LWJGL/lwjgl3/blob/master/modules/lwjgl/opengl/src/generated/java/org/lwjgl/opengl/GL20C.java#L1030), which [returns the address of the current position in the buffer](https://github.com/LWJGL/lwjgl3/blob/master/modules/lwjgl/core/src/main/java/org/lwjgl/system/MemoryUtil.java#L780)

I changed it to absolute puts for now, but I think maybe relative puts would be better.

Darkyenus: I thought it was a libGDX convention, but I don't know what is the correct way to do this. Anyway, good job.

Haspamelodica:
Do you think the absolute puts should be changed back to relative puts (in the Android backend), before merging this PR?
If yes, should I force-push to my fork to "delete" the latest commit or should I create a revert commit?

Darkyenus:
After re-reading your research comment, I think that you are right: the fact that it used absolute puts was another deviation from the norm. So you can probably revert it. (And feel free to squash all commits into one.)

The situation is still a bit complicated, because even if all backends respect the buffer's position, only those that call `put` move the position. (i.e. MOE/RoboVM/LWJGL3 keep the old position, all others advance position.) But I'm pretty sure that this is inconsistent all over libGDX, so I wouldn't worry about it in this PR.

PR #6109, opened 2020-07-16
[OisController] Handle return of methods

### LINKED ISSUE #6108: OisControllers ignores the return value of ControllerListener methods
#### Issue details
The methods in ControllerListener (eg, `buttonDown()`) let you return true to say the event is handled, in which case it's not handed on to any other listeners. The AndroidControllers class [correctly breaks from the loop](https://github.com/libgdx/libgdx/blob/f929278ff3b4bf1a73a7be07f498ccbf6e72dffa/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidControllers.java#L94) so that later listeners are not called, but the OisControllers code on Desktop [continues to call every single listener](https://github.com/libgdx/libgdx/blob/f929278ff3b4bf1a73a7be07f498ccbf6e72dffa/extensions/gdx-controllers/gdx-controllers-desktop/src/com/badlogic/gdx/controllers/desktop/OisControllers.java#L125), based on the initial size of the listeners array.

This is causing me crashes, because I'm calling Controllers.removeListener() inside a controller event, which changes the size of the array and causes it to read past the end of it since OisControllers uses the initial size of that array, not the current size, to see if it is finished.

```
Exception in thread "LWJGL Application" java.lang.IndexOutOfBoundsException: index can't be >= size: 2 >= 2
	at com.badlogic.gdx.utils.Array.get(Array.java:154)
	at com.badlogic.gdx.controllers.desktop.OisControllers$OisController$1.buttonPressed(OisControllers.java:125)
	at com.badlogic.gdx.controllers.desktop.ois.OisJoystick.buttonPressed(OisJoystick.java:53)
	at com.badlogic.gdx.controllers.desktop.ois.OisJoystick.update(Native Method)
	at com.badlogic.gdx.controllers.desktop.ois.OisJoystick.update(OisJoystick.java:83)
	at com.badlogic.gdx.controllers.desktop.ois.Ois.update(Ois.java:42)
	at com.badlogic.gdx.controllers.desktop.OisControllers$1.run(OisControllers.java:100)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.executeRunnables(LwjglApplication.java:267)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:214)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:128)
```

I'm returning true to say the event is handled, so it should stop, but that is ignored so it carries on regardless.

I think the fix should be a simple case of changing `allListeners.get(ii).buttonUp(OisController.this, buttonIndex);` to `if (allListeners.get(ii).buttonUp(OisController.this, buttonIndex)) break;` for each event type. That shouldn't be a breaking change as it's expected behaviour that an event won't be passed on if you return true. Apologies if I should just have submitted this as a pull-request, that's not something I've done before.

1.9.10, gdx-controllers, but it's still that way in the current master branch.

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

As a side-issue, the ControllerListener documentation repeatedly says "@return whether to hand the event to other listeners", which is unclear. It sounds like you should return true to hand the event to other listeners, when the opposite is the case.

yuripourre:
Closes #6108

### DIFF
--- a/extensions/gdx-controllers/gdx-controllers-desktop/src/com/badlogic/gdx/controllers/desktop/OisControllers.java
+++ b/extensions/gdx-controllers/gdx-controllers-desktop/src/com/badlogic/gdx/controllers/desktop/OisControllers.java
@@ -114,50 +114,50 @@ public class OisControllers {
 				public void buttonReleased (OisJoystick joystick, int buttonIndex) {
 					Array<ControllerListener> allListeners = manager.listeners;
 					for (int ii = 0, nn = allListeners.size; ii < nn; ii++)
-						allListeners.get(ii).buttonUp(OisController.this, buttonIndex);
+						if (allListeners.get(ii).buttonUp(OisController.this, buttonIndex)) break;
 					for (int ii = 0, nn = listeners.size; ii < nn; ii++)
-						listeners.get(ii).buttonUp(OisController.this, buttonIndex);
+						if (listeners.get(ii).buttonUp(OisController.this, buttonIndex)) break;
 				}
 
 				public void buttonPressed (OisJoystick joystick, int buttonIndex) {
 					Array<ControllerListener> allListeners = manager.listeners;
 					for (int ii = 0, nn = allListeners.size; ii < nn; ii++)
-						allListeners.get(ii).buttonDown(OisController.this, buttonIndex);
+						if (allListeners.get(ii).buttonDown(OisController.this, buttonIndex)) break;
 					for (int ii = 0, nn = listeners.size; ii < nn; ii++)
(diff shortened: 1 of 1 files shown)


yuripourre:
@NathanSweet this is a quick one. If you have some time to review, be welcome.

yuripourre:
@NathanSweet Thank you!

PR #6110, opened 2020-07-18
Private constructors utils

payne911:
#6087 was slightly "controversial" due to having its main goal being intertwined with a bunch of JavaDoc stuff.

This PR is to:

1.  Ensure that all utility classes have a private constructor to prevent users from instantiating those classes needlessly. (Basically, I replaced the implicit no-arg constructor by a private one.)
2.  Ensure that all utility classes are final to prevent users from needlessly extending them.


payne911:
I'm not quite sure why the very first commit's diff ended up being calculated as if I had redone the whole classes: all I did was use `Alt+Insert` in IntelliJ to get the constructors to be generated automatically before changing the `public` modifiers to `private` ones.

Darkyenus:
I think that the diffs are big because you have changed line endings. I suspect that this may happen because IntelliJ does a partial refactor of the class, which changes it, but I'm not sure. The PR with whitespace changes ignored looks OK. It would be great if you managed to fix that whitespace issue, so that it does not mess with file history.

payne911:
To support my claim, here is the diff for the ``GeometryUtils`` class as seen in IntelliJ:

image

Yet GitHub says I made over 500 modifications in the file...

Darkyenus:
Try `git diff` command. Or maybe your git linebreak settings are wrong, somehow. This whole linebreak thing is a mess.

payne911:
@Darkyenus confirmed to be a problem with linebreaks.

image

On my editor, it's configured to be `CRLF`. Not sure what it *should* be in the libGDX context...

Anyways, I merged `master` to pull changes from the `1.9.11` release. Took the time to create the new `CHANGES` section and fixed a typo in there too.

I'm not sure how I'm suppose to revert the commit properly. I tried to do it from within IntelliJ but that lead to many other issues.  Any ideas?

shathor:
Create a commit that reverts the lineendings. Use `git rebase -i` to move that commit one after your initial change commit and squash / fixup those two together.

shathor:
Usually my go to command is `git rebase -i` for those sort of issues. E.g. drop the Merge, create a commit that reverts the lineendings, move the commit down to your initial changes and squash / fixup the commits together. However as the PR is already open, I don't know how github handles a history rewrite.

MrStahlfelge (approved):
This looks good for me. If no one has a problem with forbidding instantiation of the classes, we could merge this.

You won't be able to change the file without the line ending problem, Git automatically converts crlf to lf nowadays on commit (which is good), but the current version of the file has crlf endings. You can switch the view by clicking on the settings icon and check "Hide whitespace changes".

payne911:
So I can't avoid this "pollution" of the logs?

(Will `git blame` point to me for those files? That would be rather unfortunate in terms of historical tracebacks.)

MrStahlfelge:
> (Will git blame point to me for those files which has their line-endings modified? That would be rather unfortunate in terms of historical tracebacks.)

AFAIK yes, but there's no way around this and we had this before. The only thing we could do is convert every file to lf file endings in a single commit.

payne911:
Ouch.

> The only thing we could do is convert every file to lf file endings in a single commit.

Maybe do that before, and then if I merge the commit, then I won't specifically be on that `git blame`. It'd allow you guys to potentially finally run a "Code Formatter" on the whole codebase, and have a meaningful commit message associated with it.

crykn (approved):
I'm in favour of introducing the private constructors as well, this prevents accidental mistakes. The line ending changes are unfortunate, but we can't do anything about that.

MrStahlfelge:
Looks like it's set, you can either clean up conflicts (addition to CHANGES not needed in my opinion) or close this PR @payne911 :)

payne911:
> addition to CHANGES not needed in my opinion

Oops. *Just* did that. Were you commenting because you saw it?

PR #6117, opened 2020-07-23
Remove constructor with preFill argument from Pool

obigu:
Unfortunately I've realised that the constructor based mechanisms that https://github.com/libgdx/libgdx/pull/5773 introduced to prefill a Pool is bad and should be removed (we can keep calling `fill()` after instantiation).

Calling a protected method (`newObject()`) on the constructor is an anti pattern and can cause state consistency issues as the object may not have been fully initialized. A simple example can be seen on `ParticleEffectPool` in which if we use the constructor passing `preFill()` as `true`, `newObject()` will be called before `effect`has been set throwing a `NullPointerException`.

I should have seen this issue before, sorry for that :(

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+[1.9.12]
+- API Change: Removed Pool constructor with preFill parameter in favor of using Pool#fill() method. See #6117
+
 [1.9.11]
 - Update to MobiVM 2.3.8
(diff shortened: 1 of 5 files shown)


NathanSweet:
Can you add a line to CHANGES?

obigu:
@NathanSweet Done, thanks

PR #6118, opened 2020-07-24
Cache Box2D Fixture filter data to avoid JNI overhead

obigu:
This PR fixes performance issues caused by JNI overhead in certain Box2D scenarios.

Box2D resolves if 2 objects should collide or not in `World#contactFilter`. Every time 2 fixtures overlap, Box2D needs to know wether they need to collide or not. In certain scenarios with lots of bodies, even if they don't collide with one another, that method will be called many times every time there's an overlap.

In the libGDX Box2D integration we have several ways to make the "should collide" resolution.
1. Mask/Category bits
2. Group id
3. Custom `ContactFilter`

Checking the `World#contactFilter` implementation we see that unless we use option 3, 2 JNI calls to `jniGetFilterData` will be made. The only thing `jniGetFilterData` does is to get the `Filter` data from the `Fixture`. The problem is the JNI calls cause quite an overhead which have a very important impact.

The solution is to cache the Filter data on the Java side of the `Filter`class.

I've created a modified version of `Box2DTest` to profile the issue that shows the problem with current libGDX by taking advatage of the fact that using a `ContactFilter`we avoid the JNI calls so we can switch it or on off. It starts with 100 boxes. Every time you tap the screen it adds 100 more boxes. https://github.com/obigu/libgdx/blob/box2dtests/tests/gdx-tests/src/com/badlogic/gdx/tests/Box2DTest.java

The results are very noticeable on a mobile device.

- You can switch between JNI calls (resolution 1) and No JNI calls (resolution 3) changing the constant `USE_CONTACT_FILTER` to `false` and `true`.
- There's always a small initial time after tapping in which fps drops until it stabilizes due to initialization/creation.

On an Android Pixel 3:
- 1000 boxes 60fps with USE_CONTACT_FILTER = true (no JNI calls)
- 500 boxes 60fps with USE_CONTACT_FILTER = false (JNI calls to `jniGetFilterData`)

Screenshot_20200724-193233

I've also confirmed the culprit is `jniGetFilterData` by using Android Profiler and analysing CPU usage.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
@@ -28,4 +28,10 @@ public class Filter {
 	/** Collision groups allow a certain group of objects to never collide (negative) or always collide (positive). Zero means no
 	 * collision group. Non-zero group filtering always wins against the mask bits. */
 	public short groupIndex = 0;
+
+	public void set(Filter filter) {
+		categoryBits = filter.categoryBits;
+		maskBits = filter.maskBits;
+		groupIndex = filter.groupIndex;
+	}
 }
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Fixture.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Fixture.java
@@ -31,6 +31,9 @@ public class Fixture {
 	/** the address of the fixture **/
 	protected long addr;
 
+	/** the fixture filter data, initialized lazily **/
+	private Filter filter;
+
 	/** the shape, initialized lazy **/
 	protected Shape shape;
 
@@ -141,6 +144,9 @@ public class Fixture {
 	 * awake. This automatically calls Refilter. */
 	public void setFilterData (Filter filter) {
 		jniSetFilterData(addr, filter.categoryBits, filter.maskBits, filter.groupIndex);
+		if (this.filter == null)
+			this.filter = new Filter();
+		this.filter.set(filter);
 	}
 
 	private native void jniSetFilterData (long addr, short categoryBits, short maskBits, short groupIndex); /*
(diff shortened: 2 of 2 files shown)


NathanSweet:
Should we remove the `short[] tmp` field? Seems like it'd be fine to allocate it the one time per fixture.

Actually, seems like the only way the filter data ever gets set is through Java. If so, do we ever need to call `jniGetFilterData`? Don't we already have the data in the `Filter filter` field?

Do we need the `Filter tmpFilter` field? `getFilterData` could just return its filter object, with javadoc that says if it's changed, `setFilterData` must be called. That seems like the simplest thing.

obigu:
@NathanSweet My initial implementation went on that direction but I opted for this implementation for a couple of reasons.

Regarding default values and the need for `jniGetFilterData`. When creating a Fixture, Box2D sets default values for filter data (haven't checked what they are but almost sure they are the same as the libGDX `Filter` initialization values). If we assume those values will never change in Box2D (and I admint, they will probably never do) we could get rid of `jniGetFilterData`. The problem I have with that is that I think we would be breaking the separation between the library and libGDX. Assuming values in libGDX that should be provided by Box2D couples both libraries in a fragile way.

And a similar argument of aiming for robustness goes with the `Filter` returned in `getFilterData`. `Filter` is mutable. If we return the `Fixture`'s `Filter` instance to the user and modifies it, it is causing an inconsistency between internal/native Box2D and Java states. Afaik filter data is only used by Box2D to evaluate the collision which is ultimately evaluated on Java side in `World#contactFilter` but do we want to make the API javadocs-level fragile?

I agree that the changes on the PR are not nice to look at but that's the best solution I could come up keeping backwards compatibility. 

An alternative if we accept breaking backwards compatibility is to make `Filter` immutable.

NathanSweet:
Cool, if you thought about all the things then OK. I don't have much skin in this particular game anyway, as I haven't used Box2d in the past decade or so.

> Assuming values in libGDX that should be provided by Box2D couples both libraries in a fragile way.
FWIW, we could set the defaults from Java rather than rely on whatever Box2d uses for defaults.

> If we return the Fixture's Filter instance to the user and modifies it, it is causing an inconsistency between internal/native Box2D and Java states.
Yep. My general theory for similar things is that if it's easy to make the API safe, do it. If doing that is awkward, a little gotcha is usually not a huge problem. It's impossible to keep a user from screwing things up and putting too much effort into making it idiot proof is usually not a good investment. There are similar situations elsewhere, eg scene2d.ui List `getItems`, Batch `getColor`, or others.

> An alternative if we accept breaking backwards compatibility is to make Filter immutable.
When things make sense to be immutable, it's great. Otherwise it tends to be a whole lot of jumping through hoops (and allocations).

I'm fine merging it as is, though as mentioned I'm not currently a Box2D user. Do any Box2D users have input?

crykn (approved):
This looks like a good change.

I think making Fixture immutable would make sense as well, but then we should also add a `setFilterData(short, short, short)` option.

SebastianSchlegel:
Hi there :-) I found this issue while searching for a reason why my box2d collision group filtering seems to be broken in libgdx 1.9.12. Is this the reason why this cache implementation has been reverted?

obigu:
Yes, unfortunately this PR messed up Box2d collision filtering on 1.9.12. We reverted it back for the time being (and probably an emergency release) and there's a proposed fixed here that may or may not be merged at some point https://github.com/libgdx/libgdx/pull/6262

PR #6122, opened 2020-08-03
[scene2d.ui] set culling area for scrollpane's widget if it only wrap a Group

RUSshy:
Solves https://github.com/kotcrab/vis-ui/issues/266

Maybe there is a better way to handle that on vis-ui part, but i'm pretty sure there is something else to do with the ScrollPane implementation, what do you think?


NathanSweet:
I think it doesn't make sense to add this special case. It is surprising behavior because why should the ScrollPane dig through the children of its widget? Why would we only handle the case when it's a group that has one child that is Cullable? The Cullable could be in 2 tables instead of one, or 100 tables, or could be something else entirely, something with multiple children, etc.

If you have a Cullable in a ScrollPane, it already receives the culling area. It makes sense for that Cullable to do whatever is necessary with the culling area (set it on a child or whatever else).

MobiDevelop:
So, what's the decision here? Seems like a veto from @NathanSweet. Can we close or continue discussion?

MrStahlfelge:
We can close, I also don't think it is a good idea to handle such things in the parent widgets.

PR #6126, opened 2020-08-10
Make external files work on Android 10 and up

MrStahlfelge:
Since Android 6, the use of the sd card is only possible by showing a runtime permission dialog. This is not easy to implement and somehow scary for most users. Since Android 10, access to the sd card is not possible at all.

This PR changes the used path to the app external storage dir that is intended for the app to write to for other apps to read from. No runtime or other permissions are needed to use it.


MobiDevelop:
Definitely should be noted in CHANGES.

noblemaster:
That makes sense. +1 for CHANGES and then let's merge.

MrStahlfelge:
Done

PR #6132, opened 2020-08-19
iOS RoboVM: Support hardware keyboard events

MrStahlfelge:
Since iOS 13.5, iOS gives us real key events from hardware keyboards. I've added that to our robovm backend.
It is downwards compatible, older iOS versions since 9 have the used callback but don't call it (it was used on tvOS only) and therefore on older iOS versions nothing happens. But the game must be build with iOS SDK 13.5 or newer, that means robovm 2.3.10 is needed and therefore the dependency changed.

Since key event handling is used very different in libGDX (keys vs keyTyped, polling vs event handling, not to speak about Scene2d text fields, I was not able to test every type of key input. However, I tested with my game, with InputTest and SoftKeyboardTest on iOS 12, 13.2 and 13.4. As this is an addition, situation is in any case better with than without it. :-)

EDIT: Can be tested conveniently on Simulator.

Hint: [iOS 14 will add the ability to access hardware keyboard connection state and handle mouse events/lock mouse cursor as well.](https://developer.apple.com/videos/play/wwdc2020/10617)

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -648,7 +648,7 @@ public class DefaultAndroidInput implements AndroidInput {
 
 	@Override
 	public boolean isCatchKey (int keycode) {
-		return keysToCatch.contains(keyCount);
+		return keysToCatch.contains(keycode);
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


noblemaster:
It's merged! Please update the CHANGES file to and indicate that RoboVM 2.3.10+ is required. - Thanks! 👍

MrStahlfelge:
Thanks, done!

PokeMMO:
https://github.com/libgdx/libgdx/blob/master/fetch.xml#L5 was not updated, which is causing ant/maven based build to fail.

MrStahlfelge:
I am sorry, was not aware of a second definition.

MrStahlfelge:
The dependencies are updated, unfortunately using snapshots does not work in fetch.xml :( 
It points only to release repo, but changing this to snapshot repo would not help because of the timestamps in the snapshots.

PR #6133, opened 2020-08-20
Improved TextField and TextArea text, cursor and selection rendering

piotr-j:
Improved TextField and TextArea text, cursor and selection rendering in addition to fixed pref height of TextArea when setPrefRows is used. 

Main issue was that TextArea used calculated textHeight for prefHeight instead of lineHeight from the font like the GlyphLayout. That resulted in prefHeight not matching actual text, depending on declared line height in the font. 
Another issue was selection/cursor drawing in TextArea that did not account for xAdvance of first glyph on each line. It used first glyph of the entire text for all lines which resulted in selection/cursor appearing in wrong planes on some lines. 
Additionally TextField height is based on lineHeight which results in it being less tall in most cases

Current TextField and TextArea | PR TextField and TextArea
:-------------------------:|:-------------------------:
text-area-old | text-area-new

See included `TextAreaTest3.java` for sample.

Let me know if I missed anything.


NathanSweet:
I don't believe the changes to TextField are an improvement. In my app:



In UITest:



Before the PR the space below and above the cap height is the same and the selection is centered vertically. After the PR there is space above the selection and the text is too high.

FWIW, here is a Windows text field which positions the text and selection similar to before this PR:

NathanSweet:
I reverted the TextField changes and removed the `lineHeight` field. I left the TextArea changes, though I don't believe the selection looks right there. For example, in TextAreaTest3 TextField looks like:

And TextArea looks like:

piotr-j:
I agree that the `TextField` in your example looks better, however I'm no so certain about the way it is achieved. Currently is is done by adding extra `-descent` to text height, which is a decent hack if the fonts line height is very close to its text height. Perhaps that's a common solution for this problem, but I think it was a reasonable assumption to let the font decide how tall a line should be. Existing solution is good enough, so Ill let it be.

Ill fix `TextArea` text location in a PR.

PR #6139, opened 2020-08-22
Add intersection between ray and obb

yuripourre:
Add method to check intersection between ray and oriented bounding box

I've been using this method for some years now, I didn't notice any issues yet.

Additional explanation at: http://www.opengl-tutorial.org/miscellaneous/clicking-on-objects/picking-with-custom-ray-obb-function/
Original code at: https://github.com/opengl-tutorials/ogl/blob/master/misc05_picking/misc05_picking_custom.cpp#L83


mgsx-dev (changes requested):
Overall looks good to me and this is a nice addition!, but few minor changes are required IMO : 
* using temp vectors instead of creating new ones (you can use existing ones : eg. tmp, tmp1, ...).
* avoiding hardcoded limits like 100000 and 0.001 : could we use Float.MAX_VALUE and MathUtils.FLOAT_ROUNDING_ERROR in this case?
* Applying formatter on your changes (and remove some empty lines)
Thank you :-)

yuripourre:
@mgsx-dev Thank you for the review. PR updated.

mgsx-dev (approved):
It looks good to me, thank you!

obigu:
As a nice to have it'd be great having a test for it (in `IntersectorTest`). `Intersector` methods are very testable and we have a history of wrong implementations. Also we get double coverage for half the work as we can use the same tests for both `intersectRayOrientedBounds` and `intersectRayOrientedBoundsFast `.

yuripourre:
@obigu that's a valid point, I still don't know the best cases to test, I will try to find some time next week to update it.

obigu:
@yuripourre If you don't find the time it's ok, it's not required, please just let us know.

PR #6144, opened 2020-08-28
Fix asset manager path conversion and unload ordering

Crunkle:
Uses `replace` to convert path separators in `AssetDescriptor` as this will avoid the pattern compile. Adds similar path separator conversion in `unload` to fix use of backslash in asset paths, as shown:

```java
AssetManager assetManager = new AssetManager();

// No runtime exception using expected paths
assetManager.load("resource/image.png", Texture.class);
assetManager.unload("resource/image.png");

// Runtime exception on unload (as unload does not convert)
assetManager.load("resource\\image.png", Texture.class);
assetManager.unload("resource\\image.png");
```

Reorders the `unload` method to look at loaded assets before removing from the pending queue. The current logic can cause unpredictable dangling assets in larger projects (e.g. when unloading and abandoning an asset from callback):

```java
AssetManager assetManager = new AssetManager();

// Add first asset to load queue and wait
assetManager.load("resource/image.png", Texture.class);
assetManager.finishLoading();

// Add identical asset to load queue
assetManager.load("resource/image.png", Texture.class);

// Unload first asset from storage (keep queue)
assetManager.unload("resource/image.png");

// Our first asset is left and second was removed instead
```

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetDescriptor.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetDescriptor.java
@@ -38,14 +38,14 @@ public class AssetDescriptor<T> {
 	}
 
 	public AssetDescriptor (String fileName, Class<T> assetType, AssetLoaderParameters<T> params) {
-		this.fileName = fileName.replaceAll("\\\\", "/");
+		this.fileName = fileName.replace('\\', '/');
 		this.type = assetType;
 		this.params = params;
 	}
 
 	/** Creates an AssetDescriptor with an already resolved name. */
 	public AssetDescriptor (FileHandle file, Class<T> assetType, AssetLoaderParameters<T> params) {
-		this.fileName = file.path().replaceAll("\\\\", "/");
+		this.fileName = file.path().replace('\\', '/');
 		this.file = file;
 		this.type = assetType;
 		this.params = params;
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -200,35 +200,41 @@ public class AssetManager implements Disposable {
 	/** Removes the asset and all its dependencies, if they are not used by other assets.
 	 * @param fileName the file name */
 	public synchronized void unload (String fileName) {
-		// check if it's currently processed (and the first element in the stack, thus not a dependency) and cancel if necessary
-		if (tasks.size() > 0) {
-			AssetLoadingTask currentTask = tasks.firstElement();
-			if (currentTask.assetDesc.fileName.equals(fileName)) {
-				log.info("Unload (from tasks): " + fileName);
-				currentTask.cancel = true;
-				currentTask.unload();
-				return;
-			}
-		}
-
-		// check if it's in the queue
-		int foundIndex = -1;
-		for (int i = 0; i < loadQueue.size; i++) {
-			if (loadQueue.get(i).fileName.equals(fileName)) {
-				foundIndex = i;
-				break;
-			}
-		}
-		if (foundIndex != -1) {
-			toLoad--;
-			loadQueue.removeIndex(foundIndex);
-			log.info("Unload (from queue): " + fileName);
-			return;
-		}
+		// convert all windows path separators to unix style
+		fileName = fileName.replace('\\', '/');
 
 		// get the asset and its type
-		Class type = assetTypes.get(fileName);
-		if (type == null) throw new GdxRuntimeException("Asset not loaded: " + fileName);
+		Class<?> type = assetTypes.get(fileName);
+
+		if (type == null) {
+			// check if it's currently processed (and the first element in the stack, thus not a dependency) and cancel if necessary
+			if (tasks.size() > 0) {
+				AssetLoadingTask currentTask = tasks.firstElement();
+				if (currentTask.assetDesc.fileName.equals(fileName)) {
+					log.info("Unload (from tasks): " + fileName);
+					currentTask.cancel = true;
+					currentTask.unload();
+					return;
+				}
+			}
+
+			// check if it's in the queue
+			int foundIndex = -1;
+			for (int i = 0; i < loadQueue.size; i++) {
+				if (loadQueue.get(i).fileName.equals(fileName)) {
+					foundIndex = i;
+					break;
+				}
+			}
+			if (foundIndex != -1) {
+				toLoad--;
+				loadQueue.removeIndex(foundIndex);
+				log.info("Unload (from queue): " + fileName);
+				return;
+			}
+
+			throw new GdxRuntimeException("Asset not loaded: " + fileName);
+		}
 
 		RefCountedContainer assetRef = assets.get(type).get(fileName);
 


NathanSweet:
Converting the slashes looks good.

Github is bad at diffs. Here's a better diff of your other change:


> Our first asset is left and second was removed instead

I don't understand this part. No asset is removed before it still has a ref count > 0.

When an asset is loaded a second time, instead of loading it twice, the ref count is incremented. When an asset is unloaded, the ref count is decremented, then only if the ref count is zero is it actually unloaded.

Your code loads an assets, then loads it again. At this point the second load is in the load queue -- it will be actually loaded (meaning the ref count will change `1 -> 2`) the next time the asset manager is updated. You don't update the asset manager though. You unload the asset once and that removes it from the load queue.

That works correctly without your proposed change:

```
assetManager.load("resource/image.png", Texture.class);
assetManager.finishLoading();
System.out.println("first load: " + assetManager.getDiagnostics());

assetManager.load("resource/image.png", Texture.class);
System.out.println("second load: " + assetManager.getDiagnostics());

assetManager.unload("resource/image.png");
System.out.println("unloaded: " + assetManager.getDiagnostics());
```

Output:

```
first load: resource/image.png, Texture, refs: 1
second load: resource/image.png, Texture, refs: 1
unloaded: resource/image.png, Texture, refs: 1
```

You have 2 loads and 1 unload, so the ref count should be 1.

If we add an update, then the 2nd load is processed and the ref count is 2 after the second load, then 1 after the unload:

```
assetManager.load("resource/image.png", Texture.class);
assetManager.finishLoading();
System.out.println("first load: " + assetManager.getDiagnostics());

assetManager.load("resource/image.png", Texture.class);
assetManager.update();
System.out.println("second load: " + assetManager.getDiagnostics());

assetManager.unload("resource/image.png");
System.out.println("unloaded: " + assetManager.getDiagnostics());
```

Output:

```
first load: resource/image.png, Texture, refs: 1
second load: resource/image.png, Texture, refs: 2
unloaded: resource/image.png, Texture, refs: 1
```

With your proposed change the behavior is wrong: the unload removes the asset even though it was loaded twice and removed only once.

If you are leaking assets in your app, you need to ensure you balance all loads with an unload.

I'll close and merge the slash improvements manually. If you can show there is an actual error, please post here and we'll reopen.

NathanSweet:
There are a number of methods that take a file name string. Needing to replace slashes for all of those seems a bit lame. It seems odd you can pass backslash to AssetDescriptor (via `load(String, ...)`) but can't pass backslash to any of the many AssetManager methods, like `get`.

Maybe it would be best if we never replace slashes for strings, and only replace slashes in the path from FileHandle in `AssetDescriptor(FileHandle,Class)`?

Crunkle:
@NathanSweet Thank you for adding the path conversion change. As for the other change: I understand the ref count logic keeping the resource in memory, but this causes the load callback to never be invoked on the second load. By removing the asset from the pending queue, the load callback will never be invoked for the pending item if removed between updates. This may not seem like a big issue, but can cause bugs in some games where the load callback is a vital component of logic.

One of the following needs to happen to fix this:

- On `load` the ref count is increased and callback is immediately invoked if already loaded (no queue)
- On `unload` the pending queue is not first priority and the loaded asset unloads then reloads (this fix)
- On `unload` the ref count does not change and callback is invoked for removed pending task

If this does not make sense, consider this minimized scenario from a larger project before disregarding:

```java
AssetManager assetManager = new AssetManager();

{
    // If true, draw and update the first zombie
    AtomicBoolean firstZombieLoaded = new AtomicBoolean();

    TextureLoader.TextureParameter parameter = new TextureLoader.TextureParameter();
    parameter.loadedCallback = (_assetManager, fileName, type) -> firstZombieLoaded.set(true);

    assetManager.load("resource/texture/internal/Act 1/cart_1.png", Texture.class, parameter);
}

// Load callback invoked, first zombie is now active
assetManager.finishLoading();

// A second zombie is spawned on this frame

{
    // If true, draw and update the second zombie
    AtomicBoolean secondZombieLoaded = new AtomicBoolean();

    TextureLoader.TextureParameter parameter = new TextureLoader.TextureParameter();
    parameter.loadedCallback = (_assetManager, fileName, type) -> secondZombieLoaded.set(true);

    assetManager.load("resource/texture/internal/Act 1/cart_1.png", Texture.class, parameter);
}

// First zombie dies on the same frame, and unloads
assetManager.unload("resource/texture/internal/Act 1/cart_1.png");

// Asset manager then updates again
assetManager.finishLoading();

// assetManager.getLoadedAssets() = 1 (but callback was not invoked!)
// assetManager.getQueuedAssets() = 0

// Second zombie is still waiting for their callback, but never receives one
```

Please note that the load callback may not be so trivial in a larger project and this is the simplest example to explain a scenario where removing from the pending queue causes bugs. The second zombie will never be notified that they were loaded and are ready to be updated.

Yes, the asset manager saves an unload with your current implementation - but it fails to invoke the load callback on the pending asset which is often a vital element of loading and can not simply be skipped as the current logic dictates. One of my three proposed solutions would be appreciated unless there are objections.

The asset manager could be redesigned to return wrapped assets and avoid the use of string lookup entirely, but this is a larger change.

Crunkle:
And here is the state of a game relying on those callbacks, now waiting for a callback that will never arrive:

```
zombie1 -> queued
assetManager::update
zombie1 -> loaded
zombie2 -> queued
assetManager::unload (zombie1 died)
zombie2 -> unloaded
assetManager::update
```

NathanSweet:
Ah, I see. I don't use the callback mechanism, instead I poll to check if the asset is loaded.

The problem with the callback is that `unload("image.png")` is ambiguous. Do you mean to unload the asset for the first or second zombie? If you are unloading for the first zombie, you'd expect the callback to be invoked so the second zombie stops waiting. If you are unloading for the second zombie, invoking the callback might be surprising.

We could leave it as is (callback invoked on asset load and ref count increment) and also invoke the callback when removing from the load queue if the asset is already loaded. Doing that makes some sense since the asset is indeed available, even though we don't know if the callback still cares.

>    On load the ref count is increased and callback is immediately invoked if already loaded (no queue)

It's a little weird to invoke the `finishedLoading` callback from `unload`, right before unloading it. We could invoke the callback immediately but the ramifications of that are not clear. It's designed so everything goes through the load queue and most blocking work happens in `update`. It may be surprising to have callbacks fired when calling `load` and not in the order they were loaded.

>    On unload the pending queue is not first priority and the loaded asset unloads then reloads (this fix)

This isn't acceptable because it breaks ref counting.

>    On unload the ref count does not change and callback is invoked for removed pending task

This seems to make the most sense, even though as mentioned it's a little weird to invoke the callback from `unload`. I pushed a commit for this, can you please try it out?

>    On unload the asset is added to an unload queue which only processes after load queue is empty

Not sure about this. I guess it complicates the (already complicated!) asset life cycle and would mean in some cases we load an asset unnecessarily, then immediately unload it.

Crunkle:
I agree, this is the least intrusive solution and seems to work well in both projects with no negative side effects. It definitely makes sense to invoke the callback instead of skipping it - much more predictable for game logic to follow.

It would be a nice feature to have the asset manager provide wrapped assets (i.e. `Asset<Texture>`) that can be loaded, cancelled, and unloaded selectively to remove the ambiguity.

All resolved now, thanks much.

Issue #5826, opened 2019-10-31
Lwjgl3 FPS oddities


Ktar5:
**Issue Details**
I'm using the LWJGL3 backend.

Previously, when using the LWJGL2 backend, I experienced similar experiences to #1386  and #3405. However, when in fullscreen, these issues would go away and I'd have a beautiful, smooth game. This was a constant bug over the course of time.

One of the above issues suggests moving to the LWJGL 3 backend. So I did, and have been testing it over the past week.

Instead of the odd stuttering, this time around I have FPS / delta time that goes wild, and stays below 60fps, even though my computer (a recent, expensive gaming laptop) is more than capable of hitting 60 fps. It's not only my game doing this, though, it seems all the LWJGL3 games running on my machine experience the same issue. HOWEVER, once again, in fullscreen, the issue doesn't exist at all, for all the games I've tried.

Here's an interesting gif showing some stats
https://i.imgur.com/9y9e1ZT.gifv

And another of someone else's game in LWJGL3 (same game in lwjgl2 gets smooth 60fps)
https://i.imgur.com/ftbOJ1R.gifv

Here is another of a LWJGL2 game, which shows the normal, expected behaviour
https://i.imgur.com/U0QzL5t.gifv

Here's a paste of a bit of deltaTime printout to show the variation
https://pastebin.com/tcxvSt5k

And heres the same printing out of deltaTime in fullscreen:
https://pastebin.com/eFiF3J51

When in fullscreen with vsync disabled and fps unlimited, I get over 2400 FPS, which is just something to show that my machine is more than capable of 60fps windowed.

We have tried vsync on, vsync off, Display.sync stuff from lwjgl2, manually setting vsync, profiling, and other countless efforts.

Is there any information you could possibly supply me on this topic?

**Expected results**
A smooth, consistent FPS in the range of 58-62

**Version of LibGDX**
1.9.11-SNAPSHOT, LWJGL3 backend

**Platforms tested on**
Windows

Yes i have the latest drivers.

tommyettinger:
[...]

PR #6150, opened 2020-08-30
Set vsync setting again after enabling fullscreen mode.

crykn:
On Windows, the vsync setting is ignored if fullscreen mode is enabled. Setting the vsync mode again fixes this, see #5826.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
@@ -163,10 +163,11 @@ public class Lwjgl3Window implements Disposable {
 		this.tmpBuffer2 = BufferUtils.createIntBuffer(1);
 	}
 
-	void create(long windowHandle) {
+	void create(long windowHandle, boolean vsync) {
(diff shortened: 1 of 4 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
```diff
@@ -163,10 +163,11 @@ public void run() {
 		this.tmpBuffer2 = BufferUtils.createIntBuffer(1);
 	}
 
-	void create(long windowHandle) {
```
tommyettinger, line 166: As long as we know the old `Lwjgl3Window.create(long)` was only used here (which is likely but not certain, since it's package-private and could be accessed by reflection), this probably won't break anything. It isn't a public API, so we don't need to maintain any sort of compatibility, but as long as this builds it means changing the method signature won't be an issue, I think.

crykn:
In multi-window setups this does not work as intended. 

- The initial vsync setting is set via the application config and thus saved in the application
- Each new window is handed this vsync setting upon creation. Each window has its own Lwjgl3Graphics instance (with `setVsync()`). However, calling `setVsync()` affects _all_ windows. The variable saving the state though is only set for the current windows' graphics instance.

PR #6155, opened 2020-09-01
remove Directional shadows deprecation

mgsx-dev:
DirectionalShadowLight has been deprecated for a long time (since the beginning), considered as "experimental". However, some people usie it, it works well and it didn't changed for years.

Most devs was confused about its deprecation state and thought it will be removed in a future version.

Also, in order to add improvements in the future (and probably mark some methods as deprecated), i thought it would be better to cleanup this to avoid confusion.


mgsx-dev:
@Tom-Ski i added you as reviewer because i think a maintainer have to decide if we can consider it as non experimental now.

Tom-Ski:
Yeah, it's being used as more of a disclaimer than anything. Happy to remove, docs should be enough

Tom-Ski (approved):

PR #6160, opened 2020-09-03
Create Pixmap from video element on gwt

SimonIT:
You guys can surely guess why it's needed :)


SimonIT:
Here a video that this change is working https://cdn.discordapp.com/attachments/348229413785305089/751348815910666260/LibGDX_application_-_Google_Chrome_2020-09-04_09-50-28.mp4

mgsx-dev:
for reference, https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Tutorial/Animating_textures_in_WebGL

@SimonIT could you please format only your changes? i understand files wasn't properly formatted before but it'll mess with history. If we really want to fix formatting for some files, i think it should be done in a separated PR.

Other than that it looks good to me. Thank you for your work on gdx-video :-)

mgsx-dev (changes requested):
please fix formatting :-)

MrStahlfelge (approved):
LGTM

Issue #6162, opened 2020-09-04
Wiki should be adapted to the new branding rules


payne911:
Considering the [new branding rules](https://libgdx.com/brand/):

> The name “libGDX” always starts with a lower-case ``l``, even at the beginning of a sentence. ``GDX`` is always written in upper-case letters.

Someone should figure out a way to automatically run a simple change throughout the entirety of our Wiki: swap all the instances of ``LibGDX`` for ``libGDX`` to be compliant with this rule.

crykn:
I cloned the wiki and replaced most of the mentions of libGDX. There are most certainly still some missing due to "libgdx" also being part of some URLs and find/replace being kinda difficult to use in that context. But those remaining instances can be replaced manually, whenever someones stumbles upon them.

PR #6163, opened 2020-09-04
Applying latest branding rule

payne911:
In the spirit of #6162.

### DIFF
--- a/CLA.txt
+++ b/CLA.txt
@@ -1,7 +1,7 @@
         Individual Contributor License Agreement ("Agreement") V2.0
                    http://www.apache.org/licenses/
 
-Thank you for your interest in libgdx. In order to clarify the intellectual 
+Thank you for your interest in libGDX. In order to clarify the intellectual 
 property license granted with Contributions from any person or entity, Mario  
 Zechner and Nathan Sweet ("Project Owners") must have a Contributor License Agreement ("CLA") 
 on file that has been signed by each Contributor, indicating agreement to the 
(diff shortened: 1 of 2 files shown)


MrStahlfelge:
Is this really the only incidence of wrong capitalization in the repo?

MrStahlfelge (changes requested):
I found a dozen other incidents of wrong capitalization. I think we should either change them all, or not.

crykn:
I agree. @payne911 can you do all of them, except the mentions in the README (those are handled by #6142)?

payne911:
> Is this really the only incidence of wrong capitalization in the repo?

Absolutely not: there are hundreds of instances.

I had started going through the whole code-base before giving up thinking random comments weren't all that important. But also because it's not just "replace all instances" because there are URLs, file-names and other logic-related Strings which misspelled ``libGDX``: potentially breaking the code for that didn't feel like it was worth it.

So in the end, I thought it'd be good to *at least* update the "legal" document's header.

cypherdare:
IMO class and member names shouldn't change. It's standard for coding naming conventions to modify the capitalization of words, whether they are proper nouns or not. No one should expect these to match the branded name.

MobiDevelop:
I personally don't think it is time we'll spent to go through the code and find places to change the capitalization. If you happen to be working in a file that should be updated, go for it.

MrStahlfelge:
@MobiDevelop I think that's a good idea. So I suggest wie close this PR and change the capitalization in #6142 ?

crykn:
Sounds reasonable not to change every instance of "libGDX" at once. But I still think we can merge this PR just to get those two documents (that won't be touched usually) up to date.

PR #6165, opened 2020-09-04
Gdxsetup config

MaanooAk:
Support more location for the preferences file, using the same conventions as git. With no user action the behavior remains the same.

- $HOME/.gdxsetup
- $XDG_CONFIG_HOME/gdxsetup/config
- $HOME/.config/gdxsetup/config

(The corresponding of XDG_CONFIG_HOME for windows and mac could be added)

Also minor visual improvements, mainly a tooltip for the ambiguous destination field (and updated the screenshot).

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
@@ -268,7 +268,7 @@ public class GdxSetupUI extends JFrame {
 					}
 				}, ui.settings.getGradleArgs());
 				log("Done!");
-				log("To import in Eclipse: File -> Import -> Gradle -> Gradle Project");
+				log("To import in Eclipse: File -> Import -> Gradle -> Existing Gradle Project");
 				log("To import to Intellij IDEA: File -> Open -> build.gradle");
 				log("To import to NetBeans: File -> Open Project...");
 				SwingUtilities.invokeLater(new Runnable() {
@@ -493,6 +493,12 @@ public class GdxSetupUI extends JFrame {
 				extensionPanel.setOpaque(true);
 				extensionPanel.setBackground(new Color(46, 46, 46));
 			}
+
+			nameLabel.setToolTipText("The name of the application used in gradle");
+			packageLabel.setToolTipText("The Java package name of the application");
+			gameClassLabel.setToolTipText("The name of the main Java class implementing ApplicationListener");
+			destinationLabel.setToolTipText("The root directory of the project to write the generated files to");
+			sdkLocationLabel.setToolTipText("The location of your Android SDK");
 		}
 
 		private void uiLayout () {
@@ -611,10 +617,10 @@ public class GdxSetupUI extends JFrame {
 			add(showMoreExtensionsButton, new GridBagConstraints(0, 12, 0, 1, 0, 0, CENTER, WEST, new Insets(10, 0, 10, 0), 0, 0));
 		}
 
-		File getDirectory () {
+		File getDirectory (String dialogTitle) {
 			if (System.getProperty("os.name").contains("Mac")) {
 				System.setProperty("apple.awt.fileDialogForDirectories", "true");
-				FileDialog dialog = new FileDialog(GdxSetupUI.this, "Choose destination", FileDialog.LOAD);
+				FileDialog dialog = new FileDialog(GdxSetupUI.this, dialogTitle, FileDialog.LOAD);
 				dialog.setVisible(true);
 				String name = dialog.getFile();
 				String dir = dialog.getDirectory();
@@ -623,7 +629,7 @@ public class GdxSetupUI extends JFrame {
 			} else {
 				JFileChooser chooser = new JFileChooser();
 				chooser.setFileSelectionMode(JFileChooser.DIRECTORIES_ONLY);
-				chooser.setDialogTitle("Choose destination");
+				chooser.setDialogTitle(dialogTitle);
 				int result = chooser.showOpenDialog(null);
 				if (result == JFileChooser.APPROVE_OPTION) {
 					File dir = chooser.getSelectedFile();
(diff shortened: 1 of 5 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -493,6 +493,12 @@ private void uiStyle() {
 				extensionPanel.setOpaque(true);
 				extensionPanel.setBackground(new Color(46, 46, 46));
 			}
+
+			nameLabel.setToolTipText("The name of the application used in gradle");
+			packageLabel.setToolTipText("The Java package name of the application");
+			gameClassLabel.setToolTipText("The name of the main Java class implementing ApplicationListener");
+			destinationLabel.setToolTipText("The root directory of the project to write the generated files to");
```
tommyettinger, line 500: This is phrased very awkwardly. If I didn't have the context that it was the destination directory, I would be lost. How about "Where this will generate the project's root directory"?

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -648,7 +654,7 @@ public void actionPerformed (ActionEvent e) {
 			sdkLocationButton.addActionListener(new ActionListener() {
 				@Override
 				public void actionPerformed (ActionEvent e) {
-					File path = getDirectory();
+					File path = getDirectory("Choose Android SDK");
```
tommyettinger, line 657: It took me a second to notice this, but this is a great change. The previous code should not have used "Choose destination" for the Android SDK, good catch.

extensions/gdx-setup/src/com/badlogic/gdx/setup/SetupPreferences.java
```diff
@@ -29,9 +29,30 @@
...
+	private static final File[] files = new File[] {
+		// $HOME/.gdxsetup
+		new File(new File(System.getProperty("user.home")), ".gdxsetup"),
+		// $CONFIG_HOME/gdxsetup/config
+		new File(new File(new File(findConfigHomePath()), "gdxsetup"), "config")};
+
+	private static final String findConfigHomePath () {
+		Map<String, String> env = System.getenv();
+		if (env.containsKey("XDG_CONFIG_HOME")) return env.get("XDG_CONFIG_HOME");
+		// TODO add windows and mac config home paths
```
tommyettinger, line 43: I'm not sure about this TODO; is this something you want to add to the PR or something that would be added after the PR by someone else? As I understand it, this step is optional anyway, and the code will work on Windows and Mac with the current preferences location.

extensions/gdx-setup/src/com/badlogic/gdx/setup/SetupPreferences.java
```diff
@@ -29,9 +29,30 @@
...
 import java.util.Properties;
 
-public class SetupPreferences {	
+public class SetupPreferences {
+
+	private static final File[] files = new File[] {
+		// $HOME/.gdxsetup
+		new File(new File(System.getProperty("user.home")), ".gdxsetup"),
+		// $CONFIG_HOME/gdxsetup/config
+		new File(new File(new File(findConfigHomePath()), "gdxsetup"), "config")};
```
tommyettinger, line 38: These should probably be in reverse order; because findFile() returns early if it finds a directory, and any users who have ever run gdx-setup in earlier versions will have a `.gdxsetup` folder in their home directory, the option to look in a configurable path is ignored for them.

extensions/gdx-setup/src/com/badlogic/gdx/setup/SetupPreferences.java
```diff
@@ -29,9 +29,30 @@
...
+		if (env.containsKey("XDG_CONFIG_HOME")) return env.get("XDG_CONFIG_HOME");
+		// TODO add windows and mac config home paths
+		return System.getProperty("user.home") + "/.config";
+	}
+
+	private static final File findFile () {
+		for (File i : files) {
+			if (i.exists()) return i;
+		}
+		return files[0]; // default to $HOME/.gdxsetup
```
tommyettinger, line 51: (If the array above has its order reversed, this would have to be `files[1]`.)

tommyettinger (changes requested):
I mostly have minor nitpicks, but there's some very good code in this PR too.

extensions/gdx-setup/src/com/badlogic/gdx/setup/SetupPreferences.java
MaanooAk: I know there is the env var APPDATA in windows, but I'm not a Windows user so I will leave it for any windows user that is as ** as me and cannot handle random files in their ~.

extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -493,6 +493,12 @@ private void uiStyle() {
 				extensionPanel.setOpaque(true);
 				extensionPanel.setBackground(new Color(46, 46, 46));
 			}
+
+			nameLabel.setToolTipText("The name of the application used in gradle");
+			packageLabel.setToolTipText("The package name of the application");
+			gameClassLabel.setToolTipText("The name of the main class implementing ApplicationListener");
+			destinationLabel.setToolTipText("The root directory of the project, it will be created if it does not exist");
```
MaanooAk, line 500: I simplified it a bit. The problem with "Where this will generate the project's root directory" (with me) is that suggests that the root directory will be generated, but what in fact it will be filled with files. So you need to set `path/to/projects/gamename` and not `path/to/projects/`

tommyettinger: Looks good.

tommyettinger (approved):

PR #6167, opened 2020-09-05
More efficient way of setting bone state 

MarcinSc:
Each matrix is copied in one go, rather than 1 element at a time.


mgsx-dev (approved):
looks good to me, thank you :-)

PR #6168, opened 2020-09-05
Adds scaleChanged callback to Actor.

ccmb2r:
Adds scaleChanged callback to Actor similar to the ones for size, position and rotation.

The lack of a scaleChanged callback seemed like an oversight to me that became apparent when drawing an Actor via a FrameBuffer that depended, among others, on an Actor's scale. This would help my and similar issues.

If required for the pull request to be accepted, I'll gladly sign the CLA.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -589,6 +589,10 @@ public class Actor {
 	protected void sizeChanged () {
 	}
 
+	/** Called when the actor's scale has been changed. */
+	protected void scaleChanged () {
+	}
+
 	/** Called when the actor's rotation has been changed. */
 	protected void rotationChanged () {
 	}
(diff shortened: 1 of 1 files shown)


mgsx-dev (approved):
looks good to me :+1: 
no need to sign anything :-)

MrStahlfelge:
Nothing to mind here, except that I am not sure if it is really **needed** or what it is needed for. 
`sizeChanged` is needed because size can changed by relayouting, I don't know any case where the scale is changed automatically. 

@NathanSweet what is your opinion here?

NathanSweet:
I vaguely remember the main reason these `*Changed` methods were added was because there are a number of ways to change the properties and it's a pain to hook them all, so this is a centralized place to hook.

I've only overridden `sizeChanged` once, ever, and have never used the others. It's true that size and position are affected by layout and rotation and scale usually aren't. We have `rotationChanged` though, so I guess `scaleChanged` is OK.

MrStahlfelge (approved):
So that's an approve by Nate, I checked the code and it is good for me, too.

PR #6169, opened 2020-09-06
Gwt Backend support resizable application

MrStahlfelge:
In the past, we had frequent questions how an application can be made resizable on GWT, that means taking up the available space in browser. This was solved eventually with some example code in HTMLLauncher.

However, this was more a workaround then a real solution, and I am moving the code from HTMLLauncher into the backend with this PR. It is easy to see that the behaviour is unaffected if we start a fixed size application with set width and height, therefore I mark this as non-breaking. 

I also improved the Gwt tests: there was no way to enter superdev debugging mode before, and the windowed mode test had a wrong button label often.

I tested resizeability with usePhysicalPixels on and off on Firefox Desktop, Chrome Desktop, Chrome Mobile, Firefox Mobile and Safari Mobile. I discovered during testing that Firefox Mobile's full screen mode is broken, but this is independant from this PR and I am confident to say that it's Firefox, not our code. Surprisingly, even Safari works well. :-)

Would be great if this would go into the snapshot before the jam starts!

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -44,6 +44,8 @@ import com.google.gwt.dom.client.Document;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.dom.client.Style;
 import com.google.gwt.dom.client.Style.Unit;
+import com.google.gwt.event.logical.shared.ResizeEvent;
+import com.google.gwt.event.logical.shared.ResizeHandler;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.HasHorizontalAlignment;
 import com.google.gwt.user.client.ui.HasVerticalAlignment;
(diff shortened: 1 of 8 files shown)


SimonIT (approved):
Works like a charm

mgsx-dev:
I had this kind of issue with my games on itch.io : when user click the fullscreen button (called from itch.io JS code), my viewport wasn't updated (probably missing a call to the resize method).
I don't know if this PR fixes this kind of issue, the case fullscreen switch is not done by libgdx code. If it's the case then, it's perfect :-) 

My workaround about this issue was : 
```java
        @Override
        public void onModuleLoad () {
            super.onModuleLoad();
            com.google.gwt.user.client.Window.addResizeHandler(new ResizeHandler() {
                  public void onResize(ResizeEvent ev) {
                	  final int width = ev.getWidth();
                	  final int height = ev.getHeight();
                	  getCanvasElement().setWidth(width);
                	  getCanvasElement().setHeight(height);
                	  Gdx.app.postRunnable(new Runnable() {
    					@Override
    					public void run() {
    						Gdx.app.getApplicationListener().resize(width, height);
    					}
    				});
                  }
                });
        }
```

MrStahlfelge:
Yes, it fixes it. :) The issue was caused by the HTMLLauncher onResize code:

        //     public void onResize(ResizeEvent event) {
        //         if (Gdx.graphics.isFullscreen())
        //             return;

It was needed because the launcher had no access to the canvas directly, but had to use setWindowedMode to resize it - which causes an exit of fullscreen mode.

It was exactly this kind of issue why I made this PR. :)

mgsx-dev:
That's great then :+1: :heart:

PR #6170, opened 2020-09-06
Improved GdxTests and enable distribution for desktop tests

mgsx-dev:
this PR adds some improvement to Gdx tests and enable distribution for both Lwjgl and Lwjgl3 launchers : 
* gl30 switch can now be enable via CLI argument (--gl30)
* non compatible tests are disabled : 
    * with Lwjgl3 launcher, items are disabled (grayed)
    * with Lwjgl2 launcher, items are removed (i didn't want to mess with Swing, please forgive me)
    * a new Annotation allow to say if a test requires gl30 or not.
* InstancedRenderingTest was missing in the list but it fails with Lwjgl2 (glVertexAttribDivisor method is not properly mapped apparently)

Notes: 
* this new mechanism (argv options and annotation) may be used for future needs, when application configuration needs some special setting for a particular test.
* gl30 options enable GLES 3.2 to be compatible with MacOSX.

Gdx tests dists can be build like this : 
* `./gradlew :tests:gdx-tests-lwjgl:dist`
* `./gradlew :tests:gdx-tests-lwjgl:dist`
* or simply `./gradlew dist` to build them all including gwt tests

I enable distribution to test some bugs only testable within a jar (relative path) so even if we don't plan to "distribute it publicly" it help for some fixes. Also, gl30 option was a bit annoying before, having to temporarily change the code...

### DIFF
--- a/tests/gdx-tests-android/assets/data/shaders/instanced-rendering.frag
+++ b/tests/gdx-tests-android/assets/data/shaders/instanced-rendering.frag
@@ -1,5 +1,3 @@
-#version 300 es
-
 precision mediump float;
 
 in vec4 v_color;
--- a/tests/gdx-tests-android/assets/data/shaders/instanced-rendering.vert
+++ b/tests/gdx-tests-android/assets/data/shaders/instanced-rendering.vert
@@ -1,5 +1,3 @@
-#version 300 es
-
 in vec4 a_position;
 in vec2 i_offset;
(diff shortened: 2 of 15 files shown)


crykn:
Looks good to me!

MrStahlfelge (approved):
I like the changes. Given the fact that only tests are concerned, there should be no problem with merging it.

PR #5369, opened 2018-09-10
Sprite class optimization

riki137:
After getting a suspicion, i modified my in-game code like this: (where o is a variable that gets incremented every frame) (this is part of a render() method)
```kotlin
if (o++ % 2 == 0) {
            shadow.setPosition(render.x, render.y - 3.px)
            body.setPosition(render.x + 4.5f.px, render.y + .5f.px)
            shirt.setPosition(render.x + 6f.px, render.y + 8f.px)
            pants.setPosition(render.x + 6f.px, render.y + 2f.px)
            if (!isCustomer) card.setPosition(render.x + 7.5f.px, render.y + 11.5f.px)
            head.setPosition(render.x + 3.5f.px, render.y + 13.5f.px + breathingY)
            face.setPosition(
                render.x + 5.px + bodyPartDiff.x * .5f,
                render.y + 15.px + bodyPartDiff.y * .5f
            )
            boots.rotation = MathUtils.sin(bootsSin) * 2f
            boots.setPosition(render.x + 4.5f.px, render.y)
        } else {
            shadow.setBounds(render.x, render.y - 3.px, shadow.width, shadow.height)
            body.setBounds(render.x + 4.5f.px, render.y + .5f.px, body.width, body.height)
            shirt.setBounds(render.x + 6f.px, render.y + 8f.px, shirt.width, shirt.height)
            pants.setBounds(render.x + 6f.px, render.y + 2f.px, pants.width, pants.height)
            if (!isCustomer) card.setBounds(render.x + 7.5f.px, render.y + 11.5f.px, card.width, card.height)
            head.setBounds(render.x + 3.5f.px, render.y + 13.5f.px + breathingY, head.width, head.height)
            face.setBounds(
                render.x + 5.px + bodyPartDiff.x * .5f,
                render.y + 15.px + bodyPartDiff.y * .5f,
                face.width, face.height
            )
            boots.rotation = MathUtils.sin(bootsSin) * 2f
            boots.setBounds(render.x + 4.5f.px, render.y, boots.width, boots.height)
        }
```
and ran the profiler:
image
my suspicions were confirmed - using `setBounds()` for changing position of a `Sprite` is faster than using `translate` with diff calculation - probably because `translate()` uses (almost?) twice as much memory access.

I've also modified `Sprite.setBounds()` and `Sprite.setSize()` function but i haven't profiled that. I just hope you guys agree it should make it faster in certain cases.

PR #6171, opened 2020-09-06
Sprite minor performance improvements

mgsx-dev:
I didn't benchmarked it but it sounds self speaking to me. Not sure it'll improve performance drastically but i think the calls are more consistent now.
This is a replacement for PR #5369


tommyettinger (approved):
Looks like a clear improvement; the vertices are set anyway when the Sprite is drawn (if it's dirty), so setting them twice in the old version was a total waste of time. Do we have any tests for Sprite behavior? It would be reassuring just to know that rotation, scaling, origin change, and any other vertex-based changes (including color?) have the same result.

mgsx-dev:
yeah : `./gradlew launchTestsLwjgl --args="SpriteBatchTest"`

PR #6172, opened 2020-09-07
Minor MathUtils constants improvements

obigu:
Minor cosmetic improvements to some of the `MathUtils` public constants not to use hardcoded values. All constants have the same value as before.

I've also added a new constant `MathUtils.HALF_PI` which doesn't introduce extra floating precision in comparison to using `MathUtils.PI / 2f` but it's a common enough value to have it there.

NOTE: I had originally changed `radiansToDegrees` and `degreesToRadians` to improve the float precision error but, even if technically the value is closer to the real one, when working with floats for everything (when using `MathUtils` API), conversions that you'd expect to be exact such as `MathUtils.PI / 6f  * radiansToDegrees = 30`  ended up having a precision error like `30.000001`.


MrStahlfelge (approved):
Looks good to me, no mistake spotted.

PR #6173, opened 2020-09-07
GWT PreloaderBundleGenerator don't use Apache Commons 

MrStahlfelge:
to calculate MD5 hash, relates to #4314

Commons is on Classpath for Gradle, but not for Ant. It is not really needed, so replace it with own implementation.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
@@ -28,6 +28,7 @@ import java.util.Map.Entry;
 
 import com.badlogic.gdx.backends.gwt.preloader.AssetFilter.AssetType;
 import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.StreamUtils;
 import com.google.gwt.core.ext.BadPropertyValueException;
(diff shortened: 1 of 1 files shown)


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/FileWrapper.java
```diff
@@ -196,11 +196,17 @@ public String readString (String charset) {
 	/** Reads the entire file into a byte array.
 	 * @throw GdxRuntimeException if the file handle represents a directory, doesn't exist, or could not be read. */
 	public byte[] readBytes () {
+		return readBytes(read());
+	}
+
+	/** Reads the entire file into a byte array.
+	 * @param input inputstream to use. The inputstream is closed after use.
+	 * @throw GdxRuntimeException if the file handle represents a directory, doesn't exist, or could not be read. */
+	public byte[] readBytes (InputStream input) {
```
NathanSweet, line 205: Could maybe use StreamUtils `copyStreamToByteArray`?
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/StreamUtils.java#L83

MrStahlfelge: Ah yes, that would be an even better way.

MrStahlfelge:
Now it's an easy one line change. :) Thanks for the hint @NathanSweet

Issue #4489, opened 2016-12-16
DelaunayTriangulator is occasionally incorrect (6-point example)


SpexGuy:
#### Issue details
I am attempting to use the `DelaunayTriangulator` to calculate adjacency information in a voronoi diagram.  Unfortunately, it appears that the triangulator occasionally triangulates quads in the wrong direction.  Here's a picture of one of the failing cases I'm encountering:

![delaunay_fail](https://cloud.githubusercontent.com/assets/2281818/21246526/f6f9abb0-c2ee-11e6-863f-5d43f53d1001.png)

The background is the rasterized voronoi of the points, to emphasize the problem.  The quad made up of points `(0, 3, 1, 2)` should be triangulated into `(0, 3, 1)` and `(1, 2, 0)`, but instead it gets cut in the other direction.  Adding another point somewhere else in the diagram (even far away) can sometimes flip this particular quad to be correct.

#### Reproduction steps/code
This test case reproduces the issue exactly as shown in the picture.
```java
import com.badlogic.gdx.math.DelaunayTriangulator;
import com.badlogic.gdx.utils.ShortArray;

public class DelaunayTest {

	public static void main(String[] args) {
		float[] points = new float[12];
		points[0] = (float) 2000.7620849609375;
		points[1] = (float) 3044.865234375;
		points[2] = (float) 3121.152099609375;
		points[3] = (float) 2331.909912109375;
		points[4] = (float) 1909.64794921875;
		points[5] = (float) 1924.1240234375;
		points[6] = (float) 3096.233642578125;
		points[7] = (float) 2911.776611328125;
		points[8] = (float) 776.0308227539062;
		points[9] = (float) 658.64501953125;
		points[10] = (float) 3619.53564453125;
		points[11] = (float) 163.5050811767578;

		DelaunayTriangulator delaunay = new DelaunayTriangulator();
		ShortArray trigs = delaunay.computeTriangles(points, false);

		for (int c = 0, n = trigs.size; c < n; c += 3) {
			System.out.format("(%d, %d, %d)\n", trigs.get(c), trigs.get(c+1), trigs.get(c+2));
		}

		// Prints out:
		// (4, 2, 5)
		// (2, 3, 1) // wrong!
		// (2, 4, 0)
		// (3, 5, 1)
		// (2, 0, 3) // wrong!
		// (5, 2, 1)
	}

}
```

#### Version of LibGDX and/or relevant dependencies
master (as of 12/15/2016, commit 52bbab3)

[...]

PR #6177, opened 2020-09-09
Two-line fix for #4489 (DelaunayTriangulator bug)

tommyettinger:
I'm not fully confident that this solves every case where the uncommon bug in #4489 occurs, so I'm requesting review from anyone proficient with sort algorithms or who can maybe optimize some of the geometry code (this is unrelated, but while someone's looking through this class they might see some opportunities). The bug in #4489 appears to boil down to a sometimes-incorrect sort of the x-positions in an array called `points`. The repro case there has the last x-position at index 10, but it isn't the highest x-position in `points`. That seems to somehow swerve into affecting some other parts of the calculation, like the assumption in `circumCircle()` that `COMPLETE` must only be returned if the point is fully to the right of the circumcircle, and points are processed in what should be sorted order (either left to right or right to left, but not mixing the highest value in the middle like it currently does). The check in this PR seems to correct the sort order for the given repro case, but I don't know if it is robust for other cases.


tommyettinger:
Sorry about the Git history; it doesn't show up as a changed file, so it shouldn't change anything, but I don't know how to erase part of the PR's history.

MrStahlfelge:
No problem, when we squash merge, only the actual changes are added to the history in a single commit.

NathanSweet:
Nice work! It's been so long since I worked on this, I'm afraid I'm not very useful for determining if this is the right fix in all cases. A simple quicksort implementation doesn't have the check, but this one appears to have some optimization. FWIW, I tried it out with a number of meshes in Spine and the triangulations looked right. I also applied it to ConvexHull and it appears to continue to work correctly.

PR #6181, opened 2020-09-09
update some eclipse classpath files

mgsx-dev:
@NathanSweet i had some classpath issues after fetching jars, i was able to fix them with these changes. However, i'm not sure about these changes because i don't use robovm stuff and i don't have Android device to test, so i can't validate them.


NathanSweet:
I'm afraid Android and iOS are not my forte. I try to stay away from those as much as possible because they are murder for productivity. I have only [one app](http://esotericsoftware.com/gloomhaven-helper) that is deployed on those platforms. For that on Android it is built with Gradle and Mario handles making it build on iOS (which I'm sure is wrought with suffering). I don't usually even have the Android/RoboVM Eclipse projects open, unless I'm working on core libgdx API stuff. Given that, I'm probably not the best one to review. If no one with more skin in the game steps up, it's probably fine to merge.

NathanSweet:
2 weeks and no one chimed in, so let's-a-go.

PR #6182, opened 2020-09-09
Net: New method downloadPixmapFromHttpUrl

MrStahlfelge:
Downloading an image from HTTP is easy on JVM platforms, but it is difficult - especially for beginners - to do on GWT. Ironically, GWT backend fetches all images needed for the game from an url before the game starts, so everything needed is already implemented. 

Thus, the new method downloadPixmapFromHttpUrl is on the one hand just a convinience method, on the other hand desperately requested.

Please check the DownloadTest. On GWT, it will throw an expected failure for the first URL to test error handling and will download the next image in the queue.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import com.badlogic.gdx.backends.gwt.GwtFileHandle;
+import com.badlogic.gdx.backends.gwt.preloader.AssetDownloader;
 import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.utils.BufferUtils;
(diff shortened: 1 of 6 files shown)


mgsx-dev (approved):
Nice! looks good to me.

MobiDevelop:
Hmm, I'm not sure on this one. It feels odd to call out a specific resource type with a dedicated interface method across all backends.

MrStahlfelge:
Do you have a suggestion how to do it without a dedicated interface method? 

Give the fact that GWT is the platform needing another implementation, you could use a supersource instead. In my opinion, super sources are more a hack and should be avoided if there's another way to get the task done.

MrStahlfelge:
Discussing it, we came up with the possibility of a `Pixmap.downloadFromUrl(url, downloadListener)`. Pixmap already has its emu file, so at least there's no new emulation file introduced. The interface would be indeed cleaner. Would you prefer that?

MobiDevelop:
That sounds like a good approach to me.

MrStahlfelge:
Please recheck, now done via Pixmap/Pixmap-emu

MobiDevelop:
That is much cleaner, could move the listener interface out of the class if you didn't want to have to duplicate that, but it's probably fine as is.

MobiDevelop (approved):

PR #6185, opened 2020-09-11
- Fixes Issue #5048. The function Intersector.overlapConvexPolygons now should return the right minimum translation vector values.

kdenzel:
Hello everyone, this should fix #5048

Sorry for not adding a non graphical test but the magic happens in the SAT Algorithm. If you are familiar with the algorithm the placement and the orientation of the minimum translation vector happens on the one dimensional axis.
The magic happens when we project the axis of the max and min points of the polygon and based on the scalar value (dot product) of this values i calculate the direction of the minimum translation vector. You can see this with the red line in the graphical Test.

TL;DR i added no functional Test because of math magic. It is hard to predict the assert value so a graphical test with debug output says more than 1mio asserts.

Also you see in the code and in the graphical test i also handle edge cases like when one polygon contains another.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -1025,7 +1025,7 @@ public final class Intersector {
 		return closestX + closestY < c.radius * c.radius;
 	}
 
-	/** Check whether specified counter-clockwise wound convex polygons overlap.
+	/** Check whether specified convex polygons overlap (clockwise or counter-clockwise wound doesn't matter).
 	 * @param p1 The first polygon.
 	 * @param p2 The second polygon.
 	 * @return Whether polygons overlap. */
@@ -1033,7 +1033,7 @@ public final class Intersector {
 		return overlapConvexPolygons(p1, p2, null);
 	}
 
-	/** Check whether specified counter-clockwise wound convex polygons overlap. If they do, optionally obtain a Minimum
+	/** Check whether convex polygons overlap (clockwise or counter-clockwise wound doesn't matter). If they do, optionally obtain a Minimum
 	 * Translation Vector indicating the minimum magnitude vector required to push the polygon p1 out of collision with polygon p2.
 	 * @param p1 The first polygon.
 	 * @param p2 The second polygon.
@@ -1048,7 +1048,7 @@ public final class Intersector {
 		return overlapConvexPolygons(verts1, 0, verts1.length, verts2, 0, verts2.length, mtv);
 	}
 
-	/** Check whether polygons defined by the given counter-clockwise wound vertex arrays overlap. If they do, optionally obtain a
+	/** Check whether polygons defined by the given vertex arrays overlap (clockwise or counter-clockwise wound doesn't matter). If they do, optionally obtain a
 	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by verts1 out of the
 	 * collision with the polygon defined by verts2.
 	 * @param verts1 Vertices of the first polygon.
@@ -1082,6 +1082,7 @@ public final class Intersector {
 	}
 
 	/**
+	 * Implementation of the separating axis theorem (SAT) algorithm
 	 * @param verts1        the verts1
 	 * @param offset1       offset of verts1
 	 * @param count1        count of verts1
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -1049,158 +1049,121 @@ public final class Intersector {
 	}
 
 	/** Check whether polygons defined by the given counter-clockwise wound vertex arrays overlap. If they do, optionally obtain a
-	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by verts1 out of the
-	 * collision with the polygon defined by verts2.
-	 * @param verts1 Vertices of the first polygon.
-	 * @param verts2 Vertices of the second polygon.
+	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by shapeA out of the
+	 * collision with the polygon defined by shapeB.
+	 * @param shapeA Vertices of the first polygon.
+	 * @param shapeB Vertices of the second polygon.
 	 * @param mtv A Minimum Translation Vector to fill in the case of a collision, or null (optional).
 	 * @return Whether polygons overlap. */
-	public static boolean overlapConvexPolygons (float[] verts1, int offset1, int count1, float[] verts2, int offset2, int count2,
+	public static boolean overlapConvexPolygons (float[] shapeA, int offsetA, int countA, float[] shapeB, int offsetB, int countB,
 		MinimumTranslationVector mtv) {
-		float overlap = Float.MAX_VALUE;
-		float smallestAxisX = 0;
-		float smallestAxisY = 0;
-		int numInNormalDir;
-
-		int end1 = offset1 + count1;
-		int end2 = offset2 + count2;
-
-		// Get polygon1 axes
-		for (int i = offset1; i < end1; i += 2) {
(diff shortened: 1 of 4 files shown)


tests/gdx-tests/src/com/badlogic/gdx/tests/IntersectorOverlapConvexPolygonsTest.java
```diff
@@ -0,0 +1,94 @@
...
+        shapeRenderer.setProjectionMatrix(camera.combined);
+        shapeRenderer.begin(ShapeRenderer.ShapeType.Line);
+        shapeRenderer.setColor(Color.GOLD);
+        shapeRenderer.polygon(triangle1.getTransformedVertices());
+        shapeRenderer.setColor(Color.CYAN);
+        shapeRenderer.polygon(triangle2.getTransformedVertices());
+        shapeRenderer.setColor(Color.RED);
+        shapeRenderer.line(mtv.normal.x, mtv.normal.y, mtv.normal.x * 10, mtv.normal.y * 10);
+        shapeRenderer.set(ShapeRenderer.ShapeType.Filled);
+        shapeRenderer.circle((mtv.normal.x), (mtv.normal.y), 0.5f);
```
obigu, line 91: The test is great, I don't understand what the circle represents. Wouldn't it be more useful (instead of having that circle) to display the MTV depth with a green line over the current red line. Something like:
`shapeRenderer.line(0, 0, mtv.depth * mtv.normal.x , mtv.depth * mtv.normal.y);`

I would also make the red line start at (0,0).

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -1049,158 +1049,121 @@ public static boolean overlapConvexPolygons (float[] verts1, float[] verts2, Min
...
-	 * @param verts1 Vertices of the first polygon.
-	 * @param verts2 Vertices of the second polygon.
+	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by shapeA out of the
+	 * collision with the polygon defined by shapeB.
+	 * @param shapeA Vertices of the first polygon.
+	 * @param shapeB Vertices of the second polygon.
 	 * @param mtv A Minimum Translation Vector to fill in the case of a collision, or null (optional).
 	 * @return Whether polygons overlap. */
-	public static boolean overlapConvexPolygons (float[] verts1, int offset1, int count1, float[] verts2, int offset2, int count2,
+	public static boolean overlapConvexPolygons (float[] shapeA, int offsetA, int countA, float[] shapeB, int offsetB, int countB,
```
obigu, line 1058: Some missing arguments java docs.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -1049,158 +1049,121 @@ public static boolean overlapConvexPolygons (float[] verts1, float[] verts2, Min
...
 
 	/** Check whether polygons defined by the given counter-clockwise wound vertex arrays overlap. If they do, optionally obtain a
-	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by verts1 out of the
-	 * collision with the polygon defined by verts2.
-	 * @param verts1 Vertices of the first polygon.
-	 * @param verts2 Vertices of the second polygon.
+	 * Minimum Translation Vector indicating the minimum magnitude vector required to push the polygon defined by shapeA out of the
+	 * collision with the polygon defined by shapeB.
+	 * @param shapeA Vertices of the first polygon.
+	 * @param shapeB Vertices of the second polygon.
```
obigu, line 1055: I don't particularly like renaming `verts` to `shape` but if we do, we should rename the other `overlapConvexPolygon` method(s) that receive vertices as an argument to follow the same naming.

obigu (changes requested):
It looks good to me. It could be done in a more object oriented style (by defining a `Projection` class or using `Vector2` for example) but it follows the previous implementation style and it's easy to read. I agree the interactive test is very powerful even if some discrete unit tests would also be nice (not necessary though).

gdx/src/com/badlogic/gdx/math/Intersector.java
MobiDevelop: I'd prefer we leave the original naming.

tests/gdx-tests/src/com/badlogic/gdx/tests/IntersectorOverlapConvexPolygonsTest.java
kdenzel: Hello @obigu thanks for your reply. I made the circle to better visualize the direction the line is pointing at. Without the point you would only see a line but it is harder to predict where the start of this line is. Maybe it would be better to draw the red line also in the oposite direction then the dot will become the origin of the line.

Yes you are right, i can translate the position of the line to 0,0. Be careful, the line represents the minimum projecton axis where the points min and max are projected from the shapes and also the direction. Showing the green line as depth is a good idea. At the moment i just show the axis with a length of 10. I will fix that.

gdx/src/com/badlogic/gdx/math/Intersector.java
kdenzel: Yes i will change it to the original naming scheme.

gdx/src/com/badlogic/gdx/math/Intersector.java
kdenzel: Yes, they are missing in the original file too but i will add them ^^.

tests/gdx-tests/src/com/badlogic/gdx/tests/IntersectorOverlapConvexPolygonsTest.java
obigu: Thanks. It's a bit better now but why do you substract `-mtv.normal.x` and `mtv.normal.y` to all the lines (red, grey and green) y coordinates?. The idea of the green line is to visually displays the MTV depth, now it goes negative when it's small.

kdenzel: Oh yes my bad. I forgot it is a unit vector and not a coordinate in space. Tried translating it (the line) to point 0,0 which of course doesn't make any sense because the vector is just pointing in a direction. Did not see it because i used a larger scale but now i saw it reducing the scale. Thanks for reviewing and mentioning it.

obigu (approved):
Changes look good to me, thanks

kdenzel:
Thanks for reviewing the code. What also comes into my mind is the java doc. The statement that the polygons must be in counter clockwise order is no longer true i guess. I would go as far as saying this was never the case.

Let me explain. The SAT checks every axis on each shape and it shouldn't matter in which direction you check first. Wether you check the first axis last or the last axis first. I also tested it with the visual test class and the results stay the same because the direction of the unit vector "mtv.normal" depends on the points projected on the one dimensional axis of the shapes.

Also [this](https://gist.github.com/nyorain/dc5af42c6e83f7ac6d831a2cfd5fbece) shows, that i am not the only one who thinks that the direction of the vertices doesn't matter.

obigu:
> Thanks for reviewing the code. What also comes into my mind is the java doc. The statement that the polygons must be in counter clockwise order is no longer true i guess. I would go as far as saying this was never the case.
> 
> Let me explain. The SAT checks every axis on each shape and it shouldn't matter in which direction you check first. Wether you check the first axis last or the last axis first. I also tested it with the visual test class and the results stay the same because the direction of the unit vector "mtv.normal" depends on the points projected on the one dimensional axis of the shapes.
> 
> Also [this Link](https://gist.github.com/nyorain/dc5af42c6e83f7ac6d831a2cfd5fbece) shows, that i am not the only one who thinks that the direction of the vertices doesn't matter.

You're right, the polygons don't need to be defined counter-clockwise. Do you mind updating the docs to remove that statement from all `overlapConvexPolygons` methods?

PR #6187, opened 2020-09-14
Fix sound and music not playing because of changed file names

SimonIT:
With #4314 the file names contain the hash. This breaks the web audio implementation because it loaded the files with the original file names


MrStahlfelge (approved):
Good fix, thanks

PR #6188, opened 2020-09-14
GwtFileHandle lists all asset types and directories instead of only texts

MrStahlfelge:
This is a followup of #3721 - original PR author disappeared, but the change is good. So I merged against current master and fixed a small bug.


tommyettinger:
We don't need a CLA signed for this, do we? It's a non-original bugfix, that's OK right?

MrStahlfelge:
I'm sure his signed CLA is in Mario's drawer, right next to mine.

SimonIT (approved):
Works like a charm
For testing: Run FilesTest on gwt, look in the browser console if more than 23 files are listed

PR #6194, opened 2020-09-24
Gdx Setup minor Gradle fixes

obigu:
A couple of minor improvements to the Gradle scripts generated by Gdx Setup.

1. On `core` and `desktop` modules, `build.gradle` generated scripts contained the line `apply plugin: "%LANG%"` which is unnecessary as this plugin is already defined per module on root `build.gradle`. Also, in the case of Java it was adding `apply plugin: java` instead of `apply plugin: java-library`.

2. Replaced deprecated `apply plugin: "android"` by `apply plugin: "com.android.application"`.


mustii82:
Can we use latest the AndroidPlugin Version with this fix?

PR #6200, opened 2020-10-01
GWT: Use Gretty for serving files in superdev mode

MrStahlfelge:
Since Jetty plugin was removed, we use Gradle's SimpleFileHttpServer to serve the assets in superdev mode.

This has unfortunately some drawbacks:
* Files without extension aren't served, causing older projects (including our test project) to fail in superdev mode
* The server is not killed when super dev process is stopped, so task manager has to be your regular friend while developing
* SimpileFileServer is very simple (hence the name) and does not support caching or default file mechanisms

This PR changes our tests and the template project to use Gretty (Gradle Jetty Plugin), as we used Jetty before. The problems listed above are fixed with this.

I know from the history @Tom-Ski tried to integrate Gretty before, but eventually removed it. I suppose it was because Gretty fails to run out of the box in our environment. This is because gwt-dev pulls in at least some parts of Tomcat causing conflicts. This can be avoided by adding an exclusion on the gwt backend project, which I did.

### DIFF
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
@@ -15,30 +15,39 @@ gwt {
 }
 
 import org.wisepersist.gradle.plugins.gwt.GwtSuperDev
+import org.akhikhl.gretty.AppBeforeIntegrationTestTask
 
-def HttpFileServer server = null
-def httpFilePort = 8080
+gretty.httpPort = 8080
+gretty.resourceBase = project.buildDir.path + "/gwt/draftOut"
+gretty.contextPath = "/"
+gretty.portPropertiesFileName = "TEMP_PORTS.properties"
 
 task startHttpServer () {
     dependsOn draftCompileGwt
 
-    String output = project.buildDir.path + "/gwt/draftOut"
-
-    doLast {
+    doFirst {
         copy {
             from "webapp"
-            into output
+            into gretty.resourceBase
         }
 
         copy {
             from "war"
-            into output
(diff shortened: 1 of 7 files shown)


MobiDevelop (approved):
Seems fine to me. Not a fan of 'somewhat of a hack' but I am also not a fan of having to kill the gradle daemon every time I have to run superDev.

SimonIT (approved):
Fixes the annoying bug

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
```diff
@@ -15,30 +15,39 @@ gwt {
...
         }
+    }
+}
 
-        server = new SimpleHttpFileServerFactory().start(new File(output), httpFilePort)
+task beforeRun(type: AppBeforeIntegrationTestTask, dependsOn: startHttpServer) {
+    // The next line allows ports to be reused instead of
+    // needing a process to be manually terminated.
+    file("build/TEMP_PORTS.properties").delete()
+    // Somewhat of a hack; uses Gretty's support for wrapping a task in
```
tommyettinger, line 45: I should mention that the "hack code" was kinda the last resort I could come up with, after trying various ways around the very-brittle SimpleHttpFileServer code. The `integrationTestTask` isn't used as intended, but it does allow running two servers at once, a necessity for superDev because it has two different ports in use doing different things. Most of this change happened in May in gdx-liftoff, and as far as I can tell no one has had issues with superDev, including more than a few people who used liftoff in the latest jam.

tommyettinger (approved):
Looks good. I don't see anything that looks incorrect.

PR #6205, opened 2020-10-07
Improvements on Vector classes (readability)

MarcinSc:
Extracted two common methods to Vector interface.
Changed implementations to use common code for methods with different signatures.


mgsx-dev:
i think it was like that to avoid some method calls (micro optimization), i don't think it's a good idea to revert that.

PR #6206, opened 2020-10-08
Added Launchscreen storyboard and icon to gdx-ios tests

obigu:
`gdx-tests-ios` tests were still using old static launch screen images and icon. Now it uses the same as default app generated by Setup app that are StoryBoard based. 

I've also moved these resources from `gdx-tests-android` to the `gdx-tests-ios` module because they are not shared resources.


tests/gdx-tests-iosrobovm/robovm.xml
```diff
@@ -5,6 +5,7 @@
   <arch>thumbv7</arch>
   <target>ios</target>
   <iosInfoPList>Info.plist.xml</iosInfoPList>
+  <treeShaker>conservative</treeShaker>
```
MrStahlfelge, line 8: Are you sure this is needed? As far as I know, the default setting is `conservative`.

MrStahlfelge (approved):
Works good and was a thing that bothered me for some time, thanks for the change.

tests/gdx-tests-iosrobovm/robovm.xml
obigu: It's not "needed" here because IPA size is not that relevant for tests but I think it's good to have the same one that we have on `robovm.xml` gdx-setup. `conservatiove`setting doesn't harm and is totally save because it only prunes especifically anootated classes. Tree shaker on latest Mobivm is currently `none` by default iirc.

PR #6210, opened 2020-10-08
Extracted two common methods to Vector interface.

MarcinSc:


mgsx-dev (approved):
LGTM it's a breaking change but i don't think a lot of user code implements Vector interface, so i think it's a minor break that could be merged for upcoming release.

obigu:
I don't think this methods should be part of the `Vector` interface. These 2 methods (to add or substract the same scalar to each component) are not a "vector operation". They are also not currently implemented on `Vector2` so we are not really extracting it either. Imo it should not even be in `Vector3`.

mgsx-dev:
@obigu i think it's fine, it's helpful when dealing with some vectors converting from 0;1 to -1;1, range. For instance : normal vectors, procedural generation, etc... i can understand it's not mathematically common but it doesn't hurt to have that imo.

obigu:
@mgsx-dev we're talking about an interface here that defines for vector as an abstract entity, these 2 methods are, at most, convenience methods that could be placed in an implementation even if I doubt they are ever used (we already have `add(float x, float y)` for the cases you mention). 

It may not seem like it but, imo, we hurt the API when we pollute it adding methods that don't belong there, we increase complexity for no gain and we break current implementations (even if rare it's also rare to use these methods). I have to vote against merging this.

tommyettinger:
@obigu Yeah, nobody needs the two-argument `-` operator anyway, they should take it out of math. `5 + -1 = 4`, easy right? We could go further and define all of our math in bit operations! Closer to the metal!

Anyway, I vote to approve.

obigu:
I've checked other game engines and several math libraries that have Vector (2 or 3) APIs to see if I was being missing something and, except from one from a minor java library that looked suspiciously similar to libGDX `Vector3`, I haven't found any implements such methods.

mgsx-dev:
There are some algorithms that operate on abstract vectors (eg. all libgdx Path implementations). If some need to use add(float) or sub(float) in such generalized algorithms, they are stuck and have to workaround it by using matrices.

The sub method makes API consistent, we already have sub(T) and sub(float, float, float), so it makes sense to have it.

obigu:
> There are some algorithms that operate on abstract vectors (eg. all libgdx Path implementations). If some need to use add(float) or sub(float) in such generalized algorithms, they are stuck with dirty workarounds.
> 
Are they currently needed on those places? are we wing forced to do those workarounds now?

> The sub method makes API consistent, we already have sub(T) and sub(float, float, float), so it makes sense to have it.
Substracting vectors has a geometric meaning, it's a basic operation with vectors and should be there. It's not the same case with substracting the same scalar from all the components.

MarcinSc:
@obigu Also, in GLSL you can do vec3+float, and it makes sense there. All I wanted, is to those two classes Vector2 and Vector3 bring a bit closer to their GLSL equivalents.

MrStahlfelge:
What's the state of this issue? I see an approvement and some discussion, but don't understand enough to have a conclusion. Can we merge this to have it in the next version?

crykn:
@MrStahlfelge: The question is, whether Vector should only contain methods that have a geometrical meaning or also methods that may prove useful in certain scenarios.

@mgsx-dev and @tommyettinger (and obviously @MarcinSc) think that we should add `sub(float)` and `add(float)`, as there may be some uses for doing those operations and this is also possible, for example, in GLSL. @obigu, on the other hand, is against adding those methods, because there is no geometrical meaning of those operations and the PR thus just pollutes the API. 

I myself don't really have an opinion on this: I think both sides have good arguments. However, without a good use-case I would tend to agree with @obigu.

MrStahlfelge:
Thanks for the good summary.

I also don't have an opinion, and as no one else chimed in, this seems this doesn't matter for the majority here. Now we have the problem that all opinianators are appreciated members of the club, and 2:1 is a very tight majority.

To counterbalance your vote, I would tend to agree with @tommyettinger and @mgsx-dev because, if my memory is not completely wrong, scalar operations were part of vector analysis course for me.

crykn:
Hey @MarcinSc, thank you very much for your contribution! Your PR certainly sparked some discussion 😄

Since there currently is no clear majority in favour of merging, the team thought it was best to close this for now. But we certainly hope that we didn't discourage any future contributions made by you!

PR #6211, opened 2020-10-09
Extracted the setup of ScrollPane's Listeners

payne911:
This way, the API is more flexible: it will not require people to potentially copy-paste the whole class just to tweak its Listeners.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
@@ -103,9 +103,10 @@ public class ScrollPane extends WidgetGroup {
 	}
 
 	/** Adds all the default listeners. */
-	protected void addListeners() {
+	private void addListeners() {
 		addCaptureListener();
-		addFlickScrollListener();
+		this.flickScrollListener = getFlickScrollListener();
+		addListener(this.flickScrollListener);
 		addScrollListener();
 	}
 
@@ -186,8 +187,8 @@ public class ScrollPane extends WidgetGroup {
 		});
 	}
 
-	protected void addFlickScrollListener() {
-		flickScrollListener = new ActorGestureListener() {
+	protected ActorGestureListener getFlickScrollListener() {
+		return new ActorGestureListener() {
 			public void pan (InputEvent event, float x, float y, float deltaX, float deltaY) {
 				setScrollbarsVisible(true);
 				amountX -= deltaX;
@@ -218,7 +219,6 @@ public class ScrollPane extends WidgetGroup {
 				return false;
 			}
 		};
-		addListener(flickScrollListener);
 	}
 
 	protected void addScrollListener() {
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
@@ -99,11 +99,6 @@ public class ScrollPane extends WidgetGroup {
 		setActor(widget);
 		setSize(150, 150);
 
-		addListeners();
-	}
-
-	/** Adds all the default listeners. */
-	private void addListeners() {
 		addCaptureListener();
 		this.flickScrollListener = getFlickScrollListener();
 		addListener(this.flickScrollListener);
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
@@ -99,6 +99,17 @@ public class ScrollPane extends WidgetGroup {
 		setActor(widget);
 		setSize(150, 150);
 
+		addListeners();
+	}
+
+	/** Adds all the default listeners. */
+	protected void addListeners() {
+		addCaptureListener();
+		addFlickScrollListener();
+		addScrollListener();
+	}
+
+	protected void addCaptureListener() {
 		addCaptureListener(new InputListener() {
 			private float handlePosition;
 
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
```diff
@@ -173,7 +184,9 @@ public boolean mouseMoved (InputEvent event, float x, float y) {
 				return false;
 			}
 		});
+	}
 
+	protected void addFlickScrollListener() {
 		flickScrollListener = new ActorGestureListener() {
```
MrStahlfelge, line 190: Overriding this with a no-implementation is not possible, will give NPE.

So this should be a method returning a listener instead, annotated with `NonNull`. Ah wait, we don't have a `NonNull` annotation and don't need it because in libgdx returning null is not allowed anyway.

MrStahlfelge (changes requested):
Generally, I embrace every way to adjust scene2d behaviour. 
And there is a need for this, I have this code in my project:
```
    /**
     * HACK
     */
    private void removeMouseScrollListener() {
        for (EventListener e : getListeners()) {
            if (e instanceof InputListener) {
                removeListener(e);
                break;
            }
        }
    }
```
However, this should be changed, see my comment.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
payne911: Oh, good catch! I did the editing in the GitHub text editor a bit too hastily, apparently.

But now you make me wonder about my approach to allow `addListeners()` to be overridable: if `addFlickScrollListener()` returned a Listener to try to reduce the possibility that a user ends up with an NPE, what about the possibility that the user inadvertently overrides `addListeners()` and forgets to assign `flickScrollListener ` ?

Tell me what you think: I'd do what you suggested, and also change `addListeners()`'s visibility to `private`.

payne911: Oh, good catch! I did the editing in the GitHub text editor a bit too hastily, apparently.

But now you make me wonder about my approach to allow `addListeners()` to be overridable: if `addFlickScrollListener()` returned a Listener to try to reduce the possibility that a user ends up with an NPE, what about the possibility that the user inadvertently overrides `addListeners()` and forgets to assign `flickScrollListener ` ?

Tell me what you think: I'd do what you suggested, and also change `addListeners()`'s visibility to `private`.

payne911:
I also had to "hack" the `ScrollPane` in my [other project](https://github.com/payne911/gdx-setup-generator/blob/master/src/test/java/com/github/payne/gui/components/ScrollPayne.java):

```java
        addCaptureListener(new InputListener() {
            @Override
            public boolean touchDown(InputEvent event, float x, float y, int pointer, int button) {
                if (pointer == 0 && button != 0) {
                    event.cancel(); // ensure only a left-click will pan
                    return false;
                }
                return super.touchDown(event, x, y, pointer, button);
            }
        });
```

I was thinking I could add that functionality to be supported out-of-the-box by the ScrollPane, but thought the PR might be refused because that's not really a required feature. However, my idea would be backward-compatible, and allow extra customization by allowing the user to change which Mouse Button can be used to interact with the Listeners.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ScrollPane.java
```diff
@@ -99,6 +99,18 @@ public ScrollPane (@Null Actor widget, ScrollPaneStyle style) {
...
 
+		addListeners();
+	}
+
+	/** Adds all the default listeners. */
+	private void addListeners() {
+		addCaptureListener();
+		this.flickScrollListener = getFlickScrollListener();
+		addListener(this.flickScrollListener);
+		addScrollListener();
```
Darkyenus, line 110: What is the point of this method when it is private and called only from one place? Wouldn't it be simpler to just inline it?

MobiDevelop: I agree, there is no benefit to extracting this to a method.

payne911: Oh, you are right: that is an artefact of my initial idea of having this method as `protected`.

Will fix!

crykn:
I think it looks good now. What do you think @MrStahlfelge, @NathanSweet?

MrStahlfelge (approved):
It looks good for me

obigu:
This kind of requirement has popped out before and it's complicated because these kind of changes add limited extensibility as most fields are either private or package.

Independently of that I would make some changes if this were to be merged. I think the 3 listeners should be returned by a `protected getXXXXListener()` (like FlickScroll one) but I would initialize them lazily so that calling `getXXXXXListener()` returns the actual reference or creates it if null.

Generally speaking calling protected methods from the constructor is risky but I don't see much room to change that without changing the whole thing.

Probably @NathanSweet should give his opinion on this one.

MrStahlfelge:
Merged as I think this makes it a bit more flexible, does not break anything and every officially marked change request was fulfilled. Everyone with further improvement suggestions feel free to open another PR. :)

NathanSweet:
Sorry I'm a bit late.

The protected methods to customize listeners look OK to me. Could use javadocs to warn they are called from the constructor, but then it's usually pretty obvious if that is problematic for a subclass.

It may be slightly better to use eg `getCaptureListener` (like `getFlickScrollListener`) rather than `addCaptureListener`, as then it's easier to wrap the default listener. It can still be done by using `getCaptureListeners().peek()` but that's a little more awkward.

PR #6215, opened 2020-10-11
ShaderProvider extends Disposable

JackRainy:
Since `ShaderProvider` interface does already contain declaration of the `public void dispose()`, it makes sense to allow working with these instances same as with `Disposable` ones.

Also, `import` declarations, which are never used, have been removed here.


mgsx-dev (approved):
LGTM

PR #6217, opened 2020-10-11
Add method to calculate intersection point between three planes

### LINKED ISSUE #6214: [Enhancement] Intersection between three planes
This one can be assigned to me. It would be extremely helpful to allow visualization of some methods in Intersector (since it can be used to render frustum as a mesh)

Example algorithm from [MonoGame](https://github.com/MonoGame/MonoGame/blob/master/MonoGame.Framework/BoundingFrustum.cs#L529) (MIT License)

```
Vector3 IntersectionPoint(ref Plane a, ref Plane b, ref Plane c)
{
    // Formula used
    //                d1 ( N2 * N3 ) + d2 ( N3 * N1 ) + d3 ( N1 * N2 )
    //P =   ---------------------------------------------------------------------
    //                             N1 . ( N2 * N3 )
    //
    // Note: N refers to the normal, d refers to the displacement. '.' means dot product. '*' means cross product

    Vector3 v1, v2, v3;
    float f = -Vector3.Dot(a.Normal, Vector3.Cross(b.Normal, c.Normal));

    v1 = (a.D * (Vector3.Cross(b.Normal, c.Normal)));
    v2 = (b.D * (Vector3.Cross(c.Normal, a.Normal)));
    v3 = (c.D * (Vector3.Cross(a.Normal, b.Normal)));

    Vector3 vec = new Vector3(v1.X + v2.X + v3.X, v1.Y + v2.Y + v3.Y, v1.Z + v2.Z + v3.Z);
    return vec / f;
}
```

Reference: https://stackoverflow.com/a/28995460

yuripourre:
This PR simply adds a method to calculate the intersection point between three planes.

Returns false if there is no intersection or the intersection is not a point.
Returns true if there is an intersection point.

This post might help to illustrate the math behind it: http://geomalgorithms.com/a05-_intersect-1.html

I need this code (among other reasons) to prove with images that the PR #6086 is not breaking anything and is doing the same calculation as before.

**Tags**
- Non breaking changes
- Math

Closes #6214

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,7 @@
 - API Change: Table#round uses ceil/floor and is applied during layout, rather than afterward.
 - Fixed blurry NinePatch rendering when using a single center region.
 - API Change: Upon changing the window size with the lwjgl3 backend, the window is centered on the monitor.
+- Added Intersector#intersectPlanes to calculate the point intersected by three planes
 
 [1.9.11]
 - Update to MobiVM 2.3.8
--- a/CHANGES
+++ b/CHANGES
@@ -47,7 +47,7 @@
 - API Change: Table#round uses ceil/floor and is applied during layout, rather than afterward.
 - Fixed blurry NinePatch rendering when using a single center region.
 - API Change: Upon changing the window size with the lwjgl3 backend, the window is centered on the monitor.
-- Added Intersector#intersectPlanes to calculate the point intersected by three planes
+- Added Intersector#intersectPlanes to calculate the point intersected by three planes, see https://github.com/libgdx/libgdx/pull/6217
 
 [1.9.11]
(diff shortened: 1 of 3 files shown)


yuripourre:
@NathanSweet @mgsx-dev
Hey, if you have some time, can you please review?

mgsx-dev:
@yuripourre no need to mention MonoGame in the code since it's now based on the article you linked here (no need to put article link in the code neither ;-) )
Except that, it looks good to me.

yuripourre:
@mgsx-dev thanks for reviewing it, I just removed references to MonoGame

mgsx-dev (approved):
Nice! it looks good to me.

yuripourre:
@NathanSweet thanks!!

PR #6218, opened 2020-10-11
Added Input:setCursorHidden(). Resolves #5546

### LINKED ISSUE #5546: Allow hiding of the mouse cursor.
#### Issue details
The LWJGL3 backend does not allow for fully transparent images to be set as the mouse cursor. We would like to be able to make the cursor invisible without capturing it within the window. The solution is to use the following LWJGL3 method:
`glfwSetInputMode(((Lwjgl3Graphics) Gdx.graphics).getWindow().getWindowHandle(), GLFW_CURSOR, GLFW_CURSOR_HIDDEN);` It would be nice to have a method to do this in Input such as the following: `Gdx.input.setCursorHidden(true)`

#### Reproduction steps/code
- Create an LWJGL3 project
-cursor.png is a fully transparent 32x32 image. Add the following line to the core project:
`Gdx.graphics.setCursor(Gdx.graphics.newCursor(new Pixmap(Gdx.files.internal("cursor.png")), 0, 0));`
-Run the game. Notice that the cursor is now an opaque black rectangle.

1.9.10-SNAPSHOT

- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

raeleus:
Please see #5546. Catching the cursor and hiding the cursor are two separate goals. Catching cursor makes the cursor invisible _and_ keeps the cursor locked to the window. Hiding the cursor will only make the cursor invisible and not restrict the positioning of the mouse. There are use cases for this such as rendering animated mouse cursors in game.

The previous technique to hide the mouse (and the technique that is still valid for LWJGL2) is to set the cursor to a completely transparent image. This does not work on LWJGL3 per the mentioned issue (verified on Windows 10). The modern technique is to use `GLFW.GLFW_CURSOR_HIDDEN` which has no such issues.

I have implemented this feature with `Gdx.input.setCursorHidden(true)`. I am somewhat at odds with myself in regards to adding to the Input interface. This is really only applicable to LWJGL3, but the option would be visible on all platforms. However, there is precedent with this considering `setCursorCatched()`. I do not want to add it as a configuration option because users may want to Hide/Unhide during runtime. Adding the method to Input will allow others in the future to emulate the behavior on other platforms if they see fit.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -940,6 +940,16 @@ public class DefaultAndroidInput implements AndroidInput {
 	public boolean isCursorCatched () {
 		return false;
 	}
+    
+    @Override
+    public void setCursorHidden (boolean hidden) {
+    
+    }
+    
+    @Override
+    public boolean isCursorHidden () {
+        return false;
+    }
 
 	@Override
 	public int getDeltaX () {
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -940,16 +940,16 @@ public class DefaultAndroidInput implements AndroidInput {
 	public boolean isCursorCatched () {
 		return false;
 	}
-    
-    @Override
-    public void setCursorHidden (boolean hidden) {
-    
-    }
-    
-    @Override
-    public boolean isCursorHidden () {
-        return false;
-    }
+
+	@Override
+	public void setCursorHidden (boolean hidden) {
+
+	}
+
+	@Override
+	public boolean isCursorHidden () {
+		return false;
+	}
 
 	@Override
 	public int getDeltaX () {
--- a/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/input/MockInput.java
+++ b/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/input/MockInput.java
@@ -264,6 +264,16 @@ public class MockInput implements Input {
 	public boolean isCursorCatched() {
 		return false;
 	}
+    
+    @Override
+    public void setCursorHidden (boolean hidden) {
+    
+    }
+    
+    @Override
+    public boolean isCursorHidden () {
+        return false;
+    }
 
 	@Override
 	public void setCursorPosition(int x, int y) {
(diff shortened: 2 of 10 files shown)


crykn:
First things first: I just tested it on MacOS and `setCursorHidden()` works there as well 👍 . However, I'm not sure, whether `Input` is the right place for this method. I see the similarity to `Input#setCursorCatched()`, but imo it's a method dealing with graphics and should be in `Graphics` (next to methods like `setSystemCursor()`). 

Also I'm sceptical, whether `setCursorHidden` calls should really be handed over to `setCursorCatched` on GWT, etc. If we're adding `setCursorHidden` as an alternative option next to cursor catching, there should be a difference between both that should be equal on all platforms.

raeleus:
I was a little confused there on what cursorCatched actually does on GWT since it didn't seem to work on any browsers I have. I'd be happy enough to make cursorHidden do nothing on that platform.

I agree on that graphics part. I wish cursorCatched had nothing to do with the visibility of the mouse, but  that's what we have to work with. Setting cursorHidden unsets cursorCatched, so I ultimately feel the option should not go into graphics.

MrStahlfelge:
> I was a little confused there on what cursorCatched actually does on GWT since it didn't seem to work on any browsers I have.

If it does not, it is broken and should get fixed. Pointer lock works on HTML5 ([information with demo](https://www.html5rocks.com/en/tutorials/pointerlock/intro/)). 

>  Catching the cursor and hiding the cursor are two separate goals. 

Is this really true? What's the point in catching the cursor and still show it? HTML5 hides the cursor when it is catched, as well as iOS does AFAIK.

raeleus:
> If it does not, it is broken and should get fixed. Pointer lock works on HTML5 (information with demo).

I'll look at that sometime. The examples on that link work fine, so I guess I'll look at fixing that in a separate issue.

> Is this really true? What's the point in catching the cursor and still show it? HTML5 hides the cursor when it is catched, as well as iOS does AFAIK.

Yes. I gave a use case in my issue. Basically you want to have the option between the locking the cursor or not. This is legitimized by the fact that LWJGL3 offers separate options for these.

MrStahlfelge:
Yes, I can see a use case for hiding the cursor, but not catch it. However, that's not what the implementations in this PR do:
```
	public void setCursorHidden (boolean hidden) {
	setCursorCatched(hidden);
	}
```

raeleus:
Yeah I put that in there to mimic the same behavior across platforms even though it may cause locking. I think I should just redo this and make a setting only applicable to LWJGL3

Frosty-J:
For GWT we could toggle the `cursor: none` style on the game instead of catching the cursor.

Unsure why it catches on AWT and not LWJGL2, but they have the transparent pixmap technique (could be added into this PR - a pixel isn't going to use much memory even if people don't use it) and are in my opinion a lower priority anyway.

raeleus:
This has been resolved with the better solution of #6822

PR #6219, opened 2020-10-11
Fixed TextField tab traversal on HTML5.

raeleus:
Tab traversal along TextFields in HTML5 does not work for a number of reasons. The first obvious reason is that the browser reserves the key for navigation. That is easily resolved by adding the following to the script block of the index.html:
```
window.onkeydown = function(event) {
  if(event.keyCode == 9) {
    event.preventDefault();
    return false;
  }
};
```

It's the responsibility of the user to handle this code and is not the point of this PR. This operation is covered in the [wiki](https://github.com/libgdx/libgdx/wiki/HTML5-Backend-and-GWT-Specifics#preventing-keys-from-triggering-scrolling-and-other-browser-functions). However, by doing the above you break the ability of TextField to detect Tab keys. This is because the detection is in the keyTyped method. If we move the detection to keyDown, it works across all the backends.

I would appreciate some testing on iOS/Android if anyone is available.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -886,7 +886,10 @@ public class TextField extends Widget implements Disableable {
 
 			if (!hasKeyboardFocus()) return false;
 
-			if (checkFocusTraversal(keycode)) next(UIUtils.shift());
+			if (checkFocusTraversal(keycode)) {
+				next(UIUtils.shift());
+				return true;
+			}
 			
 			boolean repeat = false;
 			boolean ctrl = UIUtils.ctrl();
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -885,8 +885,8 @@ public class TextField extends Widget implements Disableable {
 			if (focused) Timer.schedule(blinkTask, blinkTime, blinkTime);
 
 			if (!hasKeyboardFocus()) return false;
-            
-            if (checkFocusTraversal(keycode)) next(UIUtils.shift());
+
+			if (checkFocusTraversal(keycode)) next(UIUtils.shift());
 			
 			boolean repeat = false;
 			boolean ctrl = UIUtils.ctrl();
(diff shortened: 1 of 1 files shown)


MrStahlfelge (approved):
The only problem I found reviewing this is that in case Ctrl key is pressed, handled will set to false although an operation is done.

This will work on iOS and Android. Both systems send the keydown event for enter.

In my opinion, keyDown event is the better place to trigger the behaviour. This is what people are used to, because on Windows and Linux native text fields, it is triggered on keyDown event, too.

raeleus:
Okay Mr. Stahlfelge, I made the adjustment. Do you have any comments on this PR, @NathanSweet?

NathanSweet:
The difference is pretty subtle, but could break existing code that is expecting traversal to occur in keyTyped, not keyDown. In Spine I have a bunch of listeners for keyTyped and some are related to focus traversal. It's not clear if this PR would require app code changes. I can try it tomorrow and report back.

> on Windows and Linux native text fields, it is triggered on keyDown event, too.

This doesn't appear to be true. If I hold tab on Windows, the focus changes repeatedly, indicating it uses the equivalent of keyTyped and matches TextField's current behavior. If it used keyDown then when the key goes down the focus would change and then not change again while the key is down, until another keyDown occurred, which is what we'll get from this PR.

raeleus:
An interesting point about holding the key down. If you don't think the code looks too ugly to handle the behavior based on platform, I'll makes the changes tomorrow. Thanks.

MrStahlfelge:
The question is if this feature is really needed. I don't see a real use case for holding tab down and let the cursor cycle - that's why I never noticed that this is the behaviour on Windows.

raeleus:
Yeah, I wouldn't imagine anyone actually using it like that. I am concerned about breaking changes though. I dislike messing with the base widgets unless absolutely necessary, but this thing is important especially since we're doing more web based utilities now.

MrStahlfelge:
I have taken another look on this, and instead of fixing this behaviour in Textfield's core code, I would rather change backend's `DefaultGwtInput` class to simulate a keyTyped event on tab keyUp. This way, we have same behaviour on GWT as on other backends. If you are with me, I would open another PR changing this and fixing the problem that way.

raeleus:
That is an excellent idea and thank you for following up on this. I didn't feel I could make an adequate, clean solution with this route.

MrStahlfelge:
Solved with #6296

PR #6220, opened 2020-10-11
Fix shader providers overflow and depth shader skinning configuration

mgsx-dev:
Few issues i tried to fix with the less breaking changes as possible :

* Sometimes, a shader created with a given configuration (renderable), cannot render the same configuration (canRender method). This causes an overflow of shader creation at each drawing making shaders list growing indefinitely. Now, an exception is thrown.
* Depth shader configured with maxBones=0 is one of those case : it can't render a skinned mesh even if it has been created with a skinned renderable.
* While i was fixing it, i saw that a new DepthShader is created depending on renderable bones count. DefaultShaderProvider doesn't care about bones count, it provides 2 variants : one with skinning, one without. I changed depth shader to use the same approach. Several models with different bones count means a lot of shaders and switches. 

The downside of this last change is for some case like : 1 model with 100 bones and a lot of models with only 1 or 2 bones. One shader switch might be better than sending 100 uniform matrices per renderable. I didn't benchmarked it to see how bad it could be. Note that devs could customize this by subclassing DepthShaderProvider.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,7 @@
 - API Change: Table#round uses ceil/floor and is applied during layout, rather than afterward.
 - Fixed blurry NinePatch rendering when using a single center region.
 - API Change: Upon changing the window size with the lwjgl3 backend, the window is centered on the monitor.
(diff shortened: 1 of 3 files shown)


mgsx-dev:
i force pushed to trigger a new build

xoppa:
Thanks!

PR #6222, opened 2020-10-13
Added support ios split screen. Fixes #5814

### LINKED ISSUE #5814: iOS Splitscreen support?
#### Issue details
When using split screen mode on iOS the app does not get called with the new dimensions of the screen.

#### Reproduction steps/code
Create a new libgdx app. Add `LaunchScreen.storyboard` and update the `Info.plist.xml` with the following
```xml
<key>UILaunchStoryboardName</key>
<string>LaunchScreen</string>

<!-- 
<key>UIRequiresFullScreen</key>
<true/>
 -->
```

- Libgdx 1.9.11
- Robovm 2.3.7

#### Logging
I've added logging to the `public void resize(int width, int height)` callback
```
## App start
Resize 2388 1620
Planned content area: 0, 0, 2388, 1580
## Enter splitscreen with 50:50 split
Resize 2388 1620
Planned content area: 0, 0, 2388, 1580
```

I would expect that libgdx reports the smaller screen size after the screen has been split.
Any ideas about this issue or is this mode not even supported?

- [X ] iOS (robovm)

obigu:
Fixes #5814 

To support iOS Split screen, the only required change is to calculate the graphics bounds using the iOS window object instead of the screen. Any app that doesn't want to support split screen just has to add the `UIRequiresFullScreen` flag to their `Info.plist`.

In addition to changing that, there's several things related to it that have been improved.
1. Window is now created and set before libGDX is setup and creates Graphics. This way, when Graphics are created and `computeBounds()` called, `window` is not null.
2. The `window` property on `AppDelegate` is now explicitely set with the main window we set key by calling `uiApp.getDelegate().setWindow(uiWindow)`. In libGDX we keep a reference to this window on `IOSApplication` (called `uiWindow`) but on standard iOS apps it is kept on the `AppDelegate` (in our case `IOSLauncher`) that can access it by calling `getWindow()`. Before this change, calling `getWindow()` from `IOSLauncher` returned `null`. Some more info on the differences between app window and app delegate windows here: https://stackoverflow.com/questions/21698482/what-is-the-difference-between-uiapplication-sharedapplication-delegate-window-a.
3. There was some code that was added for backwards compatibility for iOS 7- before windows bound sizes and status bar were made device orientation dependent (https://stackoverflow.com/questions/24150359/is-uiscreen-mainscreen-bounds-size-becoming-orientation-dependent-in-ios8). Now that the minimum supported iOS version is 8 we can get rid of them.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -119,7 +119,7 @@ public class IOSApplication implements Application {
 		this.uiApp = uiApp;
 
 		// enable or disable screen dimming
-		UIApplication.getSharedApplication().setIdleTimerDisabled(config.preventScreenDimming);
+		uiApp.setIdleTimerDisabled(config.preventScreenDimming);
 
 		Gdx.app.debug("IOSApplication", "iOS version: " + UIDevice.getCurrentDevice().getSystemVersion());
 		Gdx.app.debug("IOSApplication", "Running in " + (Bro.IS_64BIT ? "64-bit" : "32-bit") + " mode");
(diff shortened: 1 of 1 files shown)


MrStahlfelge:
Code looks good for me so far, I will test it for normal apps soon.
Only problematic thing that came to my mind is if you checked if the `Gdx.graphics.getSafeInset...` methods are still correct? (This might be a theoretical question, I am not aware of an iPad with insets.)

obigu:
> Code looks good for me so far, I will test it for normal apps soon.
> Only problematic thing that came to my mind is if you checked if the `Gdx.graphics.getSafeInset...` methods are still correct? (This might be a theoretical question, I am not aware of an iPad with insets.)

It would depend on what the iOS returns as insets values when calling the method in a windowed view. I don't think there's anything we could do to handle this on our side even if iPhones started to support split screen.

MrStahlfelge (approved):
Made a short test and this worked (for normal apps without splits).

MrStahlfelge:
Tested sucessfully by @davidgiga1993 - see https://github.com/libgdx/libgdx/issues/5814#issuecomment-709404698. We are waiting some time for his second test before merging.

PR #6225, opened 2020-10-14
Return iOS Insets in logical pixels

obigu:
Insets API was still returning physical pixels after the changes on https://github.com/libgdx/libgdx/pull/3709. Made the Insets API consistent by returning logical pixels.


MrStahlfelge (approved):
Good change. I like how clean this now is.

PR #6229, opened 2020-10-15
Add Octree

yuripourre:
Binary Space Partition (BSP) allows developers to work with just a subset of geometries at a given moment, it can be used to improve rendering and/or check collisions. This is a must-have feature for many 3d games.

I really appreciate all feedbacks here!

@mgsx-dev this is the implementation I mentioned. I removed the triangle part so we can start discussing the structure.

Steps
- [X] Simple implementation
- [X] Gdx test
- [X] Generic implementation
- [X] BoundingBox octree
- [x] Add collision methods for frustum and ray to Collider
- [X] Query methods (AABB, Frustum)
- [x] Query handlers (optional param) for query methods
- [x] RayCast implementation (with Handler)
- [x] Add documentation

References:
https://github.com/mgsx-dev/dl13/blob/master/core/src/net/mgsx/dl13/navmesh/NavMesh.java#L105


mgsx-dev (changes requested):
I made a first quick review, don't take it badly, i didn't take gloves for the sake of review efficiency :D 

I'd like to see a first implementation of a query(Frsutum) method to see how it looks, especially userObject/geometry pairing.

to me the nominal case of geometry is BoundaryBox. Vector3 is an edge case IMO.
i'm not even sure we would need more than boundary box approximation, user code could fine grain check collisions over their original shapes. Even for the rayCast scenario, they could supply a handler to perform perfect collision over their original shape. Octree still an approximation to reduce complexity.

Also, maybe we could make it generic : Octree<T>, it would be cleaner. User could always use Octree<Object> if they want to mix apples with oranges.

Also i'm a bit balance between using java collection over libgdx collections. I know java one could performs better in most cases but i think we should stay consistent with libgdx code (i know i'll get a shit storm because of what i say here, i'm ready for it :D )

Feel free to disagree with my comments, i could be wrong. I'll probably be more confident about what i said as we going further with the implementation.

Thank you for your work.

yuripourre:
@mgsx-dev Thank you so much for taking the time to review it. For the next days I will work on your comments, and build a working example.

> I made a first quick review, don't take it badly, i didn't take gloves for the sake of review efficiency :D
Don't worry at all, we don't have time for that.
 
> I'd like to see a first implementation of a query(Frsutum) method to see how it looks, especially userObject/geometry pairing.
> 
> to me the nominal case of geometry is BoundaryBox. Vector3 is an edge case IMO.
Yep, I didn't think about it before you suggested, I had plans for geometry only and do the mapping outside but makes sense to have it.

> i'm not even sure we would need more than boundary box approximation, user code could fine grain check collisions over their original shapes. Even for the rayCast scenario, they could supply a handler to perform perfect collision over their original shape. Octree still an approximation to reduce complexity.
Yes, let's start with AABB and see how it goes. I am still not sure how to integrate this structure for the broad phase and a different structure for the narrow phase. I really would like to hide technical details as much as possible.

> Also, maybe we could make it generic : Octree, it would be cleaner. User could always use Octree if they want to mix apples with oranges.
> 
> Also i'm a bit balance between using java collection over libgdx collections. I know java one could performs better in most cases but i think we should stay consistent with libgdx code (i know i'll get a shit storm because of what i say here, i'm ready for it :D )
Let's go full libgdx then! I've never tried libgdx collections before that's why I didn't think about it.

> Feel free to disagree with my comments, i could be wrong. I'll probably be more confident about what i said as we going further with the implementation.
> 
> Thank you for your work.

gdx/src/com/badlogic/gdx/math/spatial/SpatialDatabase.java
```diff
@@ -0,0 +1,18 @@
+package com.badlogic.gdx.math.spatial;
+
+import com.badlogic.gdx.math.Frustum;
+import com.badlogic.gdx.math.Vector3;
+
+import java.util.Set;
+
+public interface SpatialDatabase {
```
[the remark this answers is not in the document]
yuripourre: Removed

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
```diff
@@ -0,0 +1,139 @@
...
+import com.badlogic.gdx.math.spatial.SpatialDatabase;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class Octree implements SpatialDatabase {
+
+    private int maxDepth = 5;
```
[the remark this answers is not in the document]
yuripourre: I will add that in the next commits, too tired atm

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
```diff
@@ -0,0 +1,139 @@
...
+    protected static final int TopRightBack = 2;
+    protected static final int TopLeftBack = 3;
+    // Bottom Level
+    protected static final int BottomLeftFront = 4;
+    protected static final int BottomRightFront = 5;
+    protected static final int BottomRightBack = 6;
+    protected static final int BottomLeftBack = 7;
+
+    private OctreeNode root;
+    private Map<Object, Object> geometryMap = new HashMap<Object, Object>();
```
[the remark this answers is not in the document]
yuripourre: I just completely removed userObject from it. We can add it back later.

gdx/src/com/badlogic/gdx/math/spatial/octree/OctreeNode.java
```diff
@@ -0,0 +1,116 @@
...
+    protected BoundingBox box;
+
+    protected OctreeNode[] children;
+    private Set<Object> geometries = null;
+
+    public OctreeNode(Vector3 min, Vector3 max) {
+        this(min.x, min.y, min.z, max.x, max.y, max.z);
+    }
+
+    public OctreeNode(float x1, float y1, float z1, float x2, float y2, float z2) {
```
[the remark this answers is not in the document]
yuripourre: This method can be useful because of children initialization:
```
children[TopLeftFront] = new OctreeNode(
            box.min.x, midy, midz,
            midx, box.max.y, box.max.z)
            .depth(depth + 1);
```

gdx/src/com/badlogic/gdx/math/spatial/octree/OctreeNode.java
```diff
@@ -0,0 +1,116 @@
...
+import com.badlogic.gdx.math.collision.BoundingBox;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.badlogic.gdx.math.spatial.octree.Octree.*;
+
+public class OctreeNode {
+
+    protected int depth = 0;
```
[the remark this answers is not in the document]
yuripourre: I would never think about it. I will change the code.

gdx/src/com/badlogic/gdx/math/spatial/octree/OctreePointHandler.java
```diff
@@ -0,0 +1,101 @@
...
+        // for each ordinate
+        float midx = (node.getMax().x + node.getMin().x) * 0.5f;
+        float midy = (node.getMax().y + node.getMin().y) * 0.5f;
+        float midz = (node.getMax().z + node.getMin().z) * 0.5f;
+
+        int pos;
+
+        // Deciding the position
+        // where to move
+        if (p.x <= midx) {
```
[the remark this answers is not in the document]
yuripourre: Removed with the rest of point logic

yuripourre:
@mgsx-dev I just updated the code following your precious comments. I didn't added the maxItemsPerNode logic yet.

I quickly ran a debug and it seems to be working.

I will try to find some time this weekend to create a visual "test" example.

Thanks again!

yuripourre:
I just realized that the query method would not work as expected, I will work on that as well.

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
```diff
@@ -0,0 +1,139 @@
...
+import com.badlogic.gdx.math.Frustum;
+import com.badlogic.gdx.math.Vector3;
+import com.badlogic.gdx.math.spatial.SpatialDatabase;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class Octree implements SpatialDatabase {
```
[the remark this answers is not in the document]
yuripourre: I promise I can think about it.

tommyettinger (changes requested):
Mostly, I think there's some rough edges on this but they should be easy to polish up. Making the docs use punctuation would be nice.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+            if (frustum.boundsInFrustum(bounds)) {
+                if (geometries != null) {
+                    result.addAll(geometries);
+                }
+            }
+            return result;
+        }
+
+        Vector3 tmp = new Vector3();
+        public void rayCast (Ray ray, float maxDistance, Collider<T> narrowPhase, RayCastResult<T> result) {
```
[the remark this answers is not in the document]
tommyettinger: I think it could be OK as protected, instead, but it can't have more restrictive access than a method that requires it.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+            if (!isLeaf()) {
+                for (OctreeNode node : children) {
+                    node.getBoundingBox(bounds);
+                }
+            }
+            bounds.add(this.bounds);
+        }
+    }
+
+    public interface Collider<T> {
```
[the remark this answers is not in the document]
mgsx-dev: it should be static

tests/gdx-tests/src/com/badlogic/gdx/tests/math/OctreeTest.java
```diff
@@ -0,0 +1,178 @@
+/*******************************************************************************
```
mgsx-dev, line 1: i think we need a real example : 
* Octree<GameObject> where GameObject class holds a model instance and a cached boundary box.
* Query with frustrum to know which object to render (could be a faked camera to see how octree work ?)
* RayCast on click that could highlight our game object.
* Displaying internal AABB should be enabled/disabled via a toggle button or a keyboard key.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+        super();
+
+        Vector3 realMin = new Vector3(Math.min(minimum.x, maximum.x), Math.min(minimum.y, maximum.y), Math.min(minimum.z, maximum.z));
+        Vector3 realMax = new Vector3(Math.max(minimum.x, maximum.x), Math.max(minimum.y, maximum.y), Math.max(minimum.z, maximum.z));
+
+        this.root = new OctreeNode(realMin, realMax, maxDepth);
+        this.collider = collider;
+    }
+
+    public void add (T aabb) {
```
mgsx-dev, line 26: T variable should be named object

gdx/src/com/badlogic/gdx/math/Octree.java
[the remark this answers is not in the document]
Darkyenus: > it should be static

It is an interface, static is not needed.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+        public OctreeNode (Vector3 min, Vector3 max, int level) {
+            this(min.x, min.y, min.z, max.x, max.y, max.z, level);
+        }
+
+        public OctreeNode (float x1, float y1, float z1, float x2, float y2, float z2, int level) {
+            bounds = new BoundingBox(new Vector3(x1, y1, z1), new Vector3(x2, y2, z2));
+            this.level = level;
+        }
+
+        protected void split (int maxItemsPerNode, Collider<T> collider) {
```
mgsx-dev, line 116: maxItemsPerNode is not needed here, this information is already accessible from parent class.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+
+    protected class OctreeNode {
+
+        protected int level;
+
+        // Represents the boundary of the node
+        private BoundingBox bounds;
+
+        private Array<OctreeNode> children;
+        private ObjectSet<T> geometries = null;
```
mgsx-dev, line 105: should be final and with maxItemsPerNode capacity

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+
+        public OctreeNode (Vector3 min, Vector3 max, int level) {
+            this(min.x, min.y, min.z, max.x, max.y, max.z, level);
+        }
+
+        public OctreeNode (float x1, float y1, float z1, float x2, float y2, float z2, int level) {
```
mgsx-dev, line 111: this constructor is not needed

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+                    child.add(geometry, maxItemsPerNode, collider);
+                }
+            }
+        }
+
+        public boolean contains (Collider<T> boundingBoxHandler, T geometry) {
+            return boundingBoxHandler.intersects(bounds, geometry);
+        }
+
+        public void add (T geometry, int maxItemsPerNode, Collider<T> collider) {
```
mgsx-dev, line 148: collider is not needed, it's already accessible from parent class

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+                for (OctreeNode child : children) {
+                    child.add(geometry, maxItemsPerNode, collider);
+                }
+                return;
+            }
+
+            if (!collider.intersects(bounds, geometry)) {
+                return;
+            }
+
```
mgsx-dev, line 160: if/else statement could be more readable and cleanr than returning in the middle of the process

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+                    node.remove(aabb);
+                }
+            }
+        }
+
+        public boolean isLeaf () {
+            return children == null;
+        }
+
+        private void addGeometry (T geometry) {
```
mgsx-dev, line 183: not necessary if geometries is final (see my previous comment)

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+            geometries.add(geometry);
+        }
+
+        public ObjectSet<T> query (BoundingBox aabb, ObjectSet<T> result) {
+            if (!isLeaf()) {
+                for (OctreeNode node : children) {
+                    node.query(aabb, result);
+                }
+            }
+            if (aabb.contains(bounds)) {
```
mgsx-dev, line 196: only if isLeaf

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+                if (geometries != null) {
+                    result.addAll(geometries);
+                }
+            }
+            return result;
```
mgsx-dev, line 201: else if aabb intersect with bounds, we need to add some geometries using "collider" to filter them

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+
+        Vector3 tmp = new Vector3();
+        public void rayCast (Ray ray, float maxDistance, Collider<T> narrowPhase, RayCastResult<T> result) {
+            // Check intersection with node
+            boolean intersect = Intersector.intersectRayBounds(ray, bounds, tmp);
+            if (!intersect) {
+                return;
+            } else {
+                float dst2 = tmp.dst2(ray.origin);
+                if (dst2 > maxDistance) {
```
mgsx-dev, line 226: superior to maxDistance squared

gdx/src/com/badlogic/gdx/math/Octree.java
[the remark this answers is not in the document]
mgsx-dev: also, narrowPhase is not needed, collider is accessible from parent class.
also maxDistance could be in result

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+     * @param boxes
+     */
+    public ObjectSet<BoundingBox> getNodesBoxes (ObjectSet<BoundingBox> boxes) {
+        root.getBoundingBox(boxes);
+        return boxes;
+    }
+
+    protected class OctreeNode {
+
+        protected int level;
```
mgsx-dev, line 99: should be final

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+
+        // Represents the boundary of the node
+        private BoundingBox bounds;
```
mgsx-dev, line 102: final as well

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+     * @param frustum - The frustum to query
+     * @param result set populated with objects near from the frustum
+     */
+    public ObjectSet<T> query (Frustum frustum, ObjectSet<T> result) {
+        root.query(frustum, result);
+        return result;
+    }
+
+    public T rayCast (Ray ray) {
+        return rayCast(ray, Float.POSITIVE_INFINITY);
```
mgsx-dev, line 69: Float.MAX_VALUE instead

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    public T rayCast (Ray ray, float maxDistance) {
+        RayCastResult<T> result = new RayCastResult<>();
```
mgsx-dev, line 73: should be provided by the caller. User code may need ray distance : boolean RayCast(ray, result, maxDistance)

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+        result.distance = Float.NEGATIVE_INFINITY;
```
mgsx-dev, line 74: maxDistance

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+        root.rayCast(ray, maxDistance, collider, result);
+        return result.geometry;
+    }
+
+    public void setMaxDepth (int maxDepth) {
+        this.maxDepth = maxDepth;
+        this.root.setLevel(maxDepth);
```
mgsx-dev, line 81: it's not enough, all nodes should be updated but i'm not sure maxDepth and maxItemsPerNode have to be modifiable after construction, we'll see later if it's a necessary feature, i doubt it.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    public void setMaxItemsPerNode (int maxItemsPerNode) {
+        this.maxItemsPerNode = maxItemsPerNode;
+    }
+
+    /**
+     * Method to get nodes boxes. Useful for render purpose
```
mgsx-dev, line 89: for debug purpose

mgsx-dev (changes requested):
we also need to think about the "update" case and a proper "remove" : 

* some dynamic objects could move and we would need to update it in the tree (probably remove it and add it again for now, but it could be optimized in the future).
* when removing an object, we may want to merge some nodes in order to reduce memory footprint. maxItemsPerNode should be true at any time (except for the last depth level of course).

Considering these 2 use cases, we don't want to generate garbage so we may want to handle nodes lifecycle a bit differently : 

* by using a node pool (Pool<Node>)
* by having final Array<Node> children. An empty children array would means it's a leaf node.

gdx/src/com/badlogic/gdx/math/Octree.java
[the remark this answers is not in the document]
mgsx-dev: my bad :+1:

mgsx-dev:
@yuripourre i prefer to keep it as a draft PR until it becomes stable.

gdx/src/com/badlogic/gdx/math/Octree.java
Darkyenus: Why?

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Only leaf nodes have geometries. If I make it final, all the nodes will have it allocated.

gdx/src/com/badlogic/gdx/math/Octree.java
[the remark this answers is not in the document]
yuripourre: @tommyettinger I made it public, it can be useful.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Nice catch!

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: I did it in a previous implementation. Ideally, collider should also be capable of filtering by frustum so they can have a similar behavior, what do you think?

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Right! What if `maxDistance = Float.MAX_FLOAT;`
Should I set the default maxDistance to `Math.sqrt(Float.MAX_FLOAT)` instead?

tests/gdx-tests/src/com/badlogic/gdx/tests/math/OctreeTest.java
yuripourre: Done!
Screenshot from 2020-11-11 22-47-59

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Added a method to the collider, makes sense to have it there.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Nvm, fixed.

gdx/src/com/badlogic/gdx/math/Octree.java
mgsx-dev: we need pooling anyway because we don't want GC pollution during updates, so that would be better to have them always allocated actually.

yuripourre: Yessir

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: I removed the final so it can be used by the node pool. Is it possible to have both?
@mgsx-dev

mgsx-dev: no, we can't have both, this kind of changes are indeed necessary when working with pools.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,307 @@
+package com.badlogic.gdx.math;
+
+import com.badlogic.gdx.math.collision.BoundingBox;
+import com.badlogic.gdx.math.collision.Ray;
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.ObjectSet;
+
+public class Octree<T> {
```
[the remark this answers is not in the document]
yuripourre: Done

gdx/src/com/badlogic/gdx/math/Octree.java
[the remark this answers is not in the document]
yuripourre: Docs added.

yuripourre:
@mgsx-dev
Do you think this PR is ready for review?
Should I squash the commits?

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -347,6 +347,32 @@ public static boolean intersectSegmentCircle (Vector2 start, Vector2 end, Circle
 		return v2a.len2() <= circle.radius * circle.radius;
 	}
 
+	/** Returns whether the given {@link Frustum} intersects a {@link BoundingBox}.
+	 * @param frustum The frustum
+	 * @param bounds The bounding box
+	 * @return Whether the frustum intersects the bounding box*/
+	public static boolean intersectFrustumBoundsFast(Frustum frustum, BoundingBox bounds) {
```
NathanSweet, line 354: Does it need `Fast` in the name?

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+		return boxes;
+	}
+
+	protected class OctreeNode {
+
+		private int level;
+
+		// Represents the boundary of the node
+		private final BoundingBox bounds;
+		private final Array<OctreeNode> children = new Array<>(8);
```
NathanSweet, line 158: I suppose it could be ` new OctreeNode[8]` or null for a leaf, but this is probably fine too.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+		private final ObjectSet<T> geometries = new ObjectSet<T>(maxItemsPerNode);
```
NathanSweet, line 159: Note all nodes will allocate space to hold `maxItemsPerNode * 1.25` (as the default load factor is 0.8), even if they have much fewer or none. If it's not worth optimizing then just ignore this comment.

Does it need to be a set? I assume to avoid duplicates? If the number of items is low, an array may be sufficient. If a set is really needed to avoid n^2 when de-duping and the occtree is intended to be static, then maybe the set is only needed for construction. The drawback to using a set after construction is iteration (which appears to be done a lot) must check every entry in the backing table. One way to mitigate that is to use an OrderedSet, which keeps an array along with the set, then iteration uses the array. When using an OrderedSet for iteration when the order (ironically) doesn't actually matter, consider using `orderedItems().ordered = false;` for faster removal, as mentioned in the OrderedSet class javadocs.

NathanSweet (commented):
I gave her a once over. Not sure my comments are terribly useful, but here they are! Overall it looks nice.

NathanSweet:
> Should I squash the commits?

There's no need to squash PRs. Whoever merges this PR can (should!) choose to squash (if they forget, they will be beaten).

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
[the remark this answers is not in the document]
NathanSweet: > I'm sure 2D people would want a QuadTree too (we could add this when the Octree is completed)

@mgsx-dev Who said 2D people?! I'm 2D people! I have a QuadFloatTree class I could contribute if others would find it useful.

gdx/src/com/badlogic/gdx/math/Intersector.java
yuripourre: I usually use the `Fast` suffix when it returns a boolean. I can remove it.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: That's interesting, I can try to run a few (quick) benchmarks.

Thank you for the tip.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: I swear I tried that but the code doesn't compile. it says `Generic array creation`.

gdx/src/com/badlogic/gdx/math/Intersector.java
NathanSweet: Not sure how fast in the name is related to returning a boolean? :) Seeing "fast" in the name makes me think the function is taking shortcuts and maybe losing precision or accuracy compared to some other way of doing the same operation.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: @NathanSweet I changed to array. Thank you for the suggestion.

gdx/src/com/badlogic/gdx/math/Intersector.java
mgsx-dev: typically "non fast" methods returns more information and need more calculations (eg. collision point on a sphere), "fast" method only detect collision and simply returns a boolean.
I can't see what kind of information frustum/AABB collision detection could return so i agree, "fast" could be removed.

gdx/src/com/badlogic/gdx/math/Octree.java
mgsx-dev: problem is, Octree is not only used for static scene, we could have a lot of moving objects, not sure moving from ObjectSet to Array is a good idea.

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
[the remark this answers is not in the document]
mgsx-dev: @NathanSweet I think it would be great to have it! PR welcome! ;-)

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+	/** Method to get nodes as bounding boxes. Useful for debug purpose.
+	 *
+	 * @param boxes
+	 */
+	public ObjectSet<BoundingBox> getNodesBoxes (ObjectSet<BoundingBox> boxes) {
+		root.getBoundingBox(boxes);
+		return boxes;
+	}
+
+	protected class OctreeNode {
```
Darkyenus, line 152: Is there a reason why is this nested class not static? If it was, the node pool could be static as well and I suspect that it would allow `children` to be raw array as well.

mgsx-dev: we need to avoid garbage as much as possible especially for dynamic stuff, children as raw array wouldn't be better.

sharing pool across multiple Octrees could be an issue, geometry array could be small in some case an big in other case which could lead to unnecessary heavy allocations. This could happen within a single octree though but it would be less annoying (isolated) than with a shared pool.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+			}
+
+			// If is not leaf, check children
+			if (!isLeaf()) {
+				for (OctreeNode child : children) {
+					child.add(geometry);
+				}
+			} else {
+				geometries.add(geometry);
+				if (geometries.size > maxItemsPerNode && level > 0) {
```
mgsx-dev, line 227: better check size before adding new item to avoid unnecessary array resizing.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+			return node;
+		}   
+
+		private void freeNode(OctreeNode node) {
+			node.geometries.clear();
+			node.children.clear();
+			nodePool.free(node);
+		}
+
+		protected boolean contains (Collider<T> boundingBoxHandler, T geometry) {
```
mgsx-dev, line 211: not sure to understand what this method is and what it's used for.. ?
also, contains / intersects is a bit confusing.

gdx/src/com/badlogic/gdx/math/Octree.java
```diff
@@ -0,0 +1,382 @@
...
+				}
+			}
+		}
+
+		protected boolean remove (T object) {
+			if (!isLeaf()) {
+				boolean removed = false;
+				int childrenSum = 0;
+				for (OctreeNode node : children) {
+					childrenSum += node.geometries.size;
```
mgsx-dev, line 238: size may change after removing, i think you can swap those lines.

gdx/src/com/badlogic/gdx/math/Octree.java
Darkyenus: > we need to avoid garbage as much as possible especially for dynamic stuff, children as raw array wouldn't be better.

I was referring to the related discussion around the `children` above, I am not arguing for or against that change, as I have not read the code that thoroughly. That comment was only about the static-ness of `OctreeNode`.

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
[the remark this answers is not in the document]
NathanSweet: Alrighty! It's relatively simple and barebones, but I do have (visual) tests for it. I'll just commit it though. If there's feedback, can post it on the commit or PR to change it.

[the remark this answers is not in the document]
NathanSweet: 546a7a4487818285993d97a544b61eae631ed224

gdx/src/com/badlogic/gdx/math/Octree.java
mgsx-dev: Sorry i have deep knowledge of that class now :D, so i got directly to the point of shared pool. 
To answer your question, OctreeNode use some information from Octree (collider, maxItems), we could pass them to OctreeNode, but not sure it worth it.

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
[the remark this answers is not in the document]
NathanSweet: Sorry for the noise, but I should have cleaned it up better before committing. Have done so now. 1a1cadfc34360837d7f4196cbaf71b784dce5958 Cheers!

tommyettinger (approved):

gdx/src/com/badlogic/gdx/math/spatial/octree/Octree.java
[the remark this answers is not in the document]
mgsx-dev: @NathanSweet apparently your last commit breaks CI, see https://github.com/libgdx/libgdx/pull/6306/checks?check_run_id=1520336735#step:7:632

no offense but a PR would be highly appreciated next time (CI compilation and checks, nice review workflow, less noise in other PR, etc) :-)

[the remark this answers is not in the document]
NathanSweet: It was supposed to be a quick and easy addition, but it seems you are right. Fixing...

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: Nice catch, this method is not necessary anymore.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: You are right! Done.

gdx/src/com/badlogic/gdx/math/Octree.java
yuripourre: @mgsx-dev 
About that, may not sound optimal, but adding the geometry first, if there is a split, the geometry is added to the leaf nodes inside the split method. If I change the order, the geometry needs to be added to the children. Do you think it's better?

```
if (geometries.size > maxItemsPerNode && level > 0) {
    split();
    for (OctreeNode child : children) {
        child.add(geometry);
    }
} else {
    geometries.add(geometry);
}
```

mgsx-dev: `if (geometries.size >= maxItemsPerNode && level > 0)` :-)
yeah, that's fine, it's optimal. The problem was that we initialize geometries with 
maxItemsPerNode capacity, so adding new item before required to resize and copy the whole array for nothing.

yuripourre: Oh, I missed that, you are right!
This change is hard to decide, I am more inclined to leave it as is.

mgsx-dev: the code you proposed in your comment does fix it except it should use >= check instead of >

yuripourre: Done!

mgsx-dev: i think you forgot the else from your snippet

yuripourre:
@mgsx-dev 
> nice catch, i forgot about duplicates. However this strategy is not optimal : temporary ObjectSet will produce garbage and doing this for each removal is maybe not a good idea, maybe we should introduce a new method "pack" called by user code when it feels necessary to optimize the tree, what do you think?

I think we can leave it as-is atm and update in the future. Do you agree with that?

yuripourre:
Fixed conflicts.

NathanSweet:
Is she all done? :)

yuripourre:
@NathanSweet 
> Is she all done? :)
I think so, I don't have any plans to change the implementation. I hope it can be useful for other developers

NathanSweet:
@yuripourre It is neat, thanks for sharing! I made a few updates, see 85db7faefd396c84c516c6ea5ba098da8f62083c. Let me know if you hate any of my changes and we'll revert them.

yuripourre:
@NathanSweet not at all! Thanks for the upgrade! ❤️

obigu:
There's this warning when compiling libGDX which probably causes Octree not to work on GWT:
```
   Resolving com.badlogic.gdx.math.Octree.OctreeNode
      Found type 'com.badlogic.gdx.math.Octree.OctreeNode'
         Resolving field children
            [ERROR] Unable to resolve inner class OctreeNode in com.badlogic.gdx.math.Octree[]
com.google.gwt.core.ext.typeinfo.NotFoundException
    at com.google.gwt.dev.javac.typemodel.JArrayType.getNestedType(JArrayType.java:157)
    at com.google.gwt.dev.javac.asm.ResolveTypeSignature.visitInnerClassType(ResolveTypeSignature.java:162)
    at org.objectweb.asm.signature.SignatureReader.a(Unknown Source)
    at org.objectweb.asm.signature.SignatureReader.a(Unknown Source)
    at org.objectweb.asm.signature.SignatureReader.acceptType(Unknown Source)
    at com.google.gwt.dev.javac.CompilationUnitTypeOracleUpdater.resolveField(CompilationUnitTypeOracleUpdater.java:1073)
```

I don't do GWT or Octrees but setting OctreeNode to public seems to fix it but haven't tested properly.

yuripourre:
@obigu thanks for reporting, I will try that!

PR #6231, opened 2020-10-15
Update gdx-setup's Gradle version to 6.7, AGP to 3.4.3, official extension versions

tommyettinger:
This allows gdx-setup to actually complete its generation if the user only has Java 13, 14, or 15, or has one of those recent versions in their JAVA_HOME. The wrapper files are also updated. No changes appear to be needed to any build.gradle files for desktop or android, but I can't try ios to see if Gradle 6.7 is compatible. The html platform works when using libGDX 1.9.12-SNAPSHOT, but of course fails on the current stable version (1.9.11) for the same reasons it failed before. The Android Gradle plugin is updated to 3.4.3 for maximum compatibility while avoiding a warning from IDEA about 3.4.1 being outdated (4.0.2 is newer, but it might require a newer IDE version). Lastly, gdx-ai is updated to 1.8.2, which is a year old instead of the 3-and-a-half-year-old 1.8.0 version.

EDIT: This also now updates Ashley to 1.7.3 and box2dLights to 1.5, which appear to be the most recent releases for each. The Box2d Lights readme still points to 1.4, so I'm not sure if 1.5 is discouraged for some reason, or if no one updated the readme. That is all of the outdated official extensions I could find.

It would be best if someone tested this on iOS at some point, but that and Android Studio support are the only areas I'm especially concerned about breakage. It is possible Eclipse can't yet handle Gradle 6.7, but I don't know if it does, doesn't, or soon will. I've generated a project with this and it works fine even using Java 15 to run gdx-setup.jar and for the project itself in IDEA.


tommyettinger:
I think it would be good if this could be tested by at least a few people with different usage scenarios before the planned 1.10.0 release, since I can only anticipate new users being confounded by the "must have old Java" requirement.

extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
```diff
@@ -42,14 +42,14 @@
 	//Project plugins
 	static String gwtPluginImport = "org.wisepersist:gwt-gradle-plugin:1.0.13";
 	static String grettyPluginImport = "org.gretty:gretty:3.0.2";
-	static String androidPluginImport = "com.android.tools.build:gradle:3.4.1";
+	static String androidPluginImport = "com.android.tools.build:gradle:3.4.3";
```
MrStahlfelge, line 45: Just as a note, there's already 4.0.2 available.

extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
```diff
@@ -42,14 +42,14 @@
...
 	static String roboVMPluginImport = "com.mobidevelop.robovm:robovm-gradle-plugin:" + roboVMVersion;
 
 
 	//Extension versions
 	static String box2DLightsVersion = "1.4";
 	static String ashleyVersion = "1.7.0";
-	static String aiVersion = "1.8.0";	
+	static String aiVersion = "1.8.2";
```
MrStahlfelge, line 52: As this has nothing to do with Gradle, this change should be mentioned in PR title, too.

MrStahlfelge (commented):
I have tested on Mac with Android Studio 4.0:
* project is configured successfully
* running Android from IDE works
* running iOS form IDE works

MrStahlfelge:
Requested review by @mgsx-dev to check Eclipse compatibility.

extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
tommyettinger: I was just wondering why gdx-ai is stuck on a 3-or-more-year-old version. It's supposed to be a minor update, anyway, and I always use 1.8.2 in my code that uses gdx-ai.

extensions/gdx-setup/src/com/badlogic/gdx/setup/DependencyBank.java
tommyettinger: I wasn't sure if all versions of AS were compatible yet. I didn't want to invite any unnecessary breakage by changing a major version.

obigu:
> I have tested on Mac with Android Studio 4.0:
> 
> * project is configured successfully
> * running Android from IDE works
> * running iOS form IDE works

Just to confirm, when you say running from IDE you mean explicitly running Gradle tasks from IDE or using IDE plugins and built-in features?

In the case of iOS MobiVM, IDEA plugin doesn't use Gradle.
On Android, depending on how you run stuff, it will also not use Gradle directly.

MrStahlfelge:
Using ide plugins and built-in features. Using Gradle itself was tested by tettinger, for me it was important that everything still works with AS 4.0 (as 4.1 is still not very widely used)

obigu:
> Using ide plugins and built-in features. Using Gradle itself was tested by tettinger, for me it was important that everything still works with AS 4.0 (as 4.1 is still not very widely used)

ok, have iOS tasks been tested with Gradle @tommyettinger?

tommyettinger:
@obigu I have done zilch with iOS; I do not own anything that runs MacOS, and the only iOS devices I have nearby are not mine (my parents' relatively recent iPhones and iPads). If the tasks are usable (meaningfully) on Windows (7) or Linux, I can test there.

obigu:
I've tested launching the app on iOS device and simulator (with default RoboVM version 2.3.10) with Gradle and it has worked.

mgsx-dev (approved):
Tested successfully with Eclipse. (sorry for the delay guys)

Side node (not related to this PR): since #6169 introduced a breaking change in HTML template, i needed to adapt code a bit. I guess it'll be in sync when 1.10 get released. So right now gdx-setup artifact is broken.

Issue #6230, opened 2020-10-15
glDrawBuffers ignored buffer size for LWJGL backends 


FreakingBarbarians:
#### Issue details
`glDrawbuffers(int n, java.nio.IntBuffer bufs)` has some implementations that ignore `n` and works off the size of `buf` exclusively.
Thus if you pass in a larger buffer, but only use the first n positions you'll most likely get an invalid error inside of OpenGL.

#### Reproduction steps/code
```kotlin
val buf = glUtils.createIntBuffer(32)
for(i in 0 until 32) {
            colorAttachments[i]?.let {
                buf.put(GL20.GL_COLOR_ATTACHMENT0 + i)
            }
        }
        buf.rewind()
gl.glDrawBuffers(colorAttachments.size, buf)
```

The following code causes the error. The `colorAttachments.size` argument in glDrawBuffers is ignored.

A workaround would be to just let the `intBuffer` be the size of `colorAttachments.size` but it's a non-perfect solution since other first-time users or users familiar with openGL call semantics would expect `n` to limit the elements touched inside of the buffer.

```java
Lwjgl3GL30.java::glDrawBuffers line 183
	@Override
	public void glDrawBuffers (int n, IntBuffer bufs) {
		GL20.glDrawBuffers(bufs);
	}
(same deal in LwjglGL30.java)
```

#### Version of LibGDX and/or relevant dependencies
1.9.12-SNAPSHOT

#### Stacktrace
```java
updateDrawTargets:62, BasicFrameBuffer (com.nanite.graphics.impl)
...
initializeListener:433, Lwjgl3Window (com.badlogic.gdx.backends.lwjgl3)
update:381, Lwjgl3Window (com.badlogic.gdx.backends.lwjgl3)
loop:141, Lwjgl3Application (com.badlogic.gdx.backends.lwjgl3)
<init>:117, Lwjgl3Application (com.badlogic.gdx.backends.lwjgl3)
main:22, DesktopLauncher (com.cellular.game.desktop)
```

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] MacOS

Although I imagine other platforms would be affected as well.

As far as solutions are concerned I thought of either making a glDrawbuffers without n as an argument, and simply throwing an exception where its called in implementations that are not LWJGL, and vice versa for LWJGL. But that also doesn't feel quite right.

mgsx-dev:
[...]

PR #6232, opened 2020-10-16
glDrawBuffers no longer ignores bufsize (n) argument

FreakingBarbarians:
fixes https://github.com/libgdx/libgdx/issues/6230

Previously in LWJGL backends `glDrawbuffers` would rely on the `remaining()` of the `IntBuffer` passed in, ignoring the bufsize argument (`n`). This would possibly cause errors when users would expect (`n`) to limit the number of args passed to opengl from the buffer.

Now the buffer is limited by (`n`) properly

I've tested this with framebuffers on my own project and they work.


mgsx-dev (approved):
LGTM

@FreakingBarbarians next time you can add "fixes #6230" in your commit message, it'll automatically close the issue when PR get merged ;-)

PR #6233, opened 2020-10-16
Fix #6228. Removed delta smoothing on Android backend

### LINKED ISSUE #6228: Smoothed Delta times on Android and overall improvements
`Graphics` has 2 different methods to get delta time, `float getDeltaTime()` and `float getRawDeltaTime();`. According to java docs, the difference is that `getDeltaTime()` "Might be smoothed over n frames.".

Checking all backend `Graphics` implementations, the only backend that does smoothing on the delta is the Android backend, that applies a simple size 5 window mean. This is probably due to legacy reasons when libGDX was created some Android devices were really unrealiable (we're talking Android 1.5 or 1.6) and my guess is that this was a workaround for those old devices with very inconsistent delta times.

Averaging the delta over 5 frames has a significant impact on how the game behaves during lag spikes. Let me show with an example.

Imagine we have the following sequence of Raw delta times
1/60 1/60 1/60 1/60 1/60 **1/5** 1/60 1/60 1/60 1/60 1/60

In every backend except from Android, in a typical setup in which delta is capped to 1/30 (Scene2D does that by default) the resulting deltas used by the game would be  
1/60 1/60 1/60 1/60 1/60 **1/30** 1/60 1/60 1/60 1/60 1/60 1/60
In which only the 1/30 frame would have a "slow motion" distortion caused due to real elapsed time being higher than delta time

On Android, on the other hand, we would get
1/60 1/60 1/60 1/60 1/60 **1/30 1/30 1/30 1/30 1/30** 1/60
In which the first 1/30 frame would be distorted exactly like before but the next 4 1/30 ones they would be "fast forward" (as if it was catching up).

Understanding that
- Different games have different needs (not the same a multiplayer racing game than a puzzle game) and may want to deal differently with lag spikes or delta inconsistencies.
- It is a complex topic and there's not a perfect solution. See for example https://blogs.unity3d.com/2020/10/01/fixing-time-deltatime-in-unity-2020-2-for-smoother-gameplay-what-did-it-take/ for recent Unity improvements or https://frankforce.com/frame-rate-delta-buffering/ for a strategy working on a specific game.

I think there are several things of this design decision that are worth revisiting:

1. I'm not aware of any reason why Android should be treated differently from the other backends. Considering the algorithm applied is not particularly good (most smoothing approaches go for a weighted average or some kind of low filter) and in many scenarios the current solution will impact how lag is perceived in games and feel worse than raw delta. I think we should remove the smoothing from Android.
2. The "maybe smoothed" version `getDeltaTime()` is used across libGDX code base by different systems to get the delta time (with calls to `Gdx.graphics.getDeltaTime()`). This means that, even if a user didn't want the smoothed delta to be used on his game, some parts of libGDX (from Scene2D to particles among others) would still use the smoothed one causing unpredictable issues.
3. As an larger scoped improvement (that would entail breaking backwards compatibility) I think it's worth evaluating adding optional Delta capping and the ability to provide a Smoothing strategy at framework level. We could also consider having `getDeltaTime()` to return raw delta time and add a `getSmoothedDeltaTime()` (Unity has something on these lines).

obigu:
Fixes #6228.

- `getRawDelta()` has been deprecated as, at least for now, it's a duplicate of `getDelta()`.
- the line `mean.clear()` on `setContinuousRendering()` can be safely removed because it was added to fix an issue caused by the smoothing (https://code.google.com/archive/p/libgdx/issues/962).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@
 - [BREAKING CHANGE] iOS: Changed how Retina/hdpi handled on iOS. See #3709.
 - [BREAKING CHANGE] API Change: InputProcessor scrolled method now receives scroll amount for X and Y. Changed type to float to support devices which report fractional scroll amounts. Updated InputEvent in scene2d accordingly: added scrollAmountX, scrollAmountY attributes and corresponding setters and getters. See #6154.
 - [BREAKING CHANGE] API Change: Vector2 angleRad(Vector2) now correctly returns counter-clockwise angles. See #5428
+- [BREAKING CHANGE] Android: getDeltaTime() now returns the raw delta time instead of a smoothed one. See #6228.
 - Fixed vertices returned by Decal.getVertices() not being updated
 - Fixes Issue #5048. The function Intersector.overlapConvexPolygons now should return the right minimum translation vector values.
 - Update to MobiVM 2.3.10
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -76,7 +76,6 @@ public class AndroidGraphics implements Graphics, Renderer {
 	protected long frameId = -1;
 	protected int frames = 0;
 	protected int fps;
-	protected WindowedMean mean = new WindowedMean(5);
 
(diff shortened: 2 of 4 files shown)


crykn (approved):
Looks good to me!

PR #6236, opened 2020-10-16
Add protected modifier to mainLoop method

Dgzt:


crykn:
Is there any reason in particular why you want to override/call `mainLoop()`?

Dgzt:
Hi!

I want to use it for tests and I want to disable continuous rendering. I created a custom Application and use these "renderUntilUnitMoving", "renderUntilBattleFinish" methods. 

And if this method will be protected I can extends HeadlessApplication and override this method.

crykn (approved):
That sounds reasonable. It is the same on the other backends: `LwjglApplication#mainLoop()`, for instance, is protected as well. 

Alternatively, we could provide an implementation of non-continuous rendering for the headless application, but I'm not sure whether that would be used that often.

PR #6238, opened 2020-10-17
Replace travis with github actions build

SimonIT:
This is basically the same build which currently runs on travis.

We have to wait probably on #5976 or #6240, because the current build fails without the fixes included there.


.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Setup ccache
+        run: |
+          brew install ccache
+          echo "/usr/local/opt/ccache/libexec" >> $GITHUB_PATH
+      - name: Build macOS natives
+        run: ant -f build-mac-ios.xml
+
+  linux:
+    needs: macos
```
PokeMMO, line 36: Since this does not depend on uploading macOS natives for linux to use, we do not need this line.

PokeMMO (approved):

PokeMMO:
https://github.com/libgdx/libgdx/actions/runs/320268639

There's some deprecation warnings that we should fix.

SimonIT:
I saw them. The warnings are from the setup-android action. I don't know how to fix them because the functions are used internally

PokeMMO:
Oh, sorry, you're right. Nothing for us to do there then.

PR #6240, opened 2020-10-17
Fixes libGDX Gradle build error

obigu:
Building libGDX project with Gradle was failing (just run `gradlew clean build` to test) due to a bad workaround I added to fix Android and Desktop tests not running on IDEA (https://github.com/libgdx/libgdx/commit/a30ecf0c1f41f9f2e6a7403520e0bdcd7f4bcb72).

This alternative workaround fixes the Gradle issue and keeps IDEA happy (thanks to PokeMMO).

### DIFF
--- a/gdx/build.gradle
+++ b/gdx/build.gradle
@@ -41,7 +41,7 @@ test {
 
 // Workaround needed for IDEA to have resources on classpath when running tests (like gdx-tests-lwjgl3)
 task copyIdeaResources(type: Copy) {
-    from "${projectDir}/src/main/resources"
+    from "${projectDir}/res"
     into "${buildDir}/classes/java/main/"
 }
 processResources.dependsOn copyIdeaResources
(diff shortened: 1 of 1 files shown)


SimonIT (approved):
`gradlew build` works now🎉

gdx/build.gradle
```diff
@@ -42,3 +38,10 @@ test {
         events "passed", "skipped", "failed", "standardOut", "standardError"
     }
 }
+
+// Workaround needed for IDEA to have resources on classpath when running tests (like gdx-tests-lwjgl3)
+task copyIdeaResources(type: Copy) {
+    from "${projectDir}/src/main/resources"
```
PokeMMO, line 44: Should probably be `${projectDir}/res` for libgdx since it still uses  non-standard folders.

obigu: thanks, your're right, fixed

tommyettinger:
Looks good; I'm glad to hear the Gradle build is working again. I never know which build tool to primarily use with the libGDX repo...

Issue #6207, opened 2020-10-08
MSAA is not applied in LWJGL3 when new window is created


fgnm:
#### Issue details
According to that lines : https://github.com/libgdx/libgdx/blob/5b4aa04c73d995c40932348bd7e16f0f30e82982/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L428-L436 when a new LWJGL3 window is create all of that `glfwWindowHint` are not set. Indeed, only the very first main window (that has `sharedContextWindow = 0`) will receive that instructions, but newly created ones doesn't.

This means that MSAA samples are applied only to the first window.

I've tried to force the `sharedContextWindow == 0` to `true` with a debugger to get `glfwWindowHint` executed even on other windows and MSAA is finally applied properly.

#### Reproduction steps/code
```Java
import com.badlogic.gdx.*;

public class Barebones extends ApplicationAdapter {
	public void create () {
		Lwjgl3Application app = (Lwjgl3Application) Gdx.app;
                Lwjgl3ApplicationConfiguration config = new Lwjgl3ApplicationConfiguration();
                config.setBackBufferConfig(8,8,8,8,16,8,16);
                app.newWindow(new Barebones2(), config);
	}

	public void render () {
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
		// your code here
	}

	public static void main (String[] args) throws Exception {
                Lwjgl3ApplicationConfiguration config = new Lwjgl3ApplicationConfiguration();
                config.setBackBufferConfig(8,8,8,8,16,8,16);
		new LwjglApplication(new Barebones(), config);
	}
}
```
In this case the main first window has correctly applied MSAA while the second one not.

In my case the difference is highly appreciable, this is a screen of my application of a secondary window:
![immagine](https://user-images.githubusercontent.com/5543339/95475853-f2e08480-0986-11eb-8935-3a7cfc34699e.png)

This is instead how it looks if rendered on the first main window:
![immagine](https://user-images.githubusercontent.com/5543339/95475666-c2004f80-0986-11eb-9462-329bf0b01de2.png)

[...]

PR #6241, opened 2020-10-18
Set back buffer configs for any window

fgnm:
As point out in #6207, I can't see any downside to enable back buffer for all created windows rather then the only first one.
This commit will fix the MSAA issue for multiple windows and allow users to set custom back buffer configuration for any window.

I've tested this in some environments and I haven't noticed any unexpected behavior, also, I can't find online any reason why this was made in the past, please check it.

Thanks


crykn:
Hey @fgnm, I also couldn't see any unexpected side-effects when using your workaround, both on Windows and on MacOS. However, I am not sure and would like to know, why this change was introduced in the first place. My guess would be some open gl context related stuff, but I couldn't find any information on that either, except that GLFW_RED_BITS, GLFW_STENCIL_BITS, GLFW_DEPTH_BITS, etc. are all framebuffer related hints, while GLFW_RESIZABLE, etc. are window related hints.

fgnm:
I've looked in depth into this and I can't find any reason (a part from testing purposes) to not set framebuffer related hints to any windows. To point out some documentation:

- https://www.glfw.org/docs/3.0/window.html
- https://www.glfw.org/docs/3.0/group__window.html#ga4fd9e504bb937e79588a0ffdca9f620b

As I understand well, any hint could be set to any window without any particular attention (just set them before window creation).
Anyway I'd like to know why this was introduced in the past, but I think that only @badlogic (sorry for ping) can help us to understand 😃

crykn:
@fgnm did you test this on Linux? If someone could confirm that it works there as well, I would be inclined to approve/merge it.

fgnm:
Yes, I use ubuntu as daily driver machine and works as expected

crykn (approved):

crykn:
Another two weeks and no-one has commented, so I'm going to merge this. If there were any unexpected problems, the snapshot should give us time to identify those.

PR #6243, opened 2020-10-21
Alternative asynchronous sound  Android Audio implementation

obigu:
This PR adds an alternative audio implementation (`AsynchronousAndroidAudio`) aimed at improving sound performance.

Current sound implementation has a very poor performance in scenarios in which several sounds are played at the same time. This is caused by the main thread being locked each time a method that returns a "sound id" such as `float play()` is called. For some reason, Android takes much longer to return a sound id in these scenarios and no configuration of the `SoundPool` seems to fix it.

The proposed approach is to play the sounds on a separate thread. The drawback of this approach is that operations that use `soundId` are no longer supported but that's not a big deal in most scenarios. I personally use this implementations in all my games and the results are great.

For an app to use this alternative implementation it just needs to override `createAudio()` on `AndroidApplication` and return a new `AsynchronousAndroidAudio` instance.

There are no profiling measurements or tests linked on this PR atm. I did a lot long time ago for my implementations and, if really necessary, I can try to provide some but I'd rather confirm the interest on this first.

**This issue has been reported plenty of times on the forums, Discord and SO and is one of the most reported and recurrent issues on Android backend.**

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -51,6 +51,7 @@
 - API Change: Upon changing the window size with the lwjgl3 backend, the window is centered on the monitor.
 - Fixed DepthShaderProvider no longer creates one DepthShader per bones count. Now it creates only one skinned variant and one non-skinned variant based on DepthShader/Config numBones.
 - API Addition: Added Intersector#intersectPlanes to calculate the point intersected by three planes, see https://github.com/libgdx/libgdx/pull/6217
+- API Addition: Added alternative Android Audio implementation for performant sound. See https://github.com/libgdx/libgdx/pull/6233.
 
 [1.9.11]
 - Update to MobiVM 2.3.8
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
@@ -5,6 +5,13 @@ import android.os.Handler;
 import android.os.HandlerThread;
 import com.badlogic.gdx.audio.Sound;
 
+/** A performance oriented implementation of the {@link AndroidAudio} interface.
+ * 
+ * Sounds are played on a separate thread. This avoids waiting for sound ids on methods that can 
+ * potentially lock main thread for considerable amount of time, especially when playing several
+ * sounds at the same time. The limitation of this approach is that methods that require a sound id
+ * are not supported.
+*/
 public class AsynchronousAndroidAudio extends DefaultAndroidAudio {
 
 	private final HandlerThread handlerThread;
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
@@ -1,0 +1,37 @@
+package com.badlogic.gdx.backends.android;
+
+import android.content.Context;
+import android.os.Handler;
+import android.os.HandlerThread;
+import com.badlogic.gdx.audio.Sound;
+
+public class AsynchronousAndroidAudio extends DefaultAndroidAudio {
+
+	private final HandlerThread handlerThread;
+	private final Handler handler;
+
+	public AsynchronousAndroidAudio (Context context, AndroidApplicationConfiguration config) {
+		super(context, config);
+		if (!config.disableAudio) {
+			handlerThread = new HandlerThread("libGDX Sound Management");
+			handlerThread.start();
+			handler = new Handler(handlerThread.getLooper());
+		} else {
+			handler = null;
+			handlerThread = null;
+		}
+	}
+
+	@Override
+	public void dispose () {
+		super.dispose();
+		if (handlerThread != null) {
+			handlerThread.quit();
+		}
+	}
+
+	@Override
+	protected Sound postProcessSound (AndroidSound sound) {
+		return new AsynchronousSound(sound, handler);
+	}
+}
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.os.Handler;
 import android.os.HandlerThread;
 import com.badlogic.gdx.audio.Sound;
+import com.badlogic.gdx.files.FileHandle;
 
 /** A performance oriented implementation of the {@link AndroidAudio} interface.
  * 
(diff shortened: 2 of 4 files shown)


MrStahlfelge (changes requested):
I don't think the new method `postProcessSound` is necessary.
You can override `newSound` in the new backend directly and achieve the same:

```
public Sound newSound (FileHandle file) {
    Sound newSound = super.newSound(file);
    return new AsynchronousSound(sound, handler);
}
```

I think this way it is more clear what's happening here, I find the term "postProcess" a bit misleading as it awakened hopes for some nice frequence filtering post processing effects immediately. :D

obigu:
> I don't think the new method `postProcessSound` is necessary.
> You can override `newSound` in the new backend directly and achieve the same:
> 
> ```
> public Sound newSound (FileHandle file) {
>     Sound newSound = super.newSound(file);
>     return new AsynchronousSound(sound, handler);
> }
> ```
> 
> I think this way it is more clear what's happening here, I find the term "postProcess" a bit misleading as it awakened hopes for some nice frequence filtering post processing effects immediately. :D

Thanks, definitely unnecessary, I originally created it to avoid a blind casting to `AndroidAudio` but `AsynchronousSound` can work with a `Sound`. Fixed now.

MrStahlfelge (approved):
Code is good, backwards compatibility guarantueed, and test on Android was successful

MrStahlfelge:
Please add information about this on the wiki page. As a price, your avatar will be added to the main repo page under "Contributors".
By the way, I think it is even possibile to return some self-generated sound ids and have a mapping to the real sound id as soon as it is available. But this can be done in the future by people needing it.

MichalLebeda:
amazing, this problem is really old, hope it works well

obigu:
> Please add information about this on the wiki page. As a price, your avatar will be added to the main repo page under "Contributors".
> By the way, I think it is even possibile to return some self-generated sound ids and have a mapping to the real sound id as soon as it is available. But this can be done in the future by people needing it.

Sure, I've added some info on the Wiki https://github.com/libgdx/libgdx/wiki/Audio

ezBeanie:
How to work with sound ids? There is no way to change the pitch of a looping sound without the sound id, atleast i can't manage to achieve that, using this asnych audio on Android.

obigu:
It's not currently possible, that's the limitation of this backend.

aleksdev:
I have the problem with this implementation on my device with Android 11. After a while all sounds disappear, but then sounds can return. If the application is paused during the "silent" state (screen is turned off or interstitial ads showed) I got ANR. If to play a lot of sounds simultaneously can get the "silent" state very soon.

"main" prio=5 tid=1 Waiting
  | group="main" sCount=1 dsCount=0 flags=1 obj=0x7248e6a8 self=0x776e622010
  | sysTid=6595 nice=-10 cgrp=default sched=0/0 handle=0x7894f464f8
  | state=S schedstat=( 0 0 0 ) utm=153 stm=38 core=1 HZ=100
  | stack=0x7fe68f4000-0x7fe68f6000 stackSize=8192KB
  | held mutexes=
  at java.lang.Object.wait(Native method)
  - waiting on <0x05eb4418> (a android.opengl.GLSurfaceView$GLThreadManager)
  at java.lang.Object.wait(Object.java:442)
  at java.lang.Object.wait(Object.java:568)
  at android.opengl.GLSurfaceView$GLThread.onPause(GLSurfaceView.java:1730)
  - locked <0x05eb4418> (a android.opengl.GLSurfaceView$GLThreadManager)
  at android.opengl.GLSurfaceView.onPause(GLSurfaceView.java:579)
  at com.badlogic.gdx.backends.android.AndroidGraphics.onPauseGLSurfaceView(SourceFile:2)
  at com.badlogic.gdx.backends.android.AndroidApplication.onPause(SourceFile:12)
  at mypackage.AndroidLauncher.onPause(SourceFile:1)
  at android.app.Activity.performPause(Activity.java:8174)
  at android.app.Instrumentation.callActivityOnPause(Instrumentation.java:1510)
  at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:4730)
  at android.app.ActivityThread.performPauseActivity(ActivityThread.java:4691)
  at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:4642)
  at android.app.servertransaction.PauseActivityItem.execute(PauseActivityItem.java:46)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
  at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
  at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
  at android.os.Handler.dispatchMessage(Handler.java:106)
  at android.os.Looper.loop(Looper.java:223)
  at android.app.ActivityThread.main(ActivityThread.java:7661)
  at java.lang.reflect.Method.invoke(Native method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:594)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

"libGDX Sound Management" prio=5 tid=56 Native
  | group="main" sCount=1 dsCount=0 flags=1 obj=0x14842140 self=0x776e656270
  | sysTid=21391 nice=0 cgrp=default sched=0/0 handle=0x755071dcc0
  | state=S schedstat=( 0 0 0 ) utm=24 stm=12 core=4 HZ=100
  | stack=0x755061a000-0x755061c000 stackSize=1043KB
  | held mutexes=
  native: #00 pc 00000000000a0108  /apex/com.android.runtime/lib64/bionic/libc.so (__epoll_pwait+8)
  native: #01 pc 0000000000019d88  /system/lib64/libutils.so (android::Looper::pollInner(int)+184)
  native: #02 pc 0000000000019c68  /system/lib64/libutils.so (android::Looper::pollOnce(int, int*, int*, void**)+112)
  native: #03 pc 0000000000113314  /system/lib64/libandroid_runtime.so (android::android_os_MessageQueue_nativePollOnce(_JNIEnv*, _jobject*, long, int)+44)
  at android.os.MessageQueue.nativePollOnce(Native method)
  at android.os.MessageQueue.next(MessageQueue.java:335)
  at android.os.Looper.loop(Looper.java:183)
  at android.os.HandlerThread.run(HandlerThread.java:67)

When playing a lot of sounds on default implementation I can get ANR just during gameplay. Also tried this library https://github.com/barsoosayque/libgdx-oboe, got sound distortion on my second device (Redmi Note 2, Android 9).

obigu:
> When playing a lot of sounds on default implementation I can get ANR just during gameplay. 

So your problem seems to affect both implementations `DefaultAndroidAudio` and `AsynchronousAndroidAudio` right?

Depending on what "a lot" refers to it will be impossible for hardware the harwdware to reproduce it and even for `SoundPool` to process it. I'd say that could cause an ANR on some devices (would have to test). 

One approach is to limit the minimum amount of time (in milliseconds) that can pass between playing the same sound once it has been played. This decreases the amount of concurrent sounds and should solve most issues.

If the proposed solution doesn't work or is not an option please share some reproducible code/project with the issue.

PR #6244, opened 2020-10-21
Updated IOSDevice list with last year devices

obigu:
Updated `IOSDeviceList` with last year devices.

I think we could consider removing unsupported devices (max iOS version < 8). The list keeps growing and I don't think it makes sense to have devices not supported by libGDX.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSDevice.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSDevice.java
@@ -30,6 +30,7 @@ public class IOSDevice {
 		this.ppi = ppi;
 	}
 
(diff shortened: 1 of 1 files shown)


MrStahlfelge:
To review this, I need to know from where these values can be obtained. What's the source? This information should be added in a comment to the file as well in my opinion.

obigu:
I got the info from this library https://github.com/lmirosevic/GBDeviceInfo but in order to follow previous style with the namings (Wifi, Cellular) and confirm I used https://gist.github.com/adamawolf/3048717. It should be good but it's not official, shall we add it anyway?

MrStahlfelge:
Better to have the information for the future than to not have it, so yes!

obigu:
Ok, added!

MrStahlfelge (approved):
No problem spotted

PR #6245, opened 2020-10-21
Fix wrong reported density on iOS

obigu:
Density not only was being calculated using `IOSDevice` information (which is risky in case there's a mistake) but it was using Android's 160 dpi value which reported wrong density values on iOS (326 ppi / 160dpi = 2,0375 when it should be 326ppi / 163dpi = 2).

We already obtain points per pixel programatically (API available since iOS 8), there's not reason to depend on `IOSDevice` information afaict.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -29,6 +29,7 @@ import org.robovm.apple.uikit.UIApplication;
 import org.robovm.apple.uikit.UIApplicationDelegateAdapter;
 import org.robovm.apple.uikit.UIApplicationLaunchOptions;
 import org.robovm.apple.uikit.UIDevice;
+import org.robovm.apple.uikit.UIUserInterfaceIdiom;
 import org.robovm.apple.uikit.UIInterfaceOrientation;
 import org.robovm.apple.uikit.UIPasteboard;
 import org.robovm.apple.uikit.UIScreen;
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -168,6 +168,17 @@ public class IOSApplication implements Application {
 		 return new DefaultIOSInput(this);
 	}
 
+	/** Returns device ppi using a best guess approach when device is unknown. Overwrite to customize strategy. */
+	protected int guessUnknownPpi() {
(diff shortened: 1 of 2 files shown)


Darkyenus:
What else is `IOSDevice` used for except for this? A quick github search suggests that this is the only usage and it would be nice if we could drop it altogether. It is a public API, so users may depend on it, but I am not sure why would they.

obigu:
@Darkyenus We use it to populate the ppi which can't be obtained programatically afaik.

Darkyenus:
Interesting, I overlooked that. In that case the question is: Should we provide PPI to users? Is it worth the hassle? I think that it isn't, because most applications should not care about the actual PPI, but only about the point/pixel ratio. However, this is not the place for this discussion.

obigu:
We can move the discussion elewhere but I think it's useful (as long as it's kept up to date). You may want to build UI elements that have the same size independently of the device the app runs in or game controls that trigger after a specific touch has dragged some distance in centimeters.

obigu:
I've added some changes to improve the guessed PPI in case the device is unknown (currently it was always returning 163). The method for this guess can be overwritten by apps to choose a different strategy.

MrStahlfelge (approved):
LGTM

davidgiga1993:
The iOS density is now always either 1 or 2, causing `Gdx.graphics.getDensity()` to return the same. This doesn't make any sense whatsoever. According to the documentation it should return `a scaling factor for the Density Independent Pixel unit, following the same conventions as android.util.DisplayMetrics#density`

Therefore the assignment of pixel per point to density is wrong and breaking the existing API (as it is the case for my app which relies on DPI for proper device independent scaling)

MrStahlfelge:
Did you take into account that the reported screen size is now different?
For me, this was broken before btw. I could not rely on iOS density information before.

davidgiga1993:
My app was working fine on all platforms, densities and scaling. Now it's completely broken on ios.
The density is wrong (it's either 1 or 2, nothing in between) and the reported resolution behavior is different from lwjgl on retina screens... so I need to implement another workaround on ios due to this change

obigu:
> The iOS density is now always either 1 or 2, causing `Gdx.graphics.getDensity()` to return the same. This doesn't make any sense whatsoever. According to the documentation it should return `a scaling factor for the Density Independent Pixel unit, following the same conventions as android.util.DisplayMetrics#density`
> 
> Therefore the assignment of pixel per point to density is wrong and breaking the existing API (as it is the case for my app which relies on DPI for proper device independent scaling)

@davidgiga1993 You are right, the `getDensity()` method is defined to follow Android dpi definitions. The other changes related to ppi calculations should be good though. What I miss in this case is a method that returns the number of physical pixels per logical pixel on the Graphics interface for backends that suppor hdpi. Will work on a PR. Thanks.

PR #6246, opened 2020-10-21
Cache FileType

PokeMMO:
* Android is relative to Context#getCacheDir()
  * See https://developer.android.com/reference/android/content/Context#getCacheDir()
* iOS is relative to <Application_Home>/Library/Caches 
  * See https://developer.apple.com/icloud/documentation/data-storage/index.html
* Desktop is relative to application's root directory + /cache

Desktop might be better to use a proper temporary folder for but a subfolder is much simpler. Finding a proper temporary folder in Windows/Linux/macOS which have a similar semi-persistent nature is probably more work than it is worth.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSFiles.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSFiles.java
@@ -27,6 +27,7 @@ public class IOSFiles implements Files {
 	static final String appDir = System.getenv("HOME");
 	static final String externalPath = appDir + "/Documents/";
 	static final String localPath = appDir + "/Library/local/";
+	static final String cachePath = appDir + "/Library/Caches/";
 	static final String internalPath = NSBundle.getMainBundle().getBundlePath();
 
 	public IOSFiles () {
@@ -65,6 +66,11 @@ public class IOSFiles implements Files {
 	}
 
 	@Override
+	public FileHandle cache (String path) {
+		return new IOSFileHandle(path, FileType.Cache);
+	}
+
+	@Override
 	public String getExternalStoragePath () {
 		return externalPath;
 	}
@@ -82,5 +88,10 @@ public class IOSFiles implements Files {
 	@Override
 	public boolean isLocalStorageAvailable () {
 		return true;
+	}
+
+	@Override
+	public String getCacheStoragePath () {
+		return cachePath;
 	}
 }
--- a/gdx/src/com/badlogic/gdx/Files.java
+++ b/gdx/src/com/badlogic/gdx/Files.java
@@ -45,7 +45,11 @@ public interface Files {
 		Absolute,
 
 		/** Path relative to the private files directory on Android and to the application's root directory on the desktop. */
-		Local;
+		Local,
+
+		/** Path to store files which can be regenerated. Path relative to Context#getCacheDir() on Android, <Application_Home>/Library/Caches on iOS,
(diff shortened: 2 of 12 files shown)


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSFiles.java
```diff
@@ -27,6 +27,7 @@
 	static final String appDir = System.getenv("HOME");
 	static final String externalPath = appDir + "/Documents/";
 	static final String localPath = appDir + "/Library/local/";
+	static final String cachePath = appDir + "/Library/Caches/";
```
MrStahlfelge, line 30: This should be retrieved with from the OS by calling somethink like
`NSFileManager.getDefaultManager().getURLsForDirectory(NSSearchPathDirectory.CachesDirectory, NSSearchPathDomainMask.AllDomainsMask).get(0).getPath();` (untested)
See https://www.swiftdevcenter.com/get-document-directory-temp-directory-and-cache-directory-path-ios-swift-5-1/

gdx/src/com/badlogic/gdx/Files.java
```diff
@@ -82,4 +89,7 @@
 
 	/** Returns true if the local storage is ready for file IO. */
```
MrStahlfelge, line 90: This JavaDoc could be changed to reflect that it also returns if cache path is available. (The current doc is misleading btw)

MrStahlfelge (changes requested):
I think the solution for desktop is discussable but everyone using this will be able to deal with it.
All in all a good enhancement in my opinion.

crykn:
I suppose this addition is mostly for mobile platforms, so I'm not really in a position to judge how often it is needed and whether it makes sense to include this in libGDX. However, some thoughts regarding the desktop implementation:
- Just using the application's root dir + `/cache` is not really satisfying
- Using a temporary folder (maybe via `System.getProperty("java.io.tmpdir")`) has the problem that (at least on Linux) this folder is emptied on reboot. And then what's the point in having a cache?
- Using some other (semi-)permanent location sounds like the best solution: on Windows we could use `%localappdata%`; on macOS `/Library/Caches` in the user's home dir; on Linux [this repo](https://github.com/harawata/appdirs/blob/master/src/main/java/net/harawata/appdirs/impl/UnixAppDirs.java#L49) suggests looking for the `XDG_CACHE_HOME` env var and then falling back to `/.cache` in the user's home dir (however, according to a quick google search some distros seem to be using `/var/cache` as well)

PokeMMO:
There's some issues in finding a persistent directory for linux/macos/windows due to this code being in the framework.
Simply using appdata or something similar would cause all libgdx games/apps to share a cache folder.

On a related note, this is exactly what happens with Preferences in libgdx, unless you change the config in Lwjgl(3)ApplicationConfiguration

https://github.com/libgdx/libgdx/blob/bc314368a18e0c916437f3dd7b4172714e6abf8b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java#L62-L63

MrStahlfelge:
If it is that controversial on desktop, maybe we should add a `isCacheStorageAvailable` method and don't implement the thing on desktop. After merging in #6358 every game could change the behaviour if it is needed anyway.

I still think this is a good addition for the mobile platforms (when my remark regarding iOS is processed) and is not really needed on desktops, so we should not block this because there's no perfect desktop solution.

PR #6248, opened 2020-10-24
Fixed pooled rectangle for clipping never being freed if clip was successful. 

piotr-j:


payne911:
Nice catch!

obigu:
Isn't it freed on `clipEnd()` on that case?

piotr-j:
Ah yes, you are right. I didnt notice that it is stored in there. Or looked below. A bit confusing design, but what can you do

PR #6250, opened 2020-10-25
Support specifying which keyboard type to use for getTextInput

MrStahlfelge:
This is a follow-up to #6235 and adds support for different keyboard types on iOS and Android for `getTextInput` method.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -228,17 +228,19 @@ public class DefaultAndroidInput implements AndroidInput {
 	}
 
 	@Override
-	public void getTextInput(final TextInputListener listener, final String title, final String text, final String hint, final OnscreenKeyboardType type) {
+	public void getTextInput(final TextInputListener listener, final String title, final String text, final String hint, final OnscreenKeyboardType keyboardType) {
 		handle.post(new Runnable() {
 			public void run () {
 				AlertDialog.Builder alert = new AlertDialog.Builder(context);
 				alert.setTitle(title);
 				final EditText input = new EditText(context);
-				input.setInputType(getAndroidInputType(type));
+				if (keyboardType != OnscreenKeyboardType.Default) {
+					input.setInputType(getAndroidInputType(keyboardType));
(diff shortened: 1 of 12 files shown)


MrStahlfelge:
When merged, we should add a line to CHANGES for this and #6235

obigu (commented):
I think we can implement Password type on iOS by setting `UITextField.secureTextEntry` boolean to true on that case. The rest LGTM.

MrStahlfelge:
> I think we can implement Password type on iOS by setting `UITextField.secureTextEntry` boolean to true on that case.

That is a good idea, this should be done for Android, too.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
```diff
@@ -221,15 +223,24 @@ public float getGyroscopeZ () {
...
+	}
+
+	@Override
+	public void getTextInput(final TextInputListener listener, final String title, final String text, final String hint, final OnscreenKeyboardType type) {
 		handle.post(new Runnable() {
 			public void run () {
 				AlertDialog.Builder alert = new AlertDialog.Builder(context);
 				alert.setTitle(title);
 				final EditText input = new EditText(context);
+				input.setInputType(getAndroidInputType(type));
```
PokeMMO, line 237: Since this uses EditText, would specifying `InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_FLAG_AUTO_CORRECT | InputType.TYPE_TEXT_FLAG_CAP_SENTENCES` to enable auto suggest and auto correct work?

MrStahlfelge: Thanks for the hint. Auto correction and suggestions were already present before, but the change removed it. I changed the behaviour to only overwrite the input type if something else than the default keyboard is needed.

PokeMMO (approved):
Not tested personally, but I looked over the code and everything looks good.

obigu (approved):
Tested on Android only but iOS looks good.

Not related to your changes (but maybe worth fixing), the `TextInputDialogTest` doesn't work as expected. In theory you should be able to open new input dialogs after the first one when touching the screen.

PR #6254, opened 2020-10-26
Adding logging interval parameter

yigitozdemir:
Logging interval addition to fps logger. Also if bound parameter has been given, logs on error level.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
+++ b/gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
@@ -39,11 +39,19 @@ public class FPSLogger {
 
 	/** Logs the current frames per second to the console. */
 	public void log () {
+		log(1);
+	}
+	
+	public void log(int interval) {
 		final long nanoTime = TimeUtils.nanoTime();
-		if (nanoTime - startTime > 1000000000) /* 1,000,000,000ns == one second */ {
+		if (nanoTime - startTime > (1000000000 * interval)) /* 1,000,000,000ns == one second */ {
 			final int fps = Gdx.graphics.getFramesPerSecond();
 			if (fps < bound) {
-				Gdx.app.log("FPSLogger", "fps: " + fps);
+				if(bound != Integer.MAX_VALUE) {
+					Gdx.app.error("FPSLogger", "fps: " + fps);
+				} else	{
+					Gdx.app.log("FPSLogger", "fps: " + fps);
+				}
 				startTime = nanoTime;
 			}
 		}


gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
```diff
@@ -39,11 +39,19 @@ public FPSLogger (int bound) {
 
 	/** Logs the current frames per second to the console. */
 	public void log () {
+		log(1);
+	}
+	
+	public void log(int interval) {
```
tommyettinger, line 45: This overload of log() should be documented, including the interval parameter. I also strongly think interval should be a double (not a float; these numbers are way outside precise float range), because it allows users to refresh frame rate more often than once per second.

gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
```diff
@@ -39,11 +39,19 @@ public FPSLogger (int bound) {
...
-		if (nanoTime - startTime > 1000000000) /* 1,000,000,000ns == one second */ {
+		if (nanoTime - startTime > (1000000000 * interval)) /* 1,000,000,000ns == one second */ {
 			final int fps = Gdx.graphics.getFramesPerSecond();
 			if (fps < bound) {
-				Gdx.app.log("FPSLogger", "fps: " + fps);
+				if(bound != Integer.MAX_VALUE) {
+					Gdx.app.error("FPSLogger", "fps: " + fps);
+				} else	{
+					Gdx.app.log("FPSLogger", "fps: " + fps);
+				}
```
tommyettinger, line 54: This whole section is strange and appears incorrect -- bound can be configured by the user, so why is a user-set bound writing to error while the default no-arg constructor makes a FPSLogger that writes to log? I think this should be reverted.

tommyettinger (changes requested):
I've requested some changes, but I'm not very positive on this PR right now. If @yigitozdemir could explain why some logging goes to error and other logging to log, that would help. And documentation is a must.

yigitozdemir:
Hello @tommyettinger,

I'll add documentation to the code. You mean Javadoc block, right? I am pretty new to contribution.

I converted log to error for bound values because as a user, if i am binding a value to there, i would definitely be checking an issue. That is why i converted it to error, not a plain log. If that does not make a sense for general users, i can revert this while creating documentation. 

Thanks.

crykn:
I agree with @tommyettinger: The changes regarding error/log don't make much sense (the class should be as straight-forward as possible) and the javadoc documentation needs improvement.

crykn:
I'm closing this for now. If you are willing to do the requested changes, @yigitozdemir, please feel free to reopen.

PR #6255, opened 2020-10-26
add missing res folder as source for Eclipse

mgsx-dev:
some missing stuff for Eclipse users after #5956 changes.


NathanSweet:
Sorry I'm late. I've never needed them, but it looks OK.

PR #6257, opened 2020-10-30
Added "contains" scaling mode.

Emperorlou:
Scaling.contain was missing from the scaling options. It basically functions like Scaling.fit if the image is too big to fit in the defined space, but will act like Scaling.none if it does fit. This means it will only scale the image down and never up. Useful for icons that are usually all the same size but also allowing you to use icons that are oversized without having to worry about the image spilling over into the rest of the UI.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Scaling.java
+++ b/gdx/src/com/badlogic/gdx/utils/Scaling.java
@@ -1,12 +1,12 @@
-/* ******************************************************************************
+/*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- *
+ * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * 
  *   http://www.apache.org/licenses/LICENSE-2.0
- *
+ * 
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


yuripourre:
@Emperorlou thank you for you contribution.

Before one can dig deeper into the code, can you fix the style? So it is easier to review?

Emperorlou:
To be honest I don't know how. I even tried editing the file directly in
GitHub and pasting in only the changed lines and it still did this
nonsense. I assume it's a line ending issue..

SimonIT:
I tried fixing them for you, but unfortunately you don't have _Allow edits and access to secrets by maintainers
_ enabled

MrStahlfelge:
It is the old crlf problem, it is fine for me. You can set GH to ignore whitespaces to see the real changes. 

I like the addition, and it is missing. The only thing that could be improved is naming and JavaDoc. I know the JavaDoc goes conform with the other ones, but I always have to read twice to understand.
I checked, Android names the same scaling approach "centerInside" with the following doc:
> Scale the image uniformly (maintain the image's aspect ratio) so that both dimensions (width and height) of the image will be equal to or less than the corresponding dimension of the view (minus padding). The image is then centered in the view.

Emperorlou:
Ah that's too bad about the naming convention because Android's variant
isn't as sussinct imo. Contain is used in CSS for the background-size
attribute, which also includes fit and cover iirc.

As for javadoc, I'm not sure where that lives.

MrStahlfelge (approved):
Tested this and works like a charme. My naming proposal `centerInside` is misleading if alignment is set to something different than center and this scaling is used, so we keep `contain`.
Our current Scaling.java file is CRLF formatted so this is fixed with this change, too.

PR #6258, opened 2020-10-31
Fix the TiledDrawable color.

AndEditor7:
Fixed by updating the packedColor in `SpiteBatch`.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
@@ -41,7 +41,7 @@ public class TiledDrawable extends TextureRegionDrawable {
 
 	public void draw (Batch batch, float x, float y, float width, float height) {
 		float oldColor = batch.getPackedColor();
-		batch.getColor().mul(color);
+		batch.setColor(batch.getColor().mul(color));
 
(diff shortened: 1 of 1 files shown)


crykn:
This PR doesn't really change anything, does it? `Color#mul(Color)` modifies the colour instance it is called on, so there isn't really a need to use `batch.setColor(Color)`.

AndEditor7:
Yes it does, and I have already tested. [This](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java#L207) won't update the `colorPacked` in SpriteBatch, so it can't update the color in `SpriteBatch`'s draw. Here's the last change by Nathan Sweet: https://github.com/libgdx/libgdx/commit/fd89994a6f4e2128accd68930e1c1d1742ebb220#diff-566f266554f566719acf83386b6e8c6b57dd2e43758ae6cd14fab97255963980

AndEditor7:
Here's the `setColor()` if you wondering: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java#L201

crykn (approved):

tommyettinger (approved):
Seems like a clear fix. This change only affects TiledDrawable, which wasn't using the color correctly anyway, so nothing should regress from this.

PR #6261, opened 2020-11-03
Revert 6118 "Cache Box2D Fixture filter data to avoid JNI overhead"

obigu:
Unfortunately this change has unintended consequences on Box2D making collision logic unreliable. 

I pushed a live app with libGDX 1.9.12 and have received several user reports about objects traversing walls in some cases.

I was able to reproduce incorrect collision resolution by changing the filter data of an existing object by calling `setFilterData` but there may be other scenarios. My guess is that this is caused because the filter data is not applied immediately when calling `setFilterData` as stated on this comment:
```
/** Set the contact filtering data. This will not update contacts until the next time step when either parent body is active and
	 * awake. This automatically calls Refilter. */
```
causing a mismatch between Box2D internal state and libGDX state.

**EDIT: The cause of the issue is probably because libGDX pools the `Fixture`s in `World` and the `reset()` doesn't set the filter back to default.**

**@Tom-Ski @NathanSweet This is a bad issue. I think we whould release a patch version as currently Box2D is broken.**

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
@@ -28,10 +28,4 @@ public class Filter {
 	/** Collision groups allow a certain group of objects to never collide (negative) or always collide (positive). Zero means no
 	 * collision group. Non-zero group filtering always wins against the mask bits. */
 	public short groupIndex = 0;
-
-	public void set(Filter filter) {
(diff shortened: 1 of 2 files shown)


obigu:
There is an alternative fix that I'm currently testing that is to set the `filter` to `null` on `Fixture` `reset` so that it doesn't get reused. Will keep the updates.

obigu:
I confirm the fix fixes the issue. We can discuss in a future release after the emnergency release if it's worth applying the fix considering the important performance gains (and if we should maybe go for an allocation approach as discussed in #6118 ).

PR #6262, opened 2020-11-04
 Cache Box2D Fixture filter data to avoid JNI overhead V2

obigu:
This is an attempt to fix the issues caused by #6118 described in #6261.

Instead of setting `filter` to `null` on `Fixture` `reset()` which would cause an allocation every time a `Fixture` is obtained (`reset()` is called after `obtain()` in this case), I've opted to go for a no memory allocation approach by using a boolean flag to indicate if `filter` data is dirty. The drawback of this approach is that `getFilterData()` now returns the `Filter` reference used by `Fixture` meaning that, if  it's changed without calling `setFilterData()`, it will cause inconsistent behaviour. This also means the change may break backwards compatibility.

The approach suggested by @NathanSweet on #6118 asking if it was possible to remove `jniGetFilterData()` alltogether would require some hacks because fixtures can be created using a `FixtureDef` directly (see `Body#createFixture (FixtureDef def)`) in which case the `Filter` instance would not have the information provided.

I've tested it on my games and haven't found any issue.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/Filter.java
@@ -28,4 +28,10 @@ public class Filter {
 	/** Collision groups allow a certain group of objects to never collide (negative) or always collide (positive). Zero means no
 	 * collision group. Non-zero group filtering always wins against the mask bits. */
 	public short groupIndex = 0;
+
+	public void set(Filter filter) {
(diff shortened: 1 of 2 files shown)


mgsx-dev (approved):
LGTM, we discussed about this issue yesterday on discord, i checked the code, i think it should be OK now.

obigu:
This is a good time to merge this fix now that we have a safe fallback release (1.9.13). I'm pretty confident it's good this time but before merging I'll wait a few days in case anybody else wants to review/test (maybe @MobiDevelop ?)

PR #6263, opened 2020-11-04
Fix getDensity() on iOS to be consistent with the rest of backends

obigu:
#6245 changes on `getDensity()` were not following `Graphics` API which specifies:

```
	/** This is a scaling factor for the Density Independent Pixel unit, following the same conventions as
	 * android.util.DisplayMetrics#density, where one DIP is one pixel on an approximately 160 dpi screen. Thus on a 160dpi screen
	 * this density value will be 1; on a 120 dpi screen it would be .75; etc.
	 *
```

As a separate PR we can add a new interface method to easily get the amount of physical pixels per logical pixel, maybe promoting the already present on iOS `getBackBufferScale()`.


MrStahlfelge (approved):
Yep, that was wrong.

PR #6269, opened 2020-11-14
java 8

kennethgaz:
update the java version

### DIFF
--- a/extensions/gdx-setup/build.gradle
+++ b/extensions/gdx-setup/build.gradle
@@ -28,3 +28,14 @@ jar {
 		attributes 'Main-Class': project.mainClassName
 	}
 }
+
+sourceSets {
+    main {
+        java {
+            srcDir 'src/'
+        }
+        resources {
+            srcDir 'res/' 
+        }
+    }
+}
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
@@ -18,7 +18,7 @@ android {
     }
     defaultConfig {
         applicationId "%PACKAGE%"
-        minSdkVersion 14
+        minSdkVersion 26
         targetSdkVersion %API_LEVEL%
         versionCode 1
         versionName "1.0"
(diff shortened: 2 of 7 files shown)


MrStahlfelge:
One step after the other, see #5839

kennethgaz:
Hello,
but I changed the java version only in gdx-setup, not in the project as a whole.
I did the tests and it's working perfectly.

MrStahlfelge:
Try on an Android 5 device, does it work?

kennethgaz:
But the play store no longer accepts apks with old versions.

SimonIT:
https://developer.android.com/distribute/best-practices/develop/target-sdk Here is only the target API level mentioned or am I overlooking something?

kennethgaz:
that's right :+1:

kennethgaz:
Well, I'm going to speak my opinion about that.

Developers who don't know libgdx have no idea what this library is capable of. The initial project is very simple and makes it look like it's not a game library. I believe that most developers stop using the library because they don't have a solid initial project and it is necessary to do several searches before finishing the first game (I can open a pull request with these improvements later).

And many of the developers are prejudiced against old versions of java (they usually don't know that it is possible to upgrade the version or don't know how to do that). So, I believe it would be a good idea to change the initial java version because it would bring more game developers using libgdx. But if it is necessary to wait, we will wait.

Translated by Google Translate, hahaha.

crykn:
Hey @kennethgaz!

> The initial project is very simple and makes it look like it's not a game library.  I believe that most developers stop using the library because they don't have a solid initial project [...]

This is not really relevant to this PR, but being "very simple" is kind of the point of the initial project. It is not meant as an example, but as a barebones structure to make setting up a project (gradle configs, folder structure, etc.) less tedious. 

> [...] and it is necessary to do several searches before finishing the first game

That very much depends on where you are looking. There is a step-by-step tutorial for a basic game (that is literally one page) on our [website](https://libgdx.com/dev/simple_game/) and this tutorial is also linked in the README and in the wiki.

> And many of the developers are prejudiced against old versions of java (they usually don't know that it is possible to upgrade the version or don't know how to do that). So, I believe it would be a good idea to change the initial java version because it would bring more game developers using libgdx. 

Phew, I kinda doubt that is true. If someone knows enough about Java to recognise that Java 7 is an old version, they should be capable of changing the version for themselves. And would they really opt to not use libGDX just because their first project is using an old java version by default?

Also, upping the default java version is only the first step: Android, iOS and GWT are all very delicate, so any changes need to be thoroughly tested.

PR #6271, opened 2020-11-14
fix proguard rules

kennethgaz:


SimonIT:
Can you please explain why you want these rules to be default. Did you test the controllers and box2d extension if they work because you are removing their rules.
Why do you set the sourceSets? It's inherited from the parent project

kennethgaz:
This change obfuscates only the core project. The external libraries remain intact (since libgdx is open source, I believe it is not a problem to expose the project's classes)

image

I was unable to run gdx-setup project on my machine without sourceSets (via intelliJ or gradle commands).

obigu:
Why would you add a rule not to apply optimization to anything except from your app's package? Proguard (or R8 now), not only obfuscates, it's also shrinks and optimizes code.

I'm closing this now for bad PR format, lack of description, file changes not related to PR topic and no apparent testing.

PR #6273, opened 2020-11-18
Upgrade Android SDK lib we compile against to 30

obigu:
Upgraded Android SDK we build against to 30. This will allow support for new features such as Haptics.

Having the android.jar on the repo uploaded by a contributor is not ideal, hopefully https://github.com/libgdx/libgdx/pull/5976 will fix this.

### DIFF
--- a/backends/gdx-backend-android/.classpath
+++ b/backends/gdx-backend-android/.classpath
@@ -3,7 +3,7 @@
 	<classpathentry combineaccessrules="false" exported="true" kind="src" path="/gdx"/>
 	<classpathentry excluding="**/.svn/*" kind="src" path="src"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry kind="lib" path="libs/android-9.0.jar"/>
+	<classpathentry kind="lib" path="libs/android-30.jar"/>
 	<classpathentry kind="lib" path="libs/activity-1.1.0.jar"/>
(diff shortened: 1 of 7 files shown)


MrStahlfelge:
Comparing with 7693da46147cde4d67, at least dependencies.gradle should be changed. DependencyBank is useful to change, too.

obigu:
> Comparing with [7693da4](https://github.com/libgdx/libgdx/commit/7693da46147cde4d67c8ee8c16cf40a7778bfe69), at least dependencies.gradle should be changed. DependencyBank is useful to change, too.

Wouldn't that set target SDK to 30 on new gdx setup projects? That may require some more testing. The motivation for this is to be able to compile API 29 code on libGDX project. IMO, upgrading the default target SDK to 30 on libGDX projects may be done separately.

MrStahlfelge (commented):
- [X] Checked the jar file against android sdk 30 - same
- [X] Android Tests are still compiling and running
- [ ] I can't test if Maven builds successfully

I still think it would be better to raise compileSdk for the tests to 30 as well, but that might not be necessary.

obigu:
Thanks for the tests. Ok, I agree we can update `dependencies.gradle` compile sdk to 30 for consistency sake between Maven/Ant and Gradle builds. Considering they're independent, Id rather set it on this other PR https://github.com/libgdx/libgdx/pull/6289 that already makes some changes on the names of the variables and other Gradle build related stuff, is that ok?

MrStahlfelge:
Maven and me, we have some kind of hate relationship, that's why I fell in love with Gradle so fast.

PR #6277, opened 2020-11-19
GWT use setCatchKey to prevent browser's default behaviour

MrStahlfelge:
Typically, every libGDX HTML project has to tweak around with index.html to prevent browser's default behaviour on certain key presses (F1, Alt, Space, Arrow Keys and stuff).

We already have `setCatchKey` to prevent system or background app behaviour on key presses, but this is only implemented on Android so far. This PR adds the ability to use the setCatchKey methods to prevent browser default behaviour on GWT.

Additionally, this changes `keyForCode` from private static to protected. Since we now can override the input processing class, it makes sense to have this method overridable as well.


SimonIT (approved):
Works perfectly. Tested with the Tab Key in GwtInputTest

PR #6278, opened 2020-11-19
GWT fix some key mappings

MrStahlfelge:
This fixes key mappings for

* Right shift
* Right ctrl
* Right alt
* Numpad star 
* Numpad divide

on GWT backend


tommyettinger (approved):
Wow, I wonder how long those `//FIXME` comments were in there. This looks solid, as long as nothing overrides keyForCode(), which would be an incredibly rare use-case.

SimonIT (approved):
Works like a charm

PR #6284, opened 2020-11-24
Simplifiications to Proguard config

obigu:
Removed some unnecessary Proguard rules:
- `dontwarn android.support` rule is default on Android `proguard-android.txt` platform file, no need to add it again (and with migration to AndroidX I don't think is relevant for us anyway).
- `AndroidControllers` rule is not needed after removing reflection code
- `AndroidInput` rule is not needed either (legacy reflection), actually that class is now an Interface, the old one was renamed to `DefaultAndroidInput`.

### DIFF
--- a/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidControllers.java
+++ b/extensions/gdx-controllers/gdx-controllers-android/src/com/badlogic/gdx/controllers/android/AndroidControllers.java
@@ -57,12 +57,7 @@ public class AndroidControllers implements LifecycleListener, ControllerManager,
 		
 		// use InputManager on Android +4.1 to receive (dis-)connect events
 		if(Gdx.app.getVersion() >= 16) {
-			try {
-				String className = "com.badlogic.gdx.controllers.android.ControllerLifeCycleListener";
-				Class.forName(className).getConstructor(AndroidControllers.class).newInstance(this);
-			} catch(Exception e) {
-				Gdx.app.log(TAG, "Couldn't register controller life-cycle listener");
-			}
+			new ControllerLifeCycleListener(this);
 		}
 	}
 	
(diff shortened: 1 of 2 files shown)


MrStahlfelge:
`AndroidControllers` is instantiated with `ClassReflection.forName`. I know that Proguard can identify some reflections like this, but did you make sure it does in this case?

obigu:
> `AndroidControllers` is instantiated with `ClassReflection.forName`. I know that Proguard can identify some reflections like this, but did you make sure it does in this case?

Good point, I see it's used on `gdx-controllers` (I've searched for `Class.forName` references only). As you say, `Class.forName` calls are handled automatically by ProGuard without the need of a rule but I'm not sure if ProGuard/R8 is able to infer it when it's passed to the utility class. Will have a look.

MrStahlfelge:
Yep, here is the code
https://github.com/libgdx/libgdx/blob/8d514ce74e46af80641dc70c2a37f2cfb800ed6b/extensions/gdx-controllers/gdx-controllers/src/com/badlogic/gdx/controllers/Controllers.java#L83

obigu:
Unfortunately R8 is not able to deal with this case and the rule is needed. I've reverted the change and added some comments.

There is a separate discussion to be had if we can improve extensions current dependency management not to have to use reflection to instantiate backend especific code. There is a circular dependency between extensions `core` modules and the extension backends which leads to this particular need.

MrStahlfelge (approved):
Looks good to me now!

PokeMMO:
I'm not sure this is the best place to bring this up, but isn't this proguard rule file woefully insufficient to enable obfuscation for many parts of libgdx?

I see nothing in here which would allow Scene2d/particles/serialization based loaders to work with the obfuscated results.

obigu:
It should be enough, yes, Proguard is a bit smart and handles basic reflection, these rules haven been used for a long time and have passed the test of time. Scene2D and particles (2d at least) work without issues. I don't use libGDX serialization but for any serialization framework you use, at least you'd need to add a rule for the classes you serialize but that's up to each project. That doesn't mean there can't be an uncommon functionality that requires a missing rule that we haven't noticed yet.

PR #6286, opened 2020-11-25
Get IntMap and LongMap to correctly set removed values to null

### LINKED ISSUE #6285: `IntMap.containsValue()` returns true for values that have been removed.
`IntMap.containsValue()` returns true for stale values. This creates problems with the Gwt backend on touch screens when trying to determine if a touch identifier has been seen before as it thinks that there are multiple touches on the screen. This causes the pinch handlers to be called rather than the pan handler.

#### Reproduction steps/code

* Add an item to an `IntMap<>`.
* Remove the item.
* Call `containsValue()` with the value.
* `containsValue()` returns true.

_Please provide the version(s) affected._

1.9.12

#### Test Case
The following test fails:

```java
class IntMapTest {
    @Test
    void test_contains_value_returns_false_after_removal() {
        IntMap< Integer > map = new IntMap<>();
        Integer value = 123;
        map.put( 1, value );
        map.remove( 1 );

        assertFalse( map.containsValue( value, false ) );
    }
}
```

- [ ] Android
- [ ] iOS
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

tommyettinger:
ObjectMap already does this, but somehow IntMap and LongMap got hit by a wave of changes to the primitive-valued maps even though IntMap and LongMap have references for values. This is a four-line change that makes remove() act like it already does in ObjectMap. Without this, removed values won't get GCed until the whole IntMap or LongMap does, and containsValue() gives incorrect results. This fixes #6285 .


yuripourre (approved):
Seems good to me.

PR #6289, opened 2020-11-25
Maintenance on versions and build configurations

obigu:
This is a bit of a mixed bag of maintenance changes and improvements to config files aiming at going in the direction of having all versions defined in fewer places, keeping consistency between `tests` and new libGDX project defaults and updating some outdated versions.

Highlights:
- Upgraded libGDX Gradle to `6.7.1` which allows us to increase Android Gradle plugin.
- Android Gradle Plugin version set to `4.0.2`  both on new projects and libGDX `tests` Android module (`4.1.X` is currently not supported by some IDEs).
- Set `minifyEnabled` to `true` for `release` builds by default for new projects.

I've tested both `tests` and new project and have found no issues.


gradle/dependencies.gradle
```diff
@@ -31,10 +31,12 @@ versions.jlayer = "1.0.1-gdx"
...
 versions.appleExtension = "1.4"
-versions.androidPlugin = "3.1.0"
+versions.androidPlugin = "3.5.4"
 versions.multiDex = "2.0.1"
-versions.androidSdk = 29
-versions.androidBuildTools = "28.0.3"
+versions.androidCompileSdk = 30
+versions.androidTargetSdk = 29
+versions.androidMinSdk = 14
+versions.androidBuildTools = "29.0.3"
```
MrStahlfelge, line 39: I think it would be better to have build tools and compile sdk on the same version, but it works this way without warnings.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
```
MrStahlfelge, line 6: Is there something that is holding us back from using 6.7? It is the version we use in freshly generated game projects, and for me it looks like it works without problems on the main repo, too.

MrStahlfelge (approved):
All in all good changes and the project is working for me with the changes. I have just two remarks,.

gradle/dependencies.gradle
MobiDevelop: Interestingly, it hasn't been required to specify a build tools version in quite some time. The tools will automatically choose the minimum required for the plugin. One less thing to have to keep track of.

obigu: As @MobiDevelop says it's not required anymore to explicitly set it because it automatically picks the minimum supported required by the Android gradle plugin (in the case of `3.5.3` the minimum build tools is `28.0.3` (https://developer.android.com/studio/releases/gradle-plugin)). My understanding is that if we want to use a different one like in this case, as long as we use a version higher than the minimum , it should be ok independently of the `compileSDK`. I'm ok removing it alltogether or updating it to a different version, I see the advantage of not having to deal with another version (and let it uise 28.0.3) but there may be a better argument for using the latest.

gradle/wrapper/gradle-wrapper.properties
obigu: Just playing safe, I feel confident upgrading to 5.4.1 because it's in theory fully backwards compatible with 5.4 and the basic tests I ran worked. I agree it would be ideal upgrading to 6.7 at some point but the potential of breaking something is higher there and we would require more tests. Do you want to fo all in?

MrStahlfelge: Yes, I use 6.7 for quite some time in my projects without problems and we don't do the full build with it anyway.

obigu: I've upgraded Gradle to `6.7.1` (the Gradle docs recommend to always use the latest patch version) and haven't found any issue. This allows us to upgrade the Android Gradle plugin, I've chosen 4.0.2 instead of 4.1 due to not being supported on IDEA 2020.2. I'll run some more tests on iOS and Android, it'd be great if somebody could test GWT stuff.

MrStahlfelge (approved):
Tested changes to Gradle 6.7.1 and Android build tools 4.0.2. Works for me with Android and GWT.

SimonIT (approved):
Everything's working also from my side

obigu:
I'm not happy that projects don't work out of the box after importing them and require defining an ndk installation directory, especially considering that this will be fixed on `4.1.X` (https://issuetracker.google.com/issues/144111441). We can't upgrade to 4.1.X because it's still not supported by IDEA. 

My suggestion is, for the time being, we set the Android Gradle plugin to 3.5.4 which is the latests not to require defining the ndkVersion and only once 4.1.X is more widely supported we upgrade. I will roll back the android plugin version upgrade and merge if everybody is ok with it.

MrStahlfelge:
Strangely, 4.1 works without defining an ndk version for me. But I am fine with 3.5.4 as well.

obigu:
I guess it may work without having to define it explicitly on new projects if you have an environment variable like `ANDROID_NDK_HOME` or your IDE defaults generate a `local.properties` than define.  The other condition is that you have the requested NDK version already installed in your system.

PR #6290, opened 2020-11-26
Add hasContents to Clipboard interface, to reduce clipboard notifications on iOS 14

casid:
Apple has added a new banner alert to iOS 14 that lets users know if an app is pasting from the clipboard and therefore able to read the clipboard’s contents ([More information](https://9to5mac.com/2020/06/24/ios-14-clipboard-notifications/))

I've added a `boolean hasContents()` method to the `Clipboard` interface and adjusted the iOS and Android implementations, so that the actual clipboard content is not read by the app when calling this method.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -21,6 +21,7 @@ import java.io.File;
 import com.badlogic.gdx.ApplicationLogger;
 import com.badlogic.gdx.backends.iosrobovm.objectal.OALIOSAudio;
 import org.robovm.apple.coregraphics.CGRect;
+import org.robovm.apple.foundation.Foundation;
 import org.robovm.apple.foundation.NSMutableDictionary;
 import org.robovm.apple.foundation.NSObject;
 import org.robovm.apple.foundation.NSProcessInfo;
@@ -435,7 +436,12 @@ public class IOSApplication implements Application {
 
 			@Override
 			public boolean hasContents () {
-				return UIPasteboard.getGeneralPasteboard().hasStrings();
+				if (Foundation.getMajorSystemVersion() >= 10) {
+					return UIPasteboard.getGeneralPasteboard().hasStrings();
+				}
+
+				String contents = getContents();
+				return contents != null && !contents.isEmpty();
 			}
 
 			@Override
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -434,6 +434,11 @@ public class IOSApplication implements Application {
 			}
 
 			@Override
+			public boolean hasContents () {
+				return UIPasteboard.getGeneralPasteboard().hasStrings();
+			}
+
+			@Override
 			public String getContents () {
 				return UIPasteboard.getGeneralPasteboard().getString();
 			}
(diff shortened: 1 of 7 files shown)


casid:
Thanks @MrStahlfelge for labeling!

Let me know if I should do any adjustments. I tried to stick to the existing formatting of the altered files, but might have missed something (it's my first PR to libgdx).

MrStahlfelge:
Code looked good, we just have to test before merging. You could make this a bit easier (and demonstrate that you tested yourself) by adding a call to the new method in `ClipboardTest`

casid:
@MrStahlfelge will do add some automatied tests. So far, I only tested manually on iOS 14.

MrStahlfelge:
The libgdx tests aren't automated ;)

casid:
Ah I think I found the TestStarter classes.

casid:
I added a button to `ClipboardTest`, that checks if there's something on the clipboard.

Handy when testing on windows: `cmd /c echo.|clip` :-D

MrStahlfelge (approved):
Tested on Android and iOS, works well.

casid:
Wow that was quick! Thanks for taking the time to look into and test it.

obigu:
I understand the motivation for this is avoiding iOS 14 notifications. Would it make sense to make this check on the `getContents()` iOS implementation itself instead of adding a separate method for all backends?

Something like
```
void getContents() {
  if hasContents()
    return getContents()
  else
    return null (or empty, whichever we define)
}
```

casid:
@obigu that only helps if the user has no content on the clipboard.

For example, I use a custom text field that displays a 'Paste' button when it gains focus and the clipboard is not empty.

```java
if (Gdx.app.getClipboard().hasContents()) {
    // Show paste button
}
```

If the user has something on the clipboard, this displays the paste button without showing a notification.

If I use this instead:
```java
if (Gdx.app.getClipboard().getContents() != null) {
    // Show paste button
}
```

iOS will show a notification like 'This app just read your clipboard' as soon as the text field gains focus.

I think the new method isn't that bad, since it shows the clear intent that we're not interested in reading the user's content. I've adjusted the Android implementation as well, so if they follow Apple libgdx would already be prepared :-)

obigu:
I understand, but you would also avoid the notification by modifying the `getContents` implementation to do the check on iOS. I agree the method is not bad, just considering whether it is needed on the interface.

casid:
@obigu no, it would only avoid the notification in case the user has nothing on the clipboard.

Steps to reproduce with the `getContents()` method adjusted as you suggested:

1. User hast something on the clipboard
2. I call `getContents() != null` to check if I need to display a paste button
3. `getContents()` calls `hasContents()` internally, returns `true`
4. We read the content from the clipboard
5. iOS detects that this app just read the user's clipboard and displays a warning to the user

iOS is not aware that we don't do anything with the returned content, it just noticed we read that content and could potentially do whatever with it. Therefore the privacy warning. This can only be avoided by using their SDK method that returns a boolean instead of the entire content.

obigu:
Ok, ok, true, thanks, I was just thinking in the case the clipboard was empty.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
```diff
@@ -433,6 +433,11 @@ public void setContents (String content) {
 				UIPasteboard.getGeneralPasteboard().setString(content);
 			}
 
+			@Override
+			public boolean hasContents () {
+				return UIPasteboard.getGeneralPasteboard().hasStrings();
```
obigu, line 438: This seems iOS10+, I think we need an OS check here (we support up to minOS 9)

obigu (changes requested):

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
casid: Oh @obigu, you are right, thank you for spotting this!

I will push a fix.

obigu: Thanks!

Issue #5993, opened 2020-04-09
Graphical artifacts on OnePlus 6 phone


funkyfourier:
#### Issue details
On OnePlus 6 phones, and possibly other models, there are graphical artifacts similar to low quality JPEG compression in images. It is especially obvious if there are large single colored areas. It is not equally obvious with all colors, and I have found that the color `#2c2c2c` reveals the issue well. 

Interestingly, the artifact does not show up in screenshots, so I have tried to photograph it using another phone. It is barely visible in the photos, but on the screen of the OnePlus phone they are clearly obvious. I have marked where the issue occurs, you can (barely) see lines across the screen.

https://i.imgur.com/E8rccLa.png
https://i.imgur.com/ONtum3B.png 

In the test project there are lines, but in my app they are more like cubes like in badly compressed JPEG images.

The phone runs on Android 10. I have also tried on the Pixel emulator with Android 10 and Google services and have not been able to reproduce it on that. On a Sony XZ1, Samsung S5, Nexus 5, Honor 7 as well as the desktop version it runs fine.

Now I guess if none of you have a OnePlus 6 it will be hard to debug this. If someone would give me an idea of where to begin to look that would be greatly appreciated.

#### Reproduction steps/code
1. Generate a blank LibGDX project
2. Replace the code of the main ApplicationAdapter with this:
```
public class MyGdxGame extends ApplicationAdapter {
	SpriteBatch batch;
	Color bgColor;
	
	@Override
	public void create () {
		batch = new SpriteBatch();
		bgColor = Color.valueOf("2c2c2c");
	}

	@Override
	public void render () {
		Gdx.gl.glClearColor(bgColor.r, bgColor.g, bgColor.b, bgColor.a);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
	}
	
	@Override
	public void dispose () {
		batch.dispose();
	}
}
```
3. Run on a OnePlus 6 
4. If the issue does not occur, try exiting and entering the app a couple of times. Maybe also kill and restart it.

#### Version of LibGDX and/or relevant dependencies
gdxVersion = '1.9.10'

#### Please select the affected platforms
- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

funkyfourier:
[...]

PR #6291, opened 2020-11-26
Use platform default color bit depth on Android and iOS

obigu:
After a post from user "tann" on Discord #android channel we've realised some inconsistencies on the libGDX default color format (RGB565) and the platform defaults on Android (RGB888) and iOS (RGBA8888).

### Android
It is currently set to RGB565 because that was the default of `GLSurfaceView` before but, currently (not sure since when), it's RGB888 (https://developer.android.com/reference/android/opengl/GLSurfaceView.html).

> By default GLSurfaceView will create a PixelFormat.RGB_888 format surface.

I've kept RGB565 as fallback in case the provided color format configuration is not supported by the device. This is the safest approach as, in the rare case a device did not support RGB888 it would fallback to previous default but it's not necessarily the best one. It depends if there are actually devices in the wild (with min SDK 14+) that don't support RGB888, if there aren't, we should default to RGB888. Unfortunately I haven't been able to find that information.

Regarding the changes a couple of comments:
- The change in the `BufferFormat` on `AndroidGraphics` is aesthetic. A new instance with the appropriate configurations is assigned to `bufferFormat`, I'm not sure if instantitating it on declaration is needed at all but I've kept it just in case.
- Something similar happens on `GLSurfaceView20` `init()` method changes. Even if we set the color format there (either RGB888 or RGBA8888) it doesn't matter because we set it manually afterwards in every case later on (check for example `AndroidGraphics.createGLSurfaceView()`):

```
GLSurfaceView20 view = new GLSurfaceView20(application.getContext(), resolutionStrategy, config.useGL30 ? 3 : 2);
if (configChooser != null)
	view.setEGLConfigChooser(configChooser);
else
	view.setEGLConfigChooser(config.r, config.g, config.b, config.a, config.depth, config.stencil);
```

**Known issues and Tests**
1. Using RGB565 has some known issues on especific devices such as reported here https://github.com/libgdx/libgdx/issues/5993. I've also personally observed issues when running an OpenGL gradient on a Nexus 7 2013 which displays bad banding and artifacts that get fixed using RGB888.
2. Current default RGB565 causes dithering to be applied, especially visible when zooming in on any libGDX game with default settings.
3. @cypherdare says he's been using RGBA8888 on his apps for 5 years without any issue.

### iOS
On iOS the current default is RGBA8888 (https://developer.apple.com/documentation/glkit/glkview/1615587-drawablecolorformat).

iOS is weird because the color format seems to be ignored when running on a device (it always looks like RGB888) but it is correctly applied on Simulator. I've confirmed this and has been reported on the forums before: https://badlogicgames.com/forum/viewtopic.php?f=11&t=29129. I don't know the reason for this and the change will not have any impact on apps running on device but at least we improve consistency between Simulator and Device.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
@@ -29,7 +29,7 @@ import com.badlogic.gdx.backends.android.surfaceview.ResolutionStrategy;
  * @author mzechner */
 public class AndroidApplicationConfiguration {
 	/** number of bits per color channel **/
-	public int r = 5, g = 6, b = 5, a = 0;
+	public int r = 8, g = 8, b = 8, a = 0;
 
 	/** number of bits for depth and stencil buffer **/
 	public int depth = 16, stencil = 0;
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -90,7 +90,7 @@ public class AndroidGraphics implements Graphics, Renderer {
 	private float density = 1;
 
 	protected final AndroidApplicationConfiguration config;
-	private BufferFormat bufferFormat = new BufferFormat(5, 6, 5, 0, 16, 0, 0, false);
+	private BufferFormat bufferFormat = new BufferFormat(8, 8, 8, 0, 16, 0, 0, false);
 	private boolean isContinuous = true;
 
 	public AndroidGraphics (AndroidApplicationBase application, AndroidApplicationConfiguration config,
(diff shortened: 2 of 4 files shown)


PokeMMO:
Setting to 888 on min sdk 21 android has no errors in production. I think at this point merging and seeing if issues are reported later is fine.

obigu:
I've set defaults on an iOS app and Android apps and haven't seen errors in production either. I think it's safe to merge.

MobiDevelop:
Yep, seems fine to me.

tarslangil:
@obigu @MobiDevelop I'm having an issue on iOS when i run the project on iPhone.Left is desktop screenshot and right is iOS screenshot.On iOS everything looks like desaturated.I switched to 1.9.13-SNAPSHOT version but it didn't changed anything.

Screen_Shot_2020-12-20_at_14 27 41

obigu:
@tarslangil I think your issue is unrelated to this PR (the change only affects new created apps using GDX Setup tool). You can try setting `config.colorFormat = GLKViewDrawableColorFormat.RGBA8888;` on your `IOSApplicationConfiguration`. If that doesn't help, I recommned asking on the official Discord channel for support (https://libgdx.com/community/discord/) or create a new issue if you think it's a bug and can provide an executable example (https://github.com/libgdx/libgdx/wiki/Getting-help).

PR #6299, opened 2020-12-04
Android: Fix wrong constants for Escape, End, Insert, F1 to F12

MrStahlfelge:
Escape, End, Insert, F1 to F12 were defined with the wrong keycodes for Android. Pressing F1 was registered as pressing Escape, others were not recognized at all.

It is necessary to change the core constants as there is no code mapping on the Android backend. As long as games only use the defined constants - which they should -, this will not break any projects.


NathanSweet:
Changing the constants is a little scary, as it's possible someone relies on the values rather than the names, eg maybe in serialized data. I'm OK with the PR as is and changing the constants, but if we wanted to make the fix more safely we'd add a mapping to the Android backend.

MrStahlfelge:
I agree and I am not fan of changing the constants myself. However, we historically have no mapping on Android and it is a big change to introduce one. 
As long as everyone uses the constants in the code, there should be no problem caused by a change. If someone does not use the constants, this should be changed.
So the only problem I see is that changing the constants will cause incompatiblity for user-saved keybindings. If someone used the concerned keys, he will to reconfigure. I use user-saved keybindings in my games but doubt anyone used the concerned keys in my games - and if, I am okay with the users to reconfigure after the change -, so if no one vetos I suggest to merge this in and see what happens after the release. If there is no reaction at all, we were going good, if there are some issues raised by developers, we know we should introduce a mapping system on the Android backend for future changes.

SimonIT (approved):
From the functional site: Works perfectly

MrStahlfelge:
No veto -> merge

yairm210:
Somewhat belatedly, but this is causing issues for me.
I'm on Windows 10, and when debugging, saw that catching the Escape was no longer working after upgrading LibGDX.
Diving deeper, I saw a mismatch between the keycode given (111) and the one defined (131).
Finding this PR, I feared this was part of a wider problem.
As a simple test, I created an InputListener that does nothing but print the keycodes (pardon my Kotlin, but I'm sure it's basically the same in Java):

```kotlin
val listener = object : InputListener() {
            override fun keyDown(event: InputEvent?, keycode: Int): Boolean {
                println(keycode)
                return false
            }
        }
stage.addListener(listener)
```

As I suspected, F1, F2, End etc. all gave the pre-PR keycodes.
Basically, although this fixed the keycodes for Android, it ruined them for Windows :(

It's possible I'm not using the keycodes correctly (I basically check `keycode == Input.Keys.ESCAPE`) in which case I'd be glad to hear of the correct usage I'm not sure what the right solution is... I can think of 2 possible ones

A. Having the keycodes as part of Gdx.input.Keys rather than statically referencing Input.Keys (so you would have a default Keys, but each specific Input override - AndroidInput for example - could override the defaults) - this is probably (?) the more correct way if you were starting from scratch, but currently would mean breaking the existing API in a big way

B. adding ESCAPE_ANDROID, F1_ANDROID or something - super ugly, but wouldn't affect existing projects.

I want to use this situation to give a huge shoutout to you guys, you are AMAZING! :D

MrStahlfelge:
are you sure you updated your platform dependant dependencies? Check if your lwjgl2 or lwjgl3 backend is on the same version than the core dependency.

yairm210:
You were right, this was a result of caching :(

Apparently, even after running `gradle clean` and `gradle cleanBuildCache`, Android Studio itself saves its own build cache. After I ran the cleans **and restarted Android Studio**, the old code resumed working as usual. I'll put 'restart Android Studio after every gradle change' in my internal checklist.

Thank you again for all your work, I cannot put into words how amazing you guys are :)

Issue #5389, opened 2018-10-03
F13-F24 keys missing


EtK2000:
These keys are usually not on keyboards, but are often used in custom bindings for multiple input devices or mice with more than 5 buttons (which Windows doesn't natively support).

Here are the backend keycodes I've found:
• Android: https://source.android.com/devices/input/keyboard-devices
• GLFW (LWJGL3): http://www.glfw.org/docs/latest/group__keys.html
• LWJGL 2: https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/input/Keyboard.java

I'm sure other keys are missing on LWJGL2 (see https://github.com/libgdx/libgdx/issues/4739 for example).

FyiurAmron:
related: https://github.com/libgdx/libgdx/issues/2524, https://github.com/libgdx/libgdx/pull/2592, https://github.com/libgdx/libgdx/pull/5207 ...

EtK2000:
I might work on a pull request adding in all of these missing keys (including the related issues).
I've never used GWT or any of the iOS backends so those will require me to research.

Are there any other keys that are missing that I should add in?

Also, what should I do about keys that only exist on some backends?

FyiurAmron:
@EtK2000 to know what's still missing, you'd have to research the related issues. Try to begin your work based on https://github.com/libgdx/libgdx/pull/5207 (and probably https://github.com/libgdx/libgdx/pull/2592) instead of doing a entirely new PR.

As to the keys that exist only on some backends - I guess the idea is to implement all of them in the root Input class, and simply leave them unmapped in the backends that miss them.

EtK2000:
None of those pulls seem to actually map the keycodes on other backends, no wonder they weren't accepted.

FyiurAmron:
@EtK2000 firstly, you're wrong; my PR didn't add them to backends on purpose (so that everyone can agree on the exact `Input` changes before implementing them in backends), and yousiftouma actually did add them at least to LWJGL backend.

Secondly, you miss the main issue. Until you have the keys defined in `Input`, you can't use them in backends. It's completely mandatory to do the preliminary work in the `Input` class beforehand. The PRs weren't accepted for multitude of reasons (well, actually, mine got accepted, but got reverted due to a dumb typo), not implementing changes in backends weren't one of them.

EtK2000:
Ok, I'm confused as to why you wouldn't implement them in the backends if you're already adding them.
[...]

PR #6300, opened 2020-12-05
Add support for hardware keys F13 to F24

MrStahlfelge:
This adds keybindings for F13 to F24 as requested in #5389 

I couldn't test this because I have not such a keyboard. I derived the necessary keybindings from the given constants in the backends. For GWT and Android, there were no constants defined in code. [This page](https://source.android.com/devices/input/keyboard-devices) lists them for Android, [this one](https://keycode.info/) for JS. Lwjgl2 only supports F13 to F18, I did not forget the other ones. :)


crykn (approved):
I don't have a suitable keyboard either, but the new constants are correct. 

Just for future reference, when using Mozialla's Gecko on Web, the F13+ keys may be detected as different keys on Mac OS and Linux, see [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). But considering the low adoption of keyboards with those keys, this shouldn't really be an issue.

MrStahlfelge:
Interesting link you found! But the information is not quite correct or outdated on some combinations, for example I know for sure that the Print key worked for me on Chrome and Gecko on Linux. So hopefully this is the same for F13 up, too.

PR #6149, opened 2020-08-30
AnimationController action end up at duration time, fixes #6145

mgsx-dev:
Issue #6145 describe the edge case where action (queued animation) was not removed when just finished.
The extra update is not necessary with zero delta time in this case.
@asu-talk thank you a lot for your test case, i rewrote it as a JUnit.

Issue #6145, opened 2020-08-28
3D main looped animation is not queued to play after action animation is complete


asu-talk:
Relates to a corner case with timing - action animation should be complete at the exact time to be able to observe this issue.

#### Reproduction steps/code
Test case reproducing the issue:

```java
package com.badlogic.gdx.tests.g3d.utils;

import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.graphics.g3d.Model;
import com.badlogic.gdx.graphics.g3d.ModelInstance;
import com.badlogic.gdx.graphics.g3d.model.Animation;
import com.badlogic.gdx.graphics.g3d.utils.AnimationController;
import com.badlogic.gdx.tests.utils.GdxTest;
import com.badlogic.gdx.utils.StringBuilder;

public class AnimationIssueTest extends GdxTest {

    private final StringBuilder sb = new StringBuilder(1024);

    @Override
    public void create() {
        Model model = new Model();

        Animation mainAnimation = new Animation();
        mainAnimation.id = "main";
        mainAnimation.duration = 1f;

        Animation actionAnimation = new Animation();
        actionAnimation.id = "action";
        actionAnimation.duration = 0.2f;

        ModelInstance modelInstance = new ModelInstance(model);
        modelInstance.animations.add(mainAnimation);
        modelInstance.animations.add(actionAnimation);

        Gdx.app.log("DBG", "-----------------------------------------");
        Gdx.app.log("DBG", "Correct behaviour test case");
        Gdx.app.log("DBG", "-----------------------------------------");
        testCase(modelInstance, actionAnimation.duration + 0.001f);

        Gdx.app.log("DBG", "-----------------------------------------");
        Gdx.app.log("DBG", "Test case with issue");
        Gdx.app.log("DBG", "-----------------------------------------");
        testCase(modelInstance, actionAnimation.duration);

        model.dispose();
    }

    private void testCase(ModelInstance modelInstance, float actionUpdateTime){
        AnimationController animationController = new AnimationController(modelInstance);
        Gdx.app.log("DBG", "Play main animation");

        animationController.setAnimation("main", -1);
        animationController.update(animationController.current.duration);
        dumpAnimation(animationController.current);
        animationController.update(0.01f);
        dumpAnimation(animationController.current);

        Gdx.app.log("DBG", "Start action animation");
        animationController.action("action", 1, 1f, null, 0f);
        dumpAnimation(animationController.current);

        animationController.update(actionUpdateTime);
[...]

PR #6303, opened 2020-12-07
fix AnimationController and AnimationDesc update

mgsx-dev:
This PR both fix a regression introduced in 1.9.12 and fix some edge cases.

fix #6149 introduced a regression, there was a misunderstanding (my fault) with AnimationDesc#update return value : it was returning 0 if animation still playing which conflicts with a remaining time of 0 returned. That was actually the origin of the first issue #6145.

Now the value -1 is returned in case of animation still playing. It's a breaking change but this class is rarely used directly or subclassed by user code, so i think it's fine.

I also rewrote AnimationDesc#update a bit in a cleaner way in order to handle some edge cases (junit tests provided)


mgsx-dev:
regression was found by @JackRainy see https://github.com/libgdx/libgdx/pull/6149#issuecomment-737469424 i checked Animation3DTest : attack animation is playing correctly now.

MobiDevelop (approved):

PR #6304, opened 2020-12-07
Media Keys for gwt

SimonIT:
Media key mapping for gwt.
I also tried to map lwjgl and lwjgl3, but both of them return 0 or -1 for each key.
Robovm has already some of the keys mapped. I couldn't find enum entries for the missing ones.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
@@ -1110,13 +1110,17 @@ public class DefaultGwtInput implements GwtInput {
 		case KEY_SCROLL_LOCK:
 			return Keys.SCROLL_LOCK;
 		case KEY_AUDIO_VOLUME_DOWN:
+		case KEY_AUDIO_VOLUME_DOWN_FIREFOX:
 			return Keys.VOLUME_DOWN;
 		case KEY_AUDIO_VOLUME_UP:
+		case KEY_AUDIO_VOLUME_UP_FIREFOX:
 			return Keys.VOLUME_UP;
 		case KEY_MEDIA_TRACK_NEXT:
 			return Keys.MEDIA_NEXT;
 		case KEY_MEDIA_TRACK_PREVIOUS:
 			return Keys.MEDIA_PREVIOUS;
+		case KEY_MEDIA_STOP:
+			return Keys.MEDIA_STOP;
 		case KEY_MEDIA_PLAY_PAUSE:
 			return Keys.MEDIA_PLAY_PAUSE;
 		case KeyCodes.KEY_PRINT_SCREEN:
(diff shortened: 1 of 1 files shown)


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```diff
@@ -1223,6 +1233,11 @@ protected int keyForCode (int keyCode, int location) {
 	private static final int KEY_F24 = 135;
 	private static final int KEY_NUM_LOCK = 144;
 	private static final int KEY_SCROLL_LOCK = 145;
+	private static final int KEY_AUDIO_VOLUME_DOWN = 174;
```
MrStahlfelge, line 1236: https://keycode.info/ states this is 182 on Firefox, 183 for Volume up

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```diff
@@ -1223,6 +1233,11 @@ protected int keyForCode (int keyCode, int location) {
...
+	private static final int KEY_AUDIO_VOLUME_UP = 175;
+	private static final int KEY_MEDIA_TRACK_NEXT = 176;
+	private static final int KEY_MEDIA_TRACK_PREVIOUS = 177;
+	private static final int KEY_MEDIA_PLAY_PAUSE = 179;
```
MrStahlfelge, line 1240: It came to my mind that we are totally missing STOP, just as a remark.

MrStahlfelge (commented):
Unfortunately, I don't have a fancy keyboard with the asked keys so I could not test. The code looks good and will not break anything, so it is safe to merge. However, it might be Firefox has other keys for vol up/down, so would be good for someone to test this.

As remarked, we should probably add KEY_MEDIA_STOP for completeness sake.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
SimonIT: Good find, will add it

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
SimonIT: That's a key that isn't on my keyboard, but I can add it

crykn: My keyboard has one and it is reported as `178`. The rest of the PR looks good to me!

MrStahlfelge (approved):
Green light from me for the code, this is safe to merge as it does not break anything. But take the approval with a grain of salt as I could not test it with a hw keyboard, maybe @crykn can do a real test.

crykn (approved):
Works with Chrome. With Firefox, the pause, stop, previous and next keys do not work, but that is Firefox' fault, as the JS events just report 0 as keycode. 

Another key that could be added is AUDIO_VOLUME_MUTE (181 in Firefox; 173 in other browsers), but that one would have to be added for the other backends as well.

MrStahlfelge:
We can do this in another PR, this one is good as it is.

PR #5763, opened 2019-08-18
Add GLES 3 support for GWT via WebGL2.

intrigus:
Currently still wip, but I'd appreciate some initial feedback :)

PR #6308, opened 2020-12-09
fix/improve WebGL 1.0 backend

mgsx-dev:
These changes was inspired by #5763. I prefer to split changes from original PR. More will come later.

Current GwtGL20 missing some buffer flips and ByteBuffer support for some methods.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java
@@ -497,6 +497,7 @@ public class GwtGL20 implements GL20 {
 			int id = this.buffers.add(buffer);
 			buffers.put(id);
 		}
+		buffers.flip();
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


tommyettinger:
Huh, I hadn't noticed anything wrong with GWT's GL20 implementation. What does this change? Performance?

mgsx-dev:
@tommyettinger nothing about performances, it fixes functionality:
* calling buffer.get(index) doesn't matter if buffer was flipped or not. That's how libgdx code access the result so that's why there are no noticeable issue.
* calling buffer.get() several times, buffer should be flipped in this case. User code may access the result  like this. Since other backends does flip the buffer, it wouldn't work for GWT backend.

tommyettinger:
So there's a difference in behavior between other platforms and GWT? I am shocked to my very core 😱 . If this bit of extra code doesn't harm framerate on GWT (or affect it at all), then I see no problems merging.

MrStahlfelge (commented):
Can't say I understand what's going on here completely, but I deeply trust the contributor and clicking through around 20 tests didn't bring up any problems.

PR #6317, opened 2020-12-17
Initial. AlphaTestGroupStrategy and associated test.

jeffcragg:
AlphaTestGroupStrategy. DecalAlphaTest shows intersecting overlapping translucent decals without artifacts with stipple/screen door transparency. The test is based on the CameraGroupStrategy one switched to perspective camera.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
@@ -1,0 +1,151 @@
+package com.badlogic.gdx.graphics.g3d.decals;
+
+import com.badlogic.gdx.*;
+import com.badlogic.gdx.graphics.*;
+import com.badlogic.gdx.graphics.glutils.*;
+import com.badlogic.gdx.utils.*;
+
+/**
+ * <p>Single Group strategy (all Decals same) using Z-buffer to render using screen door transparency to avoid having
+ * to presort the decals. Materials are still sorted.</p>
+ * <p>Avoids artifacts when the Decals are sufficiently proximate
+ * and orientated such that sorting fails, and also from transparent decals overlapping
+ * in plane (z-collisions) or by orientation.
+ * </p>
+ */
+public class AlphaTestGroupStrategy implements GroupStrategy, Disposable {
+    private static final int GROUP_ALPHA_TEST = 0;
+
+    final Pool<Array<Decal>>                     arrayPool      = new Pool<Array<Decal>>(16) {
+        @Override
+        protected Array<Decal> newObject() {
+            return new Array();
+        }
+    };
+    final Array<Array<Decal>>                    usedArrays     = new Array<Array<Decal>>();
+    final ObjectMap<DecalMaterial, Array<Decal>> materialGroups = new ObjectMap<DecalMaterial, Array<Decal>>();
+
+    Camera        camera;
+    ShaderProgram shader;
+
+    public AlphaTestGroupStrategy(final Camera camera) {
+        this.camera = camera;
+        createDefaultShader();
+    }
+
+    public Camera getCamera() {
+        return camera;
+    }
+
+    public void setCamera(Camera camera) {
+        this.camera = camera;
+    }
+
+    @Override
+    public int decideGroup(Decal decal) {
+        return GROUP_ALPHA_TEST;
+    }
+
+    @Override
+    public void beforeGroup(int group,
+                            Array<Decal> contents) {
+        for (int i = 0, n = contents.size; i < n; i++) {
+            Decal        decal         = contents.get(i);
+            Array<Decal> materialGroup = materialGroups.get(decal.getMaterial());
+            if (materialGroup == null) {
+                materialGroup = arrayPool.obtain();
+                materialGroup.clear();
+                usedArrays.add(materialGroup);
+                materialGroups.put(decal.getMaterial(),
+                                   materialGroup);
+            }
+            materialGroup.add(decal);
+        }
+
+        contents.clear();
+        for (Array<Decal> materialGroup : materialGroups.values()) {
+            contents.addAll(materialGroup);
+        }
+
+        materialGroups.clear();
+        arrayPool.freeAll(usedArrays);
+        usedArrays.clear();
+    }
+
+    @Override
+    public void afterGroup(int group) {
+    }
+
+    @Override
+    public void beforeGroups() {
+
+        Gdx.gl.glEnable(GL20.GL_DEPTH_TEST);
+        shader.begin();
+        shader.setUniformMatrix("u_projectionViewMatrix",
+                                camera.combined);
+        shader.setUniformi("u_texture",
+                           0);
+    }
+
+    @Override
+    public void afterGroups() {
+        shader.end();
+        Gdx.gl.glDisable(GL20.GL_DEPTH_TEST);
+    }
+
+    private void createDefaultShader() {
+        String vertexShader = "attribute vec4 " + ShaderProgram.POSITION_ATTRIBUTE + ";\n" //
+                              + "attribute vec4 " + ShaderProgram.COLOR_ATTRIBUTE + ";\n" //
+                              + "attribute vec2 " + ShaderProgram.TEXCOORD_ATTRIBUTE + "0;\n" //
+                              + "uniform mat4 u_projectionViewMatrix;\n" //
+                              + "varying vec4 v_color;\n" //
+                              + "varying vec2 v_texCoords;\n" //
+                              + "\n" //
+                              + "void main()\n" //
+                              + "{\n" //
+                              + "   v_color = " + ShaderProgram.COLOR_ATTRIBUTE + ";\n" //
+                              + "   v_texCoords = " + ShaderProgram.TEXCOORD_ATTRIBUTE + "0;\n" //
+                              + "   gl_Position =  u_projectionViewMatrix * " + ShaderProgram.POSITION_ATTRIBUTE + ";"
+                              + "\n" //
+                              + "}\n";
+        String fragmentShader = "#ifdef GL_ES\n" //
+                                + "precision mediump float;\n" //
+                                + "#endif\n" //
+                                + "varying vec4 v_color;\n" //
+                                + "varying vec2 v_texCoords;\n" //
+                                + "uniform sampler2D u_texture;\n" //
+                                + "float[] bayerMatrix ={1.0f / 17.0f,  9.0f / 17.0f,  3.0f / 17.0f, 11.0f / 17.0f," //
+                                + "13.0f / 17.0f,  5.0f / 17.0f, 15.0f / 17.0f,  7.0f / 17.0f,"//
+                                + "4.0f / 17.0f, 12.0f / 17.0f,  2.0f / 17.0f, 10.0f / 17.0f, " //
+                                + "16.0f / 17.0f,  8.0f / 17.0f, 14.0f /17.0f,  6.0f / 17" + ".0f};\n" //
+                                + "\n" //
+                                + "void main()\n"//
+                                + "{\n" //
+                                + "int xPos = int(gl_FragCoord.x) % 4; \n" //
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,151 @@
+package com.badlogic.gdx.graphics.g3d.decals;
+
+import com.badlogic.gdx.*;
+import com.badlogic.gdx.graphics.*;
+import com.badlogic.gdx.graphics.glutils.*;
+import com.badlogic.gdx.utils.*;
+
+/**
+ * <p>Single Group strategy (all Decals same) using Z-buffer to render using screen door transparency to avoid having
+ * to presort the decals. Materials are still sorted.</p>
```
mgsx-dev, line 10: presort decals is not an issue, the reason of this class is to avoid overlap artifacts in some situation.

tests/gdx-tests/src/com/badlogic/gdx/tests/DecalAlphaTest.java
```diff
@@ -0,0 +1,135 @@
...
+import com.badlogic.gdx.graphics.*;
+import com.badlogic.gdx.graphics.g2d.*;
+import com.badlogic.gdx.graphics.g3d.decals.*;
+import com.badlogic.gdx.math.*;
+import com.badlogic.gdx.tests.utils.*;
+import com.badlogic.gdx.utils.viewport.*;
+
+import java.util.*;
+
+public class DecalAlphaTest extends GdxTest {
```
mgsx-dev, line 29: This test is a bit complicated, a load test is not really what we need. Maybe a better approach would be to switch from/to different strategy at runtime and see how much FPS we can get (vsync can be disabled manually by testers when needed). Also adding/removing decals manually (via keyboard keys) would be better.

Before performance measurement, what we really need is to test that the strategy works (eg. the stipple effect, etc)

mgsx-dev (changes requested):

mgsx-dev (changes requested):
also you forgot to include a class to GWT xml (that's why the build fails), and also need to add an entry to the CHANGES file

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,151 @@
...
+                                + "float[] bayerMatrix ={1.0f / 17.0f,  9.0f / 17.0f,  3.0f / 17.0f, 11.0f / 17.0f," //
+                                + "13.0f / 17.0f,  5.0f / 17.0f, 15.0f / 17.0f,  7.0f / 17.0f,"//
+                                + "4.0f / 17.0f, 12.0f / 17.0f,  2.0f / 17.0f, 10.0f / 17.0f, " //
+                                + "16.0f / 17.0f,  8.0f / 17.0f, 14.0f /17.0f,  6.0f / 17" + ".0f};\n" //
+                                + "\n" //
+                                + "void main()\n"//
+                                + "{\n" //
+                                + "int xPos = int(gl_FragCoord.x) % 4; \n" //
+                                + "int yPos = int(gl_FragCoord.y) % 4; \n" //
+                                + "float stipple= bayerMatrix[ xPos * (yPos*4) ]; \n" //
```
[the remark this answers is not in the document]
tommyettinger: That does look incorrect, and it also won't run on GWT (can't use variable indices into arrays). It should be documented that this is GWT-incompatible.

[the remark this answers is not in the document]
mgsx-dev: Good catch! indeed, variable array index requires GLSL 130 (GLES 3.0+ / WebGL 2.0). There are possible workarounds either using a for loop to find the entry or using a texture lookup.
I was balanced to include this stipple feature from the beginning mainly because it's not really a concern of group strategy. Since implementation would bring more complexity or limitation, i'm in favor to fully remove this feature from the PR.
It could be interesting to have it but i would separate it from Decals stuff (and do it in another PR), user code could plug it in this class easily by overriding this method anyway.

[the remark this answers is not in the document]
jeffcragg: From Tettinger's post the bayer stuff is no longer going to be present anyway. I have been looking at this https://bartwronski.com/2016/10/30/dithering-part-three-real-world-2d-quantization-dithering/. I think stipple should be included though, because it is a part of the group strategy idea, that within a strategy some groups are translucent, and others opaque etc but if you remove stippling then the group strategy has no support for translucency at all when it could easily be a setting in the constructor. And i think that probably people would prefer CameraGroupStrategy or AlphaTestStrategy working entirely without needing extra stuff for the second choice, because its easy to just replace one with the other to see which is preferable.

gdx/res/com/badlogic/gdx.gwt.xml
```diff
@@ -147,6 +147,7 @@
 		
 	<!-- graphics/g3d/decals -->
 		<include name="graphics/g3d/decals/CameraGroupStrategy.java"/>
+		<include name="graphics/g3d/decals/AlphaTestGroupStrategy.java"/>
```
mgsx-dev, line 150: please use alphabetical order (A before C)

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,151 @@
+package com.badlogic.gdx.graphics.g3d.decals;
```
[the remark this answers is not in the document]
mgsx-dev: it looks like there is an issue with the formatted code (some spaces everywhere and some space padded statements)

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+			  + "varying vec2 v_texCoords;\n" //
+			  + "\n" //
+			  + "void main()\n" //
+			  + "{\n" //
+			  + "   v_color = " + ShaderProgram.COLOR_ATTRIBUTE + ";\n" //
+			  + "   v_texCoords = " + ShaderProgram.TEXCOORD_ATTRIBUTE + "0;\n" //
+			  + "   gl_Position =  u_projectionViewMatrix * " + ShaderProgram.POSITION_ATTRIBUTE + ";" + "\n" //
+			  + "}\n";
+
+		  String dfragmentShader = "#ifdef GL_ES\n" //
```
mgsx-dev, line 164: dead code

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+	 }
+
+	 /**
+	  * Renders translucency with screen door effect/stippling.
+	  *
+	  * @param camera
+	  * @param bayerFilter Specifies the size of the dithering filter used to discard pixels
+	  *                    of translucent textures.
+	  */
+	 public AlphaTestGroupStrategy (final Camera camera, final BayerFilter bayerFilter) {
```
mgsx-dev, line 87: better use an integer (keep it simple)

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+	 Camera camera;
+	 ShaderProgram shader;
+
+	 /**
+	  * Renders translucency with screen door effect/stippling.
+	  *
+	  * @param camera
+	  */
+	 public AlphaTestGroupStrategy (final Camera camera) {
+		  this(camera, BayerFilter.MODERATE);
```
mgsx-dev, line 77: should be disabled by default

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+			  + "precision mediump float;\n" //
+			  + "#endif \n" //
+			  + "varying vec4 v_color;\n" //
+			  + "varying vec2 v_texCoords;\n" //
+			  + "uniform sampler2D u_texture;\n" //
+			  + "\n" //
+			  + "#define MAX_LEVEL 4\n" //
+			  + "float GetBayer() {\n" //
+			  + "  float finalBayer = 0.0;\n" //
+			  + "  for(int i = 1 - "+bayerFilter.size +"; i<= 0; i++) {\n" //
```
mgsx-dev, line 213: better have an optimized version when stipple is disabled

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+			  + "    float bayerSize = exp2(float(i));   // negative exponent makes an inverse\n" //
+			  + "    vec2 c = mod(floor(gl_FragCoord.xy * bayerSize), 2.0);\n" //
+			  + "    float bayer = 2. * c.x - 4. * c.x * c.y + 3. * c.y;  // borrows factors of the summed geometric series soon to"
```
mgsx-dev, line 216: comments in generated shader code is not necessary, you could move it to java comments instead

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+		  //Tetting Bayer shader
+		  String fragmentShader = "#ifdef GL_ES\n" //
+			  + "precision mediump float;\n" //
+			  + "#endif \n" //
+			  + "varying vec4 v_color;\n" //
+			  + "varying vec2 v_texCoords;\n" //
+			  + "uniform sampler2D u_texture;\n" //
+			  + "\n" //
+			  + "#define MAX_LEVEL 4\n" //
+			  + "float GetBayer() {\n" //
```
mgsx-dev, line 211: why calculating for every fragment when we could use a lookup table (like in your original PR)?

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
```diff
@@ -0,0 +1,251 @@
...
+		  String vertexShader = "attribute vec4 " + ShaderProgram.POSITION_ATTRIBUTE + ";\n" //
+			  + "attribute vec4 " + ShaderProgram.COLOR_ATTRIBUTE + ";\n" //
+			  + "attribute vec2 " + ShaderProgram.TEXCOORD_ATTRIBUTE + "0;\n" //
+			  + "uniform mat4 u_projectionViewMatrix;\n" //
+			  + "varying vec4 v_color;\n" //
+			  + "varying vec2 v_texCoords;\n" //
+			  + "\n" //
+			  + "void main()\n" //
+			  + "{\n" //
+			  + "   v_color = " + ShaderProgram.COLOR_ATTRIBUTE + ";\n" //
```
mgsx-dev, line 159: missing `v_color.a = v_color.a * (255.0/254.0);`, see CameraGroupStrategy

gdx/src/com/badlogic/gdx/graphics/g3d/decals/AlphaTestGroupStrategy.java
jeffcragg: this one works on webgl and the lookup one can't unfortunately. variable access to arrays not allowed and for four iterations wtih this you get the equivalent of 16x16 dither block, although a single pass looks ok imo

PR #6319, opened 2020-12-18
Update intersectRayBoundsFast to handle degenerated boxes

### LINKED ISSUE #5128: IntersectRayBoundsFast does not give correct results when testing against bounding boxes with zero height
Using the `intersectRayBoundsFast` method with a boundingbox that is zero in one dimension results in false negatives.

#### Issue details
Consider the example below: Creating a bounding box from (3,3,0) to (4,4,0) and casting a ray from the origin (0,0,0) in the direction (1,1,0) should result in `true` when invoking `Intersector.intersectRayBoundsFast`. However, it returns `false`.

A workaround is to give the box "volume", as the bounding box from (3,3,-1) to (4,4,1) is intersected correctly.

`Intersector.intersectRayBounds` works as desired, even with the zero-coordinates.

#### Reproduction steps/code
```
import com.badlogic.gdx.math.Intersector;
import com.badlogic.gdx.math.Vector3;
import com.badlogic.gdx.math.collision.BoundingBox;
import com.badlogic.gdx.math.collision.Ray;

public class TestClass {
        Vector3 bottomLeft = new Vector3(3f, 3f, 0f);
        Vector3 topRight = new Vector3(4f, 4f, 0f);
        //Bounding box from (3,3) to (4,4)
        BoundingBox boundingBox = new BoundingBox(bottomLeft, topRight);
        //Ray from (0,0), direction (1,1)
        Ray ray = new Ray(Vector3.Zero, new Vector3(1f, 1f, 0f));
        System.out.println(Intersector.intersectRayBoundsFast(ray, boundingBox)); //prints "false", should print true
        System.out.println(Intersector.intersectRayBounds(ray, boundingBox, new Vector3())); //prints "true", should print true
    }
}
```

1.9.8

- [X] Windows
- Possibly others

I'll be happy to provide more clarification if needed - thank you for your time!

yuripourre:
The fix here is to set a minimal dimension when the value is zero.
It adds a few more operations (because of Math.max) but also removes three multiplications.

Closes #5128

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -621,25 +621,29 @@ public final class Intersector {
 		final float divY = 1f / ray.direction.y;
 		final float divZ = 1f / ray.direction.z;
 
-		float minx = ((center.x - dimensions.x * 0.5f) - ray.origin.x) * divX;
-		float maxx = ((center.x + dimensions.x * 0.5f) - ray.origin.x) * divX;
+		final float halfDimX = Math.max(dimensions.x, MathUtils.FLOAT_ROUNDING_ERROR) * 0.5f;
+		final float halfDimY = Math.max(dimensions.y, MathUtils.FLOAT_ROUNDING_ERROR) * 0.5f;
+		final float halfDimZ = Math.max(dimensions.z, MathUtils.FLOAT_ROUNDING_ERROR) * 0.5f;
+
+		float minx = ((center.x - halfDimX) - ray.origin.x) * divX;
+		float maxx = ((center.x + halfDimX) - ray.origin.x) * divX;
 		if (minx > maxx) {
 			final float t = minx;
 			minx = maxx;
 			maxx = t;
 		}
 
-		float miny = ((center.y - dimensions.y * 0.5f) - ray.origin.y) * divY;
-		float maxy = ((center.y + dimensions.y * 0.5f) - ray.origin.y) * divY;
+		float miny = ((center.y - halfDimY) - ray.origin.y) * divY;
+		float maxy = ((center.y + halfDimY) - ray.origin.y) * divY;
 		if (miny > maxy) {
 			final float t = miny;
 			miny = maxy;
 			maxy = t;
 		}
 
-		float minz = ((center.z - dimensions.z * 0.5f) - ray.origin.z) * divZ;
-		float maxz = ((center.z + dimensions.z * 0.5f) - ray.origin.z) * divZ;
-		if (minz > maxz) {
+		float minz = ((center.z - halfDimZ) - ray.origin.z) * divZ;
+		float maxz = ((center.z + halfDimZ) - ray.origin.z) * divZ;
+		if (minz >= maxz) {
 			final float t = minz;
 			minz = maxz;
 			maxz = t;
--- a/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
+++ b/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
@@ -3,6 +3,8 @@ package com.badlogic.gdx.math;
 
 import com.badlogic.gdx.math.Intersector.SplitTriangle;
 
+import com.badlogic.gdx.math.collision.BoundingBox;
+import com.badlogic.gdx.math.collision.Ray;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
@@ -184,5 +186,16 @@ public class IntersectorTest {
 		assertEquals(-102.63903f, intersection.x, 0.1f);
 		assertEquals(-57.7337f, intersection.y, 0.1f);
 		assertEquals(100, intersection.z, 0.1f);
+	}
+
+	@Test
+	public void testIntersectRayAABBDegenerated() {
+		Vector3 bottomLeft = new Vector3(3f, 3f, 0f);
+		Vector3 topRight = new Vector3(4f, 4f, 0f);
+		//Bounding box from (3,3) to (4,4)
+		BoundingBox boundingBox = new BoundingBox(bottomLeft, topRight);
+		//Ray from (0,0), direction (1,1)
+		Ray ray = new Ray(Vector3.Zero, new Vector3(1f, 1f, 0f).nor());
+		assertTrue(Intersector.intersectRayBoundsFast(ray, boundingBox));
 	}
 }


mgsx-dev:
I'm not sure it's a good idea to rely on operations with infinity floats, it might not work with all backends/platforms.
I run the same test with intersectRayOrientedBoundsFast by passing identity matrix and it works.
I'm more in favor to use the same algorithm as intersectRayOrientedBoundsFast (without matrix transformation) instead in order to have a more robust solution.
What do you think?

yuripourre:
@mgsx-dev I think you are right, I am not happy fully satisfied with the solution I proposed.
This is the algorithm proposed by Christer Ericson, what do you think?
```
	/** Intersects a {@link Ray} and a {@link BoundingBox}, returning the intersection point in intersection. This intersection is
	 * defined as the point on the ray closest to the origin which is within the specified bounds.
	 * 
	 * <p>
	 * The returned intersection (if any) is guaranteed to be within the bounds of the bounding box, but it can occasionally
	 * diverge slightly from ray, due to small floating-point errors.
	 * </p>
	 * 
	 * <p>
	 * If the origin of the ray is inside the box, this method returns true and the intersection point is set to the origin of the
	 * ray, accordingly to the definition above.
	 * </p>
	 * @param intersection The intersection point (optional)
	 * @return Whether an intersection is present. */
	public static boolean intersectRayBounds (Ray ray, BoundingBox box, Vector3 intersection) {
		float tmin = 0;
		float tmax = Float.MAX_VALUE;

		Vector3 p = ray.origin;
		Vector3 d = ray.direction;

		// Calculate x
		if (Math.abs(d.x) < MathUtils.FLOAT_ROUNDING_ERROR) {
			if (p.x < box.min.x || p.x > box.max.x) {
				return false;
			}
		} else {
			float ood = 1f / d.x;
			float t1 = (box.min.x - p.x) * ood;
			float t2 = (box.max.x - p.x) * ood;

			if (t1 > t2) {
				float temp = t1;
				t1 = t2;
				t2 = temp;
			}

			if (t1 > tmin) {
				tmin = t1;
			}
			if (t2 > tmax) {
				tmax = t2;
			}
			if (tmin > tmax) {
				return false;
			}
		}

		// Calculate y
		if (Math.abs(d.y) < MathUtils.FLOAT_ROUNDING_ERROR) {
			if (p.y < box.min.y || p.y > box.max.y) {
				return false;
			}
		} else {
			float ood = 1f / d.y;
			float t1 = (box.min.y - p.y) * ood;
			float t2 = (box.max.y - p.y) * ood;

			if (t1 > t2) {
				float temp = t1;
				t1 = t2;
				t2 = temp;
			}

			if (t1 > tmin) {
				tmin = t1;
			}
			if (t2 > tmax) {
				tmax = t2;
			}
			if (tmin > tmax) {
				return false;
			}
		}

		// Calculate z
		if (Math.abs(d.z) < MathUtils.FLOAT_ROUNDING_ERROR) {
			if (p.z < box.min.z || p.z > box.max.z) {
				return false;
			}
		} else {
			float ood = 1f / d.z;
			float t1 = (box.min.z - p.z) * ood;
			float t2 = (box.max.z - p.z) * ood;

			if (t1 > t2) {
				float temp = t1;
				t1 = t2;
				t2 = temp;
			}

			if (t1 > tmin) {
				tmin = t1;
			}
			if (t2 > tmax) {
				tmax = t2;
			}
			if (tmin > tmax) {
				return false;
			}
		}

		intersection.set(d).scl(tmin).add(p);
		return true;
	}
```

mgsx-dev:
it looks similar to [intersectRayOrientedBoundsFast](https://github.com/libgdx/libgdx/blob/f63c2aff1b5f60e6dd09e279af1f57c2ee3d21f5/gdx/src/com/badlogic/gdx/math/Intersector.java#L659) :-)

yuripourre:
@mgsx-dev I am very of surprised, running a simple benchmark, the `intersectRayBoundsFast` is always worse than `intersectRayBounds`

yuripourre:
@mgsx-dev nvm, I was wrong. I created a more realistic benchmark and the algorithm I posted is around 50% slower than the current `intersectRayBoundsFast`. Do you think worth replace it?

mgsx-dev:
i think it would be safer to code something similar to intersectRayOrientedBoundsFast which is close to the code you posted i guess. A fast algo doesn't worth anything if it's buggy.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -609,46 +609,111 @@ public static boolean intersectRayBounds (Ray ray, BoundingBox box, Vector3 inte
...
-		float miny = ((center.y - dimensions.y * 0.5f) - ray.origin.y) * divY;
-		float maxy = ((center.y + dimensions.y * 0.5f) - ray.origin.y) * divY;
-		if (miny > maxy) {
-			final float t = miny;
-			miny = maxy;
-			maxy = t;
+			if (t1 > tmin) {
+				tmin = t1;
+			}
+			if (t2 > tmax) {
```
mgsx-dev, line 654: how t2 could be greater than tmax which is initialized to max float value?
i think there is a typo for the 3 axis.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -609,46 +609,111 @@ public static boolean intersectRayBounds (Ray ray, BoundingBox box, Vector3 inte
...
-		}
+		// Half dimensions
+		tmp.set(dimensions).scl(0.5f);
+
+		Vector3 min = tmp1.set(center).sub(tmp);
+		Vector3 max = tmp2.set(center).add(tmp);
+		return intersectRayBoundsFastMinMax(ray, min, max);
+	}
+
+	static public boolean intersectRayBoundsFastMinMax (Ray ray, Vector3 min, Vector3 max) {
```
mgsx-dev, line 628: missing documentation

mgsx-dev (changes requested):

gdx/src/com/badlogic/gdx/math/Intersector.java
yuripourre: Whoa, nice catch! The book has a typo.

gdx/src/com/badlogic/gdx/math/Intersector.java
yuripourre: Oops, it supposed to be private.

PR #6321, opened 2020-12-19
added shader compilation check for ImmediateModeRenderer20

mgsx-dev:
It might sounds like a breaking change but i can't see a case where user code check compilation and fallback to something else so i think it's fine to merge it as is.


crykn:
Is an `IllegalArgumentException` the right type of exception in this case? Because I think the compilation only fails if there are some weird platform specific issues -> Maybe a GdxRuntimeException would fit better. But anyway, looks good to me.

MobiDevelop:
Yeah, I'd agree - IllegalArgumentException implies that the user supplied something invalid, so if that's not really the case here, it'd be better to use a different exception.

crykn:
SpriteBatch seems to be throwing an `IllegalArgumentException` as well, so I think it should be changed there, too.

mgsx-dev:
I simply copied what we have at other places (eg. SpriteBatch, CameraGroupStrategy) to be consistent but i agree with you, that was stupid :D. I'm not sure it worth changing exception type at other places since it'd be a not so valuable breaking change. Anyway it's beyond the scope of this PR.

crykn (approved):

tommyettinger (approved):
Looks absolutely fine. An extra check for correctness is nice.

PR #6322, opened 2020-12-19
Add faster ray aabb intersection

yuripourre:
The proposed method is borrowed from the book "Real-Time Collision Detection By Christer Ericson".

This algorithm is around 20% faster than the current algorithm.


gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -498,112 +498,105 @@ public static boolean intersectRaySphere (Ray ray, Vector3 center, float radius,
...
+			if (t2 > tmax) {
+				tmax = t2;
+			}
+			if (tmin > tmax) {
+				return false;
 			}
 		}
-		return hit;
+
+		intersection.set(d).scl(tmin).add(p);
```
mgsx-dev, line 598: if intersection is not null :-)

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -498,112 +498,105 @@ public static boolean intersectRaySphere (Ray ray, Vector3 center, float radius,
...
-	 * 
-	 * <p>
-	 * If the origin of the ray is inside the box, this method returns true and the intersection point is set to the origin of the
-	 * ray, accordingly to the definition above.
-	 * </p>
+	 *
 	 * @param intersection The intersection point (optional)
 	 * @return Whether an intersection is present. */
 	public static boolean intersectRayBounds (Ray ray, BoundingBox box, Vector3 intersection) {
-		if (box.contains(ray.origin)) {
```
mgsx-dev, line 515: better keep original behavior (and related documentation) instead of introducing an unnecessary breaking change.

mgsx-dev:
any copyright/licence for the source you copied from?

yuripourre:
@mgsx-dev I've just made the changes based on your comments. About the copyright I searched a lot in the book but didn't find anything related to the codes.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -499,111 +499,116 @@ public static boolean intersectRaySphere (Ray ray, Vector3 center, float radius,
...
-		// max y
-		if (ray.origin.y >= box.max.y && ray.direction.y < 0) {
-			t = (box.max.y - ray.origin.y) / ray.direction.y;
-			if (t >= 0) {
-				v2.set(ray.direction).scl(t).add(ray.origin);
-				if (v2.x >= box.min.x && v2.x <= box.max.x && v2.z >= box.min.z && v2.z <= box.max.z && (!hit || t < lowest)) {
-					hit = true;
-					lowest = t;
-				}
+			if (t2 > tmax) {
```
mgsx-dev, line 546: same mistake as in the other PR :-) less than instead of greater than

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -499,111 +499,116 @@ public static boolean intersectRaySphere (Ray ray, Vector3 center, float radius,
 
 	/** Intersects a {@link Ray} and a {@link BoundingBox}, returning the intersection point in intersection. This intersection is
 	 * defined as the point on the ray closest to the origin which is within the specified bounds.
-	 * 
+	 * defined as the point where the ray hits the box.
```
mgsx-dev, line 502: i think you can revert javadoc changes (both formatting and this statement), current definition is more accurate.

mgsx-dev (changes requested):

mgsx-dev:
considering this algo is very close to what we already have in intersectRayOrientedBoundsFast, let say we took inspiration from there ;-).

FYI, when there is no licence or copyright notice, it's copyrighted by default.

PR #6325, opened 2020-12-20
Remove controllers extension

MrStahlfelge:
We now have controllers in an own repository and released the first version, so we can remove the code from here and change the setup dependencies.

Would be great if someone checks if I found all Eclipse stuff that should be removed.

For the setup, we should consider to add the following to robovm.xml:

`forceLinkClasses`:

        <pattern>com.badlogic.gdx.controllers.IosControllerManager</pattern> 

`frameworks`:

	<framework>GameKit</framework>

(might be added by #6342 anyway)


tommyettinger:
Looks fine from here. I didn't try to open the project in IDEA or Eclipse, though. My only remotely possible areas of concern are that the tests had some extra lines removed that didn't seem related to gdx-controllers, though they were probably not used and your IDE removed those imports automatically. One test had an import for super-koalio removed, another no longer imports Array, etc. If the tests run, this is not an issue, it's just mildly surprising. I think this can be merged once someone tries it with Eclipse.

intrigus:
I've been out of the loop on libgdx developement for some time, but why did you remove `ControllersTest`?

MrStahlfelge:
> I've been out of the loop on libgdx developement for some time, but why did you remove `ControllersTest`?

Because it tests the code that is removed as well.

intrigus:
> > I've been out of the loop on libgdx developement for some time, but why did you remove `ControllersTest`?
> 
> Because it tests the code that is removed as well.

Ok, I guess I would have left the tests in core so they are not scattered over many projects but this is also fine.

Issue #6212, opened 2020-10-10
[Feature Request] Zip file handle


yuripourre:
It may be a specific request but since there is a zip file handler for Android, doesn't hurt to ask for a desktop (maybe multiple backends?) implementation.

I would like to try that in the future but if anyone has time before feel free to do it.

Example code:
https://gist.github.com/MobiDevelop/5514357#file-archivefilehandle-java

MrStahlfelge:
Personally, I achieved this by unpacking the zip to a local (temp) folder. Let's ping @MobiDevelop - it is his code, and his opinion if this should be added to the framework is important imo.

PokeMMO:
I also use an implementation similar to MobiDevelop's on Desktop. This seems like it would be a good addition to libgdx.

crykn:
Just as a quick note: According to Discord, @MobiDevelop is okay with adding ZipFileHandle to libGDX. If anyone is interested in taking this on, please feel free to do so. We also need implementations for iOS and GWT.

Armaxxx:
I will try to tackle this issue, this will be my first issue on an open source project so please be patient.

SaraWinter1:
Geht mir genauso. Ich würde auch lieber Ferienhäuser mit einer guten Internetanbindung buchen und dabei meine Privatsphäre haben, wie bei https://www.sporthotel-oberhof.de/hotel/ferienhaeuser/

anbu93:
For html module
you have error: `No source code is available for type java.util.zip.ZipFile; did you forget to inherit a required module?` 
because `java.util.zip` not emulated JRE http://www.gwtproject.org/doc/latest/RefJreEmulation.html#Package_java_util. 

_A possible workaround is to use JSZip in html backend_

mailinh19:
Thanks for your topic, I am also very interested in it!
[URL=https://trexrun.co]t rex run[/URL]

tugbacelik07:
Teşekkürler, daha kaliteli bir kaynak bulamazdım. Antalya escort olarak bundan daha iyisini görüntüleyemem. http://www.antalyaucak.com

Jensen0000:
I will try to understand this issue, this will be my next open source arrangement

[url=https://cebuinsights.com/]Cebu Travel Guide[/url]

igor98bn:
Good

igor98bn:
Thanks for the solution anbu93

Projectslib:
[...]

PR #6326, opened 2020-12-22
Implementation of zip handling for headless and lwjgl

Armaxxx:
fixes partially #6212

I'm having trouble understanding after weeks of reading but it seems this works for this particular backend. Might be unstable though.


backends/gdx-backend-headless/pom.xml
```diff
@@ -18,6 +18,12 @@
       <artifactId>gdx</artifactId>
       <version>${project.version}</version>
     </dependency>
+      <dependency>
+          <groupId>com.google.android</groupId>
```
MobiDevelop, line 22: Why does the headless backend need to depend on Android?

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/ZipResourceFile.java
```diff
@@ -0,0 +1,389 @@
...
+import java.nio.ByteOrder;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Vector;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+public class ZipResourceFile {
```
MobiDevelop, line 29: Hmm, not super keen on copying this, and then duplicating it in multiple backend.

Armaxxx: right, It seemed that this one was pretty much unchanged, but will refactor them properly, thanks.

backends/gdx-backend-headless/pom.xml
Armaxxx: You're right, I'll refactor them properly. Thanks.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/ZipResourceFile.java
```diff
@@ -0,0 +1,389 @@
...
+			buf.clear();
+			ze.setOffsetFromFile(f, buf);
+
+			// put file into hash
+			mHashMap.put(str, ze);
+
+			// go to next directory entry
+			currentOffset += kCDELen + fileNameLen + extraLen + commentLen;
+		}
+		if (LOGV) {
```
yuripourre, line 386: Empty block: is it on purpose?

Armaxxx: Yeah, but I ought to remove them since they're just empty space.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/ZipResourceFile.java
MobiDevelop: I guess what I am getting at is I don't understand why it is necessary to have this class at all. The Android team created it to solve an Android problem, it's not required in order to read things from a zip file.

Armaxxx: Yeah, now that I read into it it's clear this wasn't needed, I'll keep at it making it possible to read zip files on non android backends.

PR #6327, opened 2020-12-22
Ios delta pause bugfix like on android

### LINKED ISSUE #6093: Large delta after pause/resume (iOS)
Similar to #1196

#### Issue details
If a game is paused for a long time, after resuming, Gdx.graphics.getDeltaTime() will give the delta of the entire pause duration. This causes freeze/stutter on resume as the game tries to process possibly hours worth of delta. This was fixed for the android version but not for iOS which makes it inconsistent across backends.

_Please provide the version(s) affected._
1.9.10

- [ ] Android
- [x] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Armaxxx:
Should fix #6093


MobiDevelop:
Please use separate branches for separate PRs.

Armaxxx:
Oh you're completely right, how silly of me, thought I'd branch them before the fact. Will redo the PR.

Armaxxx:
There, sorry for the revert, now this branch only holds the fix.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -59,7 +59,7 @@
 public class IOSGraphics extends NSObject implements Graphics, GLKViewDelegate, GLKViewControllerDelegate {
 
 	private static final String tag = "IOSGraphics";
-
+	volatile boolean resume = false;
```
obigu, line 62: We can move it with the rest of variable declarations.

I wonder if we need it `volatile`. I'm not sure `resume()` or `draw()` can be called from any other than the main thread. I guess it's not a bad idea to be on the safe side do the same as we do with `paused`.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -308,6 +308,7 @@ public void resume () {
 			for (LifecycleListener listener : listeners) {
 				listener.resume();
 			}
+			resume = true;
```
obigu, line 311: Can we move it outside the synchronized block?

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -354,9 +355,18 @@ public void draw (GLKView view, CGRect rect) {
...
+		if (!resume) {
+			deltaTime = (time - lastFrameTime) / 1000000000.0f;
+		} else {
+			deltaTime = 0;
+		}
 		lastFrameTime = time;
 
+		if (resume) {
+			resume = false;
+		}
```
obigu, line 368: Minor thing but we can probably remove this block and set resume to false on the previous conditional that checks resume value.

obigu (changes requested):
Thanks for the PR. Still have no tested it, just some style stuff.

Regarding the code format changes, we usually try to only apply style changes to new code to keep git history and PRs easier to read.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -354,9 +355,18 @@ public void draw (GLKView view, CGRect rect) {
 		}
 
 		long time = System.nanoTime();
-		deltaTime = (time - lastFrameTime) / 1000000000.0f;
+		// After pause deltaTime can have somewhat huge value that destabilizes the mean, so let's cut it off
```
obigu, line 358: Remove " that destabilizes the mean", this had sense on Android when we applied a mean on `delta`

Armaxxx: Good as done.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Armaxxx: It's been moved but it was kept volatile

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Armaxxx: Good as done

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Armaxxx: You're right, it's kind of redundant.

obigu: thanks for the changes, I'll test it as soon as I can

obigu (approved):

obigu:
Tested, seems to we working, thanks!

MrStahlfelge (changes requested):
There are a lot of unneccesary whitespace changes, please reduce the changes to the necessary ones

PR #6329, opened 2020-12-25
GWT: Initial call to set viewport size fails for resizable application

MrStahlfelge:
d3cbe98eb72ec68bdd8c8630034b50166e8da4f8 introduced resizable applications with no configured with and height, therefore `config.width` and `configh.height` being 0.
The initial call to set the viewport size still based on this constants, causing this call to set an invalid size for resizable applications. This was not noticed because our tests and most games use an own viewport, but of all things, our starter template generated by gdx-setup fails to render correctly without this fix.


MobiDevelop (approved):

SimonIT (approved):

MrStahlfelge:
Thanks for the quick approvals guys ❤️

PR #6336, opened 2020-12-29
Add an Abstract Graphics class

obigu:
Having an abstract class that the rest of the `Graphics` implementations extend from allows us to keep implementations clean while adding the option of adding useful common methods to it.

The following methods have been added to `Graphics` that are implemented on `AbstractGraphics`:
1. ~~`void clear(float r, float g, float b, float a)`~~
2. ~~`void clear(Color color)`~~
3. `float getBackBufferScale ()`

The most important change has been to `iOSGraphics` that has required a bit of refactoring not to extend `NSObject` but it's a positive change. There are also several other minor improvements (documentation, code style, etc...) that made sense to include.

Tested on Android, iOS and Desktop (LwjGL2).


MrStahlfelge:
I like to have abstract core classes holding the shared code of platform implementations, I also introduced such an [abstract super class in gdx-controllers](https://github.com/libgdx/gdx-controllers/blob/master/gdx-controllers-core/src/com/badlogic/gdx/controllers/AbstractController.java) for the same purpose. 

I agree the change on the iOS delegate is a good one, however I still don't like when unchanged inner classes are moved for within the file for no reason. Makes git history and diffs hard to read.

obigu:
> I agree the change on the iOS delegate is a good one, however I still don't like when unchanged inner classes are moved for within the file for no reason. Makes git history and diffs hard to read.

Thanks for having a look. Yeah, me neither, but in this case there's a reason, the class had inner classes declarations both at the top and the bottom and it was messy and hard to read.

MrStahlfelge:
Since it is class that needs to be overriden in projects from time to time, maybe it makes sense to extract it completely instead?

gdx/src/com/badlogic/gdx/AbstractGraphics.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.GL20;
+
+public abstract class AbstractGraphics implements Graphics {
+
+	@Override
+	public void clear(float r, float g, float b, float a){
+		Gdx.gl.glClearColor(r, g, b, a);
+		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
```
mgsx-dev, line 11: I'm not sure it's a good idea to have these clear methods, for 3D the standard way is to clear both color and depth, this could be confusing for users and it doesn't add so much value.
Also, i don't think it's the right place for this kind of methods, Graphics is a backend abstraction while clear method wouldn't be more like a utility around GL interface.
I'm in favor to simply remove these 2 clear methods if you don't mind.

obigu: Ok, I will remove them, the 3D argument is strong.

I've always found that those 2 lines were out of place, in libGDX you don't need to care about the GL interface for anything to do a simple game but you have to copy and paste those 2 ugly strange lines on every screen.  From a non advanced user's perspective, the only thing they do is to clear the screen.

Putting it on a utility class is an option but IMO putting the `Graphics` backend abstraction seems the right place. Clearing the screen every frame seems like part of the `Graphics` functionality any backend would require, in an ideal world (I know it would take more than that) if we had a Non-OpenGL based backend we would also need clear the screen and the `clear` method abstraction would apply.

NathanSweet: I agree the two lines to clear are cryptic. I think the clear methods are pretty OK on Graphics since clearing the screen is part of the boilerplate for all libgdx apps. It's a valid point that some use cases wouldn't use them, but it doesn't seem terribly confusing to me. FWIW I could go either way, with a lean toward having the convenience methods.

crykn: Alternatively, you could add a boolean parameter to the method (and then offer an overloaded version of the method with the param set to false), which if true also clears `GL20.GL_DEPTH_BUFFER_BIT`.

obigu: We all seem to agree it would be good to have the `clear` methods somewhere (with maybe an overloaded version for 3D we can discuss later). 

I've been thinking about it a bit a bit more and I see how `Graphics` is not perfect because it doesn't really deal with rendering stuff, but I'm not sure the alternatives are better.

These are the possible options:
1. Keep it in `Graphics`
2. Utility class. We need some specifics. De we create new class for this? Is there any existing class where this may fit? Are we talking about static utility methods?
3. `GL20`/`GL30` interfaces. If we saw the GLXX interfaces as our Rendering abstraction interface (instead of just a wrapper for OpenGL ES methods), I think this is probably where `clear` would really belong. Alternatively, if we want to keep interfaces that just wrap the OpenGL methods, we can create new Decorator interfaces that have the extra utility methods and are returned by the `Gdx.glXX` methods.

1. and 3. are my preferred options, I probably lean towards 1 for simplicity but I accept 3 is more correct.

@msgx-dev Maybe you can share some more details on how you'd do the utility class option. 

For those who don't like option 1., do you prefer the option 3 alternative?

mgsx-dev: `GLFrameBuffer` could be a reasonable place to put this, we already have a static `unbind` method there and "clearing screen" actually means clearing the current frame buffer, so it'd makes sense.

`ScreenUtils` could be a good place too, class javadoc would need few adaptations but still a good candidate.

`GL20/GL30` would be the worse option IMO. It shouldn't do more than abstracting GLES calls. 

I'm still a bit balanced to put that in `AbstractGraphics`, there is no reasons to override them, these should be static IMO. But if you really want it here, i won't fight against it.

About the depth buffer, a boolean is a good idea, the same way we have one for FrameBuffer constructor: clear(r,g,b,clearDepth) and clear(color,clearDepth). Preferably without overloads, because it's not clear what could be the default. And this way, user knows what he's doing.

I really hope we find a concensus before 2021 :joy: Cheers! :champagne:

crykn: I think `ScreenUtils` would be a good place if we want it in a utility class, good find! Also, I'm against adding it to `GLFrameBuffer`, as I think even `unbind` shouldn't be static. This would allow overriding it in nestable framebuffer implementations.

obigu: Ok, thanks for the feedback, I've removed them for now, `ScreenUtils` seems a good fit, we can take the final decision on a separate issue/PR.

Consensus has been achieved before 2021! (in most parts of the globe at least :D)

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -659,14 +547,136 @@ public void setCursor (Cursor cursor) {
 	public void setSystemCursor (SystemCursor systemCursor) {
 	}
 
+	public static class IOSUIViewController extends GLKViewController {
```
MrStahlfelge, line 550: Please extract this to an own class. It serves not only for graphics, but also for input, and needs to be overriden in game projects often. The only reason to keep it an inner class is to keep the git history, but we erase it with this move so no reason to keep it inner any more.

obigu: Done. The reason was not to break backwards compatibility on this PR but, considering the change makes sense and that  `IOSUIViewController` deserves a refactor, I think it's a good opportunity to move it to its own class now and apply the refactor on the new class later so that it's easier to review.

MrStahlfelge (approved):
I think it is good the way it is now

crykn (approved):
Looks good to me!

PR #6337, opened 2020-12-29
iOS catch exception for unknown keys from hardware keyboard

MrStahlfelge:
I have this crash in my logs:

```
java.lang.IllegalArgumentException: No constant with value 669 (0x29d) found in org.robovm.apple.uikit.UIKeyboardHIDUsage
	at org.robovm.rt.bro.ValuedEnum$AsLongMarshaler.toObject(ValuedEnum.java:142)
	at org.robovm.rt.bro.ValuedEnum$AsMachineSizedSIntMarshaler.toObject(ValuedEnum.java:158)
	at org.robovm.apple.uikit.UIKey.$m$keyCode(Native Method)
	at org.robovm.apple.uikit.UIKey.getKeyCode(UIKey.java)
	at com.badlogic.gdx.backends.iosrobovm.DefaultIOSInput.onKey(DefaultIOSInput.java:660)
```

A UIKeyboardHIDUsage with value 669 is currently not present in RoboVM bindings. Even if it is added, the IllegalArgumentException might be thrown when Apple adds another newly supported key and the game is compiled with a RoboVM version still not knowing it. To avoid crashes, this PR catches the thrown exception and maps to Keys.UNKNOWN instead.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
@@ -699,7 +699,7 @@ public class DefaultIOSInput implements IOSInput {
 			return false;
 		}
 
-		int keyCode = getGdxKeyCode(key.getKeyCode());
+		int keyCode = getGdxKeyCode(key);
 
 		if (keyCode != Keys.UNKNOWN)
 			synchronized (keyEvents) {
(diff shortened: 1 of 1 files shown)


MobiDevelop (approved):
Seems reasonable.

obigu:
LGTM. 

As discussed, the constant is missing on the Apple class itself (https://developer.apple.com/documentation/uikit/uikeyboardhidusage) so I would make sure with MobiVM guys in Gitter the bindings are correct or if it would make sense to change them so that `UIKey.getKeyCode` returns a number instead.

obigu (approved):

PR #6341, opened 2021-01-01
Added clear methods to ScreenUtils

obigu:
Proposal for new `clear` methods on `ScreenUtils`. 

Replaced calls on most test classes.


crykn:
I think this is a good addition! One small thing: I would welcome an override for `clear(...)`, which sets the depth to `1`. This is the default value for `glClearDepthf` and probably by far the most common use case, when one wants to clear the depth buffer.

gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
```diff
@@ -27,12 +28,41 @@
...
+	 * @param colorClear Clears the color buffer if true.
+	 * @param depthClear Clears the depth buffer if true.
+	 * @param color The color to clear with, used only if clearColor is true.
+	 * @param depth The depth to clear the z-buffer with, used only if clearDepth is true. The valid range is from 0 to 1.
+	 * */
+	public static void clear (boolean colorClear, boolean depthClear, Color color, float depth) {
+		int mask = 0;
+		if (colorClear)
+			mask = mask | GL20.GL_COLOR_BUFFER_BIT;
+			Gdx.gl.glClearColor(color.r, color.g, color.b, color.a);
```
NathanSweet, line 58: Indentation like this is not good, makes it look like the if has curlies, but it doesn't.

obigu: Yep, curlies were missing, removed them at some point and forgot to add them back when I added the second line (I'm not used of not adding them). Thanks!

obigu:
> I think this is a good addition! One small thing: I would welcome an override for `clear(...)`, which sets the depth to `1`. This is the default value for `glClearDepthf` and probably by far the most common use case, when one wants to clear the depth buffer.

Sure, whatever is more useful for the 3D guys, I understand you want an overloaded `clear` that doesn't receive `depth` as parameter right? If we add that, do we need to call `Gdx.gl.glClearDepthf ` (considering 1 is default)? 
@mgsx-dev It's be great to have your feedback on this as well.

mgsx-dev:
@obigu to me a single method is enough and preferable : `clear(float r, float g, float b, float a, boolean clearDepth)`

* Using color object is confusing because you usually want to clear alpha channel with zero (especially when dealing with FBOs).
* Messing with glClearDepthf is not a good idea, it's very rare to change it.

If you really want 2 methods, i'm not against having both `clearColor (float r, float g, float b, float a)` and `clear (float r, float g, float b, float a, boolean clearDepth)`.

Also, to be accurate, i would reword javadoc a bit, something like this (with valid English) : `Class with static helper methods related to currently bound OpenGL frame buffer. Can be used to get the entire screen content or a portion thereof`

obigu:
@mgsx-dev Thanks for your input. 

I think it's important to focus the API on the most common use case, in this case it is just to clear your screen with a solid color and that method must be as simple as possible (shouldn't have a boolean about `depth` that would cause confusion). Having the option to pass a `Color` is very useful because many users store their background color as a `Color` somewhere else in their game making the call even simpler/more compact.

For more advanced users, considering from your experience that there's no need for a versatile method that allows choosing which buffers to clear and with which parameters and that the only other useful helper method would be a specific one that always clears the color buffer with the provided color components and that allows clearing the depth buffer with depth 1 (with no need to call `glClearDepthf`), I'd go for the simpler solution @crykn proposed of calling them all `clear` and overload them as they all are similar in that they are specific implementations that assume default values:

`clear(Color color)`
`clear(float r, float g, float b, float a)`
`clear(float r, float g, float b, float a, boolean clearDepth)`


> Also, to be accurate, i would reword javadoc a bit, something like this (with valid English) : `Class with static helper methods related to currently bound OpenGL frame buffer. Can be used to get the entire screen content or a portion thereof`

I'll change that.

mgsx-dev (changes requested):
It looks like we found an agreement :-) 

Also, maybe it's the right time to fix javadoc for other unmodified methods, they refer to "default framebuffer" while they apply to "current framebuffer". The "default framebuffer" still a framebuffer.

Sorry to be nit picky (as usual :D)

obigu:
Great! Thanks for taking the time to review. I've applied the requested changes. Next will be to replace current calls to to glClear with the utility methods when possible.

obigu:
Ready for review/merge now

crykn (approved):

mgsx-dev (approved):

MrStahlfelge (commented):
Let's change [setup template](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/core/src/MainClass) as well using the new methods!

obigu:
Sure, but we need to wait till it's release right? Gdx Setup uses 1.9.13 currently.

MrStahlfelge:
I would prefer to have the template changes linked to the corresponding PR, but we had problems with the approach in the past (y'all know the epic discussions on setup). So we can postpone it, yes.

obigu:
@NathanSweet With your green light we're ready to merge!

PR #6342, opened 2021-01-02
iOS 14 detect connected hardware keyboard reliably

MrStahlfelge:
Since iOS 14, there is now a reliable way to detect a connected hardware keyboard. Until now, it is asumed that a keyboard is connected when a key was pressed. Therefore, we cannot detect a keyboard before the first key press which might confuse users, and we can't detect disconnection of the keyboard. Both is fixed with this PR.

The newly used method is part of GameKit framework, which I added to linked frameworks. It could be discussed if improving such a minor feature justifies the addition a new framework. But I am confident that most games out there will add GameKit sooner or later anyway because of GameCenter or other game related features.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
@@ -31,9 +31,11 @@ import com.badlogic.gdx.utils.Pool;
 import org.robovm.apple.audiotoolbox.AudioServices;
 import org.robovm.apple.coregraphics.CGPoint;
 import org.robovm.apple.coregraphics.CGRect;
+import org.robovm.apple.foundation.Foundation;
 import org.robovm.apple.foundation.NSExtensions;
 import org.robovm.apple.foundation.NSObject;
 import org.robovm.apple.foundation.NSRange;
+import org.robovm.apple.gamecontroller.GCKeyboard;
 import org.robovm.apple.uikit.UIAlertAction;
 import org.robovm.apple.uikit.UIAlertActionStyle;
 import org.robovm.apple.uikit.UIAlertController;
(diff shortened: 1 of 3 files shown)


obigu:
I think it'd be better if we keep libGDX dependencies to the strict minimum, in the case by just adding `GameController` framework (instead of the whole `GameKit`).

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
```diff
@@ -643,7 +645,8 @@ public boolean isPeripheralAvailable (Peripheral peripheral) {
 		if (peripheral == Peripheral.Compass) return compassSupported;
 		if (peripheral == Peripheral.OnscreenKeyboard) return true;
 		if (peripheral == Peripheral.Pressure) return pressureSupported;
-		if (peripheral == Peripheral.HardwareKeyboard) return hadHardwareKeyEvent;
+		if (peripheral == Peripheral.HardwareKeyboard)
+			return (Foundation.getMajorSystemVersion() >= 14) ? GCKeyboard.getCoalescedKeyboard() != null : hadHardwareKeyEvent;
```
MobiDevelop, line 649: Why put the parens around it?

MrStahlfelge:
> I think it'd be better if we keep libGDX dependencies to the strict minimum, in the case by just adding `GameController` framework (instead of the whole `GameKit`).

I was not aware it is possible to link subframeworks. I'll try if it works.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
MrStahlfelge: I just got used to write it this way because I am using Kotlin on other projects. I can remove it here.

MrStahlfelge:
Tested successfully.

obigu (approved):

PR #6346, opened 2021-01-05
move ScreenUtils GWT specifics to Pixmap

mgsx-dev:
ScreenUtils was emulated for GWT but only one method needed it, i made this PR in order to reduce emulated classes and avoid some duplications/desynchronizations between classes. It'll be better if we plan to [add some features to this class in the future](https://github.com/libgdx/libgdx/pull/6341).

As you might not see, there was [a fix in core class](https://github.com/libgdx/libgdx/compare/master...mgsx-dev:fix/screenutils?expand=1#diff-836626286da08c0cec2e3c8e0bf7cffbc4f20b378f89cb5a6d3e3735071fd92eL62) not reported in emu class. This is now fixed de facto.

I moved getFrameBufferPixmap to Pixmap (renamed as createFromFrameBuffer), i hope it make sense to you.

Bonus point, we now have a setPixels method for Pixmap.

Note that ScreenUtils#putPixelsBack (public) was removed but considering user code shouldn't use emu classes directly, we probably don't need to consider it as a breaking change.

PixmapTest and FramebufferToTextureTest validates that all is working correctly. (tested on desktop and html)


gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
```diff
@@ -70,13 +70,7 @@ public static TextureRegion getFrameBufferTexture (int x, int y, int w, int h) {
 	}
 
 	public static Pixmap getFrameBufferPixmap (int x, int y, int w, int h) {
```
MrStahlfelge, line 72: We could add a deprecation annotation here to remove this eventually and keep this non-breaking, or remove the method at all

tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
```diff
@@ -668,6 +668,10 @@ public GdxTest instance () {
 			public GdxTest instance () {
 				return new PixmapPackerTest();
 			}
+		}, new Instancer() {
```
MrStahlfelge, line 671: You can also add `AlphaTest` which tests the changed method

MrStahlfelge (approved):
Code changes are comprehensible, `AlphaTest` and `PixmapTest` work on GWT and Desktop

tests/gdx-tests/src/com/badlogic/gdx/tests/gwt/GwtTestWrapper.java
mgsx-dev: already there ;-)

MrStahlfelge (approved):

obigu (approved):
Thanks! Tested on Android and iOS (`PixmapTest` on iOS doesn't look exactly how it should but it's unrelated to this PR (orientation and notch stuff)).

PR #6348, opened 2021-01-06
Fix missing checks for zero key during Values iteration in IntIntMap, IntFloatMap.

### LINKED ISSUE #6347: IntIntMap.Values and IntFloatMap.Values not handling INDEX_ZERO
I started getting some ArrayIndexOutBounds errors using IntIntMap after upgrading to 1.9.13 from 1.9.10.

Looking into the error, it seems IntIntMap.Values#next() and IntFloatMap.Values#next() don't handle INDEX_ZERO as IntMap.Values does.

IntIntMap.Values ->
```
public int next () {
   if (!hasNext) throw new NoSuchElementException();
   if (!valid) throw new GdxRuntimeException("#iterator() cannot be used nested.");
   int value = map.valueTable[nextIndex];
   currentIndex = nextIndex;
   findNextIndex();
   return value;
}
```

vs. IntMap.Values ->

```
public @Null V next () {
   if (!hasNext) throw new NoSuchElementException();
   if (!valid) throw new GdxRuntimeException("#iterator() cannot be used nested.");
   V value;
   if (nextIndex == INDEX_ZERO)
      value = map.zeroValue;
   else
      value = map.valueTable[nextIndex];
   currentIndex = nextIndex;
   findNextIndex();
   return value;
}
```

tommyettinger:
This fixes #6347 ; big thanks to @tomcashman for making the affected code so easy to spot.

In case what this does isn't obvious, IntMap and LongMap correctly check for whether their Values iterators are currently iterating through the "zero value," which is not in the regular array. When iterating over the value for a key that is zero, the iterator is at position `INDEX_ZERO`, which is -1. While IntMap.Values and LongMap.Values don't try to look up INDEX_ZERO in their value table, IntIntMap.Values and IntFloatMap.Values do, which currently causes an index out of bounds exception in 1.9.13. This makes IntIntMap and IntFloatMap match IntMap's correct behavior.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/MathUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/MathUtils.java
@@ -108,16 +108,40 @@ public final class MathUtils {
 		return y < 0f ? atan - PI : atan;
 	}
 
-	/** Returns acos in radians, less accurate than Math.acos but may be faster. */
-	static public float acos (float a) {
-		return 1.5707963267948966f - (a * (1f + (a *= a) * (-0.141514171442891431f + a * -0.719110791477959357f)))
-			/ (1f + a * (-0.439110389941411144f + a * -0.471306172023844527f));
+	/** Returns acos in radians; less accurate than Math.acos but may be faster. Average error of 0.00002845 radians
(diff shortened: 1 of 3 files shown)


NathanSweet:
Cheers!

PR #6350, opened 2021-01-06
Fix keycode for FOWARD_DEL on GWT and iOS

MrStahlfelge:
Playing around with TextAreaTest, I noticed that pressing forward delete had the same behaviour than pressing backspace.

Cause is a wrong mapping to DEL which is NOT forward del. A check turned out that RoboVM also has a wrong mapping.


SimonIT (approved):
Works on the gwt side

obigu (approved):
Works on iOS (I've tested on Simulator commenting the line preventing KeyType event from being thrown when soft keyboard is visible).

PR #6351, opened 2021-01-06
Introduce an AbstractInput class and consolidate some methods

MrStahlfelge:
After we introduced an abstract graphics class, this PR introduces an abstract input class which aims to consolidate code that is implemented on all (or most) platforms.
We start with pressed key and caught key handlings.

This fixes some minor inconsistencies (keycodes going to 260 on Android, 255 on other platforms). We have a minor behaviour change, too: Setting caught keys will report them as caught when checked with `isCatchKey` (which is not really important, but good imo) even on platforms that have no behaviour change when keys are set to be caught.

With this PR, we also get rid of the two-sided mappings for keycodes on Lwjgl and Lwjgl3 which had inconsistencies as well. And the static methods for key mapping are converted to protected to make overriding in game projects possible.

Testing with SuperKoalio recommended.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -39,13 +39,14 @@ import android.view.View.OnKeyListener;
 import android.view.WindowManager;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
+
+import com.badlogic.gdx.AbstractInput;
 import com.badlogic.gdx.Application;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Graphics.DisplayMode;
 import com.badlogic.gdx.Input;
 import com.badlogic.gdx.InputProcessor;
 import com.badlogic.gdx.backends.android.surfaceview.GLSurfaceView20;
-import com.badlogic.gdx.utils.IntSet;
 import com.badlogic.gdx.utils.Pool;
 
 import java.util.ArrayList;
@@ -56,7 +57,7 @@ import java.util.List;
  * 
  * @author mzechner */
 /** @author jshapcot */
-public class DefaultAndroidInput implements AndroidInput {
+public class DefaultAndroidInput extends AbstractInput implements AndroidInput {
 
 	static class KeyEvent {
 		static final int KEY_DOWN = 0;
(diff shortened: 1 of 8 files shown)


tommyettinger (approved):
This really looks quite a lot cleaner! There's much less duplicate code, so fewer places a mistake could hide later on. Is there a reason the conversion methods between LWJGL and GDX keys all became `protected`, even if `public` before? I doubt those were used heavily, if at all, by user code, but I'm not sure. It might be safer to keep public methods public.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
```diff
@@ -359,28 +354,6 @@ public void setKeyboardAvailable (boolean available) {
 		this.keyboardAvailable = available;
 	}
 
-	@Override
-	public synchronized boolean isKeyPressed (int key) {
```
obigu, line 363: Is it safe to remove this `synchronized`?

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
```diff
@@ -359,28 +354,6 @@ public void setKeyboardAvailable (boolean available) {
...
-			return keyCount > 0;
-		}
-		if (key < 0 || key >= SUPPORTED_KEYS) {
-			return false;
-		}
-		return keys[key];
-	}
-
-	@Override
-	public synchronized boolean isKeyJustPressed (int key) {
```
obigu, line 374: Same for this one

obigu:
This is a good change, nice to see all the Android specific catchXXXKey moved away.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
MrStahlfelge: In my opinion it is. The method only reads a field from a fixed-size array and considering that it should only get called from the main thread, there should be no simulatneous calls.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
MrStahlfelge: Yes, same

crykn (approved):
I really like this change; it makes the code easier to understand, especially in the desktop classes. I only tested the LWGL 2+3 backends, but the rest looks good as well!

PR #6353, opened 2021-01-07
Android raise touchup events when touched array is reset on pause

### LINKED ISSUE #4866: `ClickListener` not working after game lost focus while touching down
#### Issue details
Issue occurs to scene2d Button. Click event doesn't fired after losing focus to screen.

#### Reproduction steps/code
  - Create a simple app: one `Screen` with one `Stage` with two `Button` (see code)
  - Step 1: Touch down both 2 buttons at the same time (don't release yet)
  - Step 2: Press `Home` button on your phone (2 buttons still touch down)
  - Step 3: Reopen app, now 1 button can't fire click event anymore (the button that you touch last)

Below code registers click events that do change background randomly for testing. After go through these reproduce steps, one button just can't change background anymore.

```
class MyGdxGame extends ApplicationAdapter {
    Stage stage;

    @Override
    public void create() {
        stage = new Stage(new StretchViewport(800, 480));

        TextButton.TextButtonStyle style = new TextButton.TextButtonStyle();
        style.font = new BitmapFont();

        TextButton button1 = new TextButton("11111", style);
        button1.setSize(stage.getWidth() / 2, stage.getHeight());
        button1.addListener(new ClickListener() {
            @Override
            public void clicked(InputEvent event, float x, float y) {
                Gdx.gl.glClearColor(MathUtils.random(), MathUtils.random(), MathUtils.random(), 1);
            }
        });

        TextButton button2 = new TextButton("22222", style);
        button2.setSize(stage.getWidth() / 2, stage.getHeight());
        button2.setX(stage.getWidth() / 2);
        button2.addListener(new ClickListener() {
            @Override
            public void clicked(InputEvent event, float x, float y) {
                Gdx.gl.glClearColor(MathUtils.random(), MathUtils.random(), MathUtils.random(), 1);
            }
        });

        stage.addActor(button1);
        stage.addActor(button2);
    }

    @Override
    public void render() {
        stage.act();
        stage.draw();
    }
}
```
LibGdx 1.9.6

- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

MrStahlfelge:
This fixes #4866

When game was sent to background, the touched array was reset, but no touchup events were raised. Thus, listeners and scene2d buttons hung in a clicked state.

Best to test with Scene2dTest and InputTest.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidTouchHandler.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidTouchHandler.java
@@ -23,6 +23,8 @@ import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Input.Buttons;
 import com.badlogic.gdx.backends.android.DefaultAndroidInput.TouchEvent;
 
+import java.util.Arrays;
+
 /** Multitouch handler for devices running Android >= 2.0. If device is capable of (fake) multitouch this will report additional
  * pointers.
  * 
@@ -33,9 +35,9 @@ public class AndroidTouchHandler {
 		int pointerIndex = (event.getAction() & MotionEvent.ACTION_POINTER_INDEX_MASK) >> MotionEvent.ACTION_POINTER_INDEX_SHIFT;
 		int pointerId = event.getPointerId(pointerIndex);
 
-		int x = 0, y = 0;
-		int realPointerIndex = 0;
-		int button = Buttons.LEFT;
+		int x, y;
+		int realPointerIndex;
+		int button;
 
 		long timeStamp = System.nanoTime();
 		synchronized (input) {
@@ -161,5 +163,14 @@ public class AndroidTouchHandler {
 
 	public boolean supportsMultitouch (Context activity) {
 		return activity.getPackageManager().hasSystemFeature("android.hardware.touchscreen.multitouch");
+	}
+
+	void cancelTouchState(DefaultAndroidInput input) {
+		long timeStamp = System.nanoTime();
+		for (int i = 0; i < input.realId.length; i++) {
+			if (input.button[i] != -1 && input.touched[i])
+				postTouchEvent(input, TouchEvent.TOUCH_UP, input.touchX[i], input.touchY[i], i, input.button[i], timeStamp);
+		}
+		Arrays.fill(input.touched, false);
 	}
 }
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -1047,7 +1047,7 @@ public class DefaultAndroidInput implements AndroidInput {
 		Arrays.fill(realId, -1);
 
 		// erase touched state. this also sucks donkeyballs...
-		Arrays.fill(touched, false);
+		touchHandler.cancelTouchState(this);
 	}
 
 	@Override
@@ -1062,11 +1062,7 @@ public class DefaultAndroidInput implements AndroidInput {
 
 	@Override
 	public void onDreamingStopped () {
-		unregisterSensorListeners();
-		// erase pointer ids. this sucks donkeyballs...
-		Arrays.fill(realId, -1);
-		// erase touched state. this also sucks donkeyballs...
-		Arrays.fill(touched, false);
+		onPause();
 	}
 
 	/** Our implementation of SensorEventListener. Because Android doesn't like it when we register more than one Sensor to a single
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/InputTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/InputTest.java
@@ -86,7 +86,7 @@ public class InputTest extends GdxTest implements InputProcessor {
 
 	@Override
 	public boolean touchDown (int x, int y, int pointer, int button) {
-		Gdx.app.log("Input Test", "touch down: " + x + ", " + y + ", button: " + getButtonString(button));
+		Gdx.app.log("Input Test", "touch down: " + x + ", " + y + ", button: " + getButtonString(button) + ", pointer: " + pointer);
 		return false;
 	}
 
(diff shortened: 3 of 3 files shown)


obigu:
If I understand correctly this change performs a touchUp automatically on activity pause which can lead to undesired actions or clicks if a user is touching the screen and the activity changes state (like receiving a call). For instance, if in your game you have to drag your finger to aim and release to shoot, if the activity is paused while aiming, with this change the shot will occur right after resume on the direction your finger was last before pause, correct?

I think Android registers an ACTION_CANCEL event when it goes onPause() which means any gesture that was occuring is cancelled. On libGDX we lack a `touchCancel` event which would be useful to have control over what to do on this scenario.

MrStahlfelge:
Good objection, this will indeed happen. But this is "correct" (as we have no cancel): 
* If a game polls the touch state, it will also register a touch up at the moment. With this change, event based input matches that behaviour
* It is implemented this way on RoboVM, too [1]. What you describe applies for libGDX on iOS, I just rechecked. This is barely noticed and in my opinion better than on Android where the pressed state hangs. 

So I think we are better off with this change.

[1] https://github.com/libgdx/libgdx/blob/f43e2d7f4d19cea6cc01089caf8c8083b6ac4a62/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java#L774

obigu:
I don't think this is the "correct" and that this should be the default. It will cause unintended behaviour in more occasions than situations that may end up buggy (you need to be touching with 2 fingers when pause occurs to trigger the issue).

> * If a game polls the touch state, it will also register a touch up at the moment. With this change, event based input matches that behaviour

Well, it's not really a touch up is it? it just resets all inputs to default. I made that PR for a very specific issue on some devices but wouldn't affect gesture management with InputListeners.

> * It is implemented this way on RoboVM, too [1]. What you describe applies for libGDX on iOS, I just rechecked. This is barely noticed and in my opinion better than on Android where the pressed state hangs.

The major differencee is that on iOS you don't get a CANCEL event when you receive a call or other non user-triggered situations like you would on Android making it a not-issue.

I think we should have a thought on how to handle Cancel, this change would break bakwards compatibility heavily (it would on my 5 games). 

Maybe @NathanSweet can help?

MrStahlfelge:
> Well, it's not really a touch up is it? 
It is if games poll state and the state changes from touched to untouched, it is same as a touch up. 

> The major differencee is that on iOS you don't get a CANCEL event when you receive a call
The code snippet I linked is called on iOS when app goes into background, that's why the issue with the hanging down state is not present on iOS. Please try it, you'll see that the touch up is called on iOS. By the way, it is the same on GWT: https://github.com/libgdx/libgdx/blob/24a6c9fb03eb7c8f8de17dcb71238a41a04cf4dc/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtInput.java#L748

I think introducing a touch cancel is a too big change for this minor problem. Granted, it is a minor problem, but at the moment we have different behaviour on the platforms which is fixed with this PR. This is not a scene2d problem, by the way. I can just repeat the problems we have at the moment:

* We have different behaviour on the platforms at the moment
* We have the problem that polling and event based input differ on Android on pause while touched, which is why I think Android is the one with the problem

If 5 of your games are concerned by this change, I am afraid that they already are with your iOS versions.

obigu:
I've tested it on iOS devices. On iOS no CANCEL touch events are registered for non user triggered interruptions like a call. This means that, even if you get a CANCEL event if you do something like pressing the lock button on the device it's not an issue because that's a user triggered action so no unintended behaviour will occur.

The problem on Android is that events not initiated by the user can trigger a CANCEL and cause unintended behaviour. This change would actually make platform to behave differently.

MrStahlfelge:
For me, the touchUp events are fired on iOS when the game comes back to foreground. I am sure that @piotr-j can acknowledge as he is familiar with the bug on Android and knows iOS well, too.

NathanSweet:
I don't know anything about how events are handled on mobile, but I don't think it's unreasonable to send a touch up event in order to uphold the contract that every touch down has a corresponding touch up.

FWIW, in scene2d Stage `cancelTouchFocusExcept` cancels touches for this reason by simulating a touch up, similar to this PR. It's a bit hacky, but I couldn't come up with a less clumsy way of doing it. Sometimes listeners need to know if the touch up was actually for cancelling focus. To do that InputEvent has to do this magicke nasty:
```
public boolean isTouchFocusCancel () {
	return stageX == Integer.MIN_VALUE || stageY == Integer.MIN_VALUE;
}
```

obigu:
> I don't know anything about how events are handled on mobile, but I don't think it's unreasonable to send a touch up event in order to uphold the contract that every touch down has a corresponding touch up.
> 

@NathanSweet On mobile (both Android and iOS) the contract that every touch down has a corresponding touch up doesn't apply. A touch down can be followed by a touch cancel without the need of a touch up. The reason for this is to prevent unwanted gestures to be triggered under several scenarios such as an external interruption (depending on the device can be a phone call or a notification) or "strange" touch gestures such as starting with a touch down and progressively openning your hand palm on top of the screen.

You can test this on any Android or iOS native app,  while touching the screen (maybe a touching down a button), press the Home button on your device. When you get back the button will not have been pressed (not touch up occured).

Do you think we can add this behaviour to libGDX?

NathanSweet:
Canceling with a touch up event is more convenient, requiring less special cases. It can still be surprising though and needs a way of checking that a touch up is actually a touch cancel. If we want it in libgdx, that is probably what I would suggest.

If canceling touch is done with a touch up, like in scene2d, then everything mostly just works. Where it doesn't, something needs to check if the touch up is a cancel, usually to abort what it was doing. For example, a slider see it's a touch cancel and reverts the slider position to before the touch down. Other things don't need special behavior, eg buttons/click listeners see the touch up at Integer.MIN_VALUE and so don't trigger a click. Everyone always sees a touch up for every touch down, which is very convenient.

If canceling touch is done by some other mechanism, then almost everyone who cares about touch down/up needs to also care about touch cancel. For example, click listeners need to know about touch cancel so they can adjust their state.

obigu:
Sounds like a good approach. I would say that it'd be worth also going through classes that wouldn't require special handling such as ClickListener as when it's a "touch cancel" we could probably optimize by avoiding some logic (like determining if touch is over) . 
`Input` implementations on backends that support touch cancel would fire the touch up set to Integer.MIN_VALUE when a real CANCEL is received. As a side effect we may even be able to remove the logic to reset touch states on `onPause()`.

MrStahlfelge:
Closed in favor of #6871

Issue #6190, opened 2020-09-16
LWJGL3 don't have LwjglAWTCanvas


kowlown:
#### Issue details
LwjglAWTCanvas is not present is LWJGL3

#### Version of LibGDX and/or relevant dependencies
Version 1.9.10

- [ ] Android
- [ ] iOS (robovm)
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] MacOS

crykn:
Hey @kowlown, sadly GLFW (which is used by LWJGL3) doesn't really play nice with AWT. There is nothing much we can do about this from the libGDX side, but there have been some general efforts to remedy this situation (see the [lwjgl3-awt](https://github.com/LWJGLX/lwjgl3-awt) project). The problem is that up until recently lwjgl3-awt didn't support Mac OS and a proper release supporting Mac OS is still missing. As long as that is the case, our hands are tied.

yuripourre:
Just an idea, this library is awesome and is built over OpenGL 2, maybe it can be used as a starting point.

https://github.com/brandonborkholder/glg2d

cordelleus:
@kowlown 
AWT is meant only for web browser support.  It figures in automatically, if you compile for AWT as your target of choice.  You only need to test Google Chrome for compatibilities.

tommyettinger:
I'm going to keep deleting any gibberish messages that are completely off-topic and factually wrong. Monarchis, we know it is you, you were blocked for a reason. Again, please seek help from a mental health professional; your messages are nonsense.

cordelleus:
@tommyettinger 
We shall see about removing you from GitHub.

NathanSweet:
Our fearless leader, EL MARIO, has made a GLFW fork so it works with AWT:
https://github.com/badlogic/glfw

What it do:
> Makes GLFW not start on the main thread, then dispatches all GLFW calls made on the non-main thread to the main thread.

What is needed is the forked `libglfw.dylib` and:
https://gist.github.com/badlogic/bac3284aa4c93378c3d7ec9e2c0f4fda#file-gistfile1-txt-L63-L64

shaun-wild:
https://github.com/LWJGLX/lwjgl3-awt

This now has support for windows, mac and linux, may it be worth revisiting this?

kowlown:
Thanks for the feedbacks. I will close this issue

PR #6357, opened 2021-01-10
Input improvements for the LWJGL2 backend

NathanSweet:
- InputEventQueue: no longer implements InputProcessor (never meant to be used as one), added time to the event methods rather than using the current time (LWJGL2 uses that), synchronization for setProcessor.
- LWJGL2 input: use InputEventQueue, missing `@Override` (if we're going to use the silly thing in a file, should at least be consistent :p), removed obsolete methods (supportsMultitouch, etc).
- LWJGL3 input: pass time to InputEventQueue, source format, avoid synthetic accessors for AbstractInput fields (in GLFWKeyCallback).
- Made getGdxKeyCode public (made protected in #6351).


NathanSweet:
Oh boy, the PR included lots of commits for some reason. The actual files changed are correct, just need to be sure we squash when merging.

MrStahlfelge:
Yeah we discussed on the other PR if the method is needed public, so you do use it not from subclasses but from somewhere else? What for (this is just out of interest).

Also out of interest, why the changes in initializing the variables in AbstractInput?

crykn:
Uhm, with the PR `Gdx.input.isKeyPressed(...)` no longer works with the LWJGL2 backend. This is because the `pressedKeys` array is no longer updated.

NathanSweet:
> so you do use it not from subclasses but from somewhere else? What for (this is just out of interest).
@MrStahlfelge I have a fork for LWJGL2 input that replaces the event queue with one that processes only one touch up per frame (simplifies some things in my app).

> Also out of interest, why the changes in initializing the variables in AbstractInput?
@MrStahlfelge It's not a big difference, but the idea is less code is easier to read code. If I see values assigned I have to read what they are. If there are no values it's the tiniest bit easier to see they are the default.

> with the PR Gdx.input.isKeyPressed(...) no longer works with the LWJGL2 backend
@crykn Yikes, good catch! Fixed.

tommyettinger:
Is there a plan for how long we'll maintain the LWJGL2 backend? The plan could be "forever," I guess, or at least until OpenGL stops being supported. It certainly isn't going to work on new Macs natively, due to the ARM libraries not existing and no one writing new LWJGL2 code, but could be emulated probably for as long as Apple decides to permit that.

This PR looks good, also. I don't really understand InputEventQueue or how to use it, but if it's mostly internal to libGDX, that makes more sense. I also have not figured out why Git sometimes remembers previous PRs and keeps their history in a later one, but since we always squash-merge now, it doesn't pollute the repo.

NathanSweet:
I'm stuck with LWJGL2 until LWJGL3 can be used with AWT. Seems like that's in progress. We don't need to keep LWJGL2 for me though, I can fork it if it's removed from libgdx. Though if it isn't hurting anything, it may as well stay.

InputEventQueue stores input events that can occur on any thread and later fires them on the game thread.

crykn:
> I'm stuck with LWJGL2 until LWJGL3 can be used with AWT. Seems like that's in progress. 

It's not only in progress; #6190 contains a working Lwjgl3AWTFrame based on [lwjgl3-awt](https://github.com/LWJGLX/lwjgl3-awt).

NathanSweet:
Yeah that looks cool. I'll need to find time to dig into it deeper.

Issue #6345, opened 2021-01-05
AndroidFiles initialization does disk I/O on UI thread (Android)


cthomis:
#### Issue details
Initializing the AndroidFiles object (for example, from com.badlogic.gdx.backends.android.AndroidFragmentApplication#initializeForView) will cause a disk read on the UI thread. 
As of 1.9.12, there is also a potential write to disk from the UI thread at com.badlogic.gdx.backends.android.AndroidFiles#<init> -- since https://github.com/libgdx/libgdx/pull/6126 causes the AndroidFiles external path to be eagerly initialized, this can cause the application to try and create those directories on the UI thread if they haven't yet been initialized. 

#### Reproduction steps/code
Enable StrictMode on an Android application (https://developer.android.com/reference/android/os/StrictMode), disallow disk reads and writes, and observe errors on application startup when initializing AndroidFiles.

#### Version of LibGDX and/or relevant dependencies
Disk read seems to have been an issue for a while, disk write issue introduced as of 1.9.12.

#### Please select the affected platforms
- [x ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

MrStahlfelge:
Initializing asynchronously might lead to problems when applications access the file system at startup. We already have similar issues when playing sounds on application startup on Android. 

To keep backwards compatiblity and the ability to access the file system on startup, but to avoid access when not necessary, we could add a new configuration option externalFileAccess: Disabled, Enabled, EnabledOnStartup or make the AndroidFiles class initialization overridable like we do with input and graphics.

obigu:
IMO I don't think it's a big deal the fact that an empty file is created on Android's external path executing on the main thread the first time an app runs by `contextWrapper.getExternalFilesDir(null);`. Except from the general rule of not being a good practice to do disk or network stuff on the main thread I don't think it's worth the hassle and complexity of making it asynchronous.

@cthomis Have you come across any issue due to this or can tell us more about the risks you see or you just found out when enabling Strict mode?

@MrStahlfelge Independently from this issue, I think that initializing `Files` the same way we do with `Audio` or `Input` to improve extensibility would be a good change.

MrStahlfelge:
@cthomis Please check the linked PR. It gives you the possibility to use `NoExternalAndroidFiles` to disable external file access completely, or to use your own async implementation to access files.

MrStahlfelge:
[...]

PR #6358, opened 2021-01-10
Make instantiation of Files overridable

MrStahlfelge:
In reaction to #6345, this makes the instantiation of Files implementation overridable on Android. GWT is done, too, because overriding it is useful in projects often to change the handling of local files. Discussing with @obigu, we never had the need to change Files on desktop or iOS, so this is not changed here.

Testing discovered that `FilesTest` was broken on non-GWT platforms by #6188 and is fixed hereby.

Entry suggestion for CHANGES file:
```
* API Change: AndroidFiles and GwtFiles instantiation can be overriden by own classes. NoExternalAndroidFiles implementation can be used to avoid unnecessary write to external files dir if not needed (see #6345)
```


obigu:
I think that even if we have never had the need to change them, we should follow the same approach on all backends (same as `Audio` and `Input`.

MrStahlfelge:
Apart from that, what's your opinion on this change?

obigu:
I like giving the option of extendibility, I think it's a good change. Regarding the details of the Android implementation I would probably just pass a `boolean` to the `DefaulAndroidFiles` constructor to indicate if it requires "external" and initialize accordingly and avoid the protected `initExternalFilesPath` method and alternative `NoExternalAndroidFiles` that is a bit too complex IMO.

MrStahlfelge:
Good idea with the param and get rid of NoExternalAndroidFiles. I would like to keep the initExternalFilesPath though. It is possible to override it to use another path.

MobiDevelop:
I agree on ditching the `NoExternalAndroidFiles` subclass.

MrStahlfelge:
Changes done

crykn:
I think you should add `@Override` annotations to make it consistent with `createInput()` and `createAudio()`. And for the same reason: maybe make the method public as well? Apart from that, it looks good to me.

MrStahlfelge:
Thanks for looking at this. 

Just adding an `Override` annotation would break the compile because the new methods don't override anyhting, they are meant to be overriden. :) 

I don't know why the two other factory methods were added to the super interface and were made public. I don't see any use case to call them from outside to instantiate a second or third audio or files handler. Maybe @obigu can tell us his intention behind this.

PR #6365, opened 2021-01-22
New IosTestWrapper derived from GwtTestWrapper

MrStahlfelge:
Ios Test project had no wrapper for the tests to start them conviniently, so I refactored the GwtTestWrapper to AbstractTestWrapper for use with iOS:

* Renamed GwtTestWrapper to AbstractTestWrapper and moved package, fixed viewport
* Moved filling the instancer list for Gwt to a new GwtTestWrapper class
* Added IosTestWrapper with another list creation implementation based on GdxTests list

Unfortunately, going back to the list with ESC does only work on real devices with attached keyboard. If someone has a good idea what key or gesture we could use to switch back, I am all ears. :)


obigu (approved):
Thanks for this! Tested on a device and works well. 

You can go back to the test list by touching the top left corner of the screen on any test, I think it's good enough. 

Even if that touch area might interfere with some tests that require interaction on that area of the screen but it's a minor issue (I guess it may also affect GWT). We can deal with these cases on a separate PR and fix other iOS releated issues (some tests for instance don't take notches and test is hidden behind them or show unexpected results).

MrStahlfelge:
Ha, I never noticed you can click/touch in the upper corner to go back! I will change the label to give a hint when I resolve the conflict.

PR #6366, opened 2021-01-22
Add HdpiMode option to iOS backend

JoelOtter:
This change adds a HdpiMode option to the iOS backend's IOSApplicationConfiguration, which is used to calculate pixel coordinates for input and return heights and widths in raw pixels on the `getWidth()` etc calls. Safe areas are also returned in raw pixel sizes. The default is left as Logical pixels, the current iOS behaviour.

The rationale behind this is that some LibGDX games - including my own - use pixel scaling for things like UI positioning, texture tiling, and font generation. Porting to the new "logical pixels" system proved to be very difficult. This gives developers the same option they have on the LWJGL3 backend, and more closely matches the Android backend which to my knowledge always uses raw pixels. I've tested it with my own game and it seems to work.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
@@ -796,16 +796,15 @@ public class DefaultIOSInput extends AbstractInput implements IOSInput {
 		for (int i = 0; i < length; i++) {
 			long touchHandle = NSArrayExtensions.objectAtIndex$(array, i);
 			UITouch touch = UI_TOUCH_WRAPPER.wrap(touchHandle);
-			final int locX, locY;
+			int locX, locY;
 			// Get and map the location to our drawing space
 			{
 				CGPoint loc = touch.getLocationInView(app.graphics.view);
+				locX = (int) (loc.getX() - screenBounds.x);
+				locY = (int) (loc.getY() - screenBounds.y);
 				if (config.hdpiMode == HdpiMode.Pixels) {
-					locX = (int) ((loc.getX() * app.pixelsPerPoint) - (screenBounds.x * app.pixelsPerPoint));
-					locY = (int) ((loc.getY() * app.pixelsPerPoint) - (screenBounds.y * app.pixelsPerPoint));
-				} else {
-					locX = (int) (loc.getX() - screenBounds.x);
-					locY = (int) (loc.getY() - screenBounds.y);
+					locX = (int) (locX * app.pixelsPerPoint);
+					locY = (int) (locY * app.pixelsPerPoint);
 				}
 				// app.debug("IOSInput","pos= "+loc+"  bounds= "+bounds+" x= "+locX+" locY= "+locY);
 			}
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
@@ -23,6 +23,7 @@ import com.badlogic.gdx.backends.iosrobovm.custom.UIAcceleration;
 import com.badlogic.gdx.backends.iosrobovm.custom.UIAccelerometer;
 import com.badlogic.gdx.backends.iosrobovm.custom.UIAccelerometerDelegate;
 import com.badlogic.gdx.backends.iosrobovm.custom.UIAccelerometerDelegateAdapter;
+import com.badlogic.gdx.graphics.glutils.HdpiMode;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.Pool;
@@ -799,8 +800,13 @@ public class DefaultIOSInput extends AbstractInput implements IOSInput {
 			// Get and map the location to our drawing space
 			{
 				CGPoint loc = touch.getLocationInView(app.graphics.view);
-				locX = (int)(loc.getX() - screenBounds.x);
-				locY = (int)(loc.getY() - screenBounds.y);
+				if (config.hdpiMode == HdpiMode.Pixels) {
+					locX = (int) ((loc.getX() * app.pixelsPerPoint) - (screenBounds.x * app.pixelsPerPoint));
+					locY = (int) ((loc.getY() * app.pixelsPerPoint) - (screenBounds.y * app.pixelsPerPoint));
+				} else {
+					locX = (int) (loc.getX() - screenBounds.x);
+					locY = (int) (loc.getY() - screenBounds.y);
+				}
 				// app.debug("IOSInput","pos= "+loc+"  bounds= "+bounds+" x= "+locX+" locY= "+locY);
 			}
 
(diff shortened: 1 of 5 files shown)


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
```diff
@@ -799,8 +800,13 @@ private void toTouchEvents (long touches) {
 			// Get and map the location to our drawing space
 			{
 				CGPoint loc = touch.getLocationInView(app.graphics.view);
-				locX = (int)(loc.getX() - screenBounds.x);
-				locY = (int)(loc.getY() - screenBounds.y);
+				if (config.hdpiMode == HdpiMode.Pixels) {
```
MrStahlfelge, line 803: I think it is better understandable what's going on if we would have code like this:

			locX = (int) (loc.getX() - screenBounds.x);
			locY = (int) (loc.getY() - screenBounds.y);
			if (config.hdpiMode == HdpiMode.Pixels) {
				locX = locX * app.pixelsPerPoint;
				locY = locY * app.pixelsPerPoint;
			}

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -433,33 +448,53 @@ protected void updateSafeInsets() {
 		safeInsetLeft = 0;
 		safeInsetRight = 0;
 		safeInsetBottom = 0;
+		safeInsetRawTop = 0;
```
MrStahlfelge, line 451: The new raw values are the same as the other ones, multiplied with a (more or less) constant... would it be better to just have the four logical values and multiply when getter method is called?

MrStahlfelge (commented):
This looks save to merge for the existing logical functionality, I did not test if using Pixels works without problems so far.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
JoelOtter: Sure, updated.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
JoelOtter: I thought to leave these "cached" as it's a slight performance hit doing a float multiplication if this is called in every `render()`. Acknowledge it's negligible so happy to change if you prefer.

MrStahlfelge: Background of the save fields is less the performance hit but more that otherwise we would have the need to call the native methods 4 times to get the values. I don't think there is any performance hit here if the values are multiplied, so I would prefer the less polluted graphics class. But that's my opinion, let's see if others have an opinion on this.

Darkyenus: Is there a good reason to not just reuse the existing fields to store pixel offsets instead of point offsets? I don't see any just by looking at the file in GitHub and it would solve this whole discussion.

MrStahlfelge: You are completely right, they can be reused.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSUIViewController.java
```diff
@@ -79,7 +80,11 @@ public void viewDidLayoutSubviews () {
 		if (graphics.created && (newBounds.width != oldBounds.width || newBounds.height != oldBounds.height)) {
 			graphics.makeCurrent();
 			graphics.updateSafeInsets();
-			app.listener.resize(newBounds.width, newBounds.height);
+			if (graphics.config.hdpiMode == HdpiMode.Pixels) {
+				app.listener.resize(newBounds.width, newBounds.height);
+			} else {
+				app.listener.resize(newBounds.backBufferWidth, newBounds.backBufferHeight);
+			}
```
obigu, line 87: Shouldn't it be the opposite?

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -319,11 +328,17 @@ public void setGL30 (GL30 gl30) {
 
 	@Override
 	public int getWidth () {
+		if (config.hdpiMode == HdpiMode.Pixels) {
+			return getBackBufferWidth();
+		}
```
obigu, line 333: Minor nitpick but can we have an `if`  `else` structure on all the conditionals?

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
obigu: Yep, and even better, we could have a single int[] to hold the 4 values.

obigu: Yep, and even better, we could have a single int[] to hold the 4 values.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSUIViewController.java
JoelOtter: 🤔 You're right, it probably should. Not sure how this worked...

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
JoelOtter: Sure, done

obigu (approved):
Thanks. Tested on a real project and have not found any issue.

@JoelOtter I agree this makes it consistent with the libGDX API and the option to select the HDPI mode so it's a good change. I don't share, though, that using logical pixels instead of physical pixels may be difficult on any scenario (including the ones you've mentioned) considering the current API. I'm happy to discuss it on Discord if you want to provide a use case.

JoelOtter:
@obigu It's probably because I've built my game in a very strange way, but all my UI scaling and font generation is done in pure pixel terms, and the move to logical pixels broke my game in a way I wasn't able to fix by simply changing `getWidth` to `getBackBufferWidth` and so on. I can't speak for anyone else but I feel it likely that if I managed to (mis?)use libGDX in this way others may well have done so too. Also happy to chat on Discord!

MrStahlfelge:
Let's move the entry in the CHANGES file to the correct position and we can merge.

JoelOtter:
> Let's move the entry in the CHANGES file to the correct position and we can merge.

Ooh, I missed the version bump. Done. :)

devAbreu:
@JoelOtter  Thanks for this collaboration. I was having the same behavior. I spent three days trying to find out how to solve this until I found this PR.

PR #6369, opened 2021-01-25
Replace eclipse project files with gradle task

PokeMMO:
Instead of having .project/.classpath/.settings in git we should be using gradle's eclipse plugin.

You don't have to use gradle in eclipse, a `./gradlew cleanEclipse eclipse` will generate .project/.classpath/.settings files for you to use like normal.


build.gradle
```diff
@@ -52,6 +62,36 @@ configure(subprojects - project(":tests:gdx-tests-android")) {
     }
 }
 
+
+
+subprojects {
+    //This has to be dine in afterEvaluate due to java/android plugin dependency
```
tommyettinger, line 68: Typo: dine -> done .

build.gradle
```diff
@@ -40,10 +43,17 @@ allprojects {
...
+        options.encoding = 'UTF-8'
+    }
+
+    tasks.withType(Test) {
+        systemProperty 'file.encoding', 'UTF-8'
+    }
 }
 
 configure(subprojects - project(":tests:gdx-tests-android")) {
-    apply plugin: "idea"
```
tommyettinger, line 46: Why is the `idea` plugin getting removed here? Is it no longer needed?

extensions/build.gradle
```diff
@@ -20,8 +20,6 @@ subprojects {
     sourceCompatibility = 1.7
     targetCompatibility = 1.7
 
-    compileJava.options.encoding = 'UTF-8'
```
tommyettinger, line 23: This line is somewhat concerning; I've hit all sorts of issues when the encoding is not explicitly set to UTF-8, typically on Windows. Recall that Java uses the platform encoding if unspecified.

tests/build.gradle
```diff
@@ -22,8 +22,6 @@ configure(subprojects - project(":tests:gdx-tests-android")) {
     sourceSets.main.java.srcDirs = ["src"]
     sourceSets.main.resources.srcDirs = ["res"]
 
-    compileJava.options.encoding = 'UTF-8'
```
tommyettinger, line 25: Again, why is the encoding being reverted to platform-specific here? Is it set in the root maybe? I use
```gradle
[compileJava, compileTestJava, javadoc]*.options*.encoding = 'UTF-8'
```
in all of my projects, personally.

build.gradle
```diff
@@ -40,10 +43,17 @@ allprojects {
...
         maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
     }
+
+    tasks.withType(JavaCompile) {
+        options.encoding = 'UTF-8'
+    }
+
+    tasks.withType(Test) {
+        systemProperty 'file.encoding', 'UTF-8'
+    }
```
tommyettinger, line 53: Ah, I just saw the UTF-8 getting set up here. This is a good place to set this. I'll resolve my earlier comments.

tommyettinger (approved):
I'm only slightly concerned about one place removing its `idea` plugin, but I can't remember if IDEA even uses that plugin anymore. If it still loads correctly in IDEA, I approve.

build.gradle
PokeMMO: It got moved to where I added eclipse, so that all projects get it.

https://github.com/libgdx/libgdx/blob/3defef117ac6ffd328788d634fc2162c9f300525/build.gradle#L20

build.gradle
PokeMMO: The idea plugin is still used, it's now in a better spot next to eclipse's apply.


https://github.com/libgdx/libgdx/blob/3defef117ac6ffd328788d634fc2162c9f300525/build.gradle#L20

tommyettinger: Great! That does seem like a better place.

PokeMMO:
@NathanSweet Can you review this PR when you get a chance? I'm not willing to merge it until you give it an ok.

PokeMMO:
- I'm not sure what you mean by export. Do you mean lwjgl natives aren't being added to projects which depend on gdx-backend-lwjgl?
- The projects have different names in gradle apparently? https://github.com/libgdx/libgdx/commit/446e26910cd30fbcda8518d979f75c43bcf44c82
  - This would result in the eclipse project name being different too.
- All tests should be UTF-8 when run via gradle, is this an issue when running from eclipse?
- Not sure why I never go that issue, but I'll verify/fix.
- The formatter settings were taken from the old `org.eclipse.jdt.core.prefs` files only removing a couple generatable values. I dont know why git is struggling to diff it properly.

PokeMMO:
UTF-8 is now used for all projects, as well as fixed the gwt classpath issue when running the eclipse task multiple times without using cleanEclipse.

NathanSweet:
Sorry, I missed your response!

> I'm not sure what you mean by export. Do you mean lwjgl natives aren't being added to projects which depend on gdx-backend-lwjgl?

In Eclipse, dependencies are only provided to dependent projects if they are "exported". On an Eclipse project, properties, Java Build Path, Order and Export tab, only the dependencies that are checked will be seen by projects that depend on this project.

If the other issues have been addressed I can take it for a spin again.

PokeMMO:
I've marked all classpath entries as exported, so that should solve the last issue.

Make sure to remove any bin folders before testing, I had some issues with copies of old .project files being left there.

PokeMMO:
@NathanSweet Any chance you can give this another review so it can be merged soon?

NathanSweet:
1) Just curious, why does `gradlew cleanEclipse eclipse` need to compile? I'd expect it to only generate Eclipse files, as Maven does. It seems to do that first and quickly, but then proceeds to compile things. The first time it took 46s. I started over with a fresh clone and the second time it got stuck here for 13+ minutes, until I killed it:
```
<=========----> 72% EXECUTING [13m 25s]
> :extensions:gdx-bullet:compileJava > Resolve files of :extensions:gdx-bullet:compileClasspath
```
Third time it finished in 28s. Weird.

2) I cloned your branch into a new folder, so I have no old bin/etc files, then generated the Eclipse files with the above. The resulting Eclipse projects have dependencies from `~/.m2`. Good so far, except gdx-tools, gdx-tests, gdx-tests-lwjgl, and gdx-tests-lwjgl3 each have 4 of these errors:
```
Project 'gdx-*' is missing required library:
```
For these paths:
```
C:\Dev\libgdx\extensions\gdx-box2d\gdx-box2d\libs\gdx-box2d-natives.jar
C:\Dev\libgdx\extensions\gdx-bullet\libs\gdx-bullet-natives.jar
C:\Dev\libgdx\extensions\gdx-freetype\libs\gdx-freetype-natives.jar
C:\Dev\libgdx\gdx\libs\gdx-natives.jar
```
Is this on purpose? I can use `fetch.xml` to get them, but I'm not sure that is intended. I would guess the corresponding project for each of these JARs should have the JAR from `~/.m2` as a dependency with export checked, so all projects that depend on the project get the JAR transitively.

3) gdx-bullet still has 5 of these warnings:
```
Description: The resource is a duplicate of jni/swig-src/collision/classes.i and was not copied to the output folder
Resource: classes.i
Location: Unknown
Type: Java Problem
```
One for each of these 5 paths:
```
/gdx-bullet/jni/swig-src/dynamics
/gdx-bullet/jni/swig-src/extras
/gdx-bullet/jni/swig-src/inversedynamics
/gdx-bullet/jni/swig-src/linearmath
/gdx-bullet/jni/swig-src/softbody
```
Each of those are source folders (in addition to `/gdx-bullet/jni/swig-src/collision`) so their contents are copied to the bin dir. The warning is that they all contain `classes.i` and are copied to the bin dir with the same name. I would guess we don't want `classes.i` in the output folder at all, so we should exclude it from all 6 source folders (along with the `.cpp` and `.h` files in those folders).

PokeMMO:
1. I've noticed the same behavior with other gradle tasks. It's related to `:tests:gdx-tests-gwt:warTemplate` Running `./gradlew cleanEclipse eclipse -x :tests:gdx-tests-gwt:warTemplate` and excluding that task causes it to skip compiling. I'll look into this more shortly.
2. It's not fully intended, I guess I had run fetch before in my testing directory.  These are specified in https://github.com/libgdx/libgdx/blob/master/gradle/dependencies.gradle#L123-L135
   I'm not sure it's in scope of this PR to change this, but maybe I can make it automatically fetch. Do you have any suggestions here?
3. I'm not sure why I don't get those errors, but I've excluded those files from the sourceset.

PokeMMO:
https://github.com/steffenschaefer/gwt-gradle-plugin/issues/93

Seems to be a common issue, that I can work around easily enough.

PokeMMO:
You cant remove tasks like that since gradle 6, so I guess we have to put up with it.

https://github.com/steffenschaefer/gwt-gradle-plugin/blob/087a94377fbd83cab938ffffe51ce8fab871be35/gwt-gradle-plugin/src/main/java/de/richsource/gradle/plugins/gwt/GwtEclipsePlugin.java#L67-L68

I don't see an easy way around this unless we stop using the war plugin with gwt.

Tom-Ski:
Is there an up to date replacement we can use? Do we even need it if we aren't packaging?

MrStahlfelge:
Not needed, I suppose it is used for gwt's server side components which we don't use for libgdx.

PokeMMO:
war plugin removed, as far as I know this should resolve any concerns, without breaking any compatibility with `./gradlew tests:gdx-tests-gwt:superDev`

I'm not sure if gwt-tests-gwt can be run from eclipse directly to verify it still functions correctly.

MrStahlfelge (commented):
I can't judge if this works on Eclipse.
With the branch here and in my own project there is no problem removing the war plugin and running or distributing gwt.

PokeMMO:
Rebased to latest master

@NathanSweet If there's no objections, I'm going to merge this soon.

NathanSweet:
I haven't tried it since my last feedback. If that's all resolved, I'm OK with merging.

PokeMMO:
The only thing not addressed is the native jars still rely on fetch.
They are defined here, and changing this is probably not in the scope of this PR.

https://github.com/libgdx/libgdx/blob/master/gradle/dependencies.gradle#L122-L127

I'll merge this and then if you run into any issues, let me know and I'll look into them asap.

NathanSweet:
First error running `gradlew eclipse`:

> Unsupported class file major version 60

After changing `JAVA_HOME` from 16 to 13, gradle then downloaded a lot of things. That took several minutes, seeming to spend most of the time for tests and robovm. Next error:

> A problem occurred configuring project ':extensions:gdx-box2d-parent:gdx-box2d-gwt'.
> Could not create task ':tests:gdx-tests-android:lint'.
> SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at 'C:\Dev\libgdx\local.properties'.

Is the Android SDK necessary to generate Eclipse project files? Is that avoidable when using libgdx only for desktop?

After setting `ANDROID_SDK_ROOT` it generated the Eclipse project files and everything seems to be compiling and working as expected!

PokeMMO:
>After changing JAVA_HOME from 16 to 13, gradle then downloaded a lot of things. That took several minutes, seeming to spend most of the time for tests and robovm. Next error:

We can tell Gradle to download and use (or just use, and disable automatic downloads) of specific JDKs using Gradle's [Toolchain support](https://docs.gradle.org/current/userguide/toolchains.html) 8 might be the minimum supported, but I would have to check.

I can make a PR if you're comfortable with specifying this. Would you want to specify a toolchain only, or enable automatic provision too?

>Is the Android SDK necessary to generate Eclipse project files? Is that avoidable when using libgdx only for desktop?

It is to use gradle at all, due to us having an android subproject. We could optionally/conditionally skip the include of the [android subprojects](https://github.com/libgdx/libgdx/blob/master/settings.gradle#L19) but I really wouldn't recommend it.

NathanSweet:
I think the class file version error message is OK. I mentioned it just in case others have the issue or we want to document somewhere the Java version needed. Downloading a JDK seems a bit extreme.

Re: Android SDK, a CLI parameter to skip it could be useful, but I suppose it's OK as is.

@PokeMMO I ran into an issue: the gdx project doesn't have the gdx natives in the `.classpath` file. Other projects, like gdx-backend-lwjgl, do have the natives in their `.classpath` (but only eg for LWJGL, not for gdx).

PokeMMO:
It looks like only the test projects are given the `gdx-natives.jar`. This seems to make sense to me, as not all platforms (namely iOS) should have a direct dependency on the `gdx-natives.jar`. It also seems like idea users wouldn't have had these natives available on classpath for a while either?

This is easy enough to fix by just adding a new dependency like the ones defined here:
https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/gradle/dependencies.gradle#L122-L127
https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/tests/gdx-tests-lwjgl/build.gradle#L25

Or just having it explicitly add the `gdx-natives.jar` during `.classpath` generation.

What use case do you have for using the `gdx-natives.jar` directly without using a test project?

NathanSweet:
The use case is eg a project depending on `gdx-backend-lwjgl` and using LwjglApplication to run an ApplicationListener. It fails because `getResourceAsStream` returns null in SharedLibraryLoader `readFile` for the gdx native lib.

NathanSweet:
I guess the backends that need `gdx-natives.jar` should have the dependency?

PokeMMO:
#6656 should resolve this

NathanSweet:
I'm seeing the generated `gdx-tests` Eclipse project not having a dependency on `gdx-tools`, which is needed only for BitmapFontWriterTest. FWIW, the `build.gradle` has `compileOnly project(":extensions:gdx-tools")`.

PokeMMO:
> I'm seeing the generated `gdx-tests` Eclipse project not having a dependency on `gdx-tools`, which is needed only for BitmapFontWriterTest. FWIW, the `build.gradle` has `compileOnly project(":extensions:gdx-tools")`.

I think this is just an issue with our handling of this specific test.
`gdx-tests-lwjgl3` depends on `gdx-tests`, while `gdx-tools` depends on `:backends:gdx-backend-lwjgl`

Having both lwjgl2 and lwjgl3 on classpath at runtime is an issue. I think the way past manual projects handled this just never ran into this issue, while gradle's generated classpath would pollute this.

We could manually add a dependency in the generated eclipse `.classpath` to resolve this but I'm not sure that is the best way to handle this.

NathanSweet:
I guess we could do as you said or remove `BitmapFontWriterTest`. We definitely want everything to compile out of the box.

NathanSweet:
As mentioned, the generated `gdx-tests` doesn't have a dependency on `gdx-tools` and so doesn't compile. It DOES have LWJGL2 JARs. If I delete `BitmapFontWriterTest` so it compiles, then `gdx-tests-lwjgl3` gets LWJGL2 on its classpath and `Lwjgl3TestStarter` fails to run:
```
java.lang.SecurityException: sealing violation: can't seal package org.lwjgl: already loaded
```

PokeMMO:
I'll figure out a solution and make a PR with a fix later today. I think I know how to fix it.

PR #6370, opened 2021-01-25
Readd accidently removed <imports> comments for robovm.

PokeMMO:
Whoops, I didn't expect any comments to be removed in #6368


tommyettinger:
Did we figure out if these are needed or not anymore? I'm not sure who to ping about RoboVM things...

Tom-Ski:
These are auto generated comments when running brogen binding generation

PR #6375, opened 2021-01-27
Convert Scaling to a object from enum

PokeMMO:
This allows custom Scaling methods.

We have a need for multiple new custom scaling elements, some of which have "fixed" parameters. These wouldn't be very useful for much of libgdx. This change would allow us to add them ourselves.

Scaling doesn't really need to be an Enum, it's never used in any switch statements, it only is to map String names to Scaling objects.

This is a pretty basic proof of concept, and I'm extremely open to feedback.

TODO: 
* [ ] Verify gwt reflection.


MobiDevelop:
I think calling something that clearly is not an enum should be avoided.

PokeMMO:
That's a good point, and I agree.

NathanSweet:
Agreed. Not sure we need ExtendableEnum. Can just have static fields, like the Interpolation class.

PokeMMO:
Converting it to a class with just static fields has a breaking change in serialization. Custom implementations would still serialize fine, which I was not expecting.

```
static class TestWrapper {
    Scaling scaling;
}
```

Scaling as a regular class would now be serialized as 
`{scaling:{class:com.badlogic.gdx.utils.Scaling$1}}` 

while it used to be
`{scaling:fit}`

I've done this and also added support for serializing/deserializing based on constant names. Deserializing support would allow us to make this change without breaking existing code, but I'm not sure if we should continue to serialize based on names.

Can I get some opinions on this?

PR #6384, opened 2021-01-30
implement loading Tiled Point objects

happy-bracket:
This pull request adds Tiled Point objects (https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#point).

 As of now, TmxMapLoader parses points into rectangles with zero width/height. This behavior is probably fine as it is, but one could argue that it's counter-intuitive and non-explicit, so, here we are.
 
 It's also a breaking change for all the people who rely on that their points are parsed into zero-sized rectangles. So, if this PR has any value, I'm open to suggestions on how to work around that.
 
 TiledMapObjectLoadingTest is updated to reflect a new type of objects.

### DIFF
--- a/gdx/res/com/badlogic/gdx.gwt.xml
+++ b/gdx/res/com/badlogic/gdx.gwt.xml
@@ -300,6 +300,7 @@
 		<include name="maps/objects/PolylineMapObject.java"/>
 		<include name="maps/objects/RectangleMapObject.java"/>
 		<include name="maps/objects/TextureMapObject.java"/>
+        <include name="maps/objects/PointMapObject.java"/>
 	
 	<!-- maps/tiled -->
 		<include name="maps/tiled/AtlasTmxMapLoader.java"/>
--- a/gdx/src/com/badlogic/gdx/maps/objects/PointMapObject.java
+++ b/gdx/src/com/badlogic/gdx/maps/objects/PointMapObject.java
@@ -1,0 +1,29 @@
+package com.badlogic.gdx.maps.objects;
+
+import com.badlogic.gdx.maps.MapObject;
+import com.badlogic.gdx.math.Vector2;
+
+/** @brief Represents 2D points on the map */
+public class PointMapObject extends MapObject {
+
+    private final Vector2 point;
+
+    /** creates a 2D point map object at (0, 0) */
+    public PointMapObject() {
+        this(0, 0);
+    }
+
+    /** Creates a 2D point map object
+     * @param x X coordinate
+     * @param y Y coordinate
+     * */
+    public PointMapObject(float x, float y) {
+        point = new Vector2(x, y);
+    }
+
+    /** @return 2D point on the map as {@link Vector2} */
+    public Vector2 getPoint() {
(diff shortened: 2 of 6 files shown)


SimonIT:
You have to add the `PointMapObject` there for gwt
https://github.com/libgdx/libgdx/blob/3d1f76f574bec71afc8e2e9afa38ec7e83d944f1/gdx/res/com/badlogic/gdx.gwt.xml#L296-L303

happy-bracket:
@SimonIT fixed, thank you

gdx/src/com/badlogic/gdx/maps/objects/PointMapObject.java
```diff
@@ -0,0 +1,29 @@
+package com.badlogic.gdx.maps.objects;
```
mgsx-dev, line 1: could you please add licence header, you can find it in any other libgdx java file.

mgsx-dev (changes requested):
Could you please fix formatting, indentation seams off. You can use provided formatter, see [contribution guidelines](https://libgdx.com/dev/contributing/).
You also need to update CHANGES file with this API addition.

Apart from these details, it looks good to me.

gdx/src/com/badlogic/gdx/maps/objects/PointMapObject.java
happy-bracket: PR updated

gdx/res/com/badlogic/gdx.gwt.xml
```diff
@@ -300,6 +300,7 @@
 		<include name="maps/objects/PolylineMapObject.java"/>
 		<include name="maps/objects/RectangleMapObject.java"/>
 		<include name="maps/objects/TextureMapObject.java"/>
+        <include name="maps/objects/PointMapObject.java"/>
```
mgsx-dev, line 303: could you please fix formatting here as well :-)

mgsx-dev:
I think it's good to have it but considering it's a breaking change without much value for now, i propose to merge it when we will implement some other big changes related to Tiled 1.x (eg. infinite map support, template support) and/or a new major version of libGDX.

BoBIsHere86:
Since the latest libGDX release. I'm assuming we have a year till the next one. I've been going through tiled PRs starting with the oldest. Just to give some update on the current status. I pulled down this change, built it locally and despite being over 3 years old and with many changes to the map codebase since then. It rebased without any issue and the tests still ran perfectly fine. 
So for what it's worth. This one still looks good to go.

*One thing to note, and this goes for a lot of the tiled PRs stacking up. Depending on the order in which they get approved and merged. Certain ones may need to be modified after the fact.

Ex. If the Tiledmap JSON support PR were merged before this one. This would need to be updated to be placed into the loadObject method for the json map load codebase or vice versa.

BoBIsHere86:
@Tom-Ski Just a reminder, You recently merged PR #7523, which was meant to replace this outdated PR.
We can now freely close out this PR as it is no longer relevant, a nice two for one PR close! 👍

PR #6386, opened 2021-01-31
Split GWT reflection cache into two generated classes

TheLogicMaster:
Not sure if this is the best solution to the issue, but I constantly find myself running into the "too many constants" error while compiling for GWT, so this is a "dumb" solution that splits the collected classes in half and allocates them to separate ReflectionCacheSourceCreator instances. The ReflectionCache class, in turn, proxies to both generated caches while first searching for a class then stores a reference to that cache in the created Type for later invocations.


piotr-j:
Very useful! Any change it could be made dynamic? Perhaps extra setting to set number of caches.

TheLogicMaster:
@piotr-j 
I don't think it is possible to arbitrarily generate caches due to limitations with GWT deferred binding. As far as I know, a dedicated interface is required for each generated class. In addition, to actually instantiate one of the caches at runtime, a class has to be hard coded into the `GWT.create(IReflectionCache2.class);` function calls. This would mean that we could make the code a bit more dynamic to support 3 caches, for instance, but the maximum number would need to be set in code by creating that many interfaces in actual source files, which would probably negate the usefulness of such a feature completely, since there shouldn't be much of a performance difference between having three caches and one cache, since the only slight difference in functionality occurs when acquiring a Class. Instead of hashing the class name and using it in an enormous switch statement once, we hash the name n times and use it in n switch statements that are (1/n)th the size (though if it really mattered we could re-use the hash for all of the caches).

[GWT Deferred Binding Documentation](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsDeferred.html)

piotr-j:
Ah thats a shame! If there is no significant its probably fine as is. Perhaps make it 3 caches just in case?

TheLogicMaster:
Honestly, I think 2 caches are all that is needed. I just did a test by creating a project with Gdx-liftoff and included every GWT compatible library available. As expected, with a single cache it failed to compile, but with 2 caches, both generated caches were a little under 34000 lines. Since you normally have to reach over 60000 lines, give or take a bit, two should be sufficient.

MobiDevelop:
It's possible to generate multiple source files from the generator and make it dynamic, such that it only creates multiple files if needed, might not be worth the hassle but certainly possible to do.

TheLogicMaster:
That does seem to be the case. I had assumed it wasn't possible because the generate function just returned a single string with the name of the generated class, but looking at the JavaDocs for that function does seem to indicate that it isn't required to do that. It looks like the only way to accomplish something like that would be having the GWT.create function return a factory of some sort to return the different caches. 
It does look like it's overkill, though, since two caches should be enough for just about anyone. 
[Stack Overflow Answer](https://stackoverflow.com/a/4178673)

TheLogicMaster:
I figured out how to do the dynamic generation and got some working code to do so, so it wouldn't be too difficult to implement if we wanted the functionality. It could be specified via GWT config properties to add any number of caches and equally split the classes between them. Might not be worth the effort, though, since two should be enough and doing it this way probably adds unnecessary complexity. There seems to be dead code in ReflectionCacheSourceCreator class, though, which made it seem a bit more complicated than it was. Or, at the very least, I can't tell what it is supposed to do and removing it seems to have no effect.
Here's an extremely rough test that only serves to demonstrate that the dynamic generation works: https://github.com/TheLogicMaster/libgdx/commit/cf7c5c6984ec4a7c1d335d0370e4b4ae1ad7f4de

MobiDevelop:
Yeah, there is dead code in there. I used the createProxy method to test generating multiple files and consuming them before making the last comment. It likely is more complexity than it is worth.

tommyettinger:
I'm so glad this GWT limitation can be overcome, and without any really hairy code, too. I am not 100% sure if just selecting libraries in liftoff will cause the reflection cache to be expanded in the generated project, but that would make sense because the .gwt.xml files with the include-into-the-reflection-cache instructions probably get evaluated regardless of whether the code is used. I mean, the 3D model parts of libGDX's reflection cache get added even for purely 2D games.

piotr-j:
Perhaps the default config for included types could be split?
So it could be in user xml config:
```xml
	<inherits name='com.badlogic.gdx.backends.gdx_backends_gwt' />
	<inherits name='com.badlogic.gdx.backends.gdx_backends_gwt_tiled' />
	<inherits name='com.badlogic.gdx.backends.gdx_backends_gwt_2d' />
	<inherits name='com.badlogic.gdx.backends.gdx_backends_gwt_3d' />
```
Ability to exclude entire folders/packages would be an alternative of sorts, but that is probably harder to do.

TheLogicMaster:
I feel like inheriting gwt files like that is a better option than excluding packages, since it allows categorizing the libgdx systems beyond their containing packages and is more elegant. It sounds like the project should be generated with all of the subsystems inherited by default and you remove any ones you don't need to hopefully reduce the build size but still have it build by default. The only negative effect of having additional cache entries, assuming we are going to split it in half, should be an increase in load times. I haven't tested the difference, but the built cache easily exceeds 6MB before being obfuscated, so that could be a real concern. The runtime performance difference of adding more cached entries should be negligible since everything is nicely hashed.

TheLogicMaster:
I did a small test to see if minimizing the reflection cache would be beneficial in terms of build size and it looks like it would be a good idea to categorize the includes by subsystem. By removing all of the gdx packages from the cache, around 1.5MB was saved from the build, which could make a substantial difference for load times if Scene2d or 3D isn't needed, for instance.

crykn:
Splitting the libGDX module file sounds like a good idea, but a quick google search didn't bring anything up on how I would actually go about excluding any modules.  Regardless, I think this discussion is beyond the scope of this PR and should be split into its own issue.

Back to the original PR: are there any opinions on whether or not to merge this?

tommyettinger:
Merge! Merge! Merge!

I think just the fact that this makes a previously-impossible build failure, and now makes it **effortless** for users to get building, is great. We can worry about splitting up the GWT modules in libGDX in another commit, since that might be trickier than it seems (some 3D-related code is used even by 2D code, and UIUtils in scene2d.ui is used in other places).

TheLogicMaster:
Since there's talk of actually merging it, I cleaned up a copied helper function from when I didn't think splitting it would actually work.

tommyettinger (approved):
Just looking through the source here, this is an ingenious solution. IReflectionCache2 can act as an IReflectionCache by extending it, but doesn't force the 64K fields to be repeated in the `2` class, and it can have its own many fields. Approving again.

tommyettinger:
It looks like this PR was half-forgotten in the process of releasing 1.10.0 . Does anyone else want to review? @MrStahlfelge , who uses GWT heavily and is an expert on the topic? I think this is a pretty important step towards mitigating one of the harshest limitations of GWT, effectively "Game scale cannot exceed X", where X is tied to the reflection cache.

Tom-Ski:
We can ping @MobiDevelop. Or if you have any other guys that do use this extensively please ping also for a review. If @tommyettinger says its good, then its good. Are there any limitations or problems with this approach that we need to be worried about or warn people about?

tommyettinger:
Well no one has ever hit the limitation of 2 whole ReflectionCaches being exhausted, so I'm going to try to do that, and see what it takes.

tommyettinger:
So, I can confirm that this does have issues when given a very very large pool of constants. This 2-part pool is actually exhaustible if you just stick whole, large, packages in the reflection cache without rhyme or reason.

I tried adding the [latest squidlib and squidlib-util](https://jitpack.io/#yellowstonegames/SquidLib/26b81caa7c), large libraries, in full to the cache. Of course, this is a terrible idea, but I'm testing.
```xml
<!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit 2.8.2//EN"
        "http://www.gwtproject.org/doctype/2.8.2/gwt-module.dtd">
<module rename-to="html">
    <source path="" />
    <inherits name="com.badlogic.gdx.backends.gdx_backends_gwt" />
    <inherits name="com.github.tommyettinger.demos.TsarDemo" />
    <inherits name="regexodus" />
    <inherits name="squidlib" />
    <inherits name="squidlib-util" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidmath" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidgrid" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidai" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidgrid.mapping" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidgrid.mapping.styled" />
    <extend-configuration-property name="gdx.reflect.include" value="squidpony.squidgrid.gui.gdx" />
    <entry-point class="com.github.tommyettinger.demos.gwt.GwtLauncher" />
    <set-configuration-property name="gdx.assetpath" value="../assets" />
    <set-configuration-property name="xsiframe.failIfScriptTag" value="FALSE"/>
    <set-property name="user.agent" value="gecko1_8, safari"/>
    <collapse-property name="user.agent" values="*" />
</module>
```
On the first `dist`, I get this:
```
      Tracing compile failure path for type 'com.badlogic.gwtref.client.IReflectionCache2Generated'
         [ERROR] Errors in 'generated://C12ED3E00A94C3FA603D417F1D3DAC89/com/badlogic/gwtref/client/IReflectionCache2Generated.java'
            [ERROR] Line 7: Too many constants, the constant pool for IReflectionCache2Generated would exceed 65536 entries
   [ERROR] Errors in 'generated://C12ED3E00A94C3FA603D417F1D3DAC89/com/badlogic/gwtref/client/IReflectionCache2Generated.java'
      [ERROR] Line 7: Too many constants, the constant pool for IReflectionCache2Generated would exceed 65536 entries
   Tracing compile failure path for type 'com.badlogic.gwtref.client.IReflectionCache2Generated'
      [ERROR] Errors in 'generated://C12ED3E00A94C3FA603D417F1D3DAC89/com/badlogic/gwtref/client/IReflectionCache2Generated.java'
         [ERROR] Line 7: Too many constants, the constant pool for IReflectionCache2Generated would exceed 65536 entries
   [ERROR] Hint: Check that the type name 'com.badlogic.gwtref.client.IReflectionCache2Generated' is really what you meant
   [ERROR] Hint: Check that your classpath includes all required source roots
   [ERROR] Errors in 'com/badlogic/gwtref/client/ReflectionCache.java'
      [ERROR] Line 23: Rebind result 'com.badlogic.gwtref.client.IReflectionCache2Generated' could not be found
```
Even weirder, when I set the user-agent as gdx-setup currently does, and re-run `dist`, I get this bizarre compiler crash:
```
   [ERROR] An internal compiler exception occurred
com.google.gwt.dev.jjs.InternalCompilerException: Unexpected error during visit.
	at com.google.gwt.dev.jjs.ast.JVisitor.translateException(JVisitor.java:111)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:276)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:265)
	at com.google.gwt.dev.jjs.ast.JVisitor.accept(JVisitor.java:118)
	at com.google.gwt.dev.jjs.ast.JCastOperation.traverse(JCastOperation.java:76)
	at com.google.gwt.dev.jjs.ast.JModVisitor.traverse(JModVisitor.java:361)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:273)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:265)
	at com.google.gwt.dev.jjs.ast.JVisitor.accept(JVisitor.java:118)
	at com.google.gwt.dev.jjs.ast.JDeclarationStatement.traverse(JDeclarationStatement.java:49)
	at com.google.gwt.dev.jjs.ast.JModVisitor$ListContext.traverse(JModVisitor.java:88)
	at com.google.gwt.dev.jjs.ast.JModVisitor.acceptWithInsertRemove(JModVisitor.java:331)
	at com.google.gwt.dev.jjs.ast.JBlock.traverse(JBlock.java:94)
	at com.google.gwt.dev.jjs.ast.JModVisitor.traverse(JModVisitor.java:361)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:273)
	at com.google.gwt.dev.jjs.ast.JVisitor.accept(JVisitor.java:139)
	at com.google.gwt.dev.jjs.ast.JVisitor.accept(JVisitor.java:135)
	at com.google.gwt.dev.jjs.ast.JMethodBody.traverse(JMethodBody.java:83)
	at com.google.gwt.dev.jjs.ast.JModVisitor.traverse(JModVisitor.java:361)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:273)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:265)
	at com.google.gwt.dev.jjs.ast.JMethod.visitChildren(JMethod.java:786)
	at com.google.gwt.dev.jjs.ast.JMethod.traverse(JMethod.java:778)
	at com.google.gwt.dev.jjs.ast.JModVisitor.traverse(JModVisitor.java:361)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:273)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:265)
	at com.google.gwt.dev.jjs.impl.UnifyAst.mainLoop(UnifyAst.java:1401)
	at com.google.gwt.dev.jjs.impl.UnifyAst.exec(UnifyAst.java:896)
	at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.unifyJavaAst(JavaToJavaScriptCompiler.java:1410)
	at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.constructJavaAst(JavaToJavaScriptCompiler.java:1222)
	at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.precompile(JavaToJavaScriptCompiler.java:1140)
	at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.precompile(JavaToJavaScriptCompiler.java:255)
	at com.google.gwt.dev.Precompile.precompile(Precompile.java:255)
	at com.google.gwt.dev.Precompile.precompile(Precompile.java:202)
	at com.google.gwt.dev.Precompile.precompile(Precompile.java:143)
	at com.google.gwt.dev.Compiler.compile(Compiler.java:204)
	at com.google.gwt.dev.Compiler.compile(Compiler.java:155)
	at com.google.gwt.dev.Compiler.compile(Compiler.java:144)
	at com.google.gwt.dev.Compiler$1.run(Compiler.java:118)
	at com.google.gwt.dev.CompileTaskRunner.doRun(CompileTaskRunner.java:55)
	at com.google.gwt.dev.CompileTaskRunner.runWithAppropriateLogger(CompileTaskRunner.java:50)
	at com.google.gwt.dev.Compiler.main(Compiler.java:125)
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0
	at org.objectweb.asm.ClassReader.readClass(Unknown Source)
	at org.objectweb.asm.ClassReader.accept(Unknown Source)
	at org.objectweb.asm.ClassReader.accept(Unknown Source)
	at com.google.gwt.dev.javac.BytecodeSignatureMaker.visitCompileDependenciesInBytecode(BytecodeSignatureMaker.java:228)
	at com.google.gwt.dev.javac.BytecodeSignatureMaker.getCompileDependencySignature(BytecodeSignatureMaker.java:209)
	at com.google.gwt.dev.javac.CompiledClass.getSignatureHash(CompiledClass.java:165)
	at com.google.gwt.dev.javac.Dependencies$Ref.<init>(Dependencies.java:41)
	at com.google.gwt.dev.javac.Dependencies$Ref.<init>(Dependencies.java:36)
	at com.google.gwt.dev.javac.Dependencies.resolve(Dependencies.java:100)
	at com.google.gwt.dev.javac.CompilationStateBuilder$CompileMoreLater.compile(CompilationStateBuilder.java:352)
	at com.google.gwt.dev.javac.CompilationStateBuilder.doBuildGeneratedTypes(CompilationStateBuilder.java:638)
	at com.google.gwt.dev.javac.CompilationStateBuilder$CompileMoreLater.addGeneratedTypes(CompilationStateBuilder.java:228)
	at com.google.gwt.dev.javac.CompilationState.addGeneratedCompilationUnits(CompilationState.java:127)
	at com.google.gwt.dev.javac.StandardGeneratorContext.finish(StandardGeneratorContext.java:551)
	at com.google.gwt.dev.jjs.impl.UnifyAst$UnifyVisitor.createStaticRebindExpression(UnifyAst.java:526)
	at com.google.gwt.dev.jjs.impl.UnifyAst$UnifyVisitor.createRebindExpression(UnifyAst.java:487)
	at com.google.gwt.dev.jjs.impl.UnifyAst$UnifyVisitor.maybeHandleMagicMethodCall(UnifyAst.java:415)
	at com.google.gwt.dev.jjs.impl.UnifyAst$UnifyVisitor.visit(UnifyAst.java:402)
	at com.google.gwt.dev.jjs.ast.JMethodCall.traverse(JMethodCall.java:265)
	at com.google.gwt.dev.jjs.ast.JModVisitor.traverse(JModVisitor.java:361)
	at com.google.gwt.dev.jjs.ast.JModVisitor.accept(JModVisitor.java:273)
	... 40 more
      [ERROR] at ReflectionCache.java(23): GWT.create(IReflectionCache2.class)
      [ERROR] at ReflectionCache.java(23): (IReflectionCache) GWT.create(IReflectionCache2.class)
      [ERROR] at ReflectionCache.java(23): static IReflectionCache instance2 = (IReflectionCache) GWT.create(IReflectionCache2.class)
      [ERROR] at ReflectionCache.java(21): {
  Object.$clinit();
  static IReflectionCache instance1 = (IReflectionCache) new IReflectionCacheGenerated();
  static IReflectionCache instance2 = (IReflectionCache) GWT.create(IReflectionCache2.class);
}
      [ERROR] at ReflectionCache.java(21): {
  Object.$clinit();
  static IReflectionCache instance1 = (IReflectionCache) new IReflectionCacheGenerated();
  static IReflectionCache instance2 = (IReflectionCache) GWT.create(IReflectionCache2.class);
}
      [ERROR] at ReflectionCache.java(21): com.badlogic.gwtref.client.ReflectionCache.$clinit()V
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0


> Task :html:compileGwt FAILED
```

So, I'm not sure what this means, but it is certainly possible to exhaust both 65536-item caches.

DaanVanYperen:
A perspective of a third party lib: https://github.com/junkdog/artemis-odb depends heavily on reflection for configuration.  I personally hit the 65535 limit with gdx+odb jam games fairly often. GDX+odb libs fill a decent part of it, so it easy to blow over the limit by including too much by accident. 

Double the cache would fit most games, so why not start with that? It would also help if developers could spot scoping mistakes more easily during compilation. The current reflection cache generation logging could report on scope when the limit is hit for example.

(As a side note, we actually have run a copy of the GDX reflection logic in odb for our non-GDX users. We'd love for the reflection logic to be available as a separate package so our libGDX users save some memory).

MrStahlfelge:
All we need is some people actually testing. The only approval here is "just looking through the source here", that's not enough to merge. Given the fact that there were often complaints about the problem, I wonder why no one fired up their IDE to test for nearly 5 months now.

tommyettinger:
> All we need is some people actually testing. The only approval here is "just looking through the source here", that's not enough to merge. Given the fact that there were often complaints about the problem, I wonder why no one fired up their IDE to test for nearly 5 months now.

Hi! Did you see https://github.com/libgdx/libgdx/pull/6386#issuecomment-848655518 ? I actually did test! I found that it is still possible to exhaust the pool. I don't have any games that are pushing the limit of the used reflection pool, so I'm not sure what more you want.

MrStahlfelge:
So lang ago, yes, now I remember that this was the reason why this is stalled. I missed it scrolling up because it is not a review.
So, looks like it is not only testing, but actually doing :)

tommyettinger (approved):
I tested this with a probably-normal-sized cache here, using a locally-built libGDX with this PR: https://tommyettinger.github.io/CaveCops/pr6386/index.html

I also tested it with the same code and without this PR here, for reference:
https://tommyettinger.github.io/CaveCops/index.html

The results are pretty much identical. There seems to be a difference of 20 to 30 kilobytes in total file size, which could simply be the added code in 1.10.1-SNAPSHOT relative to 1.10.0.

I don't see any problems, but I also don't have any code where the cache size is more than 1.10.0 can handle, and less than what this PR is limited to (which is still a lot). You can [poke around the built files here](https://github.com/tommyettinger/CaveCops/tree/master/docs) if you want to compare the output more carefully, but I'm pleased with this. Approved.

backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/ReflectionCache.java
```diff
@@ -19,22 +19,40 @@
 import com.google.gwt.core.client.GWT;
 
 public class ReflectionCache {
-	private static IReflectionCache instance = GWT.create(IReflectionCache.class);
+	private static IReflectionCache instance1 = GWT.create(IReflectionCache.class);
+	private static IReflectionCache instance2 = GWT.create(IReflectionCache2.class);
 
 	public static Type forName (String name) throws ClassNotFoundException {
```
DaanVanYperen, line 25: Has a throws clause for `ClassNotFoundException` but will throw a `RuntimeException` on class not found. Perhaps remove the clause?

backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheGenerator.java
```diff
@@ -16,31 +16,207 @@
...
 		}
 
-		if (type.isInterface() == null) {
+		if (jType.isInterface() == null) {
 			logger.log(ERROR, "Type '" + typeName + "' must be an interface");
 			throw new UnableToCompleteException();
 		}
 
-		ReflectionCacheSourceCreator source = new ReflectionCacheSourceCreator(logger, context, type);
+		TypeOracle typeOracle = context.getTypeOracle();
```
DaanVanYperen, line 64: level of abstraction, would be easier to grok if type collection had it's own class or encapsulated in method.

backends/gdx-backends-gwt/src/com/badlogic/gwtref/gen/ReflectionCacheGenerator.java
```diff
@@ -16,31 +16,207 @@
...
+		gatherTypes(typeOracle.findType("java.lang.Object").getErasedType(), types);
+
+		// sort the types so the generated output will be stable between runs
+		Collections.sort(types, new Comparator<JType>() {
+			public int compare (JType o1, JType o2) {
+				return o1.getQualifiedSourceName().compareTo(o2.getQualifiedSourceName());
+			}
+		});
+
+		List<JType> split;
```
DaanVanYperen, line 108: Could use a line of background.
`// Split types between two classes as a workaround for the java 65535 const limit. `

DaanVanYperen (approved):
Some minor suggestions but nothing major. 
Merged most of this into odb's version of libgdx reflection and it passes our reflection dependent GWT test suite.

backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/ReflectionCache.java
tommyettinger: Good catch, this should probably use ClassNotFoundException for when a Class is not found.

MrStahlfelge (approved):
Tested ReflectionTest and ReflectionCorrectnessTest sucessfully. 

@DaanVanYperen I suggest you make a new PR with your suggestions, looks like @TheLogicMaster is not around at the moment.

backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/ReflectionCache.java
TheLogicMaster: Removing ClassNotFoundException from the signature would probably force ClassReflection.forName to catch and throw runtime exceptions as Reflection exceptions, which seems more awkward. Throwing a ClassNotFoundException would make it inconsistent with getType, which can't be easily refactored without changing the public-facing API.

tommyettinger: Great, if this matches the public API then it should need no changes. I'm happy if this doesn't change anything for users.

tommyettinger:
I was wondering why this showed me as a contributor to this PR -- I think I just merged the current commits into the version from March, and I didn't realize that would affect the actual PR. I did that to evaluate what it would take to exhaust the cache, and merging to the current version makes sense for that test, but I need to figure out how my Git configuration is set up so I don't accidentally affect other PRs...

backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/ReflectionCache.java
DaanVanYperen: Ok I'll skip the PR. Most reflection issues fail fast and on in libGDX/GWT context typically are developer errors, so any exception works.

Just documenting for GWT users who stumble upon this ticket: GWT `ClassReflection.forName` throws a RuntimeException, on desktop it throws ReflectionException.

DaanVanYperen:
> @DaanVanYperen I suggest you make a new PR with your suggestions, looks like @TheLogicMaster is not around at the moment.

I'm happy this got merged! Given that the API doesn't need fixing (see above) I'll save you guys the effort of reviewing a PR that just improves readability.

MrStahlfelge:
> I was wondering why this showed me as a contributor to this PR -- I think I just merged the current commits into the version from March, and I didn't realize that would affect the actual PR.

Yes, it is because of your merge commit.

PR #6388, opened 2021-02-01
Add a ParticleEffectActor

MrStahlfelge:
This adds a simple `ParticleEffectActor`. One could discuss if it is useful to show particle effects in Scene2d, but searching on GitHub for ParticleEffectActor shows that there is a need for such an actor in projects. Menus should look good, too, and we have some games completely built on scene2d.

Most ParticleEffectActors out there are inspired by a code from StackOverflow and update the particle effect in `update()` by converting the local coordinates to stage coordinates. This does not work if the ParticleEffectActor is member of a transformed group. The implementation here handles the update in draw(), because only in draw the position is set correctly by the parent.

I am sure this actor will not satisfy everyone's needs and has to be completed in the future. But it is a start.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
@@ -62,8 +62,9 @@ public class ParticleEffectActor extends Actor implements Disposable {
 
     public void start() {
         isRunning = true;
-        if (resetOnStart)
-            particleEffect.reset();
+        if (resetOnStart) {
+            particleEffect.reset(false);
+        }
         particleEffect.start();
     }
 
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
@@ -18,6 +18,7 @@ public class ParticleEffectActor extends Actor implements Disposable {
     protected boolean isRunning;
     protected boolean ownsEffect;
     private boolean resetOnStart;
+    private boolean autoRemove;
 
     public ParticleEffectActor(ParticleEffect particleEffect, boolean resetOnStart) {
         super();
@@ -58,6 +59,10 @@ public class ParticleEffectActor extends Actor implements Disposable {
         // don't do particleEffect.update() here - the correct position is set  just while we
         // are in draw() method. We save the delta here to update in draw()
         lastDelta += delta;
+
+        if (autoRemove && particleEffect.isComplete()) {
+            remove();
+        }
     }
 
     public void start() {
@@ -72,8 +77,18 @@ public class ParticleEffectActor extends Actor implements Disposable {
         return resetOnStart;
     }
 
-    public void setResetOnStart(boolean resetOnStart) {
+    public ParticleEffectActor setResetOnStart(boolean resetOnStart) {
         this.resetOnStart = resetOnStart;
+        return this;
+    }
+
+    public boolean isAutoRemove() {
+        return autoRemove;
+    }
+
+    public ParticleEffectActor setAutoRemove(boolean autoRemove) {
+        this.autoRemove = autoRemove;
+        return this;
     }
 
     public boolean isRunning() {
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
@@ -1,0 +1,96 @@
+package com.badlogic.gdx.scenes.scene2d.ui;
+
+import com.badlogic.gdx.files.FileHandle;
+import com.badlogic.gdx.graphics.g2d.Batch;
+import com.badlogic.gdx.graphics.g2d.ParticleEffect;
+import com.badlogic.gdx.graphics.g2d.TextureAtlas;
+import com.badlogic.gdx.scenes.scene2d.Actor;
+
+/**
+ * ParticleEffectActor holds an {@link ParticleEffect} to use in Scene2d applications.
+ * The particle effect is positioned in the centered in the ParticleEffectActor. Its bounding box
+ * is not limited to the size of this actor.
+ */
+public class ParticleEffectActor extends Actor {
+    private final ParticleEffect particleEffect;
+    float lastDelta;
+    boolean isComplete;
+    private boolean resetOnStart;
+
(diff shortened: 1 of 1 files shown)


TCreutzenberg:
You could already call "preAllocateParticles()" after loading the ParticleEffect I guess.

raeleus (approved):
It's simple and it works reliably. It's good enough to add a little sparkle to a user interface. If a user wanted something more advanced, they can just build on this source.

obigu (approved):
Looks good to me, just a couple of things you may consider.
1. I don't think setting `width` and `height` apply in this case. Considering they only affect positioning and may lead to confusion, I would get rid of it and keep it simple `x` and `y`.
2. In my implementation I have a property that enables auto removing the `Actor` from the `Stage` when the `ParticleEffect` is `complete`, I think it's pretty useful.

piotr-j:
Good idea, but can we not make new classes impossible to extend with private/package private fields?

MrStahlfelge:
Width and height might be useful if the Actor is added to a Table. This way, it gets the set size reserved without the need to wrap another Group around.

The idea with removing the Actor optionally is good, do you want to add your enhancements after a first merge of this one or right here on this branch?

MrStahlfelge:
> Good idea, but can we not make new classes impossible to extend with private/package private fields?

I am all for extensibility, the final variables have getters. The two variables that are not accessible by subclasses are helper variables, do you think it makes sense to tweak them?

piotr-j:
Getter are nice for composition, but when I extend a class I usually want to change how it does something without ending up with `ParticleEffectActor2`. Which is par the course for most `Actors`, so checks out I guess ;)

obigu:
> Width and height might be useful if the Actor is added to a Table. This way, it gets the set size reserved without the need to wrap another Group around.

Unless we implement `Layout` to meet the requirements to participate in one I think we should not worry about as it will not act as expected in many cases when added to a `Table`.

> The idea with removing the Actor optionally is good, do you want to add your enhancements after a first merge of this one or right here on this branch?

Well, I was hoping you would add it to this PR if you liked it :)

MrStahlfelge:
Okay, but it doesn't make much sense to change the inner ParticleEffect, does it? I would use another Actor when I had the need to change the effect.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
```diff
@@ -0,0 +1,100 @@
...
+import com.badlogic.gdx.graphics.g2d.ParticleEffect;
+import com.badlogic.gdx.graphics.g2d.TextureAtlas;
+import com.badlogic.gdx.scenes.scene2d.Actor;
+
+/**
+ * ParticleEffectActor holds an {@link ParticleEffect} to use in Scene2d applications.
+ * The particle effect is positioned in the centered in the ParticleEffectActor. Its bounding box
+ * is not limited to the size of this actor.
+ */
+public class ParticleEffectActor extends Actor {
```
SimonIT, line 14: Maybe implement the disposable interface for the dispose method?

MrStahlfelge:
> Unless we implement `Layout` to meet the requirements to participate in one I think we should not worry about as it will not act as expected in many cases when added to a `Table`.

Okay, will remove it. Positioning at the center might not be the expected behaviour anyway

> Well, I was hoping you would add it to this PR if you liked it :)

I like the idea but if you already have a working implementation I'd prefer not to reinvent the wheel ;)

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
obigu: Regarding the `dispose` implementation, I think it only makes sense in case it owns the `ParticleEffect`, that is if it's instantiated by the `ParticleEffectActor` (second and third constructors).

obigu:
> I like the idea but if you already have a working implementation I'd prefer not to reinvent the wheel ;)
 Inside `act()`
```
if (particleEffect.isComplete())
            remove();
```

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
```diff
@@ -0,0 +1,100 @@
...
+    private final ParticleEffect particleEffect;
+    float lastDelta;
+    private boolean isComplete = true;
```
obigu, line 17: Shouldn't it be `false`?

MrStahlfelge: isComplete is false while the effect is running in this implementation, and it does not autostart

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
MrStahlfelge: It makes sense to call it if the ParticleEffect is just instantiated for this Actor and not referenced elsewhere, because some other ways to initialize it were needed. It should be called manually, and of course `getParticleEffect().dispose()` could used, too. I am open to remove it if others think it doesn't make sense here.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
obigu: yep, I think there are 3 different states, "Not started", "Running" and "Complete", we want the property to tells us if it's "Running".

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/ParticleEffectActor.java
obigu: Yes, like `Stage` `ownsBatch` we could have `ownsParticleEfffect` that would be `true` only when `ParticleEffect` is not provided but created internally.

raeleus:
> In my implementation I have a property that enables auto removing the Actor from the Stage when the ParticleEffect is complete, I think it's pretty useful.

In that case, can we have an overridable method that is called when the particle effect is complete? In that case, the user can trigger a remove, a new particle effect, loop, etc.

obigu:
> In that case, can we have an overridable method that is called when the particle effect is complete? In that case, the user can trigger a remove, a new particle effect, loop, etc.

Not sure about that. For looping the `ParticleEffect`, `continuous` can be set to `true`. Triggering new particle effects from a ParticleEffectActor when is complete or any other actions seems a very specific requirement that can be easily achieved by extending the class.

MrStahlfelge:
I think I've processed all requests

PR #6389, opened 2021-02-02
FloatCounter, add tostring and implement poolable

fabiitch:
At start i just wanted to add toString method on FloatCounter

And i see the reset method,  so im think 'why not poolable' .

Open to remove the implements Poolable, i just need the toString ^^

Thanks for reading

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/FloatCounter.java
+++ b/gdx/src/com/badlogic/gdx/math/FloatCounter.java
@@ -16,11 +16,13 @@
 
 package com.badlogic.gdx.math;
 
+import com.badlogic.gdx.utils.Pool.Poolable;
+
 /** Track properties of a stream of float values. The properties (total value, minimum, etc) are updated as values are
  * {@link #put(float)} into the stream.
  * 
  * @author xoppa */
-public class FloatCounter {
+public class FloatCounter implements Poolable {
 	/** The amount of values added */
 	public int count;
 	/** The sum of all values */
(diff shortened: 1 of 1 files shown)


MrStahlfelge (approved):
I don't see a problem here. One could discuss if the toString() method is needed here but it is handy for debugging purposes so why not.

crykn (approved):

gdx/src/com/badlogic/gdx/math/FloatCounter.java
```diff
@@ -76,4 +79,18 @@ public void reset () {
...
+	public String toString() {
+		return "FloatCounter{" +
+				"count=" + count +
+				", total=" + total +
+				", min=" + min +
+				", max=" + max +
+				", average=" + average +
+				", latest=" + latest +
+				", value=" + value +
+				", mean=" + mean +
```
obigu, line 93: Can we remove `mean`? `value` already holds the `mean` value.

fabiitch: done

MobiDevelop:
I'm not going to lie, it feels like an odd type to be pooling - so not sure about adding the Poolable interface (even if it happens to have a reset method).

obigu (approved):
Seems good to me. I don't see an issue with adding the `Poolable` interface even if it's probably going to be rarely pooled.

NathanSweet:
This isn't super important nitpick, but we'd not like this to run away from us and make lots of classes poolable. :)

PR #6390, opened 2021-02-03
Fix getIndices() crash in Mesh

tommyettinger:
This was probably a copy/paste error, but the getIndices() code mistakenly changed the vertices buffer where it formerly changed only the indices buffer. This caused crashes in ModelCacheTest when the Cache checkbox was selected and any model was added.

This crash was found by Agueliethun (mentioned on Discord) initially when the same model was added twice to the cache, but it affects more than just that case, as can be seen in ModelCacheTest. Agueliethun went through the recent history of Mesh.java and found that in the Buffer casts change, the calls to change the position of the indices buffer instead changed the position of the vertices buffer, which doesn't seem related in getIndices(). This bug affects 1.9.14 only, as far as I can tell.


SimonIT (approved):
My bad, I went thru the changes multiple times to prevent such copying mistakes but it seems that I still overlooked it

crykn (approved):

tommyettinger:
I'm guessing parts of the issue are that both a) no one tested a 3D project that used ModelCache on the 1.9.14-SNAPSHOT, and b) even if using 1.9.14 and running the tests, the ModelCacheTest doesn't actually use the cache until a checkbox is selected, so just "going through each relevant test, seeing if it crashes, trying the next" doesn't find the crash. There may be another test that does use ModelCache right away, though it apparently didn't crash (or wasn't run) either. I didn't personally test 1.9.14-SNAPSHOT because I don't have a game project that would appropriately test it right now. I think right now the best advice is to stay on 1.9.13 if doing 3D code, or use the nightly after this is merged.

PR #6391, opened 2021-02-03
Convert Scaling from an Enum to an Object.

PokeMMO:
This allows custom Scaling methods.

We have a need for multiple new custom scaling elements, some of which have "fixed" parameters. These wouldn't be very useful for much of libgdx. This change would allow us to add them ourselves.

Scaling doesn't really need to be an Enum, it's never used in any switch statements, it only is to map String names to Scaling objects.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.9.15]
+- [BREAKING CHANGE] API Change: Scaling is now an object instead of an enum. This may change behavior when used with serialization.
 - Scene2d.ui: Added new ParticleEffectActor to use particle effects on Stage
 
 [1.9.14]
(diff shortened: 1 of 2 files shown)


obigu (approved):
I've never had the need to add a custom scaling method but I think it adds extensibility and I can't see any drawback.

Just a personal preference but you could make it a bit more compact by inlining the 3 lines into 1 in each `apply()` implementation like this:
`return temp.set(x, y);`

raeleus:
This is something I've wanted actually.

mgsx-dev:
It looks like a breaking change since serialization was removed, no? if so, we need to update CHANGES file.

PokeMMO:
Yeah, it seems like Scaling is never really serialized, but we could readd serialization based on constant names if we wanted to. #6375 did this, but since scaling is never serialized in generic libgdx, I'm not sure if it's needed.

I was going to add CHANGES right before/after it was merged.

mgsx-dev:
i think it's fine like this, probably only a few people (or maybe no one) serialize it. It would be better to update CHANGES within this PR though, if you don't mind.

mgsx-dev:
sorry to be nit picky but i think it would be great to explain how it could break user code, this CHANGES file is useful for users to know what they should take care of and how to migrate their code, maybe you can add few words about serialization?

PokeMMO:
Sure, I've updated the CHANGES entry.

PR #6393, opened 2021-02-04
jcenter is shutting down, replace by Gradle Plugins

MrStahlfelge:
jcenter is shutting down in less than 90 days, so we need to remove it from our build scripts. All dependencies we needed from there are available on Gradle Plugins repo as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,9 +23,9 @@ allprojects {
     buildscript {
         repositories {
             google()
-            jcenter()
             mavenLocal()
             mavenCentral()
+            maven { url 'https://plugins.gradle.org/m2/' }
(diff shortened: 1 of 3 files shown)


MrStahlfelge:
Build fails on fetching the natives, but that shouldn't be caused by this change.

obigu (approved):
Oh, that's a bit of a short notice. 

I've tested it both on libGDX and a new project and all dependencies are resolved correctly.

As a side information, when we update the Android Gradle Plugin to 4.X.we may need to also add the Gradle repo to new projects but not needed yet.

MrStahlfelge:
The Gradle repo is already present on generated project, and it is needed for Gretty. So no todo left here. :)

PR #6309, opened 2020-12-10
GLES 3.0 glMapBufferRange implementation and PBO example

mgsx-dev:
This PR is open for discussion (i'll keep it as draft until we're all set).

I investigated deeply to know why glMapBufferRange was commented from the beginning. Some reasons could be found [in original PR comments](https://github.com/libgdx/libgdx/pull/2527#issuecomment-74588121).

In short, WebGL 2.0 doesn't provide this method for [performance and security reasons when writing to a mapped buffer](https://stackoverflow.com/questions/43530082/how-can-i-upload-a-texture-in-webgl2-using-a-pixel-buffer-object). For read only buffers there is a replacement method (getBufferSubData). So we could provide a limited implementation (read only) for the WebGL 2.0 backend later.

PBO (PixelBufferObject) is one of the use cases using glMapBufferRange. PBO is basically useful to asynchronously transfer pixels from CPU to GPU and vice versa.
I'm currently working on TransformFeedback which is another use case when you want to retrieve transformed data CPU side, that's why i did this PR.

I only tested Desktop (linux) Lwjgl2 and Lwjgl3 backends, it needs to be tested on Android and iOS (i don't have required hardware).

PR #6394, opened 2021-02-04
GLES3 added glTexImage2D and glTexSubImage2D with offset parameter

mgsx-dev:
This should fix regression introduced in #6309.

Backends status : 
* LWJGL 2 and 3: tested
* IOS: not tested, i'm not confident with my C++ code, @Tom-Ski feel free to fix it if needed.
* Android: not tested, apparently it requires native Android to call related methods but it worked in the previous PR by passing null buffer.
* WebGL: N/A, i checked specifications, WebGL2 have the related methods [texImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D) and [texSubImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D), we could implement them in the future.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL30.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL30.java
@@ -17,7 +17,6 @@
 package com.badlogic.gdx.backends.android;
 
 import android.annotation.TargetApi;
-import android.opengl.GLES20;
 import android.opengl.GLES30;
 
 import com.badlogic.gdx.graphics.GL30;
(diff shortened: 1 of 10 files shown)


mgsx-dev:
I rebased the PR but i think IOS implementation needs recent modification to be applied (see #7014).

@Tom-Ski I let you see if you can apply these changes because i don't have IOS dev environement.

Tom-Ski:
Sorry, made a little bit of a mess with a commit. 
iOS is compiling now, and I've tested the PBO test on iPhone 8, and it worked as desktop does.
Android also passes tests for me. So I think this is good.

PR #6397, opened 2021-02-05
Make PolygonSpriteBatch and SpriteBatch overridable

fgnm:
I'd like to extend `PolygonSpriteBatch` to add features from `CpuSpriteBatch`, after tried some solutions I finally have a version that works good, however I need it as a drop-in replacement of `PolygonSpriteBatch`, but I can't extend it directly because there are some private fields I need. I've also some troubles with GWT because the tricky method I've used to get everything working. So, this commit expose these fields in subclasses.

Thanks!

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSpriteBatch.java
@@ -58,12 +58,12 @@ import com.badlogic.gdx.math.Matrix4;
 public class PolygonSpriteBatch implements PolygonBatch {
 	private Mesh mesh;
 
-	private final float[] vertices;
-	private final short[] triangles;
-	private int vertexIndex, triangleIndex;
-	private Texture lastTexture;
-	private float invTexWidth = 0, invTexHeight = 0;
-	private boolean drawing;
+	protected final float[] vertices;
+	protected final short[] triangles;
+	protected int vertexIndex, triangleIndex;
+	protected Texture lastTexture;
+	protected float invTexWidth = 0, invTexHeight = 0;
+	protected boolean drawing;
 
 	private final Matrix4 transformMatrix = new Matrix4();
 	private final Matrix4 projectionMatrix = new Matrix4();
@@ -80,7 +80,7 @@ public class PolygonSpriteBatch implements PolygonBatch {
 	private boolean ownsShader;
 
 	private final Color color = new Color(1, 1, 1, 1);
-	float colorPacked = Color.WHITE_FLOAT_BITS;
+	protected float colorPacked = Color.WHITE_FLOAT_BITS;
 
 	/** Number of render calls since the last {@link #begin()}. **/
 	public int renderCalls = 0;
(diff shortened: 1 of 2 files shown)


tommyettinger:
I think I've proposed this before, and the request was declined each time. I've been able to get around the package-private fields by copying the entire SpriteBatch class, and since it implements Batch in the same way, it works for that case. I don't know about PolygonSpriteBatch. I did benchmark a Batch class with private or package-private fields that are commonly used, against one with (if I remember this right) public fields, and there was a small performance dip with the public fields. I believe the JDK can do extra analysis on fields with restricted visibility. There may be no performance dip or a major one on some JDK vendors or versions, and Android could be very different.

fgnm:
I understand, currently I've a copy of `PolygonSpriteBatch` too but I don't like it much as a final solution.. Should be better to have some some protected getters and leave fields private?
How can I do a benchmark for this class?

tommyettinger:
Unfortunately it's pretty hard to get an accurate benchmark, and I don't know how accurate my results were. I was overdrawing quite a lot of small TextureRegions, which taxes SpriteBatch in ways it doesn't normally see. Ideally you'd want to check the framerate of a purely-automated scene (to rule out user input as affecting the performance), maybe with varying amounts of sprites or polygons drawn. Since in your case you have a replacement for PolygonSpriteBatch, maybe there are benchmarks in Spine? It should be within the permitted terms of its license to run one of their demos for personal use. And, Spine is probably one of the more intense usages of a Batch... Rendering several detailed Spine animations would be a good way of seeing when a game starts to slow down. You'd need to test with vsync disabled, no framerate limit, and so on, so any changes in framerate from a changed Batch would be noticeable. There's all sorts of weird things in the JVM that can affect small delays like this, like making a function call across JAR boundaries, or JIT deoptimization happening after a certain number of optimized calls. Normally I'd recommend JMH or BumbleBench for benchmarks in a heartbeat, but I don't think either works at all with graphical applications.

The problem here is that people have been copying Batch classes for long enough now that they're used to it, and to make SpriteBatch and other similar Batches different in even minor ways needs to be shown that it won't have unexpected consequences. On top of that, copying and editing a Batch class may not be convenient, but it's just as powerful in most cases. I would normally be all in favor of more `protected` and less package-private APIs, but I think this particular class could be too sensitive to performance issues to edit unless we know these changes will be painless. I can help construct benchmarks if you want.

fgnm:
Thanks for the explanation, I understand, and it's much more complex then expected :)
I can try to create the benchmark you described, but I'm not very familiar with different JVM implementation, so I think I need some time to understand everything properly. If you can show me the right way to do that I'll appreciate a lot.
Meanwhile, this PR should be better in a Draft state

mgsx-dev:
i don't think it's a good idea to expose things like this, even via getters/setters. If you want something similar to CpuSpriteBatch, maybe you could PR a CpuPolygonSpriteBatch and change needed fields to package-private the same way it was done for SpriteBatch?

fgnm:
Thanks for suggestions, I see that there are too many downside in this PR, so the better choice I think is to close this one and made another for the `CpuPolygonSpriteBatch`. Currently, the version I have seems to work good, but it's untested and there may be some problems, I'll PR that when properly tested :)
Thanks again

PR #6399, opened 2021-02-05
Rename gdx-box2d projects in gradle to match maven

PokeMMO:
https://github.com/libgdx/libgdx/commit/446e26910cd30fbcda8518d979f75c43bcf44c82 renamed the subproject to "gdx-box2d-core" which does not match the maven/artifact id. This was done to prevent the issue caused by having two projects named the same. We should instead rename the parent project to "gdx-box2d-parent" like it [is in maven](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-box2d/pom.xml).

This is necessary to be able to publish from gradle eventually™

Addresses an issue brought up in #6369

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -23,8 +23,10 @@ include ":backends:gdx-backend-lwjgl3"
 include ":backends:gdx-backend-robovm"
 include ":backends:gdx-backends-gwt"
 
-include ":extensions:gdx-box2d:gdx-box2d"
-include ":extensions:gdx-box2d:gdx-box2d-gwt"
+include(":extensions:gdx-box2d")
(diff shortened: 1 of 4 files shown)


SimonIT (changes requested):
I'm not very happy how you renamed it. The result is a mixture of renamed and not renamed subprojects:
image
I made some changes in my fork https://github.com/libgdx/libgdx/commit/e8177b3108db4b73efe4ae0451e637291a207420. If you are content, I would merge them in you PR

PokeMMO:
I like your way of renaming them better, feel free to merge them into the PR.

SimonIT (approved):

Tom-Ski (approved):

PR #6400, opened 2021-02-05
Exclude apache-jsp from gwt-tests.

PokeMMO:
Gretty pulls it in too, so it's not enough to have it on other classes.

Not sure why it isnt an issue for everyone, but https://mvnrepository.com/artifact/org.gretty/gretty-runner-jetty93/2.3.0 clearly lists it as a dependency.

### DIFF
--- a/tests/gdx-tests-gwt/build.gradle
+++ b/tests/gdx-tests-gwt/build.gradle
@@ -21,6 +21,11 @@ buildscript {
     }
 }
 
+configurations.all {
+    // gwt-dev pulls in apache-jsp (Tomcat), but we don't need it and it messes with gretty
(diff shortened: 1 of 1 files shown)


MobiDevelop:
I've found that if you just specify a servlet container on the gretty config in Gradle, this issue goes away.

MrStahlfelge (commented):
Merging this won't hurt, but I would like to understand why I don't have the issue without the line. 

@MobiDevelop how to you set such a config?

MobiDevelop:
@MrStahlfelge You'd add
`gretty.servletContainer = "jetty9"`
Or
`gretty.servletContainer = "jetty8"`
Depending on the gretty version. I'm not at my computer to look which I added.

PR #6401, opened 2021-02-06
Updated SpriteBatch and BillboardParticleBatch to allow for up to 16383 quads

MarcinSc:
I believe Mesh can now have 64k-1 vertices, therefore both SpriteBatch and BillboardParticleBatch should increase their maximum supported quads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![logo](libgdx_logo.svg)
 
-[![GitHub Actions build status](https://img.shields.io/github/workflow/status/libgdx/libgdx/Build%20and%20Publish/master?label=GitHub%20Actions)](https://github.com/libgdx/libgdx/actions?query=workflow%3A%22Build+and+Publish%22)
+[![GitHub Actions build status](https://img.shields.io/github/actions/workflow/status/libgdx/libgdx/build-publish.yml?branch=master&label=GitHub%20Actions)](https://github.com/libgdx/libgdx/actions?query=workflow%3A%22Build+and+Publish%22)
 
 [![Sonatype Nexus (Releases)](https://img.shields.io/nexus/r/com.badlogicgames.gdx/gdx?nexusVersion=2&server=https%3A%2F%2Foss.sonatype.org&label=version)](https://search.maven.org/artifact/com.badlogicgames.gdx/gdx)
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
```diff
@@ -36,6 +36,7 @@
  * @author mzechner
  * @author Nathan Sweet */
 public class SpriteBatch implements Batch {
+	public static final int MAX_SPRITE_COUNT = 16383;
```
tommyettinger, line 39: I like that this is in a constant field. Maybe it could have some snippet of Javadocs?
```java
/** The maximum number of sprites that can be rendered in a single batch. */
```

tommyettinger (dismissed):
Other than maybe being able to use some Javadocs in one place, this is a great and welcome change.

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
MarcinSc: I've added lengthy JavaDoc

tommyettinger: Perfect, it explains why the number is what it is.

mgsx-dev:
maybe you could apply this change to PolygonSpriteBatch as well?
also you need to update CHANGES and apply formatter to your changes.

MarcinSc:
I hope I found all the instances where the old limitation was in code...

mgsx-dev (changes requested):
raising the limit is one thing, ensuring it works above 32k is another thing and i think it will crash at some places, (eg. DelaunayTriangulator): conversion from int to short can be done with a cast but conversion from short to int require a mask, eg. `int int_Index = short_index & 0xFFFF`
 
we need some unit tests to confirm it works.

also it's a breaking change since user code may not properly handle this conversion, so we need to add something in CHANGES to explain how to migrate code.

tommyettinger:
I think I could do the steps @mgsx-dev requested. I don't think this code should be discarded, and the Batches should probably all use the higher vertex limit. I might need some guidance from mgsx about what exactly he wants, but other than that this should just be some extra casts and masks, plus a test.

PR #3816, opened 2016-02-02
Support for launching libGDX from a native iOS app

Dzamir:
I implemented the support for launching libGDX from a native iOS application. 

**Example usage:**

The host application must include a framework like in the [MyJavaFramework example](https://github.com/robovm/robovm-samples/tree/master/MyJavaFramework) that embeds the libGDX application.

In the libGDX application I added a class in the iOS module:

```
public class TestApplication {
    private IOSApplication iosApplication;

    public TestApplication() {
    }

    public void createApplication() {
        IOSApplicationConfiguration config = new IOSApplicationConfiguration();
        iosApplication = new IOSApplication(new Main(), config);
        iosApplication.launch("yourFrameworkName");
    }

    public void destroyApplication() {
        Singletons.close();
        iosApplication.destroy();
        iosApplication = null;
        System.gc();
    }

}
```

Inside the iOS application I can now create the TestApplication object and launch it by using the JNI, like in the [MyJavaFramework example](https://github.com/robovm/robovm-samples/tree/master/MyJavaFramework) :

```
- (IBAction)startLibGDX:(id)sender
{
    JavaVMInitArgs vm_args;
    vm_args.version = JNI_VERSION_1_4;
    vm_args.nOptions = 0;
    vm_args.options = NULL;

    jint res;
    JNIEnv *env;
    res = JNI_CreateJavaVM(&vm, &env, &vm_args);
    if (res != JNI_OK)
    {
        [NSException raise:@"JNI_CreateJavaVM() failed" format:@"%d", res];
    }

    _javaVM = javaVM;

    JNIEnv *env;
    jint res = (*[_javaVM vm])->AttachCurrentThread([_javaVM vm], (void**) &env, NULL);
    if (res != JNI_OK) {
        [NSException raise:@"AttachCurrentThread() failed" format:@"%d", res];
    }

    // get the class and methods reference
    testApplicationServiceClass = (*env)->FindClass(env, "net/yourcompany/yourproject/TestApplication");
    if (!testApplicationServiceClass) {
        [NSException raise:@"Failed to find class net.yourcompany.yourproject.TestApplication" format:@""];
    }
    createApplicationMethod = (*env)->GetMethodID(env, testApplicationServiceClass, "createApplication", "()V");
    if (!createApplicationMethod) {
        [NSException raise:@"Failed to find method net.yourcompany.yourproject.TestApplication.createApplication()" format:@""];
    }
[...]

PR #6410, opened 2021-02-08
iOS Robovm backend in UI view

dzolo:
Adds a new `IOSViewApplication` that can be used to insert a libgdx game to a view as child UI view controller. This allow to create an iOS application with native UI which displays a game on one of its screens. 

> Explanation of my intentions: I am using the libgdx for an application, which displays a libgdx map with naive control buttons and includes multiple additional native forms screens. We support the Desktop and the Android for now. This patch would enable us to do the same thing on iOS a well.

A new `BaseIOSApplication` with shared parts of code was created. Most parts of the implementation are shared. Thus adding this feature should not decrease maintainability of the code.

An example, how a embedded game can be created is placed in new test module **gdx-tests-iosrobovm-view**. The example is similar to original *gdx-tests-iosrobovm* (was partially copied from it).


MrStahlfelge:
Thanks, this is similar to #3816 which we closed because it was outdated and the original author left.

obigu:
While I can see the use case for this, imo it's a niche one and, being honest, the chances of this PR being merged are very low. 

The PR is quite large with lots of changes and new classes, several formatting changes that need cleaning up and the new `IOSViewApplication` seems to require a lot of code duplication. Having two `Application`s means extra maintenance work for any future changes that would need to be made and tested in both implementations.

I'm going to close it for the time being, if there's renewed interest in the future it can always be brought back.

PR #6421, opened 2021-02-17
Enhance ObjLoader

yuripourre:
ObjLoader can now handle ambient color and multiple textures.


gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -462,7 +462,7 @@ private void addTexture (ModelMaterial mat, String texFilename, int usage) {
 		}
 
 		public void reset () {
-			ambientColor = Color.WHITE;
```
mgsx-dev: default libGDX shader doesn't support ambient color in material anyway, so i think default to BLACK is probably better.

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -439,4 +420,58 @@ public ModelMaterial getMaterial (final String name) {
 		materials.add(mat);
 		return mat;
 	}
-}
+
+	private static class CurrentMaterial {
```
mgsx-dev, line 424: ObjMaterial would be a better name imo.

also, it would be better to avoid code duplication by calling reset in contructor.

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -439,4 +420,58 @@ public ModelMaterial getMaterial (final String name) {
...
+			addTexture(mat, ambientTexFilename, ModelTexture.USAGE_AMBIENT);
+			addTexture(mat, diffuseTexFilename, ModelTexture.USAGE_DIFFUSE);
+			addTexture(mat, specularTexFilename, ModelTexture.USAGE_SPECULAR);
+			addTexture(mat, shininessTexFilename, ModelTexture.USAGE_SHININESS);
+
+			return mat;
+		}
+
+		private void addTexture (ModelMaterial mat, String texFilename, int usage) {
+			if (texFilename == null) {
```
mgsx-dev, line 455: if texFilename != null then add to material would be better in term of readability

CHANGES
```diff
@@ -4,6 +4,7 @@
 - Scene2d.ui: Added new ParticleEffectActor to use particle effects on Stage
 - API addition: iOS: Added HdpiMode option to IOSApplicationConfiguration to allow users to set whether they want to work in logical or raw pixels (default logical).
 - Fix for #6377 Gdx.net.openURI not working with targetSdk 30
+- ObjLoader: Added ambient color and better texture handling
```
mgsx-dev, line 7: it should be more precise, something like "ObjLoader now support ambientColor, ambientTexture, etc..."

Thing is default libGDX shader doesn't support most of these (only specular texture if i'm not wrong). It could be useful for custom shaders though.

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
yuripourre: Good point, I end up changing to null so it will not change the current behavior.

yuripourre:
@mgsx-dev I updated the PR following your comments.

Thanks for reviewing it!

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -439,4 +420,61 @@ public ModelMaterial getMaterial (final String name) {
...
+
+			return mat;
+		}
+
+		private void addTexture (ModelMaterial mat, String texFilename, int usage) {
+			if (texFilename != null) {
+				ModelTexture tex = new ModelTexture();
+				tex.usage = usage;
+				tex.fileName = texFilename;
+				mat.textures.add(tex);
```
xpenatan, line 463: textures array is not initialized. It will throw null exception

tommyettinger: As far as I can tell, yes, this is a problem. Other parts of the g3d package check if the `textures` field is null before using it.

yuripourre: @xpenatan Nice catch! I forgot this line:
`if (mat.textures == null) mat.textures = new Array<ModelTexture>(1);`

NathanSweet:
Is it good to merge now? I don't use OBJ myself.

mgsx-dev (changes requested):
CHANGES file need an update and ObjLoader header formatting to be removed. Other than that it looks good to me.

I don't use OBJ format neither and i don't know the specification for this format, so i think it'd require some other reviews before merging it.

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- *
```
yuripourre, line 3: IntelliJ is removing this.

CHANGES
```diff
@@ -14,6 +14,7 @@
 - API Addition: Desktop Sound now returns number of channels and sample rate.
 - API Addition: PointSpriteParticleBatch blending is now configurable.
 - TOOLS Features: Blending mode and sort mode can be changed in Flame particle 3D editor.
+- API Addition: ObjLoader now supports ambientColor, ambientTexture, transparencyTexture, specularTexture and shininessTexture
```
mgsx-dev, line 17: well, 1.10.0 is already out :-) so you can start a new 1.10.1 chapter at the beginning of the file

gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
mgsx-dev: sorry i don't use IntelliJ, there might be some configuration for this. So better ask community in Discord

mgsx-dev (changes requested):

mgsx-dev (approved):

PR #6423, opened 2021-02-18
Unfocus children in Group#clearChildren()

petoncle:
`Group.removeActor(Actor child, boolean unfocus)` gives the option to unfocus the removed child.
`Group.removeActor(Actor child)` unfocuses the removed child (calls `removeActor(child, true)`).

For some reason, `Group.clearChildren()` currently does not unfocus the children. This looks wrong.
This PR adds a `clearChildren(boolean unfocus)` method.

Now the question is: do you want `clearChildren()` (no arguments) to unfocus like `removeActor(child)` does, or do you want `clearChildren()` to not unfocus to avoid potentially breaking apps which may rely on the fact that clearChildren() never unfocused until now? I'm all for breaking apps, but 🤷.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@
 - Scene2d.ui: Added new ParticleEffectActor to use particle effects on Stage
 - API addition: iOS: Added HdpiMode option to IOSApplicationConfiguration to allow users to set whether they want to work in logical or raw pixels (default logical).
 - Fix for #6377 Gdx.net.openURI not working with targetSdk 30
+- API addition: Added Group#clearChildren(boolean unfocus).
 
 [1.9.14]
 - [BREAKING CHANGE] iOS: IOSUIViewController has been moved to its own separate class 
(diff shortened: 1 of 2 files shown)


petoncle:
After thinking about it some more, I believe not making `clearChildren()` (no arguments) unfocus its children is useless for me personally as I'll have to change all my `clearChildren()` calls to `clearChildren(true)`, and most of the times I don't even call `clearChildren()` but `clear()`.

NathanSweet:
Good find. I think `clearChildren` without specifying unfocus should match `removeActor` without specifying unfocus: it should unfocus. As @petoncle mentioned, this makes it a breaking change but the API consistency is likely worth it.

Some notes: Having removed actors retain focus can cause errors and surprises. We don't always remove focus though, because sometimes removal is temporary and keeping the focus is desired. Adding an unfocus boolean to a bunch of methods isn't great, but gives the flexibility needed and helps people consider the issue.

NathanSweet:
I wonder, should we have a Group `clear(boolean unfocus)`?

petoncle:
> I wonder, should we have a Group `clear(boolean unfocus)`?

We might as well for consistency with the `clearChildren` and `removeActor`.

petoncle:
@NathanSweet I've added a Group#clear(boolean unfocus) method if you're interested. I think you'll have to re-open the PR to be able to merge it in.

NathanSweet:
No worries, I've made the change.

PR #6424, opened 2021-02-18
Set window title on gwt

SimonIT:


tommyettinger (approved):
Short and sweet. I don't see why this wouldn't work.

MrStahlfelge (approved):
Tested and it worked

PR #6425, opened 2021-02-20
Get js heap size on gwt

SimonIT:
https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory

I'm not totally sure if it's safe to use `UnsafeNativeLong`.

> The compiler will then allow you to pass a long into and out of JavaScript. It will still be an opaque data type, however, so the only thing you will be able to do with it will be to pass it back to Java.

from http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html

An alternative would be to convert it to a string and then parse it to a long back.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -39,7 +39,6 @@ import com.google.gwt.animation.client.AnimationScheduler.AnimationCallback;
 import com.google.gwt.core.client.EntryPoint;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.core.client.JavaScriptObject;
-import com.google.gwt.core.client.UnsafeNativeLong;
 import com.google.gwt.dom.client.CanvasElement;
 import com.google.gwt.dom.client.Document;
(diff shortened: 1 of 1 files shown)


intrigus:
> A non-standard extension added in Chrome, this property provides an object with basic memory usage information. You should not use this non-standard API.

How big do you expect the result to be?
Why can't `getJavaHeap` return an int instead of a long?
Because I don't think using `UnsafeNativeLong` will work here.

SimonIT:
I know that it's not standard API and not supported on all browsers. That's why I added the check. I think it's better to support some browsers than none. (That's also what libgdx does with all other backends and gwt.)

The browser reports a js heap size limit of 4 294 705 152 for me. That's to large for java ints. So I thought we have to use the longs.

Why do you think it won't work? I had the number as debug output in the console when I tested it

intrigus:
> Why do you think it won't work? I had the number as debug output in the console when I tested it

Because of this:
> The compiler will then allow you to pass a long into and out of JavaScript. It will still be an opaque data type

Why don't you let the method return a `double` and cast that to a `long`?

SimonIT:
Didn't thought on returning a `double` and cast that to a `long`. Changed accordingly
Should be fine now

badlogic:
Looks good, merged.

PR #6431, opened 2021-02-23
Fix mesh implicit update when retrieving data for read only, fixes #6430

### LINKED ISSUE #6430: calculateBoundingBox unnecessarily dirties the vertex and index buffers
#### Issue details
Calling Mesh.calculatingBoundingBox (and related functions like Mesh.extendBoundingBox, Mesh.calculateRadius) calls VertexData.getBuffer() and IndexData.getBuffer().  The implementations of these methods mark the buffer as dirty (for the general case where the user has called them and may modify the data) and on the next call to VertexData.bind() the data is reuploaded to the GPU using glBufferData or similar.

This means that calculating the bounding box is made into an expensive GPU operation even though it doesn't need to go near the GPU.  I have problems with this where I am animating a small part of a big model and recalculating the bounding box, which causes my whole model to be reuploaded to the GPU.  The test case below shows that while the bounding box calculation is quite quick (5 microseconds), the reupload is much more expensive if the GPU is saturated (740 microseconds).

I suggest that VertexData should have some kind of access mechanism specifically for the Mesh bounding box methods that allows it to access the data without marking it as dirty, whether that's a getBufferReadOnly() method or some kind of instanceof check for the built-in buffer types.  It's a pain to work around as a library user - you either need to hold a duplicate copy of the mesh for the bounding box checks, or you have to duplicate the calculateBoundingBox code yourself and hack into the vertex buffer internals to get the data without marking it as dirty.

#### Reproduction steps/code
Here is a complete test case showing the times involved:
```java
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.graphics.Mesh;
import com.badlogic.gdx.graphics.VertexAttribute;
import com.badlogic.gdx.graphics.g3d.model.MeshPart;
import com.badlogic.gdx.graphics.glutils.ShaderProgram;
import com.badlogic.gdx.math.collision.BoundingBox;

public class MyGdxGame extends ApplicationAdapter {
    public static final int LOOP_AMOUNT = 1000;
    public static final int FRAME_AMOUNT = 10;
    ShaderProgram shaderProgram;
	Mesh bigMesh;
	MeshPart meshPart;
	BoundingBox boundingBox = new BoundingBox();
	int frames = 0;
	
	long nanosJustBox;
	long nanosDirtyBind;
        long nanosCleanBind;
	
	@Override
	public void create () {
		bigMesh = new Mesh(true, 30_000, 100, VertexAttribute.Position(), VertexAttribute.Normal(), VertexAttribute.Tangent(), VertexAttribute.TexCoords(0));
		// Vertex values not important here:
		float[] vertices = new float[30_000 * 11];
        short[] indices = new short[bigMesh.getMaxIndices()];
        for (int i = 0; i < bigMesh.getMaxIndices(); i++)
        {
            indices[i] = (short)i;
        }
        bigMesh.setVertices(vertices);
        bigMesh.setIndices(indices);
        meshPart = new MeshPart("", bigMesh, 0, 100, GL20.GL_TRIANGLES);
        String vs = "attribute vec4 a_position; void main(){ gl_Position = a_position;}";
        String fs = "void main(){ gl_FragColor = vec4(1.0, 1.0, 1.0, 1.0); }";
        shaderProgram = new ShaderProgram(vs, fs);
	}

	@Override
	public void render () {
	    if (frames < FRAME_AMOUNT)
        {
            long t0 = System.nanoTime();
            for (int i = 0; i < LOOP_AMOUNT; i++)
                meshPart.mesh.calculateBoundingBox(boundingBox, meshPart.offset, meshPart.size);
            long t1 = System.nanoTime();
            for (int i = 0; i < LOOP_AMOUNT; i++)
            {
                // Calls vertices.getBuffer(), which does the dirty marking without the work of calculateBoundingBox:
                meshPart.mesh.getVerticesBuffer();
                bigMesh.bind(shaderProgram);
                bigMesh.unbind(shaderProgram);
            }
            long t2 = System.nanoTime();
            for (int i = 0; i < LOOP_AMOUNT; i++)
            {
                bigMesh.bind(shaderProgram);
                bigMesh.unbind(shaderProgram);
            }
            long t3 = System.nanoTime();
            
            nanosJustBox += (t1 - t0);
            nanosDirtyBind += (t2 - t1);
            nanosCleanBind += (t3 - t2);
        }
	    else if (frames == FRAME_AMOUNT)
        {
            System.out.println("Nanos on box: " + (nanosJustBox / (FRAME_AMOUNT * LOOP_AMOUNT)));
            System.out.println("Nanos on dirty bind: " + (nanosDirtyBind / 10000L));
            System.out.println("Nanos on clean bind: " + (nanosCleanBind / 10000L));
        }
        frames += 1;
	}
	
	@Override
	public void dispose () {
		bigMesh.dispose();
	}
}

```

The above used libGDX 1.9.14

#### Output
On an Nvidia GTX 1080 I got the following output:
```
Nanos on box: 5383
Nanos on dirty bind: 739774
Nanos on clean bind: 1469
```
The difference between dirty bind and clean bind is equivalent to the effect of calculateBoundingBox unnecessarily marking the mesh as dirty.

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

mgsx-dev:
It's a minor breaking change since users rarely subclass Mesh class or underlying VertexData / IndexData in their own code.

I run every impacted tests and some tests covering changed code, all goes well, i only have a crash with a Bullet test (FrustumCullingTest) but it's not related to changes in this PR, i'll investigate further and try to see what's going on.

Note that this change is a major performance improvement for people reading mesh data.

In order to help the reviewers : 
* i made a separate commit for code simplification.
* i removed some comment duplication (same javadoc on both interface and implementation)


mgsx-dev:
Note: i'll apply this change to InstanceData for consistency

mgsx-dev:
Note that instanced rendering doesn't work with Lwjgl2 backend for me on Linux (glVertexAttribDivisor not supported). But it works with Lwjgl3 backend, i guess it's a known issue.

intrigus:
Why not use `asReadOnlyBuffer` to ensure that no one accidentally writes to the buffer?

mgsx-dev:
@intrigus i think it would produce garbage and the point of this PR is mostly to avoid marking buffers as dirty when they're only retrieved for reading.

intrigus:
> @intrigus i think it would produce garbage and the point of this PR is mostly to avoid marking buffers as dirty when they're only retrieved for reading.

You could cache the read-only buffer.
I mean you can keep it this way, but some time, some one will probably be wondering why their Mesh is not updated :D

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -134,17 +134,20 @@ public void updateIndices (int targetOffset, short[] indices, int offset, int co
...
+	/** @deprecated use {@link #getBuffer(boolean)} instead */
+	@Override
+	@Deprecated
 	public ShortBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public ShortBuffer getBuffer (boolean forWriting) {
```
rami-dabain, line 146: Shouldn't a getter be read-only by default, as the name implies (GET)
marking a buffer as dirty should be done either via a setter `setBuffer()` or just `markDirty()`

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteCache.java
```diff
@@ -241,7 +243,7 @@ public int endCache () {
 	/** Invalidates all cache IDs and resets the SpriteCache so new caches can be added. */
 	public void clear () {
 		caches.clear();
-		((Buffer) mesh.getVerticesBuffer()).clear().flip();
+		((Buffer) mesh.getVerticesBuffer(true)).clear().flip();
```
rami-dabain, line 246: `.clear()` and `.flip()` should set the buffer as dirty , not `getBuffer()`

mgsx-dev: we can't modify clear/flip methods, those are from JDK.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
mgsx-dev: It still a getter: original buffer reference is not changed. Basically any getter returning an object doesn't prevent caller to modify its state.
IMO, "forWriting" parameter is simpler to understand and more obvious for the caller to have it at one place.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
```diff
@@ -134,17 +134,20 @@ public void updateIndices (int targetOffset, short[] indices, int offset, int co
...
+		isDirty |= forWriting;
```
rami-dabain, line 147: isDirty should not be settable to `false` in this method. 
Imagine this scenario:
- 1st call with `forWriting = true`
- 2nd call with `forWriting=false`
- updating if dirty
In this case the buffer won't be uploaded to the GPU when it should have been.

proposed solution:
`if (!isDirty) isDirty |= forWriting;`

gdx/src/com/badlogic/gdx/graphics/glutils/IndexBufferObjectSubData.java
```diff
@@ -153,18 +153,20 @@ public void updateIndices (int targetOffset, short[] indices, int offset, int co
...
+	@Override
+	@Deprecated
 	public ShortBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public ShortBuffer getBuffer (boolean forWriting) {
+		isDirty |= forWriting;
```
rami-dabain, line 166: Same issue as above, the buffer could be reset as "not dirty" by mistake

gdx/src/com/badlogic/gdx/graphics/glutils/InstanceBufferObject.java
```diff
@@ -75,12 +75,20 @@ public int getNumMaxInstances () {
...
 	@Override
+	@Deprecated
 	public FloatBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public FloatBuffer getBuffer (boolean forWriting) {
+		isDirty |= forWriting;
```
rami-dabain, line 88: same as above settable to "not dirty" by mistake

gdx/src/com/badlogic/gdx/graphics/glutils/InstanceBufferObjectSubData.java
```diff
@@ -109,12 +109,20 @@ public int getNumMaxInstances () {
...
 	@Override
+	@Deprecated
 	public FloatBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public FloatBuffer getBuffer (boolean forWriting) {
+		isDirty |= forWriting;
```
rami-dabain, line 122: same as above settable to "not dirty" by mistake

gdx/src/com/badlogic/gdx/graphics/glutils/VertexBufferObject.java
```diff
@@ -93,12 +93,20 @@ public int getNumMaxVertices () {
...
 	@Override
+	@Deprecated
 	public FloatBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public FloatBuffer getBuffer (boolean forWriting) {
+		isDirty |= forWriting;
```
rami-dabain, line 106: same as above settable to "not dirty" by mistake

gdx/src/com/badlogic/gdx/graphics/glutils/VertexBufferObjectSubData.java
```diff
@@ -99,12 +99,20 @@ public int getNumMaxVertices () {
...
 	@Override
+	@Deprecated
 	public FloatBuffer getBuffer () {
 		isDirty = true;
 		return buffer;
 	}
 
+	@Override
+	public FloatBuffer getBuffer (boolean forWriting) {
+		isDirty |= forWriting;
```
rami-dabain, line 112: same as above settable to "not dirty" by mistake (sorry if this is annoying)

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/IndexBufferObject.java
mgsx-dev: `isDirty |= forWriting` is the same as `isDirty = isDirty || forWriting` which would be the same as `if (!isDirty) isDirty = true;`. all of them cover the scenario you mentioned.

Tom-Ski (approved):

mgsx-dev:
i had to rebase commits to resolve CHANGES conflicts, sorry about that.

Tom-Ski (approved):

PR #6434, opened 2021-02-25
Fixes #6433: HTTP request gets rejected if configuration option maxNetThreads is set

### LINKED ISSUE #6433: HTTP request gets rejected if configuration option maxNetThreads is set
#### Issue details
If mote HTTP requests are invoked in same time above the maximal allowed network thread count (specified by the configuration option `maxNetThread`) then they are rejected. The configuration option is useless because of this.

The reason for this is that the `NetJavaImpl` creates a `ThreadPoolExecutor` with work queue set to `SynchronousQueue` instance. The `SynchronousQueue` offer method always rejects an item, thus if active count is higher then maximal thread pool size (set by `maxNetThread`) the task is not accepted (`ThreadPoolExecutor#execute` throws `LinkedBlockingQueue`).

#### Reproduction steps/code
```java
// config settings
config.maxNetThreads = 5;

// Listener
class MyHttpResponseListener implements Net.HttpResponseListener {
  void handleHttpResponse (HttpResponse httpResponse) {}
  void failed (Throwable t) {
    System.out.println("failed: " + t.getMessage());
  }
  void cancelled () {}
}

// create more then 5 HTTP requests
for (int i = 0; i < 6; i++) {
  HttpRequest httpRequest = /* ... */;
  Gdx.net.sendHttpRequest(httpRequest, null); // the sixth request will print message `failed: ...`
}
```

_1.9.14_

#### Stacktrace
For  `maxNetThread` set to 5:
```java
Task java.util.concurrent.FutureTask@34f07602 rejected from java.util.concurrent.ThreadPoolExecutor@49035613[Running, pool size = 5, active threads = 5, queued tasks = 0, completed tasks = 0]
```

- [x] Android
- [x] iOS
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

dzolo:
Fixes queuing network tasks if configuration option `maxNetThreads` is set to non-default value (other than MAX_INTEGER).

If non-default value is set then the ThreadPoolExecutor used by NetJavaImpl is created with:
- core pool size to `maxNetThreads` (instead of 0)
- enables `ThreadPoolExecutor#allowCoreThreadTimeOut` to remove core pool threads after the specified timeout
- work queue type `LinkedBlockingQueue` (instead of SynchronousQueue which does not accept any tasks to be queued)

The behavior if `maxNetThreads` is set to default value remain unchanged. Thus this change should not cause any regression.

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
+++ b/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
@@ -25,7 +25,7 @@ import java.net.URL;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -123,7 +123,7 @@ public class NetJavaImpl {
 		}
 	}
 
-	private final ExecutorService executorService;
+	private final ThreadPoolExecutor executorService;
 	final ObjectMap<HttpRequest, HttpURLConnection> connections;
 	final ObjectMap<HttpRequest, HttpResponseListener> listeners;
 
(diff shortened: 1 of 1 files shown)


MrStahlfelge (approved):
Works with default config and changed config

PR #6435, opened 2021-02-26
Remove commented code

BobDorland:
This pull request removes commented code from three files in the `gdx/src/com/badlogic` folder, in order to make these files more neat.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/PixmapIO.java
+++ b/gdx/src/com/badlogic/gdx/graphics/PixmapIO.java
@@ -90,7 +90,6 @@ public class PixmapIO {
 			DataOutputStream out = null;
 
 			try {
-				// long start = System.nanoTime();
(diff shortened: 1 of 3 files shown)


NathanSweet:
FWIW, the commented code removed from Array.java prints about concurrent usage, which is occasionally useful if not using `Collections.allocateIterators = true`.

BobDorland:
> FWIW, the commented code removed from Array.java prints where the iterator is acquired, which is occasionally useful if not using `Collections.allocateIterators = true` and having a crash from modifications while iterating.

I see, sorry. I didn't realize these lines were still used for debugging. Would you prefer to have it in? I can add it back in a separate PR.

MrStahlfelge:
They are of no real use if they are commented and not usable via a debug flag.

NathanSweet:
I don't mind much that the comments are gone, no need to put them back.

They are certainly of use when running from source and the functionality is needed. Overhead of a debug flag is not wanted.

PR #6438, opened 2021-02-27
Fix wrong Tree batch color when icons are present

fgnm:
When icons are provided in a `Tree.Node` batch color is not correctly set. It lost parent alpha and, after drawing the icon, batch is forced to `WHITE` color instead to restore the original one. This is what happens when trying to fade out a container that has a `Tree` inside.
tree-batch-color
 
Thanks


MrStahlfelge (approved):
Looks good to me

PR #6444, opened 2021-03-09
Queue Testing and Quality improvement

ColinGeukes:
In this pull request, the following has been done to the file `utils/queue`:

- Added/Extended the test coverage 
- Removed commented code
- Updated/Improved the javaDocs
- Minor static analysis upgrades

The code coverage has been improved from

| queue (old)   | 100% | 70% | 80% |
| queue (new)      | 100% |   100% | 100% |

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -64,7 +64,7 @@ public class Queue<T> implements Iterable<T> {
 		T[] values = this.values;
 
 		if (size == values.length) {
-			resize(values.length << 1);// * 2
+			resize(values.length << 1);
 			values = this.values;
 		}
 
@@ -450,8 +450,6 @@ public class Queue<T> implements Iterable<T> {
 		int index;
 		boolean valid = true;
 
-// QueueIterable<T> iterable;
-
 		public QueueIterator (Queue<T> queue) {
 			this(queue, true);
 		}
@@ -463,7 +461,6 @@ public class Queue<T> implements Iterable<T> {
 
 		public boolean hasNext () {
 			if (!valid) {
-// System.out.println(iterable.lastAcquire);
 				throw new GdxRuntimeException("#iterator() cannot be used nested.");
 			}
 			return index < queue.size;
@@ -472,7 +469,6 @@ public class Queue<T> implements Iterable<T> {
 		public T next () {
 			if (index >= queue.size) throw new NoSuchElementException(String.valueOf(index));
 			if (!valid) {
-// System.out.println(iterable.lastAcquire);
 				throw new GdxRuntimeException("#iterator() cannot be used nested.");
 			}
 			return queue.get(index++);
@@ -498,8 +494,6 @@ public class Queue<T> implements Iterable<T> {
 		private final boolean allowRemove;
 		private QueueIterator iterator1, iterator2;
 
-// java.io.StringWriter lastAcquire = new java.io.StringWriter();
-
 		public QueueIterable (Queue<T> queue) {
 			this(queue, true);
 		}
@@ -512,13 +506,10 @@ public class Queue<T> implements Iterable<T> {
 		/** @see Collections#allocateIterators */
 		public Iterator<T> iterator () {
 			if (Collections.allocateIterators) return new QueueIterator(queue, allowRemove);
-// lastAcquire.getBuffer().setLength(0);
-// new Throwable().printStackTrace(new java.io.PrintWriter(lastAcquire));
+
 			if (iterator1 == null) {
 				iterator1 = new QueueIterator(queue, allowRemove);
 				iterator2 = new QueueIterator(queue, allowRemove);
-// iterator1.iterable = this;
-// iterator2.iterable = this;
 			}
 			if (!iterator1.valid) {
 				iterator1.index = 0;
--- a/gdx/src/com/badlogic/gdx/utils/Queue.java
+++ b/gdx/src/com/badlogic/gdx/utils/Queue.java
@@ -21,506 +21,679 @@ import java.util.NoSuchElementException;
 
 import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
-/** A resizable, ordered array of objects with efficient add and remove at the beginning and end. Values in the backing array may
+/**
+ * A resizable, ordered array of objects with efficient add and remove at the beginning and end. Values in the backing array may
  * wrap back to the beginning, making add and remove at the beginning and end O(1) (unless the backing array needs to resize when
- * adding). Deque functionality is provided via {@link #removeLast()} and {@link #addFirst(Object)}. */
+ * adding). Deque functionality is provided via {@link #removeLast()} and {@link #addFirst(Object)}.
+ */
 public class Queue<T> implements Iterable<T> {
-	/** Contains the values in the queue. Head and tail indices go in a circle around this array, wrapping at the end. */
-	protected T[] values;
-
-	/** Index of first element. Logically smaller than tail. Unless empty, it points to a valid element inside queue. */
-	protected int head = 0;
-
-	/** Index of last element. Logically bigger than head. Usually points to an empty position, but points to the head when full
-	 * (size == values.length). */
-	protected int tail = 0;
-
-	/** Number of elements in the queue. */
-	public int size = 0;
-
-	private transient QueueIterable iterable;
-
-	/** Creates a new Queue which can hold 16 values without needing to resize backing array. */
-	public Queue () {
-		this(16);
-	}
-
-	/** Creates a new Queue which can hold the specified number of values without needing to resize backing array. */
-	public Queue (int initialSize) {
-		// noinspection unchecked
-		this.values = (T[])new Object[initialSize];
-	}
-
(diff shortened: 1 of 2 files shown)


yuripourre:
@ColinGeukes hey, thank you for this PR every test is very welcome.
My only concern with this PR is that there are lost of style changes.
Do you mind using the lint command so we can review more easily?

ColinGeukes:
@yuripourre Thank you very much with your input, 
I will apply the lint command in the next commit, hope that will fix the styling changes.

ColinGeukes:
@yuripourre The libGDX formatter was applied to revert the style changes, only javadocs now start at a new line in the form of
```
/** 
* JavaDoc
*/
```
previously the file used the format of:
```
/** JavaDoc */
```

Should I also revert that? It is not being applied by the formatter.

ColinGeukes:
@mgsx-dev, @yuripourre The format changes have been reverted!

yuripourre:
Seems good to me!

yuripourre (dismissed):
Seems good to me!

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -245,19 +255,22 @@ public T removeIndex (int index) {
 		return value;
 	}
 
-	/** Returns true if the queue has one or more items. */
+	/** Returns true if the queue has one or more items.
+	 * @return true if not empty, else false */
```
mgsx-dev, line 259: no need to repeat yourself: "@return true if the queue has one or more items." is enough. You can simplify at other places too.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -16,11 +16,11 @@
 
 package com.badlogic.gdx.utils;
 
+import com.badlogic.gdx.utils.reflect.ArrayReflection;
```
mgsx-dev, line 19: no need to move imports

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -300,8 +316,8 @@ public T get (int index) {
 		return values[i];
 	}
 
-	/** Removes all values from this queue. Values in backing array are set to null to prevent memory leak, so this operates in
```
mgsx-dev, line 303: looks like you formatted manually (newline), please only format lines you changed.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -355,6 +374,9 @@ public String toString () {
 		return sb.toString();
 	}
 
+	/** Method to transform the queue into a string representation,
```
mgsx-dev, line 377: "Method to" doesn't add value, we know it's a method.

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -368,6 +390,8 @@ public String toString (String separator) {
 		return sb.toString();
 	}
 
+	/** Method to transform the queue into a hashCode.
+	 * @return the hashCode of the queue */
```
mgsx-dev, line 394: no need to javadoc overriden methods if there is no specific implementation information.
add @Override annotation instead (same for equals method)

gdx/src/com/badlogic/gdx/utils/Queue.java
```diff
@@ -438,87 +468,108 @@ public boolean equalsIdentity (Object o) {
...
-			if (itsIndex == itsBackingLength) itsIndex = 0;
+			if (myIndex == myBackingLength)
+				myIndex = 0;
+			if (itsIndex == itsBackingLength)
+				itsIndex = 0;
 		}
 		return true;
 	}
 
+	/** The QueueIterator class is used to iterate through the queue.
```
mgsx-dev, line 479: we know it's a class, an iterator, and it's used to iterate. Javadoc is not really necessary here.

mgsx-dev (changes requested):
It sounds like unnecessary over-quality to me, a lot of noise for not much value.

I didn't marked all the requested changes but the main rules are (imho) : 
* no need to javadoc overriden methods unless implementation specific information, use `@Override` annotation instead (IDE inherit automatically javadoc from super method)
* avoid the word "constructor", "method", etc..
* only add valuable information for the user, keep it simple and concise
* runtime exceptions are rarely documented (not needed for GdxRuntimeException)

I'm not fan of wording like "the queue **you want** to create an iterator of"

for the rest, as i'm not english native, i can't approve the whole thing.

Junits you added doesn't hurt though.

Sorry to be a bit rude and i don't want to discourage you, but i think it'd be better to focus on the part of libGDX that really miss documentation and tests. And i don't think collections are at the top priority.

ColinGeukes:
@mgsx-dev thank you very much with your input,
it is my first merge request with an open source project, I am trying to adapt to the project.

>  Sorry to be a bit rude and i don't want to discourage you.
Your input really helps in delivering quality merge requests that fit the project.

> Junits you added doesn't hurt though.
After this merge request I will mostly focus on the addition of test cases instead of refactoring the documentation. 

but i think it'd be better to focus on the part of libGDX that really miss documentation and tests. And i don't think collections are at the top priority.
I understand that there is little gain in updating the text, so I am continuing with mainly improving the test coverage after this merge request. If an entire file lacks documentation I might update that as well.

Thank you so far for the input, I believe I have resolved

gdx/src/com/badlogic/gdx/utils/Queue.java
ColinGeukes: I changed the text to not duplicate the return tag, but to provide information for the function. Without a body, the javaDoc might look strange when observing the function.

mgsx-dev: does it add any value? no. IMO javadoc is not even necessary for such self-explanatory methods.

ColinGeukes: It is removed 👍

NathanSweet:
Do not add @Override, diamond operators, or other stylistic changes if they do not already exist in the class you are modifying.

Queue is already thoroughly javadoc'ed before this PR, at least by my own rules. Javadoc is only used when it is needed. Adding javadoc to something that is already clear is not helpful (eg "Sets the name." for a `setName` method). Using more words to say the same thing is not helpful.

Tests are good, especially for collections. We could merge the Queue test, but not any of the changes to Queue (unless some useful change is hidden in the noise).

ColinGeukes:
@NathanSweet I reverted ALL changes to the Queue.java. Even commented code is back.

gdx/test/com/badlogic/gdx/utils/QueueTest.java
```diff
@@ -1,16 +1,29 @@
...
 
 public class QueueTest {
+
+	@Rule
+	public final ExpectedException exception = ExpectedException.none();
+
 	@Test
-	public void addFirstAndLastTest() {
+	public void initializeWithType () {
+		Queue<String> queue = new Queue<String>(1, String.class);
```
mgsx-dev, line 20: with or without type initialization, this test should pass. Maybe you need to test direct access to underlying array instead.

gdx/test/com/badlogic/gdx/utils/QueueTest.java
```diff
@@ -374,6 +402,233 @@ public void hashEqualsTest () {
...
+		queue.first();
+	}
+
+	@Test
+	public void indexOfNotFound () {
+		Queue<Integer> queue = new Queue<>();
+
+		// Check an existing element
+		queue.addFirst(1);
+		assertEquals(0, queue.indexOf(1, true));
```
mgsx-dev, line 447: you shouldn't use identity comparison with Integer class. (same remark for some other lines below)

gdx/test/com/badlogic/gdx/utils/QueueTest.java
```diff
@@ -374,6 +402,233 @@ public void hashEqualsTest () {
...
+
+		assertEquals(iterator, iterator.iterator());
+	}
+
+	@Test
+	public void resizeLargerTailThanHead () {
+		Queue<String> queue = new Queue<>(2);
+		queue.addLast("1");
+		queue.addLast("2");
+		queue.head = 1;
```
mgsx-dev, line 624: you shouldn't modify internal (protected) field in your unit tests

gdx/test/com/badlogic/gdx/utils/QueueTest.java
```diff
@@ -374,6 +402,233 @@ public void hashEqualsTest () {
...
+	@Test
+	public void queueIteratorInvalidNext () {
+		// Initialize queue and iterator.
+		Queue<String> queue = new Queue<>();
+		Queue.QueueIterator<String> iterator = new Queue.QueueIterator<>(queue);
+		iterator.valid = false;
+		queue.addLast("1");
+
+		// GdxRuntimeException will be thrown by check hasNext of invalid iterator.
+		exception.expect(GdxRuntimeException.class);
```
mgsx-dev, line 592: you shouldn't test undocumented exceptions, only test methods contract from its javadoc.
So i think you can remove this test.

mgsx-dev (changes requested):
Please remove formatting on not modified lines.

Some unit tests seems useful, i have doubts for some others, looks like over-quality to me. I let owners decide @NathanSweet .

PR #6445, opened 2021-03-09
Move CI infrastructure to GitHub Actions

badlogic:
Up until now, libGDX native libraries and snapshot releases have been build via a custom Jenkins instance at https://libgdx.badlogicgames.com/jenkins. The build artifacts have also been stored on this server for public access. The server has seen many failures and breakages and is hard to maintain.

This PR switches the entire CI infrastructure over to GitHub actions, so the entire system is in one place. It also allows us to keep the build infrastructure, e.g. compilers, up-to-date more easily in a simple, reproducible, declarative way.

User facing changes:

* Android armeabi native libraries have been dropped. They are no longer supported in the latest NDK.
* iOS/tvOS ARM 32-bit support has been dropped. When building an IPA via RoboVM, only select the ARM64 slice.
* Linux 32-bit support has been dropped.
* Linux ARM and AARCH64 native libraries are packaged and published to SonaType.
* macOS Apple Silicon support has been added in form of fat `.dylib` share libraries, with an AARCH64 and an x86-64 slice. This could be improved in the future, but requires more significant changes to gdx-jnigen.
* The native libraries are now hosted on S3. The directory structure mirrors that of the libGDX repository, e.g.`gdx/libs/gdx-natives.jar`. Users who directly downloaded individual binaries must fix their URLs. See `fetch.xml` in the root of this repository for the new URLs and structure.

Everything else works as before. 

Some more detail son the changes. The repository now features 3 GitHub Actions workflows.

* build-natives.yml, compiles all libGDX native libraries for each architecture and uploads the result to an S3 bucket. Only .dll, .dylib, .so, .a, and a.tvos, and gdx-xxx-natives.jar files are uploaded, mirroring the directory structures found in the libGDX repository, e.g. gdx/libs/gdx-natives.jar. This workflow must be triggered manually, it takes about 25 minutes, mostly due to the slow GitHub Actions macOS build node. Natives are not updated frequently, so the manual trigger is OK.
* test-pullrequest.yml, fetches the natives from S3 and compiles and tests the changes. Triggered on opening/updating a new pull request.
* release-snapshot.yml, fetches the natives from S3 and compiles and deploys the Maven artifacts SNAPSHOT builds to SonaType. Triggered by a push to the master branch.

For release builds, ant -f fetch.xml && mvn release:prepare && mvn release:perform as usual.

The build.xml Ant script used to build the libGDX distribution, as well as fetch.xml Ant script used to pull down the native libraries have been fixed to use the S3 bucket.

The Maven build has also been fixed to fetch the natives from S3.

Android no longer supports armeabi, so it has been removed from the build.

Linux x86 (32-bit) support has been removed.

Linux ARM and AARCH64 support has been added. The gdx-xxx-native.jar files now contain native libraries of these architectures as well.

macOS Apple Silicon native libraries are now also build. They are merged into a fat shared library. We can consider splitting these up again in the future. Updated to gdx-jnigen version 2.1.0 which has the needed changes.


Tom-Ski:
Just to clarify, you've taken out the openal build and are just hosting them statically? I see you've added them manually, but which build are these from? I recently disabled 32bit support for ios in the openal build due to build infrastructure issues, and haven't reverted xcode yet to be able to build those slices. I think its fine to just remove them from the build since we never change the code there, just need to make sure those natives are the 'old' versions with 32bit support.

PokeMMO:
> build-natives.yml, compiles all libGDX native libraries for each architecture and uploads the result to an S3 bucket. Only .dll, .dylib, .so, .a, and a.tvos, and gdx-xxx-natives.jar files are uploaded, mirroring the directory structures found in the libGDX repository, e.g. gdx/libs/gdx-natives.jar. This workflow must be triggered manually, it takes about 25 minutes, mostly due to the slow GitHub Actions macOS build node. Natives are not updated frequently, so the manual trigger is OK.

I think there's a *lot* of benefit in building natives on every CI, using ccache should lower the actual duration. This would allow users to use the CI runners without changes in their own forks.

>test-pullrequest.yml, fetches the natives from S3 and compiles and tests the changes. Triggered on opening/updating a new pull request.

S3 for the final output is good, but github offers an "artifacts" action which can be used to upload natives between different runners.

Final result would look like mac-os runner first, which compiles macos/ios and then uses the uploads artifacts as "macos-natives".
ubuntu runner which depends on macos runs next, which downloads the "macos-natives" artifacts and compiles linux/windows/android binaries.

This gives a lot of flexibility for forks and PR to test things since the artifacts can be downloaded manually from the action result.

See
https://github.com/PokeMMO/jnigen-template/blob/master/.github/workflows/gradle.yml#L32-L36
https://github.com/PokeMMO/jnigen-template/blob/master/.github/workflows/gradle.yml#L74-L78
https://github.com/PokeMMO/jnigen-template/blob/master/.github/workflows/gradle.yml#L86-L90

>release-snapshot.yml, fetches the natives from S3 and compiles and deploys the Maven artifacts SNAPSHOT builds to SonaType. Triggered by a push to the master branch.

I'd much rather see this as a single condition to the end of the shared runner which does everything. If we really want we can create a cache system which detects if jnigen output changed at all before building.

An example of conditional publishing can be found here (The conditions and logic itself, not the gradle part): 
https://github.com/libgdx/gdx-jnigen/blob/master/.github/workflows/main.yml#L41-L60
with the required release trigger:
https://github.com/libgdx/gdx-jnigen/blob/master/.github/workflows/main.yml#L6-L7

> Splitting native build into separate tasks for OS/project

 This is a lot more fragile and not needed, I would much rather see just using `jnigenBuild` and letting gdx-jnigen handle which can be built. Having it all from one runner would be a lot faster and simpler.

badlogic:
> Just to clarify, you've taken out the openal build and are just hosting them statically?
@Tom-Ski no, OpenAL is build along with everything else, see https://github.com/badlogic/libgdx/blob/master/.github/workflows/build-natives.yml#L42, which will eventually do this: https://github.com/badlogic/libgdx/blob/master/build-mac-ios.xml#L16

> I think there's a lot of benefit in building natives on every CI, using ccache should lower the actual duration.
All platforms build in under 10 minutes without ccache, except for macOS. Historically, ccache didn't work for macOS due to the way the macOS/iOS native builds are setup, i.e. each arch emits to the same dir, smoking the previous arch's object files. This needs to be fixed in gdx-jnigen, for which I currently do not have the time.

@PokeMMO I'm OK with just triggering the native build on every push to the repo. Since workflows can't really depend on each other, both the release-snapshot and build-natives workflow have to be merged into one workflow, which kinda sucks. I'll see what I can do.

> Final result would look like mac-os runner first, which compiles macos/ios and then uses the uploads artifacts as "macos-natives". ubuntu runner which depends on macos runs next, which downloads the "macos-natives" artifacts and compiles linux/windows/android binaries. This gives a lot of flexibility for forks and PR to test things since the artifacts can be downloaded manually from the action result.

I'm strictly against serializing the natives-xxx jobs on natives-macos. That increases the total build time by 25 minutes, for a total of 50 minutes instead of 25 minutes. IWe will have to make changes to the workflows in the future, so low overall build times are a must.

If a fork or PR wants to test things, they can simply copy the workflow files and modify them for their needs. This should not be our concern. The workflow files are for us maintainers to 1. smoke test PRs, 2. build natives and deploy them and 3. release snapshots.

> I'd much rather see this as a single condition to the end of the shared runner which does everything. If we really want we can create a cache system which detects if jnigen output changed at all before building.

This should be resolved by merging build-natives.yml and release-snapshot as outlined above.

> Splitting native build into separate tasks for OS/project
> This is a lot more fragile and not needed, I would much rather see just using jnigenBuild and letting gdx-jnigen handle which can be built. Having it all from one runner would be a lot faster and simpler.

No, that is not true. My first iteration did exactly that and it was massively slow. The split into one build per platform must remain in build-natives.yml. However, maybe I can coerce Gradle to just build all e.g. `jniGenBuildLinux64` tasks of all projects in one go. That is about the only brittleness there is.

To summarize:
* I'll amend this PR to combine build-natives and release-snapshot
* I'll remove the brittleness of calling subproject jniGenBuild goals if possible.

mgsx-dev:
Could you please update CHANGES file? Breaking changes should be on top with "BREAKING CHANGE" tag, others changes as usual.
Something like this: 
```
- [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
- [BREAKING CHANGE] Linux x86 (32-bit) support has been removed. No migration steps required. 
...
- Architecture support: Linux ARM and AARCH64 support has been added. The gdx-xxx-native.jar files now contain native libraries of these architectures as well.
```

Some other required changes:
* gdx-setup template have to be modified.
* Jenkins badge can be removed from main README file

Reminder after this PR get merged : 
- update the wiki : https://github.com/libgdx/libgdx/wiki/Dependency-management-with-Gradle

badlogic:
@mgsx-dev thank you for the suggestions, I've made the required changes.

@PokeMMO I've decreased build times by further splitting up the native builds and made the other changes outlined in my last comment. We now have a build-snapshot.yml files, which builds everything including natives on a push to master and uploads a snapshot release. And we have a test-pullrequest.yml, which is invoked on pull requests and builds the Java bits only. It can be made to build the natives as well, but that would require copying everything over from build-snapshot.yml. I'm not convinced this is necessary.

If there are not further change requests, I'd like to merge this and pull down my Jenkins server.

PokeMMO:
You can still use the multiple concurrent runners with the artifacts action, just the final pack step depends on all of the other runners.

Either way, this can always be improved later, I'm fine with merging it as long as it does what you want it to.

PokeMMO (approved):

Tom-Ski:
@badlogic Yeah but its not building the 32 bit slices. They have been removed from the script, and the version of xcode wont support building that.

Tom-Ski:
I did not see any mention of ios 32 bit being dropped. And it would be good to keep it, even if its old version :(. Xcode > 9 wont build 32 bit

badlogic:
@Tom-Ski ah, I forgot to mention that iOS 32-bit is gone. The last iOS version to support 32-bit apps was 10. The last 32-bit only iPhone was the 5C from 2013. For iOS it is more than save to drop 32-bit support.

Similar story for iPads, the last 32-bit device was the iPad 4th gen from 2012.

And AppleTV 3rd gen (2012) was the last 32-bit device from that category.

So, for all our Apple targets, 64-bit seems extremely save, as none of the 32-bit devices are supported anymore (which as I understood also means no access to the app store).

Good catch with the min-os flag, I'll fix that up.

badlogic:
Alright, all fixed up. I'll let this cook until the weekend and would like to merge it by then.

badlogic:
Here's why we shouldn't build natives on every push https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

obigu:
Thanks for your work on this @badlogic .

Just wanted to give some numbers on a real game regarding the percentage of users using 32 bit devices on iOS (that can still download and update apps without issues from the AppStore as long as include 32 bit archs). These are the numbers of active users for a period of the last X days for a game that is most popular on US with a mixed audience (kids usually use older devices). If I haven't made a mistake selecting the 32 bit devices, the percentage of users with a 32 bit only support is almost 2%.

image

If that number is low enough to be acceptable is something we will probably disagree on but, at least, we have a number to measure the impact of the decision.

badlogic:
Most popular with 80k users in the past X days? What ismthe source of this data?

obigu:
Sorry for my bad English, it's my game and the country it's more popular in is the US (that is relevant as richer countries tend to have a lower percentage of low end devices).

badlogic:
I've made additional changes.

* The jobs use GitHub Action artifacts instead of S3 to pass build artifacts between them.
* The GitHub Actions build does not require S3 or Maven credentials. If not supplied, upload to S3 and SonaType are skipped. That way forks work out of the box and can build their own natives using GitHub Actions
* The natives are now packed into a .zip. This reduces the number of PUTS/GETS to/from S3 so we can stay on the free tier.
* The Maven build now uses the fetch.xml Ant script to download natives from S3, simplifying the overall build.
* The Ant fetch.xml script has been simplified and is a lot faster, as the number of requests has been significantly reduced.

This leaves two open questsions:
1. iOS 32-bit support. @obigu I understand your concern. I'm unsure if 32-bit support should be removed. It is currently not possible to build via GitHub Actions except with a self-hosted runner, which would make this whole effort void.
2. Whether to build natives on each push. The macOS/iOS builds take 28 minutes, all other builds take 33 minutes together. The macOS minutes are multiplied by 10 for quota purposes by GitHub. The total number of minutes is thus 283 per build. The libGDX organisation has a monthly quota of 2000 build minutes. We can thus do 7 full builds per month within that quota. If we build natives on each push, we'll go above the quota in a few days. Each additional full build would cost us 28 minutes * $0.08 (macoOS/iOS, no minutes multiplier) plus 33 minutes * $0.008 (all other builds), for a total of $2.5 per build. In the past month, there were 25 commits. 7 builds for 7 commits would have come out of the quota, 18 builds would have been billed, for a total cost of $45 for that month. 25 commits in a month is on the low-end.  50-70 commits seems more likely. That'd cost $107-$157 per month. The native code does not change often. I'd therefore suggest to have a Java only build triggered on each push that releases a snapshot to SonaType, which fits easily into our quota. Additionally, we have a native + Java code build that is triggered manually in case the native code (or gdx-jnigen) changed. Alternatives:
  1. Reduce the native (macOS/iOS) build times. I see no obvious way. Only building if specific code is changing is very brittle, so we need to perform the full native build. ccache is an option, but that would require non-trivial changes to the iOS build.
  2. Using a self-hosted macOS runner. in this case I might as well just build everything on Jenkins again. The security implications of a self-hosted runner for a public repo are pretty bad. I could not host it on any of my networks and would have to dedicate an entire clean machine to it.

My answer for 1. is to kill 32-bit iOS support. My vote for 2. is to try ccache, and if that fails to give us sensible build times that fit within our budget, move to manually triggered native builds.

PokeMMO:
Does the quota actually apply to open source projects? My understanding it was only for private repositories.

badlogic:
The quota applies. 2000 minutes. See https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions.

You can also see quota usage under the Billings settings. Unhelpfully, they aren't up-to-date at all...

badlogic:
Oh, you're right. WTF Mario.

> GitHub Actions usage is free for both public repositories and self-hosted runners.

obigu:
@badlogic I'm ok with killing iOS 32 bit support if it's going to be a stopper, it's Apple's fault for making it so hard. 
I suggest we make sure we communicate properly on the release notes/wiki, etc... that libGDX 1.9.13 is the last version to support 32 bits for those of us who are still interested on supporting those devices. We will be able to hold for a while until Apple breaks something else that forces us to update.

badlogic:
@obigu I'll host the last 1.9.13 natives on S3 as well. As long as no native code changes, those will continue to work.

obigu:
Do we need tvOS? AFAIK RoboVM doesn't currently support building for tvOS. Maybe we can keep it disabled until/if it does?

badlogic:
Oh, good point. I think tvOS can go for now! I'll have to modify gdx-jnigen
for that though.

Tom-Ski:
Yeah, I think we just kill it 32 bit too. Its too much of a PITA to keep building it with other means anyway, and impossible as we are going this route. As long as we can keep providing the old natives and an easy way to hook those up in latest versions.

badlogic:
Ok, I guess we've decided then! Here's what needs to happen:

* Remove tvOS support from gdx-jnigen. Should just need to modify the ant build-ios.xml template in gdx-jnigen, and modify build-objectal.sh script in gdx.
* Upload the last natives from 1.9.13 to S3
* Update CHANGES with instructions on how to use the old natives with current libGDX.

If someone could help me with CHANGES that'd be great. It's been a while since I messed with a gdx project Gradle. What I'll upload to S3 will be a .zip of this directory https://libgdx.badlogicgames.com/ci/nightlies/dist/. The easiest way is likely just to fix this dependency to 1.9.13 in your project's Gradle setup and call it a day. https://github.com/libgdx/libgdx-demo-invaders/blob/master/build.gradle#L90

Tom-Ski:
We can modify the build-objectal.sh to fetch the old openal 32 bit slices from somewhere statically hosted, and then combine with lipo like we do for everything else. That way we keep compatibly going further, and no need for messing with local jars or hacking latest jars. 
Otherwise if we change other gdx natives, since we combine the static libraries they can no longer use the 1.9.13 version of the natives jar.  This way it will continue working even if we change the gdx natives, no need to change version, and we can keep it for as long as we want.

badlogic:
I'm against this. It would mess up the pretty clean build again, partially
nullifying the work. I think it's not too big a burden for users to change
1-4 dependencies of their Gradle build to a fixed version. There's no
messing with jars and lipo involved.

Tom-Ski:
No need to mess with the jars, its just bringing back the slice in the bash script from a wget instead of building it.


The problem is that the natives are bundled with the gdx natives. As soon as they change, you can't use 1.9.13 anymore and would have to hack the natives-ios jar instead of just modifying the trivial gradle build dependency version. 

We already have this case now, recent changes to ios gl30 implementation are not in 1.9.13. If a user wants to support a 32 bit device, but also use these features, they can't. They would have to download the jar and modify the contents to replace the static openal binary with the one from 1.9.13.

badlogic:
I don't understand. If the natives change, the statically hosted 32-bit slices can't just be lipoed in anymore. Their API and/or semantics will not match the latest natives, and we won't be able to recompile them. Unless you want to also keep a machine with Xcode 9 running somewhere. Is that what you are actually suggesting? Periodically compile the latest natives outside the GH Actions build and statically host them? I'd rather not do that, as it introduces multiple points of potential error.

badlogic:
Also, all this doesn't just apply to ObjectAL, it applies to all other natives as well, unless I'm missing something.

Tom-Ski:
I see you removed iOS 32 bit from jnigen the other day. I didn't know you did that, so perhaps this is why there is confusion.

I don't think we had any issue building gdx natives on 32bit for ios, I just checked the logs and they built fine, we also had users that could run still on 32 bit devices but only stumbling at the objectal point.


So my point was that we could still have built 32 bit gdx core natives as usual (no build issues or changes required), and then do this shim for 32 bit objectal so they get combined into the fat libObjectAl.a static binary.  At the end we package the libObjectAL.a (made up of the 32bit old .a, and the newly compiled other slices) and libgdx.a into the natives-ios jar like normal and people get compatibility for as long as objectal does not change on all future libgdx versions.

As far as I know, please correct me if I'm wrong, but there is only a problem combining the 32bit slice for objectAL with the other compiled slices (am64 and x86_64) if there is an API change. Which there hasnt been, and probably most likely wont be for a while. So this seemed like a pain free way all around to still support these devices. 

If that's not possible, then we just have to drop official support for ios32 bit here, and people cannot upgrade their libGDX safetly past 1.9.13 if they want to still support those devices.

badlogic:
Understood. I'll add iOS ARM32 support back in, remove tvOS support, and commit prebuilt ObjectAL binaries for arm32/arm64/x86_64 to the repo, which will be copied over to `gdx/libs/ios32` in every build. As long as ObjectAL doesn't change we should be good.

PokeMMO:
Why are we still concerned with minutes when minutes are free for open source repositories?

badlogic:
I made all the required changes. ObjectAL is still build for ARM32 with latest Xcode. I found no problems with that.

@Tom-Ski @PokeMMO @mgsx-dev let me know if this is good to merge and I'll do a squash.

PokeMMO:
Looks good to me.

mgsx-dev:
@badlogic could you please make a PR removing these snapshots [here](https://github.com/libgdx/libgdx/pull/6445/files#diff-02b139e755ab0e27e0b0a6f9845843ef189116955cc340c49a4022587dbb2b52R29) and [there](https://github.com/libgdx/libgdx/pull/6445/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R49)
To be merged when Lwjgl 3.3.0 is out. I hope it'll be released soon because it blocks the next libGDX release if i'm not wrong.

badlogic:
@mgsx-dev Whoopsies, fixed.

yajirobe69:
I'm sorry to drop the IQ points of this thread a bit here, a fascinating read even though it all goes over my head. Just want to put in a request that if/when iOS-32bit support is removed, now or later, can you please give us a final stable build verison just prior which still support it (as it is now I believe).

badlogic:
iOS ARM 32-bit support stays!

yajirobe69:
Even better!

crykn:
@badlogic it seems like built versions of gdx-setup are no longer uploaded to https://libgdx.badlogicgames.com/ci/nightlies/dist/ (or anywhere else for that matter).

Tom-Ski:
Yes we are working on it. They'll be hosted soon

tommyettinger:
If anyone is waiting on gdx-setup or wants LWJGL3 in their setup, there's always [gdx-liftoff](https://github.com/tommyettinger/gdx-liftoff/releases/tag/v1.10.0), which is up-to-date for 1.10.0.

PR #6451, opened 2021-03-12
enable GL errors in tests via commandLine argument, and fix a test

mgsx-dev:
Now we can enable GL errors logging when running tests, helpful for debugging.

i also fixed the VertexBufferObjectShaderTest : 
- VertexBufferObjectWithVAO must be used when gl30 is enabled
- `gl.glEnable(GL20.GL_TEXTURE_2D)` is only valid with fixed function pipeline (which is not the case with libGDX).


MrStahlfelge (approved):
Looks good to me

crykn (commented):
I don't think the formatting for the if-blocks in `VertexBufferObjectShaderTest` and `GdxTestWrapper` is correct, but apart from that LGTM.

mgsx-dev:
We usually don't format tests. If this is something we want to do from now, i could format my changes in tests, not a problem, just tell me.

crykn:
Dunno, I think it would make sense. But tbh it doesn't really matter.

PR #6455, opened 2021-03-15
Fix #6454

### LINKED ISSUE #6454: gdx-setup_latest does not produce a correct iOS project
#### Issue details
If the latest gdx-setup_latest is downloaded from here: [project-generation](https://libgdx.com/dev/project-generation/) and then is used to create an iOS project, it will have several problems that likely will be of big problems for most people that page is intended for. 

First of all the version of libGdx that the gdx-setup specifies will make the iOS project crash when packaging it. Seems to be a know problem, but for people going to that page it most likely isn't know. 

Secondly, if you manage to figure that out and reduce the gdx version to 1.9.12 for example it will still have problems. The project generated apparently does not follow whatever changes that has been made for iOS since 1.9.12. The app will only display on the lower quarter of iOS devices.

#### Reproduction steps/code
Download gdx-setup_latest from here: [project-generation](https://libgdx.com/dev/project-generation/) and attempt to build and it will fail. That problem is known, but it would probably be a good idea to not have a newbie tool available that produce faulty projects. 

Secondly, if the gdx version is changed to 1.9.12 and the image is moved to 200,200 in the Game class specified in the gdx-setup gui.

change
`		batch.draw(img, 0, 0);
`
to
`		batch.draw(img, 200, 200);
`
or any suitable position so that it would reach over the middle of any iOS device to test on.

Then build and deploy in device and the image will be clipped to the lower left quarter of the device

It would probably be good to try to make sure that the project generator really creates projects that can work, so changing the gdx version back to 1.9.11 could be good and document why.

Introducing a change (in what typically would be though of as a bug release) that seems to be breaking many if not most games on iOS does not seem like a very good idea. It is also not clear how to migrate to the new version so that iOS games still render normally. At least it would be good to update the template that is generated from the gdx-setup to work so that the whole screen renders.

1.9.12+
But the issue is really regarding gdx-setup and documentation of breaking changes.

#### Stacktrace
No crash but unexpected behaviour. See how right part of image is gone, since right half of screen is not rendering. Also top half is not rendering either.
Screenshot 2021-03-14 at 02 35 07

If app is started in landscape, then top part cuts the image given the location of it, but same here, both top and right part of screen is not rendering leaving only lower left quarter of screen rendering:
IMG_3406

- [ ] Android
- [x] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

obigu:
When we added support to HDPI mode on iOS we missed changing a `glViewport` call that is still receiving logical pixels (glViewport always requires physical pixels as explained in `HDPIUtils` javadoc). 

It is  noticeable as shown in #6454 on a simple project with no `Viewport`. When using a `Viewport` this problem did not reproduce because it makes its own `glViewport` calls on `apply()` implementation.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
@@ -28,6 +28,7 @@ import com.badlogic.gdx.graphics.GL30;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.glutils.GLVersion;
 import com.badlogic.gdx.graphics.glutils.HdpiMode;
+import com.badlogic.gdx.graphics.glutils.HdpiUtils;
 import com.badlogic.gdx.utils.Array;
 
 import org.robovm.apple.coregraphics.CGRect;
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
Looks clean; thanks for documenting the different reasons why one glViewport uses HdpiUtils and one glViewport doesn't. Approved!

MrStahlfelge (approved):

PR #6456, opened 2021-03-16
Update desktop package name to look like other projects

kendaozinho:
Update desktop package name to look like other projects (Android, iOS, HTML and core).


MrStahlfelge (approved):
Since no one has made any objections, I approve because I like the change

PR #6457, opened 2021-03-18
Add copy of macosx natives for box2d maven build

MobiDevelop:
Looks like this was just missed when switching over to Github Actions. Current snapshots are missing the macosx dylib.


badlogic:
Oh Jebus, sorry.

MobiDevelop:
Ha, no worries. Just wanted to make sure there wasn't a reason it was left out.

PR #6465, opened 2021-03-21
Fix #6464: BaseTmxLoader now imports default property "Type" properly.

### LINKED ISSUE #6464: TiledMap tile property "Type" doesn't get loaded
#### Issue details
The `Type` property on tiles gets skipped [during loading](https://github.com/libgdx/libgdx/blob/bc314368a18e0c916437f3dd7b4172714e6abf8b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java#L635) and therefore isn't available in libgdx.

image

```xml
<tileset firstgid="1" name="Test" tilewidth="256" tileheight="256" tilecount="4" columns="0">
  <tile id="0" type="AAAA"> 
    <image width="32" height="32" source="Start.bmp"/>
```

#### Reproduction steps/code
* Create a new Tilemap in Tiled
* Add a Tileset
* Assign a Type on one of the Tiles
* Load the map into libgdx
* Type isn't available

image

1.9.14

- [x] Android
- [x] iOS
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

lyze237:
Fix for #6464 

Only downside to that is that this could break some maps which use a custom "type" property.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,7 @@
 - Fix for #6377 Gdx.net.openURI not working with targetSdk 30
 - Api Addition: Added a Pool#discard(T) method.
 - Architecture support: Linux ARM and AARCH64 support has been added. The gdx-xxx-natives.jar files now contain native libraries of these architectures as well.
+- Fix for #6464: BaseTmxMapLoader now imports default property "Type" properly.
 
 
 [1.9.14]
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@
 - Api Addition: Added a Pool#discard(T) method.
 - Architecture support: Linux ARM and AARCH64 support has been added. The gdx-xxx-natives.jar files now contain native libraries of these architectures as well.
 - Fix for #6464: BaseTmxMapLoader now imports default property "Type" properly.
+- TiledMap: Default properties in tmx files get loaded with \tname now when a custom property exists with the same name.
 
 
 [1.9.14]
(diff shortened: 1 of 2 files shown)


tommyettinger:
Is it possible for existing Tiled maps to have a property called "type" and also an attribute "type"? If so, this is a potential breaking change. Is there any way to handle the attribute separately from properties? That way, it preserves the structure of the TMX file and permits having both kinds of "type".

tommyettinger:
From what I understand, @lyze237 has some improvements for this planned, to be done when he has time. Hoow will this talon-ted young programmer resowlve these minor issues? We'll have to wait and see!

tommyettinger:
Just as a commentary on what was discussed on Discord, the tab character was chosen for the tile attribute to avoid overlap with the custom property that could be called "type", because it's virtually impossible that users could accidentally have the custom property "\ttype" in a real map -- the tab key changes focus and doesn't enter a tab, so it would have to be entered in code. This is all a situation that would require a user to look into the map handling code anyway, I think, so some code docs there are probably the main necessary thing. Maybe later a wiki entry can be edited about corner cases in Tiled.

mgsx-dev:
why not adding a type property to TiledMapTile (and implementations)? It would be a lot cleaner than messing with custom properties. This tab character is a bad idea imo.

lyze237:
gonna close this for now as i don’t have any intentions to look into that further. since this was supposed to be a fix in the same style as all the other properties and nothing not really.

PR #6466, opened 2021-03-22
Align add test coverage

ColinGeukes:
In this pull request, the following has been done to the file `utils/align`:

- Added the test coverage 

The code coverage has been improved as follows

| `utils/Align` (old)   | 0% | 0% | 0% |
| `utils/Align` (new)      | 100% |   100% | 100% |

Also I found something strange with the isCenterVertical function, it gives true whenever the align is Left, Center, Right (which is horizontal rather than vertical) But this might be the case you wanted to function to be. Since there is lacking documentation I do not know that.

The changes to Align documentation are reverted but an artifact is left as I cannot push spaces to git it does not recognize it as any change.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Align.java
+++ b/gdx/src/com/badlogic/gdx/utils/Align.java
@@ -16,113 +16,58 @@
 
 package com.badlogic.gdx.utils;
 
-/**
- * Provides bit flag constants for alignment.
- *
- * @author Nathan Sweet
- */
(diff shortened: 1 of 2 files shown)


Darkyenus:
I don't think we need tests for something that basically never changes, is fairly trivial to check for correctness and is already well tested. These things will just exhaust our CI allowance, which could be used for something more useful. But others may have a different opinion.

PR #6468, opened 2021-03-22
DataBuffer add test coverage

ColinGeukes:
In this pull request, the following has been done to the file `utils/DataBuffer`:

- Added the test coverage 

The code coverage has been improved as follows

| `utils/DataBuffer` (old)   | 0% | 0% | 0% |
| `utils/DataBuffer` (new)      | 100% |   100% | 100% |


NathanSweet:
Tests add to the maintenance burden of the library and should only be added for things that are complex or relatively easy to break when changed. It is not a goal to have test coverage for all lines of code. I would say DataBuffer does not need tests.

PR #6472, opened 2021-03-25
Fix #3935 and other improvements

### LINKED ISSUE #3935: Music getPosition returns always 0.0 after played at least one time (on Android)
this happens when I first **play()** a music instance, **stop()** afterwards and **play()** again. In this case **getPosition()** returns always 0.0 even though the music is being played. This does not happen in the desktop application.

gdxVersion = 1.9.2
tested on Android 5.1.1
- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

obigu:
I've gone through our Android Music implentation and checked against MediaPlayer docs (https://developer.android.com/reference/android/media/MediaPlayer) and I've been able to make some tweaks that fix some of our long standing issues (such as #3935).

Main changes:
- `seekTo(0)` call has been removed from `stop()` and it's the line that caused #3935. This is not needed as stop() already resets the position indirectly as to play it again, `prepare()` must be called. The commit that added it is ec5ab7a323941c441acf724865cf961771a4a41e , @NathanSweet any idea what it fixed? On 2011 some Android devices had all kinds of strange issues and this might have been a workaround for one of them.

**Edit**: Found the probable cause for the fix on https://github.com/libgdx/libgdx/pull/3399 (I'm a fish, it's a good thing I let stuff written for my future me). The SO links point to Froyo problems so it should be safe to remove.

- Removed `isPlaying()` checks from `play()` and `pause()`. According to the documentation calling `play()` when player is on Play state or `pause()` when it's on Pause state has no effect so we avoid an unnecessary JNI call but I think there's a more important thing it fixes. According to the docs :

`Note that the transition from the Started state to the Paused state and vice versa happens asynchronously in the player engine. It may take some time before the state is updated in calls to isPlaying(), and it can be a number of seconds in the case of streamed content.`

This means that if we we have a paused Music and call `play()` and `pause()` in a short interval of time, it may happen that on the `pause()` call, the `isPlaying()` check returns `false` and the method is not executed.
- Improvements to the `MusicTest` that was impossible to run on a device where everything looked tiny.

Playing with the `MusicTest`, hammering buttons, positions slider, etc... on a real device (Samsung S7) I haven't been able to break it while before it was pretty easy to make it fail.


NathanSweet:
I don't remember why the seek was added, sorry. I don't see that the [prepare](https://developer.android.com/reference/android/media/MediaPlayer#prepare() docs) mention it changes the playback position, so it isn't clear if the seek is needed. Maybe it's buried in the rest of the docs? Even if you see `prepare` seeks to 0 on your device, it could have had different behavior in MediaPlayer previously or on some devices.

obigu:
Yeah, it's not explictely mentioned in the docs and in one of the SO links it shows that some specific device (unsupported Gingerbread 2.3) was not resetting the position (https://stackoverflow.com/questions/6849045/mediaplayer-instance-stop-behaves-like-a-pause) so it may very well be it was added for those buggy devices.

Expected behaviour on any player (from a Walkman to VLC Player) is that Stop resets the position of the music/video to the beginning and that is consistent with the Android devices I've tested.

To make sure the changes don't break on weird devices I've made a release of one of my games (min SDK 19) that uses Music intensively that uses a custom libGDX with this PR merged. Most of the users are kids with devices you've never heard of, if there are devices out there that don't properly set position to 0 on stop, we will hopefully know.

obigu:
@NathanSweet I've just found confirmation on the `MediaPlayer` docs that Music position should be set to 0 on `start()`
`Starts or resumes playback. If playback had previously been paused, playback will continue from where it was paused. **If playback had been stopped, or never started before, playback will start at the beginning.**`

obigu:
It's been 2 weeks since I released this fix on one of my apps and nobody has complained about Music issues (no crashes either) so I think it's safe to merge after review.

PR #6475, opened 2021-03-31
Fix for iterator() of IntMap.Values and LongMap.Values, plus a test.

tommyettinger:
As Discord user vladimir reported, there can be a discrepancy between the keys and values in some situations involving remove() called on the iterator() of IntMap.Values and LongMap.Values. In my testing, I was only able to confirm this for the zero key and the associated field, `zeroValue`. Looking at MapIterator for both of those classes, the `zeroValue` isn't assigned `null` when it is removed via the iterator. This has some implications for GC, and also affects the returned value for `put()` when placing back into the zero slot. I believe this also affected IntMap.Keys and LongMap.Keys, but the fix applies in the MapIterator code both Values and Keys share. This bug shouldn't affect IntIntMap or IntFloatMap for GC purposes because their zero values are primitive, nor should it affect ObjectMap because it doesn't have a `zeroValue`. IntIntMap and IntFloatMap return a default value instead of how IntMap works, so their put() return value isn't affected by the bug.

The only fix necessary is to assign `null` to the `zeroValue` of IntMap and LongMap when removing via an iterator. This is one line in IntMap, and one more in LongMap. The rest of this is a test that shows the error if you omit the latest commit, and shows the fix when the latest commit is included.


NathanSweet:
Good stuff as usual, cheers!

PR #6479, opened 2021-04-04
fix default shader bug

intbyte-100:
fixed a bug due to which shadows painted over  all the lights
before
after


Tom-Ski:
Please revert any formatting changes you make on the entire file

mgsx-dev (changes requested):
Your proposal can't be accepted, it's more like a hack for your current use case than a fix.
If you want to have softer shadows, you typically need to use an AMBIENT_LIGHT in your environment.

This limitation is because all lights are computed in vertex shader (Gouraud shading) into the same variable.
A real fix would be to have separated light information: one for shadow light, and another for all other lights. And apply shadows only for that light.

mgsx-dev:
feel free to open a new PR with a proper fix, i'm closing this for now.

PR #6484, opened 2021-04-06
Fix wrong sizes provided to glViewport and listener on Physical HdpiMode

obigu:
Thanks to an issue raised on Discord by M1 we've noticed https://github.com/libgdx/libgdx/pull/6455 fixed `Logical` HDPI mode but messed `Pixels` just because I didn't test properly. Now everything works as expected, tested on both modes. Sorry for all the changes on iOS HDPI mode support, this one should be the last one.

@Tom-Ski We should not release until this is merged.


tommyettinger (approved):
I like fixes. What's different about the gl20.glViewport() call above and the Gdx.gl.glViewport used after (which was HdpiUtils.glViewport() )? I'm guessing both are needed, but the Gdx,gl one is correct while HdpiUtils is not?

obigu:
No difference in this case, I've replaced it to `gl20.glViewport()` to make them consistent.

PR #6487, opened 2021-04-06
Fix #6486

### LINKED ISSUE #6486: Android Insets not updated on 180º rotations
User gerdsOn on Discord has reported that safe insets are not updated when rotating the device 180º on a landscapeSensor app. The reason is that `onSurfaceChanged()` is not triggered on that case a `width` and `height` don't change which makes sense as, until recently, screens where symmetric. This probably occurs on `portraitSensor` as well.

The solution may be to move the insets update to a `DisplayListener.onDisplayChanged()` listener that is triggered on any orientation change including 180º (It is API > 17 but that's ok). A resize() should also be triggered on this case for the app to react to the change but we need to make sure it's not triggered twice for 90º rotations.

Relevant SO post https://stackoverflow.com/questions/9909037/how-to-detect-screen-rotation-through-180-degrees-from-landscape-to-landscape-or

obigu:
This is a draft proposal to fix #6486.

Changes that require updating ppi, updating insets and calling the resize listener may be triggered both by `onSurfaceChanged()` and by `onDisplayChanged` (such as a 180º orientation change). The DisplayOrientationChange object keeps a boolean that gets evaluated once per frame and updates as necessary. In order to support older APIs there are 2 implementations.

Some more testing is required, for the moment I'll keep in draft. If feedback is positive and seems like a good idea I'll add proper javadocs and run more tests.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -69,6 +69,7 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 	EGLContext eglContext;
 	GLVersion glVersion;
 	String extensions;
+	DisplayRotationHelper displayRotationHelper;
 
 	protected long lastFrameTime = System.nanoTime();
 	protected float deltaTime = 0;
@@ -103,6 +104,11 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 		this.config = config;
 		this.app = application;
 		view = createGLSurfaceView(application, resolutionStrategy);
+		if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+			displayRotationHelper = new DefaultDisplayRotationHelper(app.getContext());
+		} else {
+			displayRotationHelper = new LegacyDisplayRotationHelper();
+		}
 		preserveEGLContextOnPause();
 		if (focusableView) {
 			view.setFocusable(true);
@@ -128,6 +134,7 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 	}
 
 	public void onPauseGLSurfaceView () {
+		displayRotationHelper.onPause();
 		if (view != null) {
 			view.onPause();
 		}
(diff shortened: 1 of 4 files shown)


MrStahlfelge:
Not sure about introducing two new classes with calls etc. Wouldn't it be simpler if the main graphics class would manage listening and getting the new callback?

obigu:
The `DisplayListener` is only available for API > 17. The idea of creating the interface is to keep `AndroidGraphics` cleaner by avoiding the conditionals that decide whether to execute legacy or new implementation logic every time orientation related code is needed. There are 4 points that would require them (listener creation, onResume, onPause and onSurfaceChange) and I think, it would look a bit messy. Maybe I can make them inner classes?

MrStahlfelge:
It is a matter of taste how to do it. In my opinion, control flow is very hidden here and I personally don't mind some more conditionals of that helps understanding what's going on. 

Would be interesting to know what others think, but as this is an iOS only topic it is most likely that no one will comment.

obigu:
> Would be interesting to know what others think, but as this is an iOS only topic it is most likely that no one will comment.

Android topic :) Hopefully somebody else will chime in

MrStahlfelge:
Ah yes, we both discuss too much ios topics. However, screen insets and a bug only appearing when rotating the device 180° don't let my hopes go up, but let's set some labels. ;)

MobiDevelop:
I think the DisplayListener implementation at least needs to be a new class, to not break older versions. So the question that remains I guess is should this be interfaced to put the legacy logic also into a class - it's probably fine, not sure it's necessary.

jkazma-logisk:
Most landscape apps support reverse landscape as well, and with newer phones having asymmetrical safe area insets, I think this feature would be useful.

obigu:
> What can I do to help get this merged into master? I've ran into a scenario where this is needed for my app.

You can test it by checking out the pull request locally. Easiest way is modifying or creating a new libgGDX `libgdx/tests` but you can also build a custom libGDX version and test it on your project.

MobiDevelop:
My Pixel 3XL supports portrait upsidedown mode.

MobiDevelop:
I don't think so, I just have auto-rotate on.

obigu:
Ok, if I explicitly set "reversePortrait" it does work but it doesn't change orientation with "sensorPortrait" as the rest of devices. The only app in the Simulator that changes orientation upside down on the Pixel 3XL is the Camera. I think it there's something else apps need to do to support portrait upside down autorotation on the Pixel 3 XL (and probably other notched devices) and it makes sense, notched iPhones don't support portrait upside down either.

obigu:
Closing this as it doesn't seem to be a good solution

PR #6492, opened 2021-04-09
Add MathUtils.norm(double, double, double)

soundasleep:
Since we have clamp(float, float, float) and clamp(double, double, double), it seems only reasonable to have a double version of norm(float, float, float).

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/MathUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/MathUtils.java
@@ -318,6 +318,15 @@ public final class MathUtils {
 	static public float norm (float rangeStart, float rangeEnd, float value) {
 		return (value - rangeStart) / (rangeEnd - rangeStart);
 	}
+	
+	/** Linearly normalizes value from a range. Range must not be empty. This is the inverse of {@link #lerp(float, float, float)}.
+	 * @param rangeStart Range start normalized to 0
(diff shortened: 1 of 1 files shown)


tommyettinger:
I'm not opposed to this, but very little of libGDX uses `double`s. For example, there's a `FloatArray`, but no `DoubleArray`, despite there being every other primitive type of Array. There's also no `lerp()` defined for `double` arguments, which is the logical counterpart to `norm()`. In general, the library strongly prefers using `float` because OpenGL works better with it than with `double`, and because GPUs are typically much slower at processing double-precision floating-point math than they are single-precision. What's your use case for `norm(double, double, double)`?

tommyettinger:
Yeah, I've encountered some resistance to using float literals and float math before. Typically the people who strongly prefer doubles also come along with other requirements for code that differ from how libGDX does things, and it isn't worth catering to this particular requirement. That said, I'd probably approve this if it covered more of MathUtils' common methods with double versions, though the scope would get tricky if the double-math spread into other classes.

soundasleep:
That's all good! I didn't know `MathUtils` was only intended for math functions used primarily by libgdx or within libgdx's point of view.

In my current project I use doubles for a bunch of stuff and I made a copy of `MathUtils.norm` for doubles, and thought others might like having that functionality in new projects, thus the PR.

Feel free to close!

PR #6495, opened 2021-04-11
Update gdx version.

crykn:
Seems like this version wasn't adapted after the last releases. Happened here (3ec1d53739f919a4b4690116b4d4cf8803be3b08) as well.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -3,7 +3,7 @@
 	<property environment="env" />
 
 	<!-- library version -->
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
Looks good to me. I'm going to do a big search for any usage of `1\.9\.(1?)\d` and see what might be left; hopefully there's nothing more. EDIT: You got everything taken care of. This is the last usage of 1.9.13; 1.9.14 is only used by gdx-setup, where it's currently correct.

PR #6496, opened 2021-04-14
check bones count when creating shaders

mgsx-dev:
It's a common pitfall to have more bones in a model than what is configured (default is 12). In this case, you only see a model partially rendered without any obvious clue why (especially beginners).


Tom-Ski (approved):

dzikoysk (approved):

MrStahlfelge (approved):
That's a good change

PR #6500, opened 2021-04-16
Stop generated .gitignore ignoring essential Eclipse metadata files

soundasleep:
By default, the generated `.gitignore` ignores two essential Eclipse files: `.project`, and `.classpath`.

`.project` define a project name (user edited) and what types of builds are permitted.

`.classpath` defines the location of runtime libraries and how to execute specific builds.

`.settings/` defines specific IDE switches, e.g. how Eclipse's JDT operates, which compiler version to target, etc.

I'm proposing these changes because after a restart of Eclipse, all of a sudden I couldn't run JUnit tests in my project. Turns out both the `.project` and `.classpath` had changed and I had no idea, because they weren't version controlled.

### DIFF
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/gitignore
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/gitignore
@@ -33,8 +33,6 @@ com_crashlytics_export_strings.xml
 
 ## Eclipse
 
-.classpath
-.project
 .metadata/
 /android/bin/
 /core/bin/
@@ -45,7 +43,6 @@ com_crashlytics_export_strings.xml
 *.bak
 *.swp
 *~.nib
-.settings/
 .loadpath
 .externalToolBuilders/
 *.launch


mgsx-dev:
These files are generated by Eclipse gradle plugin and should not be tracked with git. Project name, dependencies, compiler version and all that are already configured in gradle files.

About your issue with JUnit, you should include it as test dependency in your gradle file, it works perfectly.

For very specific cases (when gradle configuration is not possible), you can always change .gitignore for your specific project but for most users this change is not wanted.

soundasleep:
Ahh. Thank you! Still a newbie to Gradle.

soundasleep:
For reference, this is one way to add JUnit to your Gradle project:

Add the following to `/build.gradle` for each project:

```
    dependencies {
        ...
        api "junit:junit:4.13"
    }

    test {
        useJUnit()
    } 
```

Run `gradlew eclipse` to recreate your Eclipse config files. (Change `org.gradle.jvmargs` from `-Xmx1500m` to `-Xmx1024m` if you're having difficulty creating enough heap space.)

If you have separate `test/` and `src/` folders, add this to each `build.gradle`:

```
sourceSets.main.java.srcDirs = [ "src/", "test/" ]
```

If you get weird ".project is out of sync" or other Gradle errors, try closing/opening the projects in Eclipse, or going Right Click > Configure > Add Gradle Nature, or going Right Click > Gradle > Refresh Gradle Project. Eventually it works.

mgsx-dev:
No, what you describe is not recommended : 
* you should use testImplementation scope for that dependency.
* you shouldn't run `gradlew eclipse` but use eclipse gradle plugin : right click / "refresh gradle project"
* you don't need to add gradle nature if you imported your project correctly using gradle plugin. see https://libgdx.com/dev/import-and-running/#importing-the-project

Please use Discord instead of the issue tracker for questions/help : https://libgdx.com/community/

PokeMMO:
See also #6369 for how we will be changing how eclipse project files work.

PR #6501, opened 2021-04-16
Add UtfFontTest to GdxTests

fabiitch:
Hi, i see test missing in GdxTests lists.
i also add more specials chars in test.

Thanks for reading

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/UtfFontTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/UtfFontTest.java
@@ -23,9 +23,13 @@ public class UtfFontTest extends GdxTest {
 	public void render () {
 		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
 		batch.begin();
-		font.draw(batch, "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØŒŠþÙÚÛÜÝŸ\n" +
-				"àáâãäåæçèéêëìíîïðñòóôõöøœšÞùúûüýÿ\n", 20, 200);
-		font.draw(batch, "⌘  ¢ß¥£™©®ª×÷±²³¼½¾µ¿¶·¸º°¯§…¤¦≠¬ˆ¨‰", 20, 100);
+
+		 //https://github.com/libgdx/libgdx/pull/6501#issuecomment-821749417
+		font.draw(batch, "\u0089 \u0065 \u0089 \u0074", 20, 400);//Missing chars should print ' e t'
+
+		font.draw(batch, "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØŒŠþÙÚÛÜÝŸ", 20, 300);
+		font.draw(batch, "àáâãäåæçèéêëìíîïðñòóôõöøœšÞùúûüýÿ", 20, 200);
+		font.draw(batch, "⌘¢ß¥£™©®ª×÷±²³¼½¾µ¿¶·¸º°¯§…¤¦≠¬ˆ¨‰", 20, 100);
 		batch.end();
 	}
 }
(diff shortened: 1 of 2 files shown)


tommyettinger:
UtfTest was only added to test a problem with this file and similar fonts: https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-android/assets/data/utf-font.fnt In those cases, the kerning in the font file was incorrect (you can see that the kerning mentions char 137, but there is no char 137 in the font), and libGDX previously broke when attempting to render it. I think to adequately test the issue, the problem kerning pairs should be entered as part of a String, even though the font can't render parts of the pair (the test may do this already, if it includes a raw Unicode char where it should have `\u0089`, but it looks like it doesn't actually test the missing kerning pairs). Rendering other Unicode chars doesn't help confirm any qualities of the tested code, since it already rendered a high-in-Unicode char.

fabiitch:
Yeah, its true add more characs dont add value to the test himself.

It wasnt my goal when i made this commit.
I also see gdx-tests like showcases for see gdx possibilities.
I imagine this screen glows with more chars. ^^

I can repush to reset screen to display only ⌘, and just add UtfFontTest to screens test list.
Or pr can be close. no problem.

Thanks a lot for your precisions

tommyettinger:
No, it's nothing wrong with what you've done, it's just upon looking at this test 7 years after it was added, I'm not sure it accomplished its original goal. Printing an extra line that tests the original problem would be good to add:
```java
"\u0089\u0065\u0089\u0074"
```
As long as that doesn't cause horrible errors like a crash when it writes the text, the test can be considered as passing. Ideally, it should show an unknown-character glyph, then "e", then another unknown-character glyph, then "t". If it doesn't do that, it should just show "et".

fabiitch:
Done i add new line, i reference your comment  in code for context.
Thanks for your feedback and kindness.

tests/gdx-tests/src/com/badlogic/gdx/tests/UtfFontTest.java
```diff
@@ -23,9 +23,13 @@ public void create () {
 	public void render () {
 		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
 		batch.begin();
-		font.draw(batch, "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØŒŠþÙÚÛÜÝŸ\n" +
-				"àáâãäåæçèéêëìíîïðñòóôõöøœšÞùúûüýÿ\n", 20, 200);
-		font.draw(batch, "⌘  ¢ß¥£™©®ª×÷±²³¼½¾µ¿¶·¸º°¯§…¤¦≠¬ˆ¨‰", 20, 100);
+
+		 //https://github.com/libgdx/libgdx/pull/6501#issuecomment-821749417
+		font.draw(batch, "\u0089 \u0065 \u0089 \u0074", 20, 400);//Missing chars should print ' e t'
```
tommyettinger, line 28: This almost works, but the spaces between chars mean that kerning still isn't tested. Please remove the spaces, then your ordeal is over 😄. I'll merge then.

(It should be `"\u0089\u0065\u0089\u0074"`)

tommyettinger (changes requested):
Almost there! Too many spaces! The kerning only happens if the two characters are next to each other, like `"\u0089\u0065\u0089\u0074"` but *not* like `"\u0089 \u0065 \u0089 \u0074"`. I think libGDX probably handles kerning correctly now, but if something breaks in the future this will act as a way to detect it. It's also possible that this will detect a current bug in kerning, who knows...

tests/gdx-tests/src/com/badlogic/gdx/tests/UtfFontTest.java
fabiitch: done ! Tthanks for babysit  XD

tommyettinger (approved):
Looks good! This might be the first PR accepted in 1.10.1-SNAPSHOT.

PR #6502, opened 2021-04-17
Vectors change check in nor() before normalise

fabiitch:
Hi !
This commit change for Vector2/Vector3 the check for see if Vector is already normalised.

I add 2 tests 
 - for check than len2 method can return number close to 1f and nor() needs to check with MathUtils.isEqual(len2,1f)
 - Check if Arthimetic exception can appears with nor(). (i think its useless, i see in testing that float variable divide by zero, dont throw Arthimetic but an infinite float.

Im not really sure about the tests pertinence.
Thanks for reading.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -113,7 +113,7 @@ public class Vector2 implements Serializable, Vector<Vector2> {
 	@Override
 	public Vector2 nor () {
 		 final float len2 = len2();
-		 if (MathUtils.isZero(len2) || MathUtils.isEqual(len2, 1f))
+		 if (MathUtils.isEqual(len2, 1f) || MathUtils.isZero(len2))
 			  return this;
 		 return this.scl(1 / len());
 	}
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -115,7 +115,7 @@ public class Vector2 implements Serializable, Vector<Vector2> {
 		 final float len2 = len2();
 		 if (MathUtils.isEqual(len2, 1f) || MathUtils.isZero(len2))
 			  return this;
-		 return this.scl(1 / len());
+		 return this.scl(1f / (float)Math.sqrt(len2));
 	}
 
 	@Override
--- a/gdx/src/com/badlogic/gdx/math/Vector2.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector2.java
@@ -112,12 +112,10 @@ public class Vector2 implements Serializable, Vector<Vector2> {
 
 	@Override
 	public Vector2 nor () {
-		float len = len();
-		if (len != 0) {
-			x /= len;
-			y /= len;
-		}
-		return this;
+		 final float len2 = len2();
+		 if (MathUtils.isZero(len2) || MathUtils.isEqual(len2, 1f))
+			  return this;
+		 return this.scl(1 / len());
 	}
 
 	@Override
--- a/gdx/src/com/badlogic/gdx/math/Vector3.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector3.java
@@ -299,8 +299,9 @@ public class Vector3 implements Serializable, Vector<Vector3> {
 	@Override
 	public Vector3 nor () {
 		final float len2 = this.len2();
-		if (len2 == 0f || len2 == 1f) return this;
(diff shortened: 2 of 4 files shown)


Darkyenus:
The proposed code is slower (it computes length^2 twice and has more comparisons) and less accurate (almost normalized vectors will never get fully normalized, very small vectors won't get normalized at all, although the precision of such normalization is not entirely clear without testing).

Is there a concrete problem that this change is trying to solve?

fabiitch:
Hi !
 There is no real problem solved by this commit.
I was exploring source and i see difference in nor() between Vector2/Vector3.

The goal was to egalise nor() method in Vector2 and Vector3.
Previously Vector3 check len2 ==0 || len2 ==1 and Vector2 only if len2 ==0

This check with len2 permit to avoid jni call in Math.sqrt if vector is already normalised.

In test case,  Vector3(-5f, 42.00055f, 44444.32f).nor() dont has len2 equal to 1.
So recall nor() on him, renormalise a normalised vector. That was the goald with mathUtils.equal(len2,1)

Use tolerance in check 0 or 1 can avoid to normalise a vector already normalised.

> less accurate 

Its true, i did not think about vectors close to a normalised value.
I think the tolerance in MathUtils  `FLOAT_ROUNDING_ERROR = 0.000001f;` is enought.

Previously in Vector3 the check in nor compare len2 with 0 and 1 before normalise

> it computes length^2 twice  

Yeah i think i should reuse len2 value. i change it.

> more comparisons

- For Vector3 no, 2 comparaison with 0 ,1  just a math.abs called by MathUtils.equals()
- For Vector2 yes the call to len2 was added to check.

I often heards we have to use jni less for performance.

I think Vector2 and Vector3 nor() should have same logic (maybe no ?)

I gonna modify the last operation to use `return this.scl(1f / (float)Math.sqrt(len2));` like before in Vector3


>Is there a concrete problem that this change is trying to solve?
Yeah if there is no problem ... 
I think it was an optimisation, but in regards to your comment , maybe its not.

If im wrong, just close pr (fails appends !)

Thanks a lot for your feedback.

Darkyenus:
`Math.sqrt` is not a jni call, it is implemented in JVM and it is often [intrinsified](https://alidg.me/blog/2020/12/10/hotspot-intrinsics) so the performance is not far from optimal. And while it is still slower than the extra comparison, I expect that most `nor` calls will be on non-unit vectors, so the check will fail and any performance gained by the sqrt elimination will be wasted on the extra check.

The inconsistency is a bit strange, but I'd rather solve it by removing the `== 1`, since very few vectors will be exactly unit in size.

MobiDevelop:
Most of the functions in Java's Math class actually are native methods, including sqrt.

tommyettinger:
It's not exactly accurate to call `Math.sqrt()` a native function in the same sense as a JNI call to a user-written function. As Darkyenus said, Math.sqrt() is an intrinsic candidate on HotSpot (it's annotated as such on Java 9 and higher, and it's still replaced by an intrinsic on Java 8 even though it isn't documented as such). Sources: https://gist.github.com/apangin/7a9b7062a4bd0cd41fcc#file-hotspot-jvm-intrinsics-L9 https://github.com/openjdk/jdk/blob/e5304b3a994b1e291e4ac5258f473dd7874f163f/src/java.base/share/classes/java/lang/Math.java#L394-L400
Intrinsics compile down to specialized native code that doesn't involve JNI, when specific conditions are met (such as a call count threshold). When given a float and the result of sqrt() is cast to a float, as far as I am aware, this intrinsic will use the appropriate hardware instruction for single-precision square root, and in other cases it uses double-precision. You could implement Math.sqrt() with JNI, it would just not perform like it does now (as mentioned in the second link, in the OpenJDK sources).

Getting off-track here, but I agree that the existing behavior should probably be left alone.

MobiDevelop:
Sure, but Android isn't hotspot.

fabiitch:
I agree 
Thanks for your feedback

tommyettinger:
For ART on recent Android: https://cs.android.com/android/platform/superproject/+/master:art/runtime/intrinsics_list.h;l=153 https://cs.android.com/android/platform/superproject/+/master:art/compiler/optimizing/intrinsics_arm_vixl.cc;l=391-398

For Dalvik on older Android: https://cs.android.com/android/platform/superproject/+/master:libcore/dalvik/src/main/java/dalvik/annotation/optimization/CriticalNative.java;l=36-42?q=CriticalNative&ss=android%2Fplatform%2Fsuperproject (`Math.sqrt()` is annotated with `CriticalNative`.)

If you're running on Dalvik, Android is executing something like a (significantly) faster path for JNI code. On ART, it seems to be doing what HotSpot does.

PR #6504, opened 2021-04-20
Add zipfile handle (implements #6212)

dushechka:
Adds a common zip-file handle and its resolver which enable reading from zip archives.
Borrowed from: https://gist.github.com/MobiDevelop/5514357#file-archivefilehandle-java.
This code has been tested on Windows 7x64.
#6212


tests/gdx-tests/src/com/badlogic/gdx/tests/files/ZipFileHandleResolverTest.java
```diff
@@ -0,0 +1,88 @@
...
+		camera.update();
+
+		cameraController = new OrthoCamController(camera);
+		Gdx.input.setInputProcessor(cameraController);
+
+		font = new BitmapFont();
+		batch = new SpriteBatch();
+
+		try {
+			ZipFile archive = new ZipFile(Gdx.files.internal(ZIP_FILE_PATH).file());
```
joshuaskelly, line 64: `FileHandle.file()` is unsupported on the GWT backend:
https://github.com/libgdx/libgdx/blob/5917a01534d7a2aa134e61b7c7975746e5ef19e4/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtFileHandle.java#L102

MrStahlfelge (changes requested):
As you see, the GWT build fails. You can check locally by running `gradlew :tests:gdx-tests-gwt:compileGwt`

Probably when fixing the `file()` problem you will stumble across another problem. Probably you need to add an emu file for ZipFileHandle, just throwing exceptions.

dushechka:
@MrStahlfelge Maybe I'd better add a check for zip files to the [LwjglFiles](https://github.com/libgdx/libgdx/blob/5917a01534d7a2aa134e61b7c7975746e5ef19e4/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglFiles.java) (like in [DefaultAndroidFiles#getFileHandle](https://github.com/libgdx/libgdx/blob/ae57eee82b7c447abf1111c09cc6c2b439f2b9c4/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidFiles.java#L66)), so we could get access to zip files on desktop simply through [Gdx.files](https://github.com/libgdx/libgdx/blob/ae57eee82b7c447abf1111c09cc6c2b439f2b9c4/gdx/src/com/badlogic/gdx/Files.java#L25) and not break polymorphism? Then I'd rather move the test to the LWJGL and not make it common for all backends. Maybe later I could do the same for the Headless backend. GWT, in turn, will require its own implementation.

dushechka:
Submitted a separate pr 'cause this one has got too flooded because of the rebase: [#6518](https://github.com/libgdx/libgdx/pull/6518).

MrStahlfelge:
Then we can close this one?

PR #6507, opened 2021-04-21
Use gdx-jnigen v2.2.0, gdx-jnigen-loader, macOS arm64 build targets

PokeMMO:
* gdx-jnigen v2.2.0 splits macOS arm64 into it's own dylib instead of merging them into x86_64 https://github.com/libgdx/gdx-jnigen/pull/24
* Use gdx-jnigen-loader to allow for gdx-jnigen projects that depend on it to be used in libgdx projects easier without duplicate class files.


Tom-Ski (approved):

NathanSweet:
Previously gdx didn't have any runtime gdx-jnigen dependencies, now it requires gdx-jnigen-loader solely for SharedLibraryLoader. It's not a big deal but is it worthwhile for one class?

PokeMMO:
gdx-jnigen-loader is just one class with no dependencies.

This allows non-gdx projects to use gdx-jnigen while also being fully compatible with gdx due to the same shared dependency.

Tom-Ski:
Its more to completely extract the jnigen ecosystem entirely from the project so we don't have duplication and can separate the release cycle from jnigen which shouldn't be tied to libgdx.

NathanSweet:
I understand. I'm OK with it, though I'm not sure I'd add a dependency for one class.

PokeMMO:
I think the dependency is a lot better than requiring all projects that use gdx-jnigen to depend on gdx (with it's own release schedule) or to maintain SharedLibraryLoader in two locations that can't both be depended on at the same time without its own set of issues.

NathanSweet:
Depending on gdx to get SharedLibraryLoader is unreasonable. gdx-jnigen having a copy of SharedLibraryLoader isn't much of a problem. Not sure what you mean that they can't both be depended on? They'd be in a different packages.

The dependency is OK, it was just nice to not have any dependencies.

PR #6508, opened 2021-04-25
fix audio voice stealing

mgsx-dev:
When a sound is finished, a new music playback didn't remove entry in sound/source maps. In this case, any change using the soundID (volume, pan, pitch) applies to the music instead.

I included a test case in order to validate the issue, let me know if you don't want this test in libgdx repo or if you prefer that i move it somewhere else...

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OpenALLwjglAudio.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OpenALLwjglAudio.java
@@ -141,12 +141,11 @@ public class OpenALLwjglAudio implements LwjglAudio {
 			int sourceId = idleSources.get(i);
 			int state = alGetSourcei(sourceId, AL_SOURCE_STATE);
 			if (state != AL_PLAYING && state != AL_PAUSED) {
+				Long oldSoundId = sourceToSoundId.remove(sourceId);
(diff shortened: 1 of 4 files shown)


tommyettinger:
It took me a while to understand the problem here; the IDs in the current libGDX code can change while multiple sounds or music are playing, which makes changes to other audio streams unpredictable without this PR? Is that right? I like having the test here; I think the audio part of libGDX doesn't see much attention, so a test to verify behavior is good. The original code does look buggy, but I don't really know what I'm looking at.

My vote is to merge after mgsx has had a chance to look at this again and maybe think of other parts of code that might be involved. I'm slightly concerned about doing anything hastily with the audio, because it doesn't get tested by a lot of games.

mgsx-dev:
I double checked, that was the only place where maps was not properly updated.

Quick explanation (i'll try my best): 
* idleSources, sourceToSoundId and soundIdToSource form a kind of pool.
* sourceToSoundId and soundIdToSource are only used by Sound.
* idleSources are used by Sound, Music, and AudioDevice.
* Source can be explicitly freed (stop sound, dispose sound, stop music, stop device, etc): properly freed in all these cases.
* Source can be implicitly freed when obtaining a source (that was the buggy case). It scans the pool looking for a stopped source that can be used/reused. But it didn't update the maps when obtaining a source for music.

tommyettinger (approved):
You have my rubber stamp of approval! 👍

crykn (approved):
Looks good to me!

PR #6509, opened 2021-04-26
Moving music streaming to a thread for desktop

mgsx-dev:
This PR is an attempt to move music streaming to a thread for desktop backends. 
It still a work in progress, it almost work as expected but there are some corner cases to fix, maybe i'll need to change the synchronization strategy.

I only changed Lwjgl3 backend for now, i'll copy/paste to lwjgl2 backend once it's all clear.

I need to find time to finish it but i thought it was preferable to PR this WIP for discussion as a replacement of #6452.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusicProcessor.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusicProcessor.java
@@ -1,0 +1,123 @@
+package com.badlogic.gdx.backends.lwjgl3.audio;
+
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.Disposable;
+
+public class OpenALMusicProcessor implements Runnable, Disposable
+{
+	private Queue<OpenALMusic> musicsToPlay = new ConcurrentLinkedQueue<OpenALMusic>();
+	private Queue<OpenALMusic> musicsToStop = new ConcurrentLinkedQueue<OpenALMusic>();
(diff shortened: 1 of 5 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusicProcessor.java
```diff
@@ -0,0 +1,123 @@
...
+	private Queue<OpenALMusic> musicsToPlay = new ConcurrentLinkedQueue<OpenALMusic>();
+	private Queue<OpenALMusic> musicsToStop = new ConcurrentLinkedQueue<OpenALMusic>();
+	Queue<OpenALMusic> musicsStopped = new ConcurrentLinkedQueue<OpenALMusic>();
+	
+	Queue<OpenALMusic> musicsFinished = new ConcurrentLinkedQueue<OpenALMusic>();
+	private Array<OpenALMusic> musicsToUpdate = new Array<OpenALMusic>();
+	
+	private Queue<OpenALMusic> musicsToSeek = new ConcurrentLinkedQueue<OpenALMusic>();
+	
+	private boolean shouldRun;
```
Hangman, line 23: I think shouldRun should be volatile. Otherwise, at least in theory (*), it'd be undefined if and when the thread will stop running.

* I don't think it will be a problem in pratice but since volatile shouldn't introduce much of a performance penalty, I don't see a reason to take the risk that on some systems in some weird situations the thread keeps being alive for longer or forever.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusicProcessor.java
```diff
@@ -0,0 +1,123 @@
...
+	private Thread audioThread;
+	private final OpenALLwjgl3Audio audio;
+	
+	public OpenALMusicProcessor (OpenALLwjgl3Audio audio) {
+		this.audio = audio;
+		audioThread = new Thread(this);
```
Hangman, line 29: `audioThread.setDaemon(true);`
The audioThread should be made daemon.

sebaber:
I need so hard this PR merged, having the same issues that origin this. What is the status of this? Need a hand?

PR #6517, opened 2021-05-03
Improve precision for sinDeg(180f)

tommyettinger:
This PR just makes `MathUtils.sinDeg(180)` (and `MathUtils.cosDeg(90)`) correct, so they return `0` as expected. This particular angle was one of four angles given special treatment by `MathUtils`, but was still wrong, yielding `-8.742278E-8` instead of `0`. I think the updated code is actually less complex, because it doesn't risk a difference between the double-precision radians that `Math.sin()` expects and the single-precision radians `MathUtils` can provide it. I also added tests for `sinDeg()` and `cosDeg()`, which would have failed before because of the extremely low tolerance they use.

I think the right angles are the most likely to need to be correct, so special-casing them makes sense. Everywhere else, the `MathUtils.isEqual()` checks with a tolerance are perfectly fine.

Thanks to Discord user Miamoto for finding the original bug, and also working around it by using `MathUtils.isZero()`.


yuripourre (approved):
Sounds good, the code only overwrites the index for those 4 angles (0, 90, 180, 270).

The PR also has tests

tommyettinger:
Yeah, the angle 0 would not have a value of 0 in the table if something didn't overwrite it, though 90 and 270 were fine before (I believe they were exactly 1 and -1). Partly running a loop for 4 items is just overkill, though, so directly writing to the 4 angles we want to change made the most sense.

...That said...

I did make a version that used the previous loop, and it's pretty horrific.

```java
            for (int i = 0, v = 2; i < 360; i += 90, v--)
                table[(int)(i * degToIndex) & SIN_MASK] = (v & (i | -i) >> 31);
```

I figured it's better for readability to have it how it is in the PR.

gdx/test/com/badlogic/gdx/math/MathUtilsTest.java
```diff
@@ -59,4 +59,20 @@ public void testRandomLong() {
...
+		assertEquals(-1f, MathUtils.sinDeg(270f), Float.MIN_NORMAL);
+	}
+
+	@Test
+	public void testCosDeg() {
+		assertEquals(1f, MathUtils.cosDeg(0f), Float.MIN_NORMAL);
+		assertEquals(0f, MathUtils.cosDeg(90f), Float.MIN_NORMAL);
+		assertEquals(-1f, MathUtils.cosDeg(180f), Float.MIN_NORMAL);
+		assertEquals(0f, MathUtils.cosDeg(270f), Float.MIN_NORMAL);
+	}
```
obigu, line 77: Don't we expect `delta` of 0 instead of `Float.MIN_NORMAL` now?

gdx/src/com/badlogic/gdx/math/MathUtils.java
```diff
@@ -59,8 +59,15 @@ private MathUtils () {
 		static {
 			for (int i = 0; i < SIN_COUNT; i++)
 				table[i] = (float)Math.sin((i + 0.5f) / SIN_COUNT * radFull);
-			for (int i = 0; i < 360; i += 90)
-				table[(int)(i * degToIndex) & SIN_MASK] = (float)Math.sin(i * degreesToRadians);
+			// Calculating the radians from degrees is just a little imprecise; this
+			// was slightly off for 180 degrees (giving -8.742278E-8 instead of 0).
+			// It's easier to understand to assign the exact values like this.
+			// Only these right angles are given the most precise measurement, so
+			// other angles should be compared with a tolerance.
```
obigu, line 66: This comment helps understand the change but doesn't document the code that well and IMO contains unnecessary information. I would just keep why the code exists (to handle common especial cases with right degree angles caused by float precision errors).

obigu (commented):
Good stuff, thanks. This may open a debate on whether classes such as `Vector2` should use `MathUtils` to calculate angles. We would get the benefit of supposed improved performance and extra precision when dealing with right angles with degrees (rendering `rotate90` unnecessary) for the price of loosing just a bit of precision on other angles.

gdx/test/com/badlogic/gdx/math/MathUtilsTest.java
tommyettinger: 0.0f would be fine here, though `assertEquals(double, double)` (which compares exactly, with effectively 0 tolerance) is currently deprecated in JUnit, and using a tolerance of 0 just sidesteps that deprecation without addressing it. I'm not opposed to either currently, in this case.

gdx/src/com/badlogic/gdx/math/MathUtils.java
tommyettinger: I'll shorten it up.

gdx/test/com/badlogic/gdx/math/MathUtilsTest.java
obigu: In this case we expect to be exactly that value, with 0 tolerance. I think the deprecation is to encourage taking into account what's the tolerated delta on each individual assertion, not to address delta being 0 in some cases.

tommyettinger: OK, changed to 0.

obigu (approved):

yuripourre (approved):

PR #6518, opened 2021-05-04
Add zipfile handle (#6212)

dushechka:
Added instead of pr [#6504](https://github.com/libgdx/libgdx/pull/6504), because it was too dirty. Please refer to its description.
This code has been tested on Windows 7x64 and Ubuntu 20.04 x86_64.
Implements #6212.


badlogic:
This looks OK to me, but needs an entry in CHANGES (and a resolution of conflicts).

MobiDevelop:
There was a [comment on my gist](https://gist.github.com/MobiDevelop/5514357?permalink_comment_id=3142905#gistcomment-3142905) about it not working on Android. I never looked into it, but it's a bit odd I suppose to pack zip files into what's effectively a zip file. Was it tested on Android and found to be working fine?

PokeMMO:
I use similar code from your gist on android and can confirm it works.

MobiDevelop:
Cool beans, so if this is still wanted, I guess it's just a matter of a resolving of the conflicts.

PR #6520, opened 2021-05-05
Polygon add method setVertex

fabiitch:
Hi, 
Actually in Polygon , the only method to change a vertex is to use 
 ```	
 public void setVertices (float[] vertices) {
		if (vertices.length < 6) throw new IllegalArgumentException("polygons must contain at least 3 points.");
		localVertices = vertices;
		dirty = true;
	} 
 ```	
I think a method is missing for change easily a vertex.

If read pr code you can see,
- a param vertexNum -> i start to 0, so for a triangle values allowed are (0,1,2). 
Maybe its more normal to start at 1.

- I change vertex in localvertices array, but should it consider the value as worldvertices ? And transform with scale and rotation to a localvertices
- It throw IllegalArgumentException if vertices dont exist in polygon.
Its better to let an npe appears?

My use case is not  on polygon class but in my own class `public class Triangle extends Polygon `
And i want to add a method for change one vertex.
An other solution was to change localvertices to protected, but i follow @obigu advices.

Waiting your feedback
Thanks for reading

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Polygon.java
+++ b/gdx/src/com/badlogic/gdx/math/Polygon.java
@@ -121,12 +121,13 @@ public class Polygon implements Shape2D {
 
 	 /** Set vertex position
 	  *
-	  * @param vertexNum min=0, max=(vertices/2)-1
+	  * @param index min=0, max=vertices.length/2-1
 	  * @throws IllegalArgumentException if vertex doesnt exist */
-	 public void setVertex (int vertexNum, float x, float y) {
-		  if (vertexNum > localVertices.length / 2 -1) throw new IllegalArgumentException("the vertex " + vertexNum + "doesn't exist");
-		  localVertices[2 * vertexNum] = x;
-		  localVertices[2 * vertexNum + 1] = y;
+	 public void setVertex (int index, float x, float y) {
+		  if (index < 0 || index > localVertices.length / 2 - 1)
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/math/Polygon.java
```diff
@@ -119,6 +119,17 @@ public void setVertices (float[] vertices) {
 		dirty = true;
 	}
 
+	 /** Set vertex position
+	  *
+	  * @param vertexNum min=0, max=(vertices/2)-1
+	  * @throws IllegalArgumentException if vertex doesnt exist */
+	 public void setVertex (int vertexNum, float x, float y) {
+		  if (vertexNum > localVertices.length / 2 -1) throw new IllegalArgumentException("the vertex " + vertexNum + "doesn't exist");
```
crykn, line 127: I think the exception should be thrown if `vertexNum < 0` as well. And maybe an `IndexOutOfBoundsException` would fit better?

obigu: I'd just add the `vertexNum < 0` check, the rest seems good to me.

fabiitch: I add <0 check, i let you choose about exception to throw.

I got a question about to throw exception, 
Its really better to throw exception manually  than let an outofbound appears?
Or is just for stackstrace lisibility ?

obigu (approved):

PR #6521, opened 2021-05-05
iOS Call glViewport on device rotation, fixes #6514

### LINKED ISSUE #6514: LibGDX app renders incorrectly on iOS after device rotation

#### Issue details
LibGDX app renders incorrectly on iOS after device rotation

#### Reproduction steps
Create a boilerplate LibGDX app, ensure that the Info.plist.xml file permits all device orientations.

Capture screen size in `resize()`. Do some drawing in `render()` that is dependent on the aspect ratio of the screen. Check that the result is the same if the app is started in portrait vs starting in landscape then rotating to portrait.

#### Code
````
public class MyGdxGame extends ApplicationAdapter {
    int width = 0, height = 0;
    ShapeRenderer sr;

    @Override
    public void create() {
        super.create();
        sr = new ShapeRenderer();
    }

    @Override
        super.resize(width, height);
        this.width = width;
        this.height = height;
    }

    @Override
    public void render () {
        super.render();
        Gdx.gl.glClearColor(0.6f, 0.1f, 0.1f, 1f);
        sr.begin(ShapeRenderer.ShapeType.Filled);
        sr.setColor(Color.YELLOW);
        sr.rect(width/4f, height/4f, width/2f, height/2f);
        sr.end();
    }

    @Override
    public void dispose () {
        super.dispose();
        sr.dispose();
    }
}
````
Tested with 1.9.5 and 1.10.0 - same results

#### Expected result

There should be no difference in the display whether the app starts in one orientation then rotates, vs starting in the final orientation.

#### Actual result

As expected on Android. Not as expected on iOS. Although `resize()` is called correctly with the new screen size, the image drawn is wrong, as though the backend is unaware of the rotation.

- [ ] Android
- [X] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

#### Screenshots

gdx

MrStahlfelge:
Fix for #6514

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSUIViewController.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSUIViewController.java
@@ -80,6 +80,7 @@ public class IOSUIViewController extends GLKViewController {
 		if (graphics.created && (newBounds.width != oldBounds.width || newBounds.height != oldBounds.height)) {
 			graphics.makeCurrent();
 			graphics.updateSafeInsets();
+			graphics.gl20.glViewport(0, 0, newBounds.backBufferWidth, newBounds.backBufferHeight);
 			if (graphics.config.hdpiMode == HdpiMode.Pixels) {
 				app.listener.resize(newBounds.backBufferWidth, newBounds.backBufferHeight);
 			} else {


obigu:
Haven't tested it yet but I think we may be able to avoid the `glViewport()` call here by instead updating cached `IOSGLES20.width` and `IOSGLES20.height` and letting `draw()` to make the `glViewport()` call using the new cached values.

obigu (approved):
2 `glViewport()` calls will be made when resizing (when using a `Viewport`) but I think it's not worth trying to optimize it as it should be a cheap call (maybe the driver doesn't even run the second one), we may introduce undesired effects and this is the way we do it on Android.

PR #6522, opened 2021-05-06
3D : take vertex attribute unit into account when creating shaders

mgsx-dev:
Shader providers create shaders based on renderable configuration (material, lights, vertex attributes, etc). For vertex attributes, test is based on attributes mask and attributes count which wasn't enough in some cases. For instance : 
* a first shader is created for a model having 3 bone weights and 2 texture coordinates.
* when selecting a shader for a model with 4 bone weights and 1 texture coordinate, the previous shader was selected but wasn't able to render it : same attribute count but not same weight count.
* if creation order was inverted, it worked because a shader with 4 bone weights was able to render a model with less bone weights.

Another side effect was that one shader was created per bone weight count configuration : 1 shader for models with 3 weights, another for models with 4 weights, and so on. Resulting in a lot of shader switches. With this change there is only 2 variations : one with some bone weights (for skinned models) and one without (for non skinned models) just like DepthShader.

So, i introduced a new shader provider configuration setting for bone weights, limited to 4 by default. Most of tools i know export models with max 4 bone weights and i think it's a reasonable default value, in my experience you rarely need more.

This change may introduce performance issues with existing code, for instance a game where all models with 2 bone weights was producing one shader with 2 bone weights. It'll now produce one shader with 4 bone weights. They still can always change configuration (numWeights = 2) for their game.
Another corner case would be a game with a lot of models with 3 weights and one model with 8. With this change they are forced to use 8 weights for all their rendered models. A solution for them could be to use an additional shader providers and set the special shader for these special renderables... there's always a way.

I also simplified DepthShader code a bit: there are no use case i'm thinking of for weights and numBones fields outside of this class.

I can understand there are choices to debate here, i'm all ears.

### DIFF
--- a/.github/workflows/build-pullrequest.yml
+++ b/.github/workflows/build-pullrequest.yml
@@ -9,7 +9,7 @@ permissions:
 jobs:
   build-java:
     if: "!contains(github.event.head_commit.message, 'ci skip')"
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@
 - [BREAKING CHANGE] Cell getters return object wrapper instead of primitives.
 - [BREAKING CHANGE] MeshPartBuilder#lastIndex now returns int instead of short.
 - [BREAKING CHANGE] 3D API - max bone weights is now configurable and limited to 4 by default. Change this value if you need less or more. See #6522.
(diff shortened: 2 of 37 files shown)


Tom-Ski (approved):

mgsx-dev:
I had to rebase this PR because it was done prior to the overall repo formatting (all files conflicting). I think it's OK but it worth a second review i think.

Tom-Ski (dismissed):

NathanSweet:
Forgot to sqorsh, I think. :)

PR #6524, opened 2021-05-07
Remove frameCount from AnimatedTiledMapTile

TCreutzenberg:
frameCount in AnimatedTiledMapTile is private and not used anywhere. It also cannot be accessed from outside.


crykn (approved):
LGTM. The frame count can still be accessed via `getFrameTiles()#length`.

PR #6527, opened 2021-05-08
fix stereo for sounds and audio device. fixes #6526

### LINKED ISSUE #6526: PCM Audio not properly stereo with LWJGL3 backend
#### Issue details
PCM audio played via Gdx.audio.newAudioDevice does not properly play stereo content with the LWJGL3 backend, but does with the default desktop backend. It's not fully mono but it isn't stereo to the correct extent either.

#### Reproduction steps/code
```
public class Test extends ApplicationAdapter {
	AudioDevice audioDevice;
	float deviceSamples[] = new float[1024];
	float x = 0;
	@Override
	public void create () {
		audioDevice = Gdx.audio.newAudioDevice(44100, false);
	}

	@Override
	public void render () {
		new Thread() {
			@Override
			public void run() {
				while (true) {
					try {
						for (int a = 0; a < 1024; a += 2) {
							float sample = .2f * (float) Math.sin(x);
							x += 0.0005;
							deviceSamples[a] = 0; // Left speaker should be silent.
							deviceSamples[a + 1] = sample; // Right speaker should make noise.
						}

						audioDevice.writeSamples(deviceSamples, 0, deviceSamples.length);
						int latency = audioDevice.getLatency();
						sleep(500 * latency / 44100);
					} catch (Exception e) {
						e.printStackTrace();
					}
				}
			}
		}.start();
	}
}
```

Expectation: Sound only comes out of the right speaker.
Reality: Sound comes mostly out of the right speaker, but also somewhat from the left. It works fine when using the default lwjgl2 backend, but not with the lwjgl3 backend.

libgdx 1.10.0

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ X] Windows
- [ ] Linux
- [ ] MacOS

mgsx-dev:
AL_DIRECT_CHANNELS_SOFT source option was only set for Music. With this change it's applied for both Sound and AudioDevice as well. This option allow to not treat stereo sources with 3D audio stuff. When disabled, you can ear a bit of left channel in right speaker and vice-versa. When enabled, channels are perfectly separated.

This option only applies to stereo sources and has no effect on mono sources. Note that with mono sources you can ear a bit of sound in left speaker even with pan at full right (and vice-versa) because of how OpenAL implement 3D audio.

Lwjgl2 backend doesn't have this option but both have the same behavior now.

I improved the tests a bit in order to validate this change.


NathanSweet:
Cheers!

crykn:
While this change works as intended (and fixes the bug in #6526), I thought I'd mention a shortcoming of the `AL_DIRECT_CHANNELS_SOFT` option I found while looking over this PR:

Apparently it drops any extra channels if it can't find a matching output channel. For example, when playing a surround source on stereo speakers the additional channels are dropped instead of downmixed (see the AL SOFT specification [here](https://github.com/openalext/openalext/wiki/AL_SOFT_direct_channels#additions-to-specification) and this [issue](https://github.com/telegramdesktop/tdesktop/issues/2580#issuecomment-438137819) for example). Obviously, this can lead to unwanted results. 

[As a consequence](https://github.com/kcat/openal-soft/issues/209#issuecomment-401983491), OpenAL Soft introduced an alternative extension: [`AL_SOFT_direct_channels_remix`](https://openal-soft.org/openal-extensions/SOFT_direct_channels_remix.txt). However, this extension isn't a part of LWJGL 3 until the next release.

PR #6528, opened 2021-05-08
Box2D Shape implements Disposable

fabiitch:
Hey i see Shape class got this  comment
``` 
* NOTE: YOU NEED TO DISPOSE SHAPES YOU CREATED YOURSELF AFTER YOU NO LONGER USE THEM! E.g. after calling body.createFixture();
 * @author mzechner */
public abstract class Shape implements Disposable {
```
dispose() method is already here.
I check on package except in World class (already impl Disposable) i dont find other occurence of dispose() or jniDispose()

Thanks for reading.

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d-gwt/src/com/badlogic/gdx/physics/box2d/gwt/emu/com/badlogic/gdx/physics/box2d/Shape.java
+++ b/extensions/gdx-box2d/gdx-box2d-gwt/src/com/badlogic/gdx/physics/box2d/gwt/emu/com/badlogic/gdx/physics/box2d/Shape.java
@@ -16,12 +16,14 @@
 
 package com.badlogic.gdx.physics.box2d;
 
+import com.badlogic.gdx.utils.Disposable;
+
 /** A shape is used for collision detection. You can create a shape however you like. Shapes used for simulation in b2World are
  * created automatically when a b2Fixture is created. Shapes may encapsulate a one or more child shapes.
  * 
  * NOTE: YOU NEED TO DISPOSE SHAPES YOU CREATED YOURSELF AFTER YOU NO LONGER USE THEM! E.g. after calling body.createFixture();
  * @author mzechner */
-public abstract class Shape {
+public abstract class Shape implements Disposable {
 	/** Enum describing the type of a shape
 	 * @author mzechner */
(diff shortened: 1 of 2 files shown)


obigu:
Not sure it's worth adding the `Disposable` interface here, the only benefit I see is code documentation but the current docs are pretty clear. Shapes on Box2D are usually very short lived and the `dispose()` method is usually called just after fixture creation. This is different from the most common behavior of classes that implement `Disposable` in libGDX in which `dispose()` is called on longer lived resources when they are not needed anymore for example when changing to a different screen or quitting the game. 

In case we want to add it I'd suggest, at least, adding the @Override annotation to `dispose()`.

fabiitch:
Hi @obigu thanks to look here !

i added @override on dispose.

I don't agree about  life times.
In wiki [Memory Management'](https://github.com/libgdx/libgdx/wiki/Memory-management.

> They all implement a common Disposable interface which indicates that instances of this class need to be disposed of manually at the end of the life-time.

Wiki provide link to [Disposable ](https://libgdx.badlogicgames.com/ci/nightlies/docs/api/com/badlogic/gdx/utils/Disposable.html) javadoc
Add Disposable reference Box2D Shape here too.

I don't think this implements can be a bad thing.

I have already add Shape in the list in memory management wiki, i remove it if commit is rejected

Thanks for your feedbacks.

obigu:
> I have already add Shape in the list in memory management wiki, i remove it if commit is rejected

`Shape` was already in the wiki, there's no other ' Shape' class. I've reverted the change. In any case edits to the wiki should be made After PRs are merged.

fabiitch:
oh my bad , sry

PokeMMO:
> Not sure it's worth adding the `Disposable` interface here, the only benefit I see is code documentation but the current docs are pretty clear. Shapes on Box2D are usually very short lived and the `dispose()` method is usually called just after fixture creation. This is different from the most common behavior of classes that implement `Disposable` in libGDX in which `dispose()` is called on longer lived resources when they are not needed anymore for example when changing to a different screen or quitting the game.

It still has some usefulness if you were to consider debug tools which track objects which implement Disposable and throw warnings when an object is finalized without being disposed first.

badlogic:
This doesn't hurt and adds consitency.

PR #6534, opened 2021-05-12
Fix #6464: TMX built-in tile property "type" is now supported.

### LINKED ISSUE #6464: TiledMap tile property "Type" doesn't get loaded

lyze237:
Previous PR request: https://github.com/libgdx/libgdx/pull/6465

I've found some time and implemented it in the way @mgsx-dev wanted it.

This however turns the PR into a breaking change compared to previously where it wasn't the case.

image

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,8 +2,8 @@
 - API Addition: ObjLoader now supports ambientColor, ambientTexture, transparencyTexture, specularTexture and shininessTexture
 - API Addition: PointSpriteParticleBatch blending is now configurable.
 - TOOLS Features: Blending mode and sort mode can be changed in Flame particle 3D editor.
-- API Addition : Polygon have method setVertex (int index, float x, float y)
-- API Addition:
+- API Addition: Polygon have method setVertex (int index, float x, float y).
+- API Addition: TMX built-in tile property "type" is now supported.
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - API Addition: PointSpriteParticleBatch blending is now configurable.
 - TOOLS Features: Blending mode and sort mode can be changed in Flame particle 3D editor.
 - API Addition : Polygon have method setVertex (int index, float x, float y)
+- API Addition:
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
@@ -641,6 +641,10 @@ public abstract class BaseTmxMapLoader<P extends BaseTmxMapLoader.Parameters> ex
 		if (probability != null) {
 			tile.getProperties().put("probability", probability);
 		}
+		String type = tileElement.getAttribute("type", null);
+		if (type != null) {
+			tile.getProperties().put("type", type);
+		}
 		Element properties = tileElement.getChildByName("properties");
 		if (properties != null) {
 			loadProperties(tile.getProperties(), properties);


gdx/src/com/badlogic/gdx/maps/tiled/tiles/StaticTiledMapTile.java
```diff
@@ -123,4 +127,32 @@ public StaticTiledMapTile (StaticTiledMapTile copy) {
 		this.id = copy.id;
```
mgsx-dev, line 127: new fields need to be copied here.

gdx/src/com/badlogic/gdx/maps/tiled/TiledMapTile.java
```diff
@@ -63,4 +63,21 @@
 	/** @return collection of objects contained in the tile */
 	public MapObjects getObjects();
 
+	/** Sets he content of the property "terrain" */
+	public void setTerrain(String terrain);
+	/** @return the content of the property "terrain" */
```
mgsx-dev, line 68: formatting issue ?

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -634,13 +634,14 @@ protected abstract void addStaticTiles (FileHandle tmxFile, ImageResolver imageR
...
-		if (terrain != null) {
-			tile.getProperties().put("terrain", terrain);
-		}
-		String probability = tileElement.getAttribute("probability", null);
-		if (probability != null) {
-			tile.getProperties().put("probability", probability);
-		}
+		tile.setTerrain(terrain);
+
+		float probability = tileElement.getFloatAttribute("probability", 1f);
```
mgsx-dev, line 639: apparently default is 0 in tiled spec, see https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tile
however it might be better to keep it as Float (default null) in order to keep the previous behavior.

CHANGES
```diff
@@ -1,4 +1,5 @@
 [1.10.1]
+- [BREAKING CHANGE]: Removed TiledMapTile default properties (terrain, probability, type) from the properties array and added them as getter/setters to the tile classes.
```
mgsx-dev, line 2: maybe add a line (API addition) to say that builtin 'type' field is now supported?

mgsx-dev (changes requested):

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
lyze237: I've set it to 1 as it defaults to 1.000 in the tiled editor. I however could change it to 0 though.

image

Previous behavior was a string, should I change it back to that instead?

mgsx-dev: Indeed, it's maybe more safe to keep it as String. Maybe some user code use it as string and encode some stuff in it, who knows..

mgsx-dev (approved):
LGTM

MobiDevelop:
For background, the TiledMapTile was never intended to specifically map to the Tiled (TMX) idea of a tile, so I don't really agree with these changes.

tommyettinger:
Is this still compatible with Tide? Do we still need to be compatible with Tide, or other editors? I think the naming conventions across the tiled package are very misleading, but I'm used to that in libGDX by now ("No no, you want an int array, not an IntArray!"). It isn't at all clear that Tiled, in class names, doesn't refer to Tiled (TMX), but to a rectangular-grid tiling, until you've really delved into the barely-documented tiled package. I think the JavaDocs for TiledMapTile, which currently only has a `@brief` summary, could really use some attention from someone who actually knows what the intent behind TiledMapTile is, rather than the rest of us guessing at it.

As an aside, we've had lots of issues on the Discord and I believe also here on GitHub relating to newer features in the Tiled map format, such as infinite maps. If any PRs are going to be challenged by the lack of documentation in the tiled package, of intent or otherwise, then that documentation should be improved, or the challenges should be considered moot.

MobiDevelop:
Ultimately, it's y'all's project now... I suppose. If you guys want to change things, I'm not going to block it. But just because you believe the name TiledMap specifically refers to the Tiled Editor doesn't make it so.

mgsx-dev:
@MobiDevelop since we already merge "terrain" and "probability" properties into custom properties and then potentially override custom properties with same names, what's the best move? merge "type" property as well or introduce a kind of prefix for them?
I'm not fan of both solutions, i can't see an elegant solution right now, any suggestion?

MobiDevelop:
I'd probably go the route that you add the 'built-in' values first, and if they chose to add a custom property with the same name have it override the built-in.

lyze237:
soo the way how I did it in the first pr initially (ignoring the later change with /t)?

MobiDevelop:
Looks like it, yeah.

mgsx-dev (approved):

mgsx-dev:
side note: i was wrong, custom properties already override builtin ones. So this change doesn't break anything.

CHANGES
```diff
@@ -3,6 +3,7 @@
 - API Addition: PointSpriteParticleBatch blending is now configurable.
 - TOOLS Features: Blending mode and sort mode can be changed in Flame particle 3D editor.
 - API Addition : Polygon have method setVertex (int index, float x, float y)
+- API Addition:
```
MobiDevelop, line 6: I guess this line is either not needed or should be completed.

MobiDevelop (approved):

CHANGES
lyze237: No clue why it disappeared, most likely due to all the force push magic. :D

PR #6536, opened 2021-05-17
Fix asset manager deadlock on dispose and clear

Crunkle:
A deadlock occurs when an asset manager executor thread tries to acquire the lock during a `dispose()` or `clear()` call. This PR alters these methods to only acquire the lock when required (similar to `finishLoading()` having no explicit lock) and yield to the pending async tasks instead.

```
"AssetManager@2033" daemon prio=5 tid=0x17 nid=NA waiting for monitor entry
  java.lang.Thread.State: BLOCKED
	 waiting for main@1 to release lock on <0x81d>
	  at com.badlogic.gdx.assets.AssetManager.get(AssetManager.java:145)
```

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -699,43 +699,50 @@ public class AssetManager implements Disposable {
 
 	/** Disposes all assets in the manager and stops all asynchronous loading. */
 	@Override
-	public synchronized void dispose () {
+	public void dispose () {
 		log.debug("Disposing.");
 		clear();
 		executor.dispose();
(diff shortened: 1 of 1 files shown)


tommyettinger:
Huh, it seems unfair that GitHub considers you a first-time contributor; you put a bunch of work into https://github.com/libgdx/libgdx/pull/6144 and that was just changed manually. I don't know anything about deadlocks, so I hope someone else can chime in, but I will approve the GitHub workflow.

MrStahlfelge:
It looks good to me, let's see if @NathanSweet has objections.

NathanSweet:
Sorry for the delay! It does seem to make sense to me.

Issue #6537, opened 2021-05-19
Gdx.graphics.getDensity() returns infinity inside VirtualBox


ichernev:
#### Issue details
Calling `Gdx.graphics.getDensity()` inside VirtualBox (Linux guest/host) returns `Infinity` (because display size reported by Lwjgl is 0). I understand that on some platforms the display density can not be computed, but I think there should be a special value UNKNOWN or an exception thrown to force client code to handle this case. I got an exception in the JNI code of FreeType (because I use density for font size scaling), and getDensity() doesn't indicate it can return a completely invalid value.

It returns an estimate in any case, so if the value is too low / too high it can just default to 1 or something.

#### Version of LibGDX and/or relevant dependencies
GDX version `1.10.0` (latest at time of writing)

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [x] Linux
- [ ] MacOS

crykn:
Could you please try whether the issue also happens with the [LWJGL 3 backend](https://gist.github.com/crykn/eb37cb4f7a03d006b3a0ecad27292a2d)?

ichernev:
This is with Lwjgl3. It is not trivial to switch to 2 at this point :)

tommyettinger:
Good point that it already returns an estimate; I think you're right in your interpretation of the docs that it does not indicate it can return a non-finite value, and an estimate shouldn't generally be infinite or NaN. A density of 0 would clearly indicate that something is wrong with the estimation, since there's no way you could have a screen with 0 pixels per inch/cm that actually works. Same for a negative number. I'm not sure if it's better to just return 1, indicating "some normal screen," if the screen is actually unusable at that point in time. A quick workaround might be to use Float.isFinite() as your check for an invalid result, since that would pick up both infinities and also NaN (and NaN is tricky to check for normally). I don't know if getDensity() can return NaN, but if it's dividing a positive number by 0.0f already to get Infinity, it's not unreasonable that it could divide 0 by 0.0f and get NaN.

crykn:
I think you should also report this to the [GLFW project](https://github.com/glfw/glfw), since the bug seems to be [on their end](https://github.com/libgdx/libgdx/blob/add6031b113d9e255f791656581bdffbbff83f89/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java#L253).

[...]

PR #6539, opened 2021-05-23
Implemented a guard against invalid values of getPpiX in AbstractGraphics

Timvisser1994:
Gdx.graphics.getDensity() returns infinity inside VirtualBox #6537

This way, AbstractGraphics' getDensity will always return 1f as a default value (instead of Infinity).
By the way, this is my first Pull Request for LibGDX ✨

### DIFF
--- a/gdx/src/com/badlogic/gdx/Graphics.java
+++ b/gdx/src/com/badlogic/gdx/Graphics.java
@@ -216,7 +216,7 @@ public interface Graphics {
 	 * android.util.DisplayMetrics#density, where one DIP is one pixel on an approximately 160 dpi screen. Thus on a 160dpi screen
 	 * this density value will be 1; on a 120 dpi screen it would be .75; etc.
 	 *
-	 * If no pixels per inch could be determined, this returns a default value of 1.
+	 * If the density could not be determined, this returns a default value of 1.
 	 *
 	 * @return the Density Independent Pixel factor of the display. */
 	float getDensity ();
--- a/gdx/src/com/badlogic/gdx/Graphics.java
+++ b/gdx/src/com/badlogic/gdx/Graphics.java
@@ -216,6 +216,8 @@ public interface Graphics {
 	 * android.util.DisplayMetrics#density, where one DIP is one pixel on an approximately 160 dpi screen. Thus on a 160dpi screen
 	 * this density value will be 1; on a 120 dpi screen it would be .75; etc.
 	 *
+	 * If no pixels per inch could be determined, this returns a default value of 1.
+	 *
 	 * @return the Density Independent Pixel factor of the display. */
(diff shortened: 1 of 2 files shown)


tommyettinger:
Normally I'd mention that AndroidGraphics and IOSGraphics override getDensity(), and so could in theory have different behavior than what's documented -- except they also default to a density of 1 already, making this even better. The only place where the documentation might not be 100% accurate is in LwjglGraphics, because it allows an overrideDensity setting in its configuration, but because that is already documented and you'd have to look for overrideDensity to change it, I think we're fine.

Since this is a small change and fixes an annoying bug, I see no reasons that would prevent merging.

gdx/src/com/badlogic/gdx/Graphics.java
```diff
@@ -216,6 +216,8 @@ public String toString () {
 	 * android.util.DisplayMetrics#density, where one DIP is one pixel on an approximately 160 dpi screen. Thus on a 160dpi screen
 	 * this density value will be 1; on a 120 dpi screen it would be .75; etc.
 	 *
+	 * If no pixels per inch could be determined, this returns a default value of 1.
```
tommyettinger, line 219: Only minor critique I have: the term "pixels per inch" gives a subtle clue to how this works in AbstractGraphics, but the other implementing classes don't always use PPI to determine density (especially Android). Technically, this could be `If the density could not be determined, this returns a default value of 1.` but I have no major issue with how this is currently phrased. I don't think Android ever has a case where it can't determine density or PPI (although I could easily be wrong if emulators don't handle their display correctly), so the exact phrasing shouldn't matter.

tommyettinger (approved):
Just minor nitpicks; this looks good and I'll happily merge once someone else looks at it.

obigu:
I think the fix should be on the root of the issue that is `getPpiX()` and `getPpiY()` returning 0 instead of setting an arbitrary density (not allowing differentiating if it's actually the real value or default). On iOS we already have a mechanism to allow the user to define the strategy to follow when `ppix` or `ppiy` are unknown (have a look at `IosApplication#guessUnknownPpi()`). Once set, any other value depending on `ppi` is then calculated accordingly. Maybe we should add a similar mechanism to the affected backends or, if it affects most backends, consider handling it in the abstract class.

MrStahlfelge:
I agree that would be better, but I think as a first step it is a good change if density defaults to 1 if unknown.

gdx/src/com/badlogic/gdx/Graphics.java
Timvisser1994: @tommyettinger Totally agree, changed it.

MrStahlfelge (approved):
No objections were stated, let's merge it

PR #6546, opened 2021-05-30
Add user-configurable uncaughtExceptionHandler to lwjgl backend

nescohen:
Sometimes it is nice to be able to catch exceptions at the top-level. Currently it is very difficult to do this due to the way the lwjgl backend runs it's own thread that is not accessible to the outside world.

This PR simply adds an optional parameter allowing the user to set their own UncaughtExceptionHandler for the lwjgl backend's thread.

### DIFF
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglApplication.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglApplication.java
@@ -58,6 +58,7 @@ public class LwjglApplication implements LwjglApplicationBase {
 	protected ApplicationLogger applicationLogger;
 	protected String preferencesdir;
 	protected Files.FileType preferencesFileType;
+	protected Thread.UncaughtExceptionHandler exceptionHandler;
 
 	public LwjglApplication (ApplicationListener listener, String title, int width, int height) {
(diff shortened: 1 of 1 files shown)


PokeMMO:
Do you have a specific use case that `Thread.#etDefaultUncaughtExceptionHandler` is not sufficient?

nescohen:
@PokeMMO,

To answer your question: Yes, I am developing a game called BattleCasters (https://battlecasters.io). Unfortunately it is closed source at the moment.

Today, I was trying to build a top level exception handling routine that would catch runtime exceptions, log them to file/game server, and ensure the application exits cleanly. I had some trouble with this due to the way that the lwjgl application is started in libgdx. I cannot catch exceptions outside of LwjglApplication class because it runs the application in a separate thread. On the other side, to catch exceptions inside my application listener, I would have to wrap each listener method individually and even then I might miss some. It just seemed like a better solution to do something like this.

For what it's worth I am not necessarily convinced that this is the best way to expose the required functionality, but tried to come up with a reasonable way that did not break any old APIs.

crykn:
This still doesn't really answer, why `Thread#setDefaultUncaughtExceptionHandler` wouldn't suffice in your case.

nescohen:
@PokeMMO @crykn Sorry guys. Reading comprehension skills. That seems like a cleaner solution. Thanks

Issue #6547, opened 2021-05-31
FreeTypeFontGenerator produces very different result compared to Hiero


MontyHimself:
#### Issue details
I am trying to use the `FreeTypeFontGenerator` to generate a `BitmapFont` for the Calibri typeface. This is the code I use for generation:

```java
gridCoordFonts = new BitmapFont[4];
FreeTypeFontGenerator generator = new FreeTypeFontGenerator(Gdx.files.internal("assets/fonts/calibri/calibri.ttf"));
for (int i = 0; i < baseCamera.getZoomLevels().length; i++) {
    FreeTypeFontGenerator.FreeTypeFontParameter parameter = new FreeTypeFontGenerator.FreeTypeFontParameter();
    parameter.size = Math.round(16f / baseCamera.getZoomLevels()[i]);
    parameter.renderCount = 1;
    parameter.padTop = 1;
    parameter.padBottom = 1;
    parameter.padLeft = 1;
    parameter.padRight = 1;
    BitmapFont font = generator.generateFont(parameter);
    gridCoordFonts[i] = font;
}
generator.dispose();
```

The result looks much different than when I generate the `BitmapFont` through Hiero (with the default parameters) and load it in as follows:

```java
Texture fontTexture = new Texture(Gdx.files.internal("assets/calibri16.png"), true);
fontTexture.setFilter(Texture.TextureFilter.Nearest, Texture.TextureFilter.Nearest);
font = new BitmapFont(Gdx.files.internal("assets/calibri16.fnt"), new TextureRegion(fontTexture), false);
```

I tried to play with the hinting parameter, but it doesn't change the output to that of the Hiero case.

With `FreeTypeFontGenerator`:

![freetype_generator](https://user-images.githubusercontent.com/16304918/120214743-2e410380-c235-11eb-9dde-7773d2ea1466.png)

With Hiero:

![freetype_hiero](https://user-images.githubusercontent.com/16304918/120214758-313bf400-c235-11eb-883c-84e8845e728d.png)

#### Version of LibGDX and/or relevant dependencies
libgdx 1.10.0
gdx-freetype 1.10.0

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [X] Windows
- [ ] Linux
- [ ] MacOS

MontyHimself:
[...]

PR #6550, opened 2021-06-01
Update freetype to v2.11.1 and use submodule.

PokeMMO:
Github actions workflow already checks out submodules, allowing an authenticated non-duplicated copy.

~~Fixes https://nvd.nist.gov/vuln/detail/CVE-2020-15999~~

Future updates are as simple running a fetch + checkout in the subdirectory.


tommyettinger:
Some non-Chrome-specific info: https://www.freetype.org/index.html#news

That references this info on SourceForge:

> A heap buffer overflow has been found  in the handling of embedded
    PNG bitmaps, introduced in FreeType version 2.6.
      https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15999
    If you  use option  FT_CONFIG_OPTION_USE_PNG  you  should  upgrade
    immediately.

Can we confirm that we currently use or enable the use of `FT_CONFIG_OPTION_USE_PNG`?

PokeMMO:
Taking a closer look, I think we don't actually implement that option.

tommyettinger:
I mean, it could be nice to update FreeType just to get other changes in the late 2.9.x branch and the first few releases of 2.10.x, even if we don't need the security patch. #6547 might (big maybe) be affected by using newer FreeType, if there is a bug in FreeType and not in Hiero.

yuripourre:
Awesome!

Issue #5354, opened 2018-08-26
Feature request: add sRGB internal format of texture


Kright:
Internal format `sRGB` is same as `RGB`, but openGL automatically does gamma-correction when reading colors from texture. This is used in HDR rendering.

### Way to add this: 

code in [GLTexure.uploadImageData](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/GLTexture.java): 
```
Gdx.gl.glTexImage2D(target, miplevel, pixmap.getGLInternalFormat(), pixmap.getWidth(), pixmap.getHeight(), 0, pixmap.getGLFormat(), pixmap.getGLType(), pixmap.getPixels());
```
So, getGLInternalFormat is returned by [Pixmap](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Pixmap.java): 
```
public int getGLInternalFormat () {
	return toGlFormat(format);
}
```

but toGlFormat never returns sRGB:
```
public static int toGlFormat (int format) {
	switch (format) {
	case GDX2D_FORMAT_ALPHA:
		return GL20.GL_ALPHA;
	case GDX2D_FORMAT_LUMINANCE_ALPHA:
		return GL20.GL_LUMINANCE_ALPHA;
	case GDX2D_FORMAT_RGB888:
	case GDX2D_FORMAT_RGB565:
		return GL20.GL_RGB;
	case GDX2D_FORMAT_RGBA8888:
	case GDX2D_FORMAT_RGBA4444:
		return GL20.GL_RGBA;
	default:
		throw new GdxRuntimeException("unknown format: " + format);
	}
}
```

As I think, one can add a few new formats or/and special flag to the Texture constructor.

P.S. Now I am achieving this by overriding ```Pixmap.getGLInternalFormat()``` method and returning something like ```GL30.GL_SRGB8_ALPHA8``` or ```GL30.GL_SRGB8```

EtK2000:
If you want this functionality, the best way it to submit a pull request yourself.

mgsx-dev:
Another cleaner way to have sRGB texture is to do something like this, given a pixmap : 
```java
TextureData data = new GLOnlyTextureData(pixmap.getWidth(), pixmap.getHeight(), 0, GL30.GL_SRGB8, GL20.GL_RGB, GL20.GL_UNSIGNED_BYTE);
texture = new Texture(data);
texture.draw(pixmap, 0, 0);
[...]

PR #6553, opened 2021-06-03
Implemented GL30's sRGB Format

Timvisser1994:
#5354 Feature request: add sRGB internal format of texture

Work done:

1. Added `SRGB` format support for `Gdx2DPixmap`. This format lives inside of `GL30`.
2. Code cleanup (changed `if-statements` to `switch-statements` for better readability). This should also enable the compiler to optimize better.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Gdx2DPixmap.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Gdx2DPixmap.java
@@ -22,23 +22,25 @@ import java.io.InputStream;
 import java.nio.ByteBuffer;
 
 import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.GL30;
 import com.badlogic.gdx.utils.Disposable;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 
 /** @author mzechner */
 public class Gdx2DPixmap implements Disposable {
-	public static final int GDX2D_FORMAT_ALPHA = 1;
-	public static final int GDX2D_FORMAT_LUMINANCE_ALPHA = 2;
-	public static final int GDX2D_FORMAT_RGB888 = 3;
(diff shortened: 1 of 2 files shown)


mgsx-dev (changes requested):
@Timvisser1994 your code doesn't work for me, it crashes, did you tested it? it's better to provide a test case for new features. Also, as you can read in the [issue comment](https://github.com/libgdx/libgdx/issues/5354#issuecomment-867729632) i just posted, i don't think it's the right approach of implementing it.

Feel free to change your PR (and provide test case) or close it if you're not sure how to do that.

Timvisser1994:
@mgsx-dev You are right... this does not work.

image

And I agree with you comment, will rework this PR.

gdx/src/com/badlogic/gdx/graphics/g2d/Gdx2DPixmap.java
```diff
@@ -22,23 +22,25 @@
...
 
 /** @author mzechner */
 public class Gdx2DPixmap implements Disposable {
-	public static final int GDX2D_FORMAT_ALPHA = 1;
-	public static final int GDX2D_FORMAT_LUMINANCE_ALPHA = 2;
-	public static final int GDX2D_FORMAT_RGB888 = 3;
-	public static final int GDX2D_FORMAT_RGBA8888 = 4;
-	public static final int GDX2D_FORMAT_RGB565 = 5;
-	public static final int GDX2D_FORMAT_RGBA4444 = 6;
+	public static final short GDX2D_FORMAT_ALPHA = 1;
```
badlogic, line 31: Why `short`?

badlogic (changes requested):
Looks good in general, except for turning the `int` constants into `short` constants. That's an unnecessary breaking change.

Also the CHANGES file should note that this new pixmap format was added.

PR #6554, opened 2021-06-04
Fix asset manager backslash paths (and store in Unix-style)

Crunkle:
This change fixes double-backslash paths (`a\\b\\c.png`) and helps standardize the behavior of file names given to the asset manager (i.e. always prefer and store in Unix-style format). Only loading via `AssetDescriptor` and `AssetManager::unload` behave as expected before this change.

Should be the minimal change required outside of overriding the map itself and no breaking change for existing users.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -277,7 +277,7 @@ public class GlyphLayout implements Poolable {
 		for (int i = 0; i < runsSize; i++) {
 			GlyphRun run = (GlyphRun)runsItems[i];
 			float[] xAdvances = run.xAdvances.items;
-			float runWidth = xAdvances[0], max = 0;
+			float runWidth = run.x + xAdvances[0], max = 0; // run.x is needed to ensure floats are rounded same as above.
 			Object[] glyphs = run.glyphs.items;
 			for (int ii = 0, nn = run.glyphs.size; ii < nn;) {
 				Glyph glyph = (Glyph)glyphs[ii];
@@ -286,7 +286,7 @@ public class GlyphLayout implements Poolable {
 				ii++;
 				runWidth += xAdvances[ii];
 			}
-			run.width = Math.max(runWidth, max);
+			run.width = Math.max(runWidth, max) - run.x;
 			width = Math.max(width, run.x + run.width);
 		}
 		this.width = width;


Crunkle:
@NathanSweet hate to bother again but you're familiar with the asset manager so would appreciate a review.

NathanSweet:
Is this needed to fix a problem? Before the PR asset descriptor `fileName` replaced backslash with forward slash (I guess because it may come from the filesystem) and all code calling AssetManager methods pass strings that use forward slash. After the PR AssetManager methods can be passed strings that contain backslashes, but why would code be doing this?

Crunkle:
> Is this needed to fix a problem?

Yes, `AssetManager::load(String, Class)` is public and does not contain the replacement logic. Neither do any of the other methods that this covers. When loading external assets, it isn't unheard of to see backslash paths exported by programs.

Consider the following code where the logic is counter-intuitive:

```java
assetManager.load("a\\b.png", Texture.class);
assetManager.contains("a/b.png"); // fails

assetManager.load("a\\\\b.png", Texture.class);
assetManager.contains("a/b.png"); // fails
```

Or, for a more complete scenario I recently encountered, consider a custom loader where `getDependencies` returns a list of `AssetDescriptor` with backslash format then later tries to fetch that same resource, only to fail despite being loaded:

```java
getDependencies:
  return [new AssetDescriptor("my\asset.png")]

loadAsync:
  assetManager.get("my\asset.png"); // fails
```

On top of that, the previous logic did not replace double backslashes commonly found on Windows.

NathanSweet:
I understand you _could_ pass backslashes to the methods, but why would you do that? The paths passed to methods are almost always strings in app code. Forward slash is easier to type.

Doing replacement may clear up one potential confusion but causes new ones (probably just as unlikely):
```
assetManager.load("a\\b.png", Texture.class);
assetManager.contains("a/b.png"); // succeeds (surprise)
assetManager.load("a/b.png", Texture.class); // different paths are actually the same asset
```

Maybe it would be better if we completely avoid changing the paths (both in AssetDescriptor and the other methods). If we just use the paths the user provides, then we don't have to normalize the input all over the place, which seems cleaner.

In my AssetManager usage I've never passed it backslashes and relied on the current replacement in AssetDescriptor (which has been there since the beginning). In what scenarios might users be relying on that replacement? Are users getting a path from a File object and giving it to AssetManager? I'm not sure that is common. The current replacement tries to help the user in that scenario but it would reduce confusion to remove it.

Crunkle:
In our case, on a fairly sized project, we have an auto-generated JSON file describing an asset which references other assets relative to the configuration. For example, an image asset to be loaded may be described as:

```json
{ "x": 10, "y": 20, "texture": "images\\sky.png" }
```

Our loaders will read that JSON (generated by various higher-level tools or manual input) and submit a request to load `images\sky.png` which is then converted to `images/sky.png` automatically. Any future attempts to fetch the asset `images\sky.png` fail as it is being stored with a different key than requested upon load.

I would consider the separators `/` and `\` to be equivalent but agree that the problem lies within `AssetDescriptor` changing the paths seemingly without reason. I will open a second PR to address this, but am not sure what the original rationale was for adding this conversion in 2011?

PR #6555, opened 2021-06-04
Remove backslash conversion from asset manager paths

Crunkle:
Removes the asset backslash replacement for more predictable asset paths (see #6554 for discussion and rationale).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,7 @@
 - LWJGL 3 is now the default desktop backend. If you want to port your existing applications, take a look here: https://gist.github.com/crykn/eb37cb4f7a03d006b3a0ecad27292a2d
 - Brought the official and third-party extensions in gdx-setup up to date. Removed some unmaintained ones and added gdx-websockets & jbump.
 - API Fix: Escaped characters in XML attributes are now properly un-escaped
+- Bug Fix: AssetManager backslash conversion removed - fixes use of filenames containing backslashes
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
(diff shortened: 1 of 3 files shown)


tommyettinger:
This affects the ability to copy JARs or their data from Windows, where backslash or slash can separate directories, to other OSes, where backslash is a valid part of a filename. This seems like a bad idea and breaks compatibility, subtly. In the referenced issue, you could probably process paths with backslashes into paths with slashes, in your generating code, without any real issue.

tommyettinger:
I removed the `breaking change` label based on discussion with Crunkle on the Discord. I'm not 100% clear on where backslashes even work currently, in AssetManager or libGDX file handling in general. If backslashes just don't work reliably before the PR, it can't be seen as breaking what's non-usable at this time. I think any solution needs to incorporate some kind of documentation on what types of paths are permitted and/or recommended, and notes of possible pitfalls where there are any.

NathanSweet:
While Windows uses backslashes for paths, I don't think I've run into a situation on Windows where forward slashes instead of backslashes did not work. I just always use forward slash and it works on Windows and other OSes. When reading files on Windows backslashes are given by default and sometimes need replacing.

For AssetManager I think it makes sense that the path you specify when loading an asset is also the path that finds that loaded asset. The current replacing of backslashes with forward slashes prevents that. I haven't personally had a problem with that since I either don't get my asset paths from the filesystem or I replace backslash with forward slash in app code.

Also AssetManager shouldn't assume that the "path" is a file path, rather it is more like an "asset ID" which happens to often be a file path.

Given those reasons I think removing the backslash-to-forward-slash replacement in AssetManager makes sense. Once it no longer has such surprising behavior, I don't even think we need to document it: you load an asset with a path and you use the same path to find it later. Any existing app code that relies on the replace will break. Hopefully that breakage would be pretty obvious.

Let's give some time to see if anyone is against merging.

Crunkle:
@NathanSweet still keen to get this in for the sake of clarity; it has been a while with no objections.

crykn (changes requested):
Hey @Crunkle, would you mind adding a short entry to the CHANGES file?

Issue #5759, opened 2019-08-16
ParticleSystem multiple ParticleEffects with different Textures


cubesdev:
#### Issue details
The 3D ParticleSystem class does not work correctly as soon as multiple billboard ParticleEffects are loaded with different texture resources in their JSON 'assets' tag: All particles will share the same Texture.

Could be fixed if different effects were using their own BillboardParticleBatch type. But this is currently not working properly because you have to add (and construct) ParticleBatch objects manually. The ParticleEffectLoader does not take care of this step. The current `BillboardParticleBatch#setTexture()` method will override the texture for all future drawn effects, ignoring the assigned texture for a specific effect.

#### Reproduction steps/code
For example:

`   assets:[
      {
         filename:particle_type_one.png,
         type:com.badlogic.gdx.graphics.Texture
      }
   ]`

and
`   assets:[
      {
         filename:particle_type_two.png,
         type:com.badlogic.gdx.graphics.Texture
      }
   ]`

in a different billboard ParticleEffect.

I initialized the ParticleSystem using this code:  
`mParticleSystem = new ParticleSystem();`
`mBillboardBatch = new BillboardParticleBatch();`
`mParticleSystem.add(mBillboardBatch);`

#### Version of LibGDX and/or relevant dependencies
LibGDX 1.9.10

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] MacOS

Not tested on any other platform except Windows. It should be a general problem and I assume it will fail on all other platforms, too.

Timvisser1994:
When implementing this, there is another aspect coming in: the division that you want, e.q. 50/50 or 30/70. This would mean that `assets` will need to support an array of `{ filename..., type... }`.

I would suggest to append an optional third parameter there: `{ filename..., type..., weight... }` where `weight` is a float ranging from `0.0` to `1.0` and where the sum of all weights is `1.0`. When no `weight` is specified, the remaining weight (`1.0` minus the sum of the specified weights) should be divided evenly across all entries that don't have a `weight` specified.

With this `weight`, one can have multiple particle assets that are... well... weighted. E.q. 30% balloons, 45% sparkles and the remaining 25% confetti. (In that case, the weights are: `0.3`, `0.45` and `0.25`, totaling `1.0`)

[...]

PR #6558, opened 2021-06-08
ParticleSystem multiple ParticleEffects with different Textures

Timvisser1994:
ParticleSystem multiple ParticleEffects with different Textures #5759

Removed overarching state management in `BillboardParticleBatch` that forced the `RenderablePool` to lose it's already defined previous Textures.

_This needs to be tested_


tommyettinger:
This seems like a very bad way to go about this! In this PR, Renderables are cleared from the tracking Array, but not freed, even though the code right after that depends on the status of what should have been freed from the Pool. I think the solution will not be so simple as just removing one `freeAll()` call.

Also, the indentation got wrecked in FlameMain, for one large section. This just makes it harder to read the Git diff, since I think all of the code is the same, just with improper indentation now.

PR #6562, opened 2021-06-12
Fix clear method

### LINKED ISSUE #6561: Pool.clear() doesn't clear all elements from the internal Array<T> freeObjects
#### Issue details
`Pool.clear()` doesn't clear all elements from the internal `Array<T> freeObjects`. 
The size of the array is reduced inside the loop every time when freeObjects.pop() is called, which means that `freeObjects.size` will return a smaller value and `i < freeObjects.size` condition will not allow to remove all elements.

The issue leads to a very unpleasant bugs, for example if you have a pool with objects storing reference to box2d bodies.
I believe it may lead to other serious bugs because Pool is used in many parts of LibGDX itself.

This part of the Pool.java code is broken:
```
	public void clear () {
		for (int i = 0; i < freeObjects.size; i++) {
			T obj = freeObjects.pop();
			discard(obj);
		}
	}
```

The solution could be:
```
	public void clear () {
		while (freeObjects.size > 0) {
			T obj = freeObjects.pop();
			discard(obj);
		}
        }
```

#### Reproduction steps
1, Create `Pool` object with any objects inside.
2. Call `.obtain()` method of the pool multiple times (5+) to get enough objects
3. Free all obtained objects by calling `free(T object)` method of the pool (probably same will happen for `freeAll (Array<T> objects) `)
4. Call `.clear()` method of the pool once
5. Check the size of the internal `freeObjects` array by calling `getFree()` method of the pool

Expected result: 0 elements remain in the internal array
Actual results: ~half of elements remains in the internal array

#### Version of LibGDX
LibGDX 1.10.0

Related to this [commit](https://github.com/libgdx/libgdx/commit/049174598204936eab55375079944983c39530e1#diff-2b502e40add95b689c7f66b2f0a0fa97d4b1d1961881443a6e4a7d470a2c6208R115)

#### Affected platforms
- [x] Android
- [x] iOS
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

yuripourre:
Thanks @tscissors

Closes #6561

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Pool.java
+++ b/gdx/src/com/badlogic/gdx/utils/Pool.java
@@ -112,7 +112,7 @@ abstract public class Pool<T> {
 
 	/** Removes and discards all free objects from this pool. */
 	public void clear () {
-		for (int i = 0; i < freeObjects.size; i++) {
+		while (freeObjects.size > 0) {
 			T obj = freeObjects.pop();
 			discard(obj);
 		}


NathanSweet:
Oops, I fixed it when I saw tscissors issue, before I saw this one. Sorry!

tscissors:
Thank you guys for the fast response and the fix!

PR #6564, opened 2021-06-13
FirstPersonCameraController, input keys public to allow mapping changes

fabiitch:
```	private int STRAFE_LEFT = Keys.A;
	private int STRAFE_RIGHT = Keys.D;
	private int FORWARD = Keys.W;
	private int BACKWARD = Keys.S;
	private int UP = Keys.Q;
	private int DOWN = Keys.E;
	private float velocity = 5;```
Just put it public to can change it. (I got azerty keyboard) or map it with arrows

Thanks for reading

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
@@ -26,20 +26,18 @@ import com.badlogic.gdx.utils.IntIntMap;
 /** Takes a {@link Camera} instance and controls it via w,a,s,d and mouse panning.
  * @author badlogic */
 public class FirstPersonCameraController extends InputAdapter {
-	 protected final Camera camera;
-	 protected final IntIntMap keys = new IntIntMap();
-	 protected final Vector3 tmp = new Vector3();
-
-	 public boolean autoUpdate = true;
-	 public float velocity = 5;    /* Velocity in units per second for moving forward, backward and strafing left/right.*/
-	 public float degreesPerPixel = 0.5f;    /* Sets how many degrees to rotate per pixel the mouse moved.*/
-
-	 public int strafeLeftKey = Keys.A;
-	 public int strafeRightKey = Keys.D;
-	 public int forwardKey = Keys.W;
-	 public int backwardKey = Keys.S;
-	 public int upKey = Keys.Q;
-	 public int downKey = Keys.E;
+	protected final Camera camera;
+	protected final IntIntMap keys = new IntIntMap();
+	public int strafeLeftKey = Keys.A;
+	public int strafeRightKey = Keys.D;
+	public int forwardKey = Keys.W;
+	public int backwardKey = Keys.S;
+	public int upKey = Keys.Q;
+	public int downKey = Keys.E;
+	public boolean autoUpdate = true;
+	protected float velocity = 5;
+	protected float degreesPerPixel = 0.5f;
+	protected final Vector3 tmp = new Vector3();
 
 	public FirstPersonCameraController (Camera camera) {
 		this.camera = camera;
@@ -55,6 +53,18 @@ public class FirstPersonCameraController extends InputAdapter {
 	public boolean keyUp (int keycode) {
 		keys.remove(keycode, 0);
 		return true;
+	}
+
+	/** Sets the velocity in units per second for moving forward, backward and strafing left/right.
+	 * @param velocity the velocity in units per second */
+	public void setVelocity (float velocity) {
+		this.velocity = velocity;
+	}
+
+	/** Sets how many degrees to rotate per pixel the mouse moved.
+	 * @param degreesPerPixel */
+	public void setDegreesPerPixel (float degreesPerPixel) {
+		this.degreesPerPixel = degreesPerPixel;
 	}
 
 	@Override
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
@@ -99,5 +99,4 @@ public class FirstPersonCameraController extends InputAdapter {
 		if (autoUpdate)
 			 camera.update(true);
 	}
-
 }
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
@@ -28,12 +28,13 @@ import com.badlogic.gdx.utils.IntIntMap;
 public class FirstPersonCameraController extends InputAdapter {
 	private final Camera camera;
 	private final IntIntMap keys = new IntIntMap();
-	public int strafeLeftKey = Keys.A;
-	public int strafeRightKey = Keys.D;
-	public int forwardKey = Keys.W;
-	public int backwardKey = Keys.S;
-	public int upKey = Keys.Q;
-	public int downKey = Keys.E;
+	private boolean autoUpdate = true;
+	private int strafeLeftKey = Keys.A;
+	private int strafeRightKey = Keys.D;
+	private int forwardKey = Keys.W;
+	private int backwardKey = Keys.S;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by yuripourre — CHANGES_REQUESTED** [review]
They shouldn't be public, if you really want to expose them, they should be protected so you can extend the class. And if so, they should be converted to attributes (camelCase).

**Comment by fabiitch** [comment]
Hi ! 
I rename fiels in camel case.
I think they should be public, it allow mapping change after constructor and in child class.
```
FirstPersonCameraController cameraController = new FirstPersonCameraController(camera);
		 cameraController.upKey = Input.Keys.UP;
		 cameraController.downKey= Input.Keys.DOWN;
		 cameraController.strafeLeftKey= Input.Keys.LEFT;
		 cameraController.strafeRightKey= Input.Keys.RIGHT;
```

**Comment by crykn** [comment]
I think the fields should either be protected or have a public setter (like `velocity` and `degreesPerPixel` already do).

**Comment by fabiitch** [comment]
-_-
i'm not a fan of getter/setter, i really pref public fields.

I repush, all fields private and add getter/setter ( i can change to all public ! no prob!)
i also add an "autoUpdate" field like CameraInputController  
`	if (autoUpdate) camera.update();`

**Comment by fabiitch** [comment]
I "rework" it , i really pref this syle.
field  public for expose and protected for internal.
```	 
public class FirstPersonCameraController extends InputAdapter {
         protected final Camera camera;
	 protected final IntIntMap keys = new IntIntMap();
	 protected final Vector3 tmp = new Vector3();

	 public boolean autoUpdate = true;
	 public float velocity = 5;    /* Velocity in units per second for moving forward, backward and strafing left/right.*/
	 public float degreesPerPixel = 0.5f;    /* Sets how many degrees to rotate per pixel the mouse moved.*/

	 public int strafeLeftKey = Keys.A;
	 public int strafeRightKey = Keys.D;
	 public int forwardKey = Keys.W;
	 public int backwardKey = Keys.S;
	 public int upKey = Keys.Q;
	 public int downKey = Keys.E;
```
What do you think ?

Thanks for reading

**Comment by tommyettinger** [comment]
@yuripourre Are we looking at the same libGDX? I almost never see `protected`, but `public` fields are used all over. For instance, https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/Array.java#L31-L39 uses public or private but has no protected fields (and a public `size` but no `getSize()`, so you can't override its access), https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Button.java#L45-L49 uses public, private, and package-private (which is always annoying to encounter when using the library), https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Color.java#L72 uses mutable public fields even for what should be constant Color values... I'm not saying I agree with this decision where it has already been made, but "it uses public fields" is not a reason to block this PR, since it's in line with much of libGDX already.

**Comment by yuripourre** [comment]
@tommyettinger you are right, it's been a while since I navigate through libGDX I could swear that the standard was private methods.

Sorry about that.

In this case then, I am all in for the changes.

**Review by yuripourre — CHANGES_REQUESTED** [review]
Ok, I am convinced that the fields can be public. If they are public camelCase I will approve :)

**Comment by tommyettinger** [comment]
@yuripourre Confusingly, some of the graphics code does use getters and setters for interface reasons. I think some devs here have said they don't like using getters and setters (maybe because of JSON issues with non-public fields?). There is a pretty wild set of coding standards being used in libGDX. I'm in favor of getters and setters wherever interfaces are relevant, but I'm not really opposed to public fields.

Using camelCase is my preference too, for non-constant fields.

**Comment by fabiitch** [comment]
Hey, thanks for all feedbacks !

Finnally i set like it : 
``` 
public class FirstPersonCameraController extends InputAdapter {
	 protected final Camera camera;
	 protected final IntIntMap keys = new IntIntMap();
	 protected final Vector3 tmp = new Vector3();

	 public boolean autoUpdate = true;
	 public float velocity = 5;    /* Velocity in units per second for moving forward, backward and strafing left/right.*/
	 public float degreesPerPixel = 0.5f;    /* Sets how many degrees to rotate per pixel the mouse moved.*/

	 public int strafeLeftKey = Keys.A;
	 public int strafeRightKey = Keys.D;
	 public int forwardKey = Keys.W;
	 public int backwardKey = Keys.S;
	 public int upKey = Keys.Q;
	 public int downKey = Keys.E;
```

Public for keyCodes and velocity
Protected for camera, tmpVector, keysMap

File: gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java [hunk 1]
```diff
@@ -26,17 +26,20 @@
 /** Takes a {@link Camera} instance and controls it via w,a,s,d and mouse panning.
  * @author badlogic */
 public class FirstPersonCameraController extends InputAdapter {
```
mgsx-dev, line 28: API should be documented in CHANGES file, something like "FirstPersonCameraController key mapping is now configurable"

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
```diff
@@ -26,17 +26,20 @@
 /** Takes a {@link Camera} instance and controls it via w,a,s,d and mouse panning.
  * @author badlogic */
 public class FirstPersonCameraController extends InputAdapter {
-	private final Camera camera;
```
mgsx-dev, line 29: could you please keep fields ordering? it's better for history/diffs

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
```diff
@@ -26,17 +26,20 @@
...
-	private int STRAFE_LEFT = Keys.A;
-	private int STRAFE_RIGHT = Keys.D;
-	private int FORWARD = Keys.W;
-	private int BACKWARD = Keys.S;
-	private int UP = Keys.Q;
-	private int DOWN = Keys.E;
-	private float velocity = 5;
-	private float degreesPerPixel = 0.5f;
-	private final Vector3 tmp = new Vector3();
+	 protected final Camera camera;
```
mgsx-dev, line 29: indentation issue (unwanted extra space)

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
```diff
@@ -54,26 +57,13 @@ public boolean keyUp (int keycode) {
 		return true;
 	}
 
-	/** Sets the velocity in units per second for moving forward, backward and strafing left/right.
```
mgsx-dev, line 57: there is no reason to introduce a breaking change here, please revert setter removal and keep related fields protected.

yuripourre (approved):

mgsx-dev (changes requested):

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
fabiitch: done

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
fabiitch: done

fabiitch:
I hope its ok , thanks for review and interest ! 
next version i can change my mapping keys ! 🍡 🍡

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
fabiitch: done

gdx/src/com/badlogic/gdx/graphics/g3d/utils/FirstPersonCameraController.java
fabiitch: done

mgsx-dev (approved):
LGTM

yuripourre (approved):

PR #6567, opened 2021-06-14
Made "Type" and "Probability" first-class citizens on a TiledMapTile

Timvisser1994:
This way, one can access them by both `tile.getProperties().get("type")` and `tile.getType()`. This is closer to the truth given the following .tmx snippet:

```xml
<tileset firstgid="1" name="grass-tiles-2-small" tilewidth="32" tileheight="32" tilecount="72" columns="12">
  <image source="grass-tiles-2-small.png" width="384" height="192"/>
  <tile id="13" type="Grass" probability="0.5"/>
</tileset>
```

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmxMapLoader.java
@@ -197,7 +197,7 @@ public class AtlasTmxMapLoader extends BaseTmxMapLoader<AtlasTmxMapLoader.AtlasT
 					AtlasRegion region = atlas.findRegion(regionName);
 					if (region == null)
 						throw new GdxRuntimeException("Tileset atlasRegion not found: " + regionName);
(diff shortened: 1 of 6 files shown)


lyze237:
As already discussed in #6464 this is not a workable solution. As any class with "Tiled" is a generic tilemap and therefore unrelated to Tiled (tmx). So adding interface variables to that is not good and accessing them via .getProperties() is the intended way.

Timvisser1994:
@lyze237 Totally get that they are too Tiled-specific, which is not intended... But there should be a way to read out `type` from the .tmx file (just as there will be one to read out `probability`).

Is this something that we should generalize, for example: looping over all XML properties and adding them to the `MapProperties` e.q.:

```xml
<tile id="13" type="Grass" probability="0.5"/>
```
would yield properties `id = 13`, `type = "Grass"` and `probability = "0.5"`

This, of course, on top of the existing functionality:
```xml
<tile id="13" type="Grass" probability="0.5">
   <properties>
    <property name="myProperty" value="Hello there"/>
   </properties>
</tile>
```

PR #6569, opened 2021-06-16
Protect Android MediaPlayer from getting into invalid state

obigu:
On https://github.com/libgdx/libgdx/pull/6472 the check to verify `MediaPlayer` was playing before calling `pause()` was removed but it's actually needed in case `MediaPlayer` is on `Stopped` state to prevent entering `Error` state from which it doesn't recover.

To reproduce the issue just run `MusicTest` and:
1. Play a song
2. Stop it
3. Pause it
4. Try to play it again, it doesn't play (also can check stacktrace)


MrStahlfelge (approved):
Thanks for fixing it

sakib-rahman-bangladesh (commented):
ooking

PR #6571, opened 2021-06-19
Maintenance on default Setup tool and tests Android manifests

obigu:
Just a couple of minor changes
- Removed the default Android `GdxTheme`. Afaik there's currently no need to have an Android theme defined (it seems on Android 2.2 it was needed to fix some fullscreen issue). One less thing beginners need to care about and one less unnecessary file on the app. Also, the theme itself is ugly and, as a default,  IMO it's better to just let each Android device to use its default theme.
- Changed the `installLocation` of the Tests libGDX manifest to `auto`. The app is small size and for testing purposes only. Testing stuff such as Live Wallpapers on `external` may not be the best choice according the the docs (https://developer.android.com/guide/topics/data/install-location#Should). Also removed outdated Ouya references.


yuripourre (approved):
Seems good to me

MrStahlfelge (approved):
Good for me too!

PokeMMO:
After making the same changes to my app, I'm seeing a fullscreen title bar that does not go away on LineageOS Android 11.

Can anyone confirm they see the same or do not have this issue?

tommyettinger:
Maybe that default Android theme is actually necessary in some cases, like fullscreen on LineageOS? I still am having issues testing on any kind of Android system, so I'm not sure I'll be able to help much...

MrStahlfelge:
It might be, some device manufacturers (Lineage is nothing else) change the default theme and so an own one is perhaps needed. How does this title bar look like?

PokeMMO:
screen

It looks like this. Ignore the super secure blur.

It also shows a white background when doing initial (blocking) loading. Compared to the old full-black screen.

MrStahlfelge:
If it is like that we should add back the theme!

obigu:
Is that the status bar? Is immersive mode on in your config? If it’s not
and it’s a bar specific to the OS, can’t it be enabled/disabled on Settings?

PokeMMO:
Immersive mode is on. Immersive mode off still ugly, but in a different way.

screen

I'd be curious if anyone has a stock pixel phone to test on, since lineageos and pixels should both be pretty close to AOSP.

MrStahlfelge:
Emulator is like pixel phone

obigu:
I do have a Pixel 3, never seen that bar before. I removed GdxTheme from
all my games years ago and never received any complain about not being
fullscreen.

Does this bar appear on other apps/games? My guess is this is LineageOS
soecific. May it be related to the Navigation Bar, some posts talk about a
Setting in  Settings -> Display -> Expanded desktop .

DavidPDev:
Sorry for re-opening this issue...
Updating to the last Gdx build tools (gdx-setup.jar) I am having an issue because of the removal of the GdxTheme from styles, and from the manifest. Only on Android, of course.
When launching the game it appears a window with title bar during a few milliseconds.

Screenshot_1666264264 - copia

After a moment the main window of the game appears over the screen, but the effect is quite ugly.
Tested on some emulators and physical devices. Also tested in Android target sdk 30,31 and 32.

I have done a quick "fix" that  is working in my case, but not sure if it might have some side-effects.

On manifest add:
``
android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
``

image

What do you think?

Tom-Ski:
Yeah I would agree putting this back

obigu:
Sure, we can add it but I'd like to test it first. You say you reproduce it on Simulator I assume with a newly created app with Gdx-Setup. Which Device and Android API shows that behaviour?

Tom-Ski:
I don't think any default themes should be imposed, so minimal case of 2 would be better IMO. 

I have a Huawei mate 10pro that was doing it, also a pixel 4

DavidPDev:
Yes, tested on emulator with x86_64 api 29 and many other configurations.
Maybe you have to force some "delay" (Thread.sleep)  in the OnCreate before the initialize call to show the effect.

DavidPDev:
> I don't think any default themes should be imposed (added by us I mean), so minimal case of 2 would be better IMO.
> 
> I have a Huawei mate 10pro that was doing it, also a pixel 4

@Tom-Ski how did you fix it? adding the old GdxTheme file?

Tom-Ski:
I am using `android:theme="@android:style/Theme.NoTitleBar.Fullscreen"` at the application level. Not sure if there are any backwards compatibility issues here though

DavidPDev:
> I am using `android:theme="@android:style/Theme.NoTitleBar.Fullscreen"` at the application level. Not sure if there are any backwards compatibility issues here though

Perfect. Thank you!
I also think this is a clean fix.

DavidPDev:
Thank you very much @obigu for so a quick response and for taking your time to test it!
I have no idea how to do the push request. Sorry :(

obigu:
I've given it some more thought and the proposed solution is not ideal. `Theme.NoTitleBar.Fullscreen` is an old style (pre Material and even pre Holo) which means any native element on the app (such an alert native dialog) will not look good. Currently it uses default for the device which is ok.

Solution is to define separate styles for different api versions. I'll make a proposal on a PR at some point.

DavidPDev:
Yes, I agree. Anyways I think the default libGdx game should be fullscreen with no title... but, of course this is not mandatory...
Again, thank you!, for now I'm happy using the ``Theme.NoTitleBar.Fullscree``, it's even better and cleaner than the old styled GdxTheme.

MobiDevelop:
Yeah, we should provide a sensible default but, frankly, anyone actually releasing stuff on Android should probably be preparing their own theme anyway.

PR #6572, opened 2021-06-19
Fix for compile-time error with OctreeNode on GWT

tommyettinger:
I'm not sure why, but GWT gives a compilation error when it compiles Octree, and then finishes the build successfully. This PR changes the usage of `OctreeNode` to `Octree.OctreeNode` within the inner class `OctreeNode`; this shouldn't matter but apparently does for GWT. I added OctreeTest to the GWT tests to verify. Without this PR, the compile-time error is there, but with it, there is just the often-seen warning about unchecked types. OctreeNode remains `protected` with this, and all other privacy levels are the same.

I'm not sure if this matters, but it's possible the compilation error is a sign that something else may fail soon after, or will fail in newer versions of GWT. It's probably best just for appearances to not have compilation errors in libGDX's tests, even if they don't cause problems.


yuripourre (approved):

tommyettinger:
Thanks for the review, @yuripourre ! I'm glad you wrote the OctreeTest in the first place, which made verifying this easy, and I'm also glad that it runs on GWT (just kind of slowly, like most full-screen GWT apps).

I'm wondering if there could, at some point in the future, be a way to do an automated compilation check for GWT. It's the least forgiving of the platforms, with lots of missing APIs that are sometimes filled in by libGDX and sometimes not. I think a priority would be to update to the current GWT 2.9.0 release, as [my fork of MrStahlfelge's backends](https://github.com/tommyettinger/gdx-backends) does (with most of the work recently done by raeleus). I've encountered no issues so far with GWT 2.9.0, and the only GWT-related issues that needed serious work recently were related to Gradle 7.0 (now 7.1) breaking compatibility with a dependency of the HTML Gradle modules.

MrStahlfelge (approved):
Good change

MrStahlfelge:
> I'm wondering if there could, at some point in the future, be a way to do an automated compilation check for GWT.

We already have that and since we introduced it the situation became better a lot. The problem here was that this was just a compiler warning and did not stop the compile.

PR #6573, opened 2021-06-21
Polygon add method getVertex, getNbVertices, getCentroid

fabiitch:
Hi, i think this methods can be usefull.
Its my case, free to close if im solo ^^

Thanks for reading


gdx/src/com/badlogic/gdx/math/Polygon.java
```diff
@@ -175,6 +174,23 @@ public float area () {
 		return GeometryUtils.polygonArea(vertices, 0, vertices.length);
 	}
 
+	 public int getNbVertices(){
```
NathanSweet, line 177: This isn't a good name. Do you mean `getNumberVertices`? I would use `getVertexCount`.

fabiitch: yes, thanks
done

PR #6576, opened 2021-06-24
StringBuilder.toStringAndClear()

TCreutzenberg:
Hey guys,

I find myself typing this a lot:

```
    ...
    final String s = stringBuilder.toString();
    stringBuilder.clear();
    return s;
}
```

So I thought maybe somebody else would also like a convenience method for this?

```
    ...
    return stringBuilder.toStringAndClear();
}
```


sakib-rahman-bangladesh (commented):
ok

PR #6577, opened 2021-06-24
Fix GlyphLayout multiline with color markup

TCreutzenberg:
Hey guys,

this PR fixes a very specific problem: Wrong GlyphLayout for multi-line texts where a line after the first one several runs because of colormarkup. (Unfortunately this is not very obvious with the fonts provided with libgdx so I used my own BitmapFont for the screenshots where it is really obvious)

After going through the code for three hours I found the issue: While creating the runs in GlyphLayout#setText() "lastGlyph" is set to null even if there is no newline, but just a new run because of a colormarkup tag.

**Unfixed:**
One line: Everything works
unfixedOneLine

Several lines: This is not correct. The bug is the distance between the "(" and the "5"
unfixedTwoLines
unfixedThreeLines

**Fixed:**
The distance between "(" and "5" is correct now also for multiple lines
fixedOneLine
fixedTwoLines
fixedThreeLines

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@
 - API Addition: Polygon have method setVertex (int index, float x, float y).
 - API Addition: TMX built-in tile property "type" is now supported.
 - API Addition: Octree structure.
+- Fix: GlyphLayout with multiline texts with colormarkup uses the correct distance between GlyphLayout#GlyphRuns
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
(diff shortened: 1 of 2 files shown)


tommyettinger:
Wow, I feel your pain with GlyphLayout bugs like this; I've debugged them before. I'm very glad the fix is just one line! Do you have any idea why the built-in Arial-15 font doesn't show the behavior that you corrected? I wonder if this is related to an epsilon issue; we had a recent fix that involves checking equality between floats with some tolerance for imprecision. I approve, tentatively, because I haven't yet tried the code but it looks correct. I'll request Nate to look at this because any problems would directly affect his work.

tommyettinger (approved):
Tentative approval; we may need to try this out with a few different varieties of BitmapFont to make sure this works across the board.

NathanSweet (approved):

NathanSweet:
Looks like a good find! I applied it and tried it out in Spine. It doesn't break anything, though Spine doesn't use a whole lot of color markup. The fix looks correct and I'd be surprised if it broke other fonts, so I think it's good to go.

PR #6578, opened 2021-06-25
GlyphLayout: Fix width of color markup runs

TCreutzenberg:
While I was on the GlyphLayout yesterday I noticed 2 more bugs. One here and the other one will soon follow in another PR.

This PR fixes the width of color markup runs.

**Before:**
before

**After:**
after


TCreutzenberg:
Found some more issues while testing, sorry

PR #6579, opened 2021-06-25
GlyphLayout: Fix wrapping for color-markup-runs with just one character

TCreutzenberg:
Hey, I found another issue while I was doing my other GlyphLayout-PR yesterday, here is the fix for that:

Fixed wrapping of color-markup-runs with just one character (everthing works fine if the color-markup-run has more than 1 character).

I also adjusted the BitmapFontTest to include a 1-character color-markup-run and to use different colors for the debug-borders of the different runs.

State before this PR: The line is not wrapped properly if the targetWidth is at the right border of the "5".
wrappingBug

This does not happen any more with this PR, the lines get wrapped properly.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,7 +5,7 @@
 - API Addition: Polygon have method setVertex (int index, float x, float y).
 - API Addition: TMX built-in tile property "type" is now supported.
 - API Addition: Octree structure.
-- Fix: GlyphLayout with multiline texts with colormarkup uses the correct distance between GlyphLayout#GlyphRuns
+- Fix: GlyphLayout: Several fixes for color markup runs with multi-line or wrapping texts
 
 [1.10.0]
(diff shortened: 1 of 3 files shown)


tommyettinger (approved):
This looks correct; before there wasn't any check to make sure there was a second glyph in the run, but element 1 of the FloatArray's `items` was used anyway. If the FloatArray only had size 1, hat wouldn't give an IndexOutOfBoundsException, but would usually give incorrect results. I think empty runs are handled just before the changed block, so now all the bases are covered -- lengths 0, 1, and 2 or more.

NathanSweet:
The first item in `xAdvances` is the drawing X offset, so when there is one glyph there are sure to be 2 entries. The old code was assuming that the first glyph would fit (for some reason), but that was wrong since of course it may not fit. The PR looks good!

It also highlights another issue, that the glyph runs can overlap:

That should not be the case. I thought I had a simple fix, but nothing is simple in this beast. I'll create a new PR.

PR #6580, opened 2021-06-26
GlyphLayout: Removed redundant lastGlyph (also fixes overlapping GlyphRuns)

TCreutzenberg:
Hey guys,

While trying to fix overlapping GlyphRuns I found out that "Glpyh lastGlyph" is not necessary. Can you guys please confirm? Am I missing any edgecases?
I already tried a lot of fonts and I cannot see any problems.

PS: I cannot commit FreeTypeFontGenerator.java without Github thinking it's all new... the only change actually is the removal of "Glyph lastGlyph" from the method getGlyphs(GlyphRun, CharSequence, int, int) within... I tried everything already... Can somebody help or change my PR? Also reverting FreeTypeFontGenerator and committing it again won't help.
Worst case: Please somebody else recreate this PR and merge. Thanks!

fix

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
 - API Addition: TMX built-in tile property "type" is now supported.
 - API Addition: Octree structure.
 - Fix: GlyphLayout: Several fixes for color markup runs with multi-line or wrapping texts
+- API change: Removed redundant redundant "Glyph lastGlyph" from BitmapFontData#getGlyphs(GlyphRun, CharSequence, int, int) 
 
 [1.10.0]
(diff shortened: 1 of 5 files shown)


NathanSweet:
`lastGlyph` is necessary for BitmapFont `getGlyphs` to know the last glyph from the previous run. This allows kerning to be applied.

Without `lastGlyph`:

With `lastGlyph` (note the kerning for `Va`, another paid that often has kerning is `Te` or `To`):


I believe we want to keep the `lastGlyph` parameter, but note BitmapFont `getGlyphs` is public API, so changing it could break existing code. The idea is that someone could extend BitmapFont to do glyph layout differently.

I've just made a PR for improving the glyph bounds: #6581

NathanSweet:
BTW, I was also thinking the runs shouldn't overlap, but after getting back into the code I think it can't be avoided, like in the image above. The bounds were pretty poor before PR #6581 but even after runs can still overlap.

TCreutzenberg:
@NathanSweet Thanks for clarification about lastGlyph!

Issue #4672, opened 2017-03-28
word break cause by color tag


loohoo:
#### Issue details
the second draw of font should not break the word "international".  It just one word with different color.

#### Reproduction steps/code
```
	public void render(float delta) {
		Gdx.gl.glClearColor(0.6f, 0.6f, 0.6f, 1.0f);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
		
		batch.setProjectionMatrix(camera.combined);

		shape.begin(ShapeRenderer.ShapeType.Line);
		shape.setColor(Color.RED);
		shape.rect(0, VIRTUAL_HEIGHT - 600, 300, 600);
		shape.end();

		batch.begin();
		font.getData().setScale(0.5f);
		font.draw(batch, "[BLUE]Is this a bug? international string123456789abcdefghiklmn",
				0.0f, VIRTUAL_HEIGHT,
				300.f, Align.left, true);
              // [output](http://pasteboard.co/OPDave7xa.png)
		font.draw(batch, "[BLUE]Is this a bug? inter[RED]national string123456789abcdefghiklmn",
				0.0f, VIRTUAL_HEIGHT - 200,
				300.f, Align.left, true);
               // [output](http://pasteboard.co/OPuY5FCP6.png)
		batch.end();

	}


```

#### Version of LibGDX and/or relevant dependencies
libgdx 1.9.5

TCreutzenberg:
@NathanSweet @tommyettinger If you guys are not on this already I would have a look at it tomorrow. I wanted to fix this tomorrow actually anyways and didn't even know there was an issue about this already 😄

NathanSweet:
I'm not on it. Best of luck! thar be dragonz

TCreutzenberg:
Hey guys, I need some opinions to actually be able to fix this bug.

GlyphLayout#setText() at the moment does many things in one single loop:
- Adding glyphs
- Checking linebreaks and color-markups
- Truncating + (non-color-markup) line wrapping
- Handling all the special cases for the above

It does this by reusing all the same variables all the time. This makes it incredibly hard to follow and debug. And it also connects all the above tasks directly to another via the shared variables, making it even worse to debug.

[...]

PR #6581, opened 2021-06-26
GlyphLayout, improved glyph run bounds and whitespace wrapping.

NathanSweet:
**Improves glyph run bounds**
The bounds can still overlap due to kerning, but they are no longer off by an entire character. Instead of using a low x and higher xadvance to position the glyph in the right place, we now increase the x position to the start of the glyph and decrease the xadvance. The test was changed to help differentiate bounds from different runs.
Before:

After:

**Fixes bounds when whitespaces wraps**
This text has "T e". When the wrapping occurs so the space wraps, the bounds of the first wrapped glyph can be wrong. This was because the `if (wrapIndex == 0)` code path is taken instead of `wrap()` and it didn't clear `lastGlyph`.
Before:

After:

**Fixes wrapping whitespace before a newline**
This text is "AA BB \nEE". When the wrapping occurs so the space after "BB" wraps, a blank line appears. This was because we moved down a line for the wrapping, then `newLine` was true and moved down an extra line.
Before:

After:

Would be great to get eyes on this before merging. Maybe @TCreutzenberg and @tommyettinger are interested?


NathanSweet:
All the commits in the PR show I clearly don't know what I'm doing (guess I should have squashed before the PR), but the 2 files changed is right. We'll just be sure to squash before merging!

tommyettinger:
It looks like this definitely helps with word wrapping when there's a line break. What if there are two or more line breaks where it wraps? I'd think the initial whitespace would be ignored and later line breaks should be respected, right? The rest of this seems to clean up bounds all over, which may help so we don't need a patchwork of fixes with epsilon and font metrics. I like this whole PR, I'll probably approve in a review.

I always have a ton of commits backlogged in anything I do too. Not sure how to get Git to avoid that...

tommyettinger (approved):
This looks good, but I can't quite follow the logic when reading it as a git diff. The examples on the initial PR post show really solid improvement, so I'm approving. I do do encourage testing with multiple newlines and/or mixed newlines and lines full of all whitespace for wrapping; there are always some corner cases with text rendering that seem to pop up.

TCreutzenberg:
@NathanSweet Just tried it with my fonts and it works great. Yes, the kerning-overlap can't be avoided.
Code looks good.

If I may make one suggestion, I'd change the text in BitmapFontTest to:

`text = "AA BB    [RED]EE[] T [GREEN]e[] V[YELLOW]a bb[] ([CYAN]5[]FFF)";`

This includes a color-run with a space in it and the color-run which made problems for me 2 days ago. So basically it would cover more edgecases and past problems.

TCreutzenberg:
@NathanSweet I just created a new PR (https://github.com/libgdx/libgdx/pull/6582https://github.com/libgdx/libgdx/pull/6582) before looking at this one, it may lead to merge conflicts with this one, so you may want to include the comments into yours, then I can just close mine?

NathanSweet:
> What if there are two or more line breaks where it wraps?
She works properly in that case. Eg for two newlines you get one newline taken care of by the wrapping, then the other one gives a blank line as normal.

> I can't quite follow the logic when reading it as a git diff.
Indeed, it's hard to follow in an IDE with a debugger!

> Oh, and thank you btw!
Sure, no problem! Glad to have more eyes on this code.

I wish we could have an elegant way of fixing this one:
https://github.com/libgdx/libgdx/issues/4672

PR #6582, opened 2021-06-26
Comments added to lastGlyph in GlyphLayout and BitmapFont

TCreutzenberg:
Added some comments about the reason/use behind "lastGlyph" in GlyphLayout and BitmapFont to make the code easier to understand.


sakib-rahman-bangladesh (commented):
*erfect*

updated: Last glyph of the previous run on the same line, used for kerning between runs.

PR #6585, opened 2021-06-28
 GlyphLayout: Fix crash on empty glyphs

TCreutzenberg:
@NathanSweet @tommyettinger Please check: Without the check `glyphs.isEmpty()` setText() throws a NullPointerException with the text provided in the test.


NathanSweet:
When a run is empty, better to skip it entirely. See: 6c69ac2

PR #6586, opened 2021-06-28
GlpyhLayout: Cleanup

TCreutzenberg:
This PR tries to make GlyphLayout a little less intimidating and to make it easier and to reduce the cognitive load working with it.
(This PR also includes changes from https://github.com/libgdx/libgdx/pull/6585 as there will be merge conflicts else)

@NathanSweet @tommyettinger Please check! Thanks!

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -224,7 +224,7 @@ public class GlyphLayout implements Poolable {
 								run.glyphs.removeRange(0, wrapIndex - 1);
 								run.xAdvances.removeRange(1, wrapIndex);
 							}
-							xAdvances[0] = getRunOffset(run.glyphs, fontData);
+							xAdvances[0] = getLineOffset(run.glyphs, fontData);
 
 							if (previous != null) { // Previous run is now at the end of a line.
 								// Remove trailing whitespace and adjust last glyph.
@@ -414,7 +414,7 @@ public class GlyphLayout implements Poolable {
 			FloatArray xAdvances1 = second.xAdvances; // Starts empty.
 			xAdvances1.addAll(xAdvances2, 0, firstEnd + 1);
 			xAdvances2.removeRange(1, secondStart); // Leave first entry to be overwritten by next line.
-			xAdvances2.items[0] = getRunOffset(glyphs2, fontData);
+			xAdvances2.items[0] = getLineOffset(glyphs2, fontData);
 			first.xAdvances = xAdvances1;
 			second.xAdvances = xAdvances2;
 		} else {
@@ -433,7 +433,7 @@ public class GlyphLayout implements Poolable {
 		return second;
 	}
 
-	private float getRunOffset (Array<Glyph> glyphs, BitmapFontData fontData) {
+	private float getLineOffset (Array<Glyph> glyphs, BitmapFontData fontData) {
 		return -glyphs.first().xoffset * fontData.scaleX - fontData.padLeft;
 	}
 
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -98,10 +98,6 @@ public class GlyphLayout implements Poolable {
 		glyphRunPool.freeAll(runs);
 		runs.clear();
 
-		for (int i = 1, n = colorStack.size; i < n; i++)
-			colorPool.free(colorStack.get(i));
-		colorStack.clear();
-
 		BitmapFontData fontData = font.data;
 		if (start == end) { // Empty string.
 			width = 0;
@@ -109,8 +105,8 @@ public class GlyphLayout implements Poolable {
 			return;
 		}
 
-		boolean wrapOrTruncate = (wrap && targetWidth > fontData.spaceXadvance * 3) // Minimum width for wrapping is 3 * spaceXadvance.
-			|| truncate != null;
+		float minimumTargetWidthForWrapping = fontData.spaceXadvance * 3; // Avoid one line per character, which is very inefficient.
+		boolean wrapOrTruncate = (wrap && targetWidth > minimumTargetWidthForWrapping) || truncate != null;
 
 		Color nextColor = color;
 		boolean markupEnabled = fontData.markupEnabled;
@@ -286,6 +282,13 @@ public class GlyphLayout implements Poolable {
 
 		// Align runs to center or right of targetWidth.
 		alignRuns(targetWidth, halign);
+		
+		// Free colorStack resources 
+		if (markupEnabled) {
+			for (int i = 1, n = colorStack.size; i < n; i++)
+				colorPool.free(colorStack.get(i));
+			colorStack.clear();
+		}
 	}
 
 	private void calculateAndSetWidths (BitmapFontData fontData) {
@@ -431,7 +434,7 @@ public class GlyphLayout implements Poolable {
 	}
 
 	private float getRunOffset (Array<Glyph> glyphs, BitmapFontData fontData) {
-		return glyphs.isEmpty() ? 0 : -glyphs.first().xoffset * fontData.scaleX - fontData.padLeft;
+		return -glyphs.first().xoffset * fontData.scaleX - fontData.padLeft;
 	}
 
 	/** Adjusts the xadvance of the last glyph to use its width instead of xadvance. */
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -165,18 +165,24 @@ public class GlyphLayout implements Poolable {
 						glyphRunPool.free(run);
 						break runEnded;
 					}
-					if (lastGlyph != null) { // Move back the width of the last glyph from the previous run.
-						x -= lastGlyph.fixedWidth ? lastGlyph.xadvance * fontData.scaleX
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -98,24 +98,23 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
 		glyphRunPool.freeAll(runs);
 		runs.clear();
 
+		for (int i = 1, n = colorStack.size; i < n; i++)
+			colorPool.free(colorStack.get(i));
+		colorStack.clear();
+
```
TCreutzenberg, line 104: In case of an empty String the pools would never be cleared (line 109), so I moved it up here.
Also I think the pools need to be freed and cleared even if `markupEnabled = false` as this could have been changed in the meantime or another BitmapFont could have been used.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -98,24 +98,23 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
...
 			height = fontData.capHeight;
 			return;
 		}
 
-		if (truncate != null)
-			wrap = true; // Ensures truncate code runs, doesn't actually cause wrapping.
-		else if (targetWidth <= fontData.spaceXadvance * 3) //
-			wrap = false; // Avoid one line per character, which is very inefficient.
+		boolean wrapOrTruncate = (wrap && targetWidth > fontData.spaceXadvance * 3) // Minimum width for wrapping is 3 * spaceXadvance.
+			|| truncate != null;
```
TCreutzenberg, line 113: In `setText` there are so many variables which keep chaning and changing that I thought it would be good to use more variables which are final.
I didn't use the final keyword here because it is not used elsewhere in this code. Would it be ok to add more "final"s?

Besides, this reduces the linecount by 2 and makes the code a little easier to read.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -140,7 +139,7 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
 				case '[':
 					// Possible color tag.
 					if (markupEnabled) {
-						int length = parseColorMarkup(str, start, end, colorPool);
+						int length = parseColorMarkup(str, start, end);
```
TCreutzenberg, line 142: `colorPool` is static and does not need to be in the parameters for this function.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -165,7 +164,7 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
 						glyphRunPool.free(run);
 						break runEnded;
 					}
-					int n = run.xAdvances.size;
+					int xAdvancesSize = run.xAdvances.size;
```
TCreutzenberg, line 167: xAdvancesSize is easier to understand than having to remember what `n` is supposed to be all the time

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -183,19 +182,19 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
...
 						break runEnded;
 					}
 
 					// Wrap or truncate.
 					x += xAdvances[0]; // X offset relative to the drawing position.
-					for (int i = 1; i < n; i++) {
+					for (int i = 1; i < xAdvancesSize; i++) {
 						Glyph glyph = run.glyphs.get(i - 1);
-						float glyphWidth = (glyph.width + glyph.xoffset) * fontData.scaleX - fontData.padRight;
+						float glyphWidth = getGlyphWidth(glyph, fontData);
```
TCreutzenberg, line 197: The same code of `getGlyphWidth` in several places in the code.
Having it as a private mthod reduces errors and makes it easier to read imho.
Besides, I can be sure that this line does not change any variables.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -204,7 +203,7 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
 
 						if (truncate != null) {
 							// Truncate.
-							truncate(fontData, run, targetWidth, truncate, i, glyphRunPool);
+							truncate(fontData, run, targetWidth, truncate);
```
TCreutzenberg, line 206: `glyphRunPool` is static and `i` is not used in the rpivate method `truncate`

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -238,7 +237,7 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
 								adjustLastGlyph(fontData, previous);
 							}
 						} else {
-							next = wrap(fontData, run, wrapIndex, i);
+							next = wrap(fontData, run, wrapIndex);
```
TCreutzenberg, line 240: `i` is not used in the private method `wrap`

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -276,6 +275,13 @@ else if (targetWidth <= fontData.spaceXadvance * 3) //
 		height = fontData.capHeight + Math.abs(y);
 
 		// Calculate run widths and the entire layout width.
+		calculateAndSetWidths(fontData);
+
+		// Align runs to center or right of targetWidth.
+		alignRuns(targetWidth, halign);
```
TCreutzenberg, line 281: Having `calculateAndSetWidths` and `alignRuns` as private methods reduces the length of `setText` quite a bit.
Also, I was never sure if this code also changes any vaiables needed for the loop above, even if the blocks were outside the loop. It was just very hard to keep track of this monster function.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
NathanSweet: Might make sense to clear it at the end of a run, rather than at the start of the next one.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
NathanSweet: The old code seems fine to me here. The new comments are maybe not as clear.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
NathanSweet: These unused parameters are good to clean up. Leftover from when it tried to avoid looping through all the runs at the end.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -416,15 +423,23 @@ private GlyphRun wrap (BitmapFontData fontData, GlyphRun first, int wrapIndex, i
 		return second;
 	}
 
+	private float getRunOffset (Array<Glyph> glyphs, BitmapFontData fontData) {
+		return glyphs.isEmpty() ? 0 : -glyphs.first().xoffset * fontData.scaleX - fontData.padLeft;
```
NathanSweet, line 427: Shouldn't need `isEmpty`, as we should never have empty runs. See: #6585
The method is OK, since we do it in two places. `getLineOffset` may be a better name, as it's the offset for the first glyph of a line.

NathanSweet:
I made a few comments. Overall the improvements are good!

tests/gdx-tests/src/com/badlogic/gdx/tests/BitmapFontTest.java
```diff
@@ -115,7 +115,7 @@ public void render () {
 			text = "Another font wrap is-sue,  this time with multiple whitespace characters.";
 			// text = "test with AGWlWi AGWlWi issue";
 			text = "AA BB \nEE"; // When wrapping after BB, there should not be a blank line before EE.
```
tommyettinger, line 117: The first assignment to `text` is ignored because it is reassigned on the next line.

tommyettinger (approved):
My only nitpicks are minor (the redundant text assignment in the test, mostly, and maybe clarification on the first change to comments). I'd be fine if this was merged as-is, though it would be nice if those could be addressed.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -105,17 +105,12 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
 			return;
 		}
 
-		if (truncate != null)
-			wrap = true; // Ensures truncate code runs, doesn't actually cause wrapping.
-		else if (targetWidth <= fontData.spaceXadvance * 3) //
-			wrap = false; // Avoid one line per character, which is very inefficient.
+		float minimumTargetWidthForWrapping = fontData.spaceXadvance * 3; // Avoid one line per character, which is very inefficient.
+		boolean wrapOrTruncate = (wrap && targetWidth > minimumTargetWidthForWrapping) || truncate != null;
```
TCreutzenberg, line 109: @NathanSweet @tommyettinger I'm not really sure what to do, please change at will

NathanSweet:
I made a few minor changes but I think it's good! The class is still a beast, but it's nicer now. Cheers!

sakib-rahman-bangladesh (commented):
great work! erfect

PR #6587, opened 2021-07-01
GlyphLayout: Fix color-markup line wrapping

TCreutzenberg:
Here is the fix for https://github.com/libgdx/libgdx/issues/4672.

This is a working prototype. I know that some things can be optimised in a second step.

I don't know if this is a lot slower or faster than before, it doesn't really look bad. @NathanSweet please confirm.

The change that may upset people is that GlyphRun's `color` was replaced by `colors`, one rgba888 int per glyph.
Color-marup does not "break up" the GlyphRuns any more, there is exactly one GlyphRun per line now. Color changes are retrieved from `colors`.
If this is really bad and people start calling me a witch and start chasing me with pitchforks and torches, etc. this behaviour can be easily changed in the follwing way:
GlyphRun will get back the `color` field, each line's GlyphRun can be split up by color into several GlyphRuns again (at a little extra expense, to stay in line with the old code structure) at the end of `setText()`. So everything is as before, just that color-markup line wrapping is working (and at least I think that the code is easier to follow and debug with one GlyphRun per line).

fix

Other people who are most likely interested in this PR: @tommyettinger


yuripourre:
@TCreutzenberg I am not familiar with scene2d nor GlyphLayout but having one color per character seems too overkill.
Why not having color per interval? I believe the default use case is to have color for consecutive characters, right?

tommyettinger:
Kind-of agreeing with Yuri; as I understood it the concept of splitting text into runs was to draw all text with one color as a single unit, thus only needing to store the color once per run. That said, I'm only kind-of in agreement -- a color is pretty small, especially if stored as a packed float or int. A 1080p screen completely filled with text (let's say it's an 80s retro terminal, so using a tiny 8x8 font) would only use about an eighth of one megabyte to store a four-byte color per char. What's more, if using color per-char allows reducing the number of run objects allocated, which this seems to do, it might even use less memory in the case where color changes often. It is, however, much more memory (relatively) when used for text with all one color; it goes from two bytes per char to six when the chars are stored in a char array and the colors are stored in a float array (or int array, depending on how the color is expected). This looks like a similar solution to the one I made in [Textramode](https://github.com/tommyettinger/textramode), which effectively uses one `long` per character to store its codepoint (potentially going above 0xFFFF to at most 0xFFFFFF to permit emojis, though that isn't implemented yet), its color, and 8 bits for special effects like bold/italic/underline/strikethrough/superscript/subscript. I haven't touched Textramode in a little while because most usage in a regular libGDX app would need it to integrate with scene2d.ui or a parallel version of scene2d.ui that used this alternate font type. I did find that word wrapping was pretty easy with the approach that stores a color per character, so I can see why that would be a good option here.

So, I wouldn't worry about a fraction of a megabyte of memory usage (especially if it reduces the number of objects needed), but I am somewhat concerned about compatibility with other parts of the library, especially less-commonly-used ones. The pitchfork-option of using one color per char during the parsing step and then re-forming runs from that data might be better for compatibility, but I'd avoid that unless something needs it. I haven't had a chance to look through this code yet, but I am guessing parts will look similar to how I did it in Textramode; I also used one object per line instead of one per run.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -122,6 +125,7 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
 			if (start == end) { // End of text.
 				if (runStart == end) break; // No run to process, we're done.
 				runEnd = end; // Process the final run.
+				isLastRun = true;
```
NathanSweet, line 128: `isLastRun` seems identical to `runEnd == end`.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -150,129 +154,98 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
...
-						xAdvances[0] = 0;
-					}
-					run.x = x;
-					run.y = y;
-					runs.add(run);
 
-					if (newline || runEnd == end) {
-						adjustLastGlyph(fontData, run);
+					if (newline || isLastRun) {
+						adjustXadvanceOfLastGlyph(fontData, newRun);
```
NathanSweet, line 171: The descriptiveness of this private method seems a bit excessive. :)

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -529,12 +514,23 @@ public String toString () {
...
+				xAdvances.add(run.xAdvances.get(i));
+			}
+			colors.addAll(run.colors);
+		}
+		
 		public void reset () {
 			glyphs.clear();
 			xAdvances.clear();
+			x = 0;
+			y = 0;
```
NathanSweet, line 531: `x` and `y` don't need to be reset for pooling because they don't affect GC. That is assuming they are set by the code that calls `obtain`.

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
```diff
@@ -121,15 +121,22 @@ public void tint (Color tint) {
 			for (int ii = 0, nn = layout.runs.size; ii < nn; ii++) {
 				GlyphRun run = layout.runs.get(ii);
 				Array<Glyph> glyphs = run.glyphs;
-				float colorFloat = tempColor.set(run.color).mul(tint).toFloatBits();
+				final IntArray rgba8888Colors = run.colors;
+				int lastRgba8888Color = Integer.MAX_VALUE;
```
NathanSweet, line 125: I'd say it's safe to just say "color" for these locals.

NathanSweet:
My code comments are mostly pedantic stuff.

Overall it's an interesting idea! I do think it's a bit unfortunate to store a color per glyph, mainly because I expect most text to be a single color. Unless you are doing rainbow text, I'd say it's most common for text to be 1 color. Text that does use color tags is probably most often using it to highlight a portion, so would have 2 or maybe 3 colors. More than that happens, but is relatively rare in most apps.

Maybe we can still have an IntArray `colors`, but only store when the color changes? Entries could be pairs: "glyph index" and "color". The first entry could be omitted, since we know it's zero.

tommyettinger:
> Maybe we can still have an IntArray colors, but only store when the color changes? Entries could be pairs: "glyph index" and "color". The first entry could be omitted, since we know it's zero.

I like this a lot! It is like the color-per-run approach, but should interact more cleanly with line breaks, if I understand it right. An open question might be how the previous color from an earlier line should carry over into the next line (I think it should, if it does now). This acts as an optimization for the most common case of solid blocks of single-color text, with minimal cost for the much-less-common cases of rainbow text.

NathanSweet:
Aye, it's like basically the PR with run-length encoding. Likely the RLE can be built as we go, rather than building it up with one color per glyph, then compressing.

Since we store the colors on each run, I wouldn't think the color of a previous run would affect subsequent runs. Though that leads to: what if we store the IntArray `colors` on the GlyphLayout? The glyph indices used in that `colors` array would be across all glyph runs. If you cared to find out what color a particular glyph is, you'd have to start counting glyphs from the first glyph run, but that's not a problem for BitmapFontCache drawing, who always goes through all the glyphs.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
TCreutzenberg: Yes, it just is much easier to identify `isLastRun` than  `runEnd = end`. We can remove that again if you want.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
TCreutzenberg: During refactoring I  was often going back to `adjustLastGlyph` to see what exactly is changed so I renamed it to `adjustXadvanceOfLastGlyph` so I knew exactly what it does and what it changes. We can revert the naming of course.

TCreutzenberg:
Hey guys, actually I already tried the only-store-color-changes idea during refactroing, but I was not able to get it running quickly (wrapping and dropping whitespace runs ruined the easy approach) so I thought I'll leave it for the moment.

I stopped at this point with this prototype because I expected heavy resistance against `color` being dropped and didn't want to invest a lot more time into that if you say it's a no-go while trying to burn me alive.
As this is not the case, I'll try it right now.

Also, the idea with the color changes stored in the GlyphLayout instead of the GlyphRun seems great, but it will make it really hard to do other stuff with it. I'll try stroing it in the GlyphRun for the first try.

TCreutzenberg:
So, it was a lot harder than thought (thanks wrapping 😝), the GlyphRun now stores "color-intervals".

TCreutzenberg:
@NathanSweet Just out of curiosity, did you test the difference in your startup time? I'd really like to know the differences if this is called a few thousand times!

sakib-rahman-bangladesh (commented):
updated logic and value

sakib-rahman-bangladesh (commented):
updated api

TCreutzenberg:
(Sorry for all the extra commits, I just keep finding stuff... but that's it for today)

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -413,10 +377,61 @@ private GlyphRun wrap (BitmapFontData fontData, GlyphRun first, int wrapIndex) {
...
+				if (droppedGlyphCount > 0)
+					for (int i = first.colorChangeIndices.size - 1; i > 0; i--) { // i > 0 because first value is never adjusted
+						final int colorChangeIndex = first.colorChangeIndices.get(i);
+						if (colorChangeIndex > firstGlyphCount)
+							first.colorChangeIndices.set(i, colorChangeIndex - droppedGlyphCount);
+						else
+							break;
+					}
+
+				for (int i = 0, n = first.colorChangeIndices.size; i < n; i++) { // i = 1 because first color is always added
```
NathanSweet, line 399: This comment seems confusing. It's either obsolete and should be deleted, or the code should actually use 1.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -149,130 +153,95 @@ public void setText (BitmapFont font, CharSequence str, int start, int end, Colo
...
-						break runEnded;
-					}
-					int xAdvancesCount = run.xAdvances.size;
-					float[] xAdvances = run.xAdvances.items;
-					GlyphRun previous = null;
-					if (lastGlyph != null) { // Move back the width of the last glyph from the previous run.
-						previous = runs.peek();
-						x += xAdvances[0] - previous.xAdvances.peek();
-						xAdvances[0] = 0;
+				if (true) { // Can occur eg when a color tag is at text start or a line is "\n".
```
NathanSweet, line 156: Was this change accidental?

NathanSweet: Nevermind, seems the `if (true)` is right and the comment is obsolete.

NathanSweet:
> Just out of curiosity, did you test the difference in your startup time? I'd really like to know the differences if this is called a few thousand times!

Wrapping my UI creation with System `nanoTime`, I see 100ms deviation from run to run, without any code changes (~700ms total time). That unfortunately makes it pretty useless as a benchmark. FWIW, it doesn't feel slower with your new code, on my particular hardware. We'd need to setup JMH if we wanted a better idea of how changes are affecting performance.

I made a few commits. I tried to commit each time I made a decent sized change, so it's easier to see what was changed rather than one big diff.

I think the current state of the PR is great! It seems like it was the right way to solve not wrapping colors, kudos! Also the code is easier to read and doesn't appear to be doing significantly more work than before.

To improve it from here it would be cool to see:
1) The colors and indices moved to GlyphLayout. I like this because it's fewer objects and means we don't need a `0` glyph index for each run. The downside is it may be a little harder to associated a run's glyph with its color. However, 99% of the time (possibly 100%) it's only BitmapFontCache who cares, and that iterates the glyphs from the start, so knowing each glyph's color is easy.
2) The colors and indices could be a single IntArray. They are read together and are both ints, so its efficient to store them together in the same array.

BTW, add your @author tag!

tommyettinger:
Thanks for doing this, @TCreutzenberg and @NathanSweet ! I know making big PRs is tough, but this seems like a very worthwhile improvement both in terms of removing bugs and code quality.

@sakib-rahman-bangladesh , you are acting suspiciously like a bot and we will block you if you continue to review PRs without any meaningful contribution. If you are not a bot, you can either:
 - Stop reviewing PRs and stop making irrelevant comments, in which case we have no reason to block you, or
 - Verify that you are, in fact, human -- you're walking in the desert and you see a tortoise, flipped over on its back.  It can't get up, but you aren't doing anything to help.  What am I referencing? Hint: humans can use search engines, but bots cannot.
   - If you pass this test, please explain why you made the previous content-less reviews and comments.

NathanSweet:
Oops, I already blocked him for 30 days early today. His comments didn't make sense over and over, ain't nobody got time for that. He can come back if he's human, but I'm sorry I ruined your tortoise thing!

tommyettinger:
Well it's a Blade Runner reference, next time I'll try to think of another uncommon Turing test.

I'm doing some font work myself (unrelated) and it is always surprising to me, every time, how involved text rendering is.

TCreutzenberg:
Cool! Thanks @NathanSweet for the great additional commits and @tommyettinger for the kind words.
I just fixed kerning if the runs the append are separated by a color tag.

@NathanSweet So, should I go on with the improvements in this PR or in another one? And I won't get much done until Monday, weekend is family time.

TCreutzenberg:
Found some spare time, colors are now in GlyphLayout as one IntArray.

NathanSweet:
Cool! I'm away from computers until Tuesday, so I can't look at it properly until then, though I'm sure it's great!

TCreutzenberg:
I'm not 100% sure if my new or updated comments can be understood...

NathanSweet:
I finished a pass. @TCreutzenberg please check it out, I think it's great and ready to merge!

tommyettinger:
Wow, that's quite a big pass; thanks for catching all that. The changes look good to me. Fixing crashes is great. I usually prefer ABGR8888 in general, so that's good by me. Lots of corner cases seem to be handled correctly now, that's good for sure. I'm supporting a merge when @TCreutzenberg is ready.

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -94,201 +112,164 @@ public void setText (BitmapFont font, CharSequence str, Color color, float targe
...
+					lastGlyph = null;
+				} else
+					lastGlyph = lineRun.glyphs.peek();
+
+				if (!wrapOrTruncate || lineRun.glyphs.size == 0) break runEnded; // No wrap or truncate, or no glyphs.
+
+				if (newline || isLastRun) {
+					// Wrap or truncate. First xadvance is the first glyph's X offset relative to the drawing position.
+					float runWidth = lineRun.xAdvances.first() + lineRun.xAdvances.get(1); // At least the first glyph will fit.
+					for (int i = 2; i < lineRun.xAdvances.size; i++) {
```
TCreutzenberg, line 211: ~~@Nathan Sweet: As there was a change to `i = 2` here, shouldn't it also be `i = 2;` in line 242?
Testing look great with `i = 2;` in line 242 but I guess just because I don't have a font with a really wide single charater (> 3 * spaceWidth)... or do I miss something?~~

@NathanSweet Nevermind the above comment, `i = 1;` yields the correct result while wrapping in the BitmapFontTester (`i = 2;` does not, it wraps "too late" for some lines), although I don't 100% understand why.

TCreutzenberg:
I just added a new commit which fixes truncating/wrapping if there is a color tag as the last element of the input string, e.g.:
"abcd[BLUE]"

TCreutzenberg:
@NathanSweet @tommyettinger PR looks good to me!

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
NathanSweet: It's because right after the `i = 1;` the loop repeats, so `i` is incremented and is 2 for the next go around.

NathanSweet:
Great improvements, cheers!

Surely there are no edge cases we missed. :D

PR #6591, opened 2021-07-07
Make capitalization of "libGDX" consistent with the branding guidelines by replacing some instances of "LibGDX", "Libgdx" and "libgdx", plus some other very minor inconsistencies

Frosty-J:
> The name “libGDX” always starts with a lower-case `l`, even at the beginning of a sentence. `GDX` is always written in upper-case letters.

https://libgdx.com/brand/

If we're going to go with this, let's at the very least have the README follow it.

Most of these changes are in comments and the likes, so hopefully nothing will break.

This is not a Find & Replace job - I left some things alone, such as historical changelogs and someone's monologue in `AndroidLiveWallpaper.java`.

I agree to the contributor licence, yadda yadda - just forge my signature.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -77,7 +77,7 @@ If you want to poll the brains of other devs, either send a pull request and sta
 
 ### Contributor License Agreement
 
-Libgdx is licensed under the [Apache 2.0 license](http://en.wikipedia.org/wiki/Apache_License). Before we can accept code contributions, we need you to sign our [contributor license agreement](https://github.com/libgdx/libgdx/blob/master/CLA.txt). Just print it out, fill in the blanks and send a copy to [`contact@badlogicgames.com`](mailto:contact@badlogicgames.com?subject=[LibGDX]%20CLA), with the subject `[Libgdx] CLA`.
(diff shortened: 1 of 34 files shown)


.github/CONTRIBUTING.md
```diff
@@ -77,7 +77,7 @@ If you want to poll the brains of other devs, either send a pull request and sta
 
 ### Contributor License Agreement
 
-Libgdx is licensed under the [Apache 2.0 license](http://en.wikipedia.org/wiki/Apache_License). Before we can accept code contributions, we need you to sign our [contributor license agreement](https://github.com/libgdx/libgdx/blob/master/CLA.txt). Just print it out, fill in the blanks and send a copy to [`contact@badlogicgames.com`](mailto:contact@badlogicgames.com?subject=[LibGDX]%20CLA), with the subject `[Libgdx] CLA`.
+libGDX is licensed under the [Apache 2.0 license](http://en.wikipedia.org/wiki/Apache_License). Before we can accept code contributions, we need you to sign our [contributor license agreement](https://github.com/libgdx/libgdx/blob/master/CLA.txt). Just print it out, fill in the blanks and send a copy to [`contact@badlogicgames.com`](mailto:contact@badlogicgames.com?subject=[libGDX]%20CLA), with the subject `[libGDX] CLA`.
```
tommyettinger, line 80: We might need to verify that the CLA email filter is case-insensitive, otherwise using a different subject line might inadvertently bypass the inbox.

tommyettinger (changes requested):
I just think we need to verify that the CLA email process is either checked by a person or by a case-insensitive filter before we change the email procedure. Other than that, everything looks fine.

00-Evan:
I can confirm that the inbox seems to work with the title "[libGDX] CLA", I just received a reply email from Mario for my recent CLA submission.

crykn (approved):

tommyettinger (approved):

Issue #5251, opened 2018-06-05
Metal port since Apple is deprecating OpenGL


riho:
#### Issue details
Apple announced that they're pulling OpenGL support ([link](https://developer.apple.com/macos/whats-new/) and [link](https://developer.apple.com/ios/whats-new/)). What would that mean for Mac and iOS support in LibGDX going forward? Since Vulkan support is off the table (#3344), would Metal support be unfeasible for the same reasons?

- [ ] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [x] MacOS

badlogic:
We don't know when Apple is actually removing OpenGL. Metal is not feasible
for the same reasons as Vulkan: all of libGDX is based on OpenGL. We'd have
to essentially rewrite large parts of it. If someone wants to try that in a
fork, I'll be happy to check it out. Could be build on top of Sokol or bgfx.

barkholt:
Could we, when necessary, rely on a software implementation of OpenGL on top of Metal or Vulkan?

Something like what these projects have attempted?
https://github.com/doitsujin/dxvk (implements DirectX using Vulkan).
https://moltengl.com/moltengl/
https://github.com/g-truc/glo

badlogic:
We could, and especially for our 2D use cases something like MoltenGL would
do just fine. Another alternative would be ANGLE, which I'm sure will have
Metal support soonish. If not, then ANGLE serves as a good canary for when
Google thinks it's time to switch from OpenGL to Metal to implement WebGL.

julienvillegas:
@noblemaster, in regards to your point about Molten and converting shaders : "_(2) convert shader code to Metal. Mapping the function calls is probably the easy part. The shader code is harder to do but not impossible._"

According to MoltenGL product factsheet:
"_MoltenGL will automatically convert your GLSL shaders to their MSL equivalents. This can be done transparently at run time, using the Runtime Shader Conversion feature of MoltenGL, or at development time using the offline MoltenGLShaderConverter development tool, provided with MoltenGL._"

So that´s quite good for us..  That said Molten is not free.. $149 per user. so I am not sure it would even be an option-

noblemaster:
Yes, MoltenGL is definitely an interesting option. Given it costs $99/year to release on iOS, I think the 1x $149.- isn't that much of a problem. I am just wondering how easy it will be integrate into the libGDX iOS backend?

[...]

PR #6593, opened 2021-07-10
Add back MOE 2.0 as backend with MetalKit instead of OpenGL.

Berstanio:
As I understand the discussion in this pr (https://github.com/libgdx/libgdx/pull/6016) MOE was removed for three reasons:
1. It was in many aspects behind robovm
2. Nobody was maintaining the MOE backend.
3. MOE itself wasn't really maintained. 

I think these things have changed:
1/3. Moe got a new backend. Art was replaced by substratevm/ graalvm's native-image. With this new backend MOE supports Java up to Java 11 (and theoretically Java 16). Also MOE is maintained/bugs get fixed. And now, with the new backend, MetalANGLE is working with libgdx and MOE (art seems to struggle with that https://github.com/libgdx/libgdx/issues/5251)
2. I would volunteer to maintain the MOE backend on libgdx.

Now the question is, is that enough to bring back MOE support to libgdx?

Notes for building:
- The iosgl c++ files needs to be compiled and copied over to the "ios/xocde/native" dir in the MOE project App.
- The MetalANGLE framework needs to be build https://github.com/kakashidinho/metalangle. Instructions are: (**Quick build with Xcode project/ios**): https://github.com/kakashidinho/metalangle/blob/master/src/libANGLE/renderer/metal/DevSetup.md. 
The framework also needs to be embeded.
- Java GraalVM 11 is needed.

Current problems:
-  Since I edited the cpp files directly (see commit 48dbabf35dcd83da2b0b1e3b6414a72829b4b898 ) and they are the same for robovm and MOE, robovm won't work any more. The solution is that robovm and MOE needs different natives, but I wasn't able to achieve this. My attempt was to create in the MOE-backend a "jni" folder and place the cpp files for MOE there and then configure jnigen (as I see all the natives are generated with jnigen). The problem with that was, that jnigen also tried to generate files from the native methods in MOE, which it shouldn't do, and failed. I wasn't able to find a way to exclude the java files from jnigen. If someone has an idea of how to make seperate native files for moe and robovm I would be glad to hear!
- MetalANGLE currently needs to be compiled by the user. It would be nice to ship a MetalANGLE version directly with the natives.
- The setup template for MOE is currently based on OpenGL and not MetalANGLE. 

I hope everything I wrote was clear (I'm not a native english speaker.). If there are any questions, objections or suggestion I would love to answer/hear. (I'm sure there are at least some.)


tommyettinger:
I like the idea here, but this is still definitely at a draft stage. First off, we need to make sure no existing projects break because MOE was re-added, and that means we would need different sets of natives for MOE and for RoboVM instead of editing the existing ones. Is there a reason why metalangle needs to be compiled by the user? Would the latest release, such as https://github.com/kakashidinho/metalangle/releases/tag/gles3-0.0.6 work instead? Optimally, we wouldn't need any C/C++ builds by users to get this up and running, or if some is needed, it should be automated by build tools.

You said the setup template for MOE uses OpenGL and not Metal -- is there any code difference? There probably shouldn't be, at least in `core`... If any differences are only in the `moe` module, then it's fine, but if code needs to be changed in `core` to use Metal, that seems like a problem.

Using Substrate for iOS seems like a much better option than the current solution used by RoboVM with an Android core, since having Java 11 puts MOE ahead in version, past RoboVM and roughly tied with GWT 2.9.0. RoboVM doesn't really have full Java 8 support yet, although I understand dkimitsa was working on it. Using Metal should also future-proof iOS support a little more than using OpenGL.

Thank you for offering to maintain the MOE backend; we'd certainly need someone to handle any problems that arise with iOS over time (and it tends to change so much over the years...). You might need to be willing to train some other developers regarding any quirks of the MOE code, so that questions could be answered and fixes delivered by more than one person.

I hope there can be some constructive discussion about MOE now that it seems to have advanced quite a bit. We still need to support RoboVM, though, and that would probably mean having two backends again, at least for a while.

Tom-Ski:
Aside from the build problems that would need fixing, (as none of that is gonna get merged whilst there would be so many hoops for a dev to jump through to use it), its welcome to have multiple solutions as long as conditions are met.  Those including some stronger evidence that the project isn't going to fade again, or just become a burden.

It would be good to get a list of all restrictions and limitations that exist, what IDEs and versions of the most common plugins we use are compatible.

It looks like Natj hasn't changed very much at all, so the backend maintenance won't really be any different than it was before. 

Regarding natives, jnigen shouldn't have to change, but we would probably need to add a variant option probably to be able to build the same natives with slightly different inputs, which would save us a lot of work since they are going to be identical. A not so nice way would be to add a new platform to jnigen. @PokeMMO what do you think?

I don't think metal bindings belong in this repo either. We shouldn't be taking on the maintenance of the binding, and also building the frameworks for our backend which is 100% gl based.  I'm not sure we also want to be forcing the use of angle as this PR is doing, definitely should be a choice, in which case we'll probably just need to double up the gl implementations for both gl backend types.

PokeMMO:
Sorry, I don't quite understand. If they're identical why would we have to do anything? You could just have each project refer to the ios natives normally?

MrStahlfelge:
While I also think it is good to have more options, I don't like the idea of having two iOS backends on the main repo again. There was some confusion for developers which one to chose. We haven't two HTML backends for a reason, also there is at least a TeaVM backend out there, too.

Best is to develop the backend in an own repo. When/if it will be mature enough and has some practical advantages over RoboVM, we can move the repo into the libGDX organization. Time to get it into the main repo is when it is better than the current RoboVM backend, (and that's probably a long way to go).

Tom-Ski:
> Sorry, I don't quite understand. If they're identical why would we have to do anything? You could just have each project refer to the ios natives normally?

Right had a little closer look, and I thought they would be different. So in this case it wouldn't be needed. But perhaps there might be some other? 

We can just have the option to have metalangle in robovm too as an option, with the increase in side the only downside. Then we'd have identical natives there for the backend.

Berstanio:
> Is there a reason why metalangle needs to be compiled by the user? Would the latest release, such as https://github.com/kakashidinho/metalangle/releases/tag/gles3-0.0.6 work instead?

No, just it isn't tested at the moment with the prebuild framework. Later my idea would be to ship a build MetalANGLE directly with the other libgdx natives.

> You said the setup template for MOE uses OpenGL and not Metal -- is there any code difference?

No, it is just that the framework configuration in the template project.pbxproj is outdated and does not include MetalANGLE.

> What IDEs and versions of the most common plugins we use are compatible

Currently the MOE Plugin just works reliably with Android Studio/IntelliJ. But the plugin isn't really needed for libgdx-projects. The whole build/launch process can be done with gradle. Xcode works also fine for this. 
I don't know what you exactly mean by "plugins we use" but I think it is something like gdx-pay etc.?
If that is right, here is what I already tested.
Working are: gdx-box2d, gdx-bullet (some problems, explained later), gdx-freetype, gdx-ai (all the newest ones)
I implemented gdx-controllers with moe as backend but it isn't tested yet.
Not tested: gdx-pay, gdx-video etc.

>   Any critical pitfalls in the VM

Yes they are some:
1. Finalizers are not supported and they wont be. This could lead to some issues with some projects when they heavily rely on finalizers. e.g. gdx-bullet has finalizer implementations which does clean up if the gc deletes a bullet java object. But as I tried a bit with a desktop game using bullet the finalizer was never called. But still that is a problem that potentially could lead to some memory leaks/bugs.
2. It is not directly a pitfall but a difficulty when using MOE 2.0. native-image needs some configuration on what classes/fields etc. needs to be reflectible/accesible by jni etc. Most of that native-image figures out with heavy static analyzes and for the reflection/jni done in libgdx I already added some configuration (see commits d7f8fa7 and  e176e0a) but the problem here is the json parsing. Since the json api uses reflection for recreating the saved objects nearly any field and class could need to be reflectable (as I understand how Models etc. are loaded). I'm currently thinking about some solutions, maybe a tool that analyzes the assets and creates the configs accordingly. And for all the other cases where the user needs reflection etc. and native-image can't identify it it is pretty easy to create the configs based on the exception that are thrown. It is also possible with graalvm's native-image-agent but that is a bit more complicated.
3. Currently native-image with llvm as backend has a few problems with compile time and app size. But I'm sure it will be fixed. It is covered by: https://github.com/oracle/ graal/issues/3017

>    Any features not supported yet, what ios versions are currently supported
>      etc

The newest iOS versions are supported. 
I'm not sure whether this are all downsides, but if I find/remember more I will append this list.

> I don't think metal bindings belong in this repo either. We shouldn't be taking on the maintenance of the binding, and also building the frameworks for our backend which is 100% gl based. I'm not sure we also want to be forcing the use of angle as this PR is doing, definitely should be a choice, in which case we'll probably just need to double up the gl implementations for both gl backend types.

What reasons are against using MetalANGLE for translating the OpenGL calls to Metal calls as the basic OpenGL MOE backend? Metal is faster and more future save than the deprecated OpenGLES. Also MetalANGLE seems pretty done, at least OpenGL 2.0 but now also OpenGL 3.0. But maybe I'm just a bit unexperienced in serving such a broad width of use cases and maybe there is always something that breaks on such changes I don't know.  
But if MetalANGLE should be used I think the bindings are necessary. They are replacing the GLKit-Api. But I would agree that "MetalANGLE.java" should be deleted.

> We can just have the option to have metalangle in robovm too as an option, with the increase in side the only downside. Then we'd have identical natives there for the backend.

As I wrote there are some complications with robovm and MetalANGLE. I'm not so experienced with robovm so another more experienced person would need to look at this.

I hope I answered all questions/commented the necessary stuff!

obigu:
I think it's good news MOE is not dead and to see some interesting developments are being made and having an alternative option on iOS alleviates possible concerns about libGDX iOS future-proofness.

> As I wrote there are some complications with robovm and MetalANGLE. I'm not so experienced with robovm so another more experienced person would need to look at this.
> 

Just a clarification on this point. As far as I know, the issues @kakashidinho encountered with libGDX and RoboVM were related to setting up his dev environment and unfortunately he didn't had much time to spend on it at the time so progress stopped then, we didn't even get the chance to work on the bindings. 

I think nobody has tried to integrate MetalAngle to libGDX and MobiVM so far. I'm with Tomski that if we do integrate Angle, it makes sense it is shared between iOS backends and that using it should be an option (unless Apple drops openGL support obviously). AFAIK MetalAngle is still in an early stage and will need some ironing out and testing with actual games until it becomes stable. Does any published app/game use it already?

Tom-Ski:
Yes, that is my point. I'm not against using MetalAngle, and we can have that option in both ios backends. That should be separated from bringing MOE back as a backend as they are not a package deal and make this less likely to happen. We can treat them separately, and think abut a clean way to test and integrate metalangle into mobivm first and potentially the additional backend if we think it makes sense to bring it back to life also.


> What reasons are against using MetalANGLE for translating the OpenGL calls to Metal calls as the basic OpenGL...
 
I'm not against this, I'm saying that having the full Metal binding is not necessary considering that we are using it as a shim for gles translation.

Berstanio:
> Yes, that is my point. I'm not against using MetalAngle, and we can have that option in both ios backends. That should be separated from bringing MOE back as a backend as they are not a package deal and make this less likely to happen. We can treat them separately, and think abut a clean way to test and integrate metalangle into mobivm first and potentially the additional backend if we think it makes sense to bring it back to life also.

Ah okay I see I'm sorry I misinterpreted your message. Now I understand your point. I would agree with you.

kakashidinho:
> Just a clarification on this point. As far as I know, the issues @kakashidinho encountered with libGDX and RoboVM were related to setting up his dev environment and unfortunately he didn't had much time to spend on it at the time so progress stopped then. We didn't even get the chance to work on the bindings.
> 
Sorry about that, at that time asides from dev environment difficulties, porting `MetalANGLE` to `robovm` required me to study how the latter's java bindings work as well so I put it on hold. It's not `robovm`'s problems, it's just me not having much knowledge in `robovm` and its java dev environments in general (don't mind me, I don't use java that much).

>AFAIK MetalAngle is still in an early stage and will need some ironing out and testing with actual games until it becomes stable. Does any published app/game use it already?
>
Might not be relevant, but I know at least one commercial software is using MetalANGLE right now:
https://apps.apple.com/us/app/nomad-sculpt/id1519508653#?platform=ipad
It is confirmed by the app's author: https://github.com/kakashidinho/metalangle/issues/25#issuecomment-791556693

Berstanio:
So what are the next steps here for me? 
That is what I think: 
I would exclude MetalANGLE from this pr and put it in its own repo where it can be included optionaly. I would work further on the MOE backend on libgdx and try to test it against more actual games. Is there any special place where I could find Open Source libGDX games that aren't outdated since 5 years? On GitHub it was a bit hard to find something complex that was working.
I would keep the pr open and comment a bit about progress that has been made over the time or should I close it?

Any objections or other ideas?

I want to thank everyone for their detailed comments and feedback!

tommyettinger:
Recent libGDX Jam games are probably the best place to find games written that use a recent libGDX version, and most (all?) are open source. https://itch.io/jam/libgdx-jam-17/entries has the most recent entries. However, it's doubtful that any have iOS support built-in, and very few probably provide Android APKs, though some might be possible to adapt so they support mobile.

Having MetalANGLE in its own repo would be great. I imagine if Apple does remove OpenGL from a future MacOS version, then that would mostly affect AArch64 Macs, and I think those can already run iOS apps sorta-natively. In that case, being able to support Metal for rendering (even if only meant originally for iOS) would help keep compatibility with MacOS.

MrStahlfelge:
> So what are the next steps here for me?

Considering following my suggestion using an own repo for the backend and publish it so that everyone can use own games to test with. Not every game is open source.

If you want to test yourself, have a go here, its iOS is up to date and working with latest RoboVM: https://github.com/MrStahlfelge/lightblocks

Berstanio:
So I've seperated MetalANGLE support into a different repository. https://github.com/Berstanio/gdx-metalangle

Berstanio:
Sorry for the earlier miscomment. Hitting "enter" was for some reasons "comment".

So I've seperated MetalANGLE support into a different repository: 
https://github.com/Berstanio/gdx-metalangle
Installation guide can be found in the README.

But this wont work out of the box. The GLKit also needs to be replaced (https://github.com/kakashidinho/metalangle/issues/65). For MOE I already generated the bindings and replaced GLKit (see this pr) this just also needs to be done for robovm. But the problem is I'm not sure how to provide the altered robovm/moe backend. Should I also create a own repo (or include it into the existing gdx-metalangle) which than builds the altered backend and installes it into the local repository?

MrStahlfelge:
> But the problem is I'm not sure how to provide the altered robovm/moe backend.

You can create an own repo and either publish automatically with Jitpack if you have no natives. If you need to compile natives than Sonatype is the repo to publish on.

PokeMMO:
What all needs altered in the robobm backend? Is it natives only, or does it include java changes?

Berstanio:
> What all needs altered in the robobm backend? Is it natives only, or does it include java changes?

It's only java changes. The GLkit bindings needs to be replaced. I've done a initial replacing and it can be found here: https://github.com/Berstanio/libgdx/tree/metalkit-bindings/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm

PokeMMO:
These changes are probably best done by implementing a separate `IOSGraphics` class, like `IOSGraphicsMetalGL` and publishing that as it's own repository. (Until it is mature enough to merge into libgdx proper)

As well as `IOSUIViewController`.

Do the constants changed in `IOSApplicationConfiguration` match the actual values? If so, it might be safe to ignore that file's changes for now.

Berstanio:
> These changes are probably best done by implementing a separate `IOSGraphics` class, like `IOSGraphicsMetalGL` and publishing that as it's own repository. (Until it is mature enough to merge into libgdx proper)

Since I plan to already publish it in a different package "gdx-backend-robovm-metalangle" I think it would be unnecessary to also create a different class. And also there would be another IOSApplicationMetalGL class which instantiates the IOSGraphicsMetalGL and that would be a bit confusing for the user who implements "gdx-backend-robovm-metalangle".
I think both are separated enough. Or does I miss something?

PokeMMO:
The separate class is so users can have a library that just implements the changes into the existing libgdx project without conflicting classes.

Your package can simple depend on the normal robovm backend, and add additional classes. The graphics class can already be changed by overriding the `createGraphics` method when creating the `IOSApplication`

https://github.com/libgdx/libgdx/blob/f27c5685262c3ee55fa9c0267e73de021e6d4b93/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java#L166

Eventually both your class and the original `IOSGraphics` could both exist in libgdx itself allowing the user to select which they wanted to use.

MrStahlfelge:
> The separate class is so users can have a library that just implements the changes into the existing libgdx project without conflicting classes.

Would also work with a different package, but I share your opinion that a different name is better.

Berstanio:
I took a deeper look and I think there are a few problems with that approach. 

I can't easily override IOSApplication. https://github.com/libgdx/libgdx/blob/f27c5685262c3ee55fa9c0267e73de021e6d4b93/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java#L170-L172
In order to return a IOSUIViewControllerMetal and properly override the function IOSUIViewControllerMetal needs to extend IOSUIViewController. But then it would also extend GLKViewController and couldn't extend MGLKViewController. Because of that the set/getDelegate functions aren't working because they would return a GLKViewDelegate where we need a MGLKViewDelegate. However this is solvable with proxies but I think this would be a bit of a overkill.

Overriding IOSGraphics is also not easy. Since very much of the initalisation is done in the constructor of IOSGraphics, and when overriding we need to call the constructor of the super class, nearly everything initailised would need to be reinitailised. 

So I think the only way is than to create completely new classes for IOSApplication etc. which nearly the same functionality. But this is hard to maintain and keep up to date. I would need to frequently and carefully check whether there are some commits on the robovm backend and copy them over. Also when the classes are in the main libgdx repo there needs to be double edits. 
With the approach I suggested I just need to pull and merge new commits which is a lot easier and safer. But I understand that than it is harder to have both options at some point in the main repo.

Did I missed something or are there any ideas of how I could still implement it as a simple extension?

Berstanio:
First of all, is this still the right place to publish information regarding the MetalANGLE implementation?

So what have I done so far:
I created the MetalANGLE ios natives, the MetalANGLE robovm backend and the MetalANGLE moe backend. They are all published to sonatype in the snapshot repository (https://s01.oss.sonatype.org/content/repositories/snapshots/)
The groupdId is "io.github.berstanio" and the artifactsIds are just the normal ids with a "-metalangle" appended e.g. natives-ios-metalangle. The versions are "1.10.1-SNAPSHOT". They completely replace the standard artifacts. 
A basic example of how this can be done can be found here: https://github.com/Berstanio/chigginblazzder
I tested it fairly basic atm but it seems to work.

What still needs to be done:
Further testing needs to be done.
The changes @PokeMMO suggested aren't currently not implemented because of the difficulties I mentioned in my last post. The current approach is in my opinion still the easiest to maintain and wouldn't lead to conflicting classes (since the whole backend will be replaced). But if the general opinion is still that these changes should be done, I could do it.

MrStahlfelge:
> Further testing needs to be done.

What exactly do people need to do for testing with own games and what should be looked for when testing?

Berstanio:
> What exactly do people need to do for testing with own games and what should be looked for when testing?

1. Add this repository: maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
2. Change the gdxVersion to "1.10.1-SNAPSHOT"
3. Change the artifacts: 
"com.badlogicgames.gdx:gdx-backend-robovm:$gdxVersion" 
"com.badlogicgames.gdx:gdx-platform:$gdxVersion:natives-ios" 
to
"io.github.berstanio:gdx-backend-robovm-metalangle:$gdxVersion" 
"io.github.berstanio:gdx-platform:$gdxVersion:natives-ios-metalangle"
4. Add to robovm.xml the "MetalKit" and "Metal" framework (the "OpenGLES" framework currently needs to be kept.) 

These steps can also be seen in the latest 2 commits on https://github.com/Berstanio/chigginblazzder

MovementSpeed:
> > What exactly do people need to do for testing with own games and what should be looked for when testing?
> 
> 1. Add this repository: maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
> 2. Change the gdxVersion to "1.10.1-SNAPSHOT"
> 3. Change the artifacts:
>    "com.badlogicgames.gdx:gdx-backend-robovm:$gdxVersion"
>    "com.badlogicgames.gdx:gdx-platform:$gdxVersion:natives-ios"
>    to
>    "io.github.berstanio:gdx-backend-robovm-metalangle:$gdxVersion"
>    "io.github.berstanio:gdx-platform:$gdxVersion:natives-ios-metalangle"
> 4. Add to robovm.xml the "MetalKit" and "Metal" framework (the "OpenGLES" framework currently needs to be kept.)
> 
> These steps can also be seen in the latest 2 commits on https://github.com/Berstanio/chigginblazzder
> 
> It should be just currenlty looked for different behavior than without MetalANGLE or any other disadvantages.

I tried running your sample project, I also built the MetalANGLE.framework by following the MetalANGLE Dev doc, but when I launch it on the iPhone simulator with launchIPhoneSimulator, the app crashes with `Allocating missing native class MGLContext`. What am I missing?

MovementSpeed:
> > I tried running your sample project, I also built the MetalANGLE.framework by following the MetalANGLE Dev doc, but when I launch it on the iPhone simulator with launchIPhoneSimulator, the app crashes with `Allocating missing native class MGLContext`. What am I missing?
> 
> First of all, thanks for testing!
> You did all right. I could reproduce the issue. I made a mistake during the artifact deployment. But there is now also another error with the simulator that I need to fix.
> And just for your information: You couldn't use MetalAngle as a framework because of the way the bindings are generated. MetalAngle is included over a static library.

I briefly tried the new library, it is working now. I will be trying with my pet 3d engine to see how it does with more complex things. Thank you for doing this by the way!

fgnm:
Hi @Berstanio, tested your RoboVM MetalANGLE backend with a bunch of HyperLap2D tests, many custom shaders, FBOs.. 
It's awesome to see a libGDX game running at 60fps on the simulator, and I haven't found issues at moment.

On physical device (iPhone 7 with iOS 14.7) rendering looks even more stable, with OpenGL backend I usually get fps drop sometimes to 47/50 with game stuttering, this is not happening with Metal, sometimes drop to 59 but it's overall normal. Since changes to test RoboVM are quite painless I'll start using your backend more in depth to see if anything weird arises, thanks for publishing it on maven :)

As also point out by obigu, I get many warnings like that: `warning: (x86_64)  could not find object file symbol for symbol`, but then works great.

I'd love to see a separate PR for RoboVM backend only, I can't test easily MOE since on some games I've already many RoboVM native bindings.

Berstanio:
I have some new about the MOE backend. First of all, the most important:
I will from now on not limit the backend to moe 2.x. I do this because MOE 1.9 made some progress regarding supporting higher java api's. Many of the api's up to java 11 are now supported. 
The main change for the backend is now, that I removed the svm features for reflection/jni config. They are now inside a helper lib as a opt in feature. With many thanks to @anyicomplex the helper api also supports more extensions. Just as a note, these are just meant to be helper configs. It may be, that they are incomplete.
Instructions how to use them can be found here: https://github.com/Berstanio/gdx-graalhelper/
MOE now also supports AudioDevices. AudioRecorder implemenation is planned. 
If anyone wants to port this to robovm-backend: It seems like that mobivm hasn't OpenAL bindings? Because of that, I tried to keep as close as ObjectAL as possible. This was doable except for rebuffering data in the same ALBuffer. If mobivm really doesn't bind OpenAL, this might be easy to fix by extending the ALBuffer api by a simple buffer method.

Berstanio:
The current state of the pr is now, that it should be functionality wise equal to the roboVM backend. 
The backend is also currently used in production by a few applications. Because of that, I would move this pr out of the "draft" state now.

PR #6595, opened 2021-07-12
Audio: Add pause/resume methods to AudioDevice.

PokeMMO:
* Only really useful for Android to save resources when not playing any audio.

A playing AudioDevice will cause Android's audioserver to run constantly, even when the AudioDevice is not being fed any samples. Closing and recreating a new AudioDevice is not cheap, but pausing is.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudioDevice.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidAudioDevice.java
@@ -94,4 +94,14 @@ class AndroidAudioDevice implements AudioDevice {
 	public void setVolume (float volume) {
 		track.setStereoVolume(volume, volume);
 	}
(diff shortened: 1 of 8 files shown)


mgsx-dev (changes requested):
please add something to the CHANGES file (as a breaking change for custom AudioDevice implementations)

PokeMMO:
I miss having default interface methods.

mgsx-dev (approved):

crykn:
The changes look good to me, but what exactly are the comments in `OpenALAudioDevice#pause()` and `#resume()` for?

PokeMMO:
> The changes look good to me, but what exactly are the comments in `OpenALAudioDevice#pause()` and `#resume()` for?

Just to explain why no code is necessary in those methods. There's no point to actually pause the OpenAL output as it will pause itself when you stop sending it samples.

PR #6607, opened 2021-07-22
Fix HexagonalTiledMapRenderer displaying maps with the wrong stagger index

lyze237:
When trying to load a hexagonal map via tiledmap, the stagger index is apparently inverted.

It doesn't matter if you save the file in odd or even, gdx loads that property correctly, but renders it inverted.
Therefore, I concluded that the boolean flag is inverted and flipped it around.

# Before:
### Even:
image
### Odd:
image

# After:
### Even:
image
### Odd:
image

A workaround for the current libGDX version would be to do the map in one of the indexes, then save it with the other index. That loads the map correctly in the current version.

Test code:
```java
public class HexMapRenderer extends ScreenAdapter {
    private final TiledMap map;
    private final HexagonalTiledMapRenderer renderer;

    private final Viewport viewport;

    public HexMapRenderer() {
        map = new TmxMapLoader().load("hex/hex.tmx");
        renderer = new HexagonalTiledMapRenderer(map, 1);

        viewport = new ExtendViewport(16 * 32, 9 * 32);
    }

    @Override
    public void render(float delta) {
        viewport.apply();
        renderer.setView((OrthographicCamera) viewport.getCamera());
        renderer.render();
    }

    @Override
        viewport.update(width, height);
    }
}
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,7 @@
 - API Change: Enable the AL_DIRECT_CHANNELS_SOFT option for Sounds and AudioDevices as well to fix stereo sound
 - API Addition: CameraInputController#setInvertedControls(boolean)
 - API Removal: AnimatedTiledMapTile#frameCount
+- FIX: HexagonalTiledMapRenderer now displays maps with the correct stagger index.
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
@@ -30,8 +30,8 @@ public class HexagonalTiledMapRenderer extends BatchTiledMapRenderer {
 
 	/** true for X-Axis, false for Y-Axis */
 	private boolean staggerAxisX = true;
-	/** true for even StaggerIndex, false for odd */
-	private boolean staggerIndexEven = false;
+	/** true for even StaggerIndex, true for odd */
+	private boolean staggerIndexOdd = false;
 	/** the parameter defining the shape of the hexagon from tiled. more specifically it represents the length of the sides that
 	 * are parallel to the stagger axis. e.g. with respect to the stagger axis a value of 0 results in a rhombus shape, while a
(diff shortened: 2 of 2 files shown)


tommyettinger (dismissed):
This looks like it works based on the screenshots you showed, but I can't tell exactly what's happening with `staggerIndexEven`/`staggerIndexOdd`. The code and the map you used for the screenshots -- can you put them in a test? From Discord, I understand that the existing test isn't correct enough to see how maps are actually rendered; maybe fixing that test could be part of this PR? Since the staggered mode doesn't seem to be terribly usable in conjunction with the Tiled editor right now, I'm approving this just in the hope that this Tiled feature doesn't need anything else.

lyze237:
Since a hex map can be rendered in 4 ways, there are 4 options to change:
image
(Source: https://www.redblobgames.com/grids/hexagons/#coordinates)

The index bool (odd/even) flag was wrong while the axis r/q (or in tiled called x/y) flag is correctly implemented.

When switching the same map in tiled from odd to even, or the other way around, it creates bad tiles as you can't translate from one to another properly. Therefore, you need to manually fix it in tiled.
I didn't do that in the initial post to highlight that it was inverted in libgdx.

For example, here's a map now in "odd":
image

And here a similarly drawn one in "even": 
image

Here's a recording on what you need to do when you want to convert from even to odd:
WwIiSIEoEc

I'll update the test tomorrow and update this PR.

lyze237:
Additionally, just to be sure, please hold off from accepting the PR, I want to be sure that it works in every case so I want to do some more testing.

lyze237:
My initial assumptions were partially wrong, the same wrong as the initial code.

With a bit more investigating it seems like it's only sometimes wrong, and I guess because of y axis differences.
Therefore, we only need to flip the variable sometimes.

Here are tests for all combinations (even, odd, x, y and with a 3x3 and 4x4 grid):
(Expected is a image export from tiled, actual is whatever gdx renders, and stack is both overlayed)
## 3x3
3x3 (Hex_3x3-even-x tmx, Hex)
3x3 (Hex_3x3-even-y tmx, Hex)
3x3 (Hex_3x3-odd-x tmx, Hex)
3x3 (Hex_3x3-odd-y tmx, Hex)

## 4x4
4x4 (Hex_4x4-even-x tmx, Hex)
4x4 (Hex_4x4-even-y tmx, Hex)
4x4 (Hex_4x4-odd-y tmx, Hex)
4xa (Hex_4x4-odd-x tmx, Hex)

And here are two where I initially stumbled upon the issue:

Scholz (Hex_Scholz tmx, Hex)
Scholz (Hex_Scholz-4 tmx, Hex)
Scholz (Hex_Scholz-16 tmx, Hex)
50x50 (Hex_50x50 tmx, Hex)

I had no time yet to create some useful tests for the repo yet, I'll do that this week hopefully, sorry about the delay.

PR #6610, opened 2021-07-24
gdx-setup dialogs

Frosty-J:
Some super simple changes to the gdx-setup dialogues. Basically makes the extensions text match the new red, and some other minor things done for consistency sake. Inspired by #6463 and about half of this is to match what it changed - I didn't try it until it was merged.

Currently, diauloug text gets truncated on Linux due to the window size not being wide enough. It's something I'd like to fix, but that would require seeing what the situation is on macOS and confirming it's consistent across distros - or make the daiologes so they can be packed without looking silly (more robust solution but also more work) - so it might not be destined for this pull request.


crykn (approved):

tommyettinger (approved):
I also like these small clean-ups; they help gdx-setup look professional.

...Deialauggs.

PR #6611, opened 2021-07-24
Prevent repeated mipmap generation in FileTextureArrayData

armingensler:
The current code repeats the Gdx.gl20.glGenerateMipmap call for every subtexture, although a single call at the end should be sufficient. This becomes a significant performance problem if a texture array contains a large number of elements.


mgsx-dev (changes requested):
This change won't work for some custom formats eg. KTX may contains all mipmap levels, and we don't want to generate them.
I think we could generate mipmaps only if no custom data is used.

armingensler:
If I understand you correctly, the newst commit should fix this problem.

mgsx-dev (approved):
A cleaner way to manage this would need some deeper changes but i don't think it's necessary, so this change looks OK to me.

PR #6614, opened 2021-07-31
Use proper preferences directories on desktop

### LINKED ISSUE #6559: Use XDG Base Directory on Linux
#### Issue details
Currently, by default, libGDX [Preferences](https://github.com/libgdx/libgdx/wiki/Preferences) are saved to `~/.prefs` on Linux, which is nonstandard. The [XDG Base Directory](https://wiki.archlinux.org/title/XDG_Base_Directory) specification suggests that, to determine where to write configuration files to, programs should:
- First try the environment variable `$XDG_CONFIG_HOME`.
- If the environment variable is unset, default to `$HOME/.config`.

This standard serves to make the home directory cleaner, and normalize the way different programs save their configurations.

If this is adopted in libGDX, it would likely be necessary to check the (then legacy) `~/.prefs` path in addition to the above two paths, as to not break existing configurations.

This is somewhat similar to #6367, but not the same as this pertains to preferences rather than local storage.

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [X] Linux
- [ ] MacOS

Thanks!

Frosty-J:
## What this pull request does

Preferences are now stored in standard locations for each OS, which should in turn fix #6559. This doesn't consider the recent discussions in that issue, as I'd already completed the majority of this pull request by that point.

## Where are my preferences now?

Existing preferences will not be moved. New preferences will typically be at the following locations:

OS|Location
-|-
Windows 95\*/98/Me|C:\WINDOWS\Application Data
Windows 2000/XP|C:\Documents and Settings\\<YOUR USERNAME\>\Application Data
Windows Vista and newer|C:\Users\\<YOUR USERNAME\>\AppData\Roaming
macOS|/Users/\<YOUR USERNAME\>/Library/Preferences
Linux|/home/\<YOUR USERNAME\>/.config

**Note:** Locations may vary slightly between Windows/Linux installations and be subject to localisation/localization. On 2000/XP, `Application Data` may be localised to `Anwendungsdaten` or `Datos de programa` as a couple examples. Due to this pull request's use of `%APPDATA%`, that won't be an issue for those operating systems.

\*`Application Data` wasn't a standard in Windows 95, but there's no evidence anyone has got Java 7 or newer (as required by libGDX) running on it, so this literally doesn't matter. In the event someone does get it running and the folder doesn't already exist, it'll be created like any other.

Screenshots||
-|-
**Windows 98**<br>Windows 98 screenshot|**Windows XP**<br>Windows XP screenshot
**Windows 10**<br>Windows 10 screenshot|**Ubuntu 20.04**<br>Ubuntu screenshot

## Applicable backends

This pull request applies to the desktop backends (LWJGL and LWJGL3), plus headless. It shouldn't affect Android, iOS or web in any way. The libGDX core hasn't been modified.

Locations where `.prefs/` is hardwired such as in tests, `LwjglAWTCanvas` and `LwjglCanvas` have not been modified. Does anyone actually use those two classes? They don't appear to have LWJGL3 equivalents.

## Backwards compatibility

This shouldn't break existing preferences for games. For simplicity, let's say we have `save.xml` and that this pull request moves preferences from `.prefs` to `AppData`.

* If `save.xml` is found in `AppData`, changes to it will be continue to be stored in `AppData`.
* If `save.xml` is found in `.prefs`, changes to it will continue to be stored in `.prefs`.
* If `save.xml` cannot be found in either of the above locations, it will be created in `AppData`.

So, for existing preferences, it is on the developer to move the data from the old location to the new one if they so decide. `getDefaultPreferencesDirectory()` is public, so they have everything they need to do so. I opted not to move old data mainly because it's easier not to implement such a setup, but also because moving it can't go wrong if you don't move it, and users/developers may get confused if their old file moves.

As for the nitty gritty developers out there, if you're overriding this and that, things shouldn't break too badly! Classes/methods had to be adapted to fit more variables for backwards-compatibility with the old preferences directory, but you can still use them as before with fewer arguments through the power of overloads.

## New configurations

When configuring your desktop or headless game, there's a few more things you can change. Note that the below may differ slightly between backends.

* `preferencesBackupDirectory`: This is how you configure where it'll look for preferences if it fails to find them in the main directory. By default `.prefs/` - exactly how `preferencesDirectory` used to be. If you're creating a new game and don't want to waste a single millisecond on I/O operations, you can set this to `null`.
* `preferencesBackupFileType`: Same as `preferencesFileType` used to be: `Files.FileType.External`
* `getDefaultPreferencesDirectory()`: This gets the default preferences directory. As already discussed, this is probably in `AppData\Roaming` on Windows, `Library/Preferences` on macOS and `~/.config` on Linux.
* `getDefaultPreferencesFileType()`: Gets the default preferences filetype. This is important because it's not guaranteed whether the directory `getDefaultPreferencesDirectory()` returns is relative from home or absolute.

## Assumptions

The code contained within this pull request makes some assumptions. Pretty sure it'll be fine, but I'll list them here in case someone finds reason to object.

### 98/Me Application Data

On DOS-based versions of Windows, the preferences directory is set to `Application Data`. Ideally this would be obtained from special folder `CSIDL_APPDATA` or registry key `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders\AppData` as it can vary between locales. I haven't found an environment variable present on these versions that returns this location, and accessing `CSIDL` or the registry from Java appears too complicated to justify implementing for such a small market share's benefit.

### macOS preferences

It is assumed that the directory `Library/Preferences` will always be present and accessible on macOS. If it turns out there's an envrionment variable or similar that defines this, I'd prefer that, in case Apple one day decides to change this filepath, breaking compatibility. But I'm unaware of such a solution.

## Aspects that need reviewing

This pull request is functional, as far as I can tell. I've tested various Windows operating systems and the three states the preference locations can be in, across the three modified backends.

### ~~macOS~~

~~macOS remains completely untested. I do have a virtual machine of it laying around, but I have some technical issues with it and overall it's just a big pain. I would much prefer if someone here could test on macOS. All we really need to know is that saving to `Library/Preferences` works. Other behaviour should be consistent regardless of OS.~~

See [my comment](https://github.com/libgdx/libgdx/pull/6614#issuecomment-890441249) below. I think macOS is fine.

### Roaming

On Windows, the directories this pull request uses are roaming. Meaning the preferences may follow a user around from computer to computer in certain networked spaces such as offices and schools. I think this is the better setup for the majority of games, but if people disagree this can easily be changed to use local directories instead - `%LOCALAPPDATA%` on Windows 2000 and newer; possibly just `C:\` for 95/98/Me?

### $XDG_CONFIG_HOME

Support for the `XDG_CONFIG_HOME` environment variable on Linux is partial! ~~And untested.~~ This variable can include other variables, but I found it difficult to make that work properly and not worthwhile, as the majority of Linux installations don't even have this variable set, let alone changed to a custom value. So `$HOME` is supported within this variable but that's all.

### Experts

I also require that someone who is experienced with developing libGDX takes a look at the code. Due to the way the different backends are separate from each other in libGDX, there's a lot of duplicate code here, which I think is hard to avoid but unfortunately still makes developing/maintaining slower. Perhaps some of it could be moved someplace else. And enough has changed that I've probably inadvertantly broken a dozen written rules and two dozen unwritten rules while I'm at it.

## Amendments

* Whether intentional or not, Terefang has reminded me that the wiki would need to be updated if this is merged - specifically https://github.com/libgdx/libgdx/wiki/Preferences#storage but there may be other pages too.
* MrStahlfelge linked https://stackoverflow.com/a/586917 as a starting point for getting the localised `Application Data` on Windows 98/Me. It's too much, isn't it? What I already have here will work under different languages - it'll just create an English folder instead of using the correct language. Such old operating systems aren't our priority so I'm perfectly content with things working but not being perfect under them.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java
@@ -949,28 +949,38 @@ public class GwtGL20 implements GL20 {
 		if (pixels == null) {
 			gl.texImage2D(target, level, internalformat, width, height, border, format, type, null);
 		} else {
-			if (pixels.limit() > 1) {
+			if (pixels instanceof FloatBuffer) {
 				HasArrayBufferView arrayHolder = (HasArrayBufferView)pixels;
-				ArrayBufferView webGLArray = arrayHolder.getTypedArray();
-				ArrayBufferView buffer;
-				if (pixels instanceof FloatBuffer) {
-					buffer = webGLArray;
-				} else {
+				ArrayBufferView buffer = arrayHolder.getTypedArray();
+				gl.texImage2D(target, level, internalformat, width, height, border, format, type, buffer);
+			} else if (pixels instanceof IntBuffer) {
+				if (pixels.limit() > 1) {
+					HasArrayBufferView arrayHolder = (HasArrayBufferView)pixels;
+					ArrayBufferView webGLArray = arrayHolder.getTypedArray();
 					int remainingBytes = pixels.remaining() * 4;
 					int byteOffset = webGLArray.byteOffset() + pixels.position() * 4;
-					buffer = Uint8ArrayNative.create(webGLArray.buffer(), byteOffset, remainingBytes);
+					ArrayBufferView buffer = Uint8ArrayNative.create(webGLArray.buffer(), byteOffset, remainingBytes);
+					gl.texImage2D(target, level, internalformat, width, height, border, format, type, buffer);
+				} else {
+					Pixmap pixmap = Pixmap.pixmaps.get(((IntBuffer)pixels).get(0));
+					// Prefer to use the HTMLImageElement when possible, since reading from the CanvasElement can be lossy.
+					if (pixmap.canUseImageElement()) {
+						gl.texImage2D(target, level, internalformat, format, type, pixmap.getImageElement());
+					} else if (pixmap.canUseVideoElement()) {
+						gl.texImage2D(target, level, internalformat, format, type, pixmap.getVideoElement());
+					} else {
+						gl.texImage2D(target, level, internalformat, format, type, pixmap.getCanvasElement());
+					}
 				}
+			} else if (pixels instanceof ByteBuffer) {
+				HasArrayBufferView arrayHolder = (HasArrayBufferView)pixels;
+				ArrayBufferView webGLArray = arrayHolder.getTypedArray();
+				int remainingBytes = pixels.remaining();
+				int byteOffset = webGLArray.byteOffset() + pixels.position();
+				ArrayBufferView buffer = Uint8ArrayNative.create(webGLArray.buffer(), byteOffset, remainingBytes);
 				gl.texImage2D(target, level, internalformat, width, height, border, format, type, buffer);
 			} else {
-				Pixmap pixmap = Pixmap.pixmaps.get(((IntBuffer)pixels).get(0));
-				// Prefer to use the HTMLImageElement when possible, since reading from the CanvasElement can be lossy.
-				if (pixmap.canUseImageElement()) {
-					gl.texImage2D(target, level, internalformat, format, type, pixmap.getImageElement());
-				} else if (pixmap.canUseVideoElement()) {
-					gl.texImage2D(target, level, internalformat, format, type, pixmap.getVideoElement());
-				} else {
-					gl.texImage2D(target, level, internalformat, format, type, pixmap.getCanvasElement());
-				}
+				throw new GdxRuntimeException("Unsupported Buffer type");
 			}
 		}
 	}
--- a/tests/gdx-tests-gwt/src/com/badlogic/gdx/tests/gwt/client/GwtTestWrapper.java
+++ b/tests/gdx-tests-gwt/src/com/badlogic/gdx/tests/gwt/client/GwtTestWrapper.java
@@ -98,6 +98,7 @@ import com.badlogic.gdx.tests.g3d.ModelCacheTest;
 import com.badlogic.gdx.tests.g3d.MultipleRenderTargetTest;
 import com.badlogic.gdx.tests.g3d.ShadowMappingTest;
 import com.badlogic.gdx.tests.g3d.TextureArrayTest;
+import com.badlogic.gdx.tests.gles2.GlTexImage2D;
 import com.badlogic.gdx.tests.gles2.VertexArrayTest;
 import com.badlogic.gdx.tests.gles3.GL30Texture3DTest;
 import com.badlogic.gdx.tests.gles3.NonPowerOfTwoTest;
@@ -293,6 +294,11 @@ public class GwtTestWrapper extends AbstractTestWrapper {
 		tests.add(new GwtInstancer() {
 			public GdxTest instance () {
 				return new GLProfilerErrorTest();
+			}
+		});
+		tests.add(new GwtInstancer() {
+			public GdxTest instance () {
+				return new GlTexImage2D();
 			}
 		});
 		tests.add(new GwtInstancer() {
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/gles2/GlTexImage2D.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/gles2/GlTexImage2D.java
@@ -1,0 +1,198 @@
+/*******************************************************************************
+ * Copyright 2024 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.tests.gles2;
+
+import java.nio.FloatBuffer;
+import java.nio.ByteBuffer;
+
+import com.badlogic.gdx.Gdx;
+import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.Pixmap;
+import com.badlogic.gdx.graphics.glutils.ShaderProgram;
+import com.badlogic.gdx.tests.utils.GdxTest;
+import com.badlogic.gdx.utils.BufferUtils;
+
+/** Added during glTexImage2D corrections.
+ * @author Ret-Mode */
+public class GlTexImage2D extends GdxTest {
+
+	ShaderProgram shader;
+
+	int texture = 0;
+	int pixmapTexture = 0;
+
+	int vertsBuffer = 0;
+	int pixmapBuffer = 0;
+	int uvBuffer = 0;
+
+	Pixmap pixmapCheck;
+
+	String vertexShader = "attribute vec2 vPosition;                    \n" + "attribute vec2 vTexCoords;                   \n"
+		+ "varying   vec2 fTexCoords;                   \n" + "void main()                                  \n"
+		+ "{                                            \n" + "   gl_Position = vec4(vPosition, 0.0, 1.0);  \n"
+		+ "   fTexCoords  = vTexCoords;                 \n" + "}";
+	String fragmentShader = "#ifdef GL_ES\n" + "precision mediump float;\n" + "#endif\n"
+		+ "varying vec2 fTexCoords;                       \n" + "uniform sampler2D uTex2d;                      \n"
+		+ "void main()                                    \n" + "{                                              \n"
+		+ "  gl_FragColor = texture2D(uTex2d, fTexCoords);\n" + "}";
+
+	FloatBuffer verticesData = BufferUtils.newFloatBuffer(8);
+	FloatBuffer pixmapVerticesData = BufferUtils.newFloatBuffer(8);
+	FloatBuffer uvData = BufferUtils.newFloatBuffer(8);
+	ByteBuffer textureColorData = BufferUtils.newByteBuffer(12);
+
+	@Override
+	public void create () {
+
+		float[] vertices = {-1.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, -1.0f, 1.0f};
+		float[] pixmapVertices = {1.0f, 0.0f, -1.0f, 0.0f, -1.0f, -1.0f, 1.0f, -1.0f};
+		float[] uv = {0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 1.0f, 0.0f, 1.0f};
+		byte[] data = {0x7F, 0x00, 0x00, /* r */
+			0x00, 0x7F, 0x00, /* g */
+			0x00, 0x00, 0x7F, /* b */
+			0x00, 0x7F, 0x7F}; /* rg */
+
+		verticesData.put(vertices);
+		verticesData.rewind();
+		pixmapVerticesData.put(pixmapVertices);
+		pixmapVerticesData.rewind();
+		uvData.put(uv);
+		uvData.rewind();
+		textureColorData.put(data);
+		textureColorData.rewind();
+	}
+
+	@Override
+	public void render () {
+		/*
+		 * check if OpenGL context needs to be reloaded; checking only texture should be sufficient; but i guess this check is
+		 * redundant
+		 */
+		if (!Gdx.gl20.glIsTexture(texture)) {
+			reload();
+		}
+
+		/* bump bytes and upload to gpu */
+		for (int colorComponent = 0; colorComponent < textureColorData.capacity(); ++colorComponent) {
+			textureColorData.put(colorComponent, (byte)(textureColorData.get(colorComponent) + 1));
+		}
+		textureColorData.rewind();
+		Gdx.gl20.glActiveTexture(GL20.GL_TEXTURE1);
+		Gdx.gl20.glTexImage2D(GL20.GL_TEXTURE_2D, 0, GL20.GL_RGB, 2, 2, 0, GL20.GL_RGB, GL20.GL_UNSIGNED_BYTE, textureColorData);
+		Gdx.gl20.glActiveTexture(GL20.GL_TEXTURE0);
+
+		/* init drawing */
+		Gdx.gl20.glViewport(0, 0, Gdx.graphics.getBackBufferWidth(), Gdx.graphics.getBackBufferHeight());
+		Gdx.gl20.glClear(GL20.GL_COLOR_BUFFER_BIT);
+
+		/* draw texture built from bytes */
+		Gdx.gl20.glBindBuffer(GL20.GL_ARRAY_BUFFER, vertsBuffer);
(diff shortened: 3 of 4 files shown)


backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessApplicationConfiguration.java
```diff
@@ -16,13 +16,81 @@
...
+	/** Preferences directory for headless. Default depends on operating system. */
+	public String preferencesDirectory = getDefaultPreferencesDirectory();
+	public Files.FileType preferencesFileType = getDefaultPreferencesFileType();
+	public String preferencesBackupDirectory = ".prefs/";
+	public Files.FileType preferencesBackupFileType = Files.FileType.External;
 
 	/** The maximum number of threads to use for network requests. Default is {@link Integer#MAX_VALUE}. */
 	public int maxNetThreads = Integer.MAX_VALUE;
+
+	private enum OperatingSystem {
```
crykn, line 36: As stated in #6559, I think OS checks should use `UIUtils#isWindows`, etc. instead of duplicating this stuff all over the place.

Frosty-J: Thanks, will fix.

Frosty-J:
I'm marking this as ready for review now. I ran out of things to do and decided to try macOS. I don't understand why it doesn't show in Finder, but this is also true for files created by other third-party applications. I know very little about macOS. The important thing is the file gets created safe and sound without permission issues or anything nasty like that.

macOS screenshot

Would still be nice to give people a chance to comment, but at this stage I'm confident merging this won't break everything.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Preferences.java
```diff
@@ -34,18 +34,25 @@
 
 public class Lwjgl3Preferences implements Preferences {
 	private final Properties properties = new Properties();
-	private final FileHandle file;
+	private FileHandle file;
```
tommyettinger, line 37: I'm not clear why this has its own `FileHandle` but never appears to use it outside of one local usage in a constructor. I might be missing something, but could this be made local? I believe IDEA has an automatic inspection for that.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -337,4 +350,54 @@ static Lwjgl3Monitor toLwjgl3Monitor(long glfwMonitor) {
...
+
+		} else if (UIUtils.isMac) {
+			return "Library/Preferences";
+
+		} else if (UIUtils.isLinux) {
+			String configHome = System.getenv("XDG_CONFIG_HOME");
+			if (configHome != null) {
+				//TODO: Someone with regex skills should make this work with all environment variables
+				if (configHome.contains("$HOME")) configHome = configHome.replace("$HOME", System.getenv("HOME"));
+				else if (configHome.contains("$")) configHome = null;
```
tommyettinger, line 376: I can try some things for the regex. I don't like forcing `configHome` to null if any non-HOME environment variables are present, or even any escaped dollar signs.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglPreferences.java
```diff
@@ -33,21 +33,38 @@
 import com.badlogic.gdx.utils.StreamUtils;
 
 public class LwjglPreferences implements Preferences {
-	private final String name;
+	private String name;
 	private final Properties properties = new Properties();
-	private final FileHandle file;
+	private FileHandle file;
```
tommyettinger, line 38: I also wonder here if there's any reason to hold onto an inaccessible `FileHandle` when it's only used in a constructor. I'm not sure what order my comments will show in, but I also mention this for Lwjgl3Preferences.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessPreferences.java
```diff
@@ -34,18 +34,34 @@
 
 public class HeadlessPreferences implements Preferences {
 	private final Properties properties = new Properties();
-	private final FileHandle file;
+	private FileHandle file;
```
tommyettinger, line 37: Also here, again not sure if this `FileHandle` needs to be a member when it seems to be only used in a constructor.

tommyettinger (commented):
This mostly looks good; I'm just shy of approving, and mostly because of how the Linux environment variables are (not) handled. There's also (repeated 3 times because of the 3 platforms) the `private FileHandle file` that's kept in each Preferences when it's only used in a constructor; that can probably be made local. I'll start tinkering with regexes to attempt at least one level deeper of environment evaluation; I'm reluctant to allow unbounded evaluation because a cycle would make the program hang and crash.

Frosty-J:
If I try using a local `file` instead, it gets upset because it's required for `flush()`.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
tommyettinger: This is pretty gnarly, but I think it works, based on limited testing in jshell. It should ignore backslash-escaped dollar signs, treating them as literal dollar signs.

```java
Pattern p = Pattern.compile("(?<!\\\\)\\$(\\w+)");
Matcher m = p.matcher(configHome);
while(m.find()) {
  m.reset(configHome = configHome.replaceFirst("\\Q" + m.group() + "\\E",
    Matcher.quoteReplacement(String.valueOf(System.getenv(m.group(1))))));
}
```

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessPreferences.java
tommyettinger: For some reason, I didn't see that there was a reference already to `file`, and only its `final` status changed. So these should work.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
tommyettinger: The above code can replace the highlighted range in Lwjgl3ApplicationConfiguration, and the similar range with a TODO in
LwjglApplicationConfiguration. That's the `if` and its `else if`.

Frosty-J: Sorry I didn't notice this for a while. I've committed this now. Nighty night.

Frosty-J: Apparently typing `$XDG_CONFIG_HOME` in my terminal wasn't a good idea. The commit messages are screwed.

tommyettinger: Great! Consider this resolved.

MrStahlfelge:
Okay people, where are the desktop devs willing to test, review and approve so we can merge?

crykn:
I finally had some time to look at the code. It seems fine, but I'd change two minor things:

1. I think the names `preferencesBackupDir` , `preferencesBackupFileType`, `backupFile`, etc. are misleading. At least for me, "backup" implies that the location is used secondarily, if the primary directory can't be used for some reason. But actually it's the other way around: the "backup" directory is  checked first. Maybe `preferencesLegacyDir` is a better name?
2. In addition, I don't think that we need to expose the `preferencesBackupDirectory`  and `preferencesBackupFileType` variables. There shouldn't really be a need to change them, so we could just hardcode the values, which should make the whole thing more concise. If people want to disable the "legacy directory first"-behaviour, we can introduce a boolean flag for that.

Frosty-J:
I agree with point 1, especially as that would use the same terminology I added to gdx-setup. Will sort that out tonight.

Point 2: I kinda like the variables exposed, but I agree there's not much need. I'm also thinking now... there's not really a way to tell which directory was used - legacy or new. I suppose most games won't go saying "your save data is in this location", but some certainly might.

What would be ideal is if we could check the legacy location only if the project has been updated from an older libGDX version, as there's really no point checking it for new projects, but I don't see a way of doing that.

I see little reason to add that boolean flag, as you can do `config.preferencesDirectory = ".prefs/";` to make it ignore the new location... but it would still check `.prefs` twice, which is an inefficiency, albeit a very small one and at least it may already be cached in memory by the second read. Could check if the locations are identical before checking the legacy location, negating any benefit to a boolean flag.

Uh... so yes, I'll do something for point 2 tonight. It could do with more thought.

crykn:
> Point 2: I kinda like the variables exposed, but I agree there's not much need unless someone had previously changed their preferences location from the default and now wants to use the new location, which I feel is a realistic scenario.

If people want to continue using their old preferences directories they can just set `preferencesDir` and `preferencesType`.

Frosty-J:
Hardcoding it didn't make it more concise, and this is just asking for trouble if there's a reason to change the directory and/or filetype in future since it's now set in so many places instead of just one.

But what we can do with this route is have some freedom with `allowLegacyPreferences` (LWJGL2) and `setPreferencesConfig(boolean legacy)` (LWJGL3) in `DesktopLauncher`. Currently defaults to `true` - could be changed to `false` at a later date so long as developers are made well-aware (don't know how, though - I'm still seeing people confused by the removal of armeabi natives despite that being well-documented).

Thoughts, anyone? Not only to the setup idea but also crykn's suggestions, implemented above - nothing's set in stone yet.

Frosty-J:
Hurrah, a commit that has more removals than additions! This is why I marked it as draft - knew I was missing something. And I was hoping someone would discuss. Feelin' good, but so much has changed that I'll need to test it again. Will do that after sleeping on it.

crykn:
I've finally had some time to look at this again - sorry it took so long! (1.10.1 still isn't out yet, so I'm just in line with our release cycle, _ahem_)

There is one thing I'd like to change, and that is how new and legacy locations interact with eachother. I think it would be better to do the following:
1. Check first whether the new prefs file exists. If that is the case, ignore the old legacy files.
2. If the new file doesn't exist (which is the case when a legacy app has never saved their preferences with libGDX 1.10.1+) and the legacy option is enabled (which it is by default), then fall back to the legacy file. If that exists, use it to _load_ the values, but don't use it as new save location!
3. If neither the new nor the legacy files exist (or the legacy option is disabled), don't load anything, and continue to use the new locations for saving the prefs.

This ensures that legacy apps still benefit from the new locations, while keeping the old data. For example, `Lwjgl3Preferences` could be changed to something like this (I haven't tested it yet, but in theory this should work as intended):

```java
public LwjglPreferences (FileHandle fileToLoad, boolean legacy) {
	this.file = fileToLoad;
	if (!file.exists()) {
		if (legacy) { // fall back to the legacy file which was used before 1.10.1
			FileHandle legacyFile = new LwjglFileHandle(new File(".prefs", fileToLoad.name()), Files.FileType.External);
			if (!legacyFile.exists()) return;
			fileToLoad = legacyFile;
		} else {
			return;
		}
	}
	InputStream in = null;
	try {
		in = new BufferedInputStream(fileToLoad.read());
		properties.loadFromXML(in);
	} catch (Throwable t) {
		t.printStackTrace();
	} finally {
		StreamUtils.closeQuietly(in);
	}
}
```

Frosty-J:
I like this idea, only I want to avoid there being two copies of preferences floating around. It's not like it would use much disk space or anything, but it would be nice to avoid any scenarios where someone attempts to copy some game save data and ends up copying an old version of it instead because they looked in the wrong directory. I'll see what can be done another time, if that's okay.

orangepascal:
was hoping this would have made it into the newly released 1.11.0  - any reason why it's not in yet? Seems to be a great solution, and it should also make Steam's  cloud-saving options possible which is my main interest in it :)

Frosty-J:
It would be nice for it to move old preferences to the new location, but I haven't yet put the time into investigating this and don't currently have heaps of spare time. Though that functionality could be added later, so I'd be quite happy with this being merged in its current state.

Sesu8642:
Since you are already touching this, did you consider isolating the settings of different games? Right now, two games could override each other's preferences if they choose the same preferences name by coincidence. This could be solved by creating a directory named after the package name of the game.

Frosty-J:
I feel like it should be up to the developer to decide how to uniquely-identify their preferences directory/filenames, but if we were to do something about that I agree it would be easiest to change at the same time as the rest of this pull request. What does everyone else think?

orangepascal:
I'd be ok with that, to be honest, I never thought about it in any of my games! So I can imagine that at some point, this could cause problems for users/players.  So handling this automatically would certainly be a good thing for the players

Sesu8642:
I didn't think about it either when I implemented preferences for my game. I now tried to use a unique directory by  including a forward slash in the preferences name. This worked fine on desktop but caused an exception on Android. I guess in the current release I'd need to check the platform first which is pretty hacky.

tommyettinger:
I think that trying to separate preferences so they are guaranteed to be unique per-game is pretty much impossible, since a malicious or curious user could just rename a preferences folder to be what another game uses. Still, package names are unique enough to avoid accidental collisions, at least, and that's better than having some default name like "game" that everything uses unless it gets changed. My vote's for at least incorporating the package name of the main class (the current ApplicationListener, which can be retrieved from `Gdx.app.getApplicationListener()`) into the default preferences directory.

orangepascal:
Any update on this making it into a new libgdx version?

Frosty-J:
If someone communicates what the schedule is (don't think there is one - this sort of thing really needs discussing when there can be over a year between releases) I may try and push for this, perhaps including the package name if that can be done without breaking everything, to be in the next release. More free time is on the horizon.

crykn:
At the moment there isn't really a clear schedule. Regarding the PR itself: personally, I'm not a fan of including the package name in the config names, but I don't think it's that pressing of an issue anyways. The main thing that I'd change about the PR is the way legacy files are handled (see my comment [above](https://github.com/libgdx/libgdx/pull/6614#issuecomment-1073043666)). After we got that cleared up, we could go ahead and get this merged.

Frosty-J:
If anyone has an old copy of this branch, please do something with it. I may have lost all our preferences.

crykn:
I was really hoping I had a copy of the branch still lying around, but no luck. However, looking into this, it might be possible to recover the old branch (see [this](https://github.com/orgs/community/discussions/64693) guide for example), since [the commits are still there](https://github.com/libgdx/libgdx/compare/master..c60f92c48b8951638d9b6e6d3dee199b83de9e49).

Frosty-J:
Alright, I've copied this into https://github.com/Frosty-J/libgdx/tree/preferences. What a mess! Don't see much point opening another pull request right now since it's not in a state where we're happy to merge (due to the discussions above, less so the merge conflicts).

For future reference:
```
git fetch origin c60f92c48b8951638d9b6e6d3dee199b83de9e49
git checkout -b preferences c60f92c48b8951638d9b6e6d3dee199b83de9e49
```

Issue #6525, opened 2021-05-08
Gradle checkGwt task is broken


GatorQue:
Please ensure you have given all the following requested information in your report.

#### Issue details
Running gradle checkGwt task results in build error.

#### Reproduction steps/code
1. Create new project on Windows using libgdx setup tool (I included desktop, android, and html and tools extension)
2. Run gradlew checkGwt from command line
Expected result:
C:\libgdx\cavegame>gradlew checkGwt

> Task :html:checkGwt
C:\libgdx\cavegame\html\.
../android/assets
Copying resources from ../android/assets to war/
C:\libgdx\cavegame\html\..\android\assets
C:\libgdx\cavegame\html\assets
C:\libgdx\cavegame\html\.
../android/assets
Copying resources from ../android/assets to war/
C:\libgdx\cavegame\html\..\android\assets
C:\libgdx\cavegame\html\assets

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.7.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 7s
4 actionable tasks: 2 executed, 2 up-to-date

Actual result:
C:\libgdx\cavegame>gradlew checkGwt

> Task :html:checkGwt FAILED
Loading inherited module 'com.cosmicrover.cavegame.GdxDefinition'
   Loading inherited module 'CaveGame'
      [ERROR] Unable to find 'CaveGame.gwt.xml' on your classpath; could be a typo, or maybe you forgot to include a classpath entry for source?

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':html:checkGwt'.
> Process 'command 'C:\libgdx\jdk-15\bin\java.exe'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.7.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 1s
3 actionable tasks: 1 executed, 2 up-to-date

#### Version of LibGDX and/or relevant dependencies
LibGDX 1.10.0

#### Please select the affected platforms
[...]

PR #5509, opened 2019-01-12
fix #5238 asset folder classpath with eclipse

mgsx-dev:
More details about my changes in #5238

Tested with eclipse Neon.1a Release (4.6.1) and EGradle Version 2.4.0.

Test steps i made : 

* test 1 :
    * generate a new libgdx project for eclipse IDE
    * import into eclipse as existing eclipse project
    * run desktop launcher
    * build desktop jar (gradlew desktop:dist) and run it
    * verify there is no asset file duplication in the jar
* test 2 :
    * generate a new libgdx project without any IDE option
    * import into eclipse as existing gradle project
    * run desktop launcher
    * build desktop jar (gradlew desktop:dist) and run it
    * verify there is no asset file duplication in the jar

I'm not very confident with this change, there might be important things i missed, so it would be nice if several people test it with other IDE before merging it.

PR #6615, opened 2021-08-02
gdx-setup: Move assets folder to project root; add asset settings; Android/GWT fixes

### LINKED ISSUE #4955: Move the asset folder in to the  root of the Project
since all Modules uses the same assets folder. It would be better to put the asset folder into the root of the Project. This will even solve the run DesktopLauncher which where you must set the android asset folder as the working environment.

- [ x] Android
- [x ] iOS (robovm)
- [x ] iOS (MOE)
- [x ] HTML/GWT
- [x ] Windows
- [ x] Linux
- [x ] MacOS

Frosty-J:
Hi, this aims to resolve #4955 by moving the assets directory to the root of the project.

## What if I don't want my assets folder at project root?

Change can be hard, I get that. Fortunately, there's no need to stick to an old version to retain the old assets directory.

### UI

If using the gdx-setup UI, you can simply click the "Advanced" button then check the "Legacy Assets Dir" box. This reverts it to mostly the old behaviour (difference being `android`'s `assets.srcDirs` is now identical to the other modules).

### Command line

If creating projects from the command line, you can use the `--assetPath` argument. Unlike when using the UI, here you can enter any path you like. E.g. when testing I set it to `myfunkyassets` because that sounded funky.

## Whoa, duplicate content roots!

Yeah, so, in IntelliJ IDEA and Android Studio you'll get a "Duplicate content roots detected" notice when opening your project. If you click the cog icon in the top-right of it, you can disable this notice from appearing in future. Seems it's either put up with this or keep the assets in the `android` module forever.


extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
             aidl.srcDirs = ['src']
             renderscript.srcDirs = ['src']
             res.srcDirs = ['res']
-            assets.srcDirs = ['assets']
+            assets.srcDirs = ['../%ASSET_PATH%']
```
tommyettinger, line 11: I don't see anything that replaces `%ASSET_PATH%`, and as far as I know, Gradle will only replace a String if it's in double-quotes and uses the `"../$ASSET_PATH"` syntax.

tommyettinger (changes requested):
I'm not at all sure what the Windows-specific double-percent variable is doing here, of if it is always replaced anyways. If you've tested that on Linux with the default Android settings, then it could be fine, it just looks very suspicious in cross-platform config.

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
Frosty-J: gdx-setup replaces it when the project is generated. See `GdxSetup.java` line 353.

Frosty-J:
Maybe. If `sourceSets.main.resources.srcDirs` was added into core, we wouldn't need it in `lwjgl2` and `lwjgl3`. Other platforms would be unaffected.

#5509 should've fixed #5623 - I'm unsure why they were still having issues.

Frosty-J:
Can I fix #6525 while I'm here or should that be a separate PR? Literally a one-line addition.

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
tommyettinger: Ah, I didn't know it was a gdx-setup thing. Resolved!

tommyettinger (approved):

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/AndroidManifest.xml
```diff
@@ -1,18 +1,24 @@
...
 
     <application
         android:allowBackup="true"
+        android:fullBackupContent="true"
         android:icon="@drawable/ic_launcher"
         android:isGame="true"
         android:appCategory="game"
-        android:label="@string/app_name">
+        android:label="@string/app_name"
+        tools:ignore="UnusedAttribute">
```
Frosty-J, line 15: Alternatively `tools:targetApi` can be used. Either way, this stops it complaining about about `isGame` and `appCategory` only being used on newer Android versions.

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/AndroidManifest.xml
```diff
@@ -1,18 +1,24 @@
...
         <activity
             android:name="%PACKAGE%.AndroidLauncher"
             android:label="@string/app_name" 
             android:screenOrientation="landscape"
-            android:configChanges="keyboard|keyboardHidden|navigation|orientation|screenSize|screenLayout">
+            android:configChanges="keyboard|keyboardHidden|navigation|orientation|screenSize|screenLayout"
+            android:exported="true">
```
Frosty-J, line 21: Must be specified to avoid warnings in Android Studio Arctic Fox 2020.3.1. Game doesn't run if set to `false`.

Frosty-J:
Oh Lawd, I just wanted some comments in the commit - not for it to clog up this page.

Anyway, there's really nothing more worth committing at this time unless we reach the consensus of #5623 having value.

tommyettinger (approved):
I just tested the changes to the manifest in gdx-liftoff, and they seem to cause no problems at all. Everything else also seems fine.

MrStahlfelge (approved):
Tested Android, Desktop and HTML and it works.
Probably we need to change documentation, as it is stated to use android as work directory.

obigu:
Tested on iOS and it works.

I've also tested the legacy checkbox and assets are created inside `core`. I think resources were put on `android` module unless there was no Android module in which case they were put on `core`.

MrStahlfelge:
> I think resources were put on `android` module unless there was no Android module in which case they were put on `core`.

That's true, there is an if clause in the new code for this but I did not test legacy mode.

PR #6617, opened 2021-08-04
better boundary box API and documentation for manual updates

mgsx-dev:
A classic pitfall with BoundaryBox: when modifying min and/or max vectors, some internal fields (center and dimensions) need to be updated. Previous workaround was to call box.set(box.min, box.max).

Wording in documentation could probably be better, any feedback appreciated.


crykn (approved):

tommyettinger (approved):
Wording seems fine; "inferior" and "superior" are not used often in libGDX docs, but their meaning is clear. I'm all for better docs; approved.

PR #6621, opened 2021-08-06
fix #6618 move vertex array index buffer limit to backends

### LINKED ISSUE #6618: opengl call to glDrawElements() ignores numIndices parameter on PC, but works on Android
#### Issue details
Using opengl graphics library directly ...
    private static GL20 GLES20;
    GLES20 = Gdx.gl20;
 
 Everything works perfectly when launching to Android devices. I have an entire scene with thousands of triangles rendered just fine, but when running on PC there are extraneous triangles being drawn all over the scene. I tracked the problem down to  glDrawElements(). On PC desktop launch, glDrawElements() draws incorrect number of indices, ignoring the numIndices second parameter completely.

    GLES20.glDrawElements(GL20.GL_TRIANGLES, numIndices, GL20.GL_UNSIGNED_SHORT, mIndices);

For example, changing the numIndices to 3 should render only one triangle. 

    GLES20.glDrawElements(GL20.GL_TRIANGLES, 3, GL20.GL_UNSIGNED_SHORT, mIndices);

Here is the expected correct result on Android.

  imageandroid

But the desktop launch show all triangles, ignoring numIndices completely.

imagepc
#### Reproduction steps/code
An SSCCE can be found at
https://gist.github.com/arashbi/5659096

NOTE: There is one bug. The mColorData needs to be changed to floating points or the shader will crash.

1.10.0

#### Stacktrace
NA

- [ ] Windows 10

mgsx-dev:
Tested with LWJGL2 and LWJGL3 backends. Need to be tested with Android and iOS backends.

About GWT backend, it's a bit special (as usual) since it's partially implemented. As said in the code : "// FIXME this is assuming WebGL supports client side buffers...". That's why i added some documentation at interface level. Content of the Buffer parameter is not used at all, only its position. That means it uses the content of the bound index buffer (GPU side). 

Side note: the test is excluded for GLES3 context because vertex array is not supported with GLES3.


Tom-Ski (approved):

NathanSweet:
Looks OK. Can merge if good on Android and iOS.

mgsx-dev:
I resolved conflits. If someone could test on Android and iOS, it'd be awesome. Simply run VertexArrayTest. By touching screen 9 times, you should see 3 cycles of :
* full quad
* then a half quad (bottom left triangle)
* then a half quad (top right triangle)
(you can test on desktop if you're not sure about what you're seeing).

Thank you.

SimonIT:
I'm not sure if I'm doing something wrong on Android, but I only see the square no matter how often I touch the screen. Tested it on the emulator and my phone, same on both devices. If I run it on my pc it does what you described

obigu:
On iOS 15.1.1 and it works properly.
On Android, like @SimonIT has reported, I only see the square (Pixel 3 with Android 12).

obigu:
Ok, the problem is simply that `Gdx.input.isButtonJustPressed(Buttons.LEFT)` is not reported as a tap on Android :D

mgsx-dev:
Thank you guys, i fixed the test to use justTouched() instead.

PR #6627, opened 2021-08-14
README etc. changes; fix broken links; remove executable permission from non-executables

Frosty-J:
I don't want to add any more feature creep to an existing pull request, so here we go!

This is another pretty simple pull request, but **I'll number each change here** in no particular order so we're all on the same page and can easily argue.

1. Replaced the PNG libGDX logo with the SVG one. This should look a little better than a scaled PNG. Any differece to filesize is insignificant. I doubt we need to worry about compatibility issues at this point - even GitHub itself uses SVGs - but let me know if you spot any issues all the same.
2. Reference to "Mac OS X" and "MacOS" has been changed to "macOS", in line with Apple's branding since 2016.
3. "your WebGL enabled browser" has turned into "browser". This is simply what happened when I replaced the README['](https://frinkiac.com/img/S07E25/408241.jpg)s first sentence with the one from libGDX.com. As all web browsers should support WebGL these days, I feel stating that isn't required. Unless we want "WebGL" or some other term to appear on the page somewhere to clarify the technology being used.
4. Changed the "As we switched to Gradle" part, since libGDX 1.0 was a long time ago now.
5. Replaced reference to the forums and IRC with Discord and Reddit, as the former two are both dead. I think we can agree Discord is preferred, but it's nice to have a couple choices.
6. Last, and pettiest of all, minor grammatical changes.

**The "Supported By" section could do with updating**, but I'd feel more comfortable leaving that to someone who is in charge of these things.

* RoboVM and Spine are probably fine (you'd have to ask someone who is more familiar with them).
* Saikoa now redirects to Guardsquare. A little confusing.
* The Intel logo is outdated.
* Excelsior JET is dead and having a "learn more" button that just opens the image is pure bait.
* Nextpeer is also dead. I think its logo is outdated and its white name is completely invisible on GitHub's white background.
* I'm aware of #6142, which the above two points are applicable to.
* Not too happy about them being Bitly links (because it's a third-party URL shortener - I don't have anything against Bitly specifically) but I guess someone might be using them for analytics.

And, of course, I can't do anything about the Patreon page, which... in summary, you can find something outdated or wrong in nearly every sentence. If I were to give money, I'd want to be confident it's not going into the void.


tommyettinger (approved):
Looks solid; these are common-sense changes. I definitely think that removing links to the no-longer-available forums and no-longer-alive IRC are important, since people who mistakenly visit those will think libGDX is dead. I also agree that someone with authority over the financials should look at the Patreon and sponsors links/images -- is that @NathanSweet or @Tom-Ski these days?

NathanSweet:
Changes look good. Sadly I don't have access to that sweet Patreon money/account (don't particularly want the responsibility though!).

crykn (approved):

tommyettinger:
So who would we ask about the sponsors and Patreon? badlogic doesn't particularly like getting pinged, if I recall correctly. MobiDevelop? This could be merged now, but it would be great to have the sponsors cleaned up and edit the README.md in one sweep.

crykn:
IIRC badlogic was the one responsible for the Patreon account. Just as a small side-note: a suggestion for a new Patreon  description, profile picture and header image can be found on the team's ClickUp thingy.

Frosty-J:
Decided to see if there are any more references to the forums. There were some, albeit only in code comments (and changelogs, but rewriting history is uncalled for). The following commits fix dead badlogicgames.com and box2d.org forum links, and replace links that redirect with their destinations in case the redirects rot. The link in the Bullet license header is well and truly dead, but we don't have the right to alter it, so ¯\\_(ツ)_/¯.

tommyettinger:
You're really going above and beyond here, thanks again! The removal of both forum and IRC mentions in CONTRIBUTING.md is a good catch. Fixing the URLs is also valuable.

Frosty-J:
Maybe CONTRIBUTING.md should pretty much be replaced with the contents at https://libgdx.com/dev/contributing/? It looks to be a newer version of much the same thing. I'm not sure if anything even links to that file. But it has its differences from the website version, so I'm not sure.

PR #6630, opened 2021-08-16
Texture Packer page indexes

### LINKED ISSUE #6239: TexturePacker output filenames are unintuitive
The output of TexturePacker is a bunch textures, named (for example):

    tex.png
    tex2.png
    tex3.png
    tex4.png

When reading textures in a loop, this means that special-case code has to be written to handle the first (unnumbered) file.

It may be useful to include an option that can produce consistent filenames, e.g.

    tex1.png
    tex2.png
    tex3.png
    tex4.png

Frosty-J:
Don't be deceived by the number of changed lines. Only lines 204 and 238 in `TexturePacker.java` actually change any functionality. But those lines make this a breaking change! At least, I think it does - the linked issue was labelled as breaking.

It aims to fix #6239 by adding an index to the first texture page. Just like previously, an index isn't added if the atlas is only one page long.

No-one asked for this, but I also made the index start at 0 instead of 1. My reasoning behind this is it's easier to iterate over your textures and all that this way - no need to have `+ 1` all over the place.

There's also a little bit of tidy-up to reduce warnings about redundancy and whatnot, but it's a very minimal effort.

P.S. I actually have git set up a bit better now - there'll be no more "Frosty-J and Frosty-J committed" business.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
@@ -201,7 +201,7 @@ public class TexturePacker {
 		File packDir = packFileNoExt.getParentFile();
 		String imageName = packFileNoExt.getName();
 
-		int fileIndex = 1;
+		int fileIndex = 0;
 		for (int p = 0, pn = pages.size; p < pn; p++) {
 			Page page = pages.get(p);
 
@@ -233,9 +233,9 @@ public class TexturePacker {
 			page.imageHeight = height;
 
 			File outputFile;
-			while (true) {
+			do {
 				String name = imageName;
-				if (fileIndex > 1) {
+				if (pages.size > 1) {
 					// Last character is a digit or a digit + 'x'.
 					char last = name.charAt(name.length() - 1);
 					if (Character.isDigit(last)
@@ -246,8 +246,7 @@ public class TexturePacker {
 				}
 				fileIndex++;
 				outputFile = new File(packDir, name + "." + settings.outputFormat);
-				if (!outputFile.exists()) break;
-			}
+			} while (outputFile.exists());
 			new FileHandle(outputFile).parent().mkdirs();
 			page.imageName = outputFile.getName();
 
@@ -704,9 +703,8 @@ public class TexturePacker {
 			if (getClass() != obj.getClass()) return false;
 			Rect other = (Rect)obj;
 			if (name == null) {
-				if (other.name != null) return false;
-			} else if (!name.equals(other.name)) return false;
-			return true;
+				return other.name == null;
+			} else return name.equals(other.name);
 		}
 
 		@Override
@@ -719,7 +717,7 @@ public class TexturePacker {
 		}
 	}
 
-	static public enum Resampling {
+	public enum Resampling {
 		nearest(RenderingHints.VALUE_INTERPOLATION_NEAREST_NEIGHBOR), //
 		bilinear(RenderingHints.VALUE_INTERPOLATION_BILINEAR), //
 		bicubic(RenderingHints.VALUE_INTERPOLATION_BICUBIC);
@@ -802,10 +800,10 @@ public class TexturePacker {
 		return false;
 	}
 
-	static public interface Packer {
-		public Array<Page> pack (Array<Rect> inputRects);
+	public interface Packer {
+		Array<Page> pack(Array<Rect> inputRects);
 
-		public Array<Page> pack (ProgressListener progress, Array<Rect> inputRects);
+		Array<Page> pack(ProgressListener progress, Array<Rect> inputRects);
 	}
 
 	static final class InputImage {
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 [1.10.1]
 - [BREAKING CHANGE] Android Moved natives loading out of static init block, see #5795
+- [BREAKING CHANGE] Texture Packer: Filenames are now zero-indexed, including the first file (when there are multiple pages) - e.g. pack.png, pack2.png would now be pack0.png, pack1.png
 - API Addition: ObjLoader now supports ambientColor, ambientTexture, transparencyTexture, specularTexture and shininessTexture
 - API Addition: PointSpriteParticleBatch blending is now configurable.
 - TOOLS Features: Blending mode and sort mode can be changed in Flame particle 3D editor.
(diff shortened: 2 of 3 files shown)


extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
```diff
@@ -719,7 +717,7 @@ static public String getAtlasName (String name, boolean flattenPaths) {
 		}
 	}
 
-	static public enum Resampling {
+	public enum Resampling {
```
tommyettinger, line 720: I know IDEA gives some redundancy notice about how inner `enum`s are always `static`, and don't need the static modifier, but it can be a little tricky to actually know that classes aren't always static, but enums are. I'm fine with this how it is here, but I'd also be fine if the `static` stayed.

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
```diff
@@ -201,7 +201,7 @@ private void writeImages (File outputDir, String scaledPackFileName, Array<Page>
 		File packDir = packFileNoExt.getParentFile();
 		String imageName = packFileNoExt.getName();
 
-		int fileIndex = 1;
+		int fileIndex = 0;
```
tommyettinger, line 204: I like this better, but it is a breaking change. It might be best if this could be configured with a flag, to minimize breakage for older code. If `legacyPages` is true, it uses the old behavior with `fileIndex = 1`; otherwise, it uses 0. This could probably be a static variable without any issue, because the TexturePacker I think only runs on desktop platforms, and if you set it to legacy you probably want to keep it on legacy. If legacy mode is on by default, then this no longer is a breaking change.

tommyettinger (changes requested):
I think this doesn't have to be a breaking change if it is made configurable.

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
Frosty-J: Should it also exclude the first page from being given an index (like the current behaviour) or only change the starting value?

tommyettinger: It would avoid being a breaking change if it maintains the same behavior as before the PR, by default. As long as there's an option, though, you could go wild with changes if people opt in.

Frosty-J:
@tommyettinger Did you know there's a `legacyOutput` option? This defaults to `true` and isn't documented on the wiki. Results in a smaller atlas file. I guess you already do, since you commented at #6316.

I'm going to add that `legacyPages` flag and might revert some of the other changes - depends if anyone else chimes in on the matter, and if they don't it's a figurative coin flip.

I should go to bed - marking this as draft so someone doesn't go merging it in its current state, even if I end up opening it again tomorrow.

tommyettinger:
> @tommyettinger Did you know there's a `legacyOutput` option? This defaults to `true` and isn't documented on the wiki. Results in a smaller atlas file. I guess you already do, since you commented at #6316.

Indeed I do know, I set `legacyOutput` to false for the large atlases of Dawnlike tiles! I should probably edit that wiki.

Frosty-J:
Tested as working. Marking as draft was overkill, in hindsight.

tommyettinger (approved):

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
```diff
@@ -233,9 +233,9 @@ private void writeImages (File outputDir, String scaledPackFileName, Array<Page>
 			page.imageHeight = height;
 
 			File outputFile;
-			while (true) {
+			do {
 				String name = imageName;
-				if (fileIndex > 1) {
+				if (settings.legacyPages && fileIndex > 1 || !settings.legacyPages && pages.size > 1) {
```
NathanSweet, line 238: Needs parenthesis to make it clear.

NathanSweet:
Sorry, but I don't really see the usefulness of this setting. It's not difficult to read files with the naming before this PR. Most people use TextureAtlas to do the reading anyway, so the file names are unimportant. IMO for any feature, ideally default behavior should be found so a setting is not needed. Setting should only be added where necessary.

> Just like previously, an index isn't added if the atlas is only one page long.

The packer doesn't know how many page images there will be. It may pack one page the first time, then later be run again, append to the atlas file and pack additional pages.

> No-one asked for this, but I also made the index start at 0 instead of 1.

When naming files I prefer `image1.png`, `image2.png`, etc. `image0.png` just looks ugly to me personally.

It's common that an atlas has only one page image, in which case `image1.png` looks worse than `image.png`. However, I would vote to not merge this PR and to close #6618 as a non-issue.

Frosty-J:
> Settings should only be added where necessary.

I would prefer to not use a setting, but that would make this a breaking change and not everyone is a fan of that. Not a fan of discovering that we're already using a legacy output mode by default just for the sake of ensuring everything keeps ticking along unchanged.

> The packer doesn't know how many page images there will be.

This is a pretty good point. In the event indexes are used, perhaps an unindexed filename should be deleted (if present) to tidy up from previous runs. I don't think there's anything in place for if the number of pages shrinks either.

> `image0.png` just looks ugly to me personally.

Could fix this with another setting. Haha, just kidding. Yeah, it's divisive.  &nbsp;:/

> It's common that an atlas has only one page image, in which case `image1.png` looks worse than `image.png`.

Agreed. In the event it only has one page, it would be `image.png` either with or without this pull request.

NathanSweet:
> The packer doesn't know how many page images there will be.
TexturePackerFileProcessor processes a directory hierarchy of image files. It first deletes all output files, then creates and runs a TexturePacker for each directory:
https://github.com/libgdx/libgdx/blob/master/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePackerFileProcessor.java#L282-L285
Each run appends to the output of the previous runs. Checking the number of pages in the current run is insufficient to know if the final atlas will have only one page. We'd need to make that determination before running the packer the first time.

I greatly appreciate your enthusiasm, but I'm not convinced this needs changing. Does it really matter if the images are named `image.png`, `image2.png`, etc? If we can have `image.png` for one image and `image1.png`, `image2.png`, etc when there are multiple images, then I could agree that is an improvement. Doing that will take more effort though, given the above. Is it worth it? Can it be done easily? If it can, I'd prefer not to have a setting. We could have the new logic only when legacy is false, which probably affects fewer people.

Frosty-J:
Thanks for the code link and explanation. You're right - this does require more thought. And due to how it goes along one subdirectory at a time like that, I can't currently see how to solve this.

It doesn't *need* changing, though 5 people agree with the original issue with no disagreements, so it would certainly be nice to have.

PR #6640, opened 2021-08-27
gdx-setup UI: Fix Box2D being included by default

Frosty-J:
In #6463, the Box2D checkbox was set to be unselected by default, but this was only a visual change. With this pull request, now it should only be included if the user actually selects the checkbox.

This originally had stuff for removing gdx-tools from the UI, as it's unsupported in LWJGL3 so its checkbox does nothing, but following discussions on Discord that's no longer the case until we get a clearer picture on the state of gdx-tools and LWJGL3.

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
@@ -125,7 +125,6 @@ public class GdxSetupUI extends JFrame {
 		builder = new ProjectBuilder(new DependencyBank());
 		modules.add(ProjectType.CORE);
 		dependencies.add(builder.bank.getDependency(ProjectDependency.GDX));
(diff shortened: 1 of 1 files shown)


extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
```diff
@@ -125,7 +125,6 @@ public void mouseDragged(MouseEvent e) {
 		builder = new ProjectBuilder(new DependencyBank());
 		modules.add(ProjectType.CORE);
 		dependencies.add(builder.bank.getDependency(ProjectDependency.GDX));
-		dependencies.add(builder.bank.getDependency(ProjectDependency.BOX2D));
```
tommyettinger, line 128: One-line PR? Sensible change? Approved.

tommyettinger (approved):

PR #6648, opened 2021-09-13
Add bulk format commits to .git-blame-ignore-revs

PokeMMO:
* Run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to use
* Requires git 2.23 or newer


SimonIT (approved):
Works fine

Issue #6646, opened 2021-09-11
Add support for borderless fullscreen mode on supporting desktop OSes


MontyHimself:
#### Issue details
On desktop, libGDX currently offers windowed mode and fullscreen mode. There is a special third option, typically called borderless fullscreen, whereby a window with the exact characteristics of the monitor's display mode is created. There are two main advantages to this. One is smoother switching between the 'fullscreen' window and other programs (no black flickering due to display mode changes). The other is that it prevents the 'fullscreen' window from being iconified when focusing on another window on a different monitor, which is what happens when using the traditional fullscreen mode with autoIconify set to true.

For Windows (only tested on Windows 10), one can activate borderless fullscreen as follows:
```java
public static void setBorderlessFullscreenMode(Graphics.Monitor monitor) {
    var window = ((Lwjgl3Graphics) Gdx.graphics).getWindow();
    var mode = (Lwjgl3Graphics.Lwjgl3DisplayMode) Gdx.graphics.getDisplayMode(monitor);
    GLFW.glfwSetWindowMonitor(window.getWindowHandle(), 0, 0, 0, mode.width, mode.height, mode.refreshRate);
}
```

The functionality is documented here: https://www.glfw.org/docs/latest/window_guide.html#window_windowed_full_screen
What is not mentioned, however, is that one needs to set the monitor parameter to 0 for this to work (see https://github.com/glfw/glfw/issues/1267).

I think it would be great for this to be added as a method in `Gdx.graphics`.

Someone should test whether this also works on macOS and Linux.

#### Version of libGDX and/or relevant dependencies
libGDX 1.10.0 with LWJGL3 backend

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [X] Windows
- [X] Linux
- [X] MacOS

Frosty-J:
I can test stuff on Linux, bearing in mind behaviour may differ between distros and window managers. Might give that a dabble later today, might not.

#4785 may be useful to read through if you haven't already seen it, with its `setAutoIconify()` and all that.

crykn:
Borderless fullscreen is already supported, just use:
```java
Graphics.DisplayMode displayMode = Lwjgl3ApplicationConfiguration.getDisplayMode();
Gdx.graphics.setUndecorated(true);
Gdx.graphics.setWindowedMode(displayMode.width, displayMode.height);
```

MontyHimself:
[...]

PR #6649, opened 2021-09-13
Fix LWJGL 3 borderless fullscreen.

crykn:
The following code snippet is supposed to create a borderless fullscreen window:

```java
Gdx.graphics.setUndecorated(true);
Graphics.DisplayMode displayMode = Gdx.graphics.getDisplayMode();
Gdx.graphics.setWindowedMode(displayMode.width, displayMode.height);
```

However, while [testing some things for #6646](https://github.com/libgdx/libgdx/issues/6646#issuecomment-918095556), we discovered that on some systems the taskbar on Windows just stays visible, while on other systems the taskbar may be no longer visible, but is still clickable (i.e. clicking the lower part of the window just triggers taskbar shortcuts, etc.). To some extent, the bug was introduced in #6178, but even before then, in 1.9.11, this happened when the window was placed on a different monitor than the primary one. 

This PR fixes all of that, i.e. it makes borderless fullscreen (as can be achieved by the snippet above) working (again) with the LWJGL 3 backend. Tested on Windows as well as on macOS (where apparently the `glfwSetWindowPos` call has to happen after the `glfwSetWindowSize` one, or the window position doesn't take the menu bar into account).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,7 @@
 - API Fix: Escaped characters in XML attributes are now properly un-escaped
 - Bug Fix: AssetManager backslash conversion removed - fixes use of filenames containing backslashes
 - gdx-setup now places the assets directory in project root instead of android or core. See advanced settings (UI) or arguments (command line) if you don't want it in root.
+- API Fix: Resolved issues with LWJGL 3 and borderless fullscreen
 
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

PR #6650, opened 2021-09-14
Fix #6643 and improvements to iOS supported orientations logic

### LINKED ISSUE #6643: getNativeOrientation() on iOS returns the current device orientation, not the native one
On iOS backend ( see https://git.io/JEFzU ) `getNativeOrientation()` returns the current device orientation, not the native one. So when you handle device rotations (e.g. within the `ApplicationAdapter`'s resize method), you cannot know exactly what situation the device is in. BTW, on Android `getNativeOrientation()` always report the correct value (i.e. the same value independently from the current device rotation).

I would suggest a tricky code, in order to precalculate (on iOS) the native orientation:

```
int screenWidth = Gdx.graphics.getBackBufferWidth();
int screenHeight = Gdx.graphics.getBackBufferHeight();
int deviceRotation = Gdx.input.getRotation();

if (((deviceRotation == 0  deviceRotation == 180) && (screenWidth >= screenHeight)) 
    ((deviceRotation == 90 || deviceRotation == 270) && (screenWidth <= screenHeight))) {
    _nativeDeviceOrientation = Input.Orientation.Landscape;
}
else {
    _nativeDeviceOrientation = Input.Orientation.Portrait;
}
```

Feedback would be appreciated. Thanks for your hard work!

obigu:
Fixed #6643.

Also removed deprecated(on iOS6) method to handle supported orientation `shouldAutorotateToInterfaceOrientation` and improved the one that replaces it, `getSupportedInterfaceOrientations()`, to distinguish between iPad OS and iOS (following official docs recommendation only to add PortraitUpsideDown to iPad. (https://developer.apple.com/documentation/uikit/uiviewcontroller/1621435-supportedinterfaceorientations).

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java
@@ -606,13 +606,7 @@ public class DefaultIOSInput extends AbstractInput implements IOSInput {
 
 	@Override
 	public Orientation getNativeOrientation () {
-		switch (app.uiApp.getStatusBarOrientation()) {
-		case LandscapeLeft:
-		case LandscapeRight:
-			return Orientation.Landscape;
-		default:
-			return Orientation.Portrait;
-		}
+		return Orientation.Portrait;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


MrStahlfelge (approved):
LGTM, but a second review by @muratmat would be appreciated.

muratmat:
The fix seems good to me! 👍

PR #6651, opened 2021-09-14
Upgrade to RoboVM 2.3.14

obigu:


tommyettinger:
We're skipping 2.3.13, it looks like? It's good to be up-to-date, certainly. I can't say I know enough about RoboVM to be able to approve this, though I don't disapprove at all.

PR #6652, opened 2021-09-17
Replace gradle deprecations

SimonIT:
I'm not sure why the `compile` -> `api` change forces additionally adding the resources to the classpath of the gwt tests. If anyone knows a better solution or if this is the wrong way, please let me know
I upgraded gradle to 6.9 for the native M1 support


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
```
Frosty-J, line 6: Is there any reason to use 6.9 over 6.9.1? The latter's fixes are:

* [#18089] - Deprecate jcenter() repository (Backport to 6.x)
* [#17950] - Renaming and recreating the project directory causes Gradle to lose track of changes on Windows (Backport to 6.x)
* [#17949] - Gradle's up-to-date checks do not work on Windows FAT drives (Backport to 6.x)

Frosty-J (commented):
We were still using `compile`? (O_o) I've used 6.9 myself for stuff, so I don't see any obvious issue with switching to it.

gradle/wrapper/gradle-wrapper.properties
tommyettinger: Agreed, 6.9.1 seems to be a small set of changes that are valuable to some users. While we're updating, it's two more chars, might as well. It is also certainly important that people move away from jcenter(), because it is going to close down entirely soon.

tommyettinger:
Maybe the GWT resources change is required by junit 4.13.2, since that updated from 4.11 ? I'm hoping that classpath resources change only affects libGDX's tests, and not any user code that needs, say, Arial-15.fnt .

gradle/wrapper/gradle-wrapper.properties
Frosty-J: Where's my "mark as resolved" button? Anywho, resolved in a5945c2 (dead link but you can still see the commit listed on this page).

SimonIT:
There shouldn't be any impact on the deployed .jar's. I controlled the generated .jar files for the resources and everything is in place. I had the same effect at the gdx-video tests by the way. I don't the a reason why junit should have any connection to gwt, gdx-video doesn't even has junit as dependency

gradlew
```diff
@@ -28,16 +44,16 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
MrStahlfelge, line 47: Why do we have this change?

gradlew
```diff
@@ -109,10 +126,11 @@ if $darwin; then
     GRADLE_OPTS="$GRADLE_OPTS \"-Xdock:name=$APP_NAME\" \"-Xdock:icon=$APP_HOME/media/gradle.icns\""
 fi
 
-# For Cygwin, switch paths to Windows format before running java
-if $cygwin ; then
+# For Cygwin or MSYS, switch paths to Windows format before running java
+if [ "$cygwin" = "true" -o "$msys" = "true" ] ; then
```
MrStahlfelge, line 130: Technically, this has nothing to do with Gradle deprecations. I am not sure what MSYS is, is this the Linux subsystem on Windows?

I suspect the whole file is copied from another project, can you give the source?

MrStahlfelge (commented):
The changes compile -> api are safe. The changed gradlew/gradlew.bat files are not really in scope for what the PR's title says, please give the sources for the changes. If it is Gradle's current template it will be safe to use.

gradlew
SimonIT: I just run the normal upgrade command `gradlew wrapper --gradle-version 6.9.1`, nothing special

lyze237: MSYS is something similar to CYGWIN, but not WSL.
So it's cool to have that change in.

gradlew
SimonIT: I don't know why gradle changed it in their template. I just run the normal upgrade command gradlew wrapper --gradle-version 6.9.1, nothing special

PR #6655, opened 2021-09-20
TMX built-in Point Object support.

lazynessmind:
## Info:

* With tiled, you can add point objects that consist in a simple 2D coordinate, and currently, it's not supported by libgdx.

image

## Implementation:

* Added an object that holds the x and y values called PointMapObject.
  * The implementation is based on the RectangleMapObject to keep the same code design.
* While loading the map objects, check if it contains a child called "point" and adds it.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
@@ -15,6 +15,7 @@ import com.badlogic.gdx.maps.objects.EllipseMapObject;
 import com.badlogic.gdx.maps.objects.PolygonMapObject;
 import com.badlogic.gdx.maps.objects.PolylineMapObject;
 import com.badlogic.gdx.maps.objects.RectangleMapObject;
+import com.badlogic.gdx.maps.objects.PointMapObject;
 import com.badlogic.gdx.maps.tiled.TiledMapTileLayer.Cell;
 import com.badlogic.gdx.maps.tiled.objects.TiledMapTileMapObject;
 import com.badlogic.gdx.maps.tiled.tiles.AnimatedTiledMapTile;
@@ -356,6 +357,8 @@ public abstract class BaseTmxMapLoader<P extends BaseTmxMapLoader.Parameters> ex
 					object = new PolylineMapObject(polyline);
 				} else if ((child = element.getChildByName("ellipse")) != null) {
 					object = new EllipseMapObject(x, flipY ? y - height : y, width, height);
+				} else if ((child = element.getChildByName("point")) != null) {
+					object = new PointMapObject(x, y);
 				}
 			}
 			if (object == null) {
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,7 @@
 - API Fix: Escaped characters in XML attributes are now properly un-escaped
 - Bug Fix: AssetManager backslash conversion removed - fixes use of filenames containing backslashes
 - gdx-setup now places the assets directory in project root instead of android or core. See advanced settings (UI) or arguments (command line) if you don't want it in root.
+- API Addition: TMX built-in Point Object is now supported.
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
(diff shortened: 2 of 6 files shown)


MobiDevelop:
Looks good to me.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -356,6 +357,8 @@ protected void loadObject (TiledMap map, MapObjects objects, Element element, fl
 					object = new PolylineMapObject(polyline);
 				} else if ((child = element.getChildByName("ellipse")) != null) {
 					object = new EllipseMapObject(x, flipY ? y - height : y, width, height);
+				} else if ((child = element.getChildByName("point")) != null) {
+					object = new PointMapObject(x, y);
```
MobiDevelop, line 361: Does this y position need to potentially flip based on the flipY option?

mgsx-dev (changes requested):
I marked it as a breaking change because old behavior was to parse point object as rectangle object (it was somehow supported), so user code will need to adapt.
We need to add something about this in the CHANGES file (breaking change section at the beginning)

MobiDevelop:
Yeah, the default in the unknown type cases was to create a rectangle - in this case it would be a rectangle with the x/y position and w/h of 0. Good callout @mgsx-dev.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
lazynessmind: Probably not. It's a 2D coordinate, 100 from top to bottom its the same as bottom to top.

lazynessmind:
> I marked it as a breaking change because old behavior was to parse point object as rectangle object (it was somehow supported), so user code will need to adapt.
> We need to add something about this in the CHANGES file (breaking change section at the beginning)

Yeah, I never used the rectangle as point objects in my projects, so I don't know if it breaks other projects. :sweat_smile:

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
MobiDevelop: I'm not sure why it being a 2D coordinate means it wouldn't need to flip (just like the Y position of other objects).

In a map that is 100x100, a point at say (10,10) isn't going to be correct when the Y-axis when flipY.

lazynessmind: In my tests, it doesn't make any difference.

MobiDevelop: Okay, as long as it's been tested - it's been a long time since I've looked at this particular code so your understanding is probably better than mine at this point.

tommyettinger: Hmm, EllipseMapObject directly above the highlighted lines does use different y when flipY is true. What's different about a PointMapObject? Also, do the existing libGDX tests cover point objects? There's no new tests here.

mgsx-dev: height variable in this context is the height of the object itself, which is zero for a point. That's why flipping is not necessary here.
About tests, i agree, it would be better to improve the existing [TiledMapObjectLoadingTest](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java)

MobiDevelop: I just took a look - the global flip for y happens earlier in the method. The flip for EllipseMapObject (and friends) is to re-orient the object when accounting for the height of it (and the fact Rectange, Ellipse are bottoem-left origin) - if I am reading/recalling correctly.

lazynessmind: > height variable in this context is the height of the object itself, which is zero for a point. That's why flipping is not necessary here.
> About tests, i agree, it would be better to add it to the existing one: [TiledMapObjectLoadingTest](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectLoadingTest.java)

Done in 23a405b

mgsx-dev (approved):

PR #6656, opened 2021-09-20
Add gdx-natives to desktop backends for #6369

PokeMMO:
* Will have to figure out how to do this better when we do publishing from gradle.


NathanSweet:
Confirmed it works, cheers!

PR #6661, opened 2021-09-25
Improved gha

SimonIT:
Checks direct pushes to master for gwt and spotless issues. It fixes also formatting issues automatically in PRs

### DIFF
--- a/.github/workflows/build-snapshot.yml
+++ b/.github/workflows/build-snapshot.yml
@@ -392,7 +392,7 @@ jobs:
         with:
           java-version: 1.8
       - name: Build Runnables
-        run: ./gradlew clean fetchNatives buildRunnables extensions:gdx-setup:build
+        run: ./gradlew clean fetchNatives buildRunnables build
 
       - name: Upload artifacts to S3
         if: env.AWS_ACCESS_KEY_ID != null
--- a/.github/workflows/fix-formatting.yml
+++ b/.github/workflows/fix-formatting.yml
@@ -1,0 +1,42 @@
+name: Fix formatting
+
+on: [push]
+
+jobs:
+  fix-formatting:
+    if: ${{ github.repository_owner != 'libgdx' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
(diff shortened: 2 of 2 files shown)


.github/workflows/build-snapshot.yml
```diff
@@ -392,7 +392,7 @@ jobs:
         with:
           java-version: 1.8
       - name: Build Runnables
-        run: ./gradlew clean fetchNatives buildRunnables extensions:gdx-setup:build
+        run: ./gradlew clean fetchNatives buildRunnables build
```
tommyettinger, line 395: I'm glad this builds all the runnables, and not just gdx-setup!

.github/workflows/build-pullrequest.yml
```diff
@@ -28,7 +28,15 @@ jobs:
...
-
+      - name: Apply formatter
+        run: ./gradlew spotlessApply
+      - name: Commit formatting changes
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add .
+          git commit -m "Apply formatter" -a
+        continue-on-error: true
```
tommyettinger, line 39: This will still show if there is an error in spotless, but will let the build finish, right? Hopefully spotless won't encounter errors like that... I'm a little worried about cases where spotless might produce output that it itself considers invalid, so each time it runs it changes some section more and more. We've already noticed some oddities in indentation.

tommyettinger (approved):
I don't really understand GHA very well, but this looks good to me.

.github/workflows/build-pullrequest.yml
SimonIT: That's a misunderstanding. The `continue-on-error` is only there because making a commit without changes, aka correct formatted files, would let the job fail because of exit code 1

.github/workflows/build-pullrequest.yml
```diff
@@ -28,7 +28,20 @@ jobs:
...
+      - name: Push formatting changes
+        uses: ad-m/github-push-action@master
```
PokeMMO, line 41: Can you please pin a specific commit instead of blindly trusting master? If the action would be compromised so would our repository.

PR #6662, opened 2021-09-28
Remove unnecessary ThreadUtils class

obigu:


mgsx-dev:
Few questions:
* did you forgot ThreadUtils emu class?
* will this work with GWT?
* what about user code using ThreadUtils? this is a breaking change then.

intrigus:
Isn't this a rather risky/breaking change that has minimal/zero reward?

obigu:
Damn, always forget about GWT

PR #6665, opened 2021-09-29
Fix #6664: HTTP request cancels the request properly.

### LINKED ISSUE #6664: HTTP request cancel does not cancel the request
#### Issue details
Cancelling of HTTP request using `cancelHttpRequest` method in `Net` only cancels invoking of listeners, but the actual request is still performed.

#### Reproduction steps/code
```java
HttpRequest httpRequest = /* ... */;
Gdx.net.sendHttpRequest(httpRequest, null);
Gdx.net.cancelHttpRequest(httpRequest);
```
This code will not cancel the downloading that was not yet performed.

1.10.0

- [x] Android
- [x] iOS
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] macOS

dzolo:
`NetJavaImpl#cancelHttpRequest` cancels pending downloads by canceling the submitted executor service HTTP request task.

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
+++ b/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java
@@ -25,6 +25,7 @@ import java.net.URL;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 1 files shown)


crykn:
This only works if the request hasn't yet started connecting (for instance when cancelling right after sending the request). 

However, when canceling after the connection process has already begun (which is the more realistic use case imho), it won't do anything. To try this, just comment out `removeFromConnectionsAndListeners(httpRequest)` in `cancelHttpRequest` (so you'll still receive listener events even after "cancelling") and then call a slow website (e.g., http://www.fakeresponse.com/api/?sleep=1). After a couple seconds the request times out and `HttpResponseListener#failed` is called.

dzolo:
Since the `HttpURLConnection` is using blocking socket implementation there is no way to interrupt the `connect()` in `sendHttpRequest`. Calling `Future#cancel(true)` does not to the job, since `HttpURLConnection` connect cannot be interrupted. To fix this behavior the `HttpURLConnection` would have to be replaced with some other implementation (NIO). But this is a big change (and probably not welcome).

The bugfix is still very useful from my point of view as it cancels all pending requests. For me personally I found this very useful when loading map tiles together with `maxThreads` set (many queued requests).  On map zoom change it allows me to cancel downloading of tiles which are not longer needed.

PR #6670, opened 2021-10-04
gdx-lwjgl3-angle extension to add rendering support through ANGLE

badlogic:
# What and why

This PR adds and extension to support rendering through [ANGLE](https://github.com/google/angle) instead of emulating OpenGL ES 2.0 via the desktop OpenGL APIs. On Windows, DirectX 9 or 11 is used to render. On Linux calls are passed through to the desktop OpenGL implementation or, if available, a native OpenGL ES and EGL implementation. On macOS, OpenGL ES 2.0 is emulated via Metal.

This extension is especially useful on macOS, where Apple will eventually remove OpenGL entirely. And even if not, their OpenGL-on-top-of-Metal implementation is already a buggy mess, especially on Apple Silicon.

On Windows, the extension can be useful to increase compatibility with hardware that has shoddy OpenGL drivers.

ANGLE is used extensively by Chrome and Firefox on Windows and is thus battle tested and very stable. It is also a free alternative to [MoltenGL](https://moltengl.com/moltengl/) on macOS. 

# Usage
* Add `gdx-lwjgl3-angle` as a dependency to your gdx-lwjgl3 desktop project.
* Call `config.setOpenGLEmulation(GLEmulation.ANGLE_GLES20, 0, 0)` on your `Lwjgl3ApplicationConfiguration` instance before creating your `Lwjgl3Application`.

# Caveats
- ~Doesn't work on macOS at the moment, due to dylib loading shenanigans. Will fix and amend the PR.~
- ~For some reason, the frame rate drops when multiple windows are opened on macOS.~ There's either a bug in ANGLE or GLFW, in any case, there's a workaround, see https://github.com/badlogic/libgdx/commit/2c32d11dcc94ada252f0e1480798d9a964d5f677#diff-424a1076f33ca5f397d61490a4f2e1500e7a1287d24cdccf72f999eea0c5270bR69
- Wasn't tested on Linux, but should work (TM)
- Does only support OpenGL ES 2.0, as the macOS ANGLE Metal backend only supports OpenGL ES 2.0 at this point in time.
- Requires LWJGL 3.3.0-SNAPSHOT, as that contains a more recent GLFW version and binds ANGLE specific constants.
- The ANGLE binaries are commited directly into the repository. They are taken from https://github.com/xiaozhuai/angle_builds and can be updated as needed. Building them ourselfs as part of the libGDX CI adds significantly to the build time (30-50 minutes per platform).
- Sorry for the weird ass commit list, dunno what went wrong. The modified files are only concerning the actual ANGLE extension tho. May the power of Git compell you ¯\_(ツ)_/¯


tommyettinger:
This definitely seems good to have as we go forward. I think this probably shouldn't be merged while it still has to depend on an LWJGL3 snapshot, because I'm pretty sure Maven Central doesn't permit snapshot dependencies in releases. Apple Silicon support is certainly important to some users, though probably more-so to developers making iOS apps and debugging on desktop than the tiny amount of actual game-playing end-users. Could the ANGLE binaries be available via package manager in the CI system? Would they be too old if they were available? I think it's somewhat important that the whole working library can potentially be built from libGDX sources, even if in practice we are able to use pre-built binaries from another repo -- if something happens to external repositories, we just need some kind of fallback.

Is there any kind of performance gain or drop to be expected from the switch to DirectX on Windows? I'd certainly expect some performance gain on MacOS, once it works, since we're avoiding emulation.

badlogic:
I have a foro of angle_builds, which pulls in the ANGLE GitHub mirror. I can fork the ANGLE mirror as well, which should ensure longevity. I don't think ANGLE is in any package manager repo, and I wouldn't want to use those binaries.

Yes, there is likely a performance penalty. I haven't quantified it, non of our tests showed performance regressions. There is one issue regarding multi-window apps when using the Metal backend. For some reason the FPS get halfed. Haven't looked into it yet.

For iOS on device apps we'll need MetalANGLE, which should be simple to do.

PR #6672, opened 2021-10-05
gdx-lwjgl3-angle extension to add rendering support through ANGLE

badlogic:
([Original discussion](https://github.com/libgdx/libgdx/pull/6670), redid PR so commits are clean)

# What and why

This PR adds and extension to support rendering through [ANGLE](https://github.com/google/angle) instead of emulating OpenGL ES 2.0 via the desktop OpenGL APIs. On Windows, DirectX 9 or 11 is used to render. On Linux calls are passed through to the desktop OpenGL implementation or, if available, a native OpenGL ES and EGL implementation. On macOS, OpenGL ES 2.0 is emulated via Metal.

This extension is especially useful on macOS, where Apple will eventually remove OpenGL entirely. And even if not, their OpenGL-on-top-of-Metal implementation is already a buggy mess, especially on Apple Silicon.

On Windows, the extension can be useful to increase compatibility with hardware that has shoddy OpenGL drivers.

ANGLE is used extensively by Chrome and Firefox on Windows and is thus battle tested and very stable. It is also a free alternative to [MoltenGL](https://moltengl.com/moltengl/) on macOS. 

# Usage
* Add `gdx-lwjgl3-angle` as a dependency to your gdx-lwjgl3 desktop project.
* Call `config.setOpenGLEmulation(GLEmulation.ANGLE_GLES20, 0, 0)` on your `Lwjgl3ApplicationConfiguration` instance before creating your `Lwjgl3Application`.

# Caveats
- ~Doesn't work on macOS at the moment, due to dylib loading shenanigans. Will fix and amend the PR.~
- ~For some reason, the frame rate drops when multiple windows are opened on macOS.~ There's either a bug in ANGLE or GLFW, in any case, there's a workaround, see https://github.com/badlogic/libgdx/commit/2c32d11dcc94ada252f0e1480798d9a964d5f677#diff-424a1076f33ca5f397d61490a4f2e1500e7a1287d24cdccf72f999eea0c5270bR69
- Wasn't tested on Linux, but should work (TM)
- Does only support OpenGL ES 2.0, as the macOS ANGLE Metal backend only supports OpenGL ES 2.0 at this point in time.
- Requires LWJGL 3.3.0-SNAPSHOT, as that contains a more recent GLFW version and binds ANGLE specific constants.
- The ANGLE binaries are commited directly into the repository. They are taken from https://github.com/xiaozhuai/angle_builds and can be updated as needed. Building them ourselfs as part of the libGDX CI adds significantly to the build time (30-50 minutes per platform).


badlogic:
LWJGL 3.3.0 is out. I'll publish the ANGLE binaries as a separate package to Maven on which gdx-lwjgl3-angle will depend. They won't need uodating often anyways. Then I shall !erge this glorious PR.

PokeMMO:
What is the purpose/reasoning for the seperate ANGLELoader class over SharedLibraryLoader?

badlogic:
shared lib suffixes is one reason, but this is the main reason:
https://github.com/libgdx/libgdx/pull/6672/files#diff-06240c617472b87c94e1b16cee91ea906828e974b830bb64af6d26a3f1a672b9R198

badlogic:
I've addressed all issues above and rebased the fork on the current master. The ANGLE natives are now stored here: https://github.com/libgdx/gdx-angle-natives. They are fetched from there through `fetchNatives` and placed in `extensions/gdx-lwjgl3-angle`.

lyze237:
Stupid question: Doesn't the library need to be added in `publish.gradle`?
At least when I'm trying to build it myself to mavenLocal I can't figure it out without that.

Something like this:
idea64_G8bHED9Ekc

badlogic:
Yes, @pokeMMO already did that PR was merged a few hours ago, pull the
latest changes.

PR #6678, opened 2021-10-12
Fix ParticleEditor freezing and crashing

### LINKED ISSUE #6584: 2D Particle Editor freezes after Window loses then regains focus
My environment:
Intel Core i7-7700 CPU, 32.0 GB RAM, Windows 10 Pro 19041.1052 with Java(TM) SE Runtime Environment (build 1.8.0_291-b10)

This also occurs on team member's Windows 10 machine with Java build 1.8.0_291-b10 and also on my machine with Java 11.0.10.

When running the latest downloadable jar of the 2D particle editor, the program becomes unresponsive after the windows loses focuses and then regains it. No log output occurs so there is no stack trace to show.

To reproduce: Launch the jar from command-line or double-clicking. Click on a different window, click on editor again.

jaeheonshim:
The ParticleEditor crashed if you unfocused it and refocused it. For example, clicking out of the ParticleEditor window and then clicking back in would cause the ParticleEditor to freeze and not respond.

The reason is due to the `LwjglCanvas`. I suspect it has to do with the line `Display.processMessages();`. To fix the issue, in ParticleEditor I used a normal Java AWT Canvas, and created a new LwjglApplication passing the Canvas as an argument. This fixes the issue, as far as I can tell;

Tested on Windows machine, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz 

Fixes #6584

### DIFF
--- a/.github/workflows/fix-formatting.yml
+++ b/.github/workflows/fix-formatting.yml
@@ -4,7 +4,6 @@ on: [push]
 
 jobs:
   fix-formatting:
-    if: ${{ github.repository_owner != 'libgdx' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


obigu:
I couldn't reproduce with Windows 10 Java(TM) SE Runtime Environment (build 1.8.0_202) but could with AdoptopenJDK 11.0.11.9. The app seems to run as expected with the fix. Thanks!!

PR #6681, opened 2021-10-14
Add new MetalANGLE backend (ios/robovm)

Berstanio:
OpenGLES was officially deprecated by Apple a few years ago. The succesor should be Metal. But for the cross platform design of libGDX it is hard to switch to another rendering backend only on one platform.
And here comes ANGLE into play. ANGLE is a project by google that aims to translate platform independent OpenGL spec conform calls to platform specific graphic api calls. One library to rule them all, Metal, Vulkan, DirectX etc. with just OpenGL (quote was stolen).
And this is what this backend does, implementing ANGLE with a Metal backend.
To not have the burden to maintain two mobivm backends, @PokeMMO had the great idea and implemented a code generator that takes the "normal" backend and constructs the MetalANGLE backend from this. This works, because the difference is mainly only in a few binding classes.
The backend also directly ships the MetalANGLE binarys, so you need a mac to build the backend. Or make sure to run "fetchNatives" beforehand. 
To use this backend just replace the standard robovm backend with this one.
To update the backend run the gradle tasks "generate" and "spottlessApply".
The backend also autoupdates itself on every push.

This pr implements @kakashidinho really great work "MetalANGLE" (https://github.com/kakashidinho/metalangle) into libGDX to as a seperate robovm backend. It also adds MetalANGLE bindings.
Much of the discussion/tests about MetalANGLE was in this pr https://github.com/libgdx/libgdx/pull/6593.
#### Thinks that still need to be discussed/implemented
- ~How to fulfill ANGLE license conditions. ANGLE runs under BSD-3 license and libGDX under Apache 2 and iirc they are compatible so it should be fairly easy to do. Maybe it is already done because there are other pr's also related to ANGLE~ (THIRDPARTY file need to be further extended and integrated into build processes)
- ~To have a singe MetalANGLE static lib I remove from the ios sim lib the arm64 slice. This shouldn't have any downsides because the simulator always uses x86_64 arch, right?~
- ~How to effectively set the "ANGLE_DEFAULT_PLATFORM=gl" env variable in the simulator for MacOS in vm's. I exported it in the Terminal and then run xcrun from there, but maybe this is a bit annoying for regular use.~ (Will be implemented in newer version of MetalANGLE)
- ~gha currently needs to run "mvn validate" to fetch the MetalANGLE libs. Maybe this should be better done in gradle.~
- _It needs to be validated whether the gha correctly picks up the MetalANGLE framework into the natives.zip._ (The logs at https://github.com/Berstanio/libgdx/runs/3920755152 are clearly hinting that it happens correctly)
- ~Gradle build should be supported~
- ~Fix issue with MetalANGLE framework on real devices~
- Use MetalANGLE dynamic framework or static framework?
    - pro static:
        - fails on linking when Metal is not specified as framework  [see](https://github.com/libgdx/libgdx/pull/6681#issuecomment-964522836), [see](https://github.com/libgdx/libgdx/pull/6681#issuecomment-965016387)
        - produces smaller app size (5mb lesser increase)
    - contra static:
        - the libs are enormous ~2GB (is tracked in a MetalANGLE issue, but its unclear when and how much the lib size can get reduced)
        - console is spammed with symbol warnings (maybe fixable and maybe related to the point above)
     - pro/contra dynamic (just the pros/cons of static inverted)
- The "BitmapFontAtlasRegionTest" fails on simulator (https://github.com/kakashidinho/metalangle/issues/68)

For now the pr will got with the MetalANGLE framework, but this can easily be switched later

This fixes https://github.com/libgdx/libgdx/issues/5251, I think.


Berstanio:
Hey @kakashidinho,
I'm currently investigating the issue I wrote about in my previous post. Do you know what needs to happen inside MetalANGLE so that the described issue occurs? There is no log message (with MetalANGLE debug build) that gives any hints. Or is there a even more verbose compile option than just the debug build? I also sadly can't really debug the framework internally. 
It would be great if you could provide any starting point for me!

kakashidinho:
> Hey @kakashidinho,
> I'm currently investigating the issue I wrote about in my previous post. Do you know what needs to happen inside MetalANGLE so that the described issue occurs? There is no log message (with MetalANGLE debug build) that gives any hints. Or is there a even more verbose compile option than just the debug build? I also sadly can't really debug the framework internally.
> It would be great if you could provide any starting point for me!

It could be caused by wrong glGetString symbol resolution. If you link against both MetalANGLE and OpenGLES.framework the latter’s symbols would be incorrectly used by compiler/linker. Either remove OpenGLES.framework from the linking libs (MetalANGLE already links against OpenGLES by itself) or move MetalANGLE higher in the list of linking libs so that linker will pick it symbols before anything else)

Berstanio:
> It could be caused by wrong glGetString symbol resolution. If you link against both MetalANGLE and OpenGLES.framework the latter’s symbols would be incorrectly used by compiler/linker. Either remove OpenGLES.framework from the linking libs (MetalANGLE already links against OpenGLES by itself) or move MetalANGLE higher in the list of linking libs so that linker will pick it symbols before anything else)

Ah I understand! In my first tests with MetalANGLE I needed to keep OpenGLES linking for some reason, but now I just removed it and everything works fine. Many thanks!

To the libGDX maintainers: The pr is now again ready for review/tests!

obigu:
This is great, thanks!

I've tested the PR and `fetchNatives` doesn't seem to download MetalANGLE framework so it complains that MetalANGLE framework is not found when trying to run. If I execute manually the `downloadMetalANGLE` (after having run `fetchNatives` it downloads it with this folder structure:

Screenshot 2021-11-08 at 12 57 54

After that I can run libGDX tests successfully.

Berstanio:
> I've tested the PR and `fetchNatives` doesn't seem to download MetalANGLE framework so it complains that MetalANGLE framework is not found when trying to run.

Yes this is currently intended. I thought the procedure is the following: The gha will run ```jnigenBuildIOS``` which will afterwards trigger ```downloadMetalANGLE```. Then everything will be packed into the natives.zip (also the framework). Then running ```fetchNatives``` will properly unpack everything, if then ```downloadMetalANGLE``` would be run again it would override the previously unpacked frameworks.
So the issue will be gone when this is merged.

> After that I can run libGDX tests successfully.

Thats great!

obigu:
> > I've tested the PR and `fetchNatives` doesn't seem to download MetalANGLE framework so it complains that MetalANGLE framework is not found when trying to run.
> 
> Yes this is currently intended. I thought the procedure is the following: The gha will run `jnigenBuildIOS` which will afterwards trigger `downloadMetalANGLE`. Then everything will be packed into the natives.zip (also the framework). Then running `fetchNatives` will properly unpack everything, if then `downloadMetalANGLE` would be run again it would override the previously unpacked frameworks. So the issue will be gone when this is merged.

Ok, I see.

@Tom-Ski Would we be able to test the whole thing by merging this to a branch first?

Tom-Ski:
The entire build flow with actions?

obigu:
I don't know, maybe the branch doesn't make sense, the question is more if there's a good way to test the whole build changes before merging it to master so that we can test and even integrate it on our projects.

Tom-Ski:
Perhaps we just use snapshot for that? We can have a release before merge perhaps and then have the potentially volatile snapshot?

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -566,19 +573,20 @@ public void setCursor (Cursor cursor) {
...
-		public void update (GLKViewController controller) {
+		public void mglkViewControllerUpdate (MGLKViewController controller) {
 			IOSGraphics.this.update(controller);
 		}
 
-		@Override
-		public void willPause (GLKViewController controller, boolean pause) {
+		// Consider to remove since IOSGraphics.this.willPause is empty.
+		// @Override
+		public void willPause (MGLKViewController controller, boolean pause) {
```
obigu, line 584: Yeah, agree, we can remove willPause

obigu (commented):
From a first look just added some minor comments.

I think we can also update Gdx Setup (/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/ios/robovm.xml) and remove `OpenGL` and add MetalAngle dependencies.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -566,19 +573,20 @@ public void setCursor (Cursor cursor) {
 	public void setSystemCursor (SystemCursor systemCursor) {
 	}
 
-	private class IOSViewDelegate extends NSObject implements GLKViewDelegate, GLKViewControllerDelegate {
+	private class IOSViewDelegate extends NSObject implements MGLKViewDelegate, MGLKViewControllerDelegate {
 		@Override
-		public void update (GLKViewController controller) {
+		public void mglkViewControllerUpdate (MGLKViewController controller) {
```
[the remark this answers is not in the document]
Berstanio: I wasn't sure whether mobivm somewhere uses the method name of the bindings, but probably it doesn't, right?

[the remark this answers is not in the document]
obigu: As long as the `@Method` annotation point to the right `selector`, the java name can be anything.

Something like this is ok:

```
@Method(selector = "mglkView:drawInRect:")
void draw(MGLKView view, @ByVal CGRect rect);
```

Berstanio:
> I think we can also update Gdx Setup (/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/ios/robovm.xml) and remove `OpenGL` and add MetalAngle dependencies.

Done!
Just for clarification, adding MetalANGLE as framework is not needed because it is automatically added by the robovm.xml shipped with the ios natives. What should be done is pulling the OpenGLES framework linking because of the linking issues kakashidinho mentioned that could occur, depending on which robovm.xml file MobiVM evaluates first.

obigu:
> > I think we can also update Gdx Setup (/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/ios/robovm.xml) and remove `OpenGL` and add MetalAngle dependencies.
> 
> Done! Just for clarification, adding MetalANGLE as framework is not needed because it is automatically added by the robovm.xml shipped with the ios natives. What should be done is pulling the OpenGLES framework linking because of the linking issues kakashidinho mentioned that could occur, depending on which robovm.xml file MobiVM evaluates first.


Maybe we should remove the `MetalANGLE` and `Metal` from the `robovm.xml` shipped with the natives so that it doesn't get magically added and instead require apps to declare it (Gdx Setup would have it by default for new apps). Currently we don't add any other framework such as OpenGL on a libGDX embedded `robovm.xml` file even though it's also true that I think Apple frameworks don't have to be explicitly declared. The reasoning behind is that it would make it clearer for users to know which dependencies their app has and what backend their app is actually using and, maybe more important, it would force users upgrading from previous libGDX versions to update their `robovm.xml` files (otherwise it'd complain MetalANGLE is not found) avoiding the described random order-dependant linking errors that make the whole thing a bit fragile. What do you think?

Berstanio:
I also thought about that, for linking against Metal that would be easily possible. But I'm a bit unsure if we should do that also for MetalANGLE. MetalANGLE gets shipped inside the jar, the robovm.xml inside sets up the framework paths and if then the user needs to add this as a framework that seems a bit counterintuitive to me. The user would have no clue where the framework they add comes from and how the linking works since it isn't a standard build in framework. Do you know what I mean?

obigu:
> I also thought about that, for linking against Metal that would be easily possible. But I'm a bit unsure if we should do that also for MetalANGLE. MetalANGLE gets shipped inside the jar, the robovm.xml inside sets up the framework paths and if then the user needs to add this as a framework that would be acounterintuitive, I think. The user would have no clue where the framework they add comes from and how the linking works since it isn't a standard build in framework. Do you know what I mean?

I do. That would be perfect but we need to confirm that it will complain if `Metal` is not explicitly added in your `robovm.xml`, especially in the case `OpenGL` is present.

Berstanio:
> I do. That would be perfect but we need to confirm that it will complain if `Metal` is not explicitly added in your `robovm.xml`, especially in the case `OpenGL` is present.

I did some tests on this. Sadly with the current framework it wouldn't complain. One alternative would be to use a static framework instead of the dynamic one. The massiv downside of that would be that they would take ~2.5GB disk space (not on the ios device, but on the mac) and it would be needed to take overall more care that all needed frameworks are specified. Another alternative could be to somehow call a useless Metal method (e.g. MTLCopyAllDevices) in e.g. iosgl to create there the needed reference to Metal to fail at linking if Metal isn't specified, but that would probably be very hacky and complicated to do since iosgl is only a c++ file, not a ObjC one.
Another downside of removing MetalANGLE from the shipped robovm.xml would be, that it would only fail at runtime with a "Allocating missing native class" error. 

But probably removing the MetalANGLE linking from the robovm.xml is with the downsides still the best method to make sure the users properly read the changelogs. Or are there any other ideas/opinions?

obigu:
Ok, let's keep it for the moment and we can make a decision to remove it later.

I've found an issue when testing setting `useMetal` to `false`:
```
java.lang.UnsatisfiedLinkError: com.badlogic.gdx.backends.iosrobovm.IOSGraphics.disableMetal()V
	at com.badlogic.gdx.backends.iosrobovm.IOSGraphics.disableMetal(Native Method)
	at com.badlogic.gdx.backends.iosrobovm.IOSGraphics.<init>(IOSGraphics.java:91)
	at com.badlogic.gdx.backends.iosrobovm.IOSApplication.createGraphics(IOSApplication.java:167)
```

obigu:
Ok, yeah, it looks like `fetchNatives` had overwritten stuff. It runs, not sure if it's on OpenGL though, is there a way to check?

There's the small problem when testing now that if you clean native libraries, `jnigenBuildIOS` doesn't build `/ios32/libObjectAL.a` and it doesn't run without it. What I've done is run `fetchNatives` first to grab a copy of `libObjectAL.a` and then manually copy it when needed.

Thanks!

Berstanio:
> Ok, yeah, it looks like `fetchNatives` had overwritten stuff. It runs, not sure if it's on OpenGL though, is there a way to check?

In the next version of MetalANGLE I added a log line, but currently you have to look at the opening animation. If it just appears, it is OpenGLES, if it transitions from the left top corner, it is Metal.

> There's the small problem when testing now that if you clean native libraries, `jnigenBuildIOS` doesn't build `/ios32/libObjectAL.a` and it doesn't run without it. What I've done is run `fetchNatives` first to grab a copy of `libObjectAL.a` and then manually copy it when needed.

You I also noticed that annoying issue, but I wasn't sure whether is is intended that `jnigenBuildIOS` doesn't build ObjectAL or not so I left it. My quick workaround was a bash file that first runs `jnigenBuildIOS` and then the bash build script in the robovm backend dir.

Berstanio:
I added a log line showing, which render backend is effectively used. The log line in MetalANGLE I mentioned earlier would only print something if there is a forced fallback because Metal is not supported.

CoderBaron:
> In the next version of MetalANGLE I added a log line, but currently you have to look at the opening animation. If it just appears, it is OpenGLES, if it transitions from the left top corner, it is Metal.

What's the reason for that opening animation? Any way to remove that? 
It breaks the seamless transition from splash screen to first rendered scene and Apple usually doesn't like that (nor do I).

CoderBaron:
I've tried this in one of my apps (2d game) and generally it works very good. No graphical differences compared to OpenGLES (apart from the before mentioned opening animation) and a huge performance boost (at least 5 times faster on simulator and several frames faster on iPhone 13).

However I've noticed a huge issue on simulator that after some time the rendering freezes, which makes the whole app unresponsive. The update/render functions don't get called and touch input doesn't work anymore. However other iOS UI views on top of the MGLKViewController are correctly rendered and animated. I couldn't reproduce it yet on a real device.

Berstanio:
Hey,
many thanks for testing it! The opening animation can't currently be disabled. But as far as I know this is not a intended behavior and it is covered here: https://github.com/kakashidinho/metalangle/issues/49
About the simulator issue: How long does it take to freeze? Does it also freeze if you set "useMetal = false" in the IOSApplicationConfiguration object?

Berstanio:
To the libGDX contributors: Should I include the MetalANGLE headers through a submodule similiar to how it was done with freetype?

CoderBaron:
> Hey, many thanks for testing it! The opening animation can't currently be disabled. But as far as I know this is not a intended behavior and it is covered here: [kakashidinho/metalangle#49](https://github.com/kakashidinho/metalangle/issues/49) About the simulator issue: How long does it take to freeze? Does it also freeze if you set "useMetal = false" in the IOSApplicationConfiguration object?

When I switch to OpenGLES with "useMetal = false" it doesn't freeze. 
I tested using iPad mini 6th Gen Simulator with iOS 15.2 on a M1 Mac mini. When using Metal the app freezes after around 10-60 minutes.

Binkie86:
> This pr implements @kakashidinho really great work "MetalANGLE" (https://github.com/kakashidinho/metalangle) into libGDX to replace OpenGLES. It also adds MetalANGLE bindings for robovm backend.
> 
> Much of the discussion/tests about MetalANGLE was in this pr https://github.com/libgdx/libgdx/pull/6593.
> 
> #### Thinks that still need to be discussed/implemented
> 
> - ~How to fulfill ANGLE license conditions. ANGLE runs under BSD-3 license and libGDX under Apache 2 and iirc they are compatible so it should be fairly easy to do. Maybe it is already done because there are other pr's also related to ANGLE~ (THIRDPARTY file need to be further extended and integrated into build processes)
> 
> - ~To have a singe MetalANGLE static lib I remove from the ios sim lib the arm64 slice. This shouldn't have any downsides because the simulator always uses x86_64 arch, right?~
> 
> - ~How to effectively set the "ANGLE_DEFAULT_PLATFORM=gl" env variable in the simulator for MacOS in vm's. I exported it in the Terminal and then run xcrun from there, but maybe this is a bit annoying for regular use.~ (Will be implemented in newer version of MetalANGLE)
> 
> - ~gha currently needs to run "mvn validate" to fetch the MetalANGLE libs. Maybe this should be better done in gradle.~
> 
> - _It needs to be validated whether the gha correctly picks up the MetalANGLE framework into the natives.zip._ (The logs at https://github.com/Berstanio/libgdx/runs/3920755152 are clearly hinting that it happens correctly)
> 
> - ~Gradle build should be supported~
> 
> - ~Fix issue with MetalANGLE framework on real devices~
> 
> - Use MetalANGLE dynamic framework or static framework?
> 
>     - pro static:
> 
>         - fails on linking when Metal is not specified as framework  [see](https://github.com/libgdx/libgdx/pull/6681#issuecomment-964522836), [see](https://github.com/libgdx/libgdx/pull/6681#issuecomment-965016387)
> 
>         - produces smaller app size (I will add a exact amount in the next days)
> 
>     - contra static:
> 
>         - the libs are enormous ~2GB (is tracked in a MetalANGLE issue, but its unclear when and how much the lib size can get reduced)
> 
>         - console is spammed with symbol warnings (maybe fixable and maybe related to the point above)
> 
>      - pro/contra dynamic (just the pros/cons of static inverted) 
> 
> 
> 
> This fixes https://github.com/libgdx/libgdx/issues/5251, I think.

Binkie86:
> Mobivm has workarounds for xcframework via frameworkPaths. So its doable if we construct them in compatible formatting

CoderBaron:
> When I switch to OpenGLES with "useMetal = false" it doesn't freeze. I tested using iPad mini 6th Gen Simulator with iOS 15.2 on a M1 Mac mini. When using Metal the app freezes after around 10-60 minutes. Well actually the app still runs, I can see the banner ad view still animating, but the MGLKViewController freezes and doesn't update anymore.

CORRECTION: It also freezes when using OpenGLES but it takes 1-2 days before it does. It doesn't seem to happen with libGDX 1.10.0 so far but I can't tell if it's just my app, libGDX or your PR. Best would be if you try to run the a libGDX metal app on simulator for a few hours and see if the same happens to you.

Berstanio:
> CORRECTION: It also freezes when using OpenGLES but it takes 1-2 days before it does. It doesn't seem to happen with libGDX 1.10.0 so far but I can't tell if it's just my app, libGDX or your PR. Best would be if you try to run the a libGDX metal app on simulator for a few hours and see if the same happens to you.

Okay many thanks! I will try to reproduce it! But I'm a bit busy currently, so it will maybe take some time.

obigu:
> To the libGDX contributors: Should I include the MetalANGLE headers through a submodule similiar to how it was done with freetype?

Considering this comment probably not (at least for the moment) https://github.com/libgdx/libgdx/pull/6550#commitcomment-63683250.

Berstanio:
> I tested using iPad mini 6th Gen Simulator with iOS 15.2 on a M1 Mac mini.

I tried to reproduce it with the same setup and the gdx-iosrobovm tests and another app. Both worked still fine after half a day. Could you try to confirm that the bug also happens when running the "Basic3DSceneTest"  from the gdx-tests-iosrobovm? If not, would it be possible to provide a reproducer?

badlogic:
I would not put to much thought into things falling on their nose on the simulator. The simulator is notoriously buggy when it comes to both Metal and OpenGL. Since this PR allows to use both Metal and OpenGL, I see no reason to not merge this.

The only thing I am concerned about is the fact, that this will now automatically add the Metal Angle libs to a build, bloating the resulting app bundle. Do we have any way to exclude that if it's not wanted?

Berstanio:
> The only thing I am concerned about is the fact, that this will now automatically add the Metal Angle libs to a build, bloating the resulting app bundle. Do we have any way to exclude that if it's not wanted?

No, not really except having two backends. But the resulting apps with mobivm are already pretty large, I'm not sure whether the additional 5-20mb are making a big difference.

Berstanio:
> Yeah I don't really like bundling both of these when its going to be a choice at compile time what the dev wants to use. We should distribute different versions with classifiers probably.

Hm not sure if I understand you right, but MetalANGLE needs to be bundled after this pr. It is necessary for the bindings.

Tom-Ski:
I was under the impression that the GLES backend would remain, and we would provide both. GLES being what we already had, and the other through MetalAngle. Not, using MetalAngle as a shim to delegate to GLES.

This being the case in this PR as it stands, I don't really see any reason to support GLES through MetalAngle. It's just bloating the binary size at the end by going through a heavy MetalAngle lib as a shim instead of using our current GLES implementation. 

My concern was just with issues in the entire MetalAngle stack, which is why I wanted to keep our current GLES as an option, but I don't think it is worth doing it this way.  Is the fallback to GLES a pure passthrough? Doesn't that require Metal, MetalAngle, and also GLES frameworks to be statically linked?

Berstanio:
MetalANGLE already provides a backend for Metal and a backend for OpenGLES. The default backend is Metal and OpenGLES can be used as a "fallback" if the device doesn't support Metal or as a choice via a env variable. 
There is also no need for linking against OpenGLES framework, because MetalANGLE is already linked against it.

Currently there is a request to have a build without the OpenGLES backend, but I think having the fallback backend is for the beginning the best option.

And yes, if I remember correctly the OpenGLES backend is more or less pure passthrough.

Tom-Ski:
Yeah, so people that want to continue using GLES, just have to deal with the bloat from the MetalANGLE libs that we cannot strip as we are now using it as a passthrough. 

Do we know what devices do not support this? 

> Currently there is a request to have a build without the OpenGLES backend, but I think having the fallback backend is for the beginning the best option.
Do you mean there is a request on the MetalANGLE side? Because for us, there is no GLES backend, it's just transparent support through MetalANGLE

Do you know what the size difference is in the final binaries? For our tests for example.

obigu:
@Tom-Ski 
The binary size was also my main concern but I was quite surprised to discover the difference is not that big, have a look at these numbers (https://github.com/libgdx/libgdx/pull/6593#issuecomment-937727831). 

More tests may be needed to compare, especially to check the difference of the final size after the AppStore thinning.

The main problem if we want a non MetalAngle backend is that we would probably need 2 separate backends not to make the ios backend a monster module with lots of conditional logic and a pretty complex build. If we also want to keep the MOE option open that would mean 4.

Berstanio:
> Do we know what devices do not support this?

All devices with a ios version < 11 and the simulator when running mac os in a vm.

> Do you mean there is a request on the MetalANGLE side? Because for us, there is no GLES backend, it's just transparent support through MetalANGLE

Yes. MetalANGLE supports OpenGLES as backend and there was a issue created to have a configuration that doesn't include the OpenGLES backend. (https://github.com/kakashidinho/metalangle/issues/71) But I'm not sure how this impacts the final app size since it is not implemented yet.

If the app size is such a big factor it would be probably better to use static libs instead dynamic libs. Than it would be only 10mb more, this is what @obigu showed in his comment. It would be easy to switch, no big deal. The only drawback is the massive size of the static libs. The extra disk space needed would blow up during compilation to ~2gb. I'm not sure whether the saved 10mb for the app size are worth that.

badlogic:
> The binary size was also my main concern but I was quite surprised to discover the difference is not that big, have a look at these numbers (#6593 (comment)).

If Imread these results correctly, then the MetalANGLE binaries add 5mb per arch slice in a release build. That's not humongous at least. It's not ideal either, but I could live with that. Especially if we don't retain the pure GLES code, but pass through Metal in case we don't use the GLES Metal emulation. Dunno. I have an iOS app on the store which is pretty successful. I'd be fine with this.

Tom-Ski:
> Yes. MetalANGLE supports OpenGLES as backend and there was a issue created to have a configuration that doesn't include the OpenGLES backend. (kakashidinho/metalangle#71) But I'm not sure how this impacts the final app size since it is not implemented yet.

Yeah that would be a nice optimization for cases where it's known to be supported on targets.

 5mb overhead IMO is fine. App store stripping should just clear out all the other archs. Worst case scenario if there are problems people face in the wild in some scenarios, we can bring back dedicated backend for 'pure' gles and distribute it under another artifact id or classifier.


> If the app size is such a big factor it would be probably better to use static libs instead dynamic libs. 

Do you know the overhead for dynamic lib case? Size of the static libs is indeed quite a pita.

Berstanio:
I made a few more size tests:

When I e.g. test this app https://github.com/oseph/chigginblazzder I get the following results:

|       | MetalANGLE static | MetalANGLE framework | OpenGLES |
|-------|-------------------|----------------------|----------|
| .app  | 73.5             | 85.3                 | 62.9     |
| .dSYM | 309.5             | 21.9               | 10.6     |
| .ipa  | 28.9              | 31.2                 | 16.3     |

This lines up with the tests I done before and obigu's results.

And this are the results for the gdx-test, if this line is removed: https://github.com/libgdx/libgdx/blob/520d780da60c77603e10d9acf8ed2601adc7ca6c/tests/gdx-tests-iosrobovm/build.gradle#L31
This line would result in including all static libs inside the resulting bundle.

|       | MetalANGLE static | MetalANGLE framework | OpenGLES |
|-------|-------------------|----------------------|----------|
| .app  | 114.8             | 126.6                | 104.1    |
| .dSYM | 317.3             | 29.4                | 29.4     |
| .ipa  | 67.9             | 63.3                | 63.6     |

obigu:
Not sure this is the best place to discuss this but it seems Angle has officially started working on iOS support https://github.com/google/angle/commit/4a5a559d6fad3e04f5c2475fc44fc92227e538dc.

kakashidinho:
Yes, Apple has been doing a lot works since last year on their WebKit repo to support WebGL with an ANGLE fork.
Their ANGLE's WebKit version also has been tested on iOS devices and the source code has been merged to the official ANGLE repo regularly.
However, the official ANGLE repo still hasn't had any Xcode project or build script to build iOS version yet (hence it's still "in progress").
Also, as @Berstanio said both official and Webkit version won't include high level `GLKit` classes' replacements.
Additionally, Apple have done of lot of modifications (such as rewriting a new shader compiler from scratch which might compile shaders faster than MetalANGLE). And some of the changes favor WebGL's spec correctness rather than performance.

So just FYI to keep in mind. It might be worth to wait and use official ANGLE in future.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -20,6 +20,16 @@
 import com.badlogic.gdx.Application;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.LifecycleListener;
+import com.badlogic.gdx.backends.iosrobovm.bindings.metalangle.MGLContext;
+import com.badlogic.gdx.backends.iosrobovm.bindings.metalangle.MGLDrawableColorFormat;
+import com.badlogic.gdx.backends.iosrobovm.bindings.metalangle.MGLDrawableDepthFormat;
```
Tom-Ski, line 25: The changes here and in the other classes under gdx-backend-robovm should be reverted, correct?

Berstanio:
Many thanks to @PokeMMO, the pr now works like this:
- the MetalANGLE backend will be a separate backend. But it won't need to be really maintained, there is a generator which will create all MetalANGLE backend sources based on the  OpenGLES backend sources
- The MetalANGLE will be shipped somehow through a extra package or the backend. As I realised now my current approach doesn't work on non MacOS machines, because they can't build the xcframework. I will fix this.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Berstanio: Ahh yes, pushed the wrong branch!

backends/gdx-backend-robovm-metalangle/res/META-INF/robovm/ios/robovm.xml
```diff
@@ -0,0 +1,10 @@
+<config>
+    <frameworks>
+        <framework>MetalANGLE</framework>
+        <framework>Metal</framework>
```
Tom-Ski, line 4: We need a similar robovm.xml that has only the GLES framework in for GLES robovm, and we should openGL, MetalANGLE, and Metal out of the gdx-setup robovm.xml. (I think I saw it briefly on your other branch originally pushed, but seems to be missing here)

Berstanio: As far as I know OpenGLES is already linked by default and wouldn't need to be specified explicitly. Therefor it seems a bit unnecessary to have a robovm.xml file that only defines a framework, that is not needed.

Tom-Ski: Yeah, I mean we don't want GLES linked on the metal backend, and likewise for the other. So it makes sense to remove this from setup robovm.xml (which has gles already), and then make a robovm.xml for the gles robovm backend that declares it.

I think force link is fine to stay in the setup, as that doesn't change for each backend

Berstanio: I thought since I'm already on it to reduce the setup robovm.xml, it maybe would be better to have also the other framework/forcelink declaration where they come from, namely the gdx native library.

Tom-Ski: I think most belong on the backends, not the native lib, as the backend is the one that is dictating the java classes used, and the dependencies on the frameworks, but that requires duplicating them for both.

PokeMMO (approved):

Tom-Ski (approved):

davidgiga1993:
Awesome, can't wait to test this!
Will it be possible to apply the MetalAngle backend also on desktop (aka osx)?
I'm asking since the related issue has been closed, however opengl is also deprecated on osx

PokeMMO:
You can already use ANGLE on macos via https://github.com/libgdx/libgdx/pull/6672

PR #6683, opened 2021-10-20
Remove Android hideStatusBar config

obigu:
`hideStatusBar` configuration on Android implementation is wrong/incomplete and can interfere with `useImmersiveMode`.

My original plan was to work on our immersive mode support that I thought could be improved (and it can a bit but not that much) but where I've found more issues is with `hideStatusBar`.

This is what the official Android docs say about how to hide the status bar on different Android versions https://developer.android.com/training/system-ui/status.

Independently of what value we set `hideStatusBar` on our `AndroidApplicationConfiguration`, `AndroidApplication` `init()` method always hides the status bar using the `FLAG_FULLSCREEN` flag method described on the Android docs for Android < 16:
```
if (!isForView) {
			try {
				requestWindowFeature(Window.FEATURE_NO_TITLE);
			} catch (Exception ex) {
				log("AndroidApplication", "Content already displayed, cannot request FEATURE_NO_TITLE", ex);
			}
			getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN, WindowManager.LayoutParams.FLAG_FULLSCREEN);
			getWindow().clearFlags(WindowManager.LayoutParams.FLAG_FORCE_NOT_FULLSCREEN);
			setContentView(graphics.getView(), createLayoutParams());
		}
```

Our `hideStatusBar()` implementation does something similar to the Android >= 16 docs mechanism:
```
protected void hideStatusBar (boolean hide) {
		if (!hide) return;

		View rootView = getWindow().getDecorView();
		rootView.setSystemUiVisibility(0x1);
	}
```

First of all we are not setting the recommended flag  `SYSTEM_UI_FLAG_FULLSCREEN` but `0x1` which is `SYSTEM_UI_FLAG_LOW_PROFILE` (https://developer.android.com/reference/android/view/View#SYSTEM_UI_FLAG_LOW_PROFILE), not sure the reason for that but my guess is this is how it was done on legacy OS versions.

So, when tested, as expected, setting the `hideStatusBar` to `false` (default value) has no effect on Android < 16, the system bar never appears because we are ALWAYS setting the flags. The interesting bit is that it doesn't have an effect on newer versions either (tested Android 19 or Android 30) so it looks, for the moment there's backwards compatibility and the flgs are just enough to hide the status bar on all Android versions.

What happens if we make the lines that set the flags conditional to only run on Android < 16 and we test on newer Android versions by setting `hideStatusBar` to `true` expecting that system bars vibisibility is handled by `hideStatusBar()` `setSystemUiVisibility` implementation instead? A disaster, the status bar is all over the place. To properly handle system bar visibility using `setSystemUiVisibility`, as the docs indicate in the `notes`, we would need to do the kind of things we do with Immersive mode by listening to visibility change events. Even in that case, as we know from Immersive mode issues, there are cases in which the system bar would not hide properly on some weird scenarios (like closing the soft keyboard on some devices or combinations of volume key presses with other actions on Android 19 devices).

The other problem is that, even if `hideStatusBar` does nothing, if somebody sets it to `true` and also sets `useImmersiveMode` to `true`, `hideStatusBar` will clear the flags set by immersive mode on `onWindowsFocusChange()`.
```
@Override
	public void onWindowFocusChanged (boolean hasFocus) {
		super.onWindowFocusChanged(hasFocus);
		useImmersiveMode(this.useImmersiveMode);
		hideStatusBar(this.hideStatusBar);
```

I think that even if it's a breaking change, in this case, considering it has not worked properly for a long time, it can lead to issues and nowadays we have immersive mode, the best we can do is get rid of it.

**How to test?** Easiest is running `gdx-tests-android` tests such as `TextAreaTest`.


obigu:
> I'm okay with this. I assumed `hideStatusBar` was set to `true` by default, since libGDX apps by default hide the status bar, but no, it's `false`. Misleading, and it should be covered by immersive mode anyway.
> 
> I'd like `useImmersiveMode` to be `true` by default - I don't think it's standard for games to display the navigation bar, though my device has physical buttons so I wouldn't necessarily know.

Yeah, we can consider that on an Immersive Mode related PR that includes some other improvements.

> 
> I'd also like a way to hide the navigation bar prior to KitKat. Think it was introduced in Honeycomb, but libGDX doesn't support that these days, so just Ice Cream Sandwich and Jelly Bean are affected.

That is a good point. I don't think there are many devices that don't have physical buttons (have a nav bar) and are on API < 19 but there may be some. Testing on a Nexus 4 with API 16 the navigation bar currently doesn't disappear when `hideStatusBar` is set to `true` but the icons become all dots (instead of typical triangle, circle and square). So removing it doesn't really change anything here.

> 
> I'm not doing anything regarding libGDX on Android at the moment, though I was working on a non-libGDX Android game in which I display the navigation and status bars in portrait mode but not in landscape mode. If you add any code, I'll give it a good look.

The idea is for the moment to clean up the stuff that is not working at all and later we can think on fixing/adding visibility related stuff.

Frosty-J:
Been messing about with an Ice Cream Sandwich emulator today. And Gingerbread - I cannot let it go! I think I was mistaken. While the navigation bar can be hidden in 3.0-4.3, it returns when the screen is tapped and stays until the app hides it again due to the lack of immersive mode support. So not very helpful outside of things like video playback and accelerometer-controlled games.

PR #6691, opened 2021-11-06
Improve Quaternion Euler Precision

joshuaskelly:
# Summary
Changes Quaternions to use double precision under the hood when specifically working with Euler angles to improve precision/numerical stability.

## Why?
bad_euler_angles
_Wobbly/broken editor rotation gizmo._

While working on implementing a gizmo system for my editor I spent quite some time struggling with why rotations would degenerate and eventually break. After double and triple checking my math, I finally noticed that round-tripping an Euler value can change as much as 0.1 degrees!

## Floats vs Doubles
I know we typically prefer floats for performance reasons, but I'd argue that here correctness/precision would be preferred.

## My Workaround
good_euler_angles
_Nice and stable rotation gizmo._

I've worked around this by using a helper class to get/set euler angles instead of using the Quaternion instance methods. But being a good open source citizen I thought I'd contribute my findings.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Quaternion.java
+++ b/gdx/src/com/badlogic/gdx/math/Quaternion.java
@@ -129,24 +129,24 @@ public class Quaternion implements Serializable {
 	 * @param roll the rotation around the z axis in radians
 	 * @return this quaternion */
 	public Quaternion setEulerAnglesRad (float yaw, float pitch, float roll) {
-		final float hr = roll * 0.5f;
-		final float shr = (float)Math.sin(hr);
-		final float chr = (float)Math.cos(hr);
-		final float hp = pitch * 0.5f;
-		final float shp = (float)Math.sin(hp);
-		final float chp = (float)Math.cos(hp);
-		final float hy = yaw * 0.5f;
-		final float shy = (float)Math.sin(hy);
-		final float chy = (float)Math.cos(hy);
-		final float chy_shp = chy * shp;
-		final float shy_chp = shy * chp;
-		final float chy_chp = chy * chp;
-		final float shy_shp = shy * shp;
+		final double hr = roll * 0.5f;
+		final double shr = Math.sin(hr);
+		final double chr = Math.cos(hr);
+		final double hp = pitch * 0.5f;
+		final double shp = Math.sin(hp);
+		final double chp = Math.cos(hp);
+		final double hy = yaw * 0.5f;
+		final double shy = Math.sin(hy);
+		final double chy = Math.cos(hy);
+		final double chy_shp = chy * shp;
+		final double shy_chp = shy * chp;
+		final double chy_chp = chy * chp;
+		final double shy_shp = shy * shp;
 
-		x = (chy_shp * chr) + (shy_chp * shr); // cos(yaw/2) * sin(pitch/2) * cos(roll/2) + sin(yaw/2) * cos(pitch/2) * sin(roll/2)
-		y = (shy_chp * chr) - (chy_shp * shr); // sin(yaw/2) * cos(pitch/2) * cos(roll/2) - cos(yaw/2) * sin(pitch/2) * sin(roll/2)
-		z = (chy_chp * shr) - (shy_shp * chr); // cos(yaw/2) * cos(pitch/2) * sin(roll/2) - sin(yaw/2) * sin(pitch/2) * cos(roll/2)
-		w = (chy_chp * chr) + (shy_shp * shr); // cos(yaw/2) * cos(pitch/2) * cos(roll/2) + sin(yaw/2) * sin(pitch/2) * sin(roll/2)
+		x = (float)((chy_shp * chr) + (shy_chp * shr)); // cos(yaw/2) * sin(pitch/2) * cos(roll/2) + sin(yaw/2) * cos(pitch/2) * sin(roll/2)
+		y = (float)((shy_chp * chr) - (chy_shp * shr)); // sin(yaw/2) * cos(pitch/2) * cos(roll/2) - cos(yaw/2) * sin(pitch/2) * sin(roll/2)
+		z = (float)((chy_chp * shr) - (shy_shp * chr)); // cos(yaw/2) * cos(pitch/2) * sin(roll/2) - sin(yaw/2) * sin(pitch/2) * cos(roll/2)
+		w = (float)((chy_chp * chr) + (shy_shp * shr)); // cos(yaw/2) * cos(pitch/2) * cos(roll/2) + sin(yaw/2) * sin(pitch/2) * sin(roll/2)
 		return this;
 	}
 
@@ -160,40 +160,40 @@ public class Quaternion implements Serializable {
 	/** Get the roll euler angle in radians, which is the rotation around the z axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the z axis in radians (between -PI and +PI) */
 	public float getRollRad () {
-		final int pole = getGimbalPole();
-		return pole == 0 ? MathUtils.atan2(2f * (w * z + y * x), 1f - 2f * (x * x + z * z))
-			: (float)pole * 2f * MathUtils.atan2(y, w);
+		final double pole = getGimbalPole();
+		return (float)(pole == 0 ? Math.atan2(2.0 * (w * z + y * x), 1.0 - 2.0 * (x * x + z * z))
+			: pole * 2.0 * Math.atan2(y, w));
 	}
 
 	/** Get the roll euler angle in degrees, which is the rotation around the z axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the z axis in degrees (between -180 and +180) */
 	public float getRoll () {
-		return getRollRad() * MathUtils.radiansToDegrees;
+		return (float)Math.toDegrees(getRollRad());
 	}
 
 	/** Get the pitch euler angle in radians, which is the rotation around the x axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the x axis in radians (between -(PI/2) and +(PI/2)) */
 	public float getPitchRad () {
 		final int pole = getGimbalPole();
-		return pole == 0 ? (float)Math.asin(MathUtils.clamp(2f * (w * x - z * y), -1f, 1f)) : (float)pole * MathUtils.PI * 0.5f;
+		return (float)(pole == 0 ? Math.asin(MathUtils.clamp(2.0 * (w * x - z * y), -1.0, 1.0)) : pole * MathUtils.PI * 0.5);
 	}
 
 	/** Get the pitch euler angle in degrees, which is the rotation around the x axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the x axis in degrees (between -90 and +90) */
 	public float getPitch () {
-		return getPitchRad() * MathUtils.radiansToDegrees;
+		return (float)Math.toDegrees(getPitchRad());
 	}
 
 	/** Get the yaw euler angle in radians, which is the rotation around the y axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the y axis in radians (between -PI and +PI) */
 	public float getYawRad () {
-		return getGimbalPole() == 0 ? MathUtils.atan2(2f * (y * w + x * z), 1f - 2f * (y * y + x * x)) : 0f;
+		return (float)(getGimbalPole() == 0 ? Math.atan2(2.0 * (y * w + x * z), 1.0 - 2.0 * (y * y + x * x)) : 0.0);
 	}
 
 	/** Get the yaw euler angle in degrees, which is the rotation around the y axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the y axis in degrees (between -180 and +180) */
 	public float getYaw () {
-		return getYawRad() * MathUtils.radiansToDegrees;
+		return (float)Math.toDegrees(getYawRad());
 	}
 
 	public final static float len2 (final float x, final float y, final float z, final float w) {
--- a/gdx/test/com/badlogic/gdx/math/QuaternionTest.java
+++ b/gdx/test/com/badlogic/gdx/math/QuaternionTest.java
@@ -1,0 +1,23 @@
+
+package com.badlogic.gdx.math;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+public class QuaternionTest {
+    private static final float epsilon = Float.MIN_NORMAL;
+
+    @Test
+    public void testEulerPrecision () {
+        float yaw = -5f;
+        float pitch = 42.00055f;
+        float roll = 164.32f;
+
+        Quaternion q = new Quaternion().setEulerAngles(yaw, pitch, roll);
+
+        assertEquals(yaw, q.getYaw(), epsilon);
+        assertEquals(pitch, q.getPitch(), epsilon);
+        assertEquals(roll, q.getRoll(), epsilon);
+    }
+}


gdx/src/com/badlogic/gdx/math/Quaternion.java
```diff
@@ -160,40 +160,40 @@ public int getGimbalPole () {
 	/** Get the roll euler angle in radians, which is the rotation around the z axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the z axis in radians (between -PI and +PI) */
 	public float getRollRad () {
-		final int pole = getGimbalPole();
-		return pole == 0 ? MathUtils.atan2(2f * (w * z + y * x), 1f - 2f * (x * x + z * z))
-			: (float)pole * 2f * MathUtils.atan2(y, w);
+		final double pole = getGimbalPole();
```
tommyettinger, line 163: getGimbalPole() hasn't changed, as far as I can see. I don't see a reason to make `pole` a double instead of an int, since it's only checked for equality with int 0.

gdx/src/com/badlogic/gdx/math/Quaternion.java
```diff
@@ -129,24 +129,24 @@ public Quaternion setEulerAngles (float yaw, float pitch, float roll) {
...
-		final float chy_shp = chy * shp;
-		final float shy_chp = shy * chp;
-		final float chy_chp = chy * chp;
-		final float shy_shp = shy * shp;
-
-		x = (chy_shp * chr) + (shy_chp * shr); // cos(yaw/2) * sin(pitch/2) * cos(roll/2) + sin(yaw/2) * cos(pitch/2) * sin(roll/2)
-		y = (shy_chp * chr) - (chy_shp * shr); // sin(yaw/2) * cos(pitch/2) * cos(roll/2) - cos(yaw/2) * sin(pitch/2) * sin(roll/2)
-		z = (chy_chp * shr) - (shy_shp * chr); // cos(yaw/2) * cos(pitch/2) * sin(roll/2) - sin(yaw/2) * sin(pitch/2) * cos(roll/2)
-		w = (chy_chp * chr) + (shy_shp * shr); // cos(yaw/2) * cos(pitch/2) * cos(roll/2) + sin(yaw/2) * sin(pitch/2) * sin(roll/2)
+		final double hr = roll * 0.5f;
```
tommyettinger, line 132: I'm doubtful this suggestion makes a difference, but it might be more correct to use `0.5` (a double multiplier) instead of `0.5f` (a float) here. This would matter more if the multiplier wasn't a (negative) power of two.

This also applies to hp and hy.

gdx/src/com/badlogic/gdx/math/Quaternion.java
tommyettinger: Also, it's still an `int` in `getPitchRad()`.

gdx/src/com/badlogic/gdx/math/Quaternion.java
```diff
@@ -160,40 +160,40 @@ public int getGimbalPole () {
...
 	public float getPitch () {
-		return getPitchRad() * MathUtils.radiansToDegrees;
+		return (float)Math.toDegrees(getPitchRad());
 	}
 
 	/** Get the yaw euler angle in radians, which is the rotation around the y axis. Requires that this quaternion is normalized.
 	 * @return the rotation around the y axis in radians (between -PI and +PI) */
 	public float getYawRad () {
-		return getGimbalPole() == 0 ? MathUtils.atan2(2f * (y * w + x * z), 1f - 2f * (y * y + x * x)) : 0f;
+		return (float)(getGimbalPole() == 0 ? Math.atan2(2.0 * (y * w + x * z), 1.0 - 2.0 * (y * y + x * x)) : 0.0);
```
tommyettinger, line 190: I'm wondering now how much of the round-trip error is from `MathUtils.atan2()`, rather from float/double precision. I think I can find or remake some old spinning-globe GIFs that had significant distortion due to a bad `atan2()` implementation (I don't know if it was the same as MathUtils, but I think it was). The MathUtils docs for `atan2()` say that the average error is 0.1323 degrees. I've been using a rather-different implementation for `atan2()` in my code lately, and even though I know its average and max error are better, I don't know how much better (might not be enough here). It's also slightly faster, last I benchmarked.

gdx/test/com/badlogic/gdx/math/QuaternionTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.Test;
+
+public class QuaternionTest {
+    private static final float epsilon = Float.MIN_NORMAL;
+
+    @Test
+    public void testEulerPrecision () {
+        float yaw = -5f;
+        float pitch = 42.00055f;
+        float roll = 164.32f;
```
tommyettinger, line 15: There should probably be tests on more than just these 3 inputs, especially if you're using as strict an epsilon as `Float.MIN_NORMAL`. I also think this test isn't really doing the same thing as what happens in your original GIFs, where multiple rotations are applied, one after the next, to check for loss of accuracy. Performing a sequence of rotations that we can know, mathematically, should end with the same angles as the start would be the easiest to verify.

tommyettinger (commented):
I'm not totally sure on this one. MathUtils.atan2() has an average error of 0.1323 degrees, according to its docs, which I think would account for this without float precision being a culprit. The simplest check would be to take the original Quaternion and change `MathUtils.atan2(y, x)` to `(float)Math.atan2(y, x)`, then see if that resolves the issue.

gdx/src/com/badlogic/gdx/math/Quaternion.java
joshuaskelly: Good catch. I verified that making `pole` a double versus an int has no impact on precision! 👍

gdx/src/com/badlogic/gdx/math/Quaternion.java
joshuaskelly: Another good catch! I think this was an oversight be me. Changed.

gdx/test/com/badlogic/gdx/math/QuaternionTest.java
joshuaskelly: My test here was the bare minimum to exercise the case that round-tripping an Euler value introduces a high degree of imprecision. I do like the suggestion of adding an additional test to illustrate the case of multiple rotations that should result in a known value.

gdx/src/com/badlogic/gdx/math/Quaternion.java
joshuaskelly: Running my test against current implementation:
`expected:<-5.0> but was:<-5.0020123>`

Changing the `atan2()` usage results in:
`expected:<-5.0> but was:<-5.0000014>`

Switching to `Math.atan2()` would be a much smaller change and improve precision by several orders.

joshuaskelly:
@tommyettinger I really appreciate your feedback! The thing I'm currently trying to figure out is what is an acceptable epsilon of error for a series of rotations. Thoughts on that?

gdx/src/com/badlogic/gdx/math/Quaternion.java
tommyettinger: Woah, that is a significant improvement. That means the error from the entire rest of the operation is only about 1.4e-6, including the other approximations (like `MathUtils.asin()`, which I know isn't as accurate as `Math.asin()`, but isn't very bad). I know `Math.atan2()` usually isn't as fast as an approximation, particularly on Java 8, but it does get optimized pretty well on Java 11 and newer. The actual error here is almost as small as it can get; -5.0000014 has the bit representation 0xC0A00003, and we want -5.0, or 0xC0A00000 -- this is a difference only in the two least significant bits. This could still accumulate, but would do so very, very slowly in most cases. The error before, with the current `Quaternion` and -5.0020123, is 0xC0A0107C, which is wrong in its 13 least significant bits.

tommyettinger:
I'm going to guess that 0.0001 would be a good starting point -- Quaternion currently fails at that level of strictness (20x higher error), so even passing with one rotation would be an improvement. Passing with 3 rotations, one after the next, would show that the error isn't accumulating rapidly, and doing more would be nice to ensure that normal usage won't get noticeably bad.

The build check failed, so I'll look into this. It is likely to just be a formatting issue, and that can be resolved by running the `spotlessApply` task -- see the docs, it's a quick and simple step. https://libgdx.com/dev/contributing/#formatting

joshuaskelly:
Thank you for the recommendations! 👍 

I noticed the build was failing. It was a spaces vs tabs issue. I also noticed that `spotlessApply` adds an undesirable line break to the comments following setting the Quaternion component values
`// cos(yaw/2) * sin(pitch/2) * cos(roll/2) + sin(yaw/2) * cos(pitch/2) * sin(roll/2)`
Which I'd prefer to keep as they were useful in my understand of what was going on there.

tommyettinger:
I think there's a way to tell spotless to ignore a span of code; I think it is ```
// @off
codeThatYouWant = toNotBeFormatted(); // this should stay on one line
// @on
```
I've seen this in a few places in the code, but it isn't especially common, so I could be misinterpreting what it does.

The CI is failing I think as soon as it checks with Spotless, and we don't yet have a way to automatically run Spotless on PRs. I think once your tabs/spaces changes are committed and any complaints Spotless has about comments are silenced by `@off`, CI should run just fine. A maintainer needs to approve CI to run since you're a first-time contributor (GitHub needed this safeguard to block cryptocurrency mining from being done automatically by CI workflows). I'll try to check back often to get the CI to run when you make your changes. You should also be able to run the tests from source using the info on the main site, https://libgdx.com/dev/from-source/ . You should only need to fetchNatives once, and then there's different tasks for different tests (I just run them from IDEA, whatever IDE you use should have an option to run various tests).

**Comment by joshuaskelly** [comment]
No sweat. Again thanks for your help, I'll ping you when I think it's ready for a run. 👍

**Comment by tommyettinger** [comment]
I decided to look into just how much precision could be improved in MathUtils.atan2() -- it turns out, a lot. Testing against your original round-trip code, I can finally get that to pass the test with an epsilon of 0.0001, using an approximation of atan2() and no other changes. The approximation is pretty much exactly the same speed as the current MathUtils.atan2() (within a tenth of a nanosecond per call, and within the margin of error) on my laptop running Java 8 (HotSpot). That means it's about 4x faster than Math.atan2() (14.6 ns vs. 60.4 ns, so still very small times), at least on Java 8. I'm sure that the accumulating error matters more for quaternions, though, so we probably do still want Math.atan2() in Quaternion.

My code is based on another section from that 1955 RAND Corp study on digital approximations, this time sheet 11 (the fourth-fastest) instead of the one I've used before, sheet 9 (which was the second-fastest). I tried sheet 9 here and it did improve error, but not nearly enough -- the error was `-5.00103` instead of MathUtils' `-5.0020123`. With sheet 11, I initially got the roll correct but then failed on the yaw, until I started doing my intermediate calculations **with doubles**. Thanks so much for recommending that, it was the last piece in the puzzle!

I don't think I'll add my code to this PR, because I still think sheet 11 isn't precise enough for multiple rotations, and Math.atan2() is the best way. I'll put a few versions forward in another PR, I think, and let other contributors gauge how much they need precision vs. speed. The speed difference is certainly smaller than the precision difference, though.

File: gdx/test/com/badlogic/gdx/math/QuaternionTest.java [hunk 5]
```diff
@@ -0,0 +1,23 @@
+
+package com.badlogic.gdx.math;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+public class QuaternionTest {
+	private static final float epsilon = Float.MIN_NORMAL;
```
yuripourre, line 9: Is `MathUtils.FLOAT_ROUNDING_ERROR` too low here?

tommyettinger: We could try it, that's a good idea. I think the Math.atan2() version will pass FLOAT_ROUNDING_ERROR but not MIN_NORMAL (which is incredibly low, 1.17549435E-38f). We know the all-double version can pass Float.MIN_NORMAL, though I don't think it can pass Double.MIN_NORMAL simply because floats can't go that low without being subnormal (maybe even then) or 0.

tommyettinger: Nope, neither passes FLOAT_ROUNDING_ERROR. If I change `MathUtils.atan2()` to `(float)Math.atan2()`, I get

```
Expected :-5.0
Actual   :-5.0000014
```

Which is of course very close to where it needs to be, but not quite. If I didn't change MathUtils' atan2, I would have:

```
Expected :-5.0
Actual   :-5.0020123
```

And if I use my same-speed approximation (sheet 11), I get:
```
Expected :-5.0
Actual   :-5.0000577
```

The error on my approximation is 40x worse than (float)Math.atan2(), though Math should be very nearly 0 error, and the method should usually be about 4x faster than (float)Math.atan2(). The error on the current MathUtils.atan2() is about 1400x worse than (float)Math.atan2().

...And none pass with MathUtils.FLOAT_ROUNDING_ERROR as the threshold, so maybe we need to do the double-precision quaternions anyway.

joshuaskelly: I've updated the epsilon to `0.0001f` as suggested.

joshuaskelly:
@tommyettinger I've added a test that performs a 4 degree yaw rotation 90 times and validates that we are back to 0 (within the epsilon of course). Does this seem reasonable a test?

I've also noticed that the current implementation is susceptible to Gimbal lock (which is why my gizmo goes crazy in the gif), which is why I'm avoiding setting the pitch. I have a fix for this, but that will be a separate PR.

tommyettinger:
I think that test sounds OK, but would it be the same complexity to do the test for each angle (once for roll, once for pitch, once for yaw)? Seems like each one of the axes has different code that could have issues, and we certainly want to check all of the axes. Ideally we could test a sequence of more-varied rotations that we know would result in a return to the initial state, but this is more work (unless there's some source that lists example rotations like that?). Using the "simple solution" of just replacing `MathUtils.atan2()` with `(float)Math.atan2()`, I tried running the test with the 4-degree yaw rotation 90 times, and that passed, but 4-degree pitch rotations did not. Roll was fine, though. It seems like pitch is substantially less accurate than the other angles. I don't know if the more significant changes in this PR will let the pitch rotation pass tests.

Gimbal lock comes with the territory for Euler angles, I don't know of a way to avoid it in this scenario. There's some famous anecdote about the Moon lander and how, because they still wanted to use Euler angles, it had a button to reset the rotation in case of gimbal lock. Quaternions can still get gimbal lock, I believe, but not commonly.

joshuaskelly:
Removed the multiple rotation test in favor of keeping the tests simple and to the point.

aus101:
Could increasing the Sin LUT size by a factor of 4 significantly help? Keep in power of 2 to make use of bit shift addressing. Is currently 16384 elements (2^14). As a comparison, the hilarious amount of decompiled Minecraft Java MathHelpers show 65536 elements (2^16). Basic memory allocation would be 65536 x 32 bits = 0.26 MB.

I see that exact results are set for 4 cardinal directions of Sin (and Cos by extension) and that's great but shouldn't we include 8 cardinal directions and 30°/60° exact results in the private class Sin static block?

tommyettinger:
Oh wow, this has been open for a year now. MathUtils.atan2() has better precision in the current code than it did when this PR was first entered, but I think we figured out that even casting Math.atan2() to float introduces enough error to cause problems. I really have no idea what a good solution is for quaternions, since they seem so incredibly sensitive to error building up...

saphirelsnow:
If your code uses a value related to Trigonometry like Sin or Cos, then you should stay with Float if you can.  Java had that problem about 4 years ago.  The problem lingers due to a Pointer problem with C++.  If you have a Float, use a Pointer to convert to some compatible other number value, and convert back to Float, then you will see a loss in value accuracy.  It easy to find.

Very Bad Example:

Float value > Pointer integer > different Float value 

Very Bad Result:  You will get a loss of accuracy.  If you can not test to the millionth digit or even infinity, then your test case is bad.  Go try BigData.  You will find your fix I think.

tommyettinger:
I've looked into this a little more since 2022, and my conclusion is that the `getRoll()`, `getPitch()`, and `getYaw()` methods become unusable if gimbal lock occurs, for any level of precision. A solution that works some of the time is just to modify those three angles separately from the Quaternion, and never modify the Quaternion except via `setEulerAngles()` . This is not suitable in the general case, but for some usage it could be helpful. Using axis-angle form would be an alternative that doesn't seem to struggle with gimbal lock.

I really looked around for any working implementations of getPitch(), getRoll(), and getYaw(). I didn't find any satisfactory ones, even in large game engines (though I didn't look at all of them). There's some webpages linked on Wikipedia's Euler Angles article(s) that give a very cursory glance through the methods, and if they do mention gimbal lock, it's generally given in a broken link. I'm not convinced that libGDX is currently implementing gimbal lock handling correctly for those three methods, because I'm not sure it *can* be implemented correctly.

PR #6692, opened 2021-11-08
Improve MathUtils.atan2() precision.

tommyettinger:
I had looked around for alternative atan2() approximations a while ago, and somewhat recently found a good series of approximations published in 1955 by RAND Corp. I believe I already used that study for asin() and acos() here. I've ramped up the quality on the approximations here because it can go up quite a bit while still keeping pretty much exactly the same speed as the current MathUtils.atan2(). I also added atan(), because it's quite similar to atan2(). Both max and average error are substantially better here -- my tests show max error was 0.004883076937336028 radians before, and 1.9221310783024137E-6 radians now. Average error goes from 0.0023030318272527724 radians (before) to 1.0570983651010727E-6 radians now. That means in both cases, we go from two zeros (0.00n) to 5 zeros (0.00000n).

There's some part to discuss still, here -- By using a different approximation, we can speed up MathUtils.atan2() to faster than it was before, and still with lower error, but the max and average errors are still pretty close to what's in MathUtils now -- that is, not great. It also isn't drastically faster; it goes from 14.6 ns to 12.4 ns per call. I think that the speed difference is small enough, and the quality difference substantial enough, to maintain the current speed and just improve quality.

I ran spotlessApply, but I don't know if it worked out OK -- Git claimed there were ~2200 files changed, but that they had no content changes.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/MathUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/MathUtils.java
@@ -94,11 +94,10 @@ public final class MathUtils {
 
 	// ---
 
-	/** A variant on {@link #atan(float)} that does not tolerate infinite inputs for speed reasons.
-	 * This can be given a double parameter, but is otherwise the same as atan(float), and returns
-	 * a float like that method. It uses the same approximation, from sheet 11 of "Approximations
-	 * for Digital Computers." This is mostly meant to be used inside {@link #atan2(float, float)},
-	 * but it may be a tiny bit faster than atan(float) in other code.
+	/** A variant on {@link #atan(float)} that does not tolerate infinite inputs for speed reasons. This can be given a double
+	 * parameter, but is otherwise the same as atan(float), and returns a float like that method. It uses the same approximation,
+	 * from sheet 11 of "Approximations for Digital Computers." This is mostly meant to be used inside
+	 * {@link #atan2(float, float)}, but it may be a tiny bit faster than atan(float) in other code.
 	 * @param i any finite double or float, but more commonly a float
 	 * @return an output from the inverse tangent function, from {@code -HALF_PI} to {@code HALF_PI} inclusive */
 	public static float atanUnchecked (double i) {
@@ -136,13 +135,11 @@ public final class MathUtils {
 		if (x > 0)
 			return atanUnchecked(n);
 		else if (x < 0) {
-			if (y >= 0)
-				return atanUnchecked(n) + PI;
+			if (y >= 0) return atanUnchecked(n) + PI;
 			return atanUnchecked(n) - PI;
 		} else if (y > 0)
 			return x + HALF_PI;
-		else if (y < 0)
-			return x - HALF_PI;
+		else if (y < 0) return x - HALF_PI;
 		return x + y; // returns 0 for 0,0 or NaN if either y or x is NaN
 	}
 
@@ -158,7 +155,7 @@ public final class MathUtils {
 			return (float)Math.sqrt(1f - a) * (1.5707288f - 0.2121144f * a + 0.0742610f * a2 - 0.0187293f * a3);
 		}
 		return 3.14159265358979323846f
-				- (float)Math.sqrt(1f + a) * (1.5707288f + 0.2121144f * a + 0.0742610f * a2 + 0.0187293f * a3);
+			- (float)Math.sqrt(1f + a) * (1.5707288f + 0.2121144f * a + 0.0742610f * a2 + 0.0187293f * a3);
 	}
 
 	/** Returns asin in radians; less accurate than Math.asin but may be faster. Average error of 0.000028447 radians (0.0016298931
@@ -173,8 +170,7 @@ public final class MathUtils {
 			return 1.5707963267948966f
 				- (float)Math.sqrt(1f - a) * (1.5707288f - 0.2121144f * a + 0.0742610f * a2 - 0.0187293f * a3);
 		}
-		return -1.5707963267948966f
-				+ (float)Math.sqrt(1f + a) * (1.5707288f + 0.2121144f * a + 0.0742610f * a2 + 0.0187293f * a3);
+		return -1.5707963267948966f + (float)Math.sqrt(1f + a) * (1.5707288f + 0.2121144f * a + 0.0742610f * a2 + 0.0187293f * a3);
 	}
 
 	/** Arc tangent approximation with very low error, using an algorithm from the 1955 research study "Approximations for Digital
--- a/gdx/src/com/badlogic/gdx/math/MathUtils.java
+++ b/gdx/src/com/badlogic/gdx/math/MathUtils.java
@@ -94,22 +94,56 @@ public final class MathUtils {
 
 	// ---
 
-	/** Returns atan2 in radians, less accurate than Math.atan2 but may be faster. Average error of 0.00231 radians (0.1323
-	 * degrees), largest error of 0.00488 radians (0.2796 degrees). */
-	static public float atan2 (float y, float x) {
-		if (x == 0f) {
-			if (y > 0f) return HALF_PI;
-			if (y == 0f) return 0f;
-			return -HALF_PI;
-		}
-		final float atan, z = y / x;
-		if (Math.abs(z) < 1f) {
-			atan = z / (1f + 0.28f * z * z);
-			if (x < 0f) return atan + (y < 0f ? -PI : PI);
-			return atan;
-		}
-		atan = HALF_PI - z / (z * z + 0.28f);
-		return y < 0f ? atan - PI : atan;
+	/** A variant on {@link #atan(float)} that does not tolerate infinite inputs for speed reasons, and because infinite inputs
+	 * should never occur where this is used (only in {@link #atan2(float, float)}).
+	 * @param i any finite float
+	 * @return an output from the inverse tangent function, from PI/-2.0 to PI/2.0 inclusive */
+	private static float atn (final double i) {
+		// We use double precision internally, because some constants need double precision.
+		final double n = Math.abs(i);
+		// c uses the "equally-good" formulation that permits n to be from 0 to almost infinity.
+		final double c = (n - 1.0) / (n + 1.0);
+		// The approximation needs 6 odd powers of c.
+		final double c2 = c * c;
+		final double c3 = c * c2;
+		final double c5 = c3 * c2;
+		final double c7 = c5 * c2;
+		final double c9 = c7 * c2;
+		final double c11 = c9 * c2;
+		return (float)Math.copySign((Math.PI * 0.25)
+			+ (0.99997726 * c - 0.33262347 * c3 + 0.19354346 * c5 - 0.11643287 * c7 + 0.05265332 * c9 - 0.0117212 * c11), i);
+	}
+
+	/** Close approximation of the frequently-used trigonometric method atan2, with higher precision than libGDX's atan2
+	 * approximation. Maximum error is below 0.00009 radians. Takes y and x (in that unusual order) as floats, and returns the
+	 * angle from the origin to that point in radians. It is about 5 times faster than {@link Math#atan2(double, double)} (roughly
+	 * 12 ns instead of roughly 62 ns for Math, on Java 8 HotSpot). It is slightly faster than libGDX' MathUtils approximation of
+	 * the same method; MathUtils seems to have worse average error, though. <br>
+	 * Credit for this goes to the 1955 research study "Approximations for Digital Computers," by RAND Corporation. This is sheet
+	 * 9's algorithm, which is the second-fastest and second-least precise. The algorithm on sheet 8 is faster, but only by a very
+	 * small degree, and is considerably less precise. That study provides an {@link #atan(float)} method, and the small code to
+	 * make that work as atan2() was worked out from Wikipedia.
+	 * @param y y-component of the point to find the angle towards; note the parameter order is unusual by convention
+	 * @param x x-component of the point to find the angle towards; note the parameter order is unusual by convention
+	 * @return the angle to the given point, in radians as a float; ranges from -PI to PI */
+	public static float atan2 (final float y, float x) {
+		float n = y / x;
+		if (n != n)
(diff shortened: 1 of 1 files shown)


NathanSweet:
It looks good to me! Low error `atan2` is better than slightly faster. Do we really need `atn` and `atan`? MathUtils already has caveats, not supporting infinite inputs seems fine and the more common usage. If you'd like to keep it, maybe rename `atn` -> `atan` and `atan` -> `atanInfinityAndBeyond` (or something).

Couple unimportant nitpicks: `final` for locals doesn't do much, don't need `else` when the first condition returns.

tommyettinger:
I figured since atn() was private, it wouldn't matter too much. The speed difference is small between atan() and atn(), if I remember it right, and should be about the same as a call to Math.min() (this does mean there's a branch in atan() and not in atn(), but one path should only be taken extremely rarely). I think atn() does get just inside an inlining code size limit on HotSpot, where atan() does not, but I'm not certain. The inlining-like benefit was noticed in an earlier (smaller, less precise) approximation, and the two methods might not be much different now.

I think making up a new public method that acts like atan(), but tolerates infinite inputs... is less than optimal. I think users probably wouldn't know it exists, for one thing, since there's no counterpart to it in math (with atan() in math normally tolerating infinite inputs, having the method that acts like math-atan not be called atan() is unexpected). Users who do know it exists might not know their inputs necessitate it until they are already infinite. I'd rather the method called atan() act like the function atan() in math. I could make atn() public and change its name to atanUnchecked() or something, since it doesn't check for infinite inputs or handle them. Since atanUnchecked() would be a method that only exists in code, not in math they teach in school, I don't think there's a problem with the name being unusual -- and it goes from being private (unusable for library consumers) to public with an uncommon name (but usable). The few users looking for optimizations for atan(), which is a rarely-called method anyway, would probably find atanUnchecked() pretty easily.

aus101:
Good thinking @tommyettinger. I wonder if you found the PDF from the same Stack Overflow question I did. You know, the arcsin and arccos use Taylor series expansion and must be replaced with minmaxed equations for same reason. Taylor series approximation infamous issue is it's very accurate at the one point but the absolute error goes to hell for the outside half of the interval. I'm shocked using it for numerical approximation has been hiding in plain sight. Embarrassing imo.

On page 47 (50 in PDF) of Approximations it starts describing the Chebyshev polynomial method and then the subsequent minmaxing. Chebyshev is the starting "guess" for minmax because it's already so close and the non-evenly spaced sample points are ideal to avoid Runge's phenomena and spread the error evenly over the interval. The minmaxing can be done computationally in modern times with the Remez algorithm in just a few iterations.

Pro move is to find someone with MATLAB to use Chebfun and its Remez algorithm (that initially performs Chebyshev) for  whatever number of terms you want. I suspect it will be slightly / non-significantly better than Approximations book equations, but the book lacks cos, arccos and tan. Worth testing on Java 11+ if equations can beat out LUT for sin and cos at same accuracy.

tommyettinger:
I'm not totally sure if asin and acos have severe problems from using Taylor series approximation, since they're only defined for a rather small domain (-1 to 1 is about as small as it gets). I'll check the error at the edges of the domain for those two -- here's asin(), comparing the float versions of both (MathUtils only returns float, so I'm casting Math.asin() to float as well):
```
-1.000000000: Math gives -1.570796371, MathUtils gives -1.570796371 ; off by  0.000000000
-0.999899983: Math gives -1.556652904, MathUtils gives -1.556653619 ; off by -0.000000715
-0.999000013: Math gives -1.526071548, MathUtils gives -1.526073694 ; off by -0.000002146
-0.990000010: Math gives -1.429256916, MathUtils gives -1.429261923 ; off by -0.000005007
-0.509999990: Math gives -0.535184801, MathUtils gives -0.535229564 ; off by -0.000044763
-0.001000000: Math gives -0.001000000, MathUtils gives -0.001065016 ; off by -0.000065016
-0.000100000: Math gives -0.000100000, MathUtils gives -0.000167370 ; off by -0.000067370
 0.000000000: Math gives  0.000000000, MathUtils gives  0.000067592 ; off by  0.000067592
 0.000100000: Math gives  0.000100000, MathUtils gives  0.000167370 ; off by  0.000067370
 0.001000000: Math gives  0.001000000, MathUtils gives  0.001065016 ; off by  0.000065016
 0.010000000: Math gives  0.010000166, MathUtils gives  0.010044098 ; off by  0.000043931
 0.500000000: Math gives  0.523598790, MathUtils gives  0.523644924 ; off by  0.000046134
 0.990000010: Math gives  1.429256916, MathUtils gives  1.429261923 ; off by  0.000005007
 0.999000013: Math gives  1.526071548, MathUtils gives  1.526073694 ; off by  0.000002146
 0.999899983: Math gives  1.556652904, MathUtils gives  1.556653619 ; off by  0.000000715
 1.000000000: Math gives  1.570796371, MathUtils gives  1.570796371 ; off by  0.000000000
```
I think getting within 1/10000 of a radian, for the input 0.0, is sufficient for just about any game. The implementation for acos() is just an offset and sign-flip of asin(), if I remember correctly (which is pretty much the mathematical definition, too); it probably loses some precision just because it can return values with a higher magnitude than asin(), but from when I checked this the first time, I don't think it was very much.

It's pretty hard to beat the 14-bit LUT sin and cos use for precision with floats, even with the implementation here doing no interpolation. It's also quite fast, at 3x to 5x faster than Math's sin() or cos() depending on JDK version. Benchmarking MathUtils vs. Math on several Java versions: https://github.com/tommyettinger/assorted-benchmarks/blob/master/src/main/java/net/adoptopenjdk/bumblebench/examples/GDXSinBench.java#L20-L47 , https://github.com/tommyettinger/assorted-benchmarks/blob/master/src/main/java/net/adoptopenjdk/bumblebench/examples/MathSinBench.java#L19-L46 . The [best equation I have](https://github.com/yellowstonegames/SquidLib/blob/master/squidlib-util/src/main/java/squidpony/squidmath/NumberTools.java#L882-L917) to approximate sin() is half the speed of the MathUtils LUT code, and has worse precision, though I haven't needed to find a better one yet regarding precision. It would take an equation with substantially fewer operations, as far as I can tell, to beat the LUT in microbenchmarks. In an actual game, LUTs aren't thought to perform as well (I think because of contention for processor cache), but even if they were 50% slower, they would still be as fast as some equations and more precise.

The only major trigonometric function left is tan(), and I almost never need tan(), so I'm not sure we really need to replace Math.tan() with our own.

I don't have Mathematica and don't really feel competent to do many of these approximations on my own. I can evaluate the choices in front of me, for speed and precision, but I don't think I could do much more than adjust the input mapping or some other simple tweaks to an existing approximation.

@aus101 If you find an existing Remez approximation of any of the functions in MathUtils and you want to benchmark it, I'd be happy to guide you through the mess of a benchmarking repo I have, or just to benchmark whatever you throw at me. We're kind-of approaching the "generally good enough" territory with approximations, but the recent quaternion issue #6691 shows we have a ways to go in some areas.

aus101:
I admit I want the perfect solution without as much regard to the significance. Thanks for the asin comparison. The error is two orders of magnitude better than I thought. We know that sin and cos are off by a phase of 90° (pi/2). Can show arcsin(theta) + arccos(theta) = tan(1) = 90° (pi/2) => (pi/2) - arcsin = arccos. I understand the Approximations author avoiding the effort to find minmaxed cos, arccos and tan in 1955, but modern world gets to insta-generate optimal approximations with Chebfun to avoid one add/subtract per call.

Could argue saving one add/subtract isn't significant when cos uses the sin LUT with a phase addition. If there is an optimal solution though, is that not important? I admit I've never used tan XD, but may as well include with zero effort to complete the full set.

Turns out the university near where I live gives MATLAB with ALL toolkits to ALL current students, who can install or use over VPN or web browser. I need for 15 min? Yes, I'd like really help doing scientific benchmarking or asking you to do them. I never thought System nanotime before and after approach was real proof or how to fairly factor in JIT or cache miss.

In your links, I've never seen such detailed Javadoc. I think you're saying more accurate atan2 compounds over repeated calls to significantly help the quaternion library.

There is an interesting article called "Barycentric-Remez algorithms for best polynomial approximation in the chebfun system". Shows Remez's 1934 publication with 3 human calculators was accurate to ~4 decimal places compared to arbitrary precision Chebfun. I assume RAND effort was better. Significantly, article shows that with a polynomial of degree 10, max reduction in error from Chebyshev to Remez is less than a factor of 3.47, via < [2 + 2/π * ln(n)]. Same order of magnitude in base 10 but the gain approaching log(3.47)/log(2) = 1.79 bits looks good.

I'm actually more interested in fast log2, log10, ln and e^(-x) for probability distributions and information theory. The log2 in particular is bad for using a division versus reciprocal multiplication. A fast log2 can be used in fast e^(-x) for more winning.

NathanSweet:
Ah, I missed the method was private. At any rate, let's merge once you're happy with it!

tommyettinger:
@NathanSweet I think your suggestion of having both is a good idea, since most usage probably doesn't call atan() enough to notice a speed difference -- but when people need it just a bit faster and know the bounds on their input, there's the atanUnchecked() method now. I think this is good now. I'll merge.

cubesdev:
Is there really a noticable performance difference in comparison to java.lang.Math class (any data confirming that using a custom implementation has any impact on a real world applications?). My profiler sessions have never confirmed any problems when using the java.lang.Math class in my OpenGL/LibGDX applications. I had trouble with inaccuracies when I once used MathUtils class instead though...

tommyettinger:
@cubesdev it strongly depends on how many calls you make to trig functions per-frame. If you make 10 calls a frame, you will not notice a difference. Some applications really need a lot of trig, like scenes where every particle (out of thousands) is rotated differently, and taking a circular path. I've noticed slowdown when doing map projection from a globe to the screen using Math.atan2(), and subtle inaccuracy with the old MathUtils.atan2() before this PR. What we know is that Math methods are not as fast as MathUtils approximations, and sometimes we don't need the precision of Math methods. According to JMH, MathUtils.atan2() (the just-merged version) is about 4x faster than Math.atan2(). Math.sin() and Math.cos() are between 3x and 5x slower than MathUtils' (very precise) approximations, depending on JDK version. In cases where you don't need 5 digits of precision, such as just about any visual effect, it makes sense to use the fastest method that meets the demanded precision. If your use case needs perfect accuracy, sure, no one's stopping you from using Math methods. Also, if you have been using rotations with SpriteBatch, you have been using MathUtils under-the-hood and probably didn't notice any inaccuracy there.

cubesdev:
@tommyettinger okay, if you have done microbenchmarks confirming the improvement then it's good to have. And the merged PR's new approximation looks great btw. I was just asking because I had huge performance issues in the past with LibGDX Matrix4 JNI "optimizations" which were considerably slower than a plain Java implementation (at least on a modern HotSpot desktop VM). And while the Matrix4 C implementation may have been slower in previous JRE versions or at least on Android it is always worth checking whether a particular micro optimizaton is still beneficial. So I was wondering whether that may be the case here, too. But thanks for your statement on the topic.

tommyettinger:
Oh yeah, I remember the Matrix4 JNI issue; I used the same assorted-benchmarks codebase to confirm that the replacement was faster without JNI. [My results from that old test are here](https://github.com/tommyettinger/assorted-benchmarks/blob/master/src/main/java/net/adoptopenjdk/bumblebench/examples/Matrix4Benchmarks.md). Android benchmarking is unfortunately not as easy as benchmarking on a desktop, and GWT benchmarking is a nightmare. On the plus side, what Anuken showed in that benchmark set at the bottom (which is measured in time taken, instead of ops/second) was that Android acted similarly to desktop, and JNI was slower there when it was slower on desktop... just by a much larger degree. I don't think there's a ton of serious performance mistakes like that left in the codebase, though it's not really possible to say there are none left. We just haven't found them yet...

cubesdev:
@tommyettinger your benchmark looks awesome and confirms my test results I did a year ago when I tried to optimize the 3D AnimationController! Is there any reason why LibGDX has not reverted to a plain Java implementation yet? I have dealt with it by modifying the Matrix4 class on my side. I was even able to improve performance by first loading all matrix values into local float variables (float m00 = val[M00]; ...) before using them in the calculations. I guess it helps the JIT compiler. It would be great if those libgdx "hacks" were not needed and I could switch to LibGDX's official releases though.

tommyettinger:
@cubesdev it is a pure Java implementation currently except for 3 methods that need pointer manipulation to handle stride: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1209-L1245 . The rest has all been moved to pure Java, and this was done in 1.9.11 about a year and a half ago. If you have some code that behaves the same and improves performance, it would be great if you wanted to PR it. If the changes are simple enough, just describing the types of change that help might be enough.

aus101:
Pending Issues:

**1.** The approximation comes from 1955 when we didn't have IEEE floating point until 1985. It's therefore sub-optimal and needs tweaking, such as with Remez Exchange.
**2.** Dr. Salzer didn't apply Barycentric interpolation to Chebyshev approximation (that RAND uses) until 1972. While less an issue here in default [-1, 1] domain with small n of 11, using Barycentric weights on Chebyshev 2nd kind points yields weights of +/- (1/2) and (1) only. This means the weights are perfectly applied in floating point math and no risk of underflow.
**3.** While we don't have C++ fma to spam, I found other implementations check for x > 1 and use arctan(x) = pi/2 - arctan(1/x), so can avoid expensive division operation when x <= 1, versus always apply (x + 1) / ( x - 1).
**4.** Note the high relative error for very small y/x from correct Math.atan2. Could be any numerical approximation fails on small ratio but should note inaccuracy in the Javadoc for very small y/x.
=> (1/500): 0.00**19**999973333397333 correct vs 0.00**20**00728 (3.6% relative error)
=> (1/1000): 0.00**09**999996666668668 correct vs 0.00**10**011712 (11.7% relative error)

**Best approximation is MATLAB Chebfun (that auto uses Barycentric + 2nd kind) + Remez.**
I'd also trust Mathematica if someone has that. Weeks before I can borrow MATLAB.
I think I can code Remez in Java for real numbers and small n. Trick is using n+2 sample points to guarantee sign changes in equation and therefore oscillation of error.

Issue #6679, opened 2021-10-12
Intersector.intersectPolygons (Polygon p1, Polygon p2, Polygon overlap) fail with rectangles


fabiitch:
Please ensure you have given all the following requested information in your report.

#### Issue details
I create two rectangle with Polygon class, when i try Intersector.intersectPolygons (Polygon p1, Polygon p2, Polygon overlap)
it return always false, even i rotate/move them.

The two other methods from intersector return true 
- Intersector.intersectPolygons (FloatArray polygon1, FloatArray polygon2)
- Intersector.intersectPolygonEdges (FloatArray polygon1, FloatArray polygon2)
#### Reproduction steps/code
<code>
    @Test
    public void intersect2Rect() {

        float[] vertices1 = new float[]{-50, 0, 50, 0, 50, 50, -50, 50};
        float[] vertices2 = new float[]{-10, -100, 10, -100, 10, 100, -10, 100};
        Polygon poly1 = new Polygon(vertices1);
        Polygon poly2 = new Polygon(vertices2);

        boolean b1 = Intersector.intersectPolygons(poly1, poly2, null);

        boolean b2 = Intersector.intersectPolygons(new FloatArray(poly1.getTransformedVertices()), new FloatArray(poly2.getTransformedVertices()));

        boolean b3 = Intersector.intersectPolygonEdges(new FloatArray(poly1.getTransformedVertices()), new FloatArray(poly2.getTransformedVertices()));

        System.out.println("b1="+ b1 +" | b2="+b2+" | b3=" +b3);
        Assertions.assertTrue(b1);
        Assertions.assertTrue(b2);
        Assertions.assertTrue(b3);

    }
<code>
result : b1=false | b2=true | b3=true

#### Version of libGDX and/or relevant dependencies
1.9.10
- [X] Linux (only test on linux)

Some screens 
![image](https://user-images.githubusercontent.com/12832996/136926014-797c8e8d-df49-4d2b-a3d2-775a0b882535.png)
![image](https://user-images.githubusercontent.com/12832996/136926092-371e11e1-61ad-4698-b6e3-958a0054d992.png)

Thanks for reading.

fabiitch:
Polygon p2 must be clockwise.
pr https://github.com/libgdx/libgdx/pull/6695 to add it in doc method.
I close issues, no real bug

PR #6695, opened 2021-11-14
Intersector.intersectPolygons add doc p2 should be clockwise #6679

fabiitch:
Hey, i work on my issues [#6679](https://github.com/libgdx/libgdx/issues/6679)

i found after test that, the second polygon must be clockwise.

I dont know if its an error in algorithm implementation or algorithm itself need it.
I find an implementation in python  where the two polygons must be clockwise.
When i look the algorithm, a rotation dir is needed,  but it can be ccw too. 

Anyways the method seems work fine, i just add a line of doc for utilisation
Thanks for reading

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -152,6 +152,7 @@ public final class Intersector {
 
 	/** Intersects two convex polygons with clockwise vertices and sets the overlap polygon resulting from the intersection.
 	 * Follows the Sutherland-Hodgman algorithm.
+	 * p2 should be clockwise
 	 * @param p1 The polygon that is being clipped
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -151,7 +151,7 @@ public static boolean isPointInPolygon (float[] polygon, int offset, int count,
 	private final static Vector2 e = new Vector2();
 
 	/** Intersects two convex polygons with clockwise vertices and sets the overlap polygon resulting from the intersection.
```
obigu, line 153: Doesn't it already specifies it here the polygons should be clockwise?

fabiitch: XD  i read it 100 times previous month, and i never saw it.

PR #6696, opened 2021-11-14
Geometry utils add

fabiitch:
Hey again, after work on my issues [#6679](https://github.com/libgdx/libgdx/issues/6679)
I find the answer (polygon p2 muste be clockwise)

I add some methods similar to CCW but for clockWise  (check, ensure)

Thanks for reading

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,7 @@
 - Bug Fix: AssetManager backslash conversion removed - fixes use of filenames containing backslashes
 - gdx-setup now places the assets directory in project root instead of android or core. See advanced settings (UI) or arguments (command line) if you don't want it in root.
 - API Fix: Resolved issues with LWJGL 3 and borderless fullscreen
+- API Addition: GeometryUtils,polygons isCCW, ensureClockwise, reverseVertices
 
 [1.10.0]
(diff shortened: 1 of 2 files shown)


noblemaster:
This looks fine in my opinion.

noblemaster (approved):

mgsx-dev:
i would name it "CounterClockWise" instead of "CCW" for consistency with existing "ClockWise" method names.

noblemaster:
The original method is called "CCW", so you mean the new method should be called "CW"? I think typing it "Clockwise" is fine as naming it "CW" is very similar to "CCW" and could lead to problems if you forget at 'C' (very similar).

mgsx-dev:
what i meant is that existing method is named "isClockwise" and new method is named "isCCW". I think it's better to have consistency, that's why i propose to name the new method "isCounterClockwise" instead of "isCCW". 
But if it doesn't matter to you, i won't fight against that :D

noblemaster:
My bad, I just looked at the existing `ensureCCW` method. I see now that there is a new `isCCW` but it matches `ensureCCW`. I guess `isCCW` can be changed to `CounterClockwise` but then it won't match `ensureCCW` anymore.

Also, I don't mind either way. I somewhat prefer it how it is in now, but maybe there is a better solution?

mgsx-dev:
Ah ah ! yes indeed, it was already inconsistent... my bad too. I agree, let's keep the inconsistency consistent :D

mgsx-dev (approved):

Issue #6702, opened 2021-11-21
How can I check if the font file contains certain font?


mingmoe:
I use the `libgdx-freetype` extension to load .ttf font file.

I have multiple fonts files. And I want to load different fonts from different files.(For example,loading Japanese fonts from a Japanese font file,loading English fonts from a English font file)

But I did not find a way for libgdx-freetype to check whether the font glyph exists.

How can I do it? Thanks.

lyze237:
Adding to that it would be cool if the generator could  support fallback fonts on its own for this reason.

Frosty-J:
So, this is definitely hacky, but...

```java
GlyphLayout glyph = new GlyphLayout(font, char);
if (glyph.width > 0) exists = true;
else exists = false;
```

tommyettinger:
I've written some code before to iterate something similar; it used this type of thing:
```java
public static boolean exists(BitmapFont font, char c) {
  return font.getData().getGlyph(c) != null;
}
```
Fair warning, I don't know how well this behaves if there is a non-null missingGlyph set in the font's data. This would require the BitmapFont to already have been generated by gdx-freetype, so that might be an issue in your case (not sure).

Regarding the earlier answer, `glyph.width` can be 0 in some cases, but I'm not sure if gdx-freetype correctly handles them anyway. Combining marks generally have zero width, as do the appropriately-named zero-width space, zero-width joiner, etc.

mingmoe:
> I've written some code before to iterate something similar; it used this type of thing:
> 
> ```java
> public static boolean exists(BitmapFont font, char c) {
>   return font.getData().getGlyph(c) != null;
> }
> ```
> 
> Fair warning, I don't know how well this behaves if there is a non-null missingGlyph set in the font's data. This would require the BitmapFont to already have been generated by gdx-freetype, so that might be an issue in your case (not sure).
> 
> Regarding the earlier answer, `glyph.width` can be 0 in some cases, but I'm not sure if gdx-freetype correctly handles them anyway. Combining marks generally have zero width, as do the appropriately-named zero-width space, zero-width joiner, etc.

I noticed that the `FreeTypeFontParameter` has a `characters` field with comment "The characters the font should contain". (ref: https://github.com/libgdx/libgdx/wiki/Gdx-freetype#how-to-use-gdx-freetype-in-code)

And "generator.generateFont(parameter);" will only return the font specified in "characters" field.

It is still impossible to know whether a font file contains a glyph.

[...]

PR #6703, opened 2021-11-21
Add Methods for libgdx-freetype Extension

mingmoe:
#6702 

This pull request will add methods for the freetype extension.

There is a method: `FreeTypeFontGenerator#hasGlyph(int)`.

This can check whether the font file has certain glyphs.

There're no tests. According to the [documentation](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_get_char_index) of freetype, this should work on all platforms.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,7 @@
 - Bug Fix: AssetManager backslash conversion removed - fixes use of filenames containing backslashes
 - gdx-setup now places the assets directory in project root instead of android or core. See advanced settings (UI) or arguments (command line) if you don't want it in root.
 - API Fix: Resolved issues with LWJGL 3 and borderless fullscreen
+- API Addition: Added FreeTypeFontGenerator#hasCharGlyph method.
 
 [1.10.0]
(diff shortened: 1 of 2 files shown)


Frosty-J:
Tested on Linux. Works fine in my *extremely rigorous* <span title="end sarcasm">(/s)</span> testing of one font and two characters.

If I'm being really picky, the `char` version is made redundant by the `int` version as you may have noticed from calling `getCharIndex(int charCode)` with either an `int` or `char`. But it's certainly more obvious it accepts characters when the type is `char`, so I'm on the fence.

And no other methods in that class (or anywhere I've noticed in libGDX) use `final` for their parameters. I'm a `final` fan but adding it into arguments just gets too much.

There's nothing else to pick at here. It's a useful little method. Bring on the merge!

tommyettinger (approved):
Looks good to me; this is all functionality that the FreeType extension could do internally before, and is just now being made available to users. And it seems quite useful, so I'm approving.

mingmoe:
Perfect. Modified the code based on comments. :-)

@frosty-j

Frosty-J (approved):

MrStahlfelge (approved):
LGTM

PR #6707, opened 2021-11-26
GWT: Extract setupMainLoop method

yuripourre:
Extracting the method setupMainLoop allows developers to extend GWTApplication and handle the main loop externally.

It might sound very specific but this is how [WebXR works](https://developer.mozilla.org/en-US/docs/Web/API/XRSession/requestAnimationFrame).


MrStahlfelge (approved):
LGTM, no problem in opening up for enhancements

yuripourre:
@MrStahlfelge thanks for approving. I would like to try to create an experiment using the WebXR API and this is the first step.

yuripourre:
@MrStahlfelge thank you so much!

PR #6712, opened 2021-12-02
Expose more attributes from GwtApplication

yuripourre:
Expose attributes from GwtApplication (similar to what is done at AndroidApplication).

@MrStahlfelge Sorry for the extra work, those attributes are also necessary.

Related to: #6707

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -67,12 +67,12 @@ import java.util.regex.Pattern;
  * implement any loading screen drawing via GWT widgets.
  * @author mzechner */
 public abstract class GwtApplication implements EntryPoint, Application {
-	private ApplicationListener listener;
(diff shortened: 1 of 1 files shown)


MrStahlfelge (commented):
Not sure this time - for example, input can be fetched with Gdx.input and set when created by overriding createInput method. Changing it in runtime does not make sense. Same for net, audio etc.

yuripourre:
@MrStahlfelge Makes sense, but I couldn't find a better way to reuse the code at GwtApplication for WebXR. I think I will have to call `input.reset()` in the main loop at WebXRApplication (unless the is in the same package). I tried to follow the same as we have for the [AndroidApplication](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java#L45) maybe it's legacy code.

I will turn the PR as draft for now and once I have a working example I will do the proper changes. Thanks for reviewing!

yuripourre:
@MrStahlfelge Only now I realized what you were talking about, `Gdx.input = input;` so I can access it from the `Gdx` class.

This PR is not needed.

PR #6714, opened 2021-12-05
(#6480) tiled parallax support

### LINKED ISSUE #6480: Tiled 1.5 Parallax Layer Support
This is not an issue, it is a question and if possible I would like to contribute.

The latest version of Tiled (1.5) officially support parallax scrolling: https://thorbjorn.itch.io/tiled/devlog/206681/parallax-scrolling-is-here

image

In my opinion it would be great if LibGDX supports that as well. Reading the property values is most likely not that difficult and could become part of the `TiledMapTileLayer` class or whatever it is called.

But maybe also the `TiledMapTileRenderer` should support it correctly? But how do you achieve the exact same result as in Tiled?

I personally always achieved a parallax effect by multiplying the camera position with a specific value to move that layer faster/slower, meaning that before rendering a layer I would modify the camera position temporarily, render the layer, and reset the camera position to its original values.

Is this the correct way to do it? What is the exact formula for the camera position modification because a simple "camera.position.x *= parallaxFactor.X" will most likely not be sufficient.

If someone could give me a little bit feedback I am happy to prepare a PR for it, if it is something useful.

What do you think?

Quillraven:
I finally had now a little time to look into this topic (fixes #6480) and had a discussion with the guy from Tiled to understand better how Tiled is implementing parallax scrolling. You can find the discussion [here](https://github.com/mapeditor/tiled/issues/3196).

However, I seek for your help guys because I am a little bit overwhelmed by the big amount of different renderers and where exactly I should put the code.

----

For testing, I have adjusted the existing `ortho.tmx` file and added some group layers with different parallax factors to also test the correct behavior of "structured" layers where the final scrolling factor is the layer's factor multiplied by its parents.

My approach is done in the map loader directly. Whenever a parent is set, which seems to happen after its children are loaded then the parallax value is simply multiplied by the parent's parallax value. That should support that use case correctly but please note that the Tiled author mentioned that this is currently bugged and will be fixed in an upcoming version. But that should not stop us from implementing it correctly ;)

----

Next I tried to update the render logic. I see there is a `TiledMapRenderer` interface which I assume is used by all renderers related to the Tiled Editor? 

There are two implementations of this interface: `BatchTiledMapRenderer` and `OrthoCachedTiledMapRenderer`.

I saw that the second one does not correctly support group layers and also the render code seems to be copy pasted which is also not ideal. `render()` and `render(int[] layers)` has 99% the same code from what I can tell and should be merged.
But anyway, it seems like this renderer is not up to date and therefore I ignored it for now.

The `BatchTiledMapRenderer` is then subclassed by a few other renderers like `HexagonalTiledMapRenderer` or `OrthogonalTiledMapRenderer`.

Since parallax scrolling should be a generic solution for any renderer in my opinion I put my code in the `BatchTiledMapRenderer` hoping that this will then apply to all subclasses but I did not verify that yet, sorry.

My idea was to put it into the `renderMapLayer` method to also apply the logic to any type of layer because in my opinion it should always behave the same. It simply renders the layer with a specific offset no matter if it is a tiled, object or image layer.

I currently use the transform matrix of the `batch` and apply a specific translation but this has the disadvantage that I have to break the rendering process with `end` and `begin` which I guess has a negative impact on the render performance?

----

I used the existing `TiledMapLayerOffsetTest` to test the changes. It does already a parallax render effect but it does not 100% match the result I get in Tiled. The problem is I am missing some information in Tiled and therefore it is hard for me to understand where the difference is coming from.

----

Anyway, please have a look and let me know how to do it properly and efficiently.

Edit: I noticed that there is also some weird clipping happening and the parallax layer sometimes get cut off although it is within the view. Will need to have a look at that as well. I thought with using the translate functionality on the batch I should be safe but doesn't seem so 😅


Quillraven:
I went now a different route and updated the already existing `renderOffset` functionality by adding the parallax offset here as well. This has the advantage that I don't have to break the batching and also it seems to solve the issue with the clipping I had before.

I also tested it now with all map options (iso, ortho, hex) and it looks good to me. I used the `TiledMapLayerOffsetTest` for that.

lyze237:
That's a great PR! 

Though from previous discussions about tile/tiledmap stuff ( https://github.com/libgdx/libgdx/pull/6534#issuecomment-840171842 ) libGDX members don't really want to add Tiled Editor (TMX) specific elements to the base classes (For example MapLayer.java) so I guess an alternative to that is to add that feature to `TiledMapTileLayer` instead.

Quillraven:
@lyze237 I can do that but I have a question for that change. In Tiled you can have something like this:

image

This is a grouping of layers and the final parallax scrolling factor is then the multiplication of the factor of the layer itself and all of its parents.

I see there is a `MapGroupLayer` but it is not part of the `tiled` package. It is outside on the same level as `MapLayer`. However, it seems like a new instance is only created by the `BaseTmxMapLoader`. So I assume it is Tiled specific?

Since there is no common parent for Tiled specific layers I need to duplicate the parallax properties in `TiledMapTileLayer`, `MapGroupLayer` and `TiledMapImageLayer`.

I have no problem with that and could make those changes but I should I then move the `MapGroupLayer` inside the `tiled` package and rename it to `TiledMapGroupLayer`? Or should I create a new `TiledMapGroupLayer` class?

If you ask me I argue that a parallax scrolling factor should be part of the general `MapLayer` since it is a generic property which can be implemented by any map editor. I only know Tiled and cannot judge on Tide but if Tide ever supports that as well then we anyway need to get it to the generic `MapLayer` somehow.

But don't get this the wrong way. I am not an expert with the LibGDX repository and the reasoning behind certain decisions. If it is better for you to keep everything separate then as mentioned above, I can do the adjustments but please provide me with some answers to my questions :)

Thank you!

lyze237:
Hey! @MobiDevelop! Sorry for the ping.

Since you helped with previous tile d related PRs, could you check this one out quickly and see what's the best way to add that?

MobiDevelop:
I think adding parallax properties/functionality on the base MapLayer is fine. It's a pretty common/generic need whether you use Tiled (editor) for making maps. If I recall, Tide does/did have parallax support.

Quillraven:
Thanks for the feedback! Let me then do some final adjustments and testing please. I try to finish it today or tomorrow and keep it in the MapLayer.

Quillraven:
hey boys&girls - happy new year!

would be great if someone can merge that or give at least any feedback because it got broken over the last few weeks due to changes to the master and maybe changes to this formatting check stuff?

I tried to update to latest master and run the "spotlessApply" gradle task as suggested. This basically changes the entire repository for me so I only committed the files that I changed and this seems to fix the issue of the spotlessCheck task.

Not sure if this is the best way to do it but I personally am anyway not a big fan of using custom formatters and be very strict about it :P

Anyway, please let me know if this is fine and if it can be merged or if I need to do something different. I would like to close this topic please :)

tommyettinger:
I ran the automated checks and it passes there, so the spotless changes worked correctly. @lyze237 since you know a lot about the Tiled system, can you maybe look the current changeset over? This seems like a good feature to have if it's been fixed in the current Tiled release.

lyze237:
I've checked it but not tested it so far as I don't have much time currently. Looks solid though, good job!

lyze237:
(Still had no time to test)

From discord: 
> are you sure parallax factor is correctly added to the layers here? i tested it, but it works kind of in reverse to tiled map editor parallax settings - parallax scroll in game is different than in tiled map editor. like slow layers are fast, and fast layers are slow. or am i missing something?

Quillraven:
I like "are you sure?" questions haha

Of course I am not sure because I am not an expert on that topic as mentioned in the initial issue. I tried to implement the logic that the Tiled guy mentioned in the issue (mapeditor/tiled#3196) I created for the Tiled repository because the documentation was not clear.

Also, I tested it to the best of my knowledge by comparing the Tiled view with the LibGDX view by:
- moving the view to 0,0 which results in no parallax movement
- moving the view to some other points and compare it always with the Tiled view

I did not check in detail how it moves. I just made sure that it moves in the correct direction and that the result when reaching my point looks the same in LibGDX and Tiled. That is not easy in my opinion as I don't see a camera position or view information in Tiled.

However, I tested it with an older version of Tiled. I saw that 1.8 was released which contains the Bugfix for parallax factor inheritance.

I need to test that again with this version but currently I do not have time, sorry.

If anyone wants to help feel free! Also, if anyone has a good idea how to test that properly please also let me know. For me it is just a "random manual" testing.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -160,6 +160,25 @@ protected TiledMap loadTiledMap (FileHandle tmxFile, P parameter, ImageResolver
 			Element element = root.getChild(i);
 			loadLayer(map, map.getLayers(), element, tmxFile, imageResolver);
 		}
+
+		// update hierarchical parallax scrolling factors
+		// in Tiled the final parallax scrolling factor of a layer is the multiplication of its factor with all its parents
+		// 1) get top level groups
+		final Array<MapGroupLayer> groups = map.getLayers().getByType(MapGroupLayer.class);
```
MobiDevelop, line 167: I'm wondering if it's more appropriate to compound the parallax factors in the renderer, rather than in the loader. It just seems easier than the bookkeeping involved in keeping them in sync (given they are mutable).

Quillraven: Do you mean to calculate it each render call? I thought about it as well but did not like it because it is a immutable value and only needs to be calculated once in my opinion which makes sense in the loader?
Also, theoretically you can have multiple nestings I think and going up the hierarchy to get all parent factor values for the final value of the leaf layer is most likely worse performance wise.

However, you are right that right now it is a mutable property which should not be the case maybe. But making it a constructor parameter to make it immutable might be tricky because we already need the final value including the group inheritance.

mgsx-dev: i don't think that calculating global parallax factor is so costly. It's about few float multiplications which is nothing compared to layer rendering. I think people will definitely want to have it mutable and they probably don't want to calculate global parallax them-self when changing it for a single layer, so i think it would be better to consider layer parallax property to be "local parallax" and have "global parallax" calculated on the fly at rendering time as suggested by MobiDevelop. What do you think?

Quillraven: Sure, I can change that! Just be aware - if I remember correctly it is not just float multiplication. It is also the lookup of those values and if I remember correctly this can be achieved via the parent property of a MapLayer. And it is not limited to just one "jump" because GroupLayers can be nested and therefore multiple "parent.getParallaxFactor" calls can happen each frame.

As you said though it is most likely not that costy and allows then also for programmatic changes to the factor values if the user wants that.

I think it is a good idea 👍

MobiDevelop: Maybe we can have an accumulator that sits at the root and every time you run into a group layer you do apply it's factor (and unapply it after the group). That way you're not looking up parent layers down the chain.

Quillraven:
Just updated now to 1.7.90 pre-release of Tiled and made some Tests again:

1) Disable Parallax rendering in Tiled and move view in Libgdx to 0,0 which should result in no parallax scrolling.
Seems to match for me:
image

2) Tried to move the view left/right/up/down and the cloud and the green pillar move in the same directions in Tiled and LibGDX

3) Moved the left edge of the green pillar to the beginning of the right slope in Tiled/LibGDX. In general the scene and especially the cloud position seem to match (reference area highlighted in red):
image

I changed the ortho.tmx map locally to use some factor values on the group layers to test the inheritance as well. Before this delivered a different scene output but now it matches.

----

Would be interesting for me to understand what the discord people meant with it behaves differently/in reverse. Maybe my testing approach is not ideal ;)

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
Quillraven: In general I like your idea @MobiDevelop but I see the problem that the rendering code seems to be copy&pasted in multiple renderers and therefore we'd need to introduce this accumulator handling in multiple places? Or, refactor it and extract the beginning part of `renderTiledLayer` / `renderImageLayer` which seems to be always the same in every renderer?

However, I don't want that to be part of this PR because it is more effort and riskier and I don't feel comfortable with that, sorry.

----

What I can offer for now is something like this in `MapLayer`:

```Java
/** @return layer's parallax scrolling factor for x-axis including the factor of all its parents */
	public float getGlobalParallaxX () {
		if (parent == null) {
			return parallaxX;
		}

		float acc = 1f;
		MapLayer curr = parent;
		while (curr != null) {
			acc *= curr.getParallaxX();
			curr = curr.parent;
		}
		return parallaxX * acc;
	}
```

And then adjust the renderer lines like:

```Java
// from this
final float layerOffsetX = layer.getRenderOffsetX() * unitScale + viewBounds.x * (layer.getParallaxX() - 1);
final float layerOffsetY = -layer.getRenderOffsetY() * unitScale - viewBounds.y * (layer.getParallaxY() - 1);

// to this
final float layerOffsetX = layer.getRenderOffsetX() * unitScale + viewBounds.x * (layer.getGlobalParallaxX() - 1);
final float layerOffsetY = -layer.getRenderOffsetY() * unitScale - viewBounds.y * (layer.getGlobalParallaxY() - 1);
```

What I personally don't like here is that the parent loop actually needs to be executed twice which is not necessary if we introduce a `Vector2`. Maybe a static Vector2 instance in `MapLayer` to achieve something like this?

```Java
// MapLayer
private static final Vector2 TMP_VEC = new Vector2();
	public Vector2 getGlobalParallaxFactor () {
		TMP_VEC.set(1f, 1f);
		if (parent == null) {
			return TMP_VEC;
		}

		MapLayer curr = parent;
		while (curr != null) {
			TMP_VEC.set(
				TMP_VEC.x * curr.getParallaxX(),
				TMP_VEC.y * curr.getParallaxY()
			);
			curr = curr.parent;
		}
		return TMP_VEC;
	}

// and then in the renderer we can call this only once
final Vector2 globParallax = layer.getGlobalParallaxFactor()
final float layerOffsetX = layer.getRenderOffsetX() * unitScale + viewBounds.x * (layer.getParallaxX() * globParallax.x - 1);
final float layerOffsetY = -layer.getRenderOffsetY() * unitScale - viewBounds.y * (layer.getParallaxY() * globParallax .y - 1);
```

What do you think @MobiDevelop  / @mgsx-dev ?

BoBIsHere86:
It's been well over a year since the last update in this thread. [Quillraven](https://github.com/Quillraven) seems to of done a lot of great work. For it to just be sitting her stagnant is sad. We are now on a Tiled 1.9.2 release, so who knows if this would have any effect. 
Where do we stand on this, are we just looking for more people to test and give feedback?

Quillraven:
Hey,

I think people on discord mentioned that it is not behaving like it should but unfortunately noone ever answered my questions above.

According to my knowledge it was working the way it should but imo someone "bigger" in the LibGDX community should judge on it. Also, I think recent Tiled Updates did not impact this feature meaning it should still work.

Would be great if this makes it into the next official version!

lyze237:
I just looked through the examples and isn't the paralaxx in wrong direction?

More specifically take a look at the green "hill" and the cloud above it:

In Tiled the hill is moving to the right and the cloud is moving to the left:

https://user-images.githubusercontent.com/11274700/214530941-e7efebbd-8a49-416a-bd87-60d03b23749e.mp4

While in the example you provided it's the other way around, the hill is moving to the left and the cloud is moving to the right:

https://user-images.githubusercontent.com/11274700/214530935-38a0ede1-b35c-4cc3-8c8d-b0e8a9599a4a.mp4

Quillraven:
So you mean the direction needs to be the opposite? Tbh this PR is so old and unfortunately I won't have time now nor in the near future to look into it again.

Would be great if you can fix the PR when you have time. Maybe you just need to scroll the opposite direction than I do and it is already fine?

Also, I remember that Tiled had a bug with its parallax feature that got fixed in some version. Make sure to use the latest version for comparison.

BoBIsHere86:
Hey Guys, so some good news. I think the issue here was a combination of a single + sign where it should of been a - sign. And illusion caused by the slow moving objects.
I've only tested this in Ortho. But I made my own map from scratch and dropped it in to better help test. I changed this line of code right here: https://github.com/Quillraven/libgdx/blob/a59b49c5ab081c06e660b5b0cce5f742180e24fd/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthogonalTiledMapRenderer.java#L77
and changed it to have a minus instead like so:
`final float layerOffsetX = layer.getRenderOffsetX() * unitScale - viewBounds.x * (layer.getParallaxX() - 1);`

I was then able to get results that looked exactly like I saw in tiled. Here's my example:
This is in tiled, Ignore the Tint color differences.
tiledParallax

This is from the libGDX test:
libGDXParallax

As you can see everything actually moves opposite of the camera now.
Let me know what you think, in case I might of missed something. But feel free to try it out.

Quillraven:
cool stuff! Can you also verify the y-axis once more? Just to be safe. And most likely it is the same mistake in the other renderers as mentioned by @lyze237 above.

Thanks for helping out :)

BoBIsHere86:
> cool stuff! Can you also verify the y-axis once more? Just to be safe. And most likely it is the same mistake in the other renderers as mentioned by @lyze237 above.
> 
> Thanks for helping out :)

Totally, when I have some time either tonight or tomorrow I'll try and test out everything that's been mentioned. Also, I'll try out the Y axis as well and report back. Luckily, I had just finished my tiled color tint PR. So I was in a good spot to help out here.

BoBIsHere86:
Alrighty, So I'm back with some results. And it's looking pretty good. 

But first I would just like to say I think I realized why this is so hard to test and what added to all the confusion. Since default parallax is 1 and as Tiled docs say, 1 is static, anything lower than 1 would cause the layer to move ALONG with the camera, as opposed to the what is normally expected which is that the parallax objects move opposite of the direction the camera is moving. And since the parallax layers were set to 0.75  on top of needing to reverse the math AND it just being very hard to eyeball the results. Made for one hard test to verify.

For all my tests I altered all of the parallax factors on the maps to be 1.25 on both the X and Y coordinates and used the test maps provided and then my 1 map I made. I will show both what Tiled Renders, then libGDX.

For each map I will simulate: Camera goes to the Right, then to the Left, Then UP, then Down. So we should keep an eye out to make sure Camera goes left, parallax layers go right. Camera Goes up, parallax layers go down, etc. 

Also, for certain maps which are hard to follow, I explicitly mention which layer is static.


First Test: Custom ortho map:
Tiled:
tiledParallaxOrtho

libGDX:
libGDXParallaxOrtho

Second: Isometric map (Note: The blue water layer is the parallax layer, Green is the static layer.)
Tiled:
tiledParallaxISO

libGDX:
libGDXParallaxISO

Third: Iso Staggered ( The Blue Layer is the STATIC Layer here, The Green Stripes of land are Parallax ,it gets confusing to see sometimes.)
Tiled:
tiledParallaxISOSTAG

libGDX:
libgdxParallaxISOSTAG

Fourth: Hex (The Green and Orange Hexs are STATIC, the Blue Layer is the Parallax) 
Tiled:
tiledParallaxHEX

libGDX:
libgdxParallaxHEX

Anyways, that's the extent of my testing for now. If anyone wants to try there hand at it. Or maybe I missed something. Let me know.

BoBIsHere86:
One more follow up to this Thread.
The OrthoCachedTiledMapRenderer just does not work with the Parallax we have it now. I looked into it and it's because over here:
https://github.com/Quillraven/libgdx/blob/a59b49c5ab081c06e660b5b0cce5f742180e24fd/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthoCachedTiledMapRenderer.java#L117
The cached flag is only false on the first run through. So renderTileLayer() only gets called once and never again. So the map parallax effect never updates.

I have never used OrthoCachedTiledMapRenderer, but I do see that the cache will become false either through forced call of invalidateCache() or in the setView() method when the bounds are exceeded.

That being said based on how I "think" this caching system works. I actually think that we should change this:
https://github.com/Quillraven/libgdx/blob/a59b49c5ab081c06e660b5b0cce5f742180e24fd/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthoCachedTiledMapRenderer.java#L220
To base the parallax off of the viewbounds instead of the the cachebounds anyway, since it's rendering doesnt match up with tiled or any of the others when using it.
So I'd suggest it become: 'final float layerOffsetX = layer.getRenderOffsetX() * unitScale - viewBounds.x * (layer.getParallaxX() - 1);'  etc...

In my repo when testing these changes and Modifying the test to FORCE a call to invalidateCache every frame, we can actually get it to work the same as all of the other renders.

That being said, Orthcache seems out of date anyway and as you've stated doesnt even support layers. So I don't know if we would want to just skip it. :-/

Quillraven:
I think it doesn't harm to add the functionality to the Orthocache renderer as well and from what I understood you are correct that you need to call invalidateCache before rendering to get the correct result. As mentioned at the beginning of this topic (somewhere ;) ), I'd anyway suggest that this rendering logic gets refactored because it seems to be copy&pasted to the different renderers which makes it then a little bit tidious to bring in features like this one.

But I assume that this is not going to happen in the near future as this also can be a risky move to introduce bugs. So I think having it like we do it now is good enough ... just my personal point of view.

Anyway, thanks for helping out @BoBIsHere86 ! Looks good to me. I requested a re-review of the two reviewers from before. Let's hope they give it a look soon so that doesn't have to lay around another year or so :)

BoBIsHere86:
> I think it doesn't harm to add the functionality to the Orthocache renderer as well and from what I understood you are correct that you need to call invalidateCache before rendering to get the correct result. As mentioned at the beginning of this topic (somewhere ;) ), I'd anyway suggest that this rendering logic gets refactored because it seems to be copy&pasted to the different renderers which makes it then a little bit tidious to bring in features like this one.
> 
> But I assume that this is not going to happen in the near future as this also can be a risky move to introduce bugs. So I think having it like we do it now is good enough ... just my personal point of view.
> 
> Anyway, thanks for helping out @BoBIsHere86 ! Looks good to me. I requested a re-review of the two reviewers from before. Let's hope they give it a look soon so that doesn't have to lay around another year or so :)
> 
> edit: btw do I need to do something with my fork? I am not very familiar with the github merging workflow and the forks. If I need to do anything, please let me know.

I'm no git expert, I was hoping I could just propose changes in the PR for you to accept but I guess not. 
But I think you just need to make the changes I mentioned in your featured dev branch and then recommit them.
The changes being:
Going to all of the renders where you have these lines :https://github.com/Quillraven/libgdx/blob/a59b49c5ab081c06e660b5b0cce5f742180e24fd/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthogonalTiledMapRenderer.java#L77

You need to change the ' unitScale + viewBounds.x * (layer.getParallaxX() - 1);' portion to use minus like this:
'unitScale - viewBounds.x * (layer.getParallaxX() - 1);'

That was the big oversight I noticed which caused these issues.

Then if we wanted to work on the Orthocache I think we need to change :
https://github.com/Quillraven/libgdx/blob/a59b49c5ab081c06e660b5b0cce5f742180e24fd/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthoCachedTiledMapRenderer.java#L220

To used viewBounds instead of cacheBounds:
`final float layerOffsetX = layer.getRenderOffsetX() * unitScale + viewBounds.x * (layer.getParallaxX() - 1);
// offset in tiled is y down, so we flip it
final float layerOffsetY = -layer.getRenderOffsetY() * unitScale - viewBounds.y * (layer.getParallaxY() - 1);`

But that being said, in order for this to show correctly in the test, you would need to add a call to invalidate the cache when using that renderer.

Maybe I can just fork your fork, and see if I can just send you a PR with the changes to make this less painful.
I'll take a look.

Either way this PR is a great addition it will be good to try and have libGDX fully up to date with Tiled's newest features.

SimonIT:
Just cherry-picked the commits from Quillraven/libgdx#1. I hope I didn't forget something

BoBIsHere86:
Awesome, Looks like you got everything. Great to see this moving along.

Quillraven:
So I don't need to do anything here? Thank you again for the support guys!

If I still need to do something then let me know. I should have some time on the weekend 🙂

PR #6717, opened 2021-12-06
Add config to set preferred fps to max supported by device

obigu:
There's some iOS devices with 120Hz screens (iPhone 13 Pro, iPad Pro).

`IosApplicationConfiguration#preferredFramesPerSecond` provides a mechanism to pass the hint with the max fps the app should render. In theory and from what I understand from the documentation, if we set it to `120`, it would fall back to `60` on most devices that have 60Hz screens. I think a better solution is to have the option to set it to the maximum supported fps on each device and this is what I've implemented (use `0` for max).

A couple of remarks:
1. I've set the `preferredFramesPerSecond` to `0` (max) by default (Breaking change??).
2. The max setting will work on iOS 11+ even if the API is available since 10.3 (some reports say it also requires an Info.plist entry in that case and the affected devices is minimal (iPad Pros with iOS 10.3 still).


MrStahlfelge:
I think it would be better to remain at a default of 60. 120 Hz might break some games, unseen by the developer because test devices supporting it are rare.

obigu:
> I think it would be better to remain at a default of 60. 120 Hz might break some games, unseen by the developer because test devices supporting it are rare.

On Android "max" is default and I think it's good to keep it consistent. 

Regarding "breaking", performance wise, it's not probable a game breaks due to not being able to keep up with the extra work because the devices that support 120Hz are comparatively much more performant than the rest. To be honest I think this problem may be even more accentuated on Android and there's no limitations there.

badlogic:
This is a breaking change and should be documented in CHANGES before a merge is considered. I too think that going for the max refresh rate by default is OK and unlikely to break existing games. With an entry in CHANGES we have deniability!

obigu:
CHANGES added!

Little warning notice. This change may not unlock maximum refresh rate on all devices that support over 60fps until we migrate to Metal. Apparently, the behaviour on iPad Pro is different than on iPhone devices that are more conservative (probably due to battery concerns). There's no clear documentation and don't have an iPhone 13 to test but I've seen a user reporting his game on OpenGL didn't run on his iPhone 13 at 120Hz despite hinting it to.

obigu:
With the new MetalANGLE backend the possible OpenGL 60fps limitation is now gone. Unfortunately I don't have a >60fps device to test on.

obigu:
It's been open for a while and low risk, merging now.

PR #6718, opened 2021-12-07
gwtref: Throw ClassNotFoundException instead of RuntimeException in ReflectionCache.forName

### LINKED ISSUE #6708: Wrong exception type thrown in GWT backend ReflectionCache.forName
Documentation for `reflect.ClassReflection.forName` states that it throws `ReflectionException` and does so in native java implementation
https://github.com/libgdx/libgdx/blob/0acff984c84cf7a8765603865dadbe7ba987c8cd/gdx/src/com/badlogic/gdx/utils/reflect/ClassReflection.java#L29-L31
However, the GWT implementation throws a `RuntimeException`
 https://github.com/libgdx/libgdx/blob/0acff984c84cf7a8765603865dadbe7ba987c8cd/backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/ReflectionCache.java#L49-L51

This is an inconsistency and GWTs implementation should follow the doc.

This can probably be fixed by gwt reflection cache throwing a `ClassNotFoundException`, which would get caught by 
https://github.com/libgdx/libgdx/blob/0acff984c84cf7a8765603865dadbe7ba987c8cd/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/reflect/ClassReflection.java#L30-L34

- [ ] Android
- [ ] iOS
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

nojus297:
Not tested.

Should fix #6708

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/reflect/ClassReflection.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/reflect/ClassReflection.java
@@ -30,7 +30,7 @@ public final class ClassReflection {
 		try {
 			return ReflectionCache.forName(name).getClassOfType();
 		} catch (ClassNotFoundException e) {
-			throw new ReflectionException("Class not found: " + name);
+			throw new ReflectionException("Class not found: " + name, e);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


MobiDevelop:
Tested this, it has the desired effect of throwing the ReflectionException from ClassReflection.forName on GWT; however, there's an inconsistency in that it doesn't include the cause when re-throwing as ReflectionException. Would you mind updating the emulation of ClassReflection to pass along the cause, @nojus297?

nojus297:
@MobiDevelop updated, if I understood you correctly.

PR #6719, opened 2021-12-08
Lwjgl3Net: Fallback to xdg-open on Linux if Desktop.BROWSE is unavailable

kyu-n:
When using a libGDX application inside of a linux sandbox container (Snap/Flatpak), Net#openURI will fail to open any url because the Desktop.BROWSE action is unavailable. This is intended behavior of the sandbox. You're supposed to use xdg-portals to access urls in these environments.

The simplest way to workaround this limitation is to call xdg-open for the target uri. There's no guarantee it'll be able to handle your uri if a mime type is undefined for the sandbox, but it'll work most of the time. This binary is standard on freedesktop platforms.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
@@ -73,14 +73,21 @@ public class Lwjgl3Net implements Net {
 			} catch (Throwable t) {
 				return false;
 			}
-		} else {
+		} else if (Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
 			try {
 				Desktop.getDesktop().browse(new URI(uri));
 				return true;
 			} catch (Throwable t) {
 				return false;
 			}
+		} else if (SharedLibraryLoader.isLinux) {
+			try {
+				(new ProcessBuilder("xdg-open", (new URI(uri).toString()))).start();
+				return true;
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
```diff
@@ -73,14 +73,21 @@ public boolean openURI (String uri) {
...
+		} else if (Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
 			try {
 				Desktop.getDesktop().browse(new URI(uri));
 				return true;
 			} catch (Throwable t) {
 				return false;
 			}
+		} else if (SharedLibraryLoader.isLinux) {
+			try {
+				(new ProcessBuilder("xdg-open", (new URI(uri).toString()))).start();
```
noblemaster, line 85: Maybe `new URI(uri).toString()` can just be `url`?

kyu-n: This appears messy, but it has a purpose. Running the `uri` through a `URI` builder sanitizes the input to prevent arbitrary code execution on a client system by other players in online environments.

See https://github.com/libgdx/libgdx/pull/6543#issuecomment-848766738 for details

noblemaster: That makes sense. Good to keep it!

PR #6721, opened 2021-12-08
Update to LWJGL 3.3.0

### LINKED ISSUE #6716: Update lwjgl3 to 3.3.0 for M1 Mac support
LWJGL 3.3.0 was released on Nov 15 and required changes seem to already be in, just commented out via https://github.com/libgdx/libgdx/commit/7ac239c81f7d126c3be5ad18c25bcdb2ff811fd5
Perhaps it's time to uncomment them? Upgrade seems to work as expected for me and it not working on new Macs is quite a poor look.

PokeMMO:
Closes #6716


tommyettinger (changes requested):
These lines should also be changed:
https://github.com/libgdx/libgdx/blob/master/fetch.xml#L13-L14
I guess we still use Ant in places.

tommyettinger (approved):
Looks good! I'm not 100% sure of what jnigen does in this same file (line 11), but should it be changed to 2.2.1? That might be better in a separate PR, unless it's related here.

Tom-Ski (approved):

PokeMMO:
The only changes in jnigen 2.2.1 were in the gradle plugin, which this doesn't use.

PR #6722, opened 2021-12-08
fix bad camera reference by using instance variable instead of parameter

mgsx-dev:
Camera reference was wrong with some constructor, it uses the camera from parameter which is no longer valid when changing camera later.
note that It's not possible to reference instance variable when calling another constructor, so i had to move it to the final constructor.


yuripourre (approved):

obigu:
I'm not familiar with the `CameraSorter` but I understand the new default `CameraGroupStrategy (Camera camera)` constructor won't set the `sorter`, can this break backwards compatibility? The constructors seem very simple, would it make sense in this case to just duplicate the logic and avoid calling one from the other?

mgsx-dev:
@obigu you're right, i think it's more clean now, thank you.

PR #6725, opened 2021-12-10
Gradle publishing support

PokeMMO:
Should match maven for artifacts, including `$NAME-platform` artifacts for jnigen natives.

TODO:
- [x] Javadocs from GWT are broken somehow. I commented out the javadocs jar for now.
  - https://github.com/libgdx/libgdx/pull/6725/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R98
- [x] Developers listing in pom. I just copy/pasted from gdx-jnigen.
- [ ] Verify signing and release to sonatype. I've only tested `./gradlew publishToMavenLocal`
- [x] Populate `POM_DESCRIPTION` in `gradle.properties` for each project.
- [x] Decide if we care about parent/child module links in pom. I don't think these do anything?
  - https://github.com/libgdx/libgdx/blob/master/extensions/gdx-box2d/pom.xml#L15-L19
  - https://github.com/libgdx/libgdx/blob/master/extensions/gdx-box2d/gdx-box2d/pom.xml#L4-L9
  - My understanding is these are not required since gradle uses absolute version numbers instead of variables for dependencies.
- [x] Fully test and verify everything matches.
  - gdx-backend-android provides no dependencies in pom. This is because we don't use a proper android-library. We can fix this after we move to gradle publishing and merge #5976
  - gdx-backend-android from maven publishes everything as `optional` anyway, so I dont think this changes anything.
  - gdx-box2d-gwt no longer has a depdency on `gdx-box2d` exported. The project in IDE doesn't seem to depend on it anyway. So I'm assuming this is fine.
  - gwt projects no longer have a `provided` `gwt-dev` dependency listed. We use `compileOnly` which should be equivilent.


build.gradle
```diff
@@ -69,6 +93,15 @@ configure(subprojects - project(":tests:gdx-tests-android")) {
...
         options.incremental = true
     }
+    
+    java {
+        /*withJavadocJar()*/
+        withSourcesJar()
+    }
+    
+    javadoc {
+        options.addStringOption('Xdoclint:none', '-quiet')
```
tommyettinger, line 103: I guess we're requiring Java 8 to build now! 🎉 From what I can tell (given that the Java 7 man pages are no longer available on Oracle's site), the doclint option for javadoc was added in Java 8.

In my projects, I use the code:
```groovy
if (JavaVersion.current().isJava8Compatible()) {
    allprojects {
        tasks.withType(Javadoc) {
            options.addStringOption('Xdoclint:none,-missing', '-quiet')
        }
    }
}
```

tommyettinger (commented):
Looks good as long as we're OK with requiring Java 8 or higher to build. It's telling that Oracle has essentially eliminated all docs for Java 7 other than release notes and maybe API docs. If we can come to a consensus on updating to Java 8, maybe that should be in a separate issue, but it does affect the build tools in general.

build.gradle
PokeMMO: Ah whoops. I didn't realize that was Java 8 specific.

It's probably time to move to Java 8 to compile, but for now I'll add this safety check.

Thanks!

tommyettinger: Looks like you got this back to compatible with Java 7; I'll resolve this.

PokeMMO:
gwt's javadocs issue has been work around by specifying the `sourcepath` option.

I have also moved gwt-dev to a compileOnly dependency to more accurately match maven.

https://github.com/libgdx/libgdx/blob/42a59f2f7bbef736ef61df45ab89475473cffeff/backends/gdx-backends-gwt/pom.xml#L32

@MrStahlfelge can you verify this is correct?

Tom-Ski (approved):

PR #6727, opened 2021-12-11
Remove all usage of Arial assets, replace with LSans.

### LINKED ISSUE #5537: Arial font is proprietary and should be replaced.
#### Issue details
I believe Arial is a proprietary font. Doing a web search for "Arial license" comes up with places selling licenses for the font. It's [Wikipedia page](https://en.wikipedia.org/wiki/Arial) says it's proprietary, and that Liberation Sans is a "Metrically compatible" font.

Liberation Sans is licensed with SIL Open Font License v1.10. I think it would be OK to package into games. It can be downloaded from [fontsquirrel.com](https://www.fontsquirrel.com/fonts/liberation-sans). I tried it out with Hiero and libGDX and it looks pretty close to the same as Arial.

GitHub shows an error when I try to upload files, so here's a link to the Hiero project file and a screenshot comparing the results.

https://drive.google.com/drive/folders/1kzPyVWrpgerXjS2kCurI34VoUQbUz65N?usp=sharing

libGDX 1.9.9
Hiero v5

- [x] Android
- [x] iOS (robovm)
- [x] iOS (MOE)
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

tommyettinger:
This is meant to address the licensing of Arial, which is a commercial font, and replace it with a FLOSS and metric-compatible relative. LSans is an alternative name I chose for Liberation Sans, a font that's used often across Linux distros because LibreOffice defaults to it. Because "Liberation" is a reserved name for that font (ownership is held by Red Hat), any modified version needs to use a different name according to its OFL license. The arial-15 assets in every libGDX JAR have been replaced by same-image-sized, metric-compatible lsans-15 assets. The tests had some other Arial-based fonts that I think have been replaced accurately; there's some weird ones that aren't used in many tests, though. We may still want to do this again for Verdana (and maybe Garuda?), but it's much easier to replace those.

I have the Hiero files available still, so I can make changes if the shadow on the font needs adjusting. This looked extremely close to me.

This will resolve #5537 .


obigu:
Best thing is the new atlas is 6KB smaller!

I've run some libGDX tests on Lwjgl3 backend and have found a couple with apparent inconsistent behaviour.

1. `BitmapFontMetricTest` looks quite different (much more spaced characters), not sure if that's a big deal.
2. `FreeTypeMetricsTest` doesn't render all characters correctly.

tommyettinger:
I just looked into this, and the test-only lsans-32-pad.fnt was using the wrong padding info. I got it to use the same padding as before, and now BitmapFontMetricTest looks like:

image

And FreetypeMetricsTest looks like:

image

obigu (approved):

yuripourre:
@tommyettinger go for it!

yuripourre (approved):

tommyettinger:
Should there be a change marked in CHANGES? This does maybe break things that explicitly load arial-15.fnt, or that depend on the exact line width of text in the old font. I don't think it's a good idea to keep using the possible trademark(?) Arial, so we shouldn't even have an arial-15.fnt (even if it uses Liberation Sans).

yuripourre:
@tommyettinger that's a good point, it will potentially impact only a few projects (if any) but a message on CHANGES will at least give a hint for developers.

throwaway1037:
Are there still any other proprietary fonts that need replacing?

tommyettinger:
> Are there still any other proprietary fonts that need replacing?

None that are distributed with libGDX; only arial-15 was going into distributed JARs. One usage of Arial escaped me, but it's in the test assets -- multipagefont.fnt, multipagefont1.png, and multipagefont2.png use Arial. There are 3 .fnt files that use Verdana, also in test assets, plus some .ttf fonts that I don't know anything about (Garuda, Russkij, and Unbom). It's entirely possible I've missed some folder, but I don't know where that could be. I wouldn't be as concerned about test assets, personally, but of course it would be optimal to not have any commercially-licensed fonts in the codebase.

throwaway1037:
The word ["asset"](https://gnu.org/philosophy/words-to-avoid.html#Assets) is propaganda in this context.

The term ["commercial"](https://gnu.org/philosophy/words-to-avoid.html#Commercial) in this context is incorrect.

Nevertheless, all proprietary files should be replaced with libre replacements, which means Verdana has to go.
As far as I can tell, Garuda and Russkjj are proprietary too(?), and UnBom might be free(?)

tommyettinger:
> The word "asset" is propaganda in this context.

What context, pray tell, is this? You apparently haven't made a libGDX project before, or you would know that asset is actually the term used for internal resources here, so you're just trawling the open Internet looking for heresy against your idiosyncratic libre dogma? It looks like you're just spray-n-praying repos with legal junk -- https://github.com/yairm210/Unciv/issues/5770 https://github.com/angband/angband/issues/5176 https://github.com/sdiehl/write-you-a-haskell/issues/115 (bonus trolling from you there). Judging by your behavior on Mindustry ( https://github.com/Anuken/Mindustry/issues/6418 ), you need to take a chill pill.

Frosty-J:
Did this break `TextAreaTest3` or has it always been like this?

tommyettinger:
I'm comparing an older commit with the current one; the older commit (just before this was merged) had arial15.fnt use arial15_00.png (which is 1024x32 sized), while elsewhere in libGDX, arial15.fnt used arial15.png (which is the normal image). This isn't hard to fix and I'll put a PR in right away since I have it fixed locally. I just need to put the correctly-different lsans15.fnt into the test data, instead of the incorrectly-same one.

PR #6728, opened 2021-12-12
Check if we're on mobile!

Frosty-J:
Frequently I find myself fetching `GwtApplication.isMobileDevice()` and passing it into my main class. `Gdx.app.getType()` is unsuitable for this task, as it doesn't distinguish mobile web from desktop web. It's a bit of a hassle, so here's my proposal for not having to do that anymore. The code's a bit smelly, but hopefully functional - feel free to edit.

`GwtApplication.isMobileDevice()` still exists as to not break existing projects, but has been deprecated. I don't really like having two methods that do effectively the same thing, but don't see a better solution at the moment due to `isMobileDevice()` being `static`.

I've changed  `GwtApplication.isMobileDevice()` to return a `final` boolean rather than query the user agent each time the method is run. This is to ensure tip-top performance if someone calls the method each frame, such as to decide whether to draw touchscreen controls. The other backends use the same approach. It doesn't matter that it doesn't update, as the user agent shouldn't change while the game is running (every mobile browser I've used refreshes the page when switching between mobile/desktop modes).

GWT's regex for the mobile check remains completely overkill, seeing as libGDX isn't compatible with most of those devices/browsers, but I'd rather not touch it and inadvertently break something in the process.

Platform|Is mobile?
-|-
Web|Detects based on user agent like previously
Desktop|Never mobile
Android|Mobile unless running under ARC (e.g. Chromebook)
iOS|Always mobile

***

I'm having trouble properly testing my changes. The IDE doesn't recognise any libGDX classes when using `1.10.1-SNAPSHOT`, despite invalidating caches and deleting `.idea`. `superDev` just results in `User limit of inotify watches reached`. Yeah, not very helpful for you lot, but not something I can fix tonight (raising the system inotify limit didn't change anything).

Mild curiosity: `[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)` gets output twice for LWJGL3 desktop (Linux) on 1.10.1-SNAPSHOT, but the sound still works fine.

And the formatter doesn't understand the concept of hyperlinks.  :(

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@
 - API Fix: Resolved issues with LWJGL 3 and borderless fullscreen
 - API Addition: GeometryUtils,polygons isCCW, ensureClockwise, reverseVertices
 - API Addition: Added FreeTypeFontGenerator#hasCharGlyph method.
+- API Addition: Gdx.app.isMobile() can be used to determine if the user is on a mobile device.
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
@@ -62,8 +62,11 @@ public class AndroidApplication extends Activity implements AndroidApplicationBa
 	private int wasFocusChanged = -1;
 	private boolean isWaitingForAudio = false;
 
-	/* The device is mobile unless running under the Android Runtime for Chrome (e.g. on a Chromebook). Based on:
-	* https://github.com/google/talkback/blob/f5d564f/utils/src/main/java/com/google/android/accessibility/utils/FeatureSupport.java#L49-L51 */
+	/*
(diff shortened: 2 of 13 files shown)


MrStahlfelge:
The big question here is: What to use this for? It should be more interesting if multi touch is supported or a hardware keyboard. We already have methods for that. So what do you use this for?

Frosty-J:
`getMaxPointers()` always returns 20 on GWT (and Android/iOS), so in its current state that's no good for this purpose. I'm actually not sure what the point of that method is, since it makes no attempt to return the device's maximum (usually 10, I'd think; only 2 if we go back to devices that aren't supported by libGDX anymore).

I'm unaware of a libGDX method that returns if a hardware keyboard is present. Not saying there isn't one - I don't know all the ins and outs. And that gets complicated - a PC keyboard is very different from a [BlackBerry keyboard](https://i2.wp.com/www.nashvillechatterclass.com/wp-content/uploads/2016/06/Blackberry-Priv-Smartphone.jpg).

So I use this to determine if the user is on mobile and should be presented with language like "tap" instead of "click", on-screen buttons if necessary, that sort of thing. Because I haven't found a better way to do so.

MrStahlfelge:
You should probably use getPeriphalAvailable and check for multitouch, maybe multitouch combined with hardware keyboard depending of use case.

Frosty-J:
That's a pretty handy method. `Gdx.input.isPeripheralAvailable(Input.Peripheral.OnscreenKeyboard)` is effectively the same as what I've got here, bar the Android bit. Confirmed that it still returns `false` for PC users who use an on-screen keyboard. Not obvious you can do that, but yeah, this pull request is a bit unnecessary.

PR #6730, opened 2021-12-14
add back default Pool behavior when object is discarded

mgsx-dev:
as discussed in #6723, Pool change from #6379 was a breaking change. This PR add back previous behavior.


yuripourre (approved):
I remember that discussion.

crykn (approved):

obigu (approved):

PR #6731, opened 2021-12-16
Fixes some Gradle build issues (GWT Tests, Runnable Tools)

MobiDevelop:
Found some issues that look to be side effects of https://github.com/libgdx/libgdx/pull/6652.

The resources from the main gdx module were not being added as sources for the GWT Tests, which was preventing compile.
Runtime dependencies were not included in the runnable jars for tools (e.g. Hiero couldn't launch due to missing ApplicationListener class).


tommyettinger (approved):
Looks good to me; I've definitely hit the same issue while updating Gradle before (where `compile` is used in more places than just dependencies). I wonder how long the GWT tests have been non-working... I can confirm that the GWT tests without this PR are finicky at best, since they worked for me in IDEA but were totally non-functional from the command line.

PR #6732, opened 2021-12-18
Fix PixmapBlendingTest

yuripourre:
If I try to run the test for LWJGL3 as-is it gives me an error saying that cannot find test3.png or test4.png files.

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/PixmapBlendingTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/PixmapBlendingTest.java
@@ -16,9 +16,7 @@
 
 package com.badlogic.gdx.tests;
 
-import com.badlogic.gdx.Files;
 import com.badlogic.gdx.Gdx;
-import com.badlogic.gdx.InputProcessor;
 import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.g2d.Sprite;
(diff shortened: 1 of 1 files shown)


tommyettinger (commented):
I don't think this changes anything on LWJGL 2 or 3; if you go to the implementation of files.getFileHandle() for those platforms, it is the same code as would be called with files.internal(). Is this meant more for Android, iOS, or something else? The other clean-up is nice, though.

tommyettinger:
I should also say, I have just run the test without this PR on LWJGL 2 and 3 right now, so unless something changed in the latest code for Lwjgl(3)Files, I don't know if this is needed. It is cleaner, though.

yuripourre:
@tommyettinger I tried to run in Lwjgl3 and for some weird reason it couldn't find the files I mentioned. After this change, the test also worked on GWT. Would you mind trying to run this test on GWT? It is currently disabled.

tommyettinger:
I'm wondering if something changed after the commit I have checked out, since the test works without needing this PR (for me).
Working test screenshot

Thinking about it, though, this PR does just in general improve the code style, so **consider this an approval**. I'm just trying to figure out why it could be needed at all to run, since the code paths look identical for internal() and getFileHandle().

yuripourre:
Thank you for taking the time for testing. @tommyettinger I was not using the latest master indeed, maybe it was fixed at some point. I had some trouble loading files as bytes (it simply doesn't load) using GWT but I have no idea if it's related somehow.
I will start from a clean repository and try again later.

Thanks again!

tommyettinger:
Pixmap on GWT simply doesn't have a constructor that takes a `byte[]`. This is only visible if you look at the super-sourced code for Pixmap used on GWT, https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java . I'm not totally sure why this is, since there's a setPixels() that takes a ByteBuffer, but it calls into JSNI code.

Even though it might no longer be a "fix," I think this still makes PixmapBlendingTest more readable and more like the rest of libGDX's code. I won't close this unless you think the file should stay as-is.

yuripourre:
@tommyettinger agreed, let's merge this.

PR #6736, opened 2021-12-20
refactor: setWindowedMode - removed unnecessary check

marcin85:
This problem occurs only when Lwjgl3Window  is resizable (`GLFW.GLFW_RESIZABLE` is true).

Previously when I try to set windowed mode with given width and height then I got maximized window with omitted desired dimensions. Now window is properly sized and centered.

Done with this PR:

- add line `window.setSizeLimits(monitorX, monitorY, monitorWidth, monitorHeight) ` to fix issue.

additionally:

- remove misleading/unnecessary method call `storeCurrentWindowPositionAndDisplayMode();`
- minimal code refactor

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
@@ -55,8 +55,6 @@ public class Lwjgl3Graphics extends AbstractGraphics implements Disposable {
 	private int fps;
 	private int windowPosXBeforeFullscreen;
 	private int windowPosYBeforeFullscreen;
-	private int windowWidthBeforeFullscreen;
-	private int windowHeightBeforeFullscreen;
 	private DisplayMode displayModeBeforeFullscreen = null;
 
 	IntBuffer tmpBuffer = BufferUtils.createIntBuffer(1);
@@ -385,8 +383,6 @@ public class Lwjgl3Graphics extends AbstractGraphics implements Disposable {
 	private void storeCurrentWindowPositionAndDisplayMode () {
 		windowPosXBeforeFullscreen = window.getPositionX();
 		windowPosYBeforeFullscreen = window.getPositionY();
-		windowWidthBeforeFullscreen = logicalWidth;
-		windowHeightBeforeFullscreen = logicalHeight;
 		displayModeBeforeFullscreen = getDisplayMode();
 	}
 
@@ -409,15 +405,18 @@ public class Lwjgl3Graphics extends AbstractGraphics implements Disposable {
 			}
 		} else {
 			if (displayModeBeforeFullscreen == null) {
-				storeCurrentWindowPositionAndDisplayMode();
-			}
-			if (width != windowWidthBeforeFullscreen || height != windowHeightBeforeFullscreen) { // Center window
+				// Center window with given dimensions
 				Lwjgl3Monitor monitor = (Lwjgl3Monitor)getMonitor();
 				GLFW.glfwGetMonitorWorkarea(monitor.monitorHandle, tmpBuffer, tmpBuffer2, tmpBuffer3, tmpBuffer4);
-				GLFW.glfwSetWindowMonitor(window.getWindowHandle(), 0,
-					Math.max(0, tmpBuffer.get(0) + (tmpBuffer3.get(0) - width) / 2),
-					Math.max(0, tmpBuffer2.get(0) + (tmpBuffer4.get(0) - height) / 2), width, height,
-					displayModeBeforeFullscreen.refreshRate);
+				int monitorX = tmpBuffer.get(0);
+				int monitorY = tmpBuffer2.get(0);
+				int monitorWidth = tmpBuffer3.get(0);
+				int monitorHeight = tmpBuffer4.get(0);
+
+				window.setSizeLimits(monitorX, monitorY, monitorWidth, monitorHeight);
+				int newX = Math.max(0, monitorX + (monitorWidth - width) / 2);
+				int newY = Math.max(0, monitorY + (monitorHeight - height) / 2);
+				GLFW.glfwSetWindowMonitor(window.getWindowHandle(), 0, newX, newY, width, height, getDisplayMode().refreshRate);
 			} else {
 				GLFW.glfwSetWindowMonitor(window.getWindowHandle(), 0, windowPosXBeforeFullscreen, windowPosYBeforeFullscreen, width,
 					height, displayModeBeforeFullscreen.refreshRate);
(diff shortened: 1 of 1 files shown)


crykn:
Hey @marcin85! I can't seem to reproduce your issue (see below for my very simple test). Can you provide a piece of code that shows the problem? And on what OS are you? Are you using multiple LWJGL windows or just one?

<details>
  <summary><b><i>Click to expand</i></b></summary>
  
```java
public class IssueTest extends GdxTest {
	ShapeRenderer s;

	public static void main (String[] argv) {
		Lwjgl3ApplicationConfiguration config = new Lwjgl3ApplicationConfiguration();
		config.setResizable(true);
		config.setFullscreenMode(Lwjgl3ApplicationConfiguration.getDisplayMode());
		new Lwjgl3Application(new IssueTest2(), config);
	}

	@Override
	public void create () {
		s = new ShapeRenderer();
		Gdx.graphics.setWindowedMode(700, 350);
	}

	@Override
	public void render () {
		// Draw some stuff
		ScreenUtils.clear(Color.CHARTREUSE);
		s.begin(ShapeRenderer.ShapeType.Filled);
		s.circle(640 / 2, 480 / 2, 480 / 2);
		s.end();
	}

}
```
  
</details>

marcin85:
Hi @crykn :)
I used single LWJGL windows on Linux.

Part of my neofetch output:

OS: Manjaro Linux x86_64 
Kernel: 5.13.19-2-MANJARO 
DE: Xfce 4.16 
WM: Xfwm4 
CPU: Intel i7-4600U (4) @ 3.300GHz 
GPU: Intel Haswell-ULT 
Memory: 1457MiB / 15941MiB 

please check recorded video:
https://drive.google.com/file/d/10bhDk2XwdcHgLrIoE93PaWvkRp7qilZb/view?usp=sharing

scenario 1 : resizable=true; application started with windowed mode(840x524) -> changed to fullscreen -> application closed -> application started = bug is visible (840x524 is NOT applied, app is windowed, but stretched to fullscreen size) 

scenario 2 : resizable=true; application started with windowed mode(840x524) -> changed to fullscreen -> application closed -> application started = bug is  not visible (840x524 is applied)

crykn:
When you are setting the "fullscreen" toggle to "off" in your game, what is the method that is called behind the scenes?

marcin85:
I have this method called in render() to handle on/off fullscreen:
```

    public void resizeOnEnter() {
        if (Gdx.graphics.isFullscreen()) {
            Gdx.graphics.setWindowedMode(Config.WINDOW_WIDTH, Config.WINDOW_HEIGHT);
        } else {
            Gdx.graphics.setFullscreenMode(Gdx.graphics.getDisplayMode());
        }
    }
```

marcin85:
@crykn I recorded another video with your simple test:
https://drive.google.com/file/d/1fEYT4NSMh1hj3FhpANx7v-I5-ZbRq8FL/view?usp=sharing
Problem still occurring.
And I saw  `config.setResizable(true);` or  `config.setResizable(false);` does not affect issue. 
code:

```
public class IssueTestXX extends GdxTest {

    public static void main(String[] argv) {
        Lwjgl3ApplicationConfiguration config = new Lwjgl3ApplicationConfiguration();
        config.setResizable(true);
        config.setFullscreenMode(Lwjgl3ApplicationConfiguration.getDisplayMode());
        new Lwjgl3Application(new App(), config);
    }

    static class App extends ApplicationAdapter {
        ShapeRenderer s;

        public void create () {
            s = new ShapeRenderer();
            Gdx.graphics.setWindowedMode(700, 350);
        }

        @Override
        public void render () {
            ScreenUtils.clear(Color.CHARTREUSE);
            s.begin(ShapeRenderer.ShapeType.Filled);
            s.circle(640 / 2, 480 / 2, 480 / 2);
            s.end();
        }
    }
}

```

crykn:
Thanks! 

Setting the size limits (i.e., the minimum and maximum size of a window) before changing the window size should definitely not be necessary. On the contrary, by default, the size limits should be `GLFW.GLFW_DONT_CARE ` and thus aren't supposed to affect anything. Also, as stated above, I can't reproduce the bug on Windows, which seems to suggest that it is platform/window manager-specific. 

One possible problem I see with your solution of adding the `window.setSizeLimits(monitorX, monitorY, monitorWidth, monitorHeight)` line is this: `glfwGetMonitorWorkarea` returns the monitor area not occluded by the operating system task bar. If the maximum window size is set to this, this would prevent windowed fullscreen mode (i.e., undecorated windows that are the size of the _whole_ monitor, including the area of any task bar).

A better (and simpler) fix may be to use something like this in your desktop launcher:
```java
DisplayMode displayMode = Lwjgl3ApplicationConfiguration.getDisplayMode();
config.setWindowSizeLimits (1, 1, displayMode.width, displayMode.height)
``` 

Tbh, I'm reluctant to add a workaround for a platform-specific issue, when there is a decent chance this may cause problems for other (far more commonly used) platforms and there is a simple user side fix. However, other opinions on this are very welcome!

marcin85:
I agreed with you that this platform specific workaround is controversial/redundant. :) I proposed it as PR because somewhere in code I saw similar fix for mac platform.
If you like rest of the code I can remove line
`window.setSizeLimits(monitorX, monitorY, monitorWidth, monitorHeight);` and we can merge only as refactor.

marcin85:
one more thing:
Proposed fix:
```
DisplayMode displayMode = Lwjgl3ApplicationConfiguration.getDisplayMode();
config.setWindowSizeLimits (1, 1, displayMode.width, displayMode.height)
```
not fix completely my finding. With it you cannot oversize window over displayMode.width, displayMode.height, On the other hand when you set fullscreen size, the bug will is reproduced in the same way.

crykn:
I don't think the proposed changes have the same behaviour as the old code. For example, when I start a windowed application, and then set it to fullscreen, `displayModeBeforeFullscreen` is set via `storeCurrentWindowPositionAndDisplayMode()`. This means that when I now make the application windowed again, the existing code would execute the centering stuff, while the code in the PR would skip it and jump to the else clause.

marcin85:
hi:)
Yep, the _if case_
```
			if (displayModeBeforeFullscreen == null) {
				storeCurrentWindowPositionAndDisplayMode();
			}
```
is romoved, but imho is not necessary, also the method name `storeCurrentWindowPositionAndDisplayMode();` in this case is wrong.
There is no current window position, because we started with fullscreen( `displayModeBeforeFullscreen == null`).

also displayModeBeforeFullscreen = getDisplayMode(); is wrong because there was no "before fullscreen".

The method `storeCurrentWindowPositionAndDisplayMode()` is used properly in `setFullscreenMode ()`

marcin85:
@crykn can we merge this PR ?

crykn:
As I wrote above, the PR isn't just a refactor, it changes some of the old behaviour. Tbf, the changes only affect very specific cases, but the new behaviour is still different and in my opionion (others are welcome to chime in on this!) not necessarily better.

marcin85:
@crykn I named this refactor because logic isn't changed in my opinion. please check my previous comments. I just removed unnecessary/misleading code. but of course change its not necessary.

crykn:
>  ... because logic isn't changed 

But it is! See my comment above:

> [W]hen I start a windowed application, and then set it to fullscreen ... [and then] make the application windowed again, the existing code would execute the centering stuff, while the code in the PR would skip it and jump to the else clause.

While at first it seems unlikely that the windowed/fullscreen mode is changed that often, it is definitely possible. For example, when I start my game in windowed mode by default, then read the config file and change the game to fullscreen according to the users preferences, this issue would happen as soon as the user wants to switch to windowed mode again.

marcin85:
@crykn You are right when windows size will be changed in second windowed mode then the same position will be set, but without centered. You can decline Pr. thanx

PR #6738, opened 2021-12-20
Add Pixmap constructor from bytes for GWT

yuripourre:
Hey everyone, I need real help here, how can this PR look better? @lyze237 @MrStahlfelge @tommyettinger @mgsx-dev

This is how I manage to load Pixmap from bytes, I tried hard to load them sync but didn't work. I added a "loaded" attribute to pixmap (both GWT and non-GWT) it can be ignored on non-GWT but maybe could be useful for async load operations.

Thank you @lyze237 for the gist.


MrStahlfelge:
Just to check, this is not for something that's already done by `Pixmap.downloadFromUrl()`?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -633,4 +697,10 @@ private void fillOrStrokePath (DrawType drawType) {
 
 		void downloadFailed (Throwable t);
 	}
+
+	public interface LoadImageAsyncListener {
```
tommyettinger, line 701: This probably doesn't need to be public, unless it is intended for external use. The only usage I can see is fully within code here.

backends/gdx-backends-gwt/res/com/badlogic/gdx/backends/gdx_backends_gwt.gwt.xml
```diff
@@ -1,13 +1,14 @@
...
+	<inherits name='com.google.gwt.typedarrays.TypedArrays'/>
 	<!-- Inherit edited chrome theme ("gwt"-prefixed classes only) for a little bit of default styling in the text input dialogs -->
 	<inherits name='com.badlogic.gdx.backends.gwt.theme.chrome.Chrome'/>
-	<inherits name="com.google.gwt.http.HTTP"/>
-	
-	<inherits name="com.badlogic.gdx" />
-	<inherits name="com.google.gwt.webgl.WebGL" />
+	<inherits name='com.google.gwt.http.HTTP'/>
+
+	<inherits name="com.badlogic.gdx"/>
```
tommyettinger, line 10: I thought inheriting "com.badlogic.gdx" would transitively inherit TypedArrays, but I didn't ever try using TypedArrays from this package in the emu (it is used by NumberUtils).

yuripourre:
> 
@MrStahlfelge I don't think they are the same. This logic loads image from bytes, it is not the same as downloading from a server, they can be generated locally (client-side). I am not sure if internally they work the same way.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -128,6 +153,41 @@ private native static void setImageData (ArrayBufferView pixels, int width, int
...
 
+	public boolean ready = false;
+
+	/** Creates a new Pixmap instance from the given encoded image data.
+	 *
+	 * @param encodedData the encoded image data
+	 * @param offset the offset
+	 * @param len the length */
+	public Pixmap (byte[] encodedData, int offset, int len) {
+		ready = false;
```
tommyettinger, line 164: Is this meant to be assigned `true` here like it is in the non-emulated Pixmap? I don't think `ready` is used at all, in either Pixmap impl, and should probably be removed unless the JS specifically needs it.

gdx/src/com/badlogic/gdx/graphics/Pixmap.java
```diff
@@ -112,6 +112,7 @@ public static Pixmap createFromFrameBuffer (int x, int y, int w, int h) {
 	final Gdx2DPixmap pixmap;
 	int color = 0;
 
+	public boolean ready = false;
```
tommyettinger, line 115: If this is public and does something (anything), it should be documented, because I have no idea what this is supposed to do.

tommyettinger (commented):
I have some small concerns with the code, but nothing major. I agree with @MrStahlfelge that if anything can be shared with the downloading method, or there's some way the new typed array code can be used from that downloading method, that would be great.

yuripourre:
@MrStahlfelge @tommyettinger thank you for the comments. Indeed the `Pixmap.downloadFromUrl()` has a lot of similarities I believe they could be unified.

I will be working on the comments later today.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java
yuripourre: I am setting it as false here so the dev can check if this is loaded before instantiating a new texture using this Pixmap, otherwise it fails.

When `pixmap.ready` you can do `new Texture(pixmap)`.

gdx/src/com/badlogic/gdx/graphics/Pixmap.java
yuripourre: I changed to private and created a method `isReady()`. It is only useful when Pixmap is loaded async (currently only GWT).

yuripourre:
@MrStahlfelge you are right, the previous method is the same as `Pixmap.downloadFromUrl()` it wasn't working before because I had to call `ctx.drawImage()` otherwise it doesn't work. Now they are calling the same logic.

tommyettinger (approved):
Looks very clean now. I like that there's code reuse between the downloader and `byte[]` constructor. Approved!

yuripourre:
Thank you so much @tommyettinger! Looking forward to have this merged.

tommyettinger:
I have to say I don't understand the sync vs. async loading part... In the code before 3ef4ca1, it didn't look like any behavior changed to make the existing constructors no longer sync -- what did I miss? Is this just the byte[] constructor affected?

yuripourre:
@tommyettinger the main concern @mgsx-dev has is that the Pixmap should be created sync. This is how it's done in all backends but for GWT is really hard to do this sync. So devs should expect the same behavior in all backends but the code will not work on GWT.

tommyettinger:
If it's possible to load a Pixmap synchronously on GWT, then we should probably do that, but if it isn't possible for some constructor, how is it better to just omit it entirely? Isn't having a missing constructor a behavior difference too, since code that constructs a Pixmap from a `byte[]` will work on most platforms but fail to compile on GWT? I know people have struggled with the undocumented behavior difference in `getPixels()` on GWT (where it returns a `Buffer` instead of a `ByteBuffer` and backs the `Buffer` with an `IntBuffer`), but I fail to see how this PR worsens the behavior of GWT if the code that used the `byte[]` constructor wouldn't even compile before. At least we have a chance to document the newly-present GWT constructor in core, saying that it is sync or async on all platforms but X.

mgsx-dev:
The problem with this new "loaded" state is that all code that use Pixmap should check if it's loaded or not before using it. Even width and height need the data to be loaded, so it somehow breaks overall Pixmap API imho.

Maybe a cleaner solution would be to have a static method with a listener. It would ensure that you never get a half constructed Pixmap. Something like `void load(byte[], int, int, Listener<Pixmap>)` would work for all backends with the same API / same behavior and won't break anything. What do you think?

MobiDevelop:
@mgsx-dev I was just thinking that something like that would probably make more sense.

badlogic:
I agree with @mgsx-dev this would be a major API break. We do not want async constructors. The suggested static method with a callback is the better option.

PR #6740, opened 2021-12-22
Fix typo in btMultiBodyConstraintSolver.cpp

eltociear:
acutally -> actually


MobiDevelop:
While I am normally a big fan of correct spelling, this change isn't meaningful to us and introduces an unnecessary deviation from the actual Bullet sources so I'd prefer to close this.

PR #6741, opened 2021-12-22
Add new Pixmap constructor for loading from a ByteBuffer

PokeMMO:
Removes the need to create a copy if you already have an image in a **direct** ByteBuffer.


badlogic:
Neat, merged.

Issue #6634, opened 2021-08-19
trouble with AndroidGraphics.java destroy method


GolubevAlexey:
Hi everyone! I try close Activity after GDX's initialisation
but before add View to Layout.

And i have catched some proble with destroy method.

```
void destroy () {
		synchronized (synch) {
			running = false;
			destroy = true;

			while (destroy) {
				try {
					synch.wait();
				} catch (InterruptedException ex) {
					Gdx.app.log(LOG_TAG, "waiting for destroy synchronization failed!");
				}
			}
		}
	}
```

synch.wait();  - it waiting that long as can because  onDrawFrame don't call never

AndroidAplication.java
`if (isFinishing()) {
			graphics.clearManagedCaches();
			graphics.destroy();
		}`

#### Please select the affected platforms
- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

obigu:
Closed in https://github.com/libgdx/libgdx/pull/7699

PR #6745, opened 2021-12-30
fix:fix the crash or anr the view is gone before destory

TanZhiL:


tommyettinger:
Please, explain why you think this is necessary.

obigu:
Seems related https://github.com/libgdx/libgdx/issues/6634

TanZhiL:
> Please, explain why you think this is necessary.

in a normal android app it not a game app, the view maybe gone in some cases before activity onDestory() involved,   because onDrawFrame() never calls cause  anr

> Please, explain why you think this is necessary.

obigu (approved):
The fix makes sense and it's consistent with what we do in `pause()` https://github.com/libgdx/libgdx/blob/314f548806f0aa249314861f1ac896474eb15f6d/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L361-L373.

I see this rare ANR on my games so not only it affects apps:
image

obigu:
I've had a look again at this and I'm not sure this is needed. Haven't received any report of this ANR for at least the last 90 days so I'm guessing the reports that I saw in 2022 might have been caused by something not libGDX related.

fershous:
@obigu  I think this fix should be considered. I'm currently working on Jetpack Compose with LibGDX as a FragmentApplication, but whenever I try to destroy the fragment, the ANR causes the app to crash because of the lifetime of the Fragment, it blocks and doesn't allow the LibGDX instance to completely close.

tanaya177:
> I've had a look again at this and I'm not sure this is needed. Haven't received any report of this ANR for at least the last 90 days so I'm guessing the reports that I saw in 2022 might have been caused by something not libGDX related.

@obigu Even i had this issue in my application, which was fixed by this patch so, I think this patch should be considered

PR #6752, opened 2022-01-02
Upgrade Gdx-Setup generated projects Gradle version to 7.3.3

obigu:
Upgraded Gradle version of Gdx Setup generated projects to 7.3.3.

Gradle wrapper related files have been generated using Gradle `wrapper` task on a Mac.

- Java version requirements to execute Gradle: From Java 8 to Java 17 included. Yeah, we drop Java 7 but add support for Java 16 and Java 17, welcome to 2022! (https://docs.gradle.org/current/userguide/compatibility.html)
- Makes it easy for users to upgrade latest Android Gradle Plugin (AGP) 7.0.2 if they want to. I haven't updated AGP (currently on 4.1.3) to keep libGDX `gdx-tests-android` module on the same versions. We'll have to wait for libGDX to upgrade to Gradle 7 if we want to keep them in sync.
- https://gradle.org/whats-new/gradle-7/ for the rest of the benefits.

Tested running on Desktop (Lwjgl3), Android, iOS (IDEA plugin) and HTML.


gradle/dependencies.gradle
```diff
@@ -23,7 +23,7 @@ ext {
 
 versions.robovm = "2.3.14"
 versions.gwt = "2.8.2"
-versions.gwtPlugin = "1.0.9"
+versions.gwtPlugin = "1.1.16"
 versions.gretty = "3.0.6"
```
tommyettinger, line 27: Gretty is on version 3.0.7 now, at least. 3.0.6 does still work, but I think 3.0.7 will be the last update to the Gretty 3.x line for a while. This is important because Gretty 4.0.0 (at least from what I recall) needs Java 11 to run, so we can't go up to the 4.x line.

tommyettinger (changes requested):
Looks great, except the Gretty version is a tiny bit back from current -- if you know of any issues with 3.0.7, I'd really want to know about them too, but I think 3.0.7 has been working in Liftoff projects for a while.

When I did this, Gretty 3.0.6 wasn't out yet, so I had to do some sorta hacky work to get everything going. Now it seems so smooth...

gradle/dependencies.gradle
obigu: Not aware of any issues with 3.0.7, the only reason I've kept 3.0.6 is it was already being used by `gdx-tests-gwt`and I don't do any html so I wanted to touch as little as possible just to make it run. Should be fixed now.

tommyettinger (approved):

yuripourre:
Thank you @obigu!

PR #6753, opened 2022-01-06
Remove dead code in OpenALAudioDevice

Hangman:
This piece of code is dead as it never gets executed. The purpose was to fill the remaining buffers with silence and queue them on the source (the algorithm relies on it) but the code above already takes care of this, so these lines can be safely removed.


mgsx-dev (approved):

badlogic:
Looks good, cheers.

PR #6757, opened 2022-01-09
LWJGL3: Add errorStream option to initializeGlfw

PokeMMO:
Making this method public will allow a user to call it manually to override the PrintStream used for the error callback.

Due to how we call this early in [Lwjgl3ApplicationConfiguration#L263](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java#L263) we can't really make it a normal `Lwjgl3ApplicationConfiguration` option.

I'm open to feedback or suggestions on how to handle this better.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -75,10 +75,17 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	private static Callback glDebugCallback;
 	private final Sync sync;
 
-	static void initializeGlfw () {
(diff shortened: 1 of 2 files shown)


badlogic:
We can make it a static config option. The display mode methods in `Lwjgl3ApplicationConfiguration` can then call the `initializeGlfw(errStream)` method, passing whatever is currently set in the static config option. I.e.

```
Lwjgl3ApplicationConfiguration.errStream = ... whatever ...
Lwjgl3ApplicationConfiguration.getPrimaryMonitor(); // Which calls Lwjgl3Application.initializeGlfw(errStream);
```

PokeMMO:
Yeah I like that solution a lot better, I'll update the PR shortly.

PokeMMO:
Updated and much simpler change.

badlogic:
Perfection.

PR #6758, opened 2022-01-10
Update Tooltip.java

redy5:
Added option to let user set touch independency for tooltips, which prevents tooltip from showing when targetActor's hitbox is entered with screen already touched


badlogic:
Looking good, cheers.

Issue #6542, opened 2021-05-26
macOS/LWJGL3: Gdx net uses deprecated APIs


kyu-n:
#### Issue details
The [Lwjgl3Net](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java) class uses [deprecated, macos-specific APIs](https://openjdk.java.net/jeps/272) which are scheduled for removal from the JDK. The `com.apple.*` packages were deprecated in Java 9, and as-of [Java 16](https://openjdk.java.net/jeps/396), the JDK will throw an `IllegalAccessError` when calling (if using default settings).

These APIs are scheduled to become completely unusable in [JDK17](https://openjdk.java.net/jeps/403).

While OpenJDK offers the AWT-based Desktop API to replace this functionality, it's not usable for the lwjgl3 backend on macos. Calling `Desktop.getDesktop()` is enough to initialize the AWT event thread and causes any lwjgl3-based application to instantly hang, The LWJGL project often suggests using the [awt headless property](https://github.com/LWJGL/lwjgl3/issues/306#issuecomment-300481571) to workaround this issue, but this is nonsensical for the Desktop API's usage.

#### Version of LibGDX and/or relevant dependencies
1.9.10+

#### Stacktrace
```
java.lang.IllegalAccessError: class com.badlogic.gdx.backends.lwjgl3.Lwjgl3Net (in unnamed module @0x1f3f4916) cannot access class com.apple.eio.FileManager (in module java.desktop) because module java.desktop does not export com.apple.eio to unnamed module @0x1f3f4916
```

#### Please select the affected platforms
- [x] MacOS

crykn:
[This thread](http://forum.lwjgl.org/index.php?topic=6872.0) on the LWJGL forum mentions some possible solutions.

tommyettinger:
And the StackOverflow answer linked from that looks quite practical for libGDX, though I'm not sure what exactly Runtime is in this situation. https://stackoverflow.com/a/28807079/786740 (The comments seem vital, `xdg-open` on Linux and `start` on Windows should be good options, especially since I don't like listing all known browsers or using a particular name of a particular DLL that might change in some Windows 10 update.)

kyu-n:
> And the StackOverflow answer linked from that looks quite practical for libGDX, though I'm not sure what exactly Runtime is in this situation.

[...]

PR #6759, opened 2022-01-12
Update Lwjgl3Net.openURI

anyicomplex:
Avoid usage of java.awt.Desktop on LWJGL3.


tommyettinger:
I'm not a huge fan of PRs delivered with no context or rationale, but I vaguely recall a discussion about this in an issue. I'll see if I can locate the issue and link it here. The PR looks good at the surface level, but I remember all sorts of unexpected tricky aspects are associated with loading or opening URIs.

It really helps the people who have to evaluate each PR when there's any kind of explanation or reference that covers what the PR solves. Please explain (or just link to an existing explanation) what the PR does when you submit one, unless it is an extremely small and simple PR, like "Fix a typo in the README.md". If I remember it right, there is a valid issue in URI-opening code that this may indeed fully solve, but omitting any description makes me spend time explaining why descriptions are important, and doesn't save you any time because you still need to add one.

tommyettinger:
Yeah, there's a whole nest of issues associated. Most can be found from here, https://github.com/libgdx/libgdx/issues/6542 . I think this PR is meant to avoid usage of `Desktop` on LWJGL3, because right now part of NetAPITest and all of OpenBrowserExample are broken on LWJGL3 (and only there, to my knowledge).

kyu-n:
> Yeah, there's a whole nest of issues associated. Most can be found from here, #6542 . I think this PR is meant to avoid usage of `Desktop` on LWJGL3, because right now part of NetAPITest and all of OpenBrowserExample are broken on LWJGL3 (and only there, to my knowledge).

The security issues described in #6542's comment section aren't applicable here, since the PR does still wrap the untrusted text in a getURI() call. I don't have a well enough understanding of Windows' `cmd` caveats to give a good opinion on `start` though.

That said, there's no reason to avoid the JDK's `Desktop` API on Windows/Linux normally. The Linux `else if` is only present because of Snapcraft/Flatpak sandboxing (#6719), and if GLFW ever fixes their issues with macOS (lwjgl/lwjgl3#697), I'd remove that `open` command too. Popping a shell to handle a URI is fragile and has way more capabilities than this method should. It shouldn't be the default behavior.

If the goal of the PR is to resolve the broken tests, or handle a Windows LWJGL3 application which is inexplicably running in a headless mode, this can be better accomplished by testing for the Desktop API's availability and working around it if unavailable, like so:
```
		// macOS does not support the Desktop API in LWJGL3 applications. See: https://github.com/LWJGL/lwjgl3/issues/697
		// Some platforms may not provide the Desktop.Action.BROWSE function, e.g. Snapcraft/Flatpak on Linux. See: https://github.com/libgdx/libgdx/pull/6719
		if(!SharedLibraryLoader.isMac && Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.BROWSE))
		{
			try {
				Desktop.getDesktop().browse(new URI(uri));
				return true;
			} catch (Throwable t) {
				return false;
			}
		}
		else
		{
			String baseCommand;
			if (SharedLibraryLoader.isWindows) {
				baseCommand = "start";
			} else if (SharedLibraryLoader.isLinux) {
				baseCommand = "xdg-open";
			} else if (SharedLibraryLoader.isMac) {
				baseCommand = "open";
			} else
				return false;
			try {
				new ProcessBuilder(baseCommand, new URI(uri).toString()).start();
				return true;
			} catch (Throwable ignored) {
			}
			return false;
		}
```

anyicomplex:
On my machine (Fedora 35, Xfce), Desktop.getDesktop().browse(URI) is buggy, I'm sure that the application is not running in Snapcraft/Flatpak sandbox.

anyicomplex:
When I run Gdx.net.OpenURI(String), the link will not be open in browser, until the application closed.
When the application closed, links instantly open.
libGDX version: 1.10.0, Browser: Firefox

kyu-n:
XFCE has other issues with `xdg-open`. It defaults to `catfish` for file:// URLs in several distros, leading to [different broken behavior](https://archived.forum.manjaro.org/t/problem-open-folder-launches-catfish/128263/3).

What URL are you trying to browse on Fedora? http(s) or local file?

anyicomplex:
xdg-open works fine on my machine with http(s) URL.

anyicomplex:
> 

http URL.

anyicomplex:
`Desktop.getDesktop().browse(URI)` works buggy with both http(s) and local file URLs on my machine.

anyicomplex:
`xdg-open` works fine on my machine with both http(s) and local file URL.
`Desktop.getDesktop().browse(URI)` works buggy with both http(s) and local file URLs.

anyicomplex:
`xdg-open` works fine on my machine with both http(s) and local file URLs.
`Desktop.getDesktop().browse(URI)` works buggy also with both http(s) and local file URLs.

anyicomplex:
Here's my test code: 
```java
static final String GITHUB = "https://github.com";
    static final String TMPDIR = System.getProperty("java.io.tmpdir");
    static final String LOCAL = "file://" + TMPDIR + "/test.html";
    static final String HTML = "<!DOCTYPE html>\n" +
            "<html lang=\"en\">\n" +
            "<head>\n" +
            "    <meta charset=\"UTF-8\">\n" +
            "    <title>Title</title>\n" +
            "</head>\n" +
            "<body>\n" +
            "\n" +
            "</body>\n" +
            "</html>";

    static final String XDG_CMD = "xdg-open";

    public static void main(String[] args) throws URISyntaxException, IOException, InterruptedException {
        File html = new File(TMPDIR, new File(LOCAL).getName());
        if (!html.exists()) {
            if (!html.createNewFile()) throw new RuntimeException("Cannot access tmp dir.");
            FileWriter writer = new FileWriter(html);
            writer.write(HTML);
            writer.flush();
            writer.close();
        }
        Desktop.getDesktop().browse(new URI(GITHUB));
        Desktop.getDesktop().browse(new URI(LOCAL));
        Runtime.getRuntime().exec(XDG_CMD + " " + GITHUB);
        Runtime.getRuntime().exec(XDG_CMD + " " + LOCAL);
        Thread.sleep(30 * 1000);
    }
```

crykn:
This seems to be even more complicated. `Lwjgl3Net#openURI` does seem to work on Windows, but only as long as it's not used in a multi window environment (which the Lwjgl3TestStarter uses). 

Tbh there is no easy way to implement browser opening behaviour for all operating systems. The author of LWJGL 3 [suggests](http://forum.lwjgl.org/index.php?topic=6872.msg36188#msg36188) copying the old LWJGL 2 code (which did offer this functionality) and to me that seems to be the most viable solution. The Windows implementation can be found [here](https://github.com/LWJGL/lwjgl/blob/2df01dd762e20ca0871edb75daf670ccacc89b60/src/java/org/lwjgl/WindowsSysImplementation.java#L109-L117), the Linux one [here](https://github.com/LWJGL/lwjgl/blob/2df01dd762e20ca0871edb75daf670ccacc89b60/src/java/org/lwjgl/LinuxSysImplementation.java#L70-L88) (this one just iterates through different browser commands) and our already existing Mac one seems to work fine as well.

kyu-n:
> Here's my test code:

The `Desktop` API works fine by default on a fresh Fedora 35 machine. Your problem is threading. You can not block the thread which the JDK is using to execute the new process. It causes firefox to crash (I guess because it can't inherit something from the parent process, because the parent process has been hung?). It works fine for chromium, but it requires the thread be unblocked to finish executing.

Wrap your openURI call in a new Thread() or a postRunnable like `Gdx.app.postRunnable(() -> Gdx.net.openURI("URI"));`.

Spawning new processes is a platform-specific problem and comes with lots of issues like this.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java
```diff
@@ -66,28 +65,32 @@ public Socket newClientSocket (Protocol protocol, String host, int port, SocketH
 
 	@Override
 	public boolean openURI (String uri) {
-		if (SharedLibraryLoader.isMac) {
+		if (SharedLibraryLoader.isWindows) {
 			try {
-				(new ProcessBuilder("open", (new URI(uri).toString()))).start();
+				new ProcessBuilder("rundll32", "url.dll,FileProtocolHandler", new URI(uri).toString()).start();
```
badlogic, line 70: This is a sure-fire way to trigger stupid anti-virus software.

anyicomplex:
> Wrap your openURI call in a new Thread() or a postRunnable like `Gdx.app.postRunnable(() -> Gdx.net.openURI("URI"));`.

Not work for me.

anyicomplex:
> Wrap your openURI call in a new Thread() or a postRunnable like `Gdx.app.postRunnable(() -> Gdx.net.openURI("URI"));`

Not work on my machine.

PR #6761, opened 2022-01-14
deleting the ant build.xml in the extensions/gdx-setup/

cupup77:
hence gdx using gradle , deleting the old script make file clean


PokeMMO:
Thanks!

PR #6762, opened 2022-01-15
Add missing isEmpty method to emulated Timer class in GWT backend

6money:
This method was missing from the emulated Timer class in the GWT backend, so the html build failed if timer.isEmpty() was used in the project. Tested using the GWT superDev task, using a local build of libgdx on a fresh project generated from gdx-liftoff 1.10.0.10

This may be considered a change/fix of the public api, so if it should be included in CHANGES let me know.


tommyettinger (approved):
I wouldn't consider this a change of the public API, because what the API was declared as supporting (the methods in Timer in the core project) wasn't reproduced in GWT correctly before and is now. Since this was just, I presume, a mistake on the GWT support for that API, and this is a quick and simple fix, I'm all for this being merged. The only thing I'm unsure about is the behavior of `synchronized (this)`, especially on GWT, so if anyone is aware of any potential quirks of synchronized code on single-threaded JS platforms, please let me know! I think @MrStahlfelge might be the best person to ask about this type of GWT code, and might have some inkling as to why isEmpty() could have been omitted originally (if it wasn't a mistake).

6money:
Thanks for the review. I agree the omission of this method was likely just a mistake. The git blame for the core Timer class shows isEmpty was added a few years after the class was originally created, so it may be it was just forgotten to be added to the GWT backend at the time.

As for the synchronized statement, I'm also unsure how that behavior translates in JS. I wonder if it is even required due to JS being single threaded? I am currently working under the assumption it is safe to be there, based on the fact other methods in the emulated Timer class also contain synchronized statements.

tommyettinger:
Yeah, that's why I'm hoping someone is around who knows about this -- it's possible that maybe Timer in the GWT emu doesn't need the same synchronization code as other platforms, maybe just in isEmpty(), maybe in all of the class. Maybe synchronized blocks are ignored on GWT, and we can safely merge this as-is. Maybe they aren't ignored but the code in this PR is still perfectly correct, and we can safely merge this as-is. Only if the thread-related code isn't ignored, and it behaves differently on GWT in some way, do we really need to be careful. Even then, this is probably fine right now, it looks like a small check for whether a collection is empty, but the docs have me puzzled and so I'm waiting for someone who knows more about this area. The only possible worst-case is if there's a subtle difference in behavior or bug in GWT or this code, this gets merged without proper review, and a release happens immediately after, making Timer stuck with one method that, while not broken like it was before, is broken in some other way until the next release. You did test this, which is great, I just don't know if there's some particular rare circumstance that we need to be aware of (and possibly add to the tests).

MobiDevelop:
I can't imagine there is any risk in merging this as is. The synchronized piece is not necessary due to the single threaded nature of JS, so I am sure the GWT compiler just drops it.

badlogic:
No need for synchronized in singled threaded JS code. There aren't any synchronization primitives in JS that would be useful for shared state, even when web workers are involved.

PR #6734, opened 2021-12-19
Remove maven/ant build

PokeMMO:

PR #6766, opened 2022-01-15
Add CHANGES entry for maven build removal

PokeMMO:
See #6725 #6733 #6734

I don't really know how else to say this. There's no detailed migration instructions as anything more specific is implementation specific.

If anyone has any suggestions, they would be appreciated.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.10.1]
+- [BREAKING CHANGE] libGDX is now built with gradle. See https://libgdx.com/dev/from-source/ for updated build instructions.
 - [BREAKING CHANGE] Android Moved natives loading out of static init block, see #5795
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
Is it even a breaking change? For users of the library releases or nightlies, it is not. We'd have to hear from someone who actually experienced a break to know how it broke, because I don't see how that would have happened. I'm fine with it being marked as a breaking change, because, maybe, it breaks something somewhere, and I've known people who have seemed pretty upset about having to use Gradle. Pretty much anything can be marked as a breaking change, I guess.

Not like my approval matters on these things anyway, but I'm approving.

PokeMMO:
It's a breaking change for anyone building libgdx themselves. Maybe it should be tagged differently? Something like `[Building Change]` or `[Development Change]`?

PR #6770, opened 2022-01-17
Move Uberjar creation task out of Android tests.

tommyettinger:
This needs some review, but it allows the project to actually be imported by IDEA. I'm not 100% sure this is correct, which is why I'm PRing it. Without this change, @Berstanio and I both encountered the same issue in IDEA 2021.3.1, a MissingPropertyException:

Could not get unknown property 'main' for SourceSet container of type org.gradle.api.internal.tasks.DefaultSourceSetContainer.

In my case there was a line number, which pointed to a line containing `from sourceSets.main.output`, which apparently doesn't work properly with the Android support in IDEA. I moved the Uberjar creation task so it runs for all subprojects except the Android tests, since those should probably produce an APK or app bundle. I've never run the Android tests, so I'm not sure what to expect, but I'll start those up now, and get the PR in so people can actually edit files in their IDE of choice.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,21 +90,7 @@ allprojects {
         tasks.withType(Javadoc) {
             options.addStringOption('Xdoclint:none,-missing', '-quiet')
         }
-    }   
-
-    tasks.register('uberJar', Jar) {
-        archiveClassifier = ''
-        archiveVersion = ''
-        duplicatesStrategy = 'warn'
-
-
-        from sourceSets.main.output
-
-        dependsOn configurations.runtimeClasspath
-        from {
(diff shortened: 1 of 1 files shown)


badlogic:
Sorry, I broke that. I'll see how I can fix it properly. It should only be a task on gdx, the 2 LWJGL backends, freetype, and gdx-tools. Unsure how to add it to just those task without repeating myself. In the latest master state, the task is in `gradle/dist.gradle`, but still applies to all projects.

tommyettinger:
OK, if there's a better fix I'm all for it. I don't see a problem with a configure block that just lists those specific projects to run uberjar on.

tommyettinger:
OK, something changed since I wrote this, probably what you just mentioned about dist/build,gradle . Feel free to close this if it's unusable; it seems like it needs more changes to get working.

badlogic:
Was totally useable, just had to do it in dist.gradle. Fixed.

PR #6773, opened 2022-01-18
Fix eclipse compilation and running of tests.

PokeMMO:
As discussed in #6369 this PR fixes eclipse test compilation and running of tests.

This simply removes the test as it is the easiest way to fix the dependency chain. Other possibilities include:
1. Moving the test to `gdx-tests-lwjgl` and add a dependency on `gdx-tools` to that project.
2. Create new `gdx-tests-tools` for tool related tests.


NathanSweet:
Confirmed it works well now, thanks!

PR #6776, opened 2022-01-21
Add lwjgl3 extensions to publish.gradle

PokeMMO:
Was missed when adding the new projects. Normally we would have this match all subprojects, but libgdx's weird structure means specifying is better.


badlogic:
I'm a dum dum. Cheers!

PR #6777, opened 2022-01-21
Pixmap#setPixels will now verify it has been given a direct ByteBuffer

### LINKED ISSUE #6775: Pixmap doesn't specify it requires a Direct Byte Buffer
The docs for Pixmap should be updated to reflect that they only work with a DirectByteBuffer when you call setPixels(). More practically the buffer should be checked with isDirect() and throw an error rather than crashing the JVM.

PokeMMO:
Fixes #6775

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Pixmap.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Pixmap.java
@@ -448,9 +448,10 @@ public class Pixmap implements Disposable {
 		return pixmap.getPixels();
 	}
 
-	/** Sets pixels from a provided byte buffer.
-	 * @param pixels Pixels to copy from, should match Pixmap data size (see {@link #getPixels()}). */
+	/** Sets pixels from a provided direct byte buffer.
+	 * @param pixels Pixels to copy from, should be a direct ByteBuffer and match Pixmap data size (see {@link #getPixels()}). */
 	public void setPixels (ByteBuffer pixels) {
+		if (!pixels.isDirect()) throw new GdxRuntimeException("Couldn't setPixels from non-direct ByteBuffer");
 		ByteBuffer dst = pixmap.getPixels();
 		BufferUtils.copy(pixels, dst, dst.limit());
 	}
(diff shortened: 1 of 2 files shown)


tommyettinger (changes requested):
I think the GWT emu Pixmap should also have this change considered: https://github.com/libgdx/libgdx/blob/7a6a02f114c2b14a2bcacc9225c9610d4bf62f2b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java#L265-L270 ...Though the GWT Pixmap's Buffer is quite different from the one on desktop, and perhaps the fact that it uses an IntBuffer currently as its backing type should be mentioned,

tommyettinger:
> I think the GWT emu Pixmap should also have this change considered:
> 
> https://github.com/libgdx/libgdx/blob/7a6a02f114c2b14a2bcacc9225c9610d4bf62f2b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/Pixmap.java#L265-L270
> 
> ...Though the GWT Pixmap's Buffer is quite different from the one on desktop, and perhaps the fact that it uses an IntBuffer currently as its backing type should be mentioned,

This looks like it might be different from what I remembered. I guess this can take a real ByteBuffer now? Excellent. It would still be good to have this throw the same exception that desktop does if it is relevant.

PokeMMO:
I have no knowledge of if there's direct/non-direct difference in gwt.

Maybe @MrStahlfelge does?

MrStahlfelge:
I'm not into the topic so I don't know if it makes sense, but the [GWT emu file](https://github.com/libgdx/libgdx/blob/7a6a02f114c2b14a2bcacc9225c9610d4bf62f2b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/java/nio/ByteBuffer.java) provides the `isDirect()` method so it is safe to call it.

Tom-Ski (approved):

Tom-Ski:
Whats the verdict on this @tommyettinger, what changes do you want made?

tommyettinger (approved):

tommyettinger:
If GWT isn't affected anyway, I'm fine with these changes.

PR #6778, opened 2022-01-21
Remove inclusion of static libs into final app bunlde for iosrobovm test

Berstanio:
Prior to this the static libs would be included inside the final app bundle which is not a intended behavior.


obigu:
This was discussed on Discord with @Tom-Ski and there doesn't seem to be any reason why the static libs are added to Gradle (they are already linked by path on `robovm.xml`) so it's safe to merge. Thanks!

PR #6783, opened 2022-01-27
Handle Android's MotionEvent.ACTION_CANCEL touch event as TouchEvent.TOUCH_UP

maksimkalatsilin:
Resolves an issue introduced in https://github.com/libgdx/libgdx/pull/5222

We have to map MotionEvent.ACTION_CANCEL to TouchEvent.TOUCH_UP because there's no corresponding event in LibGdx's TouchEvent.

https://developer.android.com/reference/android/view/MotionEvent#ACTION_CANCEL
Android docs for MotionEvent.ACTION_CANCEL:
The current gesture has been aborted. You will not receive any more points in it. You should treat this as an up event, but not perform any action that you normally would.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidTouchHandler.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidTouchHandler.java
@@ -59,6 +59,7 @@ public class AndroidTouchHandler {
 				break;
 
 			case MotionEvent.ACTION_UP:
+			case MotionEvent.ACTION_CANCEL:
 			case MotionEvent.ACTION_POINTER_UP:
 			case MotionEvent.ACTION_OUTSIDE:
 				realPointerIndex = input.lookUpPointerIndex(pointerId);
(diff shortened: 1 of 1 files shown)


MrStahlfelge:
Does this fix #4866 ?

maksimkalatsilin:
@MrStahlfelge, does pausing the app produce `MotionEvent.ACTION_CANCEL` event in Android's `onTouch(...)`? If so, most likely it's the same issue.

As I said in description, there's no corresponding touch event in LibGdx API, so we have to use TOUCH_UP. Yes, it's a potential false-positive click. But in my app I handle this at game logic layer: simply ignore irrelevant gesture.

obigu:
This is only part of the proper solution suggested by Nate discussed in #6353 but lacks the handling of the fake cancel event. It is important if we perform the `touchUp` on `touchCancel` not to execute the `touchUp` logic on components that depend on it by checking if that `touchUp` is really a `touchUp` or a `touchCancel`, otherwise we would see unwanted false positives and break backwards compatibility. In addition to that, the fix would need to clean up the hacks we do resetting touch states on `onPause()` as those would probably not be needed anymore.

MrStahlfelge:
Closed in favor of #6871 adding touch cancel event. - thanks for the contribution

PR #6786, opened 2022-01-30
Some gwt improvements

SimonIT:
I discovered some invalid @throw tags in our gwt sources. By checking the sources I stumbled upon deprecated method usages and replaced them. I cleaned up some other warnings as well

There shouldn't be any functional differences


MrStahlfelge (approved):
LGTM

PR #6788, opened 2022-02-01
Remove copyrighted assets from tests

Frosty-J:
*Cloud Connected* is copyright Believe Music / Nucelar Blast. Despite us only using a small portion of it, it doesn't seem worth the legal risk of keeping. I've replaced it with an OGG encode of `8.12.mp3` (`ffmpeg -i 8.12.mp3 8.12.ogg`). I don't know what 8.12.08 by torque is, but it didn't hit a YouTube Content ID match so I guess it's safe.

The *Yoshi's Island DS* concept art egg is copyright Nintendo. I replaced it with a similar version based on https://pixabay.com/vectors/easter-egg-easter-egg-1324879/. I'm aware of the thin gaps between shapes in the export.

The *Kirby: Canvas Curse* wheel is copyright Nintendo. By this point I stopped trying to find a similar replacement and just used `sys.png` instead, as it's the same resolution and `wheel.png` is only used for decal tests.

I was unable to find the source of `43kJgl.jpg` - only information about it. Whatever its copyright status, I think it's low risk, but seeing as `JpegTest` no longer exists I removed it. We have plenty more JPEGs if we need one.

If there's any other non-vital material likely in breach of copyright, let's fix it here.


badlogic:
oh god, those are still in here. good catch. 8.12 is by yours truely and i
hereby grant full rights.

PR #6791, opened 2022-02-07
GWT: Turns preloadedAssets method into protected

yuripourre:
This change allows GWT applications to skip the preload step or change the way it's done.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -163,7 +163,7 @@ public abstract class GwtApplication implements EntryPoint, Application {
 		preloadAssets();
 	}
 
-	void preloadAssets () {
+	protected void preloadAssets () {
 		final PreloaderCallback callback = getPreloaderCallback();
 		preloader = createPreloader();
 		preloader.preload("assets.txt", new PreloaderCallback() {


yuripourre:
My bad, I should have searched more, I could just override the [createPreloader](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L279) method.

yuripourre:
Overriding the [createPreloader](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java#L279) method is not enough. The original is:

```
void preloadAssets() {
	final PreloaderCallback callback = getPreloaderCallback();
	preloader = createPreloader();
	preloader.preload("assets.txt", new PreloaderCallback() {
		@Override
		public void error (String file) {
			callback.error(file);
		}

		@Override
		public void update (PreloaderState state) {
			callback.update(state);
			if (state.hasEnded()) {
				getRootPanel().clear();
				if(loadingListener != null)
					loadingListener.beforeSetup();
				setupLoop();
				addEventListeners();
				if(loadingListener != null)
					loadingListener.afterSetup();
			}
		}
	});
}
```
I tired to create a dummy preloader that also creates a dummy PreloaderState, the logic doesn't break but doesn't call the `setupLoop()`. Which didn't start the application.

MobiDevelop:
I think what's missing here is the why... What's the reason you want/need to override the preloading of assets in the basic sense?

yuripourre:
@MobiDevelop I need to load all the assets asynchronously and most of them not at the start of the application. I am using my own preloader. Also the assets are served by a CDN. They are not hosted by the same server.

This is the cleanest way I found to load them.

yuripourre:
Another reason is the assets are dynamic and managed in another service so there is no way to use a hardcoded list.

MobiDevelop:
I see, I assume you must have some mechanism already to load those assets outside of the preloader mechanism.

Why not just configure the asset filter to produce an empty assets.txt? Then you'd theoretically need to make no change to libgdx itself.

yuripourre:
@MobiDevelop Yes, you are right, I have an alternative mechanism. Having the AssetFilter generating an empty file would be perfect but I have no idea how to do it.

MrStahlfelge:
You should look into the asset filter, you can use it to turn on lazy loading (= no preloading) as well.
Check out my smc-libgdx repo.

yuripourre:
@MrStahlfelge awesome! I will take a look.

https://github.com/MrStahlfelge/SMC-libgdx/blob/master/html/src/rs/pedjaapps/smc/GdxDefinition.gwt.xml#L12

yuripourre:
@MrStahlfelge I created a lazyload filter, it doesn't break the app although it requires an empty assets.txt file. I want to skip it entirely. Any suggestions?

```
GET http://0.0.0.0:8000/assets/assets.txt?etag=1645053976747 [404] File not found]
```

MobiDevelop:
May I ask why it's so important not to even have an empty file? It would have effectively zero impact to the loading time and as a bonus you _could_ preload something in the future easily if you wanted.

yuripourre:
@MobiDevelop As I mentioned, all the files are served by a CDN in an external url. That means that the libgdx app is expecting the assets.txt file to be in the same url as the javascript file, which isn't. So it gives me an error in console. I tried hard to avoid this problem but because the code on GwtApplication class is so coupled, when I try to create my own GwtApplication it has to be in the same package as the previous one otherwise graphics' attributes cannot be accessed.

I couldn't find another way to avoid preloading, maybe I am missing something. I am really open to suggestions.

MobiDevelop:
Sure but you can't host the empty assets.txt alongside the rest of the application? It just seems like your trying to solve a problem in a way that makes life harder rather than easier.

yuripourre:
@MobiDevelop the file would have to be served by a [Vue Router](https://router.vuejs.org/) and is not managed by my team. Trust me, having this method protected will make my life much easier.

MobiDevelop:
I see, well, I guess I just don't really get the problem then. Ultimately, if the easiest thing is to make it protected, there are worse things in the world.

For what it's worth, this seems to work just fine for me if you really just want to bypass all preloading and go straight to the setup sequence:

```java
@Override
public Preloader createPreloader() {
	return new Preloader(getPreloaderBaseURL()) {
		PreloaderState fakeState = new PreloaderState(new Array<Asset>()) {
			@Override
			public boolean hasEnded() {
				return true;
			}
		};
		@Override
		public void preload(String assetFileUrl, PreloaderCallback callback) {
			callback.update(fakeState);
		}
	};
}
```

You could also change the url returned by getPreloaderBaseURL() if you went with the filter method described above and point the base url at wherever the CDN hosts the files. Without something more concrete, these options are the best I can offer.

yuripourre:
@MobiDevelop I tried that but because there is no assets.txt file, when I call `preloader.preload("assets.txt", new PreloaderCallback() {...});`.

It will call the error method of the preloader therefore there is no call to `setupLoop()` and `addEventListeners()`

MobiDevelop:
I ran that exact code before posting it, so no clue what is going wrong in your case. 🤷

yuripourre:
@MobiDevelop thank you for the suggestions. Maybe when I tried I was calling the wrong URL, I will try some more tomorrow and post an update here.

yuripourre:
@MobiDevelop you are right, it does work, I just tested it! I was probably testing something else with this change before. I am closing it.

PR #6792, opened 2022-02-07
Improve WAV support with 8-bit PCM and more!

Frosty-J:
This was a fairly simple pull request to add support for 8-bit PCM WAVs on desktop. And then disaster struck! Let's go through each aspect of it in the worst possible manner for being taken seriously, shall we?

## 8-bit PCM

When dynamic range isn't a concern, 8-bit PCM sounds identical to its 16-bit counterpart, whilst being half the size. It can be loaded extremely quickly - faster than 16-bit due to being smaller, and faster than compressed formats because it can be played without having to decode anything.

It's fairly trivial to implement into libGDX's desktop backends and is already compatible with other platforms (see table near end of post), so I really see no reason not to add support for it.

But can we do better? Perhaps. IMA ADPCM is 4 bits per sample and doesn't suffer the limited dynamic range of 8-bit PCM. I was trying to implement it but it turned out to be a bit annoying to do so - wouldn't be worthwhile if the pull request would just get rejected. It has poorer compatibility - again, see table.

## Beyond 16-bit PCM

Anyone with too much time on their hands will surely be frantically yelling at their screen now about 24-bit, 32-bit and even 64-bit WAVs! Turns out it was also easy to support 32-bit and 64-bit floating-point PCM.

But let me make it clear that there's little reason to use this. If your volume is loud enough to hear digital noise in 16-bit audio, you won't have your hearing for very long. And your sound card, cables, speakers, all that, would need to be fandabidozi.

The only practical use I see for this is to be able to play your game using the master copies of your sound assets, if they're 32/64-bit, instead of having to convert them during development. On many computers, 32/64-bit floating-point sounds will be played back at less than this, often 16 or 24-bit.

Note that 24-bit integer PCM, commonly used by audiophiles (in FLAC format, but libGDX on desktop doesn't do FLAC), is unsupported. There's no OpenAL binding for 24-bit audio, and I've decided against converting it to a different bit depth. I'm not against the idea of converting it - it just doesn't seem worth anyone's while. Similarly, 32-bit integer PCM is also unsupported - floating-point only!

## MP3 WAVs

To expand our WAV support a little, I decided to use a decoder we've already got - MP3! In order to handle this, I had to do some jiggery pokery to jump between classes. Surely there's a better way - something for the Java experts to look at? Anyway, it seems to work fine with no extra work needed to traverse the RIFF header stuff.

I also looked at Vorbis, seeing as there are TwoCCs reserved for it, but didn't really get anywhere. ffmpeg can shove Vorbis in a WAV, but the result doesn't seem valid in any way, shape or form - creates garbage sounds in all software, even itself. And it uses codec 0x566f, which no-one's ever heard of. But Vorbis WAVs are obviously very rare, so I'm not going to worry about it.

Building upon this, I'd like to see libGDX adhere to file extensions less strictly when it comes to audio. Files have magic numbers at the start - use them! Not relying on file extensions could make it easier to do cross-platform development. For example, instead of loading `bang.ogg` on one platform and `bang.m4a` on another, I could simply load `bang.snd`, building each version of the game with different generic-file-extension assets.

## Surround sound

After seeing how libGDX requires WAVs to have only 1 or 2 channels, I decided to see what the situation is with other formats. The MP3 specification doesn't permit more channels, but in Ogg land it was an option. I created a 4-channel file, included in this pull request, to see what would happen. It played at half speed!

Turns out it just needed to be played with `AL_FORMAT_QUAD16` instead of `AL_FORMAT_STEREO16`. Well, that's the story on LWJGL3. On LWJGL2, I increase the sample rate to make it correct. Not an ideal solution, and I'm suspicious about the Ogg decoder's output when we consider the fact that even worked. Please test both `quadraphonic.ogg` and `quadraphonic.wav` under LWJGL3 if you have a surround sound setup.

So I guess we now have some level of surround sound in libGDX? I'm not actually sure. I don't have a surround sound setup. I might be able to fake one, though it just wouldn't be the same for properly testing things. Through my trusty old headphones, `quadraphonic.ogg` now plays at the correct speed. The front two channels get played and the rear two discarded.

If anyone with (or without, I guess) a surround sound setup wants commercial audio to be able to better test this, and can't obtain it themselves, I can provide quadraphonic ambience, 5.1 music and 7.1 movie audio. Might have to be in private - not sure how appropriate it would be to do here. I don't believe I have any 6.1 audio but can scour my DVD/BD collection for some if necessary.

Note that WAVs with more than two channels will not work when created with ffmpeg due to `Wav.java` disallowing Extensible WAVs. I've chosen to leave it this way, as allowing Extensible WAVs is a kettle of fish which could allow for ear-bleeding sounds if someone has a strange WAV. To create a WAV with more than two channels, Audacity can be used instead. You could probably just hex-edit ffmpeg WAVs to appear as PCM, or maybe use some flag - I haven't looked into that.

## LWJGL2

It's at this point I have to address the elephant in the room, which is that this pull request favours LWJGL3 over LWJGL2. If you use obsolete software... surprise, surprise, you get fewer features!

Whatever OpenAL stuff is going on in LWJGL2 doesn't seem to support many of the formats this pull request makes use of, so it simply isn't practical to fully support. But it still manages to support 8-bit audio just fine, which is ultimately what I came here for.

I don't know what the official stance is, but I see little reason to give the LWJGL2 backend much love. By all means, it should remain an option for games that were developed with it and for whatever reason can't easily use LWJGL3 (such as due to AWT integration), but new games shouldn't be developed with it (unfortunately they still are due to the gdx-setup stable download).

## A table or two

Okay, words can be confusing, so here's the state of play on my system:

Key|&nbsp;
-:|-
✔️|All good!
❌|Error or silence.
⚠️|Loud noises that hurt my ears!
2|The first 2 channels get played on my stereo setup.
4|All channels get played on my stereo setup.<br>I don't know which behaviour is preferable.

Backend|PCM8|PCM16|PCM32f|PCM64f|4ch Ogg|4ch WAV
:-:|:-:|:-:|:-:|:-:|:-:|:-:
**LWJGL2**|✔️|✔️|❌|❌|2|❌
**LWJGL3**|✔️|✔️|✔️|✔️|2|2
**Chrome**|✔️|✔️|✔️|❌|4|4
**Firefox**|✔️|✔️|✔️|⚠️|4|4
**iOS GWT**|✔️|✔️|✔️|✔️|❌|4

* I've been unable to test on Android, as entering any test crashes the app. According to Desu on Discord, some code is missing to do with the natives. I'll update the tables once that's fixed.
* It's impractical for me to test native iOS builds.

Here's another table for formats that I haven't included test files for, mostly to avoid clogging up the repository:

Backend|PCM24|A-law|µ-law|IMA ADPCM|MS ADPCM|MP3 WAV
:-:|:-:|:-:|:-:|:-:|:-:|:-:
**LWJGL2**|❌|❌|❌|❌|❌|✔️
**LWJGL3**|❌|❌|❌|❌|❌|✔️
**Chrome**|✔️|✔️|✔️|❌|❌|✔️
**Firefox**|✔️|✔️|✔️|❌|❌|✔️
**iOS GWT**|✔️|✔️|✔️|✔️|✔️|❌

Trivia: Java's crappy built-in sound support supports A/µ-law.

## Final notes

* When poking around in this many places, it's possible I may have messed something up, as much of the existing code doesn't explain why it does what it does. Existing projects should be tested, despite this not being intended as a breaking change.
* It looks like there's OpenAL bindings for ADPCM and all sorts in there. Heck, there's even `AL_FORMAT_VORBIS_EXT`. But they don't make any sound. If it's possible to replace our Java decoders with what's already in LWJGL3/OpenAL, I'd like (someone who knows what they're doing) to.
* If the formatter could not triple the length of my `switch` statements, that would be great.


tommyettinger:
There's clearly a lot of care put into this PR, judging by the excellent memes. I have headphones/earbuds that I am fairly certain support stereo sound, so I can try this on those in a little while. I kinda doubt just reviewing the source will find many bugs, so I'll pull this PR and try it out. I'm a little low on time lately but I'll try this when I can.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OggInputStream.java
```diff
@@ -454,10 +451,14 @@ public int read (byte[] b, int off, int len) {
 			if (value >= 0) {
 				b[i] = (byte)value;
 			} else {
-				if (i == 0) return -1;
-				return i;
+				if (i == 0) {
+					return -1;
+				} else {
+					return i;
+				}
```
NathanSweet, line 458: This didn't really need to change, though it's immaterial.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
```diff
@@ -61,7 +61,12 @@ public Sound (OpenALLwjglAudio audio, FileHandle file) {
 			WavInputStream input = null;
 			try {
 				input = new WavInputStream(file);
-				setup(StreamUtils.copyStreamToByteArray(input, input.dataRemaining), input.channels, input.sampleRate);
+				if (input.type == 0x0055) {
+					setType("mp3");
```
NathanSweet, line 65: This is supporting MP3 inside WAV? It doesn't seem necessary to support. Why would someone do this?

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
```diff
@@ -169,5 +157,36 @@ public int read (byte[] buffer) throws IOException {
...
+			case 0x0070:
+				return "Lernout & Hauspie CELP 4.8kbps";
+			case 0x0072:
+				return "Lernout & Hauspie CBS 12kbps";
+			case 0xfffe:
+				return "Extensible";
+			default:
+				return "Unknown";
+			}
+		}
```
NathanSweet, line 190: If you dislike what the formatter does to switches, you can use `// @off` and `// @on` (or at least you used to be able to).

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
```diff
@@ -90,40 +95,23 @@ public WavInputStream (FileHandle file) {
...
-					}
-					throw new GdxRuntimeException("WAV files must be PCM, unsupported format: " + name + " (" + type + ")");
-				}
+				type = read() & 0xff | (read() & 0xff) << 8;
 
-				channels = read() & 0xff | (read() & 0xff) << 8;
-				if (channels != 1 && channels != 2) throw new GdxRuntimeException("WAV files must have 1 or 2 channels: " + channels);
+				if (type == 0x0055)
+					return; // Handle MP3 in constructor instead
+				else if (type != 0x0001 && type != 0x0003) throw new GdxRuntimeException(
```
NathanSweet, line 102: Dangling `else`. 😛

NathanSweet:
I did a quick read through of the diff, looks good!

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/OggInputStream.java
Frosty-J: This is from copying LWJGL3's changes into LWJGL2, as are the other immaterial changes in that class. I'm not too bothered either way - just feels like life would be easier if there's not too much divergence between them.

https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OggInputStream.java#L463

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
Frosty-J: This practice was more common in older software and possibly hardware than it is today. Perfectly valid - just a bit silly when you can use the proper container and extension instead.

If we'd rather not support MP3 WAVs, more code than what you've highlighted here can be reverted (you probably already know this - just a note for everyone).

Here is one reason why one may go out of their way to put MP3 in a WAV, albeit in related to libGDX: http://www.studiodust.com/riffmp3.html#what

Sound Recorder for Windows 95 to XP used the WAV container for MP3, as did anything else that relied on ACM for encoding audio. It's a truly pants program. 
https://microsoft.fandom.com/wiki/Sound_Recorder#Features

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
Frosty-J: Probably resolved in 879acc1 if I understand what you mean. I'm still not terribly happy with how it's being handled - I think it could be easy to miss the `return` and understand the flow of the code. But all TEt told me was "nice work", so there goes all hope.

Hangman:
The non-matching channels in the 4-channel-files may be dropped on lwjgl3 due to AL_DIRECT_CHANNELS_SOFT being true. So this could possibly be changed if desired, though it was set to true for Sound recently (Music always had it) as a fix iirc.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/audio/Wav.java
tommyettinger: Yeah, this looks resolved. "Dangling else," where an `else` or `else if` is the next block that would be tried after a certain `return`, `continue`, or `break` probably has minimal to no effect on the performance of the code, but I'd love to be proven wrong. I don't think it affects readability much either... I actually feel like it hurts readability a little to remove the `else`, since there's a clear chain from `if` to `else if` to `else` without needing to look inside any of those blocks, and removing the dangling else forces you to inspect for `return` statements inside the blocks.

PR #6793, opened 2022-02-08
Copy required natives to gdx-tests-android

PokeMMO:
Not sure how this ever worked without this, but it doesn't work without this now.

### DIFF
--- a/tests/gdx-tests-android/build.gradle
+++ b/tests/gdx-tests-android/build.gradle
@@ -54,3 +54,31 @@ android {
 		versionName "1.0"
 	}
 }
+
+// called every time gradle gets executed, takes the native dependencies of
+// the natives configuration, and extracts them to the proper libs/ folders
+// so they get packed with the APK.
+task copyAndroidNatives() {
+	doFirst {
+		file("libs/armeabi-v7a/").mkdirs();
+		file("libs/arm64-v8a/").mkdirs();
+		file("libs/x86_64/").mkdirs();
+		file("libs/x86/").mkdirs();
+
+		['arm64-v8a', 'armeabi-v7a', 'x86_64', 'x86'].each { arch -> 
+			[project(":gdx"), project(":extensions:gdx-bullet"), project(":extensions:gdx-box2d-parent:gdx-box2d"), project(":extensions:gdx-freetype")].each { p ->
+				copy {
+					from new File(p.projectDir, "libs/${arch}")
+					into "libs/${arch}"
+					include "*.so"
+				}
+			}
+		}
+	}
+}
+
+tasks.whenTaskAdded { packageTask ->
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
This looks like the standard `copyAndroidNatives()` task in a setup-created project; I also wonder why this could be missing from the Android tests. If tests aren't even runnable right now for Android, and this doesn't break the entire build, it can't hurt to at least be able to see what tests are currently working. Definitely approved.

PokeMMO:
It's based on the standard version, but customized for in repository structure. The project doesn't have full normal native jars available.

Frosty-J:
Thanks. Tests are now working on my Android device. Previously, every last one would crash due to a lack of native libraries.

I had a quick run through them, and most appear to be working. Here are some of concern:

Test|Result
-|-
APKExpansion|`Error reading file: data/obbasset-DroidSerif-Regular.ttf (Internal)`
AudioRecorder|`Do you have the RECORD_AUDIO permission?`
Clipboard|Flashing static - unsafe for photosensitive epilepsy
InstancedRendering|`GLES 3.0 profile required for this test`
PixelBufferObject|`Attempt to invoke interface method 'java.nio.Buffer com.badlogic.gdx.graphics.GL30.glMapBufferRange(int, int, int, int)' on a null object reference`
ReflectionCorectness|`Expected that value of is Annotation is false but is true`
SystemCursor|Flickering static
TextureArray|`TextureArray requires a device running with GLES 3.0 compatibilty`

* The missing font can probably be fixed by using `DroidSerif-Regular.ttf` instead.
* Tests that have a horrible static background can be fixed by clearing the screen. This will also prevent the previous frame from being visible when resizing on PC.
* I don't know why some tests fail due to a lack of GLES 3.0, as `GLES30Test` works and my device has a GL Version of ES 3.0. On LWJGL3, they're greyed out.
* It's already known that libGDX's audio recording functionality is dead as of Android 6 Marshmallow. It doesn't ask for permission at runtime.

It would be good to re-add the newline at the bottom of `build.gradle`, as there was previously one and the formatter that runs on Java code adds one. In practice, it makes no odds to my life.

obigu:
I'd love to know why it used to work before but yeah, it's broken now and it makes sense it requires a similar task to the one we use on Gdx Setup template Android apps. Thanks for the fix!

MobiDevelop:
I assume it worked before because the old fetch natives (ant) copied the files to the right place and the corresponding Gradle task just invoked that ant fetch. I'm guessing when maven/ant was removed, something like this should have been added to replace what fetch.xml was doing.

PokeMMO:
Ah you're right, so it was my fault again.

https://github.com/libgdx/libgdx/blob/gdx-parent-1.10.0/fetch.xml#L86

PR #6796, opened 2022-02-16
Add support for new jnigen native generation 

Berstanio:
This add support for the changes implemented in https://github.com/libgdx/gdx-jnigen/pull/28 and https://github.com/libgdx/gdx-jnigen/pull/31
After that it should be possible to run libgdx apps on M1 arm64 simulator.
But there was a problem with arm64 simulator and OpenGLES. I can't exactly remember what was wrong and it will take some time until I have a M1 again to test it, but maybe someone else could test it on a M1 arm64 simulator.
But it was probably one of the two cases:
1. App was working there was just extreme console error spam
2. App was not working and there was extreme console error spam


PokeMMO:
gdx-jnigen v2.3.0 has been released, can you update this to no longer rely on snapshot?

Berstanio:
> gdx-jnigen v2.3.0 has been released, can you update this to no longer rely on snapshot?

Done!

PokeMMO:
Sorry I haven't had time to look into this more, my only concern left is build-objectal.sh and if it will properly copy natives in a way the build/publish can understand. Also, if it will require manually deleting the old binary to not have duplicated files.

Berstanio:
> Sorry I haven't had time to look into this more, my only concern left is build-objectal.sh and if it will properly copy natives in a way the build/publish can understand. Also, if it will require manually deleting the old binary to not have duplicated files.

build-objectal.sh will copy the xcframework into the libs folder. From there it gets picked by this line, because it contains *.a files.
https://github.com/libgdx/gdx-jnigen/blob/86c195174743794e30618b487baa2c41c0af2216/gdx-jnigen-gradle/src/main/java/com/badlogic/gdx/jnigen/gradle/JnigenIOSJarTask.java#L19

About the manual deleting, since the jnigen task clears out the libs folder I think it is not a problem. And build-objectal.sh clears the build folder.

PokeMMO:
Sounds good to me, will merge once conflicts are fixed.

PR #6797, opened 2022-02-16
GWT: Use canvasId if rootPanel is not defined

yuripourre:
Useful (required) when you embedding the GWT application in an html element that already has an id defined.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -95,10 +95,6 @@ public abstract class GwtApplication implements EntryPoint, Application {
 		return GWT.getHostPageBaseURL() + "assets/";
 	}
 
-	protected String getRootId () {
-		return "embed-" + GWT.getModuleName();
-	}
-
 	@Override
 	public ApplicationListener getApplicationListener () {
 		return listener;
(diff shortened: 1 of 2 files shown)


MobiDevelop:
Any reason not to make it either a new property of the config or just setting the rootPanel on the config?

yuripourre:
@MobiDevelop by config you mean gradle config?

MobiDevelop:
Sorry, I meant the GwtApplicationConfiguration.

yuripourre:
@MobiDevelop that's a very good idea

yuripourre:
@MobiDevelop done! What do you think?

MobiDevelop:
Hmm, not sure about that implementation - to be honest I'm not exactly sure how the rootPanel and canvasId relate but I get the impression they are two separate things.

Currently away from a computer so hard to dig into at the moment.

yuripourre:
Ok, thanks @MobiDevelop. canvasId is unused btw.
```
	/** the Panel to add the WebGL canvas to, can be null in which case a Panel is added automatically to the body element of the
	 * DOM **/
	public Panel rootPanel;
	/** the id of a canvas element to be used as the drawing area, can be null in which case a Panel and Canvas are added to the
	 * body element of the DOM **/
	public String canvasId;
```

MobiDevelop:
Ah, in that case it seems fine. I wonder what it was used for previously (and when it stopped being used).

yuripourre:
Reading the code again, the canvasId seems to be added as a property by libgdx so it can be referenced by external integrations.

Also, the root panel is not the "embed-html" div, it is the panel added inside the div. Converting to draft, I will change properly when I have some time.

PR #6801, opened 2022-02-21
Update LWJGL3 to 3.3.1

PokeMMO:


tommyettinger:
The LWJGL 3.3.1 changelog doesn't say too much that affects libGDX, but this stuck out to me:
> Core: Fixed extra NUL in string returned from SharedLibrary::getPath() on Linux & macOS. (# 713)

This doesn't change anything with how libGDX loads its natives, right?

PokeMMO:
It looks like it's just a minor bugfix when getting the path of an already loaded shared library. https://github.com/LWJGL/lwjgl3/pull/713

crykn:
LWJGL 3.3.1 is also relevant for @badlogic's gdx-lwjgl3-glfw-awt-macos backend:
> GLFW: An alternative macOS build now supports Cocoa calls on any thread.
> - A different window toolkit (AWT, JavaFX, etc.) may now run at the same time as GLFW. It must be initialized before glfwInit is called.
> - Enabled with Configuration.GLFW_LIBRARY_NAME.set("glfw_async");.

PR #6802, opened 2022-02-23
fixed forced attributes type/normalize in GWT backend

mgsx-dev:
Vertex attributes type and normalisation was forced (not respected) in GWT backend for no reason which prevent use of custom types (byte, short, etc...).

Context: We faced this issue because gdx-gltf extension now uses unpacked color with normalized unsigned short instead of floats for vertex color (due to recent changes in Blender and GLTF exporter) which breaks with GWT backend.


tommyettinger (approved):
This looks good; simpler, cleaner, and consistent with the non-GWT version. Here's the original side-by-side comparison Xpe posted, for anyone not familiar with how the bug manifests without this PR: Blue bad!

tommyettinger:
I think this should be mergeable as-is, but I requested a review from @MrStahlfelge just to get some confirmation that nothing crazy is going to happen from a merge.

MrStahlfelge:
That change is restricted to only a single emu object so definitely not going to have unintended impact.

PR #6803, opened 2022-02-24
Fixes music not looping when setting setLooping() before first play()

obigu:
#6654 Introduced an issue causing music not to loop when setting `setLooping(true)` before first `play()`. ObjectAL is a bit weird because the only thing that can't be set beforehand when calling `playFromFile()` is `numberOfLoops` and if not passed is set to 0 instead of taking the value of the property

This is the relevant code on [https://github.com/kstenerud/ObjectAL-for-iPhone/blob/master/ObjectAL/ObjectAL/AudioTrack/OALAudioTrack.m](OALAudioTrack).

```
- (bool) playUrl:(NSURL*) url
{
	return [self playUrl:url loops:0];
}
```

Basically the fix is to store if `looping` has been set in order to pass it on first `play`.

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSMusic.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSMusic.java
@@ -59,7 +59,7 @@ public class IOSMusic implements Music {
 			// OALAudioTrack needs to execute preloadURL() once to store the file path. From then on we avoid
 			// calling it again to avoid instantiating a new AVAudioPlayer every time.
 			if (!initialized) {
-				if (looping)
+				if (!looping)
 					initialized = track.playFile(filePath);
 				else
 					initialized = track.playFile(filePath, -1);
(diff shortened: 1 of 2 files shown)


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSMusic.java
```diff
@@ -58,7 +59,10 @@ public void play () {
 			// OALAudioTrack needs to execute preloadURL() once to store the file path. From then on we avoid
 			// calling it again to avoid instantiating a new AVAudioPlayer every time.
 			if (!initialized) {
-				initialized = track.playFile(filePath);
+				if (looping)
+					initialized = track.playFile(filePath);
+				else
+					initialized = track.playFile(filePath, -1);
```
tommyettinger, line 65: @Berstanio noticed this, not me, but it looks like the -1 might need to be double-checked.

> in the setLooping code it looks like "-1" needs to be set, if there should be looping:
>    ```java
>    @Override
>    public void setLooping (boolean isLooping) {
>        track.setNumberOfLoops(isLooping ? -1 : 0);
>        looping = isLooping;
>    }
> ```
> Do I miss something?

tommyettinger (changes requested):
I'm requesting changes just because I want this to be double-checked; it might be correct now.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSMusic.java
obigu: Absolutely, thanks!

tommyettinger (approved):

PR #6811, opened 2022-03-07
Update to Android Gradle Plugin 7.0.4 and API level 31

Frosty-J:
An issue that's been cropping up a fair bit lately is gdx-setup's incompatibility with Android API level 31 and higher. This is because Android Build Tools 31 removed DX (D8 has replaced it), but Android Gradle Plugin 4 isn't aware of this change due to being an old version.

A potential future issue that has cropped up since the creation of this pull request is Google has now set [the date at which API level 31 will be required](https://developer.android.com/google/play/requirements/target-sdk) for new apps: August 2022 (or November for updates to existing apps).

So I've updated AGP to version 7.0.4. The latest version is ~~7.1.3~~ 7.2.0, but this requires Android Studio ~~Bumblebee~~ Chipmunk, which IntelliJ IDEA has yet to reach feature parity with. [It is planned for 2022.2 EAP3](https://youtrack.jetbrains.com/issue/IDEA-287437/Merge-Android-Plugin-2021-1-1-Final-and-2021-2-1-Final#focus=Comments-27-6096940.0-0). So sticking with 7.0.4 should ensure everyone can use their new projects (unless on a very outdated IDE - can't do anything about that).

The target API level has also been bumped up to 31. This is Android 12, which has been out for 5 months now. Could go to 32, but that throws warnings when using AGP 7.0, and it's probably unwise to advance multiple versions at a time anyway. We don't need to do anything more to support 31 - #6615 added the required `android:exported`.


Frosty-J:
Oh. What do we do about the fact GitHub Actions is using Java 8 while Android Gradle Plugin requires 11?

tommyettinger:
It's probably not that hard to update GH Actions to use Java 11, but we would need to be careful about the difference in NIO behavior. I think this is already handled by the Gradle build, and if we use Java 9 or higher, it will set the release version to 8 (keeping NIO behavior to match 8).

Frosty-J:
Yes, I don't think it should be too hard either. I tried this in `build-publish.yml` but I don't have permission to update it.

```
- name: Set up JDK 11
  uses: actions/setup-java@v2
  with:
    distribution: 'zulu'
    java-version: '11'
```

SimonIT (dismissed):
Tested android, gwt, lwjgl and lwjgl3 with Java 17 and works fine

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Fri Sep 30 05:10:25 BST 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
tommyettinger, line 5: This could be updated to 7.4.1, but there aren't many important features in 7.4.x relative to 7.3.x.

tommyettinger (dismissed):
This looks good, but my only concern is if we now require Java 11+ to build libGDX, and whether docs will need to be updated to reflect that. I definitely think we should be using AGP 7.0.4 (or higher, as it becomes available) to avoid the many issues with compatibility on current Android installations.

SimonIT:
Two things came to my mind:

- I had to add these lines https://github.com/diffplug/spotless/issues/834#issuecomment-819118761 to be able to build after I made some local changes (with JDK 17)
- Afters seeing #6828 I ran `./gradlew javadoc` (on JDK 17) and it just worked. Maybe the fix isn't needed than anymore?

gradle/wrapper/gradle-wrapper.properties
Frosty-J: I left it as 7.3.3 for parity with what gdx-setup produces. I'd be willing to change it to 7.4 in both places if someone can promise it won't double the time it takes for this to be merged.

Frosty-J (commented):
No idea why GitHub claimed there was merge conflicts. (re: https://github.com/libgdx/libgdx/pull/6811/commits/a6f5f27219fe131b37e8f015d7e1da094def24c2 - was hoping this would have linked to that commit)

SimonIT:
> It didn't just work for me. Got the same error as reported. Ughhh, I hope this doesn't vary by distribution. I used java-17-openjdk-amd64. And Gradle 7.3.3, seeing as 6 is incompatible.

Didn't see, that you have the fix also in this PR. Doesn't work without it for me, too

gradle/wrapper/gradle-wrapper.properties
tommyettinger: 7.4.1 is a pretty minor set of changes; we should probably wait until 7.5 or whatever Gradle supports Java 18.

obigu:
This is an important fix considering Gdx Setup is now currently broken for most users but I'm concerned of the impact it may have now that we have quite a stable build.

**IMO, if merging this change means the 1.10.1 release will be delayed (considering it makes important changes to the build and requires code owners approval if I'm not wrong) I'd rather remove the milestone.**

Considering this is just critical for Gdx Setup, what we could do is merge this just after 1.10.1 release, giving us time to make sure nothing els breaks and having an available working Gdx Setup Snapshot (currently the one publicly available) immediately.

Frosty-J:
I'd be in favour of the following. Call it a fantasy if you will:

1. Merge #6828. It's a simple change that I don't see any downsides to, and would get the milestone list down to only this pull request.
2. Create a 1.11.1 milestone list.
3. Move this pull request from the 1.11.0 list to 1.11.1.
4. Release libGDX 1.11.0 ("1.10.1"). Even if there are loose ends, we can't stall its release forever.
5. Return libgdx.com to having downloads for both stable and nightly builds of gdx-setup.
6. Merge this.
7. Release libGDX 1.11.1 later this year.

While I would of course like to have this merged for libGDX 1.11.0, that release is already so overdue I wouldn't want this to be responsible for delaying it.

PokeMMO:
This seems to break `copyAndroidNatives`.

The order of operations is now incorrect and will result in the first generated APK to not contain native libraries. Changing it to the following snippet seems to fix it.

```
tasks.whenTaskAdded { packageTask ->
    if (packageTask.name.contains("merge") && packageTask.name.contains("JniLibFolders")) {
        packageTask.dependsOn 'copyAndroidNatives'
    }
}
```

This would be a breaking change requiring every user to migrate their android `build.gradle`.

Frosty-J:
Users already have to make the `duplicatesStrategy(DuplicatesStrategy.EXCLUDE)` change for Gradle 7 to work, so I guess it would be fitting for us to also provide instructions to upgrade existing projects to AGP 7 if need be.

Before changing anything, I'd like to better understand this `copyAndroidNatives` business, as I haven't had any trouble with missing natives in either AGP 4 or 7, nor have I heard of anyone else having trouble with it.

The first test I did was with Android Studio:

1. Generate an AGP 7.0.4 project using this pull request's gdx-setup.
2. Build -> Build Bundle(s) / APK(s) -> Build APK(s)

The second test I did was with IntelliJ IDEA:

1. Open this pull request's directory (fresh pull - no natives downloaded beforehand).
2. Run gdx-tests on Android.

I repeated the first test using your snippet, being sure to first remove `build` and `libs`, and that also seemed to work. So I tried removing that section altogether, after which the libraries were missing from the APK. So it definitely does something - I'm just admittedly left scratching my head on the difference of what was there before and your new version.

PokeMMO:
As discussed in discord, it seems to be limited to release builds, such as `assembleRelease`.

tommyettinger:
OK, this is crazy. Since 1.11.0 isn't released yet and, to be honest, probably won't be for a while at this rate, what do we have left to do for this PR? The request to libgdx-signing I think was accidental and unnecessary, but I'm not sure. This does remain a purely-gdx-setup thing, so it could be merged after 1.11.0 is released, but if we get a flood of updating users who all see breakage due to the antiquated AGP version, it won't be pretty. However it could mean a flood of users using [gdx-liftoff](https://github.com/tommyettinger/gdx-liftoff) instead, so selfishly I could benefit from delaying this...

crykn (commented):
Hey James, just talked to Tomski. Would you mind changing the JDK 11 stuff in the `build-publish.yml` file as well? I think it'd be best to update all the used JDKs and and not just the ones that explicitly require 11.

.github/workflows/build-publish.yml
```diff
@@ -318,10 +318,6 @@ jobs:
       - name: Pack desktop natives
         run: |
           ./gradlew jniGen
-          ant -f gdx/jni/build.xml pack-natives
-          ant -f extensions/gdx-box2d/gdx-box2d/jni/build.xml pack-natives
```
Tom-Ski, line 314: Those are generated by the jnigen task from the jnigen gradle plugin

Frosty-J:
What the fuck? We can see clearly 0dcc90d's job in this PR worked 8 days ago. What has changed?

Frosty-J:
libGDX is working with AGP 7.2.1 in IntelliJ IDEA 2022.2 EAP 3.

Screenshot

I won't make any changes right now as obviously the final release isn't yet on our doorstep, but I'd like to read the room on if all reviewers would be okay with this pull request being updated to the new versions supported by IDEA at a more appropriate time, if it's still open by then. There isn't the same urgency as with 7.0, but with gdx-setup's slow release cycle I think it wouldn't be a bad idea.

Tom-Ski (approved):

PR #6812, opened 2022-03-09
Get color depth on GWT, refresh rate on Android

Frosty-J:
Is there any particular reason why we're always returning 8 bits per pixel (256 colours) on the GWT backend?

I've set it to use `screen.colorDepth` instead, which has [supreme compatibility](https://caniuse.com/?search=colordepth). It's identical to `pixelDepth` but is supported by Internet Explorer 4 (and doesn't seem deprecated in any capacity - I don't know why there's two identical properties) so I couldn't resist.

In non-Chromium-based browsers I think it always returns 24bpp. This isn't particularly useful, I must admit, but at least it's one strange number out the way.

There's still no clean way of getting the monitor's refresh rate that I know of. Could measure the frame rate, but it's just not the same.

***

This pull request has now expanded to include Android in the mix, which was just returning 0 for refresh rate and colour depth. The refresh rate gets rounded to the nearest integer - I checked the desktop backend on Linux and the same behaviour seems to occur there, except any rounding must occur at the LWJGL/GLFW level. I don't know what behaviour we'd see on variable refresh rate devices - probably just the maximum refresh rate?

Devices on Android 4.2 and newer will additionally benefit from more accurate screen resolution detection. I don't advise people use this in their game, but there may be some circumstances where it's useful to know the device's actual resolution, including status bar and whatever that the game might be unable to access.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
@@ -229,6 +229,10 @@ public class GwtGraphics extends AbstractGraphics {
 		return $wnd.screen.height;
 	}-*/;
 
+	private native int getColorDepthJSNI () /*-{
+		return $wnd.screen.colorDepth;
+	}-*/;
+
 	private native boolean isFullscreenJSNI () /*-{
 		// Standards compliant check for fullscreen
 		if ("fullscreenElement" in $doc) {
@@ -339,7 +343,7 @@ public class GwtGraphics extends AbstractGraphics {
 	@Override
 	public DisplayMode getDisplayMode () {
 		double density = config.usePhysicalPixels ? getNativeScreenDensity() : 1;
-		return new DisplayMode((int)(getScreenWidthJSNI() * density), (int)(getScreenHeightJSNI() * density), 60, 8) {};
+		return new DisplayMode((int)(getScreenWidthJSNI() * density), (int)(getScreenHeightJSNI() * density), 60, getColorDepthJSNI()) {};
 	}
 
 	@Override
--- a/CHANGES
+++ b/CHANGES
@@ -53,6 +53,7 @@
 - API ADDITION: Emulate Timer#isEmpty on GWT
 - API Addition: Bits add copy constructor public Bits (Bits bitsToCpy)
 - API Addition: Added List#drawSelection().
+- API Fix: Display mode on GWT now attempts to return the correct color depth.
 
 [1.10.0]
 - [BREAKING CHANGE] Android armeabi support has been removed. To migrate your projects: remove any dependency with natives-armeabi qualifier from your gradle build file, this apply to gdx-platform, gdx-bullet-platform, gdx-freetype-platform and gdx-box2d-platform.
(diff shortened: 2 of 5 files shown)


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
```diff
@@ -339,7 +343,8 @@ private native void exitFullscreen () /*-{
 	@Override
 	public DisplayMode getDisplayMode () {
 		double density = config.usePhysicalPixels ? getNativeScreenDensity() : 1;
-		return new DisplayMode((int)(getScreenWidthJSNI() * density), (int)(getScreenHeightJSNI() * density), 60, 8) {};
+		return new DisplayMode((int)(getScreenWidthJSNI() * density), (int)(getScreenHeightJSNI() * density), 60,
+			getColorDepthJSNI()) {};
```
tommyettinger, line 347: This looks good, but is the color depth used anywhere in GWT? In any case, 8 appears to have been incorrect, since Chrome would have always returned 24 until a few years ago. Approving because this looks ,like it achieves the correct behavior; if there are issues with this, they're probably because something assumes 8bpp on GWT incorrectly.


...bits per pixle.

tommyettinger (approved):
Approved! It looks like browsers assume the opposite of what libGDX was ensuring, and things like Chrome used to not even notice if a monitor had less than 24bpp.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
Frosty-J: Nothing in libGDX relies on this. On GWT, I mean, where it's impossible to change the desktop display mode.

It's possible a game might use this value - I've seen titles recommend or enforce a particular colour depth, all of which predate libGDX. But they'd be getting the wrong value on GWT. I haven't checked mobile!

Frosty-J: 0fps and bpp on Android. Includes alpha on iOS. What a mess.

https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L690

https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java#L409-L410

tommyettinger: Oh, wow, that is a mess. So I suppose this means it doesn't matter currently, but could if > 24bpp monitors or mobile devices start becoming common.

Frosty-J: 10-bit displays are already a thing for computers (no surprise there), Android and iOS. Don't know why you'd want to detect them, but may as well. Think I can I can improve the Android version....

Frosty-J: So I don't think there's a way of getting an actual bit depth from an Android device, but there is `Device#isWideColorGamut` and `#Display#isHdr`. Not of much use to us, but it's interesting to note. I really need to catch some Z's.

obigu:
Looks good to me, we can get rid of the deprecated call in the future (no need to have yet another version check).

PR #6815, opened 2022-03-10
Fix vulnerability reported by CVE-2017-0702

### LINKED ISSUE #6814: Potential vulnerable code exists in LibGDX
#### Version
Libgdx master version

#### Issue description
Hi. I confirmed that the patch for CVE-2017-0702 reported by the Android team modified the LibGDX code. I'm not sure that the vulnerability affects the original LibGDX, but it would be good to check from a security standpoint.

Although I did not reproduce the vulnerability, I recommend checking for this vulnerability, as there have been cases that CVE-2017-0700, which was reported by Android, affected LibGDX (now it is patched, please see #5737 and #5741).

- Related file:
./gdx/jni/gdx2d/gdx2d.c

- Patch for CVE-2017-0702: https://android.googlesource.com/platform/external/libgdx/+/5d46e9a1784c02f347af107a978fe9fbd7af7fb2

Thank you.

yuripourre:
I didn't test it, I just copy+paste and removed the Android Log calls, the code makes sense because it verify the size of the image before trying to decode it.

Please, test it before merge.

@badlogic 

Closes #6814

### DIFF
--- a/gdx/jni/gdx2d/gdx2d.c
+++ b/gdx/jni/gdx2d/gdx2d.c
@@ -358,7 +358,18 @@ static inline void clear_RGBA4444(const gdx2d_pixmap* pixmap, uint32_t col) {
 }
 
 void gdx2d_clear(const gdx2d_pixmap* pixmap, uint32_t col) {
+	if (pixmap == 0)
+		return;
+	
 	col = to_format(pixmap->format, col);
+
+	// Check for malformed Pixmap
+	size_t requestedSize = pixmap->width * pixmap->height * sizeof(col);
+	size_t pixelsSize = sizeof(pixmap->pixels);
+
+	if (requestedSize > pixelsSize) {
+		return;
+	}
 
 	switch(pixmap->format) {
 		case GDX2D_FORMAT_ALPHA:
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/ETC1.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/ETC1.java
@@ -71,8 +71,8 @@ public class ETC1 {
 				while ((readBytes = in.read(buffer)) != -1) {
 					compressedData.put(buffer, 0, readBytes);
 				}
-				((Buffer) compressedData).position(0);
-				((Buffer) compressedData).limit(compressedData.capacity());
+				((Buffer)compressedData).position(0);
+				((Buffer)compressedData).limit(compressedData.capacity());
 			} catch (Exception e) {
 				throw new GdxRuntimeException("Couldn't load pkm file '" + pkmFile + "'", e);
 			} finally {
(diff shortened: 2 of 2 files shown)


yuripourre:
Closing in favor of #6818

PokeMMO:
I don't believe #6818 fixes anything from this PR. stb_image is separate from the clear/ETC1 code.

yuripourre:
@PokeMMO oh, got it, I thought with your changes this PR was obsolete.

PokeMMO (approved):

tommyettinger (approved):
Looks perfectly reasonable, and matches the fixes Google has apparently been using for a while.

tommyettinger:
The ETC1Test and KTXTest both pass; I don't think we have an existing test for malformed ETC files.

yuripourre:
@PokeMMO @tommyettinger @redy5 @NathanSweet this PR seems to break Pixmap rendering, what do you guys think about reverting until it is properly tested?

NathanSweet:
I mentioned the problem I'm having [here](https://github.com/libgdx/libgdx/commit/023cd9ee155d14c718cfa57dcff4799db57054b1#commitcomment-69674383), though I just made a wild guess as to which commit caused the problem so maybe it is this PR and not stb_image.

tommyettinger:
I really don't see how the checks for invalid data could be causing issues with FreeType. FreeType definitely shouldn't be relying on the buffer under/overflow that was permitted by Pixmap's internals before this PR. We can try reverting this, but we should have some way to bring back at least the changes that we want to prevent malicious images from messing with buffers.

tommyettinger:
It looks like a new version of stb_image is in the works; a preview with some kind of emergency fix for a rare type of input image was posted by nothings (STB) here: https://gist.github.com/nothings/4f15affa5aaa3afae1b9bcf972f07514

badlogic:
Reverting this temporarily for testing.

badlogic:
The ETC1 fix is correct, the fix in `gdx2d_clear` is not. There's also no clear way to fix `gdx2d_clear()` as we'd need to have info on the actual size of `pixmap->pixels`, which depending on how it was initialized may not be available. There's only a single constructor `public Gdx2DPixmap (ByteBuffer pixelPtr, long[] nativeData)` can be affected by a mismatch of `pixels->width * pixels->height * bytesPerPixel` (note `sizeof(col)` in the code is wrong as well...).

I can't find a call to `public Gdx2DPixmap (ByteBuffer pixelPtr, long[] nativeData)` in the entire libGDX code base. It may be used by some users, tho I doubt that. If we want to "fix" this CVE, that constructor could be removed. I see no scenario where there could be an invalid heap access though.

In any case, if we choose to re-merge this, with modifications to `gdx2d_clear`, we need to test that the respective `Pixmap` and `Gdx2DPixmap` methods work as intended.

badlogic:
Alright, I've revied the other "fix" as well. It too is broken and doesn't fix anything.

I decree that we will not merge this CVE fix at all. It's broken and should instead be handled on the Java side.

tommyettinger:
Sounds like a better plan!

yuripourre:
@badlogic thank you so much for looking into it and sharing your findings.

PR #6819, opened 2022-03-13
GWT: Expose xrCompatible attribute

yuripourre:
This PR exposes the xrCompatible attribute.

This change is necessary to enhance the performance of WebXR applications on devices with multiple GPUs. When this attribute is set to `true` the created WebGL context tries to use the strongest GPU on the device.

It only affects devices with multiple GPUs.

This is not a breaking change, this PR shouldn't affect any existing LibGDX app.

More details at: https://www.w3.org/TR/webxr/#contextcompatibility


yuripourre:
@MobiDevelop @MrStahlfelge sorry for bothering you again but if you guys have some time to review it I would appreciate it a lot. Please share your insights.

tommyettinger:
I'm not sure why the default is false, or when you would want to change this to true. If xrCompatible were true by default, would that cause problems in any case?

Frosty-J:
> If xrCompatible were true by default, would that cause problems in any case?

[Yes, potentially.](https://immersive-web.github.io/webxr/)

> On some systems this flag may turn on a high powered discrete GPU, for example, or proxy all commands to an on-device GPU. If you are in a situation where you may or may not be using XR, it is suggested that you only call `makeXRCompatible()` when you intend to start an immersive session.  

[It also sounds like it can crash Safari.](https://bugs.webkit.org/show_bug.cgi?id=231423)

Frosty-J:
Ha, the document I referenced has the same contents as Yuri's. You might also find this part of interest:

> This flag introduces slow synchronous behavior and is discouraged. Consider calling `makeXRCompatible()` instead for an asynchronous solution.

yuripourre:
> I'm not sure why the default is false, or when you would want to change this to true. If xrCompatible were true by default, would that cause problems in any case?

@tommyettinger This issue might be relevant, it explains some problems that may occur.

@Frosty-J thanks for flagging it. Yes, during the application creation it will definitely lag but can be handled by the container page before entering the XR mode. I didn't add the makeXRCompatible because I don't have enough time to test it properly.

yuripourre:
@PokeMMO Hey, you must be busy but if you have some time, would you mind reviewing it?

PokeMMO:
Sorry, I haven't really used GWT and am unfamiliar with it.

yuripourre:
@PokeMMO Thank you anyways!

yuripourre:
@MrStahlfelge I believe the browser will ignore unsupported tags. But I don't know how to test it.

yuripourre:
@MrStahlfelge good idea! I just tried that, the browser simply ignored it, not even a warning. The crash @Frosty-J was talking about is related to a poor implementation when Safari recognizes the flag but doesn't handle it properly.

crykn:
I'm not really sure on how to test this change, but if @yuripourre can confirm that it works, I think we can merge this.

yuripourre:
@crykn thanks for taking a look into it. 
One way to test it is using a PC with an integrated GPU + off board video board. Running with the flag set as false should not change anything nor breaking the app. Then you can set the flag as true and check if the external GPU renders the canvas.

I guess for now if the flag *true* doesn't break anything should be safe to merge it. I couldn't profile Oculus Quest 2 but it worked flawlessly.

crykn (approved):

MobiDevelop:
Is there a need to know whether you got a compatible context? If so, is that something that should be exposed? I'm not familiar enough with this to know how you use it, so maybe it's enough just to request it.

yuripourre:
@MobiDevelop that's a good question, usually if it doesn't work and the application requires immersive mode, it will fail in a later point so can be handled there (at least for WebXR).

crykn:
If no one has any objections, I'd like to merge this one as well.

tommyettinger:
No oppositions here. If it defaults to false and won't break any existing code because of that, I see no problems with the option to enable WebXR.

PR #6822, opened 2022-03-16
Allow hiding of the mouse cursor

### LINKED ISSUE #5546: Allow hiding of the mouse cursor.

### LINKED ISSUE #5725: Change Mouse Cursor/Pointer on Android
#### Issue details
Android 7.0 and newer lets developers customize the mouse cursor with different system pointers, custom bitmaps, or hiding the pointer. It would be nice if libGDX supported this.

Official Documentation:
https://developer.android.com/about/versions/nougat/android-7.0#custom_pointer_api

https://developer.android.com/reference/android/view/PointerIcon.html

I wrote a Stack Overflow answer explaining how to use it:
https://stackoverflow.com/questions/15636855/custom-mouse-pointer-on-android/55482761#55482761

libGDX v1.9.9

- [x] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] MacOS

Frosty-J:
Been thinking about cursor stuff lately. We still don't have a way of hiding the cursor without using hacks. So let's fix that. This pull request applies to LWJGL3, Android (7.0 Nougat onwards only) and GWT. LWJGL2 users should ~~stop using LWJGL2~~ carry on with their transparent pixmap hack. Might be applicable to iOS 14 onwards, but unfortunately I don't do iOS development. I also added libGDX's other existing cursor options to the Android backend.

I would like to support more system cursors than this, but GLFW is holding us back. For example, `zoom-in` and `zoom-out` have very good support but GLFW has no bindings for these. If anyone has thoughts on the possibility of supporting other cursors on Android/GWT but not desktop, or also on desktop via different means, I'd be happy to hear them.

I'm unclear on if new classes should have the `Copyright 2011` block at the top of the file. Some classes have it; others not. So I've left it out.

If merged, this resolves #5546, partially resolves #5725, and renders #6218 redundant. I chose the `setSystemCursor()` approach over `setCursorHidden()` because this aligns nicely with how things are done on HTML5 and Android. It should be possible to also support `setCursor()` on Android but I won't implement that in this pull request (you'd probably be better off using `drawable-mdpi` etc. anyway).

```java
Gdx.graphics.setSystemCursor(Cursor.SystemCursor.None); // hide cursor
Gdx.graphics.setSystemCursor(Cursor.SystemCursor.Arrow); // show default cursor
```

Discord screenshot. MrStahlfelge says "make it happen"


crykn (approved):
I like that `Graphics#setSystemCursor(SystemCursor.None)`-approach considerably better than the `Input#setCursorHidden()` one! Also it supports more platforms, which is nice. 

> I would like to support more system cursors than this, but GLFW is holding us back.

There seems to be quite an extensive discussion on this over on their issue tracker: https://github.com/glfw/glfw/issues/427

raeleus (approved):
It works as expected on my systems and this is a better change than my PR. Well done!

PR #5181, opened 2018-04-22
Added FileHandle#map(), can be used to memory map a file

PokeMMO:
Utility function to provide easy access to memory mapped files. Defaults to native byte order and can be seamlessly used in JNI contexts.

Android files must be uncompressed in APK for mapping to occur.

GWT/iOS are untested.

PR #6825, opened 2022-03-22
FileHandle#map() should use the File from FileHandle#file()

PokeMMO:
Fixes a minor mistake from #5181 which would prevent mapping files of certain `FileType`s on certain platforms.


crykn (approved):
LGTM

PR #6826, opened 2022-03-22
Set duplicatesStrategy = DuplicatesStrategy.EXCLUDE in gdx/build.gradle

PokeMMO:
copyIdeaResources gets run while compiling the jar, leading to duplicate resource entries.

https://github.com/libgdx/libgdx/blob/master/gdx/build.gradle#L53-L57


obigu (approved):
Yep, nice catch!

PR #6828, opened 2022-03-22
Don't fail on javadocs error for gwt

PokeMMO:
Java 11's javadoc seems to ignore any classpath entries in special packages. This leads to the gwt emu classes being unavailable for javadoc processing. Resulting in the following error.

```
$ ./gradlew :backends:gdx-backend-gwt:javadoc 

> Task :backends:gdx-backend-gwt:javadoc

libgdx/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL20.java:22: error: cannot find symbol
import java.nio.HasArrayBufferView;
               ^
  symbol:   class HasArrayBufferView
  location: package java.nio
1 error
```
This fix isn't ideal, but I don't really have a better idea how to make this work. Existing [Xdoclint](https://github.com/libgdx/libgdx/blob/7b7d78c175c2f3ed288215cbed9587edb35f70ab/build.gradle#L95) is applied, but doesn't prevent this being a fatal error.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,6 +95,12 @@ allprojects {
             options.addStringOption('Xdoclint:none,-missing', '-quiet')
         }
     }
+
+    if (JavaVersion.current().isJava9Compatible()) {
+        tasks.withType(Javadoc) {
+            options.addStringOption("-release", "8");
+        }
+    }
 }
 
 configure(subprojects - project(":tests:gdx-tests-android")) {


Frosty-J:
Here exists an alternative approach, but I don't know what the ideal solution is:

https://github.com/libgdx/libgdx/blob/49efe9f697470be5a78bf41e5957cf3adf4e9adf/build.gradle#L99-L103

PokeMMO:
Your solution seems to be much better, I will update the pull request.

Thanks!

tommyettinger (commented):
I'm somewhat surprised configuration like this wasn't already present; I always have similar code in my libraries, just for JavaCompile instead of Javadoc. I usually prefer using the latest possible Javadoc version because the user experience is much better with the current generated output (No HTML frames! A search box!). It seems like that isn't an option here because of the Java 11 doc bug. Do we know if more recent JDKs have this bug fixed, i.e. is it confined to Java 11?

PR #6829, opened 2022-03-24
Remove unnecessary isPlaying gwt audio

SimonIT:


crykn:
Does this still return `false` when `isPlaying()` is called before `play()`?

SimonIT:
Yes, tested it

crykn (approved):

PR #6838, opened 2022-03-29
Never stall with AssetManager on GWT

Frosty-J:
I haven't given this solution a great deal of thought, but I've just spent a long time getting to the bottom of the problem and I think you can guess what moods I'm in. I'd like it fixed so others don't have to lose sleep over the same problem. Or if this is a stupid solution but no-one comes up with a better way of doing it, at least document to never call it (in javadocs and/or wiki) with the `millis` overload on GWT.

Passing a number of milliseconds into `AssetManager#update` on GWT is a terrible idea. Not only do you get a reduced frame rate with no reduction in load time (even on localhost), you're also at great risk (only when using a custom asset filter, I think) of running into the `File not prefetched, but finishLoading was probably called` error due to the likelihood of exceeding [100,000 ticks](https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/assets/AssetLoadingTask.java#L98-L99) (one tick == one `AssetLoadingTask#update`). 

This is especially likely if the files are not cached and for some reason more likely when running in an iframe on Google Chrome (Chrome tends to be faster than Firefox for me; not sure where iframes fit into it though). If anyone does their own testing, note that incognito mode isn't a perfect testing zone - it still seems to be affected by some level of caching even if not the full-on disk cache.

My theory is the ticks accumulate extremely quickly while waiting for a file to download, since it's stuck in this loop unable to do anything else. `update(1)` would practically guarantee a crash on the first itch.io-hosted run of my game on GWT, but `update()` is perfectly fine.

### DIFF
--- a/gdx/src/com/badlogic/gdx/assets/AssetManager.java
+++ b/gdx/src/com/badlogic/gdx/assets/AssetManager.java
@@ -433,8 +433,7 @@ public class AssetManager implements Disposable {
 	 * of a single task that happens in the GL thread takes a long time.
 	 * @return true if all loading is finished. */
 	public boolean update (int millis) {
-		if (Gdx.app.getType() == Application.ApplicationType.WebGL)
-			return update();
+		if (Gdx.app.getType() == Application.ApplicationType.WebGL) return update();
 		long endTime = TimeUtils.millis() + millis;
 		while (true) {
 			boolean done = update();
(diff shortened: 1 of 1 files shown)


crykn (approved):
This seems like a sensible solution!

MrStahlfelge (changes requested):
Not opponing the "fix", but let's add this to the JavaDoc so no one is surprised.

MrStahlfelge (approved):

PR #6843, opened 2022-04-03
Tiny correction to Jdoc of disableAudio

SomeTroglodyte:
The Jdoc of Lwjgl3ApplicationConfiguration.disableAudio has it the wrong way around. Pretty obvious, and one can easily check Lwjgl3Application line 96 to see what's what, but still.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -110,7 +110,7 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		this.initialVisible = visibility;
 	}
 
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -123,11 +123,11 @@ public void setMaxNetThreads (int maxNetThreads) {
 
 	/** Sets the audio device configuration.
 	 * 
-	 * @param simultaniousSources the maximum number of sources that can be played simultaniously (default 16)
+	 * @param simultaneousSources the maximum number of sources that can be played simultaniously (default 16)
```
tommyettinger, line 126: Still, one usage of "simultaniously" snuck through, heh. Not a big enough problem to worry about.

SomeTroglodyte:
> snuck through, heh. Not a big enough problem to worry about

Nope, but that false false could cost someone five seconds of head scratching

PR #6846, opened 2022-04-11
ShapeRenderer check () method as protected visibility 

fabiitch:
void check (ShapeType preferred, ShapeType other, int newVertices)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/ShapeRenderer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/ShapeRenderer.java
@@ -1171,15 +1171,12 @@ public class ShapeRenderer implements Disposable {
 		polyline(vertices, 0, vertices.length);
 	}
 
-	 /**
-	  * Checks whether the correct ShapeType was set.
-	  * If not and autoShapeType is enabled, it flushes the batch and changes the shape type.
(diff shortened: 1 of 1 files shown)


mgsx-dev:
What's the use case? i can't see a reason why some may want to override it. Also, most of fields used in this method are private.

fabiitch:
hey, its for extends ShapeRenderer. My use case is to draw Filled triangles.

I need to call check() method before use Renderer. 
like 
`public void triangle(Triangle triangle){
super.check(ShapeType.Line, ShapeType.Filled, 6);
//draw
`

mgsx-dev:
Ah yes, my bad, it's totally legit. Could you please make it `final` as well, because there is no reason to override it.

fabiitch:
Sure, thx for your feedback

MobiDevelop:
Feels like the doc in that method needs improvement if it is going to become API.

crykn:
I agree, maybe add a small comment like: `Checks whether the correct ShapeType was set. If not and autoShapeType is enabled, it flushes the batch and changes the shape type. The batch is also flushed, when the matrix has been changed or not enough vertices remain.`

fabiitch:
hey , i do it soon !

MobiDevelop (approved):

PR #6849, opened 2022-04-13
Extract iOS audio session management into methods on IOSAudio interface

dasisdormax:
This removes direct calls to ObjectAL functions from IOSApplication and replaces it with calls to new 'activate' and 'deactivate' functions on the IOSAudio interface.

My motivation is to better support alternative IOSAudio implementations, which manage the audio session themselves. (I personally use one based on AVFoundation).

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
@@ -237,17 +237,18 @@ public class IOSApplication implements Application {
 
 	final void didBecomeActive (UIApplication uiApp) {
 		Gdx.app.debug("IOSApplication", "resumed");
-		audio.activate();
+		audio.didBecomeActive();
 		graphics.makeCurrent();
 		graphics.resume();
 	}
 
 	final void willEnterForeground (UIApplication uiApp) {
-		audio.deactivate();
+		audio.willEnterForeground();
 	}
 
 	final void willResignActive (UIApplication uiApp) {
 		Gdx.app.debug("IOSApplication", "paused");
+		audio.willResignActive();
 		graphics.makeCurrent();
 		graphics.pause();
 		Gdx.gl.glFinish();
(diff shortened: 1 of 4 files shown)


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSApplication.java
```diff
@@ -239,29 +237,13 @@ public IOSScreenBounds getScreenBounds () {
...
 	}
 
 	final void willEnterForeground (UIApplication uiApp) {
-		// workaround for ObjectAL crash problem
-		// see: https://groups.google.com/forum/?fromgroups=#!topic/objectal-for-iphone/ubRWltp_i1Q
-		OALAudioSession audioSession = OALAudioSession.sharedInstance();
-		if (audioSession != null) {
-			audioSession.forceEndInterruption();
-		}
+		audio.deactivate();
```
MobiDevelop, line 246: Hmm, feels like a misnomer to be deactivating the audio when entering the foreground.

dasisdormax:
Thanks for your review and feedback. I think 'deactivate' should be called on 'willResignActive' and the function called in 'willEnterForeground' should be called differently, perhaps 'prepare'. 

Other alternatives I could think of:
- Setting up interface methods that are called the same as the IOS Application lifecycle methods
- Not changing the interface, instead having the developer override of the functions in the Application or Delegate

MobiDevelop:
I'd probably go with following the iOS lifecycle methods myself - seems the most straightforward to know when they're going to be called.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudio.java
```diff
@@ -49,13 +53,20 @@ public void activate () {
...
+
+	@Override
+	public void willResignActive () {
+
+	}
+
+	@Override
+	public void willTerminate () {
+
 	}
```
obigu, line 70: Just a method reordering suggestion, can you move the new `IOSAudio` methods after the `Audio` methods and the `forceEndInterruption()` private method at the bottom so that the constructor is on top.

dasisdormax: Yes, that makes sense.

MobiDevelop (approved):

PR #6850, opened 2022-04-14
Set window title to project name

Frosty-J:
I don't mean to tread on any toes here, but after a few days of silence I'm proposing this solution for undoing the recent change that causes LWJGL3 projects to default to having no window title. This pull request restores the window title for old projects, corrects the javadoc, makes new projects use the project name for the window title (same as other backends) and changes the default project name in gdx-setup to clarify that the project name can and should have spaces and proper capitalisation.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3WindowConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3WindowConfiguration.java
@@ -150,7 +150,7 @@ public class Lwjgl3WindowConfiguration {
 		this.fullscreenMode = (Lwjgl3DisplayMode)mode;
 	}
 
-	/** Sets the window title. Defaults to empty string. */
+	/** Sets the window title. If null, the application listener's class name is used. */
 	public void setTitle (String title) {
 		this.title = title;
 	}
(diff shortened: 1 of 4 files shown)


tommyettinger (approved):
I like it. Nice and clean.

yuripourre (approved):

MobiDevelop (approved):

NathanSweet:
Sorry I didn't see the discussion [on the commit](https://github.com/libgdx/libgdx/commit/061e6e1dbfe548184548b89d40396a250ec1c133) until now.

The original behavior was to throw an NPE for Lwjgl3Application `newWindow`:
```
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "value" is null
	at org.lwjgl.system.MemoryUtil.memLengthUTF8(MemoryUtil.java:2254)
	at org.lwjgl.system.MemoryStack.nUTF8(MemoryStack.java:790)
	at org.lwjgl.glfw.GLFW.glfwCreateWindow(GLFW.java:2195)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createGlfwWindow(Lwjgl3Application.java:521)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:459)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application$1.run(Lwjgl3Application.java:450)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:206)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:164)
```

The javadocs said if the title wasn't set, empty string would be used, so I made that the case. The javadoc updated in this PR now makes it clear 1) the title being null is valid, and 2) the desired behavior when null.

Please check the NPE fix here:
c22cc04a3466b4ffb6be764f72cba0c99e777b2e
This also no longer makes changes to the config passed in.

Frosty-J:
Ah, I see. I haven't played around with making new windows but the new commit looks good.

PR #6851, opened 2022-04-17
Add ability to host GWT module on a different domain than the site

me4502:
This is a port of https://github.com/tommyettinger/gdx-backends/pull/1 to the main LibGDX GWT backend project. The main use case of this PR, is it allows the module and host URLs to differ, allowing for externally hosting the actual application on a CDN, S3 bucket, or similar.

For thin webapps, such as React sites, bundling the LibGDX app within the site can be a problem. 

I wrote about hosting apps with this setup a while ago here, https://madelinemiller.dev/blog/embed-gwt-in-react/ - and have been using the linked PR (which this change is a clone of) for multiple years with multiple LibGDX games without issue.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
@@ -92,7 +92,10 @@ public abstract class GwtApplication implements EntryPoint, Application {
 	public abstract GwtApplicationConfiguration getConfig ();
 
 	public String getPreloaderBaseURL () {
-		return GWT.getHostPageBaseURL() + "assets/";
+		String moduleUrl = GWT.getModuleBaseURL();
+		// Total Length - len("html") - len("/")
+		int correctLength = moduleUrl.length() - GWT.getModuleName().length() - 1;
+		return moduleUrl.substring(0, correctLength) + "assets/";
 	}
 
 	@Override
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/AssetDownloader.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/AssetDownloader.java
@@ -18,6 +18,7 @@ package com.badlogic.gdx.backends.gwt.preloader;
 
 import com.badlogic.gdx.backends.gwt.preloader.AssetFilter.AssetType;
 import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.dom.client.ImageElement;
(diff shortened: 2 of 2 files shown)


yuripourre:
Your solution looks good to me. I had the same issue some time ago. I end up creating my own preloader because of this and other requirements. Thanks for sharing.

yuripourre (approved):

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplication.java
```diff
@@ -92,7 +92,10 @@ public abstract class GwtApplication implements EntryPoint, Application {
 	public abstract GwtApplicationConfiguration getConfig ();
 
 	public String getPreloaderBaseURL () {
-		return GWT.getHostPageBaseURL() + "assets/";
+		String moduleUrl = GWT.getModuleBaseURL();
+		// Total Length - len("html") - len("/")
```
MrStahlfelge, line 96: Please add a comment here what happens exactly, it is hard to understand and "correctLength" variable name does not help here (what is "correct" in this context and why is the returned path not correct?) Somethink like "find asset path next to html path"

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/AssetDownloader.java
```diff
@@ -143,7 +144,14 @@ public void onSuccess (Blob result) {
...
 	public void loadImage (final String url, final String mimeType, final AssetLoaderListener<ImageElement> listener) {
-		loadImage(url, mimeType, null, listener);
+		String crossOrigin = null;
+
+		// Enable CORS if we're running from a different URL to the host page
+		if (!url.startsWith(GWT.getHostPageBaseURL())) {
+			crossOrigin = "anonymous";
+		}
+
+		loadImage(url, mimeType, crossOrigin, listener);
```
MrStahlfelge, line 154: This is good, however, I wonder why only images are concerned and why text files and other assets are not.

MrStahlfelge (changes requested):

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/AssetDownloader.java
me4502: The main issue is that WebGL cannot access image assets without cross-origin, other asset types (audio, text files, shaders, etc) correctly resolve.

MrStahlfelge (approved):

MobiDevelop (approved):

yuripourre:
@crykn can we have this merged? I think it has enough approvals, right?

yuripourre:
@crykn thank you!

Issue #5313, opened 2018-07-18
Expose setting for transparent windows on LWJGL3 backend.


raeleus:
#### Issue details
Please expose the window setting for transparent windows on LWJGL3 backend. This has been a feature of LWJGL since 3.1.4. It is only a matter of adding the window hint GLFW_TRANSPARENT_FRAMEBUFFER. This will allow for some neat effects with an undecorated window. I have tested and it works on my Windows 10 computer.

![Sample Image](https://ray3k.files.wordpress.com/2018/07/transparent-window.png)

#### Reproduction steps/code
Add the following to LWJGL3Application.createGlfwWindow():
`GLFW.glfwWindowHint(GLFW.GLFW_TRANSPARENT_FRAMEBUFFER, GLFW.GLFW_TRUE);`

In your LibGDX desktop app, use the following to clear the screen in the render method:
`Gdx.gl.glClearColor(0, 0, 0, 0);
Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);`

#### Version of LibGDX and/or relevant dependencies
1.9.9-SNAPSHOT

#### Please select the affected platforms
- [ ] Android
- [ ] iOS (robovm)
- [ ] iOS (MOE)
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] MacOS

Tom-Ski:
This is currently broken on windows 10. I dont think it should be added until it works on all platforms.

Tom-Ski:
I've added this as deprecated with the link to issue that breaks the functionailty of this on some systems.

PR #6852, opened 2022-04-18
Case for Removal of Deprecated status on Lwjgl3ApplicationConfiguration.setTransparentFramebuffer()

BoBIsHere86:
With the recent push to make lwjgl3 the default backend for desktop. I believe it's time to take a look at the removal of the Deprecated status that was originally placed on setTransparentFramebuffer() back in 2018 during the closure of this issue. #5313

This might seem insignificant, but having this method Deprecated is a big red flag to anyone who may like to use this pretty cool, though niche, feature. But does it even need to be anymore? I've done some digging and that's what we are here to decide.

First, The original comment in the code references an issue in the GLFW repo from 4 years ago which is still open to this day. [issue 1237](https://github.com/glfw/glfw/issues/1237) Which states that a user with an NVIDIA gpu running windows 10 was having issue with a test program using window transparency. The issue will most likely never be closed as the original user has not commented in about 4 years. Even as recently as this past week there have been other user's questioning whether it is still relevant.

Second, This same problem was brought up in a separate issue later that year in the same GLFW repo by another user with an NVIDIA card [issue 1288](https://github.com/glfw/glfw/issues/1288). Though the difference here is this user followed up on the issue a few months later stating that an NVIDIA driver update resolved this problem. Issue closed. This was about 4 years ago.

Obviously we need more proof than this, so I went about verifying for myself as best I could.
Just to cover all our bases. I created 4 different builds of the default projects and added a transparent window using setTransparentFramebuffer(true)
One for libGDX 1.10.0, one for libGDX-1.10.1-SNAPSHOT, one for libGDX with the gdx-lwjgl3-glfw-awt-macos extension and finally an ANGLE build using gdx-lwjgl3-angle. 

Since the issue originally cited nvidia cards able to test on a win10 laptop with a GTX 860M.
All 4 tests worked perfectly and showed a transparent window. I've supplied a SS below.
nvidiaTest1

I also performed tests on a Win10 AMD pc, and a Win 11 with integrated intel graphics. All Worked.
I tested on an Ubuntu instance in VMWare. All worked, EXCEPT the ANGEL build, it did not show transparency.

I was also able to run these tests on a friends 2011 Macbook Pro, you can see the results below.
macbook

3 of my tests worked but I was unable to test the ANGEL build as during testing I was unable to get it to run, regardless of transparency settings it was throwing a GLFW_API_UNAVAILABLE error. Though I think this was related to the outdated macbook itself.

Lastly I was able to test on a 2014 Mac-Mini
macmini
3 of the test's passed but the ANGLE build did not show any transparency. Just like my testing on the Ubuntu VM.

I reached out to fellow dev's on discord to see if anyone would help me do some testing.
@Frosty-J came through and helped a lot by providing some of  the following.
Seems like windows XP and Vista don't seem to make the cut.
image
image

But anything Windows 7 and above did.
image
image

And just like with my tests on Linux and the mac VM, the ANGLE build doesn't seem to show transparency.
image

That was all we were able to put together for now. But considering the original reason for deprecation stemmed from that initial issue linked in the comment about nvidia cards. I'm not sure how relevant that is 4 years later.

This PR simply removes the Deprecated annotation along with that github link as I don't believe it's been relevant for years. I did choose to leave the "Results may vary on different OS and GPUs." Only because I can't thoroughly test everything.

If we could get more participants to do their own test, it would be appreciated.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -167,8 +167,7 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		this.samples = samples;
 	}
 
-	/** Set transparent window hint
-	 * Results may vary on different OS and GPUs.
+	/** Set transparent window hint Results may vary on different OS and GPUs.
 	 * @param transparentFramebuffer */
 	public void setTransparentFramebuffer (boolean transparentFramebuffer) {
 		this.transparentFramebuffer = transparentFramebuffer;
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -168,9 +168,8 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 	}
 
 	/** Set transparent window hint
-	 * @deprecated Results may vary on different OS and GPUs. See https://github.com/glfw/glfw/issues/1237
+	 * Results may vary on different OS and GPUs.
 	 * @param transparentFramebuffer */
-	@Deprecated
 	public void setTransparentFramebuffer (boolean transparentFramebuffer) {
 		this.transparentFramebuffer = transparentFramebuffer;
 	}
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -167,7 +167,7 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		this.samples = samples;
 	}
 
-	/** Set transparent window hint Results may vary on different OS and GPUs.
+	/** Set transparent window hint. Results may vary on different OS and GPUs. Usage with the ANGLE backend is less consistent.
 	 * @param transparentFramebuffer */
 	public void setTransparentFramebuffer (boolean transparentFramebuffer) {
 		this.transparentFramebuffer = transparentFramebuffer;


tommyettinger:
Should the ANGLE backend be mentioned as being more likely to have issues? Nothing else seemed to have issues, so if people want to make a transparent-background application they should probably be advised that they should be careful with ANGLE more-so than with other backends. It's also possible that the ANGLE issues are easily fixed, but I'm not counting on it, so docs seem best for now.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -167,10 +167,8 @@ public void setBackBufferConfig (int r, int g, int b, int a, int depth, int sten
 		this.samples = samples;
 	}
 
-	/** Set transparent window hint
-	 * @deprecated Results may vary on different OS and GPUs. See https://github.com/glfw/glfw/issues/1237
+	/** Set transparent window hint Results may vary on different OS and GPUs.
```
tommyettinger, line 170: Also there's a missing period between `hint` and `Results`.

tommyettinger (changes requested):
This is extremely pedantic, but it's the only changed line; we might as well get it right.

Frosty-J:
> Should the ANGLE backend be mentioned as being more likely to have issues?

Not sure. To my knowledge the ANGLE stuff is undocumented (beyond what's written in #6672) and I think there should be something on the wiki perhaps about why and how you'd use it. Here are some pros and cons I've thought of. Anything with a question mark I'm unsure about:

<table>
<tr>
<th>Pros</th>
<th>Cons</th>
</tr>
<tr>
<td valign="top">
<ul>
<li>Better compatibility with systems that don't do OpenGL 2 by using DirectX on Windows and Metal on macOS.</li>
<li>DirectX could enable software rendering on Windows when a compatible graphics card is unavailable? (WARP)</li>
</ul>
</td>
<td valign="top">
<ul>
<li>x64 only.</li>
<li>OpenGL ES 2.0 only.</li>
<li>Larger JAR size.<ul><li>Contains x86 libraries that don't work.</li><li>I don't understand the point of the Linux libraries.</ul></li>
<li>No window transparency on macOS/Linux.</li>
<li>Worse performance on Windows/macOS?</li>
</ul>
</td>
</table>

Eh, it would only need a sentence to note its limited compatibility with gdx-lwjgl3-angle. Don't think anyone is maintaining that, unfortunately.

raeleus:
I like the proposed change. It works fine on all my OS's here, but those have been reflected in testing already. I think the documentation on the wiki would be a welcome change.

crykn:
Regarding the ANGLE backend documentation: I'll try to gather some information on it and add it to the 1.11.0 changelog post.

Issue #6848, opened 2022-04-12
[GWT] Java heap exception during compilation


yuripourre:
#### Issue details
It's not possible to have big assets in the assets folder for html. GWT throws a Java heap exception during compilation.

@mgsx-dev was very kind and found the problem, it is happening during md5 calculation, specifically this line: 
https://github.com/libgdx/libgdx/blob/665f920902e543856fdd7d496f6ec422affd46d1/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java#L170

#### Reproduction steps/code
Create a libGDX project and add a huge asset to the assets folder (I tried with a 300MB gltf file with 4k textures) and run `./gradlew html:dist`

#### Version of libGDX and/or relevant dependency 
1.10.0 and probably older versions too

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [X] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

yuripourre:
One possible solution mentioned by @mgsx-dev on Discord is buffering the file: https://stackoverflow.com/questions/9321912/very-slow-to-generate-md5-for-large-file-using-java

tommyettinger:
Having very large files in an HTML project seems like a recipe for sky-high server bills. I think even GitHub Pages restricts bandwidth past some point (and I know I've hit the limit for a repo size before, which is not that much higher than that file's size). Buffering seems like a solution, but I'm not really sure it's a problem that we need to solve. Just the amount of time it takes for users to download a 300MB asset could be, depending on connection, hours...

yuripourre:
@tommyettinger I am testing a lot of examples running it locally, I point my Oculus to localhost. I know that it's a very specific use-case but if you think that worth improving it I can think about fixing it.

yuripourre:
@tommyettinger good to know that, thanks for the tip. As I mentioned is mostly local tests but I will keep that in mind.

MobiDevelop:
Is it possible to use a DigestInputStream while copying the files from /assets to the output directory, then just rename the file afterward to include the hash?

MobiDevelop:
I just tried it... I dropped a 6.5GB file into assets and ran a build with no issues (aside from it being exceedingly slow to copy a 6.5GB file multiple times but I wanted to try a ridiculously large file, because why not. I can put up a PR for review, it's pretty simple.

PR #6853, opened 2022-04-20
GWT - Calculate MD5 hash while copying file to avoid errors on very large files

MobiDevelop:
For GWT, we calculate an MD5 hash of the file contents of each asset to assist browsers with caching. The current implementation reads the file fully into a byte array, calculates the MD5, adds it to the destination file name, then copies the source file to the output path. While this is fine for small files (albeit inefficient as we read each file multiple times), it can cause problems with large files.

This proposed implementation instead calculates the MD5 while reading the source InputStream for the copy operation, then renames (moves) the file to include the MD5 hash.

Refer to #4314 for the original addition, and #6848 for the problem this solves.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/PreloaderBundleGenerator.java
@@ -19,6 +19,8 @@ package com.badlogic.gdx.backends.gwt.preloader;
 import java.io.*;
 import java.math.BigInteger;
 import java.net.URLConnection;
+import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


yuripourre:
@MobiDevelop amazing! Have you noticed any gain on speed during GWT build?

MobiDevelop:
A note on the build failure, oddly it has started picking up the emulation of File which doesn't have the getPath() method. Weird that it never did that while I was coding it, and now it is. I wouldn't think the compile time generator would be using the emulation classes - but GWT magic is the darkest of magic.

yuripourre:
Thank you for the PR @MobiDevelop! I will try to test this new logic with the asset I was trying before and check how it performs.

Frosty-J (approved):
This is good, I like it a lot.

PR #6854, opened 2022-04-20
Add collision playground test

yuripourre:
Add this test sample to visualize frustum and ray casting. I am using this to debug collision algorithms and I am pretty confident that more people can find it useful too.

At the moment it can handle AABB and Spheres.

Screenshot from 2022-04-20 12-09-37


crykn (approved):
Looks like a nice addition!

yuripourre:
Thanks @crykn !!

PR #6859, opened 2022-04-28
Update QuadTreeFloat.Java

bilznatch:
Allow for querying of a rectangle within the area, to support using the screen or world view as culling bounds.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
+++ b/gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
@@ -160,7 +160,7 @@ public class QuadTreeFloat implements Poolable {
 	/** @param results For each entry found within the rectangle, if any, the value, x, and y of the entry are added to this array.
 	 *           See {@link #VALUE}, {@link #X}, {@link #Y}, and {@link #DISTSQR}. */
 	public void query (Rectangle rect, FloatArray results) {
-		if (!(x < rect.x + rect.height && x + width > rect.x && y < rect.y + rect.height && y + height > rect.y)) return;
+		if (!(x < rect.x + rect.width && x + width > rect.x && y < rect.y + rect.height && y + height > rect.y)) return;
 		int count = this.count;
 		if (count != -1) {
 			float[] values = this.values;
--- a/gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
+++ b/gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.utils;
 import java.util.Arrays;
 
 import com.badlogic.gdx.utils.Pool.Poolable;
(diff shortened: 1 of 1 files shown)


tommyettinger (dismissed):
Looks good to me, but you will need to run `gradlew :gdx:spotlessApply` to get this to pass the Spotless formatting check.

gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
```diff
@@ -156,6 +157,29 @@ private void query (float centerX, float centerY, float radiusSqr, float rectX,
 		}
 	}
 
+	/** @param results For each entry found within the rectangle, if any, the value, x, and y of the entry are added to this array.
+	 *           See {@link #VALUE}, {@link #X}, {@link #Y}, and {@link #DISTSQR}. */
+	public void query (Rectangle rect, FloatArray results) {
+		if (!(x < rect.x + rect.height && x + width > rect.x && y < rect.y + rect.height && y + height > rect.y)) return;
```
obigu, line 163: Can it be replaced by `Rectangle.contains(x,y)`?

obigu (commented):
I think we should aim to have as much unit test coverage on `math` classes. I don't know why this is in `utils` package but `Octree` is in `maths` and afaik they're equivalent. 
We should probably consider moving the class to the `math` package but, in the meantime, maybe you could you create a `QuadTreeFloatTest` class on `gdx/test/com/badlogic/gdx/utils` to unit test the `query `methods`?

gdx/src/com/badlogic/gdx/utils/QuadTreeFloat.java
bilznatch: > Can it be replaced by `Rectangle.contains(x,y)`?

No, it would need to be replaced with Rectangle.overlaps(Rectangle). To use the check we'd have to change the constructor for the quadtree or create a new rectangle each time we query. As for the unit tests, there are already 2 tests for the quadtree in the main testing package, would it not be better to change the existing query test to include coverage for the rectangle query as well as the radius query?

obigu: Yeah, had a better look. Instead of `x < rect.x + rect.**height**` it should be `x < rect.x + rect.**width**` right?

We can extend the current "integration test" (better than no test at all) but, imo, math classes should be covered by unit tests.

bilznatch: > Yeah, had a better look. Instead of `x < rect.x + rect.**height**` it should be `x < rect.x + rect.**width**` right?
I've committed a change for that. Thanks for finding it.

NathanSweet:
It looks good to me. I'd be OK with moving the class to the math package. Adding tests would be great, but I don't think that needs to hold up this PR.

Issue #6862, opened 2022-05-03
Sprite color setter or Color equals not working as intended


Raxorg:
#### Issue details
Either the `Sprite` color setter or the `Color` equals yield unexpected results with color 0x0000003f.

#### Reproduction steps/code
```java
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.g2d.Sprite;

public class ColorTest {

    public static void main(String[] args) {
        Color color = new Color(0x0000003f);
        Sprite sprite = new Sprite();
        sprite.setColor(color);
        System.out.println(sprite.getColor().equals(color));
        // Yields false, getColor() returns 0x0000003e
    }
}
```
#### Version of libGDX
1.10.0

#### Affected platforms
- [x] All

crykn:
Fixed by #6863.

PR #6863, opened 2022-05-04
Fix Sprite's getColor() behavior

Raxorg:
As described in [issue 6862](https://github.com/libgdx/libgdx/issues/6862) `getColor()` returns a vertex's modified color instead of the `color` variable stored in the `Sprite`.

Even if the reason for the current implementation is to keep sync when the user modifies the vertices' colors directly, current `getColor()` implementation only takes `C1` into account.

Thanks to @tommyettinger, @mgsx-dev, @Frosty-J & @MobiDevelop for helping me understand key concepts

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
@@ -630,12 +630,6 @@ public class Sprite extends TextureRegion {
 	/** Returns the color of this sprite. If the returned instance is manipulated, {@link #setColor(Color)} must be called
 	 * afterward. */
 	public Color getColor () {
-		int intBits = NumberUtils.floatToIntColor(vertices[C1]);
-		Color color = this.color;
-		color.r = (intBits & 0xff) / 255f;
-		color.g = ((intBits >>> 8) & 0xff) / 255f;
-		color.b = ((intBits >>> 16) & 0xff) / 255f;
-		color.a = ((intBits >>> 24) & 0xff) / 255f;
 		return color;
 	}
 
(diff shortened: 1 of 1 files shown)


mgsx-dev (approved):

obigu:
This was changed on [aa0d30e2dadbe36d4b549aca3c99c3725333815c](https://github.com/libgdx/libgdx/commit/aa0d30e2dadbe36d4b549aca3c99c3725333815c).

Pinging @NathanSweet

gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
```diff
@@ -630,12 +629,6 @@ public float getScaleY () {
 	/** Returns the color of this sprite. If the returned instance is manipulated, {@link #setColor(Color)} must be called
 	 * afterward. */
 	public Color getColor () {
-		int intBits = NumberUtils.floatToIntColor(vertices[C1]);
-		Color color = this.color;
-		color.r = (intBits & 0xff) / 255f;
-		color.g = ((intBits >>> 8) & 0xff) / 255f;
-		color.b = ((intBits >>> 16) & 0xff) / 255f;
-		color.a = ((intBits >>> 24) & 0xff) / 255f;
```
tommyettinger, line 638: I definitely like this removal; the scaling NathanSweet was referring to happens here and in floatToIntColor().

NumberUtils.floatToIntColor() does have some oddities because it tries to create an 8-bit alpha from a 7-bit value. My preference when doing that is to copy the MSB of alpha into the previously-vacant LSB of alpha, like:
```java
    public static int floatToIntColor (float value) {
        int intBits = Float.floatToRawIntBits(value);
        intBits |= (intBits >>> 31) << 24;
        return intBits;
    }
```

This also avoids some float math.

tommyettinger (approved):
I like this a lot; if we track a Color object inside each Sprite, we should use it.

Issue #6866, opened 2022-05-13
Android loop sound stops when playing too many sounds


KristjanP94:
#### Issue details
On Android: When playing too many sounds in a small time frame, any loop sounds might be stopped.
StreamId of looping sound gets popped from [streamId-s list](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidSound.java#L49).

#### Reproduction steps/code
1. Play a loop sound
2. Play 8+ sounds in small time frame so that the sounds overlap

#### Possible fix
Probably polished and improved version of my extremely poor code example from years back: [AndroidSound.java pastebin](https://pastebin.com/Jqx4A4na)
I added separate handling of loop sound Id-s so they are exempt from being removed by size limit.
Since I couldn't manage to build android-backend library for new LibGDX version, I haven't created new version of this fix and am using workaround - I set priorities to different sounds and limit the number manually ensuring that high priority sounds get played while low priority sounds can be played only few times in a timeframe.

#### Version of libGDX and/or relevant dependencies
All versions (tested in 1.10.0)

#### Please select the affected platforms
- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

obigu:
This can probably be "fixed" by setting a higher priority to looped sounds to avoid SoundPool to stop those streamIds.

I don't think what you see is caused by the line you indicate as that internal libGDX IntArray with 8 elements is different per `AndroidSound` instance which means you would need to have 8 instances of that same sound to be played for it to be popped from the array and, in any case, it would not stop any playing sound.

What probably happens is simply that you reach the limit of sounds that `SoundPool` is able to play concurrently and, because all have the same priorities, it just starts by stopping the oldest one to free resources. Independently of a possible fix, it's always good practice to keep track on your game on when was the last time a sound was played to avoid playing it too often in short amounts of time filling the `SoundPool` capacity.

I'll make a PR with a proposed fix.

Frosty-J:
[...]

PR #6868, opened 2022-05-13
Increase priority of looping sounds

obigu:
Fixes https://github.com/libgdx/libgdx/issues/6866


Frosty-J (approved):
Simple dimple. Don't think it can go wrong. Use case is naturally to avoid looped sounds stopping and letting players hear the deafening silence.

MrStahlfelge (approved):
LGTM

PR #6879, opened 2022-05-23
fix SortedIntList when used with allocateIterators

mgsx-dev:
SortedIntList is used with DecalBatch and doesn't work when using `Collections.allocateIterators = true`.

I didn't investigate for all other collections but i suspect there are similar issues.

Feel free to commit to this PR if you find some other.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/SortedIntList.java
+++ b/gdx/src/com/badlogic/gdx/utils/SortedIntList.java
@@ -115,7 +115,7 @@ public class SortedIntList<E> implements Iterable<SortedIntList.Node<E>> {
 	 * Use the {@link Iterator} constructor for nested or multithreaded iteration. */
 	public java.util.Iterator<Node<E>> iterator () {
 		if (Collections.allocateIterators) return new Iterator();
-		if (iterator == null) iterator = new Iterator();
+		if (iterator == null) return iterator = new Iterator();
 		return iterator.reset();
 	}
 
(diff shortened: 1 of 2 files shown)


tommyettinger (dismissed):
This looks like `reset()` was just forgotten here, and hopefully only here. Either the constructor could call `reset()` (once an Iterator constructor is added), or we could rely on calling `iterator()`... The constructor is probably a little more robust, since some usage encourages using the constructor. I'm approving this now, but if anyone feels like moving reset() into the constructor, I would be happy with that too. (The call to `iterator()` would still need to reset the iterator if it didn't construct one.)

mgsx-dev:
Thanks! Indeed other classes call reset in their constructor, so i did it for consistency.

gdx/src/com/badlogic/gdx/utils/SortedIntList.java
```diff
@@ -123,6 +123,10 @@ public class Iterator implements java.util.Iterator<Node<E>> {
 		private Node<E> position;
 		private Node<E> previousPosition;
 
+		public Iterator () {
+			reset();
+		}
```
tommyettinger, line 128: Ah, yes, this is cleaner than before. There is something to note with this: `iterator()` now calls `reset()` twice if `allocateIterators` is false and the cached Iterator is null. I don't think it's a major issue, but it's a very easy fix in `iterator()` to just `return new Iterator();` in the second if block instead of creating one and then resetting it again.

tommyettinger (changes requested):
It's not a blocking request, since I think reset() takes nearly no time, but it would be best to just call reset() once when allocateIterators is false and the cached Iterator is null.

gdx/src/com/badlogic/gdx/utils/SortedIntList.java
mgsx-dev: no problemo

tommyettinger (approved):
Looks great now!

PR #6881, opened 2022-05-29
Bitcode is deprecated on XCode 14

Berstanio:
~~This is nothing really relevant, I just noticed it by accident.
The embedded bitcode is only relevant when publishing in the App Store.~~
Bitcode appears to be deprecated on XCode 14 and won't be accepted even on AppStore submissions.

### DIFF
--- a/backends/gdx-backend-robovm/build-objectal.sh
+++ b/backends/gdx-backend-robovm/build-objectal.sh
@@ -12,8 +12,8 @@ tar xvfz $BUILD_DIR/objectal.tar.gz -C $BUILD_DIR --strip-components 1
 
 XCODEPROJ=$BUILD_DIR/ObjectAL/ObjectAL.xcodeproj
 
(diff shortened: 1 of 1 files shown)


obigu:
Bitcode appears to be deprecated on XCode and won't be accepted even on AppStore submissions. Do we need to remove it from anywhere else?

https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes

```
Starting with Xcode 14, bitcode is no longer required for watchOS and tvOS applications, and the App Store no longer accepts bitcode submissions from Xcode 14.

Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with bitcode is deprecated. Please update your project and/or target settings to disable bitcode.” The capability to build with bitcode will be removed in a future Xcode release. IPAs that contain bitcode will have the bitcode stripped before being submitted to the App Store. Debug symbols for past bitcode submissions remain available for download. (86118779)
```

Berstanio:
Probably jnigen too 
https://github.com/libgdx/gdx-jnigen/blob/c405c677c8245bfcfb9087eaf4ec70ea34385b8d/gdx-jnigen/src/main/resources/com/badlogic/gdx/jnigen/resources/scripts/build-ios.xml.template#L252

obigu (approved):

PR #6883, opened 2022-06-01
Scene2D actor, add duplicate some method to use Vector2 instead of x, y

fabiitch:
Hi i add some methods in Actor.java to use Vector2 instead of float x, float y.
I do it because is boring to do setPosition(pos.x, pos.y)
 (often my Vector2 var has a big name so its more like 
`        image.setPosition(GameHudConstants.POS_ATTACK.x, GameHudConstants.POS_ATTACK.y);`

Methods duplicate : setPosition (Vector2 pos), setPosition (Vector2 pos, int alignment), moveBy (Vector2 move)

This commit add nothing new, it just for code comfort.

thanks for reading.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 [1.11.1]
 - iOS: Add new MobiVM MetalANGLE backend
 - Javadoc: Add "-use" flag to javadoc generation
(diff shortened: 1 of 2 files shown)


Frosty-J:
I like convenience methods, but one consideration is where we draw the line. For instance, `Sprite#setPosition` and `SpriteBatch#draw` don't have `Vector2` overloads (or possibly `Rectangle` would be suitable in the latter case).

Issue #4256, opened 2016-08-18
Sound panning should work for stereo files (and if not, add it to the docs)


CreamyCookie:
#### Issue details

Pan doesn't work on Stereo files. If this can't be supported for some reason, it should at least be mentioned in the documentation of `Sound`.
#### Reproduction steps/code

Code taken from #2560:

``` java
import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.Input;
import com.badlogic.gdx.InputAdapter;
import com.badlogic.gdx.audio.Sound;

public class LibGdxScratchpad extends ApplicationAdapter {

    private Sound sound;

    @Override
    public void create() {

        sound = Gdx.audio.newSound(Gdx.files.internal("test.wav"));

        Gdx.input.setInputProcessor(new InputAdapter() {
            @Override
            public boolean keyDown(int keyCode) {
                if (keyCode == Input.Keys.LEFT) {
                    sound.play(1.0f, 1.0f, -1.0f);
                } else if (keyCode == Input.Keys.RIGHT) {
                    sound.play(1.0f, 1.0f, 1.0f);
                }
                return true;
            }
        });
    }
}
```
#### Version of LibGDX and/or relevant dependencies

1.9.4-SNAPSHOT
#### Please select the affected platforms

Tested on Ubuntu 16.04 (Linux), likely all

CreamyCookie:
@badlogic Thanks! How hard would it be to make it work for stereo files?

badlogic:
We'd have to hack OpenAL soft to support spatial stereo mixing. Or split
stereo sounds up into two mono sources and manually manage both of their
positions.

PR #6885, opened 2022-06-03
Add more docs on stereo sound panning

soundasleep:
In #4256, docs on how panning only works for mono sounds was added to the `Sound#setPan()` method.

However for anyone trying to use `Sound#play()` with the panning argument, they won't see this note in the Javadocs.

I'd like to suggest adding the same note to these method docs!

### DIFF
--- a/gdx/src/com/badlogic/gdx/audio/Sound.java
+++ b/gdx/src/com/badlogic/gdx/audio/Sound.java
@@ -51,6 +51,7 @@ public interface Sound extends Disposable {
 	public long play (float volume);
 
 	/** Plays the sound. If the sound is already playing, it will be played again, concurrently.
+	 * Note that panning only works for mono sounds, not for stereo sounds!
 	 * @param volume the volume in the range [0,1]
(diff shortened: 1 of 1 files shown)


MrStahlfelge:
Is it true for all platforms?

soundasleep:
Not sure; I don't have access to a Mac or Linux machine at the moment, only Windows.

But I suspect it's a [limitation of OpenAL](https://www.openal.org/documentation/OpenAL_Programmers_Guide.pdf):

image

Frosty-J:
macOS and Linux behaviour is usually the same as Windows. MrStahlfelge will be getting at GWT, Android and iOS, the latter of which I'm not set up to test either.

crykn:
For future readers: @Hangman noted on Discord that panning for stereo sounds can be enabled by an [OpenAL extension](https://openal-soft.org/openal-extensions/SOFT_source_spatialize.txt) on desktop: 
```
AL10.alSourcei(sourceId, SOFTSourceSpatialize.AL_SOURCE_SPATIALIZE_SOFT, AL10.AL_TRUE);
```

PR #6892, opened 2022-06-09
Set immersive mode to true by default on Android

obigu:
Immersive mode is nowadays enabled on most games and should be default. Backwards compatibility wise should not be a big deal (for those who don't read CHANGES) as the difference is obvious.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.11.1]
+- [BREAKING CHANGE] Android: Immersive mode is now true by default. Set `useImmersiveMode` config to `false` to disable it. 
 - iOS: Add new MobiVM MetalANGLE backend
 - API Addition: Added Haptics API with 4 different Input#vibrate() methods with complete Android and iOS implementations.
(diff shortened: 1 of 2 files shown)


Frosty-J (approved):
I agree that immersive mode is usually wanted. I don't know if games normally use it, as it's not applicable to my device, but I expect they do to maximise screen estate and reduce the likelihood of users accidentally closing the app.

If someone is testing on a device that has physical buttons, it can be a surprise to them that the navigation buttons are on-screen for other devices. Well, it was a surprise to me, despite seeing them in the emulator! Obviously they can just set `immersiveMode` to `false` as you've noted in `CHANGES` if the buttons are wanted, such as when creating more of an application than an actual game.

Side note: From other people's libGDX games I've tested, they often don't handle the back button, which can be very frustrating if I try to close an in-game menu and end up closing the app instead, potentially losing progress in the process.

MrStahlfelge:
Immersive mode is needed to suppress gesture navigation so it is indeed a good idea.

PR #6898, opened 2022-06-13
Fix zoomed issue on MetalANGLE backend

Berstanio:
This fixes the issue, that when setting the "Zoomed" mode, the game was also wrongly zoomed and therefor rendered partially outside the screen.
This fix applies to both backends, but it is only relevant for MetalANGLE and doesn't do any harm to the GLKit view.


obigu (approved):
Tested on real device, works as expected. On MetalANGLE, Zoomed mode not only renders outside screen but input touch coordinates are messed up. Doesn't seem to affect GLKit.

PR #6903, opened 2022-06-18
add missing g3d primitive type

haz00:
The exception message says that the lineloop is a valid type, but it looks like they forgot to parse it.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/loader/G3dModelLoader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/loader/G3dModelLoader.java
@@ -134,6 +134,8 @@ public class G3dModelLoader extends ModelLoader<ModelLoader.ModelParameters> {
 			return GL20.GL_TRIANGLE_STRIP;
 		} else if (type.equals("LINE_STRIP")) {
 			return GL20.GL_LINE_STRIP;
+		} else if (type.equals("LINE_LOOP")) {
+			return GL20.GL_LINE_LOOP;
 		} else {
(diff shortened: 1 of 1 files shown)


mgsx-dev:
apparently it's not handled by fbx-conv, see https://github.com/libgdx/fbx-conv/blob/032ec0829983f2569d636c5f2c36ab9966972172/src/modeldata/Serialization.cpp#L32

Did you managed to generate a G3D file with line loop in it?

If not i suggest to fix the exception message instead.

haz00:
I was just experimenting with [blender g3d exporter](
https://github.com/haz00/blender-g3d-exporter/blob/master/g3d_exporter/export_operator.py#L162) and wanted it to create line loops for Wire objects. But I get errors on the libgdx parser side if g3d file contains such enum. So the pull request should fix that.

mgsx-dev:
the thing is, it's not part of the official specification, see https://github.com/libgdx/fbx-conv/wiki/Version-0.3-%28draft%29

and it looks like this Blender plugin doesn't export line_loop primitive neither, see: https://github.com/haz00/blender-g3d-exporter/blob/master/g3d_exporter/builder.py#L699 so i don't know how you end up with this primitive type in your file.

haz00:
The plugin does not export this enum because it is not handled by the parser, so it makes no sense at the moment.

I see the spec, but I dont think this thing can break anything.

SimonIT:
Given that you're the author of the plugin you just could add the export :)
We could also alter the spec and add it to fbx-conv.
I think it it's a nice addition and doesn't hurt to add as it is non-breaking.

PR #6905, opened 2022-06-19
Fix iOS Keyboard type not changing immediately

Berstanio:
Currently the keyboard type does not change immediately when running "setOnscreenKeyboardVisible". This can be seen by running the "OnscreenKeyboardTest" test.
This pr fixes the issue.


obigu (approved):
Tested on real device (iPhone 11)

MrStahlfelge (approved):
LGTM

Issue #6729, opened 2021-12-12
LiveWallpaper crashing on Samsung Galaxy S21 series devices with Android 12


raichy:
#### Issue details
LiveWallpaper crashing on Samsung Galaxy S21 series devices with Android 12
p.s. works perfectly on API32 emulator (Android Studio)

#### Reproduction steps/code
Compile simple LiveWallpaper and run on Samsung Galaxy S21 (Android 12) real device. Also crashing on Samsung Remote Test Lab's devices.

#### Version of libGDX and/or relevant dependencies
libGDX version: 1.9.7 and 1.10.0 

#### Stacktrace

(stacktrace from 1.9.7, same goes for 1.10.0 but different line codes)

java.lang.NullPointerException: 
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaperService$AndroidWallpaperEngine.notifySurfaceChanged (AndroidLiveWallpaperService.java:336)
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaperService$AndroidWallpaperEngine.onSurfaceChanged (AndroidLiveWallpaperService.java:310)
  at android.service.wallpaper.WallpaperService$Engine.updateSurface (WallpaperService.java:1446)
  at android.service.wallpaper.WallpaperService$Engine.setTouchEventsEnabled (WallpaperService.java:676)
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaperService.initialize (AndroidLiveWallpaperService.java:147)
  at com.test.lwp.MainActivity.onCreateApplication (MainActivity.java:102)
  at com.badlogic.gdx.backends.android.AndroidLiveWallpaperService$AndroidWallpaperEngine.onSurfaceCreated (AndroidLiveWallpaperService.java:267)
  at android.service.wallpaper.WallpaperService$Engine.updateSurface (WallpaperService.java:1410)
  at android.service.wallpaper.WallpaperService$Engine.attach (WallpaperService.java:1613)
  at android.service.wallpaper.WallpaperService$IWallpaperEngineWrapper.executeMessage (WallpaperService.java:2512)
  at com.android.internal.os.HandlerCaller$MyHandler.handleMessage (HandlerCaller.java:44)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8582)
[...]

PR #6906, opened 2022-06-19
AndroidDaydream input null check

cypherdare:
I've had numerous crash reports on Galaxy S21 and S22 family devices with an NPE in `AndroidDaydream.onConfigurationChanged()`. My best guess of what's happening is that these devices are calling `onConfigurationChanged()` before `onAttachedToWindow()`, so `input` is still null.

This change should be safe because
1. If `onConfigurationChanged()` is called before `onAttachedToWindow()`, the keyboard state will be updated in the user's `initialize()` call.
2. It's a screen-saver. We probably don't care if the keyboard is attached anyway.

I don't have one of the offending devices to test on, but I also don't know how many of you actually use AndroidDaydream. If I don't get any feedback in a couple of weeks, I plan to go ahead and pull this.


Frosty-J:
Worth a shot. Best-case scenario this fixes it; worst it does nothing. Like you, I don't have a problematic device to test on. I don't know if you're aware that live wallpapers also seem screwed on them (#6729).

obigu:
@cypherdare Have you been able to confirm this fix works?

tommyettinger (approved):
This has taken an awfully long time to get reviewed for a simple null check being added. This is a sane practice in general that the codebase should be doing anywhere we might call a method on null. Approved.

PR #6914, opened 2022-06-28
[scene2d] Fire exit events just before an actor is removed.

### LINKED ISSUE #6910: mouseOverActor and pointerOverActors are not cleared in Stage#clear()
I would assume that when I call [`Stage#clear()`](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L591-L595) or [`Stage#dispose()`](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L841-L845), that all references to `Actor`s in the `Stage` are cleared. 

However, that is not the case. If the fields [`private final Actor[] pointerOverActors = new Actor[20];`](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L75) or [`private @Null Actor mouseOverActor;`](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L80) contain a reference to an `Actor`, those persist until the next call of [`Stage#act(float)`](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L193-L227).

I think `Stage#clear()` should set these fields to `null`.

---

Context: I had the problem that I called `Stage.dispose`, assuming that its children would get garbage-collected. However, that didn't happen as long as I kept a reference to the `Stage` itself. Only when I let go of the reference to the stage (or called `act()` one last time) did the memory the children used actually become available.

NathanSweet:
closes #6910

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
@@ -74,8 +74,7 @@ public class Stage extends InputAdapter implements Disposable {
 	private final Vector2 tempCoords = new Vector2();
 	private final Actor[] pointerOverActors = new Actor[20];
 	private final boolean[] pointerTouched = new boolean[20];
-	private final int[] pointerScreenX = new int[20];
-	private final int[] pointerScreenY = new int[20];
+	private final int[] pointerScreenX = new int[20], pointerScreenY = new int[20];
 	private int mouseScreenX, mouseScreenY;
 	private @Null Actor mouseOverActor;
 	private @Null Actor keyboardFocus, scrollFocus;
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
```diff
@@ -194,34 +193,20 @@ public void act (float delta) {
...
-					Pools.free(event);
-				}
-				continue;
+			if (pointerTouched[pointer]) {
+				// Update the over actor for the pointer.
+				pointerOverActors[pointer] = fireEnterAndExit(overLast, pointerScreenX[pointer], pointerScreenY[pointer], pointer);
+			} else if (overLast != null) {
+				// The pointer is gone, exit the over actor for the pointer, if any.
+				pointerOverActors[pointer] = null;
+				fireExit(overLast, pointerScreenX[pointer], pointerScreenY[pointer], pointer);
```
tommyettinger, line 202: This looks a lot cleaner than before. I see fireExit() was moved to its own method and is also used in actorRemoved(); could it be it called anywhere else that might be able to also use that code?

tommyettinger (approved):
This looks like it both fixes a leak and cleans up the code in Group, so that's great. I think I would wait until a nightly has built so interested people can test it quickly. My only concern is that without testing, I don't know if we can be sure the leak is fully plugged up. I'm not too concerned though, since we only had found one main culprit and it seems to be good now, so I'm approving.

gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
NathanSweet: Those are the only two places firing an `exit` event. I considered a private method to obtain and configure an event, as that is done in many places, but the code is probably a little easier to read repeating the few lines of code.

PR #6915, opened 2022-06-29
disable depth mask when rendering transparent decals

mgsx-dev:
When transparent decals collides, some transparent pixels of one could cull others because they all read from and write to depth buffer.

Since they are already sorted (back to front), there is no reason to write into the depth buffer.

While fixing th issue, i noticed that the same thing was already done in SimpleOrthoGroupStrategy, so i think it's a simple missing or an edge case in CameraGroupStrategy.

I can't see a use case where this change would be a problem. I think it'll fix a lot more cases than breaking anything.


yuripourre:
Looks good to me and could also improve performance a bit.

yuripourre (approved):

Frosty-J:
Tested this change with my game and it didn't change anything, though I don't have intersecting decals so that's to be expected. Since it didn't break my beloved decal, I'm okay with it.

PR #6919, opened 2022-07-04
Remove fallback on Android Haptics vibrate(VibrationType)

obigu:
I've been able to run some tests on some Android devices to see how consistent the new haptics/vibration API is and I've found the fallback for Android O devices for the `vibrate(VibrationType)` call is not good enough, there's no feeling of impact whatsoever.

The main problem on Android ecosystem is that AFAIK there's no way to know for sure if the device has proper haptics support or a simple vibrator with amplitude control capabilities which makes all the difference. 

I think the best approach to assume haptics support is to require both amplitude control capaibilities and Android version to be at least Q because in that case the device must implement the Haptics API (https://source.android.com/devices/input/haptics/haptics-implement) and we ensure `vibrate(VibrationType)` will work (even if there's a certain risk it doesn't feel as good as it should on some specific device).


MrStahlfelge (approved):
Good for me

PR #6920, opened 2022-07-07
[Gradle] Replace usage of 'tasks.whenTaskAdded()' to 'tasks.matching()' in Android project

leni8ec:
Replace usage of `tasks.whenTaskAdded()` to `tasks.matching()` in Android project for comply with Gradle guidelines of ["Task Configurations Avoidance"](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview)

This is implemented in order to avoid initializing ("lazy" configuration phase) all other independent tasks when starting any Gradle task. 

---
#### Performance of `gradle help` task

Before: [gradle performance link](https://scans.gradle.com/s/tlql5gfkwvgkq/performance/configuration#summary-tasks-created-during-configuration)
After: [gradle performance link](https://scans.gradle.com/s/lmnzq2xswo2jg/performance/configuration#summary-tasks-created-during-configuration)

image

### DIFF
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/android/build.gradle
@@ -59,10 +59,8 @@ task copyAndroidNatives {
     }
 }
 
-tasks.whenTaskAdded { packageTask ->
-    if (packageTask.name.contains("merge") && packageTask.name.contains("JniLibFolders")) {
-        packageTask.dependsOn 'copyAndroidNatives'
-    }
+tasks.matching { it.name.contains("merge") && it.name.contains("JniLibFolders") }.configureEach { packageTask ->
+    packageTask.dependsOn 'copyAndroidNatives'
 }
 
 task run(type: Exec) {


tommyettinger:
This does seem to speed up Android builds at least a little. After the initial build had been run and the project cleaned (so that metadata would already exist, hopefully making the comparison fair), the old code took just over 10 seconds, and this version took just over 8 seconds. That was on a trivially small project running `assembleDebug`; the amount of time it takes to build a larger project I'm sure is affected more strongly. I've merged this into gdx-liftoff, but there's no stable release with it yet (you can see the update linked at the commit just before this comment).

tommyettinger (approved):
This seems to work in projects gdx-liftoff generates; I created and ran an APK without issues. That isn't to say I tested this exhaustively, but it does build more quickly and it seems to produce the same output. Approved.

leni8ec:
Thank you!

It really doesn't make any difference exactly when building the APK, as it takes a negligible amount of time in this case.

But when performing custom tasks, which are done quickly, such as "packing textures into atlas" - it is noticeable. That's why I gave the "help" task as an example, as a benchmark.

Because in the solution, which was in libGDX, the gradle was supposed to initialize ALL the tasks at once, when performing the CURRENT operation (task)! 

And my fix - allows to initialize, only DEPENDENT tasks, required for execution of the CURRENT operation (task). Besides, it is recommended in the documentation of Gradle itself (I gave links in the first post)

leni8ec:
And one more example:

If I create a custom task, which is poorly optimized and has an initialization time of about 2-5 seconds - then this delay is added to absolutely every task in our gradle project. 
Because it will have to be initialized, while it is not required for current task.

Actually this is what pushed me to find this solution.

obigu:
LGTM, thanks!

PR #6922, opened 2022-07-11
Fix startup zoom on MetalANGLE

Berstanio:
This would be a temporary fix until https://github.com/kakashidinho/metalangle/pull/84 is merged and released.

I have tested the same code with MOE and there it worked well, but I wasn't able to test it on robovm. RoboVM didn't wanted to launch, I don't know why (java.lang.RuntimeException: No devices connected).

Documentation: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreAnimation_guide/ReactingtoLayerChanges/ReactingtoLayerChanges.html


Berstanio:
It's not working :(

Berstanio:
I close this in favor of just upgrading the MetalANGLE version

CoderBaron:
@Berstanio thanks for the fix in the latest MetalANGLE version. 
I applied it to my game and the zoom is away, however I still have black frames after the launch screen (LaunchScreen.storyboard) gets shown. 

Do you also have this issue or is it a problem on my side?
Just to clarify: The Apple launch screen is shown, then the screen fades to black and after a few frames my libGDX launch screen (identical with Apple) is shown. With the OpenGL backend the transition was seamless (no black frames).

Berstanio:
I don't have the black frames you mentioned @CoderBaron .
Is it a fading transition, like something gets animated that shouldn't?

Berstanio:
But you are also not the first person mentioning the issue. If I remember correctly @fgnm had the same problem, right?

CoderBaron:
Yeah it‘s fading but I guess that is because I‘m using a storyboard launchscreen instead of pngs? 

(I changed because Apple‘s deprecating the static pngs and it‘s space efficient and easier to have just one png image scaling to aspect fit for all screen resolutions.)

I think the problem is that the launchscreen is fading to the MGLKViewcontroller too early and there is nothing drawn yet. The console logs even say `MGLKViewcontroller resume` before libGDX is finished setting up.

fgnm:
Yes I have the same issue, but I didn't test yet the #6924 .. Can be merged so anyone can test it easily? :)

Berstanio:
> Yes I have the same issue, but I didn't test yet the #6924 .. Can be merged so anyone can test it easily? :)

@obigu What do you think?

Also, is one of you able to provide something like a minimal reproducer for me @fgnm @CoderBaron ? And maybe lets move the discussion to #6924

obigu:
Merged, won't be able to de much testing for a while though.

CoderBaron:
@Berstanio I've created a new issue for this: #6971

PR #6927, opened 2022-07-14
Update to Gradle 7.5

Frosty-J:
This seems to be okay, and should bring us Java 18 support. The warning spam about explicit/implicit task dependencies was also present with 7.3.3.

I won't do anything to gdx-liftoff as I'm far too busy. Proof:

Morbius in Morrisons


tommyettinger:
Nah, I got gdx-liftoff covered. https://github.com/tommyettinger/gdx-liftoff/releases/tag/v1.11.0.1 I've been waiting for Morbin' Time for a while now.

tommyettinger (approved):
These changes seem to work fine in gdx-liftoff. Anything broken in Gradle 7.5 should only take another 3-4 months for them to fix, so it's probably fine to update?

Frosty-J:
Before releasing libGDX 1.11.1, we should check if Gradle 7.5.1, 7.5.2 etc. has released. That'll fix them bugs.

tommyettinger:
I have a feeling I'll get yelled at if I merge this without anybody else approving, but right now the nightly build seems to be broken (see GH Actions) and maybe this would help get the Gradle wrapper working again... @Frosty-J, if you run the `gradlew wrapper` task in your libGDX fork, does it change any files that would be committed?

tommyettinger:
https://github.com/libgdx/libgdx/commit/7a11de5274424486d63a586fe5b69bd336501b5a is the most recent commit and it failed because of some Gradle wrapper validation thing. Builds since then may have succeeded but I don't know if they'd affect a nightly build, because most seem to be checks for PRs. I just ran a test in SquidLib, which now uses Gradle 7.5, using Java 8 and Java 18; both worked fine. I haven't checked Android or GWT yet.

Tom-Ski:
We should add this action to the PRs also then right?

tommyettinger:
It looks like it was added by GitHub! https://github.com/libgdx/libgdx/runs/7345118747?check_suite_focus=true

Tom-Ski:
https://github.com/libgdx/libgdx/blob/master/.github/workflows/gradle-wrapper-validation.yml its ours no? But it should be running on PRs also, so why is this PR (and others, not failing) https://github.com/libgdx/libgdx/runs/7345118747?check_suite_focus=true

Frosty-J:
Maybe some glitch in the system, since they're the same checksums both in this PR (at the time of discussion) and Nate's commit. Might as well update the wrapper anyway - it's only 60K. It adds and wants `gradle/dependencies.gradle` and `gradle/dist.gradle`... I don't know why. Actual libGDX projects don't involve this. I'll commit and let you see what you think.

Frosty-J:
> has anyone checked for the lowest java sdks for issues?

As of Gradle 5, it no longer runs on Java 7, but 8 should still be great.

tommyettinger (approved):
I've gotten similar changes when running `gradlew wrapper`; it seems totally normal. The update to 7.5 is still good.

Tom-Ski (approved):

PR #6928, opened 2022-07-15
Remove copyIdeaResources tasks

obigu:
When we upgraded to Gradle 7 (https://github.com/libgdx/libgdx/pull/6811) an issue was introduced that caused Gradle not to apply optimizations to the execution making builds much slower than they were.

There's this warnings spammed all over the place
```
Execution optimizations have been disabled for task ':tests:gdx-tests-gwt:javadoc' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: 'XXX\XXX\libgdx\gdx\build\classes\java\main'. Reason: Task ':tests:gdx-tests-gwt:javadoc' uses this output of task ':gdx:copyIdeaResources' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.5/userguide/validation_problems.html#implicit_dependency for more details about this problem.
 ```

The `copyIdeaResources` tasks was necessary to run tests from IDEA as it wasn't copying resources found in `gdx/res` to any classpath output folder which caused some tests to crash when unable to find the resource. If I remember correctly, this was an issue only on some backends such as Lwjgl3.

I've run tests  without the task on IDEA 2022.1.3 and the resources are copied to `/build/resources` and available on classpath. I've made sure I `Clean Project` (and run `clean`Gradle task as well) before testing and it seems to work. **I have no idea why it works now so, before approving or merging this please test it yourself to make sure I haven't made a mistake.**

### DIFF
--- a/extensions/gdx-lwjgl3-angle/build.gradle
+++ b/extensions/gdx-lwjgl3-angle/build.gradle
@@ -25,10 +25,3 @@ sourceSets.main.java.srcDirs = ["src"]
 sourceSets.main.resources.srcDirs = ["res"]
 
 tasks.withType(Jar) { duplicatesStrategy = DuplicatesStrategy.INCLUDE }
-
-// Workaround needed for IDEA to have resources on classpath when running tests (like gdx-tests-lwjgl3)
-task copyIdeaResources(type: Copy) {
-	from "${projectDir}/res"
-	into "${buildDir}/classes/java/main/"
-}
-processResources.dependsOn copyIdeaResources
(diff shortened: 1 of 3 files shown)


Tom-Ski:
I think that was if you run without using Gradle as the build system in the IDE?

obigu:
Yeah, true, I rememeber now! I've tested and it still fails to find resources if setting IDEA as the tool to build and run (Settings -> Build, Execution, Deployment -> Gradle -> Build and run using).

I think it's not a big deal to require IDEA to require using Gradle as build system on IDEA if you want to run tests (we can add it to the wiki). For the time being I would merge this to improve build times and get rid of all the ugly warnings.

Tom-Ski:
I think this is fine too. There are workarounds that you can do with idea to make sure those resources are processed with the other build mode, which I do a lot in projects to speed things up, so I'm fine with this.

obigu:
I'm a bit confused about something though. How a Gradle task fixes running on IDEA if it doesn't use Gradle to run?

I've now tested on Master using IDEA as "build and run" tool and after cleaning everything and I've reproduced the issue (for example run Lwjgl3 test and CullTest). Isn't it already broken independently of this change?

PR #6931, opened 2022-07-19
fix javadoc for AnimationController

mgsx-dev:
zero loopCount doesn't play animation (which make sense in term of API) and from repo history it looks like it always been like that. Also, 1 is the default for methods without loop count parameter.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/AnimationController.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/AnimationController.java
@@ -225,7 +225,7 @@ public class AnimationController extends BaseAnimationController {
 
 	/** Set the active animation, replacing any current animation.
 	 * @param id The ID of the {@link Animation} within the {@link ModelInstance}.
-	 * @param loopCount The number of times to loop the animation, zero to play the animation only once, negative to continuously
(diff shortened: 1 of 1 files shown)


MobiDevelop:
Huh, unfortunate naming on that parameter. Passing 0 to loop 0 times is obviously more natural than passing 1 to loop 0 times. I'd say playCount is a better name, except then the negative value to play continuously doesn't feel natural.

mgsx-dev:
I guess that "loopCount" could be interpreted as "looping count"  or "number of loops to play".  Considering you're playing a looped animation, it feel more naturals to me to define how many times i want it to be played. Also, in the code, you either compute the number of loops to play, or hard code it eg. `animator.animate("myAnim", 3, ...)`, which is suggesting (my interpretation) the animation will be played 3 times (and not 4).

In the javadoc we have : 
> The number of times to loop the animation, 1 to play the animation only once, negative to continuously loop the animation.

To disambiguate, it could be replaced by : 
> The number of times to play the animation, 1 to play the animation only once, negative to continuously loop the animation.

What do you think?

MobiDevelop:
I like it.

MobiDevelop (approved):

PR #6941, opened 2022-07-26
Update to Android Gradle Plugin 7.2.1 and API level 32

Frosty-J:
IntelliJ IDEA 2022.2 is now out, which means it's time for AGP 7.2! This should help with #5976. I chose to go from 7.0 to 7.2 because IDEA skipped 7.1 so we might as well too. I promise it won't be the same ordeal as #6811.

While we could go for target API 33, sources haven't been released yet and Android 13 is still in beta, so for most developers I think it's best to stick with 32 (Android 12L) for now. I've been targeting 32 for months - it's not much different from 31 and there should be no issues with it on the IDEA front anymore.

Build tools 33.0.0 is intentional - Android Studio recommends 32.1.0-rc1 or newer but I don't want to use a release candidate, so 33.0.0 was chosen as it gets along with API 32 just fine.

If libGDX 1.11.1 is releasing soon, I'm more than happy to postpone these (or similar) changes until 1.11.2 since Android/Gradle has already been shaken up quite a lot for 1.11.1, Google has yet to set a deadline for updating beyond API 31 and we should give everyone a moment to update their IDEs.

P.S. Cheers for #6936 - I think that fixed what I broke in #6927. Now, what will this one break? Hopefully nothing - libGDX projects and the tests seem to work okay.


tommyettinger:
I know nothing about this stuff, but I'll try to duplicate the changes here in gdx-liftoff when they're considered safe to merge into libGDX proper. I also think people should have a month or so to update to the latest IDEA/AS.

Frosty-J:
Sounds like a plan. Android Studio Chipmunk released 9 May, but IntelliJ IDEA 2022.2 only released today!

obigu:
I've come across some issues with RoboVM Gradle plugin 
```
"Unable to find method ''void org.apache.commons.compress.archivers.zip.ZipFile.<init>(java.nio.channels.SeekableByteChannel)''"](https://stackoverflow.com/questions/70340427/unable-to-find-method-void-org-apache-commons-compress-archivers-zip-zipfile)
```
We may need changes wither on the gdxSetup template or RoboVM Gradle plugin before merging this. Will investigate some more.

Frosty-J:
Is this something I could reproduce without access to a Mac? The `mobivm` tasks unsurprisingly die trying to run `security`. You could try updating from RoboVM 2.3.16 to 2.3.18 - I don't know what has changed between these versions.

obigu:
I'm unable to reproduce the issue consistently and after some cleaning and rebuilding it seems to disappear so it's probably nothing. I'm updating my projects, I'll update if it pops back.

obigu (approved):

PR #6945, opened 2022-07-30
Added GLES31 and GLES32 support with Lwjgl3 backend implementation

mgsx-dev:
This PR adds the base for GLES 3.1 and 3.2, and also adds implementation for Lwjgl3 backend.

Notes about other backends : 
* Lwjgl2 : i don't think we need to provide an implementation since it is supposed to disappear.
* GWT : we would need first to implement GLES 3.0 using WebGL2.
* iOS : not applicable.
* Android : I don't have the hardware but implementation should be straightforward if someone is interested.

Notes about what i did and choices i made : 
* I copied interfaces from Android sources (link to the original file included in the interface code)
* I kept Android copyright notice like it was done for GL30 interface.
* I commented out methods variant using arrays (like GL30). Those would require to copy to a buffer for the lwjgl3 backend.
* I commented out some unnecessary buffer size parameters for known types (IntBuffer, FloatBuffer). It wasn't done for the GL30 implementation but those parameters are not used in desktop backends which doesn't looks right to me. I don't want to blame anyone, but i didn't want to adopt this choice.
* I didn't found OpenGL equivalent for `glPrimitiveBoundingBox` and i don't know if it could be achieved with `glClipPlane`. I didn't investigate any further. I guess we could simply omit it for now.
* 2 methods for debug message controls has been omitted because they are related to GLSync object which we also omitted in GL30.

Note about tests : 
* I only included few tests in order to validate the implementation. I have some other tests in another repo that i need to cleanup before adding them. They will be useful to validate other backend implementations in the future (mainly Android).
* I don't plan to cover every single obscur function, but at least providing a test/example for important features.
* We now have 2 more options when running desktop tests : `--gl31` and `--gl32` in order to enable related tests.
* I tested GWT and Lwjgl2 backend to check there is no regression or impacts (all clear).

Breaking or not breaking ?
* i think the only breaking change is the new `Graphics` getter/setters which doesn't hurt so much (default interface methods would have helped with that).
* If you think it needs to be flagged as a breaking change, i'll update the CHANGES file accordingly.

What next : 
* Once this PR get merged, i could create some other PRs to add wrappers (eg. ComputeShader) and changes in utility classes in order to add some useful features (eg. Frame buffer multisample).
* I could resurrect my old PR about Geometry and Tesselation shaders (#4963). I wasn't sure how to integrate that at this time, i think i'm more confortable with GLES/OpenGL intrication in libgdx now.

Reminder : 
* The [desktop test runner](https://libgdx.com/dev/from-source/#tests) documentation have to be modified once this PR is merged (new command line options)

I guess it'll take some time to review this PR, so i plan to work on the test suite to add them in this PR while taking into account the review at the same time.

Thanks for reading.

Changes not totally related to this PR : 
* I fixed GLProfiler : with Lwjgl3 backend, we can have several windows and Gdx.glXX references was only updated by `window.makeCurrent`. So when we enable/disable the profiler, those references wasn't up to date until the next frame update.


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3GL31.java
```diff
@@ -0,0 +1,388 @@
...
+
+	@Override
+	public void glDrawArraysIndirect (int mode, long indirect) {
+		GL40.glDrawArraysIndirect(mode, indirect);
+	}
+
+	@Override
+	public void glDrawElementsIndirect (int mode, int type, long indirect) {
+		GL40.glDrawElementsIndirect(mode, type, indirect);
+		;
```
intrigus, line 56:
```suggestion
		GL40.glDrawElementsIndirect(mode, type, indirect);
```

mgsx-dev: done! :-)

mgsx-dev:
i think i'm done for the test suite, the PR is ready for reviews.

Tom-Ski (dismissed):

JamesTKhan (approved):
Ran through each test and reviewed the code. Looks good on my end. This is an exciting change, thank you.

PR #6946, opened 2022-08-02
Small fixes for gdx.gwt.xml .

tommyettinger:
In the .gwt.xml file that defines which files are usable sources for GWT, ObjectLongMap was missing as a source, and PauseableThread was spelled incorrectly. Before, if you tried to use ObjectLongMap in conjunction with reflection on GWT, it would fail because the compiler didn't know ObjectLongMap existed. I don't know anything about PauseableThread, other than the name of the class has an 'e' in the middle and the source entry previously didn't. It might not make sense as being usable on GWT, but it was originally an "include" and not an "exclude," so I fixed the spelling only.

### DIFF
--- a/gdx/res/com/badlogic/gdx.gwt.xml
+++ b/gdx/res/com/badlogic/gdx.gwt.xml
@@ -437,11 +437,12 @@
 		<include name="utils/NumberUtils.java"/> <!-- Emulated -->
 		<include name="utils/ObjectFloatMap.java"/>
 		<include name="utils/ObjectIntMap.java"/>
+		<include name="utils/ObjectLongMap.java"/>
 		<include name="utils/ObjectMap.java"/>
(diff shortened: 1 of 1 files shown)


Tom-Ski (approved):

SimonIT:
image
It seems not everything in PauseableThread is supported on gwt

tommyettinger:
> image It seems not everything in PauseableThread is supported on gwt. I'm not even using PauseableThread. I'm wondering why it wasn't catched by the CI? By searching the log it seems `gradlew build` seem not to run `gdx-tests-gwt` despite it beeing in the `settings.gradle` included. It fails on the `compileGwt` task which doesn't run here

Ha, so the typo was what allowed it to compile all this time... I guess it should never have been marked as `include` when it was initially added to gdx.gwt.xml .

PR #6954, opened 2022-08-08
Gradle 7.5.1

Frosty-J:
I've let https://github.com/tommyettinger/gdx-liftoff/commit/6582f6f0badf7ae1ecc660e51a2ec90c22678251 sit for a bit without catastrophe, so we can probably apply the same to gdx-setup. I don't think 7.5.1 vs 7.5 really makes a lick of difference to libGDX users, but might as well. Next Gradle release looks likely to be 7.6.

* [Release notes](https://docs.gradle.org/7.5.1/release-notes.html)

Budget Lyze:

Budget Lyze


tommyettinger (approved):
There doesn't seem to be a 7.5.2 on the way any time soon, though I could easily be wrong. 7.5.1 has been stable for me, I use it everywhere. The one issue I have encountered is that running `gradlew wrapper` while 7.5.1 is in use can reset the Gradle version to 7.5, so we may need to be aware of that.

PR #6957, opened 2022-08-11
Bring back getGlfwKeyCode

Hangman:
By request on [Discord](https://discord.com/channels/348229412858101762/873537733430226964/1006919687541633045) this PR re-adds the getGlfwKeyCode function. It was removed in [6351](https://github.com/libgdx/libgdx/pull/6351).

I have fixed the mentioned inconsistencies, though a real 1:1 mapping wasn't possible.

- **GLFW.GLFW_KEY_WORLD_1** and **GLFW.GLFW_KEY_WORLD_2** both map to **Input.Keys.UNKNOWN**, doesn't work the other way around
- there's no equivalent for **GLFW.GLFW_KEY_F25** in libGDX
- **GLFW.GLFW_KEY_LEFT_SUPER** and **GLFW.GLFW_KEY_RIGHT_SUPER** both map to **Input.Keys.SYM**, so I had to decide.. it's now **Input.Keys.SYM** => **GLFW.GLFW_KEY_LEFT_SUPER**
- default returns **GLFW.GLFW_KEY_UNKNOWN** (-1) instead of 0 like in the old function

Incentive for discussion: getGdxKeyCode isn't static at the moment.


crykn (approved):

MrStahlfelge:
Can you outline the use case here in short?

tommyettinger:
For context, the original Discord message before a PR was discussed:

Discord screenshot

I'd recommend clicking the "Discord" in Hangman's original post, since it goes to the conversation in question (assuming the reader has a Discord account they can use).

Personally I cannot follow all of the differences between keyboard layouts and LWJGL versions and so forth; it would be really nice to have some kind of consistent option for keybindings.

MrStahlfelge:
The consistent option was the method I introduced.
It is non-static on purpose so that it can be overridden in client code if needed, like we did in other backends before. This seems to be one of these cases here, so I don't see why we should bring back the type of code that was in my opinion a real improvement to get rid of.

Or maybe I am missing something.

crykn:
When #6351 changed the way key mapping works, we also removed (the supposedly no longer needed) `getGlfwKeyCode(int)` and with that any way to convert libGDX's keycodes to GLFW's keycodes. While we no longer need that method in `DefaultLwjgl3Input` itself, it is still needed to display the correct key names, because GLFW maps keys to their physical locations (e.g., pressing 'A' on AZERTY returns 'Q'). IIRC @PokeMMO was the first one to notice this.

PokeMMO:
Yeah, you need to map back to GLFW before calling [glfwGetKeyName](https://github.com/LWJGL/lwjgl3/blob/28cccbad117d2e7c7f550d330f32bbea1838fbee/modules/lwjgl/glfw/src/generated/java/org/lwjgl/glfw/GLFW.java#L3615-L3675) to get the proper key names.

MrStahlfelge:
The right way to get key names is the `Input.Keys.toString` method, I think we all agree on that. Why can't it be used?

The problem in introducing the removed method back are apparent; besides the maintenance problem, it does not really work because we don't have and won't ever have a 1:1 mapping. I might not see something, but in my opinion this means that you either work with the platform-agnostic keycodes and then you can use Input.Keys.toString to get key names. Or, you really want to use glfw keys and then conversion to platform-agnostic and back is plainly wrong, instead the Input implementation should be subclassed in client code.

PokeMMO:
>Why can't it be used?

Because it does not work for alternative keyboard layouts. With an AZERTY layout when you press "A" glfw sends libgdx `GLFW_KEY_Q`. The only way to map is back to key letter instead of logical position is to use `glfwGetKeyName`.

MrStahlfelge:
I see; this means Input.Keys.toString is wrong and should be fixed.

mgsx-dev:
i think the root of the issue is [here](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java#L364). If we want (and i guess we do) have consistent key mapping across all backends, we need to find a way to convert the GLFW physical keys to logical keys. I dig into that long time ago but didn't found a solution.

MrStahlfelge:
True; as often in such cases, GWT/JavaScript behaviour might give some ideas. IIRC on GWT/JS for a key there are both information available: keycode ('Q' in example) and character ('A' in example). This is IMO a good way to handle this.

MrStahlfelge:
Yes; my suggestion was to make the Key information like it is on JS, having both information available. Our current keyCode corresponds to event.code on JS: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
We are missing event.key on JS: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

This could be added in a clean and platform-agnostic way.

Hangman:
> I see; this means Input.Keys.toString is wrong and should be fixed.

> If we want (and i guess we do) have consistent key mapping across all backends

> Yes; my suggestion was to make the Key information like it is on JS, having both information available.

I agree with all of these points. Since the implementation would be completely different, I'll close this now and open an issue instead, linking to this conversation.

PR #6958, opened 2022-08-11
Align.java

Ryuu-64:
1. Add a private constructor to reinforce the utility class.
2. Rename filed in toString method StringBuilder buffer -> StringBuilder builder
    the name buffer might be ambiguous, cause there is StringBuffer in java.


Ryuu-64:
Spotless java check fails.
I'll recreate a new pull request.

PR #6970, opened 2022-08-15
`charter encoding` → `character encoding`

Frosty-J:
It's very difficult to debug why some characters don't work right in properties files when the documentation goes on about charters.

Anyway, the issue in my case is properties are ISO-8859-1 (at least for Java 8 and older) but libGDX defaults to UTF-8. As far as I understand, GWT will always want UTF-8, so to fix this [IDEA/AS' default can be changed](https://www.jetbrains.com/help/idea/properties-files.html#encoding). Setting the encoding at the bottom-right of the IDE is disabled for properties.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/I18NBundle.java
+++ b/gdx/src/com/badlogic/gdx/utils/I18NBundle.java
@@ -135,7 +135,7 @@ public class I18NBundle {
 	/** Creates a new bundle using the specified <code>baseFileHandle</code> and <code>encoding</code>; the default locale is used.
 	 * 
 	 * @param baseFileHandle the file handle to the base of the bundle
-	 * @param encoding the charter encoding
+	 * @param encoding the character encoding
 	 * @return a bundle for the given base file handle and locale
 	 * @exception NullPointerException if <code>baseFileHandle</code> or <code>encoding</code> is <code>null</code>
 	 * @exception MissingResourceException if no bundle for the specified base file handle can be found */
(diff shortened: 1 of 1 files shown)


PokeMMO:
Android will (almost?) always use UTF-8 by default.

Java18+ will use UTF-8 by default. https://openjdk.org/jeps/400

For everyone else, it's dependent on operating system defaults unless you specify it yourself. (`-Dfile.encoding=UTF-8`)

tommyettinger (approved):
Let's chart a course for Encoding Island!

Frosty-J:
I'm aware things have been moving to UTF-8 for a while. I think it may be due to libGDX's compatibility with Java 7 that IDEA enforces ISO-8859-1 by default? Either that or it's a bone-headed decision applied to everything - I haven't experimented.

> In Java SE 9, properties files are loaded in UTF-8 encoding. In previous releases, ISO-8859-1 encoding was used for loading property resource bundles. UTF-8 is a much more convenient way to represent non-Latin characters.
> 
> Most existing properties files should not be affected: UTF-8 and ISO-8859-1 have the same encoding for ASCII characters, and human-readable non-ASCII ISO-8859-1 encoding is not valid UTF-8. If an invalid UTF-8 byte sequence is detected, the Java runtime automatically rereads the file in ISO-8859-1.

https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm#GUID-974CF488-23E8-4963-A322-82006A7A14C7

(yes, I know Java properties aren't hugely applicable to libGDX properties, but the IDE doesn't know any better)

PR #6972, opened 2022-08-16
Optimize first render time of the RoboVM backend. Fixes #6971

### LINKED ISSUE #6971: iOS black frames after launch screen
#### Issue details
I'm using a storyboard launch screen in my app that then should transition into a Scene2D launch screen which looks the same but with animation. There is almost a second of black screen after the storyboard launch screen disappears and my Scene2D launch screen appears, which is not acceptable (Apple review does reject such glitches).
By using static images for the iOS launch screen the issue goes away with the RoboVM backend.

1. Static images, robovm backend: seamless transition
2. Static images, metalangle backend: 1 frame of half rendered half black screen on simulator
3. Storyboard, robovm backend: several black frames
4. Storyboard, metalangle backend: about 1 second of black screen on simulator

#### Probable cause
By design the iOS splash screen fades out once the method didFinishLaunching() of the app delegate returned. At this time the next view controller needs to be setup already. 
With libGDX this means that the OpenGL/Metal view needs to have started drawing its content **before** didFinishLaunching is complete.

Currently the way the iOS backend is setup makes this impossible. The ApplicationListener.create() method as well as the completion of the OpenGL setup are called **after** didFinishLaunching().

1.11.1-SNAPSHOT

- [ ] Android
- [X] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

CoderBaron:
This PR changes the iOS backend to have completely setup OpenGL and call `ApplicationListener.create()` within the `UIApplicationDelegate.didFinishLaunching()` method. Before this PR it was lazily done on the first draw call from `GLKView`.

Once `didFinishLaunching()` returns iOS starts a fading transition from the launch screen to the root view controller. Without this change the root view controller has no contents yet and will therefore display black frames. Please see the following gifs for comparison on the default libGDX app. It is more noticeable if you have more resources to load for your first screen.

[Before](https://media.giphy.com/media/ySB9B0QY9r9MUL12db/giphy.gif)
[After](https://media.giphy.com/media/2h5zI1BY87NQL4fxYF/giphy.gif)

One important note: With this change all code inside `ApplicationListener.create()` and `resize()` will block the return of `didFinishLaunching()`. Apple has a time limit of a few seconds before the code must have executed otherwise it will terminate the application. So developers should only load necessary resources for the first app screen and offload any heavy stuff for later (like `AssetManager.update()` inside the render method). This recommendation is not new but blocking the render thread for 10 seconds after `didFinishLaunching()` will at least not crash the app.

Fixes #6971

### DIFF
--- a/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
+++ b/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
@@ -185,6 +185,16 @@ public class IOSGraphics extends AbstractGraphics {
 		lastFrameTime = System.nanoTime();
 		framesStart = lastFrameTime;
 
+		// enable OpenGL
+		makeCurrent();
+		// OpenGL glViewport() function expects backbuffer coordinates instead of logical coordinates
+		gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);
+
+		String versionString = gl20.glGetString(GL20.GL_VERSION);
+		String vendorString = gl20.glGetString(GL20.GL_VENDOR);
+		String rendererString = gl20.glGetString(GL20.GL_RENDERER);
+		glVersion = new GLVersion(Application.ApplicationType.iOS, versionString, vendorString, rendererString);
+
 		appPaused = false;
 	}
 
@@ -215,28 +225,12 @@ public class IOSGraphics extends AbstractGraphics {
 		app.listener.pause();
 	}
 
-	boolean created = false;
-
 	public void draw (GLKView view, CGRect rect) {
 		makeCurrent();
 		// massive hack, GLKView resets the viewport on each draw call, so IOSGLES20
 		// stores the last known viewport and we reset it here...
 		gl20.glViewport(IOSGLES20.x, IOSGLES20.y, IOSGLES20.width, IOSGLES20.height);
 
-		if (!created) {
-			// OpenGL glViewport() function expects backbuffer coordinates instead of logical coordinates
-			gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);
-
-			String versionString = gl20.glGetString(GL20.GL_VERSION);
-			String vendorString = gl20.glGetString(GL20.GL_VENDOR);
-			String rendererString = gl20.glGetString(GL20.GL_RENDERER);
-			glVersion = new GLVersion(Application.ApplicationType.iOS, versionString, vendorString, rendererString);
-
-			updateSafeInsets();
-			app.listener.create();
-			app.listener.resize(getWidth(), getHeight());
-			created = true;
-		}
 		if (appPaused) {
 			return;
 		}
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - [BREAKING CHANGE] Android: Immersive mode is now true by default. Set `useImmersiveMode` config to `false` to disable it. 
 - [BREAKING CHANGE] iOS: `hideHomeIndicator` set to `false` by default (was `true`).
 - [BREAKING CHANGE] iOS: `screenEdgesDeferringSystemGestures` set to `UIRectEdge.All` by default (was `UIRectEdge.None`).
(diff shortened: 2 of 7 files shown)


Berstanio (approved):
Works fine on my test. The previos handling also seems a bit imperfect, blocking the render thread with app initialization code is maybe not the best idea. 
"Apple has a time limit of a few seconds before the code must have executed otherwise it will terminate the application." 
On my test it took 20sec until the app got killed, which is in my opinion acceptable. But maybe it should be noted in the changes anyway.

CoderBaron:
Yeah I read somewhere it‘s about 10 seconds but seems to be higher now. 
Anyways I can add a line to CHANGES. Is this considered a breaking change?

Berstanio:
I don't know, but in case of doubt I think it is better to mark it as breaking. But I'm also not a maintainer, so what do I know!
Maybe it is also worth noting, that ApplicationListener#create/resize is now called a bit earlier during the startup? But I don't think that this will lead to any errors, maybe only if some user relies on some strange specific custom callbacks/stuff initialization. I don't know, with my testing I haven't found anything relevant.

CoderBaron:
CHANGES updated

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -185,6 +185,16 @@ public void draw (CGRect rect) {
...
 
+		// enable OpenGL
+		makeCurrent();
+		// OpenGL glViewport() function expects backbuffer coordinates instead of logical coordinates
+		gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);
+
+		String versionString = gl20.glGetString(GL20.GL_VERSION);
+		String vendorString = gl20.glGetString(GL20.GL_VENDOR);
+		String rendererString = gl20.glGetString(GL20.GL_RENDERER);
+		glVersion = new GLVersion(Application.ApplicationType.iOS, versionString, vendorString, rendererString);
```
obigu, line 196: Shouldn't this just be called once on creation? Maybe we can make a call from `didFinishLaunching`.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
```diff
@@ -185,6 +185,16 @@ public void draw (CGRect rect) {
 		lastFrameTime = System.nanoTime();
 		framesStart = lastFrameTime;
 
+		// enable OpenGL
+		makeCurrent();
+		// OpenGL glViewport() function expects backbuffer coordinates instead of logical coordinates
+		gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);
```
obigu, line 191: Even if it's probably the same, I would leave the old `gl20.glViewport(IOSGLES20.x, IOSGLES20.y, IOSGLES20.width, IOSGLES20.height);` here (including the "hack" comments) so that's easy to see the only thing we're doing is moving the logic inside the `if (!created)`.

obigu (commented):
Just a few comments from looking the code, I'll test it as soon as I can.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Berstanio: Wasn't it moved directly? In the `if (!created)` block it was `gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);` too

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGraphics.java
Berstanio: Wasn't it moved directly? In the `if (!created)` block it was `gl20.glViewport(0, 0, screenBounds.backBufferWidth, screenBounds.backBufferHeight);` too

obigu: Oh, sorry, checked it too fast and the comparison view tricked me.

obigu:
Tested on real device and found no issues. Change makes sense. Thanks!

PR #6973, opened 2022-08-19
Add API to (auto) switch output device on lwjgl3

Berstanio:
This pr adds support for three main things (only for lwjgl3):
1. Auto switching the Output device if e.g. a headphone gets disconnected. This is mainly relevant on windows, because before that the audio just died.
2. It reevaluates the default device, if a new device gets connected, and therefor maybe switches to it (the os decides)
3. It adds a API to specifically select a output device. This API supports swapping too, so if the device gets disconnected and reconected, it will be choosen again.

I also added a Test for this.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,8 @@
 - iOS: Add new MobiVM MetalANGLE backend
 - iOS: Update to MobiVM 2.3.19
 - Update to LWJGL 3.3.2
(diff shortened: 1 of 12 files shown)


Berstanio:
This is now ready to get reviewed, since libGDX got updated to lwjgl3 3.3.2.

SimonIT (approved):
Works from what I tested:
- Manually setting the device
- Disconnecting a device changes the device
- Re-connecting changes the device back

HydrolienF (approved):
It's working as intended for me (Windows 11, bluetooth earphones)

PR #6978, opened 2022-09-01
Desktop AudioDevice#getLatency() now returns values in samples and not milliseconds

chrislo27:
The `AudioDevice` documentation states that the `getLatency()` method returns the latency in samples. However, the desktop implementations were returning them in milliseconds.

These changes only affect how the latency value is computed, and not the actual `AudioDevice` latency itself. The units are now in samples which is accurate to the `AudioDevice` interface spec.


imlakshay08 (approved):

crykn (approved):
LGTM.

PR #6979, opened 2022-09-03
MathUtils trigonometry improvements

tommyettinger:
This includes a bug fix and some additions, as well as minor doc fixes in 3 files. The bug fix is for MathUtils.atanUnchecked() and MathUtils.atan(), which affects MathUtils.atan2(); when given a point on the positive x-axis, they should return 0f, but didn't before. The additions are degree versions of asin, acos, atan, and atan2, plus the new atan2Deg360(), which won't return a negative degree value (its range is 0-360). The behavior of Math.atan2() where it can produce negative results often makes me do extra work with a modulus or conditional to get it back to a positive range, so it might be nice to have a version of atan2() that doesn't require that. I'm totally fine if the degree methods are removed from this commit, I just thought they helped round out the degree-based code in MathUtils. The typos seem to be mostly copy-paste-related: PixmapPacker called itself ImagePacker, an untypeable ligature got into the name of a distance field paper in DistanceFieldGenerator, and I had copied some docs from another library that still referred to libGDX as an external library (in MathUtils).

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -395,6 +395,7 @@ jobs:
 
   build-and-upload-runnables:
     runs-on: ubuntu-18.04
+    needs: pack-natives
     env:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
@@ -412,8 +413,10 @@ jobs:
           java-version: '11'
 
       - name: Build Runnables
-        run: ./gradlew clean fetchNatives buildRunnables build
-
+        run: |
+          ./gradlew clean fetchNatives
+          ./gradlew buildRunnables build
+          
       - name: Upload artifacts to S3
         if: env.AWS_ACCESS_KEY_ID != null
         run: |
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@
 - Added Stage#actorRemoved to fire exit events just before an actor is removed.
 - ScrollPane#setScrollingDisabled avoids invalidate() if nothing changed.
 - Fixed incorrect ScrollPane#scrollTo.
+- API Fix: The cursor can now be catched on Android.
 
 [1.11.0]
 - [BREAKING CHANGE] iOS: Increased min supported iOS version to 9.0. Update your Info.plist file if necessary.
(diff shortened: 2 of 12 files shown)


obigu (approved):
Looks good to me. Any chance we can add some tests to `MathUtilsTest`?

tommyettinger:
Good idea. I also just wrote a tan() approximation, and I should test that too. Not sure if I should add tan() to the PR.

obigu:
> Good idea. I also just wrote a tan() approximation, and I should test that too. Not sure if I should add tan() to the PR.
> 
> EDIT: The reason a separate tan() could be needed is because with the resolution our sine table has, just doing `sin(x)/cos(x)` has severely aliased output (limited to specific multiples).

Yeah, I think it's worth adding it if it's also faster than `Math.tan` and is more precise than sin/cos.

tommyettinger:
I spent a decent part of the day writing and refining a test just for tan(), and I haven't even added it to the PR yet. It's hard to measure the error for tan() because `tan(MathUtils.PI * 0.5f)` should either be infinite or very close to infinite by the mathematical definition (and depending on floating-point imprecision), so any results that should be approaching infinity but aren't approaching it quickly enough wind up off by millions (out of a correct result in the billions). I think normally when people are calling a method like tan(), they worry more about small absolute deviations from correctness where the relative deviation is large, than large absolute deviations where the relative deviation is small. That is, getting 2 when you expected 1 is extremely bad, but getting 1000002 when you expected 1000001 isn't bad at all.

imlakshay08 (approved):

tommyettinger:
I ran some benchmarks and the new tan() is slower than the old tan(), but not by much on recent Java, and both are quite a bit faster than Math.tan().

Measuring tan() on Java 18:
```
Benchmark                       Mode  Cnt   Score   Error  Units
MathBenchmark.measureDivideTan  avgt    5   5.217 ± 0.566  ns/op
MathBenchmark.measureMathTan    avgt    5  17.659 ± 1.176  ns/op
MathBenchmark.measureSoontsTan  avgt    5   7.824 ± 0.646  ns/op
```
And on Java 8:
```
Benchmark                       Mode  Cnt   Score   Error  Units
MathBenchmark.measureDivideTan  avgt    5   4.605 ± 0.219  ns/op
MathBenchmark.measureMathTan    avgt    5  51.494 ± 2.440  ns/op
MathBenchmark.measureSoontsTan  avgt    5  13.219 ± 0.858  ns/op
```

If the higher-speed DivideTan seems better, it might not after you see the graph of a [statistical distribution](https://en.wikipedia.org/wiki/Cauchy_distribution) using old tan() vs. new:

Old:

Old

New:

New

Changing the implementation of tan() is the only change here; both also used double precision and the old one still encountered the strange "flat lines with hairs."

obigu (approved):
Looks good to me!

imlakshay08 (approved):

NathanSweet:
Sorry I'm late, had a literal hurricane disaster to deal with.

It looks great to me. Super thorough, as always!

PR #6980, opened 2022-09-04
Remix unmatched audio channels instead of dropping them

### LINKED ISSUE #6631: Output to mono audio devices doesn't work correctly
@Hangman was having this issue with a USB headset on Discord (the part about stereo files not playing - he didn't test further). I _expect_ headsets that use a 3.5mm jack would be unaffected.

We're both on Linux - I don't know if Windows or macOS may also be affected. Both LWJGL2 and LWJGL3 are affected, but in different ways. More on that in the next section.

## Issue details

When using monaural audio output, things don't work right! Regardless of `Sound` or `Music` or file format, it seems (not fully tested - it would be a lot of combinations).

Below I have the waveforms of some recordings with a linear view. GitHub doesn't let me embed audio into the page and it's easier to see what's what with images anyway, as only volume (and quality when it inevitably clips) is affected as far as I can hear.

### Audacious - Mono/Stereo files

VLC, Chrome and Firefox also sound the same. This sample is to help show there's badly nothing wrong on my end.

audacious mono

### LWJGL2 - Mono/Stereo files

Way too loud. Stereo gets correctly downmixed to mono.

lwjgl2 mono

#### Mono files

Mono files are a little too loud.

lwjgl3 mono

#### Stereo files

Stereo files don't play at all! Well, `getPosition()` suggests they're playing, but they're silent!

lwjgl3 stereo

## Reproduction steps/code

Play some sound. I've been using the repository at https://github.com/denisk20/test-music-position though can't guarantee it'll stick around as it was made for a different issue. Here's an example for how to make some noise:

```java
Music music = Gdx.audio.newMusic("Gdx.files.internal("myMusic.mp3"));
music.play();
```
I didn't have anything like Hangman that shows up as a mono device on the computer, but I was able to reproduce the issue using [these instructions](https://www.mbeckler.org/blog/?p=472) to force monaural output:

> 1. Use `pacmd list-sinks | grep name:` to find the name of the output you want to use. Mine was `alsa_output.pci-0000_00_1b.0.analog-stereo` (strip away the `<` and `>`).
> 2. Use that string in this command: `pacmd load-module module-remap-sink sink_name=mono master=alsa_output.pci-0000_00_1b.0.analog-stereo channels=2 channel_map=mono,mono`
> 3. Use the GUI sound settings to select the new “Remapped Built-in Audio Analog Stereo”.

## Version of libGDX and/or relevant dependencies
1.10.0

## Stacktrace
N/A - No errors are produced.

## Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [X] Linux
- [ ] MacOS

Frosty-J:
This seems to fix the LWJGL3 part of #6631 (I don't care about LWJGL2 anymore). With this change, playing stereo audio on a mono output device now results in a downmixed result instead of complete and utter silence. Panning gets ignored when outputting to mono. Results can be unideal if the output device changes while the game is running (e.g. a stereo one to a mono one) but properly handling device changes is out of scope for this pull request.

It also brings benefit for #6792 in my test. Now, instead of only the left and right channels being played from the quadraphonic file, I also get a little of the rear channels, like in VLC (video players are pretty good for testing, since they have to correctly mix movie audio).

`AL_TRUE` is equivalent to `AL_DROP_UNMATCHED_SOFT`. So with the old behaviour, I guess it didn't see a direct mapping from L+R to mono so just dropped them, resulting in silence. A couple quotes from Hangman, who without I'd be unaware of this:

> Since 1.11.0 we got the bindings to the SOFTDirectChannelsRemix extension which provides AL_REMIX_UNMATCHED_SOFT.

> It's niche but I always found using direct channels without the extension quite a risky decision. It's also a problem for someone who got a 5.1 sound system. Without virtualization (AL_FALSE) or the remix extension (REMIX_UNMATCHED), all those fancy speakers besides front_left and front_right will be silent.


crykn (approved):
LGTM! I've discussed the same fix in the past in the initial [PR](https://github.com/libgdx/libgdx/pull/6527#issuecomment-881899695) that changed this behaviour.

imlakshay08 (approved):

PR #6981, opened 2022-09-05
Recreate window

aretecorp:
### What has been changed
Lwjgl3: Added `Lwjgl3RecreateWindowTest` to gdx-tests-lwjgl3
Lwjgl3: Added `recreateWindow()` to `Lwjgl3Application`; used to apply changes to settings that require a new window to be created
API Addition: `Lwjgl3Graphics` now has `setMSAASamples()`, `setColorBits()`, `setDepthBits()`, `setStencilBits()`, `setBackBufferConfig()` (these use the aforementioned `recreateWindow()` method) & `setIdleFPS()`
POSSIBLE [BREAKING CHANGE] Lwjgl3: `recreateWindow()` instantiates new `Gdx.input` & `Gdx.graphics`, which can be a problem for already existing pointers

### What does the test do?
Lets the user change the MSAA samples value at runtime, while you'd normally be able to set this option before starting the `ApplicationListener`.

### Why was this done?
It was done because most commercial games let the user change the amount of anti aliasing without asking the user to restart the game.

### What platforms has this change been tested on?
LWJGL3 on Windows - I tried to make this work with LWJGL2 but I didn't make it.
I haven't even touched HTML5 and Android / iOS, mostly because this I don't think this feature is really necessary on these platforms. That's why I didn't add the new methods in `Lwjgl3Graphics` to the `Graphics` interface; they'd do nothing for 4 platforms out of 5 (counting the LWJGL2 backend).

### Notes
If you were to make the `recreateWindow(Lwjgl3ApplicationConfiguration)` method public you could change absolutely anything about the new window, and while that was the intent originally, I figured out pretty quickly it would cause more issues than what it's worth. 
For example, if one were to recreate the window with `config.disableAudio=false` while it previously would have been `true`, all `Sound` & `Music` objects created before the new window instance would be `MockSound` & `MockMusic`, respectively - therefore they wouldn't play despite `Gdx.audio` not being `MockAudio` and the setting being `false`; I would have to have an event in the `ApplicationListener` to inform the programmer of the window change, and that seemed a little too much. 
This exact same problem manifests itself in a different way with `Gdx.input` & `Gdx.graphics` - if I were polling for input with a pointer to `Gdx.input` I obtained before the window change, then after said change the data I'd get would be stuck to however it was before the new window.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -77,6 +77,7 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	private static GLVersion glVersion;
 	private static Callback glDebugCallback;
 	private final Sync sync;
+	private boolean recreateWindow = false;
 
 	static void initializeGlfw () {
 		if (errorCallback == null) {
@@ -378,7 +379,7 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 			return preferences.get(name);
 		} else {
 			Preferences prefs = new Lwjgl3Preferences(
-					new Lwjgl3FileHandle(new File(config.preferencesDirectory, name), config.preferencesFileType));
+				new Lwjgl3FileHandle(new File(config.preferencesDirectory, name), config.preferencesFileType));
 			preferences.put(name, prefs);
 			return prefs;
 		}
@@ -418,7 +419,7 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	@Override
 	public Lwjgl3Audio createAudio (Lwjgl3ApplicationConfiguration config) {
 		return new OpenALLwjgl3Audio(config.audioDeviceSimultaneousSources, config.audioDeviceBufferCount,
-				config.audioDeviceBufferSize);
+			config.audioDeviceBufferSize);
 	}
 
 	@Override
@@ -441,18 +442,25 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 		return createWindow(appConfig, listener, windows.get(0).getWindowHandle(), true);
 	}
 
-	/**
-	 * Creates a new {@link Lwjgl3Window} using the already existing {@link Lwjgl3ApplicationConfiguration}
-	 * and {@link ApplicationListener} from the current window and replaces it. Mostly used when
-	 * some settings require the window to be created from scratch, such as MSAA sampling. */
-	public Lwjgl3Window recreateWindow() {
-		return recreateWindow(config);
+	/** Creates a new {@link Lwjgl3Window} using the already existing {@link Lwjgl3ApplicationConfiguration} and
+	 * {@link ApplicationListener} from the current window and replaces it. Mostly used when some settings require the window to be
+	 * created from scratch, such as MSAA sampling. */
+	public void recreateWindow () {
+		recreateWindow(config);
 	}
 
-	private Lwjgl3Window recreateWindow(Lwjgl3ApplicationConfiguration appConfig) {
-		//Some settings don't work, such as disableAudio, which would require the
-		//Gdx.audio object to be recreated. Changing window size this way is also not recommended
-		//as it can lead to really odd values being passed into the ApplicationListener.resize() method.
+	private void recreateWindow (Lwjgl3ApplicationConfiguration appConfig) {
+		// The createWindow() method adds the window in the array in postRunnable();
+		// because it's not instant we need this boolean to stop someone from calling
+		// recreateWindow() X amount of times, thus creating X windows.
+		if (recreateWindow) {
+			return;
+		}
+		recreateWindow = true;
+
+		// Some settings don't work, such as disableAudio, which would require the
+		// Gdx.audio object to be recreated. Changing window size this way is also not recommended
+		// as it can lead to really odd values being passed into the ApplicationListener.resize() method.
 		config.set(appConfig);
 
 		ApplicationListener listener = currentWindow.getListener();
@@ -462,15 +470,16 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 		// the window may not be created immediately, meaning getInput() would return null and crash everything
 		postRunnable(new Runnable() {
 			@Override
-			public void run() {
+			public void run () {
 				recreatedWindow.getInput().setInputProcessor(inputProcessor);
+				// Now we can recreate the window again
+				recreateWindow = false;
 			}
 		});
-		return recreatedWindow;
 	}
 
 	private Lwjgl3Window createWindow (final Lwjgl3ApplicationConfiguration config, ApplicationListener listener,
-									   final long sharedContext, final boolean callCreateOnListener) {
+		final long sharedContext, final boolean callCreateOnListener) {
 		final Lwjgl3Window window = new Lwjgl3Window(listener, config, this);
 		if (sharedContext == 0) {
 			// the main window is created immediately
@@ -487,14 +496,15 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 		return window;
 	}
 
-	void createWindow (Lwjgl3Window window, Lwjgl3ApplicationConfiguration config, long sharedContext, boolean callCreateOnListener) {
+	void createWindow (Lwjgl3Window window, Lwjgl3ApplicationConfiguration config, long sharedContext,
+		boolean callCreateOnListener) {
 		long windowHandle = createGlfwWindow(config, sharedContext);
 		window.create(windowHandle, callCreateOnListener);
 		window.setVisible(config.initialVisible);
 
 		for (int i = 0; i < 2; i++) {
 			GL11.glClearColor(config.initialBackgroundColor.r, config.initialBackgroundColor.g, config.initialBackgroundColor.b,
-					config.initialBackgroundColor.a);
+				config.initialBackgroundColor.a);
 			GL11.glClear(GL11.GL_COLOR_BUFFER_BIT);
 			GLFW.glfwSwapBuffers(windowHandle);
 		}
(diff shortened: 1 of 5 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
```diff
@@ -534,6 +534,36 @@ public void setSystemCursor (SystemCursor systemCursor) {
...
+
+	public void setDepthBits(int depth) {
+		Lwjgl3ApplicationConfiguration config = getWindow().getConfig();
+		setBackBufferConfig(config.r, config.g, config.b, config.a, depth, config.stencil, config.samples);
+	}
+
+	public void setStencilBits(int stencil) {
+		Lwjgl3ApplicationConfiguration config = getWindow().getConfig();
+		setBackBufferConfig(config.r, config.g, config.b, config.a, config.depth, stencil, config.samples);
+	}
```
tommyettinger, line 559: I have to say, I don't like the way this is structured. Each call to set some part of the back buffer config recreates the window, potentially happening more than once if something makes multiple calls to set different parts. The main way the configuration is currently used is via `setBackBufferConfig()`, so I don't know why we really need new methods that only partially set that config. I know these respect the current settings for config that isn't being changed, but if the docs are clear on the defaults, and developers save the things they want saved, I think it might avoid any potentially strange behavior from the graphics getting repeatedly created if the only way to change the back buffer config is by `setBackBufferConfig()`. If there aren't any possible issues from repeatedly creating the graphics, just ignore this.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
```diff
@@ -437,35 +438,63 @@ public Lwjgl3Window newWindow (ApplicationListener listener, Lwjgl3WindowConfigu
...
+	 * some settings require the window to be created from scratch, such as MSAA sampling. */
+	public Lwjgl3Window recreateWindow() {
+		return recreateWindow(config);
+	}
+
+	private Lwjgl3Window recreateWindow(Lwjgl3ApplicationConfiguration appConfig) {
+		//Some settings don't work, such as disableAudio, which would require the
+		//Gdx.audio object to be recreated. Changing window size this way is also not recommended
+		//as it can lead to really odd values being passed into the ApplicationListener.resize() method.
+		config.set(appConfig);
```
tommyettinger, line 456: If some settings don't work, that really, really needs to be tested and documented. Tested because we need to know what settings could be broken, and documented so users are aware that behavior is different for recreated windows.

tommyettinger (changes requested):
I'm not really sure about this PR. Adding proper documentation is a must; we can't expect users to source-dive to find out that disableAudio doesn't work anymore. We really need to know what configuration works, doesn't work as expected, or doesn't work at all when a window is recreated.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
aretecorp: >we can't expect users to source-dive to find out that disableAudio doesn't work anymore. We really need to know what configuration works, doesn't work as expected, or doesn't work at all when a window is recreated.

I should've been clearer with what I meant when explaining how some settings don't work. I brought up the example of `config.disableAudio=true` not working because that requires the `Gdx.audio` object to be reinstated and it really wasn't worth the hassle. But there's no way for the user to change that setting, the only place where `recreateWindow()` is used is in `Lwjgl3Graphics`, which uses the same `Lwjgl3ApplicationConfiguration` the programmer creates in the `DesktopLauncher` - there's no risk of anything other than the back buffer config settings getting changed.


The fact that some settings get kinda wonky when recreating the window is exactly why the main `recreateWindow()` function is private: to stop the programmer from modifying settings he really shouldn't be able to. I talked to others and the consensus was that there really is no reason for the user to be able to modify settings such as `maxNetThreads` or `audioDeviceSimultaneousSources`, but MSAA samples does make sense. If the user wants to change the window's size, he should absolutely not recreate the window, because that leads to odd results with the `width` and `height` values passes in the `ApplicationListener.resize()` method. But again, there's no possible way the programmer can change this outside of `DesktopLauncher`.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
aretecorp: >The main way the configuration is currently used is via setBackBufferConfig(), so I don't know why we really need new methods that only partially set that config.

I did it like it is now because it felt more natural to me & more in line with the rest of `Lwjgl3Graphics` but of course that can be changed - it's just how I structure code.

>If there aren't any possible issues from repeatedly creating the graphics, just ignore this.

I haven't noticed any strange behavior from repeatedly creating the window, other than the game slowing down to a crawl tbh

aretecorp: >I haven't noticed any strange behavior from repeatedly creating the window, other than the game slowing down to a crawl. It does look pretty bad though - I'll add something to Lwjgl3RecreateWindowTest later to test for this case.

I brainfarted, I take it back. It absolutely creates a problem. I'm on it.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
```diff
@@ -76,6 +77,7 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	private static GLVersion glVersion;
 	private static Callback glDebugCallback;
 	private final Sync sync;
+	private boolean recreateWindow = false;
```
crykn, line 80: I think this could use a more descriptive name like `currentlyRecreatingWindow`.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
```diff
@@ -53,6 +53,10 @@ public class Lwjgl3Window implements Disposable {
 	boolean iconified = false;
 	boolean focused = false;
 	private boolean requestRendering = false;
+	// these are used to stop create() & dispose() from getting called
+	// on the ApplicationListener in the event the window is merely being recreated.
+	private boolean callCreateOnListener = true;
+	private boolean callDisposeOnListener = true;
```
crykn, line 59: Maybe `notifyDisposeListener` would be better?

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
```diff
@@ -534,6 +534,16 @@ public void setSystemCursor (SystemCursor systemCursor) {
 		Lwjgl3Cursor.setSystemCursor(getWindow().getWindowHandle(), systemCursor);
 	}
 
+	public void setIdleFPS (int idleFPS) {
+		getWindow().getConfig().idleFPS = idleFPS;
```
crykn, line 538: Are you sure that does anything? Lwjgl3Application seems to use the inital configuration and not the one provided by the windows: https://github.com/libgdx/libgdx/blob/916663ed3d4ebd71c3bc83e6c8d3483c3aa37612/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L241

mgsx-dev:
I'm not sure this PR is a good idea for many reasons :

What you want is to change MSAA at runtime, and i suspect "most commercial games" to use multisample FBO, which you can easily recreate in your game with different settings including MSAA without messing with the window.

GLFW MSAA settings and many others like depth buffer precision are simply hints and may or may not be used depending on several things (OS window manager, GPU driver, etc). Some users reported issues about very low precision depth buffer (almost unusable) and their solution was to render into a FBO where the depth precision can be perfectly controlled. So there might be some issues with MSAA as well. 

Also, you need a GL context to know what MSAA values are acceptable (MAX_SAMPLES) and you can't know that before creating the window, so the first time you create a window, you just pray for the setting to be applyable, which sounds to me almost unusable. It's just a hint after all.

Finally, i see a lot of drawback with this window recreation : 
* creating a new window with shared context of a window previously closes sounds dangerous, i don't think we have guarantee that it will always work. (i didn't find any info on that in the GLFW doc).
* some references to Graphics will be invalid, that's the case for GLProfiler, so users have to care about re-creating some other things.
* probably some other thing i didn't see yet, like backend specific code in a game that hold some references to the window or Graphics, and maybe other stuff.

Of course, the multisample FBO solution requires GLES 3.1 (in core OpenGL since 3.0) so for desktop, i don't think that's an issue. However we'd need to have #6945 merged priori to implementing it. I already have an implementation for that ready to PR and i have a working POC that allows to change MSAA at runtime flowlessy, which does exactly what you want. Another advantage is that it's not backend specific, so it should also work with Lwjgl2 and Android backends, and also HTML backend (when we'll support WebGL 2).

I don't want to ruine all your work and the reviews, i just wanted to inform you guys that there may be some more elegant solution for the original need.

Frosty-J:
Changing antialiasing at runtime would be my use for this. I might also use it to toggle the transparent framebuffer option - I have yet to test if having that set to `true` when rendering stuff with an opaque background has any drawbacks (performance, most likely). Changing the preferences directory at runtime could be useful for some, but it's not something I'm interested in and that has nothing to do with the rendering window itself.

So if we can get antialiased FBOs, I'd just do that instead. In fact, I'd really like antialiasing in FBOs regardless of changing settings at runtime, since they're just so handy. On systems where OpenGL 3.0 isn't available, I'd just forego antialiasing, though I'd only want to do so in the case of hardware limitations rather than software (if WebGL 2.0 for libGDX never arrives).

aretecorp:
> I don't want to ruine all your work and the reviews, i just wanted to inform you guys that there may be some more elegant solution for the original need.

Don't worry, it's understandable. I was thinking that perhaps some form of anti-aliasing could be implemented through this shader I found

https://www.geeks3d.com/20110405/fxaa-fast-approximate-anti-aliasing-demo-glsl-opengl-test-radeon-geforce/3/

If I understood correctly it's also really easy to implement. I could probably subclass the `FrameBuffer` for this.

I'm going to close the PR

mgsx-dev:
@aretecorp we already have that for libgdx (FXAA) works like a charm, see https://github.com/crashinvaders/gdx-vfx

Frosty-J:
Just remembered that toggling vsync and/or frame rate limits at runtime could be useful. For example, a 3D game's benchmark could run with an uncapped frame rate, then actual gameplay could be capped.

PR #6993, opened 2022-09-28
Make libGDX buildable from source again with XCode 14

Berstanio:
https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes
"Building iOS projects with deployment targets for the armv7, armv7s, and i386 architectures is no longer supported. (92831716)"

This leads to build failures when trying to build libGDX from source. This pr fixes these in a first attempt.
The if - else blocks in the roboVM config block are not optimal. It would be best if we would have xcframework support in roboVM, or atleast in jnigen, to handle these cases better.

### DIFF
--- a/backends/gdx-backend-robovm/build-objectal.sh
+++ b/backends/gdx-backend-robovm/build-objectal.sh
@@ -3,6 +3,9 @@ set -e
 BASE=$(cd $(dirname $0); pwd -P)
 
 BUILD_DIR=$BASE/target/objectal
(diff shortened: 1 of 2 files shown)


Tom-Ski:
This should fail on building all the other natives no? Extensions are compiling armv7 still afaik

Berstanio:
> This should fail on building all the other natives no? Extensions are compiling armv7 still afaik

Nope, it only fails on XCode projects. If XCode 14 gets mandatory later, the armv7 builds should be probably removed in jnigen too, but the scope of the pr was only to get it building again.

Tom-Ski:
:O

PR #6994, opened 2022-09-28
Angle backend: Fixing vertex buffer array indexing for glDrawElements

Interrupt:
Rendering sprite batches was causing rendering artifacts when using the Angle backend, it looks like commit https://github.com/libgdx/libgdx/commit/a94b96834f541870174cf90f7c9247da6e97c0eb  broke this rendering path due to changing the other GL20 backends but not the Angle path.

Before:
Screen Shot 2022-09-27 at 9 31 46 PM

After:
Screen Shot 2022-09-28 at 2 13 52 PM


badlogic:
Cheers!

PR #6998, opened 2022-10-03
Concurrecy: allow using LIBGDX in a multicontext-mutithreaded application

MikOfClassX:
This PR allow using LIBGDX in a multicontext-mutithreaded application by getting rid of static fields.

We used thread locals where needed, which are proven to be very fast and have little or no impact to the current LIBGDX performance.

At the same time we removed a lot of unneeded static field declarations.

Plus we added an option in order to allow applications already loading LWJGL natives to avoid re-loading them from LIBGDX
Use **GdxNativesLoader.disableLwjglNativesLoading** = true;

This PR has been deeply tested and does not introduce any api/behavior change


gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -46,6 +84,10 @@ private Intersector () {
 	 * @param t3 the third vertex of the triangle
 	 * @return whether the point is in the triangle */
 	public static boolean isPointInTriangle (Vector3 point, Vector3 t1, Vector3 t2, Vector3 t3) {
+		final Vector3 v0 = new Vector3();
```
yuripourre, line 87: Please, don't instantiate new objects inside collision methods, they are called multiple times. You can get out of memory very fast. One of the reasons we have so many static final fields is to prevent multiple instantiations.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -82,7 +124,9 @@ public static boolean isPointInTriangle (float px, float py, float ax, float ay,
 	}
 
 	public static boolean intersectSegmentPlane (Vector3 start, Vector3 end, Plane plane, Vector3 intersection) {
-		Vector3 dir = v0.set(end).sub(start);
+		final IntersectorData data = tlData.get();
```
yuripourre, line 127: What happens if you call this method in parallel threads? How they will use different structures?

gdx/src/com/badlogic/gdx/math/Matrix4.java
```diff
@@ -94,6 +95,56 @@ public Matrix4 () {
 		val[M33] = 1f;
 	}
 
+	private Quaternion getQuat () {
+		if (quat == null) quat = new Quaternion();
```
yuripourre, line 99: Have you benchmark this? Checking for null in every call might slow the applications down. Have you noticed any difference?

gdx/src/com/badlogic/gdx/math/Matrix4.java
```diff
@@ -1201,54 +1252,54 @@ static inline void matrix4_rot(float* mat, float* vec) {
...
-	 * @param mat the matrix
-	 * @param vecs the vectors
-	 * @param offset the offset into the vectors array
+	 * @param mat     the matrix
+	 * @param vecs    the vectors
+	 * @param offset  the offset into the vectors array
 	 * @param numVecs the number of vectors
 	 * @param stride the stride between vectors in floats */
-	public static native void mulVec (float[] mat, float[] vecs, int offset, int numVecs, int stride) /*-{ }-*/; /*
+	public static native void mulVec(float[] mat, float[] vecs, int offset, int numVecs, int stride) /*-{ }-*/; /*
```
yuripourre, line 1260: This is not following libGDX's code style.

gdx/src/com/badlogic/gdx/math/Matrix4.java
MikOfClassX: lazy instantiation has a very very very low impact in general. We did extensive benchmarks outlining that the jvm is doing a very good job here. Anyway i just modded the class so that only getTmpMat() is lazy. The rest is instance fields.

gdx/src/com/badlogic/gdx/math/Matrix4.java
MikOfClassX: We usually pass the code through Spotless.
Maybe Spotless is not formatting that case i.e. it's nox fixing all spaces on method definitions.
As side note, Spotless is adding an extra newline in comments.

gdx/src/com/badlogic/gdx/math/Intersector.java
dar-dev: Thread local storage is very useful. It's basically a piece of memory belonging to a specific thread.
So.. if you call this method in parallel threads, each thread will refer to its own instance of _IntersectorData_

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java
```diff
@@ -36,33 +36,55 @@
...
+	/** ClassX : support for thread safety
+	 * @author dar */
+	static final class NioThreadLocalData {
+		public ByteBuffer buffer;
+		public FloatBuffer floatBuffer;
+		public IntBuffer intBuffer;
+	}
+
+	// nio thread local static instance
+	private final static NioThreadLocal nioThreadLocal = new NioThreadLocal();
```
intrigus, line 62: This should be `private static final` (per https://stackoverflow.com/questions/16731240/what-is-a-reasonable-order-of-java-modifiers-abstract-final-public-static-e/16732059#16732059).

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphicsOffScreen.java
```diff
@@ -0,0 +1,414 @@
...
+import com.badlogic.gdx.graphics.Cursor.*;
+import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.GL30;
+import com.badlogic.gdx.graphics.glutils.*;
+import com.badlogic.gdx.utils.*;
+
+/** LwjglGraphicsOffScreen implementation derived from {@link LwjglGraphics}. This class uses an external GL context.
+ * 
+ * @author ClassX */
+public class LwjglGraphicsOffScreen extends AbstractGraphics {
```
intrigus, line 35: Is this class strictly necessary?
It seems to be pretty application specific, i.e., what does it offer to the average developer?

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphicsOffScreen.java
```diff
@@ -0,0 +1,414 @@
...
+	/** The suppored OpenGL extensions */
```
intrigus, line 36:
```suggestion
	/** The supported OpenGL extensions */
```

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
```diff
@@ -113,6 +113,15 @@ public boolean validate (final BaseShader shader, final int inputID, final Rende
 	public Camera camera;
 	private Mesh currentMesh;
 
+	/** ClassX: thread-safety support */
+	public final Matrix4 tempMatrix4 = new Matrix4();
```
intrigus, line 117: Could this maybe be changed to `protected`?
At least for `ParticleShader` it should work (Didn't check all other occurrences).

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
```diff
@@ -113,6 +113,15 @@ public boolean validate (final BaseShader shader, final int inputID, final Rende
...
+
+	/** ClassX: thread-safety support */
+	public final Matrix3 tempMatrix3 = new Matrix3();
```
intrigus, line 120: `protected` also here?

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
```diff
@@ -113,6 +113,15 @@ public boolean validate (final BaseShader shader, final int inputID, final Rende
...
+
+	/** ClassX: thread-safety support */
+	public final Vector3 tempVector3 = new Vector3();
```
intrigus, line 123: `protected` also here?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
```diff
@@ -136,71 +137,73 @@ public static void buildNormals (MeshPartBuilder builder, Renderable renderable,
...
 			if (binormalOffset != -1) {
 				builder.setColor(binormalColor);
-				builder.line(tmpV0, tmpV6);
+				builder.line(data.tmpV0, data.tmpV6);
 			}
 		}
 	}
 
-	private static void ensureVerticesCapacity (int capacity) {
+	private static float[] ensureVerticesCapacity (int capacity, float[] vertices) {
```
intrigus, line 189: Maybe I'm wrong, but isn't this function unused?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
```diff
@@ -136,71 +137,73 @@ public static void buildNormals (MeshPartBuilder builder, Renderable renderable,
...
 		if (vertices == null || vertices.length < capacity) vertices = new float[capacity];
+		return vertices;
 	}
 
-	private static void ensureIndicesCapacity (int capacity) {
+	private static short[] ensureIndicesCapacity (int capacity, short[] indices) {
```
intrigus, line 194: Maybe I'm wrong, but isn't this function unused?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
```diff
@@ -123,11 +124,11 @@ public static void buildNormals (MeshPartBuilder builder, Renderable renderable,
 
 		if (mesh.getNumIndices() > 0) {
 			// Get min vertice to max vertice in indices array
-			ensureIndicesCapacity(mesh.getNumIndices());
```
intrigus, line 126: Why did you remove `ensureIndicesCapacity`?

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
```diff
@@ -136,71 +137,73 @@ public static void buildNormals (MeshPartBuilder builder, Renderable renderable,
 			verticesQuantity = renderable.meshPart.size;
 		}
 
-		ensureVerticesCapacity(verticesQuantity * attributesSize);
```
intrigus, line 139: Why did you remove `ensureVerticesCapacity`?

gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
```diff
@@ -210,13 +210,13 @@ private int loadShader (int type, String source) {
 
 		int compiled = intbuf.get(0);
 		if (compiled == 0) {
-// gl.glGetShaderiv(shader, GL20.GL_INFO_LOG_LENGTH, intbuf);
-// int infoLogLength = intbuf.get(0);
-// if (infoLogLength > 1) {
+			// gl.glGetShaderiv(shader, GL20.GL_INFO_LOG_LENGTH, intbuf);
```
intrigus, line 213: I personally see little value in *not* completely removing these comments, but this might be better for another PR.

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -33,11 +33,49 @@ public final class Intersector {
...
+		@Override
+		protected IntersectorData initialValue () {
+			return new IntersectorData();
+		}
+	}
+
+	/** ClassX: thread-safety support
+	 * @author dar */
+	static class IntersectorData {
+		final Vector3 v0 = new Vector3();
```
intrigus, line 51: Can you not make all these fields `private`?

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -240,25 +280,22 @@ static public boolean intersectPolygonEdges (FloatArray polygon1, FloatArray pol
...
 	}
 
-	static Vector2 v2a = new Vector2();
-	static Vector2 v2b = new Vector2();
-	static Vector2 v2c = new Vector2();
-	static Vector2 v2d = new Vector2();
-
 	/** Returns the distance between the given line and point. Note the specified line is not a line segment. */
 	public static float distanceLinePoint (float startX, float startY, float endX, float endY, float pointX, float pointY) {
-		float normalLength = Vector2.len(endX - startX, endY - startY);
```
intrigus, line 250: What's wrong with `Vector2.len`?

gdx/src/com/badlogic/gdx/math/Matrix4.java
```diff
@@ -73,16 +73,19 @@ public class Matrix4 implements Serializable {
...
-	static final Vector3 l_vez = new Vector3();
-	static final Vector3 l_vex = new Vector3();
-	static final Vector3 l_vey = new Vector3();
-	static final Vector3 tmpVec = new Vector3();
-	static final Matrix4 tmpMat = new Matrix4();
-	static final Vector3 right = new Vector3();
-	static final Vector3 tmpForward = new Vector3();
-	static final Vector3 tmpUp = new Vector3();
+	// avoid recursive constructors : stack overflow
+	private Matrix4 tmpMat = null;
```
intrigus, line 77: You could also change this to a thread-local, couldn't you?

gdx/src/com/badlogic/gdx/math/Matrix4.java
```diff
@@ -1124,131 +1134,132 @@ public String toString () {
 	}
 
 	// @off
-	/*JNI
```
intrigus, line 1127: I think it you will have be to revert ALL changes to the following lines.

gdx/src/com/badlogic/gdx/math/Quaternion.java
```diff
@@ -27,8 +27,8 @@
  * @author xoppa */
 public class Quaternion implements Serializable {
 	private static final long serialVersionUID = -7661875440774897168L;
-	private static Quaternion tmp1 = new Quaternion(0, 0, 0, 0);
-	private static Quaternion tmp2 = new Quaternion(0, 0, 0, 0);
+	private Quaternion tmp1 = null;
```
intrigus, line 30: Could this also be changed to a thread-local?

gdx/src/com/badlogic/gdx/math/Vector3.java
```diff
@@ -38,12 +38,20 @@ public class Vector3 implements Serializable, Vector<Vector3> {
 	public final static Vector3 Z = new Vector3(0, 0, 1);
 	public final static Vector3 Zero = new Vector3(0, 0, 0);
 
-	private final static Matrix4 tmpMat = new Matrix4();
+	private Matrix4 tmpMat = null;
```
intrigus, line 41: Could this also use a thread-local?

gdx/src/com/badlogic/gdx/utils/Scaling.java
```diff
@@ -34,9 +32,7 @@ public Vector2 apply (float sourceWidth, float sourceHeight, float targetWidth,
 			float targetRatio = targetHeight / targetWidth;
 			float sourceRatio = sourceHeight / sourceWidth;
 			float scale = targetRatio > sourceRatio ? targetWidth / sourceWidth : targetHeight / sourceHeight;
-			temp.x = sourceWidth * scale;
-			temp.y = sourceHeight * scale;
-			return temp;
+			return new Vector2(sourceWidth * scale, sourceHeight * scale);
```
intrigus, line 35: This creates a new `Vector2` on every invocation, hmmm.

gdx/src/com/badlogic/gdx/utils/Select.java
```diff
@@ -35,7 +35,7 @@ public class Select {
 	private QuickSelect quickSelect;
 
 	/** Provided for convenience */
-	public static Select instance () {
+	public synchronized static Select instance () {
```
intrigus, line 38: What's the point of `synchronized` here?

gdx/src/com/badlogic/gdx/utils/Sort.java
```diff
@@ -60,7 +60,7 @@ public <T> void sort (T[] a, Comparator<? super T> c, int fromIndex, int toIndex
 	}
 
 	/** Returns a Sort instance for convenience. Multiple threads must not use this instance at the same time. */
-	static public Sort instance () {
+	static synchronized public Sort instance () {
```
intrigus, line 63: What's the point of `synchronized` here?

gdx/src/com/badlogic/gdx/utils/GdxNativesLoader.java
```diff
@@ -18,6 +18,7 @@
 
 public class GdxNativesLoader {
 	static public boolean disableNativesLoading = false;
+	static public boolean disableLwjglNativesLoading = false;
```
intrigus, line 21: Can you explain again, when this is needed?
Also, wouldn't it make sense to move this to `LwjglNativesLoader`?
Note that `LwjglNativesLoader` already has a `load` field that you could use instead with the same effect I think.

intrigus (commented):
This is still feels like a somewhat risky changeset.
Not because it is really risky, but it touches many different files.

Did you test your changes on GWT?
GWT is probably the biggest problem.

Also, performance on Android/iOS=RoboVM may considerably vary.
Did you also try your changes there? Sadly, A desktop JVM is __way__ more powerful than a mobile JVM.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphicsOffScreen.java
tommyettinger: It looks like it's enabled by making multi-context possible, so it seems to make sense here. As far as I can tell, this would be useful for off-screen framebuffer rendering when you have many framebuffers at once. I haven't read this large PR thoroughly, though.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java
tommyettinger: Oh my God, haha... I believe I have made this point myself at least once and I have been told that `static public final` would be the way to do it here. There is no behavior difference.

obigu:
I understand this may be useful for some specific types of apps but I agree with @intrigus it's a risky PR. It introduces changes on lots of core classes that are used intensively on many projects. We would need to make sure it runs fast on GWT and older Android and iOS devices and test it intensively. I'm also not sure the extra code complexity and the the implication that from now on concurrency (or specific rules such as forbidding statics) has to be taken into account for changes or new features is worth the benefit considering libGDX itself has not been design for it.

gdx/src/com/badlogic/gdx/math/Quaternion.java
```diff
@@ -780,7 +792,7 @@ public float getAxisAngleRad (Vector3 axis) {
 		float angle = (float)(2.0 * Math.acos(this.w));
 		double s = Math.sqrt(1 - this.w * this.w); // assuming quaternion normalised then w is less than 1, so term always positive.
 		if (s < MathUtils.FLOAT_ROUNDING_ERROR) { // test to avoid divide by zero, s is always positive due to sqrt
-			// if s close to zero then direction of axis not important
+																// if s close to zero then direction of axis not important
```
Frosty-J, line 795: These indented comments only align when using a tab size of 3, which is unusual. I expect most people here use a size of 4 or 2, and GitHub defaults to 8 for some insane reason. It's unfortunate we don't all use the same tab size, as I am also a fan of aligning things this way (just I use spaces to do so if it's my own codebase).

Sorry I can't add anything particularly insightful to this pull request. Static primitives are okay - it's just the objects that are problematic?

I'm neutral on these changes if they don't break anything or greatly affect performance (a small performance impact I'd never notice and some of libGDX's optimisations are no longer important anyway) but I think it's most likely this will have to remain as a fork.

tommyettinger: Static primitives still have issues with multiple threads, if there's contention for that static variable.

gdx/src/com/badlogic/gdx/math/Intersector.java
MikOfClassX: done! Thanks!

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGL20.java
MikOfClassX: private static final should be the order. Apart from that, there is no behavior difference.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphicsOffScreen.java
MikOfClassX: yep, as Tommy writes, not every application is starting with a Display.
There may be cases where you need to create only offscreen contexts using FBOs. This is expecially true for Swing apps,

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
MikOfClassX: definitely a  good idea. We tend to not alter the fields visibility of original code.

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
MikOfClassX: yep it should be. No real reason to make temp stuff public

gdx/src/com/badlogic/gdx/utils/Scaling.java
MikOfClassX: let's look into it @dar-dev

gdx/src/com/badlogic/gdx/utils/Select.java
MikOfClassX: singleton pattern. What happens when multiple threads try to access instance() ?

gdx/src/com/badlogic/gdx/utils/GdxNativesLoader.java
MikOfClassX: basically one GL application is already loding LWJGL (i.e. from different lib folders) before LIBGDX, so we just give the option to avoid a second unnecessary loading.

gdx/src/com/badlogic/gdx/math/Quaternion.java
MikOfClassX: Most of the mods in this PR is change static fields to class instance fields.
This was easy, as there is really no need to make them static. No perf impact, better data design.
As @tommyettinger says, static primitives may have issues, so we basically checked them without finding any real danger.
We did not cleanup all the code with overrides and finals and visibility changes ofc. The result would have been a gigantic PR .. impossible to handle.

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
MikOfClassX: @dar-dev

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: probably @dar-dev

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: we did not alter the api, but @dar-dev can look into it

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: @dar-dev

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: @dar-dev

gdx/src/com/badlogic/gdx/math/Intersector.java
MikOfClassX: improvement @dar-dev ?

gdx/src/com/badlogic/gdx/math/Matrix4.java
MikOfClassX: @dar-dev

gdx/src/com/badlogic/gdx/math/Matrix4.java
MikOfClassX: @dar-dev

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
dar-dev: this matrix is referenced outside of BaseShader (i.e. ParticleShader).. so unfortunately I cannot change the visibility

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
dar-dev: _tempMatrix3_ is the only one not referenced outside, so I can change the visibility to _protected_

gdx/src/com/badlogic/gdx/graphics/g3d/shaders/BaseShader.java
dar-dev: this _Vector3_ is referenced outside of BaseShader (i.e. ParticleShader).. so unfortunately I cannot change the visibility

gdx/src/com/badlogic/gdx/math/Intersector.java
dar-dev: here I kept the visibility as before PR but yes, I can make these fields `private`

gdx/src/com/badlogic/gdx/utils/Scaling.java
dar-dev: my bad, the _apply()_ method is called multiple times. I have to use thread-local here

gdx/src/com/badlogic/gdx/math/Matrix4.java
dar-dev: I think thread-local isn't strictly necessary here. This field is not _static_ so there's no issue when using different instances of Matrix4 per thread

gdx/src/com/badlogic/gdx/math/Intersector.java
dar-dev: I didn't get the [this](https://github.com/libgdx/libgdx/commit/abb9f5b116a1470c7b23e0426a220b2515977165) commit. I'm reverting it

gdx/src/com/badlogic/gdx/math/Quaternion.java
Frosty-J: Okay, sounds good.

MikOfClassX: For our projects we are using Eclipse whose code cleanup is quite efficient and accurate.
So we're no longer bothering with self-made formatting, finals etc. We do format on save action.

BTW 'final' is crucial for a better design and.. gives some spice to the compiler optimization.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: it was originally referring to a static storage.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: it is. Let me review it

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: it is, and can be removed. It was useful for static allocation. No longer now.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: same. No longer used.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: it sas used for static re-allocation. No longer used.

gdx/src/com/badlogic/gdx/graphics/g3d/utils/shapebuilders/RenderableShapeBuilder.java
MikOfClassX: caching a float array here is a bit of overkill imho. Also I did not find any reference to RenderableShapeBuilder in LIBGDX. Am I wrong ?

MikOfClassX: Double checked it. No way to optimize unless we change the visibility of some unreferenced methods (and we want to keep the libgdx api). But we can remove the old ensureIndicesCapacity().

Frosty-J:
mgsx doesn't like it if you keep your fork up-to-date with master. I know this from experience.

MikOfClassX:
@Frosty-J ehm, I feel sorry, but I did not get you. Is there something I am missing here ?

Frosty-J:
I've had a similar commit history before.

<table>
<tr>
<td>Merge branch 'libgdx:master' into master ×14</td>
<td>Discord screenshot</td>
</tr>
</table>

MikOfClassX:
Oh I see. So each time I update my fork, a new "merge" message is sent ? Wonderful :)
This is absolutely unwanted, but ofc we do like to have our branch in sync with master.

Frosty-J:
I don't think you can prevent it. GitHub is dumb. I'm not bothered and if it were up to me it would automatically sync.

mgsx-dev:
> mgsx doesn't like it if you keep your fork up-to-date with master. I know this from experience.

What i don't like even more is people talking in my name and/or screenshot some of my conversations to post them on github. Please stop that.

PR #6999, opened 2022-10-04
Add action6

aliasifk:


Frosty-J (changes requested):
Wouldn't [`ParallelAction`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/ParallelAction.java) and [`SequenceAction`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/SequenceAction.java) also need updating? This is a silly pull request anyway.

aliasifk:
Its very beneficial as the (Object..) method creates garbage

Frosty-J:
It's not the only thing around here that creates garbage.

MobiDevelop:
I personally feel like 5 actions before varargs is sufficient. Otherwise, why stop at 6? Why not 10? It's easy enough to make your own utility class that takes as many args as you like and makes an action.

PR #7004, opened 2022-10-14
Add API to get user input with native text fields

Berstanio:
The current solution to get input from a soft keyboard is not perfect.
On iOS there is a hidden native text input field, where we intercept the characters and on android we directly intercept characters from the IME.
This has some problems:
1. On Android autocomplete/suggestion/correct is not possible. On iOS it would be theoretical possible, but not in the current design.
2. There are problems with e.g. korean symbols not getting correctly combined on iOS. I wont go into the root cause detail, since it is complicated and not relevant here.

This pr is a proposal for a new mobile input API to fix these issues.
The idea is to have a visible general native input text field where the user writes. When the keyboard is closed, the "finished" text is just extracted from the native input text field and set directly. This way, we leave all to the OS like auto correction, character combination on other languages etc.

A few notes about design decisions:
Since it is not guaranteed, who calls "closeTextInput" (the system or the app), I implemented a abstraction of a "input field" (like in scene2d). It needs to implement generic methods like "get/setText" and "get/setPosition". When the soft keyboard is closed, "setText" will be called. 
Because `getKeyboardTextField` didn't worked with some design changes, since it would be not clear what it should return, I changed it to `getActiveKeyboardTextField`. If no TextField is active, it returns null. This will probably be a breaking change. It can still all be configured the same, just only when there is a open keyboard.
On android I needed to implement a few locks and waits to prevent race conditions that could occur from user code.


Frosty-J:
> On Android autocomplete/suggestion/correct is not possible.

Additionally, it has odd behaviour for non-Latin keyboards. It may either force QWERTY or ignore keypresses altogether. Behaviour can vary between AOSP, GBoard, etc. In my view it has no use for anything entering production and games are generally better off using gdx-dialogs or potentially this PR (I haven't tried it).

Berstanio:
Updated pr, fixed lot's of bugs and made it more robust in general.

> In my view it has no use for anything entering production and games are generally better off using gdx-dialogs or potentially this PR (I haven't tried it).

As far as I'm aware gdx-dialogs fulfils a different purpose and is e.g. not feasible for a chat, right?

PokeMMO (approved):

tommyettinger (approved):
Looks fine; I like that there are tests, and I like that this significantly improves UX on mobile. @carlislefox may be interested in the outcome of this PR, as per Discord conversation! Approved, hopefully to enable documentation updates sooner.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/keyboardheight/KeyboardHeightObserver.java
```diff
@@ -0,0 +1,28 @@
...
+ * (at your option) any later version.
+ *
+ * Siebe Projects samples is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * Lesser GNU General Public License for more details.
+ *
+ * You should have received a copy of the Lesser GNU General Public License
+ * along with Siebe Projects samples.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
intrigus, line 16: Is libGDX's license actually compatible with this code?

tommyettinger: It depends on the copyrightability of interfaces. (IANAL) The US Supreme Court appeared to rule in favor of Google in Oracle v. Google, where Oracle's argument was that its interfaces were its property and copyright Oracle, and Google's argument was that an interface specifies a calling convention and does not constitute code. So, the highest court in my country (where GitHub is incorporated) says it is compatible, but I have no idea what it says in Austria.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/keyboardheight/StandardKeyboardHeightProvider.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * This file is part of Siebe Projects samples.
+ *
+ * Siebe Projects samples is free software: you can redistribute it and/or modify
+ * it under the terms of the Lesser GNU General Public License as published by
```
tommyettinger, line 5: @intrigus This block is more concerning, though it's still just from a sample, and LGPL (v3) is more compatible than GPL (v2, which OpenJDK uses) with permissive licenses. I'm not sure if LGPL v3 is compatible enough with Apache v2; I think it is, though I personally steer clear of any GNU-licensed code in my non-GNU-licensed projects.

tommyettinger: The "Siebe Projects samples" mentioned is here: https://github.com/siebeprojects/samples-keyboardheight

It doesn't seem trivial to reimplement from scratch.

PokeMMO: https://github.com/siebeprojects/samples-keyboardheight/issues/10

I don't understand this timeline, maybe we can ask him to finish license change?

PR #7005, opened 2022-10-14
Update gha to ubuntu-20.04

Berstanio:
Soo, GitHub will remove the ubuntu 18.04 image by 4/1/2023.
https://github.com/actions/runner-images/issues/6002
They are already starting to let runner fail at specific times.
This is what happened e.g. here https://github.com/libgdx/libgdx/actions/runs/3250859896/jobs/5335094874 even though date wasn't mentioned as brownout date.

So the only way is probably upgrading. However because of the dynamic linking against libc, afaik this change would make the binarys incompatible with older OS.
The only way I could think of to work around this would be, to run a nested docker container for builds, but not sure whether this would be feasible.

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -112,7 +112,7 @@ jobs:
           rm -f ~/.gradle/caches/modules-2/gc.properties
 
   natives-linux:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
@@ -162,7 +162,7 @@ jobs:
           rm -f ~/.gradle/caches/modules-2/gc.properties
 
   natives-windows:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


Frosty-J:
That action ran on 14 Oct 14:52 UTC. If you look at your linked issue's edit history, you'll see the second brownout was scheduled from 12:00 to 16:00 on the 14th, only they've since edited it to the 18th. The brownout and target dates have changed a multitude of times. I don't know if that means we'll have another one on the 18th or if that's the one we've already had. Their entire approach is a farce - just give everyone with an 18.04 action a warning email or something.

I would like to delay the upgrade, at least until after the next libGDX release (unless it's another year away, of course) if compatibility is affected, but I don't want brownout disruption either. We could limit the impact of this by upgrading `build-pullrequest.yml`, as that doesn't build natives so no difference there.

Tom-Ski:
It would be VERY nice to avoid running a container.   Any ideas on roughly how bad the backwards compatibility would be?

Berstanio:
From what I've read it would be definitly incompatible with 16.04 or debian 9 (don't know about others distros).
It *could* be incompatible with 18.04, but this is up to luck on how far away the libc versions are. This would be needed to get tested.
I don't know about debian 10, also would need a test.

Berstanio:
Just a little bump, the removal date of the runner is apparently supposed to be today, so I guess a decision on what to do should be made?

SimonIT:
We need an approve from @libgdx/libgdx-signing to merge it

PR #7006, opened 2022-10-15
Fix ExtendViewport.

dmoklyakov:
Hi there!

Recently I started to use ExtendViewport in my game and noticed strange behavior: in some cases it doesn't scale content on screen in correct way.

I use perspective camera for my 3D game and when I try to narrow down the game window, content doesn't fit on the screen horizontally (viewport works like I use "fill" scaling mode).
I looked in the library code and found a check that compares scaled viewport and screen sizes instead of viewport aspect ratio.

This PR fixes this issue.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/viewport/ExtendViewport.java
+++ b/gdx/src/com/badlogic/gdx/utils/viewport/ExtendViewport.java
@@ -68,15 +68,14 @@ public class ExtendViewport extends Viewport {
 		// Extend, possibly in both directions depending on the scaling.
 		int viewportWidth = Math.round(scaled.x);
 		int viewportHeight = Math.round(scaled.y);
-		if (viewportWidth < screenWidth) {
+		if (viewportWidth < viewportHeight) {
 			float toViewportSpace = viewportHeight / worldHeight;
 			float toWorldSpace = worldHeight / viewportHeight;
 			float lengthen = (screenWidth - viewportWidth) * toWorldSpace;
 			if (maxWorldWidth > 0) lengthen = Math.min(lengthen, maxWorldWidth - minWorldWidth);
 			worldWidth += lengthen;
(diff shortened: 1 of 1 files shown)


obigu:
I don't think this is correct, what those checks do is actually what allows the viewport to extend to the whole screen (considering max limits).

Maybe if you share some code that displays the issue we can better understand the problem.

dmoklyakov:
example
I hope this image explain situation better then words. If you really need some code, I can do a separate example project later.

dmoklyakov:
According to [wiki](https://libgdx.com/wiki/graphics/viewports#extendviewport), ExtendViewport should scale content preserving its aspect ratio, and then shorten direction should be lengthened to fill the viewport.

obigu:
Ok, in the example you show, doesn't setting `minWorldWidth` to the width of the board in world units work as expected?

dmoklyakov:
Nope, the problem is still there.

obigu:
Maybe you can use [ViewportTest1](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/ViewportTest1.java) to show the problem. As I understand, the table in the test would be the equivalent of the board in you example, and it scales respecting `minWorldWidth`.

dmoklyakov:
[Here](https://github.com/dmoklyakov/libgdx/commit/452926843265582aaf2376ac3a0f9537232bade5) I was able to reproduce exactly the same situation as it is in my game.

dmoklyakov:
Maybe there is a problem in my code which I didn't notice?

obigu:
I'm not familiar on how viewports are expected to behave on 3D games, I'm not even sure it makes sense to use them (vs working directly with the camera) but you may get some more help on Discord 3d channel. If you manage to reproduce on 2D feel free to share.

dmoklyakov:
Discussed on Discord: nothing to do here.

Issue #6934, opened 2022-07-22
Tiled 1.9 "type" converted to "class"


Quillraven:
I am just wondering if anyone else is facing a similar issue. When upgrading to Tiled 1.9 the previous "type" information of a MapObject is no longer loaded in LibGDX. I guess it is related to this change:

![image](https://user-images.githubusercontent.com/93260/180390794-e3ec8e3b-7d88-49e2-b616-356439302029.png)

I notice that the .tmx file now stores a "class" information instead of a type but it seems like LibGDX is not loading this info at all? At least I cannot find it.

Is this simply not supported at the moment and we have to work with Tiled <1.9 or am I doing something wrong?

tommyettinger:
I'd use that compatibility option if you use Tiled 1.9 or higher. This is tricky since we might need to parse both "class" and "type" to be compatible with existing Tiled maps.

Quillraven:
Okay thanks. I could not figure out how to set the first time but I will look into it again. I guess since I only have a map without a project I couldn't set it. Guess I need a project as well :D

LuoiZQ:
You can replace the class in the exported tmx file with type. This method solves the problem that the gdx program cannot read the class, but note that the class will be changed back to the class after editing with tiled again.

SimonIT:
There is already #7007

Quillraven:
There was a discussion also in Tiled repo because this change broke a couple of things also for non LibGDX users. I cannot remember if Björn wanted to revert this change or if he keeps it.

I would check that once more before merging the linked PR because maybe the PR is not needed.

MobiDevelop:
Here's the discussion on the Tiled repo, for reference: https://github.com/mapeditor/tiled/issues/3492

PR #7007, opened 2022-10-15
Added support for 'class' property in tiled tmx maps.

lyze237:
Tiled renamed the 'type' property to 'class', so I added support for that.

https://doc.mapeditor.org/en/stable/reference/tmx-changelog/

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@
 - LWJGL3 Fix: Stereo audio can now be played on mono output devices. This may also improve downmixing to stereo and upmixing to surround.
 - API Addition: Added CheckBox#getImageDrawable.
 - FIX: HexagonalTiledMapRenderer now displays maps with the correct stagger index.
+- API Addition: Added support for 'class' properties in 1.9 tiled tmx maps.
 
 [1.11.0]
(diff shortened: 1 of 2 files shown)


MobiDevelop:
Looks okay to me, but appears to have conflicts what need to be resolved.

tommyettinger:
This looks like it would solve #7047 . As always, I have no idea what I am doing with Tiled and am not a good person to review this. Do we have anyone who can review Tiled code and make sense of it?

SimonIT:
As stated in #6934 
> There was a discussion also in Tiled repo because this change broke a couple of things also for non LibGDX users. I cannot remember if Björn wanted to revert this change or if he keeps it.
>
> I would check that once more before merging the linked PR because maybe the PR is not needed.

lyze237:
So:
> So for Tiled 1.10, we're now back to using type for objects and tiles, but for other types of objects we're still using class in order to not break compatibility again.

Therefore I think this PR should be merged as is, as then whoever uses Tiled 1.9 still gets supported for the class type on tiles. (+ it doesn't new or old versions)

lyze237:
I think there is no need to implement https://github.com/libgdx/libgdx/pull/7007 anymore as the change has been [reverted over more than a year ago](https://www.mapeditor.org/2023/03/10/tiled-1-10-released.html).

Issue #7008, opened 2022-10-15
Support Java Color constructor (int r, int g, int b, int a)


HydrolienF:
#### Issue details
Create a new Color as we can do in Java cause issues
```java
new Color(8, 194, 0, 255)
```
Java developpers, knowing that they use int args think that they will get a dark green with fiew blue.
Insted it create a color full Red, full Green 0 Blue.
I don't get why but there is no error, int are used in the float constructor.

#### Reproduction steps/code
```java
package com.test;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.glutils.ShapeRenderer;
import com.badlogic.gdx.graphics.glutils.ShapeRenderer.ShapeType;
import com.badlogic.gdx.utils.ScreenUtils;

public class ColorTest extends ApplicationAdapter {
    private ShapeRenderer shapeRenderer;

	@Override
	public void create () {
		shapeRenderer = new ShapeRenderer();
		shapeRenderer.setColor(new Color(8, 194, 0, 255));
	}

	@Override
	public void render () {
		ScreenUtils.clear(1, 1, 1, 1);
		shapeRenderer.begin(ShapeType.Filled);
		shapeRenderer.circle(100, 100, (float) 100);
		shapeRenderer.end();
	}
	
	@Override
	public void dispose () {}
}

``

#### Version of libGDX and/or relevant dependencies
libGDX 1.11.0
Java 17

#### Please select the affected platforms
- [x] Android
- [x] iOS
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] macOS

**Comment by Frosty-J** [comment]
This or a bounds check sounds like a good idea to me. Using 0-255 ints instead of 0-1 floats is a common mistake. My favourite approach in this case would be `new Color(0x08c200ff)` - it isn't obvious you can use hex codes like that!

Related PR (since you didn't link them together): #7009

**Comment by HydrolienF** [comment]
Is something like
```java
public Color(int r, int g, int b, int a){
  if(r>1 || g>1 || b>1 || a>1){
    this(r/255f, g/255f, b/255f, a/255f);
  }else{
    this(r,g,b,a);
  }
}
```
a good way to solve it without breaking any current code as
```java
new Color(1,1,0,1)
```
?

Or people that try to use int will finaly check the javadoc and understand why the int color constructor don't work as the Java one ?

mgsx-dev:
[...]

PR #7009, opened 2022-10-15
Add Constructor of Color as Java have. Solve #7008

HydrolienF:


gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -82,17 +82,27 @@ public Color (int rgba8888) {
...
 	}
+	
+	/** Constructor, sets the components of the color
+	 * 
+	 * @param r the red component as int in [0,255]
+	 * @param g the green component as int in [0,255]
+	 * @param b the blue component as int in [0,255]
+	 * @param a the alpha component as int in [0,255] */
+	public Color (int r, int g, int b, int a) {
+		super(r/255f, g/255f, b/255f, a/255f);
```
MobiDevelop, line 104: I assume you intended it to be `this(...);`

HydrolienF: o yes. I'm fixing it

mgsx-dev:
it would totally break some user code (eg. `new Color(0,0,0,1)`) which is converted to floats. With this change, their color will be broken. It already breaks libgdx predefined colors (eg. Color.WHITE). You can use `Color.toIntBits` or `Color.valueOf(hexString)` instead.

Frosty-J:
Mm, good point. I might have a few of those implicit conversions. Heck, even gdx-setup does `ScreenUtils.clear(1, 0, 0, 1)` instead of `Color.RED`.

HydrolienF:
> it would totally break some user code (eg. `new Color(0,0,0,1)`) which is converted to floats.

Oh rigth. I understand.
Or we can check if one value is over 1 then use same int builder than Java else use float builder.

Anyway, that's not realy a major issues even if it's disturbing for beginners.

Frosty-J:
Using ints if over 1 else floats could cause bugs. Let's say someone fades from white to black by starting at 255 and ending at 0. Most of the fade will be fine, but at 1 there will be a flash of white. If a fade occurs over a period of 1 second at a refresh rate of 60Hz, the flash would only occur in approximately 1 in 4 fades, making the bug all the more confusing.

PR #7014, opened 2022-10-21
Convert 16bit colors to 32bit colors on MetalANGLE/Simulator

Berstanio:
Currently the iOS simulator wont work with MetalANGLE in all cases because of https://github.com/kakashidinho/metalangle/issues/68.
This is a first fix for this by converting 16bit colors manually to 32bit colors when running on simulator.
The fix can be confirmed by running the "BitmapFontAtlasRegionTest" test.

The only thing I'm atm unsure is, whether it is wrong to expect a ByteBuffer. But as far as I know, a direct buffer needs to be passed, and all direct Buffer extend from ByteBuffer.


backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
```diff
@@ -263,7 +262,55 @@ public native void glCopyTexImage2D (int target, int level, int internalformat,
...
+			}
+			converted.putInt(rgba8888);
+		}
+		converted.position(0);
+		return converted;
+	}
+
+	public void glTexImage2D (int target, int level, int internalformat, int width, int height, int border, int format, int type,
+		Buffer pixels) {
+		if (!IOSApplication.IS_METALANGLE
```
Tom-Ski, line 300: Maybe cache this in the init to a boolean so we aren't checked each time.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
```diff
@@ -263,7 +262,55 @@ public native void glCopyTexImage2D (int target, int level, int internalformat,
 
 	public native void glStencilOpSeparate (int face, int fail, int zfail, int zpass);
 
-	public native void glTexImage2D (int target, int level, int internalformat, int width, int height, int border, int format,
+	private static Buffer convert16bitBufferToRGBA8888 (Buffer buffer, int type) {
+		// TODO: 20.10.22 Can it be a different buffer type?
```
Tom-Ski, line 266: Might be better suited to take this out to some util

Tom-Ski (changes requested):
Does this problem occur on different macs in the same way?  For example M1 vs non M1?

This would also be required to do in the GL30 interfaces too no?

Berstanio:
> Does this problem occur on different macs in the same way? For example M1 vs non M1?

I think so, but I can only test on a M1.

> This would also be required to do in the GL30 interfaces too no?

You mean e.g. glTexImage3D? Yeah, probably, I oversaw that.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
Berstanio: Do you have a suggestion of a already existing util class?

Tom-Ski:
> I think so, but I can only test on a M1.

Me too atm, would be good if someone can test to see if this has adverse effects on non M1

> You mean e.g. glTexImage3D? Yeah, probably, I oversaw that.

Yeah

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
Tom-Ski: BufferUtils is probably the closest one, but this seems a bit too niche for it to be there. I guess its probably ok to have here, but it would need to be protected so GLES30 can call it too where it needs to

Berstanio:
> Does this problem occur on different macs in the same way? For example M1 vs non M1?

https://developer.apple.com/documentation/metal/developing_metal_apps_that_run_in_simulator?language=objc
Seems so according to the docs:

"Don't use the following pixel formats: [MTLPixelFormatR8Unorm_sRGB](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatr8unorm_srgb?language=objc), [MTLPixelFormatB5G6R5Unorm](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatb5g6r5unorm?language=objc), [MTLPixelFormatA1BGR5Unorm](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformata1bgr5unorm?language=objc), [MTLPixelFormatABGR4Unorm](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatabgr4unorm?language=objc), [MTLPixelFormatBGR5A1Unorm](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatbgr5a1unorm?language=objc), or any XR10 or YUV formats."

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/IOSGLES20.java
Berstanio: Done

obigu:
Tested running `BitmapFontAtlasRegionTest` on ~~EXCRUCIATINGLY SLOW~~ Intel  Mac and it crashes without the PR. With the PR test runs ok (even if it's quite an ugly test).

Tom-Ski:
Ok lets have this in then for now and see how it goes

PR #7015, opened 2022-10-23
Add a default fullscreen style to Android GDXSetup and Android tests

obigu:
This is a proposal to improve default Gdx Setup "splash screen" which displays status and action bars. The issue was raised on a https://github.com/libgdx/libgdx/pull/6571 comment.

The poposed default style inherits from standard styles and customizes the background color to white (I'm not sure if it should be set to red like the amazing GDX Setup default background). There's a fullscreen Material style for API > 21 devices and a Holo style for the rest (because we have min API suport 14 it should be ok). Apps that want a fully consistent style across Android versions can add AppCompat dependency and customize their own style.

Tested on Samsung Galaxy S7 with Android 8.0 and a Pixel 3 Android 12 (which displays a proper splash screen). I haven't been able to test on < 21 devices.


Frosty-J:
Maybe black or grey? We have #222222 for GWT. Dark mode users might not appreciate white, but I'll appreciate it. Note that I haven't checked it out for myself to see what it looks like.

obigu:
Black is not good because users may assume it's not changeable. That's why I like red (vs black or white). If the game shows a red screen after having changed the background on the Java side, you know there must be something setting it that can be edited hidden somewhere else.

Frosty-J:
I guess red would certainly encourage people to find a way to change it. I prefer sane defaults that don't need changing in the first place. Looks like Android used to default to #fff3f3f3 (#aarrggbb) for light mode and #ff000000 for dark mode, but then it changed to #ffffffff for light mode at some point.

So desktop defaults to black, GWT defaults to dark grey, Android now defaults to whatever it will be, and iOS looks like it defaults to white?

obigu:
iOS storyboard includes by default the libGDX icon in addition to the white background so it doesn't really matter because all apps have to change it. Android 12+ added default splash screens which take the background color defined on the theme and the icon drawable in which case it wouldn't really matter either and we could have white so that it always looks good by default. The problem IMO is on Android < 12 that the splash screen will be just a solid color and black would not a good default. We can leave it as it is with white to align it with iOS and with default light them then.

Tom-Ski (approved):

Berstanio (approved):
Worked fine on my test.
It's also nice to have a proper theme for the tests imo!

PR #7020, opened 2022-10-29
Fix HexagonalTiledMapRenderer displaying maps with the wrong stagger index

lyze237:
This is a reopen of https://github.com/libgdx/libgdx/pull/6607

Accidentally closed the old PR by fast forwarding the branch.


tommyettinger (approved):
For some reason I thought there were changes to the tests... Regardless, the change seems small, the previous PR commentary showed that the changes are all correct, and we just need someone else who can adequately judge Tiled map code to chime in before merging. I can't really tell much of anything about Tiled, but the test images before made sense to me. Approved because this has had a long time open without anyone finding anything wrong.

lyze237:
Here's a very old stack exchange question which solved it in the same way: https://gamedev.stackexchange.com/questions/175335/libgdx-hexagonaltiledmaprenderer-wrong-stagger-index

Tom-Ski (approved):

PR #7028, opened 2022-11-14
Avoids unnecessary flushing in case we are drawing

Betalord:
Improved on SpriteBatch.setShader(): it won't flush anymore in case we are drawing and given shader is same as current one.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
@@ -1071,6 +1071,8 @@ public class SpriteBatch implements Batch {
 
 	@Override
 	public void setShader (ShaderProgram shader) {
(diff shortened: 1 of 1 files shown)


Betalord:
A small improvement for SpriteBatch.setShader(): it won't flush anymore in case we are drawing and given shader is same as current one. This can actually happen often, for example I use a distance field fonts and some "regular" bitmap fonts - first require custom shader, second does not. So when switching shaders, I simply override BitmapFont object and the shader, and in the second case, shader is null. However, when drawing text with this font, I always bind shader - in case of normal bitmap font, such shader is null, but SpriteBatch will still call flush() which is expensive, although there was no change in shader.

tommyettinger (approved):
Looks great. This is a simple, clear optimization that I thought we must have had in SpriteBatch already -- I guess we didn't! Approved.

PR #7030, opened 2022-11-15
Apply formatter

Frosty-J:
Builds are currently busted because #6940 got merged without the formatter being run, I think. Interestingly, the formatter can run on my own fork - just not in libGDX's master branch because it's protected.


Tom-Ski:
Lets see

Tom-Ski:
Its not supported on webgl so we should have some excludes

Frosty-J:
Ah, I just got confused by the dates. I forgot that "4 months ago" doesn't mean changes to those classes were merged 4 months ago.

Tom-Ski:
@Frosty-J it was missing an exclude for the test because it had direct imports.  But also... we are on brownout day. So I guess everything is gonna fail today

Frosty-J:
The brownout's due to end in 11 minutes. It's all good.

PR #7034, opened 2022-11-22
Fix #6790; align GWT with other platforms

### LINKED ISSUE #6790: VertexBufferObjectWithVAO.unbindAttributes bug in HTML/GWT
There is a blocking bug currently in the HTML/GWT version in the VertexBufferObjectWithVAO.
The bug is not present on other platforms as the implementation is different/correct. 
The object `cachedLocations` can never be `null` as it's initialized and never reassigned.
This makes the `if` statement useless and triggers therefore a bug (IndexOutOfBoundsException).

The idea is to remove this obvious and blocking bug, to make one little step towards WebGL2 support (#5763), without implementing it entirely.

buggy  HTML/GWT:
https://github.com/libgdx/libgdx/blob/04094910142d11ee4028a4263ceb2597c2d9815f/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/VertexBufferObjectWithVAO.java#L184-L187

other platforms (working):
https://github.com/libgdx/libgdx/blob/04094910142d11ee4028a4263ceb2597c2d9815f/gdx/src/com/badlogic/gdx/graphics/glutils/VertexBufferObjectWithVAO.java#L207-L210

1.10.0

- [ ] Android
- [ ] iOS
- [X] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

intrigus:
The object `cachedLocations` can never be `null` as it's initialized and never reassigned. This makes the if statement useless and triggers therefore a bug (`IndexOutOfBoundsException`).

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/VertexBufferObjectWithVAO.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/graphics/glutils/VertexBufferObjectWithVAO.java
@@ -190,7 +190,7 @@ public class VertexBufferObjectWithVAO implements VertexData {
 	}
 
 	private void unbindAttributes (ShaderProgram shaderProgram) {
-		if (cachedLocations == null) {
+		if (cachedLocations.size == 0) {
 			return;
 		}
 		int numAttributes = attributes.size();


mgsx-dev (approved):
It doesn't hurt since this code is not currently used (will be when WebGL 2 implementation is ready)

Tom-Ski (approved):

PR #7037, opened 2022-11-28
WebGL2 Backend

JamesTKhan:
PR for WebGL2 back-end. I started off with a fresh branch off master and manually cherry picked changes from Intrigus original WebGL2 Draft PR #5763 and then continued implementing the GWT GL30 interface and WebGL2 context methods.

Apologies for such a large PR, but this is a large feature, suggestions welcome. You can see the PR in use for my latest game jam entry which is using this PR's code to run WebGL2: https://wjamesfl.itch.io/santa-is-coming

Some interface methods could not be implemented due to either not existing on WebGL2 (or I cannot find them) like glMapBufferRange, glUnmapBuffer. In those cases, an exception is thrown.

**GL30 Tests that are now enabled and working with WebGL2 back-end enabled:**
- FloatTextureTest
- GL30Texture3DTest
- InstancedRenderingTest
- MultipleRenderTargetTest
- TextureArrayTest

**GL30 Tests not currently working in WebGL2 back-end:**

- VBOWithVAOPerformanceTest - It uses some things that are not emulated, (BufferUtils newUnsafeByteBuffer, disposeUnsafeByteBuffer)
- PixelBufferObjectTest - Uses threading, map/unmapbuffer

I have not looked into what it would take to enable those tests yet, and this PR is already quite large so I do not believe that should happen here.

**New Tests added:**
- UniformBufferObjectsTest for Uniform Buffer Objects to make sure the UBO code I wrote for GWT is working but it also runs on other GL30 platforms.
- NonPowerOfTwoTest - Simple, just renders a Non-POT texture with mipmapping. Visible on all non-GWT platforms without GL30, only visible on GWT when GL30 is available.

**Questions I am unsure of and could use help with:**

1. ~~GwtGraphics.java GwtGL20Debug? Is a GwtGL30Debug needed?~~ Resolved

2. ~~To enable WebGL2, I currently set `config.useGL30 = true;` in GwtTestStarter.java. That is not pushed to the branch. Not sure how we should handle enabling WebGL2 for the GWT tests?~~ Resolved

3. ~~In GwtTestStarter the prependVertexCode and prependFragmentCode are modified for WebGL2. Is this the right place to do this? Should it go somewhere else?~~ Resolved

4. ~~In GwtGL30 there are some methods I was not sure how/if should be implemented on WebGL2. The parameters don't match up with what WebGL2 provides~~  Resolved


Frosty-J:
> VBOWithVAOPerformanceTest - It uses some things that are not emulated, like System.nanoTime()

Bring on the TimeUtils.nanoTime()!

JamesTKhan:
Good call I did not even realize it was already emulated, I updated the test to use TimeUtils but unfortunately looks like there are two other methods which are not emulated on GWT in that test 

```
newUnsafeByteBuffer(int)
disposeUnsafeByteBuffer(ByteBuffer)
```

fgnm:
I'm very positive about this, I tested the PR in my dev environment and all features I use from GLES3 are perfectly fine! Considering that libGDX web implementation has fallen far behind any other backend I'm for merging it... That would be a very good starting point for new features (I'm looking at GELS3.1 and 3.2)

mgsx-dev:
@JamesTKhan regarding the 4 remaining questions: 

1) i don't think it's necessary since GLProfiler already cover that but since it's already part of the API, i think it would be great to have it (when people enable GL debug they expect GL30 methods to be part of the error checking). You could inherit from GwtGL20Debug and make checkError protected. 

2) For desktop, we have command line options in order to enable that. I think we could have a parameter in the URL in order to enable it for web backend (eg. /.../index.html?useGL30=true) idk if it's hard to plug that in though.

3) i think it's ok, we have the same thing for [desktop](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/Lwjgl3TestStarter.java#L58)

4) I also think it's fine, there are few differences like that for other backends, while it's some extra parameters in the implementation, it's OK. When it's some missing parameters in the implementation, then that's more problematic.

Note that i didn't review the PR, i only answered your 4 questions.

Once you think the PR is OK, you can remove the draft flag to make it ready for reviews.

JamesTKhan:
@mgsx-dev Thanks for the feedback!

1. I Added GwtGL30Debug. I ran into inheritance issues trying to inherit from GWTGL20Debug, since we cannot multiple inheritance on GwtGL30 as well. What I did instead was inherit from GwtGL30, override the GL30 methods, and then copy over the overridden GL20 debug methods from GwtGL20Debug as well. Not ideal as its duplicated code..
but I am not sure what other approach to take, let me know if there is a way I am overlooking. 

2. Good idea, I was able to implement the useGL30 URL parameter.
3. All good.
4. Sounds good. I implemented those remaining uniform methods, just using 0 for the offsets. 

I have made the PR ready for review. I notice the GH built action is failing but it is not clear to me why. It looks like the task  `:tests:gdx-tests-gwt:classes` is throwing invalid symbols within the GWT jars but the task passes on my local. Not sure what the deal is.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30.java
```diff
@@ -0,0 +1,917 @@
...
+import java.nio.ByteBuffer;
+import java.nio.FloatBuffer;
+import java.nio.HasArrayBufferView;
+import java.nio.IntBuffer;
+import java.nio.LongBuffer;
+
+/** @author Simon Gerst
+ * @author JamesTKhan */
+public class GwtGL30 extends GwtGL20 implements GL30 {
+	final IntMap<WebGLQuery> queries = IntMap.create();
```
intrigus, line 49: Can we make all/some of these attributes private/protected?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30.java
```diff
@@ -0,0 +1,917 @@
...
+	Uint32Array uIntBuffer = TypedArrays.createUint32Array(2000 * 6);
+
+	final protected WebGL2RenderingContext gl;
+
+	protected GwtGL30 (WebGL2RenderingContext gl) {
+		super(gl);
+		this.gl = gl;
+	}
+
+	public Uint32Array copyUnsigned (IntBuffer buffer) {
```
intrigus, line 66: Can we make this method private?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30Debug.java
```diff
@@ -0,0 +1,1629 @@
...
+
+import java.nio.Buffer;
+import java.nio.FloatBuffer;
+import java.nio.IntBuffer;
+import java.nio.LongBuffer;
+
+/** Does not extend from GwtGL20Debug as we would not have GL30 methods to override meaning we would need to implement all of GL30
+ * here as well.
+ *
+ * Instead, GL30 override methods are first, then GL20 override methods are beneath which were copied from GwtGl20Debug. */
```
intrigus, line 30:
```suggestion
 * Instead, GL30 override methods are first, then GL20 override methods are beneath which were copied from GwtGL20Debug. */
```
Or maybe change this to a proper `{@link GwtGL20Debug}` (hope this was the right syntax)?

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
```diff
@@ -55,7 +56,8 @@ public String getName () {
 	CanvasElement canvas;
 	WebGLRenderingContext context;
 	GLVersion glVersion;
-	GL20 gl;
+	GL20 gl20;
```
intrigus, line 59: Can we also make these two private/protected?

backends/gdx-backends-gwt/src/com/google/gwt/webgl/client/WebGL2RenderingContext.java
```diff
@@ -0,0 +1,555 @@
...
+		} catch (e) {
+			console.log(e);
+		}
+		return null;
+	}-*/;
+
+	protected WebGL2RenderingContext () {
+	}
+
+	public static JsArrayString toJsArray (String[] input) {
```
intrigus, line 64: private?

intrigus (approved):

backends/gdx-backends-gwt/src/com/google/gwt/webgl/client/WebGL2RenderingContext.java
SimonIT: Maybe even put it here https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtUtils.java

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30.java
JamesTKhan: Can do. I mostly followed with what GWTGL20 is doing here, which is not private but I can make them private here.

backends/gdx-backends-gwt/src/com/google/gwt/webgl/client/WebGL2RenderingContext.java
JamesTKhan: I liked the idea of putting it in GwtUtils but it looks like it would not be available unless I update inherited GWT modules. For now will mark it private unless we want to move it into GwtUtils and update module inheritance.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
JamesTKhan: Note: also removing an unused string variable directly underneath these two variables

backends/gdx-backends-gwt/src/com/google/gwt/webgl/client/WebGL2RenderingContext.java
SimonIT: I don't see a reason why it wouldn't be available... But I don't really care

JamesTKhan:
Thanks for the feedback so far! Added some changes based on feedback.

backends/gdx-backends-gwt/src/com/google/gwt/webgl/client/WebGL2RenderingContext.java
JamesTKhan: Maybe it is a false positive in the IDE. This is warning I get but when I just tried moving it over I was able to still start up tests without updating the modules.

image

SimonIT: Ah, I didn't see that it's in the Google package. Good possible that your warning is true then

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30.java
JamesTKhan: Done.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30.java
JamesTKhan: Done

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGL30Debug.java
JamesTKhan: Good catch, done.

SimonIT:
When going back from the `NonPowerOfTwoTest` (left top corner) I get this exception. Not sure why
```
Uncaught Error: java.lang.RuntimeException: com.google.gwt.core.client.JavaScriptException: (TypeError) : Cannot read properties of undefined (reading 'dispose_1_g$')
    at ioe_g$.Ane_g$ [as createError_0_g$] (Throwable.java:120:1)
    at ioe_g$.Kne_g$ [as initializeBackingError_0_g$] (Throwable.java:112:1)
    at ioe_g$.wne_g$ (Throwable.java:80:1)
    at ioe_g$.aoe_g$ (Exception.java:37:1)
    at new ioe_g$ (RuntimeException.java:37:1)
    at Qm_g$.Rm_g$ [as execute_0_g$] (GwtApplication.java:252:1)
    at AnimationSchedulerImplStandard.java:39:1
    at dAi_g$ (Impl.java:309:1)
    at gAi_g$ (Impl.java:368:1)
    at Impl.java:78:1
```
Otherwise both tests work fine for me (I think)

JamesTKhan:
> When going back from the `NonPowerOfTwoTest` (left top corner) I get this exception. Not sure why
> 
> ```
> Uncaught Error: java.lang.RuntimeException: com.google.gwt.core.client.JavaScriptException: (TypeError) : Cannot read properties of undefined (reading 'dispose_1_g$')
>     at ioe_g$.Ane_g$ [as createError_0_g$] (Throwable.java:120:1)
>     at ioe_g$.Kne_g$ [as initializeBackingError_0_g$] (Throwable.java:112:1)
>     at ioe_g$.wne_g$ (Throwable.java:80:1)
>     at ioe_g$.aoe_g$ (Exception.java:37:1)
>     at new ioe_g$ (RuntimeException.java:37:1)
>     at Qm_g$.Rm_g$ [as execute_0_g$] (GwtApplication.java:252:1)
>     at AnimationSchedulerImplStandard.java:39:1
>     at dAi_g$ (Impl.java:309:1)
>     at gAi_g$ (Impl.java:368:1)
>     at Impl.java:78:1
> ```
> 
> Otherwise both tests work fine for me (I think)

Thanks for catching that. It was caused by an NPE on a left over ShaderProgram variable on Dispose that ended up not being needed for the test. Removed that variable and should be able to back out of the test via top left corner now.

SimonIT (approved):

mgsx-dev (approved):
It looks good to me, and since there are no breaking changes and it's an experimental feature, i think we could merge it as is. We might found few issues on edges cases, but those are hard to spot with a code review or missing tests.

mgsx-dev:
@JamesTKhan could you add a line in the CHANGES file? something like "WebGL backend added" before merging?

mgsx-dev:
NVM, i'll do it, Simon merged it already

mgsx-dev:
Done.

SimonIT:
Oops

PR #7042, opened 2022-12-06
[Gradle] Replace task creating with a 'tasks.register' to comply with 'Task Configuration Avoidance'

Bonbeart:
Replace immediate **task creating** with a lazy `tasks.register` to comply with [**Task Configuration Avoidance**](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html)

Related to #6920 

image


ice1000:
This should be done already by #7134

obigu:
Fixed in https://github.com/libgdx/libgdx/pull/7134, thanks!

PR #7043, opened 2022-12-08
[BackendAndroid] create Application inherit from ComponentActivity

MichaelOliveiraBx:
- Copy AndroidApplication to an ComponentAndroidApplication which inherit from ComponentActivity
- it permit to have an activity with lifecycle owner 
- some library need to have this functionality

**It enable the use of Jetpack Compose**


MichaelOliveiraBx:
I want to see with you there is some option
- keep like this, with lot of duplicate code
- have an common class and create 2 implementations, one inherit from Activity and one from ComponentActivity
- replace the AndroidApplication inheritance to ComponentActivity

MobiDevelop:
Why not just use an AndroidFragmentApplication and drop it into a ComponentActivity?

MichaelOliveiraBx:
@MobiDevelop That's definitely a good idea 💡 
Currently our architecture is in full compose and we have no fragments.
But it solves my problem, I will delete the Pull Request. 
Thanks for your feedback 👏

PR #7044, opened 2022-12-08
Clear list pointers of child before adding it to a JsonValue

dasisdormax:
This PR clears the "prev" and "next" pointers of a JsonValue when adding it as a child to another object.

I agree, this is unnecessary for newly created JsonValues. However, adding a previously existing object can (and for me did) cause confusing, hard to find bugs such as elements looping or appearing in duplicate locations.

See the newly added test for a short example. (Previously, this would cause "c" to be part of both objects)

This change might be considered breaking, if others depended on the behaviour to add multiple children in one step. However, this would have left the size and the other children's parent in an inconsistent state.


obigu:
@NathanSweet Can you have a look?

NathanSweet:
I think it makes sense, I don't see any reason to keep the old `next` and `prev` values. Sorry you had to suffer such a sneaky bug!

PR #7061, opened 2023-01-10
fix: make angle loader compatible with osx app bundles

### LINKED ISSUE #7058: [gdx-lwjgl3-angle] Library extraction not compatible with osx packaging
#### Issue details
The angle libraries are extracted in the current directory of the app, see [source code](https://github.com/libgdx/libgdx/blob/999291f7169bf463cee517d911499153cf8f2574/extensions/gdx-lwjgl3-angle/src/com/badlogic/gdx/backends/lwjgl3/angle/ANGLELoader.java#L207)

This causes issues with various package formats since it changes the overall package signature.
On OSX for example, the `*.app` package won't run anymore because of that.

There is no issue when running the plain jar file, but as soon as you're running from a package application everything breaks.

On a more general note it looks like a lot of the native library load code is duplicated from [gdx-jnigen](https://github.com/libgdx/gdx-jnigen/blob/master/gdx-jnigen-loader/src/main/java/com/badlogic/gdx/utils/SharedLibraryLoader.java).

Wouldn't it be possible to simply reuse it instead of re-inventing the wheel?

I noticed the same issue duplicate code issue in the [GlfwAWTLoader](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-lwjgl3-glfw-awt-macos/src/com/badlogic/gdx/backends/lwjgl3/awt/GlfwAWTLoader.java).

Merging the load code would also make sure that the folder/file structure for the native libraries is the same. Currently it's a big mix of everything: Subfolders per os/arch, files per os/arch.

#### Reproduction steps/code
1) Use the angle extension
2) Package the libgdx app into an `*.app` bundle
3) Sign all files
4) Notarize and stamp the bundle
5) Try to start the app

1.11.0

#### Stacktrace
```java
EGL: Library not found
```

- [X] macOS

davidgiga1993:
This changeset fixes the `ANGLELoader` not working correctly when being executed via an osx app bundle.

The following changes have been made:
1. Do not extract libraries if executed via app bundle (working directory is `/` or a `*.app/` folder)
2. Remove duplicate code by using the `SharedLibraryLoader`
3. Add documentation how to include angle in an app bundle

The changeset has been tested on:
- Windows 11 (x64)
- OSX 12.6 (arm64)
- Linux Mint 21.1 (x64)

closes #7058

### DIFF
--- a/extensions/gdx-lwjgl3-angle/src/com/badlogic/gdx/backends/lwjgl3/angle/ANGLELoader.java
+++ b/extensions/gdx-lwjgl3-angle/src/com/badlogic/gdx/backends/lwjgl3/angle/ANGLELoader.java
@@ -17,12 +17,11 @@
 package com.badlogic.gdx.backends.lwjgl3.angle;
 
 import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.SharedLibraryLoader;
 
-import java.io.*;
-import java.lang.reflect.Method;
-import java.util.Random;
-import java.util.UUID;
-import java.util.zip.CRC32;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
 
 public class ANGLELoader {
 	static public boolean isWindows = System.getProperty("os.name").contains("Windows");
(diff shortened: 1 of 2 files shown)


colorizenl:
@davidgiga1993 I'm running into the same issue when submitting to the Mac App Store, amazing work! I am however surprised you're not running into the same issue with `GlfwAWTLoader`, which has similar logic and also doesn't use `SharedLibraryLoader`. Is that something you also encountered while making these changes?

davidgiga1993:
> @davidgiga1993 I'm running into the same issue when submitting to the Mac App Store, amazing work! I am however surprised you're not running into the same issue with `GlfwAWTLoader`, which has similar logic and also doesn't use `SharedLibraryLoader`. Is that something you also encountered while making these changes?

The main issue with angle is that is dynamically loads additional libraries from within the native code. This is outside of the JVM and is handled by MacOS. In that case macos prevents the library from being loaded as it's outside of the dynamic library paths defined in the java binary (thus the change required in the JVM).

extensions/gdx-lwjgl3-angle/src/com/badlogic/gdx/backends/lwjgl3/angle/ANGLELoader.java
```diff
@@ -34,144 +33,21 @@ public class ANGLELoader {
...
-
-			return (Boolean)canExecute.invoke(file);
-		} catch (Exception ignored) {
-		}
-		return false;
+	/** Checks if the current jar/code location is inside an macOS app bundle
+	 *
+	 * @return True if app bundle */
+	static boolean isAppBundlePath () {
+		URL jarSource = ANGLELoader.class.getProtectionDomain().getCodeSource().getLocation();
```
Berstanio, line 43: This does not work reliable on AOT compiled images using native-image: https://github.com/oracle/graal/issues/10773

For the specific case, I would think some customisation option for the ANGLELoader might be preferable? A user can globally configure, how `ANGLELoader` should behave, depending on their project structure.
I did related work in https://github.com/libgdx/gdx-jnigen/pull/124, not 100% applicable tho. But maybe there is also a possibility to extend SharedLibraryLoader loading override proposed in https://github.com/libgdx/gdx-jnigen/pull/124, to cover also this use-case?

PokeMMO: Afaik angle is a special case as both natives need to be loaded at the same time. So they both have to be extracted before one is attempted to be loaded.

davidgiga1993: > This does not work reliable on AOT compiled images using native-image: [oracle/graal#10773](https://github.com/oracle/graal/issues/10773)
> 
> For the specific case, I would think some customisation option for the ANGLELoader might be preferable? A user can globally configure, how `ANGLELoader` should behave, depending on their project structure. I did related work in [libgdx/gdx-jnigen#124](https://github.com/libgdx/gdx-jnigen/pull/124), not 100% applicable tho. But maybe there is also a possibility to extend SharedLibraryLoader loading override proposed in [libgdx/gdx-jnigen#124](https://github.com/libgdx/gdx-jnigen/pull/124), to cover also this use-case?

This isn't really related to the project structure, but the distribution method. Given that the only official way to distribute on macOS is using ".app" bundles, the only way to get angle working is using the method I've outlined in the MR.

Happy to adjust the `isAppBundlePath` method for AOT binaries, just need a reliable way to get the path to the own binary (so we know if its being executed as an app bundle or not)

PR #7073, opened 2023-01-24
Improved on Skin serializer for BitmapFont

Betalord:
Added "integerPositions" boolean value which sets "useIntegerPositions" for BitmapFont. Also changed scaledSize from int to float as it offers more precision this way.


Betalord:
Oops, I forgot to change string value for json... Can I edit this PR somehow?

Line:
Boolean integerPositions = json.readValue("markupEnabled", Boolean.class, true, jsonData);

Should be:

Boolean integerPositions = json.readValue("integerPositions", Boolean.class, true, jsonData);

crykn (commented):
I'd change the name of `integerPositions` to the more descriptive `useIntegerPositions`, but apart from that those seem like two sensible changes.

crykn (approved):

PR #7074, opened 2023-01-25
Use setup-java v3 to manage gradle caches

fourlastor:
This PR removes the manual setup for caching gradle and delegates it to the [`setup-java` v3 action](https://github.com/actions/setup-java#caching-packages-dependencies).

I also added `--no-daemon` to the gradlew invocations as this is suggested for the caching (and it tends to run faster in gh actions, in my experience), alternatively, I could set it globally via [an environment variable](https://docs.gradle.org/current/userguide/gradle_daemon.html#disable_globally)

If you're happy with this, I can also cover the other workflows :)

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -19,20 +19,11 @@ jobs:
           submodules: 'recursive'
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
           java-version: '11'
-
-      - name: Cache Gradle packages
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


Berstanio:
Maybe we could use this plugin too?
https://github.com/gradle/gradle-build-action
Seems to have better caching and also takes care of daemons. It's in my opinion a bit clunky to add "--no-daemon" to all gradle invocations.

fourlastor:
:+1: makes sense, I have since started using the action myself, will do tomorrow!

.github/workflows/build-publish.yml
```diff
@@ -19,20 +19,14 @@ jobs:
           submodules: 'recursive'
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
           java-version: '11'
+          cache: 'gradle'
```
SimonIT, line 26: I'm not sure if we need caching here because the gradle-build-action takes care of caching now

SimonIT (approved):

SimonIT:
@libgdx/libgdx-signing

Tom-Ski:
Looks nice and clean! Thanks!

Tom-Ski (approved):

Issue #6503, opened 2021-04-19
Support layer tint color for TiledMap 


thcrock:
Tiled, in version 1.4, added an option called Tint Color ( https://doc.mapeditor.org/en/stable/manual/layers/#tinting-layers ). This allows mapmakers to add a color that is multiplied into all tiles in the layer. Supporting this would be useful in libgdx.

It is stored as an attribute on the layer, not as a property, so both the map loader and renderers need to be made aware of it. As far as I can tell, the implementation as it's done within the Tiled editor itself would be as easy as multiplying by this tint color in this line in each renderer: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthogonalTiledMapRenderer.java#L69

I haven't tested this out yet but that's the first thing I'm going to try. I plan on opening a PR when I get around to this but I wanted to open this issue to track it and give anybody the chance to weigh in before I do so.

tommyettinger:
It might make sense to change MapLayer's internal `opacity` float to be a full tint `Color`, using the Color's alpha as the actual opacity in the getter and setter. This would avoid an odd duplication of alpha variables, unless that duplication is also present in Tiled. It does seem useful to be able to tint layers.

(I don't currently use Tiled, so feel free to disregard any suggestions I make.)

thcrock:
The duplication seems to be present in Tiled. There is an opacity float and a tint color, and they appear to be combined with multiplication within the editor.

The original author of the tint color PR did propose replacing the (older) opacity option with tint, but that's not how it's implemented as of now.  You can read the original discussion if interested: https://github.com/mapeditor/tiled/issues/2683

SimonIT:
Implemented by #7076

PR #7076, opened 2023-01-26
Support for Layer TintColor in TiledMap

BoBIsHere86:
After coming across this issue https://github.com/libgdx/libgdx/issues/6503 I decided to try and give it a go.
This PR gives the ability for the Tint Color attribute which was added in Tiled 1.4 to be recognized.

As I was researching this I also found and implemented a fix for the following issue:
When a layer was in a group which had it's opacity set, that opacity attribute was being ignored as the layer's own opacity was not affected by the parent's. Which is NOT how it is rendered in tiled. 

Overview of PR Changes:
- Implemented TiledMap tintcolor attribute
- Fixed issue where opacity of parent group was ignored when rendering.
- Insure tintcolor and opacity attribute changes work in all of the following maptypes:
OrthogonalTiledMapRenderer, OrthoCachedTiledMapRenderer, IsometricTiledMapRenderer, IsometricStaggeredTiledMapRenderer, HexagonalTiledMapRenderer
- ImageLayers are rendered in a different method compared to normal TileLayers. So I made effort to include a test map with them.
- Created a new GDXTest, TiledMapLayerTintOpacityTest based on the TiledMapLayerOffsetTest as it was most similar. Created 6 separate test maps which showcase different opacity and tinting combinations.
- End result is that maps generated should match 1:1 with their tiled counterpart
- *also note: I fixed a comment where I found a typo in Color.class

One thing to mention specifically about the tintcolor attribute, that I learned.
I saw that we already have some code in place which converts tile's color format to a libGDX color. For custom properties which could be of type Color.
https://github.com/libgdx/libgdx/blob/9fb9353c99ae5b30b455262832c13a84c2736bd4/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java#L440

Tiled uses the format #AARRGGBB but for some reason when a tintcolor is selected and has an alpha of 255 (the default), tiled generates the color code without adding in the alpha for some reason. For example, a layer with a tint of Solid red, with an alpha of 255 generates: ` <layer id="1" name="Tile Layer 1" width="10" height="10" tintcolor="#ff0000">`
While Solid Red with an alpha of 254 generates: `<layer id="1" name="Tile Layer 1" width="10" height="10" tintcolor="#feff0000">`
#ff0000(alpha not included) vs #feff0000
But for the custom color property, that is not the case, it always passes the full color code.
Anyways, I specifically setup a check for it in tintcolor based on this possibility here:
https://github.com/BoBIsHere86/libgdx/blob/927528e9e22402c834fdf97c5ef6fc606327c486/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java#L315

Examples comparing tiled and the result in GDXtest:
OrthogonalTiledMapRenderer:
test1

OrthoCachedTiledMapRenderer:
(Note: the OrthoCached map does not seem to have support for groups, so I am using a different testmap then the one above)
test2

IsometricTiledMapRenderer:
test3

IsometricStaggeredTiledMapRenderer:
test4

HexagonalTiledMapRenderer:
test5

OrthogonalTiledMapRenderer (with ImageLayers):
Since I made a modification to all the renderImageLayer() methods  I thought it far to have a test show working ImageLayers
test6

This is my first large PR for libGDX, so please any comments, thoughts or complaints. Feel free to yell at me, it's the only way I'll learn.


BoBIsHere86:
Now that Quillraven's Parallax update is in place. Some modifications need to be made. But I will wait to see if the ImageLayer Repeat PR is accepted (https://github.com/libgdx/libgdx/pull/7080). As I will need to add tint functionality to those new areas of the renderer.

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
```diff
@@ -186,7 +186,8 @@ public void renderObject (MapObject object) {
 	@Override
 	public void renderImageLayer (TiledMapImageLayer layer) {
 		final Color batchColor = batch.getColor();
-		final float color = Color.toFloatBits(batchColor.r, batchColor.g, batchColor.b, batchColor.a * layer.getOpacity());
+		final float color = Color.toFloatBits(batchColor.r * layer.getTintColor().r, batchColor.g * layer.getTintColor().g,
```
tommyettinger, line 189: If my suggested change for MapLayer.getTintColor() was applied, these repeated lines could multiply by layer.parent.getTintColor().r , .g , .b, etc.

tommyettinger (changes requested):
This looks mostly good, but I don't think the parent's tint color should always be factored into any child layer's tint color. There could be a separate method that gets the combined parent + children (potentially multiple layers) tint color, but I think getTintColor() should be a normal getter, like setTintColor(). The parent color can be factored in at the usage sites, though this would only be one level deep, or the separate method(s) could call their separate counterparts for getting combined colors.

gdx/src/com/badlogic/gdx/maps/MapLayer.java
```diff
@@ -46,14 +49,32 @@ public void setName (String name) {
...
 	/** @param opacity new opacity for the layer */
 	public void setOpacity (float opacity) {
 		this.opacity = opacity;
 	}
 
+	/** @return layer's tint color */
+	public Color getTintColor () {
+		if (parent != null) {
+			tempColor.set(tintColor);
+			return tempColor.mul(parent.getTintColor());
```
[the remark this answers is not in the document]
BoBIsHere86: Thanks for taking the time to check this out. It's been a year and a half since I last looked, so I'm giving the code a deeper once over. And with some recent changes I think I'm going to also add in changes to the test maps as well since I should probably add in the Image Layer tinting on the other maps as well.

BUT in reference to your suggestions these all are great. I hope I am following you correctly.
So more in line with something like this for example:
```
	/** @return layer's tint color combined with the parent's tint color */
	public Color getCombinedTintColor(){
		 if (parent != null) {
			  tempColor.set(tintColor);
			  return tempColor.mul(parent.getCombinedTintColor()).cpy();
		 } else {
			  return tintColor.cpy();
		 }
	}

	/** @return layer's tint color */
	public Color getTintColor () {
			return tintColor.cpy();
	}

	/** @param tintColor new tint color for the layer */
	public void setTintColor (Color tintColor) {
		this.tintColor.set(tintColor);
	}

```
Since our worry would be people messing with the tintcolor. Would using Color's .cpy() method be an acceptable solution to lock it down? At most each layer's tint should only be called once per frame.
As for getting the combination I wasn't sure if getCombinedTintColor() would be an acceptable name, since getCombinedParentTintColor() or getCombinedTintWithParentColor(), might be a bit wordy?

Either way I'll gladly take any suggestions and get to work on this and some test map changes a bit later today.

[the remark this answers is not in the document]
tommyettinger: I think you'd be fine not cpy()-ing colors here. If it's understood that getTintColor() returns the direct Color being modified, that doesn't need to cpy(). I think since the combined color is always a temporary color, you could change it to:
```java
	/** @return layer's tint color combined with the parent's tint color */
	public Color getCombinedTintColor(){
		 if (parent != null) {
			  // This assumes there's a way to get the combined tint color on parent; you might use the parent's normal color if there's nothing it would combine with.
			  return tempColor.set(tintColor).mul(parent.getCombinedTintColor());
		 } else {
			  return tempColor.set(tintColor);
		 }
	}
```

I'd consider documenting getCombinedTintColor() as returning a temporary Color that should have its value set somewhere else, such as with `Color.set(Color)`, if you need to keep it, since it gets reused. I think it's a good name, it describes what it is. I'm just not sure we... even need to provide the combined color... map renderers could be doing some of this already, at least for alpha. In scene2d.ui, there's no way to get the combined color including parentAlpha; the parentAlpha is provided where it can change a color. Maybe the tint should just be considered by map renderers as part of what they process, but that might not be great with multiple layer tints.

BoBIsHere86:
So I spent a bit of time getting reacquainted with the Tiled Map code.
So just to reiterate. The purpose of this PR was to make sure we are 1 to 1 with the results we see in the TiledMap editor.
Specifically related to the Tint Color property.
For starters there was one oversight I had missed related to tinting image layers.
In TiledMap when you use the Tint Property on an Image layer it renders differently then if you did so on a normal TileLayer.

ex. Setting an ImageLayer tint to Yellow (255,255,0) with a decreasing alpha value from 255 to 0. You would get a darker and darker until it's fully black. As seen in the example image below.
img_alpha_ex
What this means is that you need opacity on your ImageLayer, you need to specifically use the Opacity property in the TiledMapEditor.

On a TileLayer however. The Alpha value of the Tint will actually affect the resulting opacity. And be affected by the opacity property of the layer itself and any parent layers. 

All this means is that when it came renderTileLayer() and  renderImageLayer() sections for rendering each map we had to handle it differently.

I have also modified all of the old tests to contain a progression of 4 ImageLayers with alpha tints from 255 to 0 and also added in a new test map with a ton of parent groups of varying tints opacity.

The representations of what you see in tiled, vs in the tests can be seen below.
ortho_img

Tints and Opacity with parent groups:
deep_group

One thing to remember is that we still have that "Parsing imagelayer inside a group Bug" pr open, so I can't put any image layers inside parent groups in the test maps. But tinting would work on them non the less once that is fixed.

Once again, feel free to scold me if any code seems a miss.

On a happy note. Once this and a few other hanging tiled prs get merged. We will probably have "almost" all of the basic features in TiledMap covered in libGDX. And once that's done we will never need to worry about it again! Until someone wants to refactor it...

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
```diff
@@ -186,7 +186,11 @@ public void renderObject (MapObject object) {
 	@Override
 	public void renderImageLayer (TiledMapImageLayer layer) {
 		final Color batchColor = batch.getColor();
-		final float color = Color.toFloatBits(batchColor.r, batchColor.g, batchColor.b, batchColor.a * layer.getOpacity());
+		final Color combinedTint = layer.getCombinedTintColor();
+		// For image layer rendering, multiply all by alpha except opacity
+		final float color = Color.toFloatBits(batchColor.r * (combinedTint.r * combinedTint.a),
```
tommyettinger, line 191: It looks like this is premultiplying alpha; if Tiled defaults to that or doesn't offer the option to change the option, this is probably good as-is. I don't think anything else in libGDX defaults to premultiplied alpha, but Tiled isn't in libGDX. And there's also the mostly-theoretical concern that none of the renderers are supposed to be Tiled-specific unless they mention they're for TMX maps... @raeleus ! You have been summoned! Since you have used other types of non-TMX maps with libGDX, would this change affect your usage in any undesirable way?

raeleus: The way I do level design is so far removed from TMX support that there is nothing you can do here that would affect my approach.
With that said, there is a basic assumption across the board that with any editor you are not going to have a 1:1 relationship between what you see in the editor and what the game is actually going to look like. I do appreciate the sentiment of this PR to bring rendering as close as possible. I don't think we should make technical choices like pre-multiplied the default.
Anyway, ignore me. I use busted old Ogmo and complain about it every time. I don't use Tiled and likely never will. Next thing for me is to learn Hyperlap as an alternative. Someday.

tommyettinger: Yeah, it sounds like you two are in agreement 6 minutes apart! Since PNGs act like libGDX normally does (without PMA), we probably shouldn't force it on, but having it as an option (like how particle effects have it) might make sense to make things act like how JPG acts in Tiled. I don't think Tiled will ever look exactly like a game will at runtime in all circumstances, but the tints can at least get closer.

BoBIsHere86: I made one final update to the get the functionality into parity with Tiled.
For what its worth, it now all works as it should. Image layers with jpgs, pngs, gifs. Since Tiled supports pngs and gifs for tint color transparency I simply check in the TiledMapImageLayer and then handle it in the renderer. Modified all the tests to have pngs now and all works and matches as it should. If it's too convoluted I could revert it all. But at least now it all matches.

Thanks for all the help and back and forth on these forsaken TiledMap Issues.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -351,6 +352,14 @@ protected void loadBasicLayerInfo (MapLayer layer, Element element) {
 		layer.setOffsetY(offsetY);
 		layer.setParallaxX(parallaxX);
 		layer.setParallaxY(parallaxY);
+
+		// tiled uses the format #AARRGGBB
+		// if the alpha of the tintcolor is set to 255, Tiled does not include it as part of the color code.
+		// ex. Red (r:255,g:0,b:0,a:255) becomes #ff0000, Red (r:255,g:0,b:0,a:127) becomes A:157 #7fff0000
+		String alpha = tintColor.length() == 9 ? tintColor.substring(1, 3) : "ff";
+		String color = tintColor.length() == 9 ? tintColor.substring(3) : tintColor.substring(1);
```
SimonIT, line 360: Maybe a new static function for `Color`? Might be useful for others, and if we need it ourselves, it might be nice to make it available to others

gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthoCachedTiledMapRenderer.java
```diff
@@ -209,7 +209,9 @@ public void renderObject (MapObject object) {
 
 	@Override
 	public void renderTileLayer (TiledMapTileLayer layer) {
-		final float color = Color.toFloatBits(1, 1, 1, layer.getOpacity());
+
+		final float color = Color.toFloatBits(1 * layer.getCombinedTintColor().r, 1 * layer.getCombinedTintColor().g,
+			1 * layer.getCombinedTintColor().b, layer.getOpacity() * layer.getCombinedTintColor().a);
```
SimonIT, line 214: Multiplying by 1 looks a bit odd to me...

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
```diff
@@ -186,7 +186,22 @@ public void renderObject (MapObject object) {
...
+		// and if the Image Layer does not support transparency, we want to multiply the combined tint values, by its alpha
+		float alphaMultiplier = supportsTransparency ? 1f : combinedTint.a;
+		// Only modify opacity by combinedTint.b if Image Layer supports transparency
+		float opacityMultiplier = supportsTransparency ? combinedTint.a : 1f;
+
+		// For image layer rendering multiply all by alpha
+		// except for opacity when image layer does not support transparency
+		final float color = Color.toFloatBits(batchColor.r * (combinedTint.r * alphaMultiplier),
+			batchColor.g * (combinedTint.g * alphaMultiplier), batchColor.b * (combinedTint.b * alphaMultiplier),
+			batchColor.a * (layer.getOpacity() * opacityMultiplier));
```
SimonIT, line 204: If I see it right, the same code is also in the other renderers. Maybe a new utility function would be nice to not have much code duplication

gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
```diff
@@ -109,7 +109,9 @@ private void init (TiledMap map) {
 	@Override
 	public void renderTileLayer (TiledMapTileLayer layer) {
 		final Color batchColor = batch.getColor();
-		final float color = Color.toFloatBits(batchColor.r, batchColor.g, batchColor.b, batchColor.a * layer.getOpacity());
+		final float color = Color.toFloatBits(batchColor.r * layer.getCombinedTintColor().r,
+			batchColor.g * layer.getCombinedTintColor().g, batchColor.b * layer.getCombinedTintColor().b,
+			batchColor.a * layer.getCombinedTintColor().a * layer.getOpacity());
```
SimonIT, line 114: This also looks duplicate to me, but I'm not sure if it's worth it

SimonIT (commented):
I didn't mean to have so many comments🥲

gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java
BoBIsHere86: Yep, you are totally correct. A lot of the tiledmap rendering code is pretty messy. Each individual map type has it's own separate rendering methods for TileLayers and ImageLayers, with a lot of duplicated code paired with minor differences within.
When I first started working on this long ago. I didn't want to rock the boat and make any big changes or refactoring, adding utility classes, etc. For just simple color tinting feature request. So I tried my best to just go along with how it was done at the time.

That being said I will now gladly take another look at this and pair it down. 

Thanks!

gdx/src/com/badlogic/gdx/maps/tiled/TiledMapImageLayer.java
```diff
@@ -27,13 +27,29 @@ public class TiledMapImageLayer extends MapLayer {
...
+	}
+
+	/** TiledMap ImageLayers can support transparency through tint color if the image provided is in PNG or GIF format. Here we
+	 * check to see if the file type supports transparency
+	 *
+	 * @param region TextureRegion of the ImageLayer
+	 * @return boolean */
+	private boolean checkTransparencySupport (TextureRegion region) {
+		String texturePath = region.getTexture().toString().toLowerCase();
+		return texturePath.endsWith(".png") || texturePath.endsWith(".gif");
```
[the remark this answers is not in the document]
BoBIsHere86: I'll take another look, this was a fairly quick change when I learned the differences in in how Tiled handles jpgs vs, png, gif.

I appreciate you taking the time to look this over. I'll look through the responses, give it another go and try to make the libGDX team proud.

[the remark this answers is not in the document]
BoBIsHere86: I wont push any changes until I feel comfortable. But I think I have a MUCH cleaner solution thanks to your comment.
I'd replace the current check in TiledMapImageLayer with something like this
```
/** TiledMap ImageLayers can support transparency through tint color if the image provided
	 *  supports the proper pixel format.
	 *  Here we check to see if the file supports transparency by checking the format of the TextureData.
	 *
	 * @param region TextureRegion of the ImageLayer
	 * @return boolean */
	private boolean checkTransparencySupport (TextureRegion region) {	
		 Pixmap.Format format = region.getTexture().getTextureData().getFormat();	
		 return format != null && formatHasAlpha(format);
	}

	 //Check if pixel format supports alpha channel
	 private boolean formatHasAlpha(Pixmap.Format format) {
		  switch (format) {
		  case Alpha:
		  case LuminanceAlpha:
		  case RGBA4444:
		  case RGBA8888:
				return true;
		  default:
				return false;
		  }
	 }
```

I think this would cover any and all cases where an image could possible be of a format containing alpha values we support.

From my testing I've made sure it works in parity visually with Tiled using JPGs, PNGs, PNGs that don't have alpha support(RGB888 format), GIFs, and Greyscale PNGs that support transparency (LuminanceAlpha).

gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
BoBIsHere86: Yes, agreed. So there's a lot of bits of repeated code throughout all the map code code. This specifically is repeated in every single map renderer except for the OrthoCachedTiledMapRenderer.
I will be making a small change in the BatchTiledMapRenderer class which all the map renderers inherit from (except orthocached)
Which should clean this specific area and the area related to the image layer transparency stuff you commented on as well. So we don't have large chunks of duplicate code pasted across 5 renderers.

gdx/src/com/badlogic/gdx/maps/tiled/renderers/OrthoCachedTiledMapRenderer.java
BoBIsHere86: Yep, this does look odd. More of an oversight on my part since the OrthoCachedTiledMapRenderer is the odd man out when it comes to all the Map Renderers. It originally passed in all 1s as it's the only renderer which didn't use the spritebatch color.
This will be cleaned up, when I push the next update.

gdx/src/com/badlogic/gdx/maps/tiled/TiledMapImageLayer.java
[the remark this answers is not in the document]
tommyettinger: Nice! I wonder if after this PR we could look at adding some `hasAlpha()` to Pixmap.Format, since this probably has uses elsewhere.

BoBIsHere86:
This PR has been updated with the latest tiledmap codebase changes that are in the master so there should be no conflict when it comes time to merge. Also, I think all changes to relevant comments and requests have been made. 🚀

tommyettinger (approved):
Looks good to me! This looks like it was a lot of work, 23 changed files...

SimonIT (approved):
Looks very clean

PR #7080, opened 2023-02-01
Support for Image Layer repeat x and y in TiledMap + Bugfix

BoBIsHere86:
This PR will accomplished two things.
1. Add the ability for Image Layers to be repeated along the X and/or Y axis to support the repeatx and repeaty attributes on a layer in Tiled.
2. Fix a Glaring position issue that has always existed when using ImageLayers in any non-orthograpic map renderer. (That I only learned about while working on this).
3. Added a new GDXTest specifically to show off this feature.

First I would like to point out a bug that seems to have always existed. As I learned while working on this PR. Any imagelayer that was added to a tiled map would render at the incorrect position in the following MapRenderers:
IsometricStaggeredTiledMapRenderer
IsometricTiledMapRenderer
HexagonalTiledMapRenderer (staggered X and/or Y)

This was because they were all using the same BatchTiledMapRenderer.renderImageLayer() which works fine for the OrthogonalTiledMapRenderer but not the rest.
While fixing this I needed to give each renderer their own specific renderImageLayer() to fit it's needs.

Using the base examples I modified with an Image Layer found in GDXTests:
Pictures comparing Tiled, with how they render in libGDX.
And what I think to be the reasoning behind the issue. Please correct me if I am mistaken.
IsometricTiledMapRenderer:
iml_Issue_Iso
So, Because of the way libGDX handles the isometric coordinates. The leftmost tile of the map begins rendering at world position 0,0, while in Tiled the y position is actually totalHeight/2
ex. Map 800px in height, leftmost tile edge starts rendering at 0,400 in tiled
To compensate for that we need to subtract half the map's height in pixels then add half of the tile's height in order to position it properly in order to get a 1 to 1 rendering as to where the imagelayer renders in tiled. 
https://github.com/BoBIsHere86/libgdx/blob/61b2384796a471d8af54eb1849861c1994f8588b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricTiledMapRenderer.java#L262

IsometricStaggeredTiledMapRenderer:
iml_Issue_IsoStag
Must offset imagelayer x position by half of tileWidth to match position in  tiled.
https://github.com/BoBIsHere86/libgdx/blob/61b2384796a471d8af54eb1849861c1994f8588b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/IsometricStaggeredTiledMapRenderer.java#L211

HexagonalTiledMapRenderer Stagger X Axis:
iml_Issue_Hex_X
If X axis is the staggered axis, must offset imagelayer y position by adding half of tileHeight to match position
https://github.com/BoBIsHere86/libgdx/blob/61b2384796a471d8af54eb1849861c1994f8588b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java#L290

HexagonalTiledMapRenderer  Stagger Y Axis:
iml_Issue_hex_y
This one was a bit complicated to figure out but ImageLayer's y position seems to be placed at an offset determined the total height if this were a normal tile map minus the height as calculated for a hexmap. We get this number and use it to counter offset our Y. Then we will have our imagelayer matching its position in Tiled. 
https://github.com/BoBIsHere86/libgdx/blob/61b2384796a471d8af54eb1849861c1994f8588b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java#L295

Anyway, once that issue was out of the way. Now I could get to the ACTUAL PR, lol.

The basic idea behind the tiling feature was to make it match 1:1 with how Tiled renders it.
Repeat X, Tiles horizontally. Repeat Y Tiles vertically. But if we have both select. It tiles all directions infinitely and not just Up/Down.

I have an example of the results in a new TiledMapImageLayerRepeatTest test I made, each with it's own map based on the maps we seem to always use.

OrthogonalTiledMapRenderer:
imagelayer_RepeatOrtho

OrthoCachedTiledMapRenderer:
imagelayer_RepeatOrthoCache
This I found to be a bit tricky, but I based it on the bounds cache. So it will render as many tiles are needed to fit within the bounds of the cache. Then when it gets to the point where the cache is invalided it get's rendered. But you will never notice this, unlike in the test. where we can see outside of the bounds, which is why this looks different than the OrthogonalTiledMapRenderer Test.

IsometricTiledMapRenderer:
iso

IsometricStaggeredTiledMapRenderer:
isoStag

HexagonalTiledMapRenderer staggered X:
hex_x

HexagonalTiledMapRenderer staggered Y:
hex_y

As for how I actually implemented this tiling function:
https://github.com/BoBIsHere86/libgdx/blob/61b2384796a471d8af54eb1849861c1994f8588b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/BatchTiledMapRenderer.java#L243

And this is replicated in each of the specific renders.

I left a giant comment in there explaining the my thought process. I may remove it. But I wasn't sure what the protocol was, as I thought it would help if someone was looking at it. I could just remove it.

Any thoughts or complaints are welcome.

Also, While I was writing up this PR it seems like the Parallax feature was reviewed and accepted. I will need to add and do some testing to make sure the Parallax effect will work on image layers and if they tile properly with it enabled. But as of right now the parallax attribute is not taken into account.


BoBIsHere86:
I've tested Quillravens parallax code in a separate merged local branch I have.
And the effect seems to work without issue. Though it might be hard to tell with this low fps gif I quickly made.
imageLayerRepeatParallaxOrtho

This PR will have to wait until https://github.com/libgdx/libgdx/pull/6714 get's merged. As I will need to modify the new renderImage Methods to support it as this branch does not contain any of his parallax code.

BoBIsHere86:
Now that Quillraven's Parallax code has been merged. I rebased against the master to pull in his changes. 
And I modified the renderImageLayer() functions that I added while doing this PR to make sure they support the parallax effect as well.
The only last hanging thread would be if my Tinting Feature PR were to be accepted as I completed that first.(https://github.com/libgdx/libgdx/pull/7076). I would need to update that PR, if this one is accepted and merged. To add tinting functionality to the new image renderers. 

Other than that there is no more work for me to do on this PR, unless someone finds and issue. 
Let me know. Cheers!

BoBIsHere86:
I felt my original code for tiling the images was a bit overly complicated and hard to understand. I realized it could be done a lot simpler so I changed it, retested and removed a giant ugly comment I had placed in there.
That is all.

tommyettinger:
I see a bunch of Tiled PRs here but I don't think anyone in the team who can review... uses Tiled. I can only just barely use Tiled well enough to replicate/verify bugs; I don't use it day-to-day in any of my code. There's been some talk of splitting the Tiled maps package into its own repo, though that would have to be after the next release. So uh... sorry these PRs are taking so long to get merged; you clearly put a lot of thought and work into the descriptions!

BoBIsHere86:
Thanks for the update. No rush on my end, libGDX is an awesome framework and I'm just doing my part to give back to this awesome community when I can.

PR #7096, opened 2023-02-20
Fix : RenderBuffer leak in GLFrameBuffer class

dar-dev:
RenderBuffer leak : depth and stencil render buffers are never released. 

Destroy any created render buffer. According to openGL specs it's possible to do so without even checking if the buffer has been created

See: #7094 

--


MikOfClassX:
this is a very important fix. Is there anything wrong, stopping a merge into GDX master ?

mgsx-dev:
Looking to the original issue, it looks like a misuse of the API, i think it would be better to throw an exception when user add non compatible/duplicate buffers instead.

BTW, you don't have to create an issue and a PR for the same topic, simply a PR is enough.

MikOfClassX:
PR came after issue.
Api can be misused anyway, so this fix is welcome.

 No bother, we keep this fix internally. It will not hurt anyone.

MikOfClassX:
> 

on that now with @dar-dev

Tom-Ski (approved):

Issue #7100, opened 2023-03-07
BitmapFont.data.setScale() is inherently flawed


Betalord:
After encountering a weird bug that can be see in the attached video (SelectBox enabling vertical scrollbar on first drop down, but not on subsequent drop downs: [https://www.youtube.com/watch?v=PJNcYH4d9gk](url)), I debugged my app and came to find out that the problem is in BitmapFontData.setScale() method.

The problem is with float calculations. Float has a limited precision, and when multiplying it and then dividing it, we risk to change the original value. Let me illustrate this.

In my example, my font has a capHeight of 28 originally. I scale it by factor 0.46, so it becomes **12.88**. Then some label scales it by factor 0.414 (in getPrefHeight(), which calls scaleAndComputePrefSize() and sizes the font for a moment to calculate width and height and then scales it back to original value) so capHeight becomes 11.592. It then scales it back by factor 0.46, and the new capHeight value becomes **12.880001**.
Here we can see the problem: when we first scaled it, the new capHeight became 12.88, the second time around becomes 12.880001 due to float precision limitations when mupltiplying and dividing floats. To make things short: after a few scalings, the capHeight values stabilizes at **12.880003**. Now this seems a small problem, but is not, since the values is not what it originally was and in some widgets that causes problems, like in my case in SelectBox, which does some check "if 12.88 >= 12.880003" which doesn't hold anymore and hence it adds a vertical scrollbar which is not required at all.

setScale() method's code should be fixed. I can make a PR if someone can confirm my issue. My solution would be to store original font attributes in e.g.:
float originalCapHeight
float originalAscent
float originalLineHeight
etc.
and then scale each time based on those original values, not the scaled values themselves. This is how setScale() would create same values each time we would call it.

#### Version of libGDX and/or relevant dependencies
1.11.0 (the latest, and older ones too)

Betalord:
Created a PR that works well (I tested it in my project):
https://github.com/libgdx/libgdx/pull/7102

PR #7102, opened 2023-03-11
Fix for font scaling problem (https://github.com/libgdx/libgdx/issues/7100)

Betalord:
Fix for font scaling problem (as suggested by 'evilentity' on Discord). Fixes a problem with limited float precision calculations which can cause some undesirable behaviour in certain widgets, as described here: https://github.com/libgdx/libgdx/issues/7100

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -983,12 +983,10 @@ public class BitmapFont implements Disposable {
 			setScale(scaleX + amount, scaleY + amount);
 		}
 
-		/**
-		 * Undo last {@link #setScale(float, float)} operation. We need this to solve problem with limited float precision
+		/** Undo last {@link #setScale(float, float)} operation. We need this to solve problem with limited float precision
 		 * calculations which can cause some undesirable behaviour in certain widgets, as described here:
-		 * <a href="https://github.com/libgdx/libgdx/issues/7100">https://github.com/libgdx/libgdx/issues/7100</a>
-		 */
-		public void undoScale() {
+		 * <a href="https://github.com/libgdx/libgdx/issues/7100">https://github.com/libgdx/libgdx/issues/7100</a> */
+		public void undoScale () {
 			scaleX = backupScaleX;
 			scaleY = backupScaleY;
 
@@ -1005,7 +1003,7 @@ public class BitmapFont implements Disposable {
 			padBottom = backupPadBottom;
 		}
 
-		public String toString() {
+		public String toString () {
 			return name != null ? name : super.toString();
 		}
 	}
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -459,6 +459,24 @@ public class BitmapFont implements Disposable {
 		 * file, it needs to be set manually depending on how the glyphs are rendered on the backing textures. */
 		public float cursorX;
 
+		/*
+		 * Here we keep backup values for those attributes that get modified in setScale() method. We need it due to a bug described
+		 * here: https://github.com/libgdx/libgdx/issues/7100. We use them with undoScale() method.
+		 */
+		public float backupScaleX;
+		public float backupScaleY;
+		public float backupLineHeight;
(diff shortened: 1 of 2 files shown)


tommyettinger:
If I'm reading this right, this will work if and only if setScale() has been called once before a call to undoScale(). If setScale() is called twice, this can only undo the most recent scaling. Other ways to handle this might involve treating prior states as `FloatArray`s to store stacks of changes, and popping off changes when they are undone. Or, and I'm not sure how viable this is here, maybe you can store one original size and make scaling relative to that original size -- this makes returning to 100% scale always accurate, since we can just reset to the original size.

Betalord:
> store stacks of changes, and popping off changes when they are undone. Or, and I'm not sure how viable this is here, maybe you can store one original size and make scaling relative to that original size -- this makes returning to 100% scale always accurate, since we can just reset to the original size.

Yes you are reading it right. I implemented the solution with storing original values and always scaling based on them, but it didn't work well (caused some other problems). The thing is, one can modify certain values manually after load() has been invoked without passing for setScale(), so it is impossible to keep track of these changes (since class doesn't provide getters/setters, as it should). I thought a lot about that solution and it isn't a good one (others on Discord confirmed it).
As far as stacking the values, I originally thought about that solution too (as I am using it throughout my project for other scenarios), however I think the way I did it is better. Because it can cause problems if some caller doesn't call undo() at the end, and if some nested stacking was going on etc. (there could be problems). The solution I presented is a very robust one - whoever wants to undo the scaling, must first call setScale() and then undoScale(). So it is completely up to caller. If someone wants to do it the old way, he is free to do it, but it doesn't work well. Currently the Label class is the only one in the library that actually uses it, so fixing the Label fixes everything as far as libGDX is concerned, other calls are up to the user.

tommyettinger:
> The solution I presented is a very robust one
I don't entirely agree; this is more robust against the accumulation of floating-point error, but as an API it isn't robust at all. For example, if a user calls undoScale() first, perhaps intended as a precaution against using a BitmapFont that could be scaled by another caller, then if setScale() hadn't already been called, the call to undoScale() sets every variable to 0. If setScale() is then called after that, you get 0 scale for x and y despite setScale() normally throwing an Exception when that could happen, and all metrics become infinite.

> fixing the Label fixes everything as far as libGDX is concerned
It doesn't, because this is now a subtle breaking change that Label users have to be aware of starting in the next release. Any user code that used setScale() multiple times in 1.11.0 on a Label would need to be changed to work with 1.11.1 (or whatever the next release is), and the change would be backwards-incompatible because undoScale() doesn't exist in 1.11.0 . I could be understanding this wrong, though.

While I agree about getters and setters, much of libGDX does not use them. I think there are a number of alternative approaches to the FP precision issue, though I don't think any are a "magic bullet" solution. Internally using fixed-point numbers would solve this problem, and potentially more, but would require getters and setters to wrap the fixnums in a usable number format. Fixnums would also probably be slower, though I doubt that matters much for UI code.

I'm hoping @NathanSweet has an opinion on this, since he wrote most of BitmapFont, I believe.

NathanSweet:
I don't actually use BitmapFontData `setScale` or Label `setScale` in any of my projects. I use a font instance per size that I want, which is generally only a few sizes. If I actually want to scale, like for dynamic effects, then I use Actor scale. I've not needed to use it yet, but if I wanted dynamic and high quality scaling then I would use DistanceFieldFont.

It's not great that this PR adds many fields that are useful only in a narrow use case, and a method to revert the last call to `setScale` makes the API a little odd. Also, if we were to have these "backup" fields, they should be used in `setScale` instead of reverting the scale with eg `float x = scaleX / this.scaleX;` before applying the new scale.

Another solution could be a method that returns an object (maybe populates an instance passed in) that has all the fields that are changed by `setScale`. You can keep that object and call a method on it to revert the scale to whatever it was when that object was populated.

I don't think Label's scaling and unscaling is a great solution, given the numerical inaccuracy that round trip can cause. Why are people scaling a font that way at all? We do have the functionality in Label though, so to avoid inaccuracy it could use locals to store the values, call `setScale`, draw, then set them back. While it is clunky, other code could do the same, though I still don't think it's a great idea to scale the BitmapFontData fields in this way at all.

NathanSweet:
Any code scaling a font and then trying to go back to how it was by scaling it the opposite amount has the problem. That should be avoided entirely. Label is just the only libgdx code that does it. Subtracting a small value doesn't fix the issue.

Issue #6882, opened 2022-05-30
isPointInTriangle returning true when it shouldn't


tylerhasman:
#### Issue details

[Intersector#isPointInTriangle](https://github.com/libgdx/libgdx/blob/8ead46fa94c9467138dc9edf0f0a2c04831fcb99/gdx/src/com/badlogic/gdx/math/Intersector.java#L48) fails for points lying on the plane of the triangle but not within it due to a floating-point rounding error in some cases.

The last check if the function:

`if (ab * bc - ac * bb < 0) return false;`

is sometimes having its arithmetic being rounded to 0.0 causing the function to return true even though the answer should be false.

#### Reproduction steps/code

```
Intersector.isPointInTriangle(
                new Vector3( -5120.8345f,8946.126f,-3270.5813f),

                new Vector3(50.008057f, 22.20586f, 124.62208f),

                new Vector3( 62.282288f, 22.205864f, 109.665924f),

                new Vector3(70.92052f, 7.205861f, 115.437805f))
```
will return true even though the point is clearly not in the triangle.

#### Version of libGDX and/or relevant dependencies
Tested on version 1.10.0 however should still be failing on all versions

tylerhasman:
I've created a more accurate function which can I've confirmed works for larger numbers:
```

    public static boolean SAFE_isPointInTriangle(Vector3 point, Vector3 t1, Vector3 t2, Vector3 t3){

        Vector3 a = t1.cpy();
        Vector3 b = t2.cpy();
        Vector3 c = t3.cpy();

        a.sub(point);
        b.sub(point);
        c.sub(point);

        Vector3 u = b.cpy().crs(c);
        Vector3 v = c.cpy().crs(a);
        Vector3 w = a.cpy().crs(b);

        if(u.dot(v) < 0f){
            return false;
        }

        if(u.dot(w) < 0f){
            return false;
        }

        return true;
    }

```

tommyettinger:
Nice, I can confirm it works for the original input as well. Your code allocates 6 Vector3 instances on each call, which isn't terrible but is avoidable. Here's a rearrangement of your code that I believe uses the same exact algorithm, but reuses the three temporary Vector3 instances that Intersector has for this purpose.

```java
    // These are declared in Intersector, and can be used by any code put into that class.
    private final static Vector3 v0 = new Vector3();
    private final static Vector3 v1 = new Vector3();
    private final static Vector3 v2 = new Vector3();

[...]

PR #7106, opened 2023-03-17
Fix isPointInTriangle in 3d (#6882)

HydrolienF:
Solution have been find by @tylerhasman & @tommyettinger but never merged. Closes libgdx/libgdx#6882

I've add fiew more tests.

I think I's usefull to warn user that because of float use, function may return a bad answer with realy small and realy big x,y,z. But if you think it's not usefull, I can revert comment of isPointInTriangle().

Let me know if anything else should be changed.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -40,11 +40,10 @@ public final class Intersector {
 	private final static FloatArray floatArray = new FloatArray();
 	private final static FloatArray floatArray2 = new FloatArray();
 
-	/** Returns whether the given point is inside the triangle. This assumes that the point is on the plane of the triangle.
-	 * No check is performed that this is the case.
-	 * 
-	 * If Vector have points with really small and realy big values (over 10^7 between the 2)
-	 * it' won't work! (because of Vector3 that use float)
+	/** Returns whether the given point is inside the triangle. This assumes that the point is on the plane of the triangle. No
+	 * check is performed that this is the case. <br>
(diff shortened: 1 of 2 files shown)


tommyettinger:
Thanks for revisiting this; I had forgotten about it. I'll run the Spotless tasks on this myself, since that's probably why the checks are failing, then commit to your PR (which I think should work). I'll also fix any typos I find.

tommyettinger (approved):
I didn't change any code; I just ran spotlessApply and the tests, as well as cleaned up some typos, unclear phrasing, and punctuation mistakes in comments. The expanded tests all pass, and I believe the algorithm has been known to work since the original issue. I don't see any reason this can't be merged.

HydrolienF:
@tommyettinger thanks for formating & english check !
Float warning is easier to understand that way.

I thought formating was done by a Github action. When I had run it on my computer (Linux) it edit all .java file. It probably only edit end line, but I was afraid that my merge request show change in every .java file.

tommyettinger:
Formatting is run by a GH Action, but to avoid the potential for spammers creating many PRs and running problematic actions on GitHub's servers, no actions will run for PRs by first-time contributors (just to be sure that it's running actions on PRs from someone it knows). You can run spotlessApply and it will claim all files have changes, but when you actually submit the PR only a few files will have actually been altered (the ones you changed by hand, usually). As you said, it is probably just changing the line endings, which Git will usually ignore when it pushes the changes.

SimonIT:
The automatic spotless fix runs on your own repository. If you don't have it enabled for the fork, it doesn't. It wasn't possible to push from the PR build

HydrolienF:
Do I need to do something else before merge ?

PR #7108, opened 2023-03-23
FIX getTextInput does nothing

deedywu:
text field was fixed with swing on lwjgl3


deedywu:
then it can work on m1 osx and x86 windows11
5ae96484548713148755a8e19e7c1ebb

PR #7110, opened 2023-03-24
FIX getTextInput does nothing (DefaultLwjgl3Input.java-getTextInput)

deedywu:
text field was fixed with swing on lwjgl3
then it can work on m1 osx and x86 windows11

5ae96484548713148755a8e19e7c1ebb

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -36,7 +36,9 @@ import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 
 import javax.swing.JButton;
 import javax.swing.JComponent;
@@ -385,12 +387,26 @@ public class DefaultLwjgl3Input extends AbstractInput implements Lwjgl3Input {
 
 	/*** input number filter */
 	private static class PlainDocumentNumber extends PlainDocument {
+
+		private final Set<Character> validChars = new HashSet<>();
+
+		public PlainDocumentNumber () {
+			for (char c : "0123456789.-".toCharArray()) {
+				validChars.add(c);
+			}
+		}
+
 		@Override
 		public void insertString (int offset, String s, AttributeSet attrSet) throws BadLocationException {
 			if (null != s) {
 				char[] chars = s.toCharArray();
-				for (char c : chars) {
-					if (c < 0x30 || c > 0x39) // use ASCII check chars
+				int point = 0;
+				for (int i = 0; i < chars.length; i++) {
+					char c = chars[i];
+					if (!validChars.contains(c)) return;
+					if (c == '-' && i != 0) // the negative sign can only be placed at the front.
+						return;
+					if (c == '.' && ++point == 2) // there is at most one decimal point symbol "." in a number.
 						return;
 				}
 			}
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
@@ -29,6 +29,25 @@ import com.badlogic.gdx.Input;
 import com.badlogic.gdx.InputEventQueue;
 import com.badlogic.gdx.InputProcessor;
 
+import java.awt.BorderLayout;
+import java.awt.Container;
+import java.awt.event.KeyEvent;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import java.util.Locale;
+
+import javax.swing.JButton;
+import javax.swing.JComponent;
+import javax.swing.JFormattedTextField;
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+import javax.swing.JPasswordField;
+import javax.swing.JTextField;
+import javax.swing.KeyStroke;
+import javax.swing.text.AttributeSet;
+import javax.swing.text.BadLocationException;
+import javax.swing.text.PlainDocument;
+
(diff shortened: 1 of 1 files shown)


tommyettinger:
Can someone with a Mac test this, or tell me if using Swing in LWJGL3 is still a no-go?

Frosty-J:
This looks similar to gdx-dialogs, which works well with LWJGL3 (can cause crashing on Linux if you invoke it before libGDX initialises, but is always fine if after).

deedywu:
> Can someone with a Mac test this, or tell me if using Swing in LWJGL3 is still a no-go?

YES,I built a game on m1pro mac used libgdx(mac of ARM support only lwjgl3),there are worked good

deedywu:
It’s been two weeks since I made this PR. I am wondering why it is not being reviewed yet. Can someone have a look?

intrigus (commented):
I really don't understand why this works.
My impression was always that Swing/AWT+LWJGL3 on macOS is a big no-no.

Is there a reason why you didn't copy the [implementation of LWJGL2](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/DefaultLwjglInput.java#L103-L194)?

Tom-Ski:
Is this tested without using the awt extension support?

deedywu:
> I really don't understand why this works. My impression was always that Swing/AWT+LWJGL3 on macOS is a big no-no.
> 
> Is there a reason why you didn't copy the [implementation of LWJGL2](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/DefaultLwjglInput.java#L103-L194)?

I am developing a small mobile game, but I need to debug it on the arm's Mac, which only supports building based on lwjgl3. Therefore, I have done the above work

deedywu:
> Is this tested without using the awt extension support?

If there is no support for awt components, it may not be possible to listen to keyboard events properly, but input can still be done

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
```diff
@@ -301,9 +323,79 @@ public void getTextInput (TextInputListener listener, String title, String text,
...
+					}
+				});
+			}
+		});
+	}
+
+	/*** input number filter */
+	private static class PlainDocumentNumber extends PlainDocument {
+		@Override
+		public void insertString (int offset, String s, AttributeSet attrSet) throws BadLocationException {
```
intrigus, line 389: Wouldn't this implementation disallow inputting negative numbers?

tommyettinger: As well as floating-point numbers, because of the `'.'`

deedywu: Haha, this is indeed a bug, because my program is more restrictive, sorry, I will fix it for general purpose now

Frosty-J:
I've been thinking about this method today and yesterday. I did an AWT implementation. This gives you a native input field with the ability to access the context menu, but no colour emoji support. Otherwise, it isn't too dissimilar to Swing - neither option is wildly appealing. I haven't tested multiple OSes. TinyFD can alternatively be used for this purpose but it offers zero customisation. I'm sharing this because concepts such as setting window icon or position can apply to either.

If you wish to test the window icon, add `winConfig.setWindowIcon("data/bobargb8888-32x32.png")` or the path of your choice into `Lwjgl3TestStarter` (at the part that creates a new window - if you add it to the application part, libGDX crashes upon launching a test). Then you can run `TextInputDialogTest`. Remember to change `"cancled by user"` to `"canceled by user"`!

I didn't want to stall the game thread but nor am I satisfied with being able to click multiple times to have multiple dialogues open at once (doing that with Swing caused rendering bugs). Maybe there's a way of disabling inputs without tampering with the user's input processor.

Unfortunately, this is one of those things where I feel any serious game would have to implement its own method of input. The libGDX API doesn't even provide a means of customising the button labels (e.g. if your game is in French, you'd probably want French buttons).

```java
@Override
public void getTextInput (TextInputListener listener, String title, String text, String hint, OnscreenKeyboardType type) {
	Thread t = new Thread(() -> {
		// Or a null frame can be used to display in taskbar
		Frame parent = new Frame();
		Dialog dialog = new Dialog(parent, title, true);

		// Works quite well (on Windows host) but some characters (e.g. Korean) look poor.
		// Inserting emoji causes font to fall back to something useless.
		dialog.setFont(new Font(Font.DIALOG, Font.PLAIN, 12));

		// If game closes, close dialog with it
		Gdx.app.postRunnable(() -> Gdx.app.addLifecycleListener(new LifecycleListener() {
			@Override public void pause() {}
			@Override public void resume() {}

			@Override
			public void dispose() {
				dialog.dispose();
				parent.dispose();
			}
		}));

		dialog.add(new Label(hint), BorderLayout.NORTH);

		TextField textField = new TextField(text, 30);
		// Demonstration of hiding input
		if (type == OnscreenKeyboardType.Password) {
			textField.setEchoChar('•');
		}

		Panel textPanel = new Panel();
		textPanel.add(textField);
		dialog.add(textPanel, BorderLayout.CENTER);

		Button okButton = new Button("OK");
		okButton.addActionListener(e -> {
			dialog.dispose();
			parent.dispose();
			Gdx.app.postRunnable(() -> listener.input(textField.getText()));
		});

		Button cancelButton = new Button("Cancel");
		cancelButton.addActionListener(e -> {
			dialog.dispose();
			parent.dispose();
			Gdx.app.postRunnable(listener::canceled);
		});

		dialog.addWindowListener(new WindowAdapter() {
			@Override
			public void windowClosing(WindowEvent e) {
				dialog.dispose();
				parent.dispose();
				Gdx.app.postRunnable(listener::canceled);
			}
		});

		Panel buttonPanel = new Panel();
		if (UIUtils.isWindows) { // Following convention
			buttonPanel.add(okButton);
			buttonPanel.add(cancelButton);
		} else {
			buttonPanel.add(cancelButton);
			buttonPanel.add(okButton);
		}
		dialog.add(buttonPanel, BorderLayout.SOUTH);

		// Inherit window icons
		if (window.getConfig().windowIconPaths != null) {
			List<Image> icons = new ArrayList<>();
			for (String path : window.getConfig().windowIconPaths) {
				try {
					FileHandle fileHandle =
						Gdx.files.getFileHandle(path, window.getConfig().windowIconFileType);
					icons.add(ImageIO.read(fileHandle.read()));
				} catch (IOException ignored) {
					// If we land here, no icon or default Java icon will be displayed
				}
			}
			dialog.setIconImages(icons);
		}

		dialog.pack();
		// Center over game window
		dialog.setLocation(
			window.getPositionX() + (Gdx.graphics.getWidth() - dialog.getWidth()) / 2,
			window.getPositionY() + (Gdx.graphics.getHeight() - dialog.getHeight()) / 2
		);

		dialog.setResizable(false);
		dialog.setAlwaysOnTop(true);
		dialog.setVisible(true);
	});
	t.setDaemon(true);
	t.start();
}
```

Berstanio:
> Unfortunately, this is one of those things where I feel any serious game would have to implement its own method of input. The libGDX API doesn't even provide a means of customising the button labels (e.g. if your game is in French, you'd probably want French buttons).

I agree, and I think the main question is, what issue on desktop this API tries to solve. Is it mainly IME composition/CJK issues? If yes, those should be rather fixed in scene2d imo (IME composition would need an SDL3 backend, as GLFW does not expose the relevant API's afaik).

PR #6772, opened 2022-01-17
gdx-lwjgl3-glfw-awt-macos extensions to add support for mixing GLFW and AWT on macOS

badlogic:
Borked the original branch unrecoverably, hence this new PR. Original PR https://github.com/libgdx/libgdx/pull/6673

# What and why
GLFW and AWT both compete for the AppKit run loop. LWJGL3 apps on macOS thus need to specify the -XstartOnFirstThread JVM option. This prevents AWT (and by proxy Swing, ImageIO, etc.) from working together with LWJGL3 on macOS.

This commit injects a modified libglfw.dylib (see https://github.com/badlogic/glfw). The GLFW fork is modified such that all calls into AppKit APIs are expected to be performed on a thread != main thread. This way, AWT can take over the run loop, while GLFW (and libGDX LWJGL3) apps can live on a separate thread.

# Usage
To use this, add `gdx-lwjgl3-glfw-awt-macos` as a dependency to your libGDX desktop project. You can now use both libGDX and AWT/Swing/ImageIO on macOS.

Remove the `-XstartOnFirstThread` JVM option from your run configurations/packr'ed apps/bash scripts.

# Caveats
* This extensions does NOT allow rendering your libGDX content to an AWT/Swing window.
* This extension requires LWJGL 3.3.0-SNAPSHOT
* This PR builds on top of this PR https://github.com/libgdx/libgdx/pull/6672

Issue #6992, opened 2022-09-27
Issues with window resizing and closing on MacOS


kopolop:
Please ensure you have given all the following requested information in your report.

#### Issue details
We encountered two issues related to application windows on MacOS:
- When the app window is resized, the app very often (but not always) crashes
- When the app window is closed, sometimes the app does not quit, but the window cannot be reopened. WindowListener's closeRequested is not triggered.
- 
#### Reproduction steps/code
- Resize an app window, the app crashes with exception "failed assertion `A command encoder is already encoding to this command buffer"
- Move the app window around a little bit, minimize, reopen the app window, then close the app window (by clicking the X button, or press Command Q). Observe that the app is not quit, but the window cannot be reopened.

#### Version of libGDX and/or relevant dependencies
libgdx = 1.11.0
roboVM version = 2.3.16
com.badlogicgames.gdx:gdx-lwjgl3-glfw-awt-macos extension is used.

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ x] macOS

crykn:
Probably related to #6987.

Thakor-Yashpal:
#2322

tommyettinger:
@Thakor-Yashpal You are not being helpful right now. I also have some suspicions that you are a bot, or are acting in an atypical way for a legitimate human user. Please send a picture of a cute cat to confirm that you are human. I also just like cats.

MobiDevelop:
Unable to reproduce on my Intel Mac, #6987 suggests that these issues may only exist on ARM Macs.

Thakor-Yashpal:
@tommyettinger
![image](https://github.com/user-attachments/assets/bc4d8541-22ed-40ad-a2c4-130400fde3b4)

PR #7117, opened 2023-04-04
Fix LWJGL3 window not updating on resize. Resolves #6808

### LINKED ISSUE #6808: LWJGL3 windows no longer update while being resized
Looks like we have a regression on our hands. Videos are by @fgnm in #6785 - that issue is closed and the callback the issue it was about is resolved. While our windows used to update when resized, they don't anymore, which looks a bit odd.

My concern is more to do with games during development than once released, as resizing your window is an excellent way of telling if UI and all that is responding to different sizes correctly.

## 1.10.0

https://user-images.githubusercontent.com/60154347/156578877-accacbc5-fd5b-47ca-99f6-9b71627fcfb7.mp4

## 1.10.1-SNAPSHOT

https://user-images.githubusercontent.com/60154347/156578881-0845af51-4a3f-4233-9661-5710cf6e547d.mp4

## Affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [X] Windows
- [ ] Linux
- [X] macOS

Windows and macOS tested by fgnm. Does not affect Linux in my test.

raeleus:
Since #6772, libGDX users have had an issue on the LWJGL3 platform when resizing the libGDX window on Windows. When players drag the window border, it does not call update. This results in black, unrendered areas until the user releases the mouse button. See #6808. This particularly impacts all tool developers who depend on UI reacting to window resize.

This fix, courtesy of @BambooBandit, addresses this problem. As noted in the issue, this only affects Windows users, so the fix should not be applied to Linux or Mac. Thank you.


BambooBandit:
Yes, this change basically stops Windows from being pulled down under with Mac

raeleus:
It was brought to my attention that Linux is actually affected by this issue, so I've included it with my fix. It is ready for review now. Thanks.

rafaskb (approved):

crykn:
This PR looks great! However, I [still](https://github.com/libgdx/libgdx/issues/6808#issuecomment-1110198514) think, we should check whether the `postRunnable` code is even needed on macOS (or maybe only when used with the gdx-lwjgl3-glfw-awt-macos extension).

HydrolienF:
`I have an old mac laptop laying around (2015 model) I can test with if that would help. What should I try out? Running this branch with and without the postRunnable code while testing the resizing?`
It seem's to be the rigth thing to test.
You can try to run master branch to compare with.

crykn:
@JamesTKhan: Yes. I guess the simplest approach is to just include `SharedLibraryLoader.isMac` in [this](https://github.com/libgdx/libgdx/pull/7117/files#diff-040a0bcb5d798c1906e75ea04d512d8a2ce33b694b8b1c4fe8edd6e6ffd9a2c5R77) if-condition and then check whether it misbehaves with and without the gdx-lwjgl3-glfw-awt-macos extension.

JamesTKhan:
I haven't forgotten about this, trying to test it today ran into some intellij version issues on the mac I gotta figure out.

JamesTKhan:
Sorry guys, my Mac is apparently older then I thought. I cannot get latest libGDX to build on the Mac and on the current version of IntelliJ due to android and when I tried updating IntelliJ to a newer version it crashes on launch as IntelliJ is trying to make API calls this old OS does not have available

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
```diff
@@ -73,23 +74,35 @@ public class Lwjgl3Graphics extends AbstractGraphics implements Disposable {
...
-					updateFramebufferInfo();
-					if (!window.isListenerInitialized()) {
-						return;
-					}
-					window.makeCurrent();
-					gl20.glViewport(0, 0, backBufferWidth, backBufferHeight);
-					window.getListener().resize(getWidth(), getHeight());
-					window.getListener().render();
-					GLFW.glfwSwapBuffers(windowHandle);
+			if (SharedLibraryLoader.isWindows || SharedLibraryLoader.isLinux) {
```
crykn, line 77: I think we should replace this with `Configuration.GLFW_CHECK_THREAD0.get(true)` as @piotr-j suggested.

raeleus:
It seems I missed the conversation since Piotr's comment. I've implemented the suggested changes, however it seems I managed to break the build. Can someone help me out? I don't understand what has changed.

Berstanio:
> It seems I missed the conversation since Piotr's comment. I've implemented the suggested changes, however it seems I managed to break the build. Can someone help me out? I don't understand what has changed.

```
import org.lwjgl.opengl.GL32;
```
The import was added, but I think, it isn't needed.

raeleus:
Oh, I see some changes have been made to this file since I've made the PR. Am I supposed to pull origin? I thought that was frowned upon.

Berstanio:
I think you can just remove the import? This whole thing will be probably squashed merged in any case.

raeleus:
Gah, Github Desktop just fetched the origin anyway. Oh well. Looks like it's ready for testing right now everyone. Thank you.

crykn (approved):

tommyettinger:
I'd like to review this, but I only have a Windows computer that I can easily test on. Has Mac been tested enough that we think it's good?

Berstanio (approved):
Works fine for me on mac M1 ventura, without glfw-awt-macos extension.
(It crashes with glfw-awt-macos, https://github.com/libgdx/libgdx/issues/6992)

Issue #6628, opened 2021-08-15
Mp3 Music setPosition works wrong


denisk20:
#### Issue details
I have an mp3 track of 135 seconds. When I set the music position to 134 seconds I expect the music to play for 1 second. Instead it plays for almost 10 seconds. 

#### Reproduction steps/code

		music = Gdx.audio.newMusic(Gdx.files.internal("24STRINGSBASSSOLO1.mp3")); // a 135 sec. track
		music.play();

		Timer.instance().scheduleTask(new Timer.Task() {
			@Override
			public void run() {
				music.setPosition(134); // expect it to play for only 1 second, instead it plays almost 10.
			}
		}, 0.1f);

Sample repository is here https://github.com/denisk20/test-music-position

#### Version of libGDX and/or relevant dependencies
1.10.0

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [x ] MacOS

denisk20:
There's no such issue for ogg files.

Frosty-J:
Tested on Linux - same issue. `getPosition()` appears accurate if you let the track play all the way through without seeking, but `setPosition()` seems to cause it to become more out-of-sync the further you seek.

I tried various re-encodes with different bitrates (including variable), sample rates and joint/stereo/mono, both with Audacity and ffmpeg's libmp3lame, none of which made any difference.

No issues when running an HTML build in the browser.

While it's for Android, #5693 sounds related:
> According to [this](https://github.com/google/ExoPlayer/issues/328#issuecomment-168532148), this is a known issue for MP3 files (linked project isn't technically the MediaPlayer bundled with Android, but I'll assume the seeking behavior is similarly implemented). Converting the file to WAV or OGG fixes the issue, which is good enough for me.

Obviously this must be possible to fix, as media players such as Audacious and VLC have no such issue. But I don't have any ideas for how - all I can really come up with is the idea to make it spew warnings if the current position is greater than the track's duration.

HydrolienF:
I have found from where it's comming from & start a pull request #7119

PR #7119, opened 2023-04-08
Fix Mp3 Music setPosition works wrong #6628

HydrolienF:
setPosition for MP3 musics wasn't setting time position at the rigth place.

**How to test this fix ?**
You can see that this fix work by running `.\gradlew launchTestsLwjgl3` then choose MusicTest, play the mp3 file go to aproximatly 3:00. File is 3:03 min long, so it should stop at 3:03, but currently it continue over 3:10. (This fix make it stop at 3:03)
You can play file with a music player at libgdx/tests/gdx-tests-lwjgl3/build/resources/main/data/8.12.mp3 at 3:00 music fall down then stop at 3:03

**How does it fix it ?**
My guess is that mp3 reader don't fill buffer as wav & ogg do.
So to calculate how much of the file have been read we need to use the return of read(int) instead of bufferSize. Otherwise we think that 40960 have been read when only 40200. So wrong offset increase over time.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@
 - [BREAKING CHANGE] 3D API - max bone weights is now configurable and limited to 4 by default. Change this value if you need less or more. See #6522.
 - [BREAKING CHANGE] Mesh#getVerticesBuffer, Mesh#getIndicesBuffer, VertexData#getBuffer, and IndexData#getBuffer are deprecated in favor to new methods with boolean parameter. If you subclassed some of these classes, you need to implement the new methods.
 - [BREAKING CHANGE] Desktop: The return value of AudioDevice#getLatency() is now in samples, and not milliseconds
+- LWJGL3 Fix: setPosision() for MP3 files.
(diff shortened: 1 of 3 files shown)


HydrolienF:
Can I do someting else to help this merge request being merge ?

Frosty-J:
Probably not. I see no issues with this but I don't have the required permissions to do anything that'll get it merged faster. Even small pull requests (such as this one) can easily take months in these parts. If you need this fix for a project of your own, best build it against your own fork for the time being.

HydrolienF:
Great idea, I will use my own libgdx patched version for now.

SimonIT (dismissed):
Works (from what I've tested)

crykn (changes requested):
Could you add the same change to the `OpenALMusic` class in the old LWJGL backend as well?

HydrolienF:
@crykn same fix apply for LWJGL.

crykn (dismissed):

PR #7120, opened 2023-04-08
Remove some duplicate code in Intercector

HydrolienF:
We have discuss about merging function that do the same thing with Vector2 or int x , int y with TEttinger & mgsx on Discord.
It make code cleanner (avoid more code lines, avoid the risk to have different behavior between the 2 functions) & less call to x & y of vector2d for a call to one more function.


mgsx-dev (approved):
LGTM

PR #7125, opened 2023-04-19
TextureUnpacker using Gdx, no AWT involvement.

MarcinSc:
However this one has to be run from Gdx application context, so that it has access to OpenGL and Gdx.file. As a result, it cannot be run from command line.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TextureUnpackerGdx.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TextureUnpackerGdx.java
@@ -1,0 +1,174 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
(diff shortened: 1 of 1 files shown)


mgsx-dev (changes requested):
I think you could replace the existing TextureUnpacker with yours. There is no reason to have both.

So it requires a main method to be run from CLI and use headless backend (see KTXProcessor which does exactly that : it only use FileHandle and Pixmap)

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TextureUnpackerGdx.java
```diff
@@ -0,0 +1,164 @@
...
+
+	private boolean quiet;
+
+	/** Splits an atlas into seperate image and ninepatch files. */
+	public void splitAtlas (TextureAtlasData atlas, String outputDir) {
+		// create the output directory if it did not exist yet
+		FileHandle outputDirFile = Gdx.files.local(outputDir);
+		if (!outputDirFile.exists()) {
+			outputDirFile.mkdirs();
+			if (!quiet) System.out.println(String.format("Creating directory: %s", outputDirFile.path()));
```
SimonIT, line 46: You could use one of these
```java
Gdx.app.log("MyTag", "my informative message");
Gdx.app.error("MyTag", "my error message", exception);
Gdx.app.debug("MyTag", "my debug message");
```
for logging. Then we don't need `quiet`.

Tom-Ski:
+1 for non AWT TexturePacker and replacement. I think I had a dream where you did that @tommyettinger.

For a solution right now for m1, I spawn a new jvm and run tools as its own runnable jar programmatically when I need to pack/unpack on m1, which is not very nice

tommyettinger:
It turns out I did make [a standalone texture packer and unpacker](https://github.com/tommyettinger/libgdx-texturepacker/tree/main); they don't use LWJGL3, but still use AWT, and I think should work on Macs. I made quite a few of the gdx-tools into standalone apps, though there's no central directory of them now, so the easiest way to find them is by looking through the repos on my GitHub profile. They all have names starting with `libgdx-` .

PR #7126, opened 2023-04-23
fix OpenGL version in tests

mgsx-dev:
i noticed a crash running InstancedRenderingTest because glVertexAttribDivisor requires OpenGL 3.3+.

Note that OpenGL 4.3 is needed in order to fully emulate GLES 3.0. (see LwjglGL30 and Lwjgl3GL30 java imports)


mgsx-dev:
@Tom-Ski here is the optimal version (with caching) : https://github.com/libgdx/libgdx/commit/df192fe24dd24e893a8f882c1f2482f2a6266020
as you can see the only breaking change is the new parameter for Mesh.bind/unbind. It's only used by BaseShader in the repo but some user may use it (it's not common to use it directly but still).

What do you think? do you want me to PR that?

PR #7127, opened 2023-04-23
fix instanced rendering when used with Renderable or Model

mgsx-dev:
ModelInstancedRenderingTest is a copy of InstancedRenderingTest but using renderables instead.

In this case, Mesh used cached vertex attributes location for instanced buffer, which was wrong : instanced attributes are not part of vertex attributes.

Without the fix, the provided test crash because it has more instanced attributes than vertex attributes (index out of bounds exception) but for other cases, it fails silently and instanced data wasn't available in the shader.

This change is not optimal, we could cache instanced attributes locations as well but i'm not sure the best way to do that without breaking API a bit : we would need something like :
* change `Mesh.bind/unbind (final ShaderProgram shader, final int[] vertexLocations, final int [] instancedLocations)`
* add `Mesh.getInstancedAttributes()`

Since it's an optimisation, it should be done in another PR anyway. Any opinion on that ?

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Mesh.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Mesh.java
@@ -512,7 +512,7 @@ public class Mesh implements Disposable {
 	 * @param locations array containing the attribute locations. */
 	public void bind (final ShaderProgram shader, final int[] locations) {
 		vertices.bind(shader, locations);
-		if (instances != null && instances.getNumInstances() > 0) instances.bind(shader, locations);
+		if (instances != null && instances.getNumInstances() > 0) instances.bind(shader);
 		if (indices.getNumIndices() > 0) indices.bind();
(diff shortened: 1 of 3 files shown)


mgsx-dev:
(sorry, my comment was an the wrong issue, sunday you know... :D )

@Tom-Ski here is the optimal version (with caching) : https://github.com/libgdx/libgdx/commit/df192fe24dd24e893a8f882c1f2482f2a6266020
as you can see the only breaking change is the new parameter for Mesh.bind/unbind. It's only used by BaseShader in the repo but some user may use it (it's not common to use it directly but still).

What do you think? do you want me to PR that?

Tom-Ski:
Any custom shader that isn't extending from defaults would probably have the issues, but I think its ok to have this breaking change. Or we add different methods to Mesh, but IMO that is worse

PR #7130, opened 2023-04-24
Upgrade gradle to 8.1.1

ice1000:
As title

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,7 +15,7 @@
 - [BREAKING CHANGE] iOS: 32 bit (armv7) builds are no longer supported. Builds must be 64 bit (arm64) only.
 - [BREAKING CHANGE] iOS: Use dynamic frameworks instead of static libs
 - [BREAKING CHANGE] optimized Mesh#bind and Mesh#unbind have a new parameter for instanced attribute locations. If you use these methods without instancing, you can pass a null value.
-- Update to jnigen 2.4.0
+- Update to jnigen 2.4.1
 - LWJGL Fix: setPosision() for MP3 files.
 - iOS: Add new MobiVM MetalANGLE backend
 - iOS: Update to MobiVM 2.3.19
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ buildscript {
 	}
 	dependencies {
 		classpath "com.diffplug.spotless:spotless-plugin-gradle:${versions.spotless}"
-		classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.4.0"
+		classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.4.1"
 	}
 }
 


ice1000:
This version of gradle supports Java 19

Frosty-J:
So, on Gradle 7 I get this if I run with `org.gradle.warning.mode=all`:

> The JavaExec.main property has been deprecated. This is scheduled to be removed in Gradle 8.0. Please use the mainClass property instead. See https://docs.gradle.org/7.5.1/dsl/org.gradle.api.tasks.JavaExec.html#org.gradle.api.tasks.JavaExec:main for more details.

I thought that meant Gradle 8 wouldn't work without further changes, but it seems they've pushed the removal of that back to Gradle 9.0. I can confirm this simply changing the version to 8.1.1 works for running desktop builds. I have not tried other platforms.

tommyettinger:
I haven't updated to 8.x for gdx-liftoff because of Android incompatibilities when it first came out. These may have been fully or partially resolved, but if they aren't, we should not update yet. Note that just because Android Studio supports an Android Gradle Plugin version that can use Gradle 8, does not mean IntelliJ IDEA supports it, and we really need to have both working.

If you just want Java 19 support, you can use Gradle 7.6.1 today; it's backwards-compatible with 7.5.1 .

I haven't looked at the exact changes yet, but the comments lacking any mention of Android was a concern for me.

ice1000:
> I thought that meant Gradle 8 wouldn't work without further changes, but it seems they've pushed the removal of that back to Gradle 9.0.

I've read their code, they simply have a function to generate a warning saying that something will be removed in the <current major version plus one> version. So we should be good.

ice1000:
I'm currently an end user, but it's slightly annoying that the default app generator installed Gradle 7.5.1 on my machine. Using 8.1.1 in an app that relies on libgdx has no problem, but that's just my own observation.

I agree that as upstream maintainers, you people need to be mroe careful. I'll leave my PR here and feel free to do whatever with it :p

tommyettinger:
Indeed, the Android Gradle Plugin version we use also needs to be updated, as the failed build may show:

```
A problem occurred configuring project ':tests:gdx-tests-android'.
> Could not create task ':tests:gdx-tests-android:mergeDebugShaders'.
   > Cannot use @TaskAction annotation on method IncrementalTask.taskAction$gradle_core() because interface org.gradle.api.tasks.incremental.IncrementalTaskInputs is not a valid parameter to an action method.
```

We should keep this PR open until an updated AGP is generally available and works in AS and IDEA. Then it can be made to include the AGP update, we can hopefully get the tests to pass without much hassle, and it can be merged.

ice1000:
> Did you upgrade gradle just by editing the file directly?
> I would use this command instead to update gradle.
> `./gradlew wrapper --gradle-version 8.1.1`
> https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper
> 
> It does the same thing but also updates the gradle wrapper scripts. When I upgraded from 7.5.1 using this command it resulted in changes to gradlew and gradlew.bat as well as updating gradle-wrapper.properties. They were very, very minor and only changed the comments but I would do that for the sake of completeness.

I just edited the versions. Sometimes they change some copyright things in the scripts which I don't like, but I can do so if nobody's against so.

PR #7131, opened 2023-04-24
Remove thumbv7 reference

obigu:
Remove 32 bit when building from Gradle as it's not longer supported.

### DIFF
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/ios/build.gradle
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/ios/build.gradle
@@ -13,7 +13,7 @@ launchIOSDevice.dependsOn build
 createIPA.dependsOn build
 
 robovm {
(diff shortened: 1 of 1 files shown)


Berstanio:
Does this mean, 32bit builds of e.g. ObjectAL can also be removed? Or is this only about a mobiVM incompatibility?

obigu:
Well, not mobiVM on itself but XCode 14 dropped support and many 3rd party SDKs are not including 32 bit anymore. The change is to have a sensible default for new Gdx Setup projects. 

I think it's still possible with Xcode 13 and MobiVM 2.3.19 to make a 32bit + 64bit. I was expecting a sooner release so it made sense to keep min deployment target to iOS 9 but now I'm not so sure. 

We can either drop 32 bit in this version and bump min deployment target to 11 or keep it so that we can brag about being the only game engine that supports iOS 9 in 2023.

Berstanio (approved):
Generated project works fine (not a suprise I think)

Issue #7104, opened 2023-03-14
MathUtils.lerpAngleDeg returns incorrect value if -720 < toDegree - fromDegree < -540


ZumiKua:
#### Issue details
As the title said. MathUtils.lerpAngleDeg returns incorrect value if ````-720 < toDegree - fromDegree < -540````, ````MathUtils.lerpAngle```` should be affected too.

#### Reproduction steps/code 
````
public void testLerpAngleDeg() {
        float fromDegree = 541;
        float toDegree = 0;
        assertThat(MathUtils.lerpAngleDeg(fromDegree % 360, toDegree, 0.5f)).isEqualTo(270.5f); //passed
        assertThat(MathUtils.lerpAngleDeg(fromDegree, toDegree, 0.5f)).isEqualTo(270.5f); //failed, lerpAngleDeg returns 90.5f
    }
````

#### Version of libGDX and/or relevant dependencies
I'm using 1.9.8, but all versions of libGDX which have this api should be affected.

#### Please select the affected platforms
- [X] Android
- [X] iOS
- [X] HTML/GWT
- [X] Windows
- [X] Linux
- [X] macOS

tommyettinger:
I tested this in my fork and the behavior is indeed incorrect right now. I can PR whenever, but I'd like to make the tests more robust, and check more combinations of large, small, negative, etc. angles.

ZumiKua:
Nice, after rounding `toDegrees - fromDegrees`, I think testing these situlations would be enough: 
````-360 < to - from < -180````
````-180 < to - from < 0````
````0 < to - from < 180````
````180 < to - from < 360````
````to - from = -180````
````to - from = 0````
````to - from = 180````
other results will be rounded anyway.

And are you plan to fix the negative return value if `fromDegrees` is negative?

tommyettinger:
> to - from = 180

This case I'm explicitly not checking because there are two valid answers when the alpha parameter is not 0 or 1. The same is true for a difference of -180.

The unfortunate part here is that any fix will require at least one extra modulus operation, on floats, which is not an especially fast operation. The best case would be for inputs to come into this in a valid range, so no fix would be needed, but asking for that tends to be wishful thinking for a library.

RE: the negative return value, that should probably get fixed. I didn't know that happened currently.

ZumiKua:
How about we check whether the difference is in the range of (-360, 360) before doing the modulus operation? Compare operation should be much faster than modulus operation? (well, this is my wild guess)

[...]

PR #7132, opened 2023-04-24
MathUtils.lerpAngle() fixes for extreme inputs.

tommyettinger:
When lerpAngle() or lerpAngleDeg() would be given a large negative angle, the result would be incorrect before, often by a large amount. This was found in #7104 by ZumiKua, and probably affects most libGDX versions so far. The discussion at that issue covers most of what went into this PR. I tested this both in libGDX's tests and in my library [digital's tests](https://github.com/tommyettinger/digital/blob/main/src/test/java/com/github/tommyettinger/digital/MathToolsTest.java), and it seems just about bulletproof for finite inputs. There could certainly be issues if any numbers get large enough to incur FP precision loss, but at some point you just have to say "your numbers are too big to be precise, so don't expect correct results." The downside here is that to support most of the reasonable number line, we need two more `%` operations, which are not incredibly cheap. However, because it's really common for angles to go above or below the 0-360 degree range when users can control an angle, I think having correct results for larger negative inputs is important.


mgsx-dev (approved):
LGTM

PR #7134, opened 2023-05-01
Use latest Gradle API to register tasks

ice1000:
Gradle now recommends using `tasks.register`. See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_guidelines

### DIFF
--- a/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
+++ b/extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
@@ -22,7 +22,7 @@ gretty.resourceBase = project.buildDir.path + "/gwt/draftOut"
 gretty.contextPath = "/"
 gretty.portPropertiesFileName = "TEMP_PORTS.properties"
 
-task startHttpServer () {
+tasks.register("startHttpServer") {
     dependsOn draftCompileGwt
 
     doFirst {
@@ -38,7 +38,8 @@ task startHttpServer () {
     }
 }
 
-task beforeRun(type: AppBeforeIntegrationTestTask, dependsOn: startHttpServer) {
+tasks.register("beforeRun", AppBeforeIntegrationTestTask) {
+    dependsOn startHttpServer
     // The next line allows ports to be reused instead of
     // needing a process to be manually terminated.
     file("build/TEMP_PORTS.properties").delete()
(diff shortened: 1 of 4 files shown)


tommyettinger (commented):
This just seems less clear; I don't see the benefit here. Is it something to do with the name of the task being a String instead of an identifier?

ice1000:
I think it's because the old behavior is different. If we want the task creation to be executed immediately (instead of when it's needed), then we use the old syntax. Code might rely on that behavior. For example, if you write

```groovy
def o = false
task ok {
  o = true
}
assert o
```

Then this code will break with the new syntax. But we're not relying on that behavior, so we're good to switch.

ice1000:
This PR is a superset of #7042

obigu (approved):

ice1000:
Thank you very much!

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
```diff
@@ -38,7 +38,8 @@ task startHttpServer () {
     }
 }
 
-task beforeRun(type: AppBeforeIntegrationTestTask, dependsOn: startHttpServer) {
+tasks.register('beforeRun', AppBeforeIntegrationTestTask) {
```
tommyettinger, line 41: This section in particular seems to entirely break HTML project imports when using gdx-setup. It's something about the integration testing using a task that causes it to fail, and I don't know what ways there are around it that still use `tasks.register()`. Using the `task` syntax from before this PR works fine.

tommyettinger (commented):
This needs to be re-evaluated, particularly the HTML section, because HTML projects this generates right now won't import or build.

extensions/gdx-setup/res/com/badlogic/gdx/setup/resources/html/build.gradle
tommyettinger: I know the task configuration is related because [gdx-liftoff](https://github.com/tommyettinger/gdx-liftoff) doesn't use it and doesn't have this problem. With the summer game jam coming up shortly, and HTML projects common there, it seems like any HTML jam submissions will have to be made with gdx-liftoff, an older gdx-setup, or an existing template.

ice1000: Let me revert this particular case?

It also seems unknown to me why this doesn't work...

PR #7137, opened 2023-05-04
Use lambda expressions in `Scaling`

ice1000:
As title. This should make the desktop version run faster due to the special JVM instructions for lambdas

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Scaling.java
+++ b/gdx/src/com/badlogic/gdx/utils/Scaling.java
@@ -20,110 +20,92 @@ import com.badlogic.gdx.math.Vector2;
 
 /** Various scaling types for fitting one rectangle into another.
  * @author Nathan Sweet */
-public abstract class Scaling {
-	protected static final Vector2 temp = new Vector2();
+public interface Scaling {
+	 Vector2 temp = new Vector2();
 
 	/** Returns the size of the source scaled to the target. Note the same Vector2 instance is always returned and should never be
 	 * cached. */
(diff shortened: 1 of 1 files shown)


ice1000:
:sob: Java 8 seems not enabled, I'm sorry.

tommyettinger:
This would be a good improvement (probably, I haven't benchmarked) if https://github.com/libgdx/libgdx/issues/5487 ever resolves and we can finally update to Java 8.

ice1000:
> This would be a good improvement (probably, I haven't benchmarked) if #5487 ever resolves and we can finally update to Java 8.

I expect a gigantic PR that does this for the entire codebase for all classes once we migrate to 8... So this is too minor to exist.

tommyettinger:
Migrating to 8 might be fairly small -- right now we have the potential to access *language level 8* from everywhere except very old Android devices, but we don't have access to the full `*standard library from Java 8* except on desktop JVMs. RoboVM, in particular, doesn't even have standard functional interfaces like `Consumer` or `Function`, though it can compile lambda expressions, method references, and default methods today. I feel like moving on to the next Java version will pick up in urgency once Java 21, an LTS release, comes out and doesn't support compiling with Java 7 as a target (I think 20, which is out now, doesn't either). Android Gradle Plugin (AGP) 8.x already requires a JDK version 17 or newer, and the version before that needed at least 11. AGP 9.x could very easily require JDK 21, which will be the latest LTS release of Java when the 9.x line of AGP is scheduled for release in 2024.

PR #7138, opened 2023-05-06
feat: Update to jnigen 2.4.0

Berstanio:
For the moment this pr relies on a snapshot. Once jnigen 2.4.0 is released, this pr will be updated and can be merged. 

Currently we need to rely on `-undefined dynamic_lookup` for the iosgl natives, since the symbol resolution is backend dependent for either MetalANGLE or OpenGLES. 

ObjectAL can be changed to a dynamic framework, if https://github.com/libgdx/ObjectAL-for-iPhone/pull/1 gets merged.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,9 @@
 - [BREAKING CHANGE] 3D API - max bone weights is now configurable and limited to 4 by default. Change this value if you need less or more. See #6522.
 - [BREAKING CHANGE] Mesh#getVerticesBuffer, Mesh#getIndicesBuffer, VertexData#getBuffer, and IndexData#getBuffer are deprecated in favor to new methods with boolean parameter. If you subclassed some of these classes, you need to implement the new methods.
 - [BREAKING CHANGE] Desktop: The return value of AudioDevice#getLatency() is now in samples, and not milliseconds
+- [BREAKING CHANGE] iOS: Removed armv7 builds
+- [BREAKING CHANGE] iOS: Use dynamic frameworks instead of static libs
+- Update to jnigen 2.4.0
 - LWJGL Fix: setPosision() for MP3 files.
 - iOS: Add new MobiVM MetalANGLE backend
 - iOS: Update to MobiVM 2.3.19
--- a/build.gradle
+++ b/build.gradle
@@ -19,10 +19,11 @@ buildscript {
 	repositories {
 		mavenCentral()
 		gradlePluginPortal()
+		maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
 	}
 	dependencies {
 		classpath "com.diffplug.spotless:spotless-plugin-gradle:${versions.spotless}"
-		classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.3.1"
+		classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.4.0-SNAPSHOT"
 	}
 }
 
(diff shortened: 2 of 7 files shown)


obigu:
This will drop 32 bit support right?

I think we should have a separate PR that increases deployment target to 11. There are no 64 bit devices that can't update at least to iOS 12 and those are 10 years old already (like the iPhone 5S). One of the benefits of iOS 11 is also that all devices support ANGLE which means we don't need the OpenGL fallback on the MetalANGLE backend which is something that even if I think was discussed a bit on Discord may be worth considering.

As a minor thing, maybe we should set the PR to Draft?

Berstanio:
Yes, this drops 32bit support through jnigen. The change for ObjectAL will be a different pr, this is only about jnigen.

I'm fine with bumping to iOS 11, I would integrate the target bump into jnigen too?
However, I would still advise against dropping OpenGLES support from MetalANGLE, there are a lot mac VM/hackintosh out there, that don't support Metal. They would completly lose simulator support.

> As a minor thing, maybe we should set the PR to Draft?

Yep, good idea!

obigu:
> Yes, this drops 32bit support through jnigen. The change for ObjectAL will be a different pr, this is only about jnigen.
> 
> I'm fine with bumping to iOS 11, I would integrate the target bump into jnigen too? 

I don't know if there could be any unintended consequences not doing it but keeping them consistent seems the safe choice to me.

However, I would still advise against dropping OpenGLES support from MetalANGLE, there are a lot mac VM/hackintosh out there, that don't support Metal. They would completly lose simulator support.

Good point but, shouldn't hackintosh use the default iOS backend instead of MetalAngle so that what they simulate is closer to what the device will actually so? Also, is there any overhead of keeping the OpenGL headers?

> 
> > As a minor thing, maybe we should set the PR to Draft?
> 
> Yep, good idea!

Berstanio:
> Good point but, shouldn't hackintosh use the default iOS backend instead of MetalAngle so that what they simulate is closer to what the device will actually do?

Theoretically yes, but you can still use a real device to do most of your testing. Also, I think ANGLE is stable enough, that you probably won't really run into ANGLE bugs. And maybe we want to make the MetalANGLE backend at some point the default, at least then it would get inconvinient.

> Also, is there any overhead of keeping the OpenGL headers?

Afaik not really, MetalANGLE just passes the call it gets further to OpenGLES, so it's only 1/a few stackframes more.

Berstanio:
jnigen 2.4.0 has been released

CHANGES
```diff
@@ -12,6 +12,9 @@
 - [BREAKING CHANGE] 3D API - max bone weights is now configurable and limited to 4 by default. Change this value if you need less or more. See #6522.
 - [BREAKING CHANGE] Mesh#getVerticesBuffer, Mesh#getIndicesBuffer, VertexData#getBuffer, and IndexData#getBuffer are deprecated in favor to new methods with boolean parameter. If you subclassed some of these classes, you need to implement the new methods.
 - [BREAKING CHANGE] Desktop: The return value of AudioDevice#getLatency() is now in samples, and not milliseconds
+- [BREAKING CHANGE] iOS: Removed armv7 builds
```
obigu, line 15: To make sure everybody understands the change maybe say something like "32 bit (armv7) builds are no longer supported. Builds must be 64 bit (arm64) only."

noblemaster (approved):

noblemaster:
This looks fine. I suggest, let's merge.

ice1000:
Do we still need the oss sonatype repository?

PokeMMO:
> Do we still need the oss sonatype repository?

Good catch, no we do not.

Issue #6635, opened 2021-08-20
Ogg files load way more slowly than wav files


hasen6:
#### Issue details
The only change I make is to switch my sound effect folder all to .ogg rather than .wav and the load time difference is:

All wav sounds:
assets 229
screen 441

All ogg sounds:
assets 1548
screen 1749

The difference in loading time is somewhat noticeable anyway, but I've provided the actual times for clarity.

#### Reproduction steps/code
Load sound effects as ordinary `Sound` files as wav and then all as ogg and test the loading time of each.

#### Version of libGDX and/or relevant dependencies
1.10.0

#### Please select the affected platforms
- [x] MacOS

tommyettinger:
Is that measured in milliseconds? If so, taking 1.5 seconds to load a small amount of .ogg files seems like a lot, but not for a large amount of .ogg files. I'm not at all familiar with audio in libGDX, but IIRC .wav is an uncompressed audio format, while .ogg compresses sounds to reduce filesize. Decompressing an .ogg then has to take some time; it's just unfortunately more than would be optimal here. This is mostly just the difference between the file types, though maybe whatever dependency we use to decode .ogg could be updated or something...

I think this is likely to be closed because we can't make .ogg into an uncompressed format; the file size improvement isn't free. I'll leave it open in case there could be something to update.

hasen6:
> Is that measured in milliseconds? If so, taking 1.5 seconds to load a small amount of .ogg files seems like a lot, but not for a large amount of .ogg files.

Yes it's in milliseconds. The issue is it takes that long to load the SAME amount of audio files with the only difference being ogg and wav format. I realise ogg is compressed but after searching it seems it's the most common format for sounds and music but if it's gonna take 3-5 times as long as loading wavs I don't see how it can be practical.

>I think this is likely to be closed because we can't make .ogg into an uncompressed format; the file size improvement isn't free. I'll leave it open in case there could be something to update.

If it's an issue with ogg itself then it probably wouldn't be the most common audio format? I noticed mp3 loads equally slow.

tommyettinger:
[...]

PR #7140, opened 2023-05-07
Use STBVorbis to decode ogg data for Sound

Hangman:
I replaced the jorbis ogg decoder by STBVorbis for Sounds (literally the Sound class, not Music) in the lwjgl3 backend. This drastically reduces the loading time, usually something between 40% and 60%, depending on the file.

"Fixes" [https://github.com/libgdx/libgdx/issues/6635](https://github.com/libgdx/libgdx/issues/6635) I guess?

Since this change will affect almost all users and I don't want to take this lightly, I'd like to mention the limitations of the decoder, that ... I don't understand ... but maybe someone else:

- floor 0 not supported (used in old ogg vorbis files pre-2004)
- lossless sample-truncation at beginning ignored
- cannot concatenate multiple vorbis streams
- sample positions are 32-bit, limiting seekable 192Khz files to around 6 hours (Ogg supports 64-bit)

I only tested on Windows 11.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/Ogg.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/Ogg.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


BambooBandit:
I gave this change a try and sheesh its a lot faster. Haven't noticed any issues either. Tried it on win 10 LWJGL3

Berstanio:
Seems to work fine on MacOS too and it is really faster.
Is there a reason to only apply this to Sounds? I think it might be good to have it consistent for Sounds and Music.

Hangman:
I agree, consistency would be nice. However, I'm not entirely convinced that this alone justifies a rewrite of `OggInputStream`.

A few thoughts on this that speak against it for me:
Compared to this rather simple change, an implementation for Music would be much more complex. Because of files packed in jars, we can't let STVorbis open/read on its own, so we'd have to buffer both the encoded and the decoded data. Performance-wise I'm unsure if it is for the worse or the better with such small buffer sizes.

**tl;dr**
If we exclude the performance aspect ([#6509 ](https://github.com/libgdx/libgdx/pull/6509) should help), I am left with:
\+ consistency
\- replace a battle-tested implementation with a new one
\- it takes more time to write, test, and review (debatable, but I'd rather spend the manpower on other PRs like the one mentioned)

Berstanio:
Yeah, I looked a bit deeper into it and it certainly is more complicated to implement.
So I agree, this PR is good on its own in my opinion.

The performance difference would be still interesting to investigate at some point though.

Berstanio (approved):

crykn (approved):

Issue #6711, opened 2021-12-01
Update box2d rope joint


HAL9000Thomas:
The rope joint on box2d has few more features which are not available in libgdx. For example using rope tuning to modify the bending of a rope. Is it possible to get the latest box2d changes into libgdx?

#### Issue details
Update rope joint. Provide methods for setting stiffness of a rope.

#### Reproduction steps/code
1. Create a box2d rope in libgdx
2. Create a box2d rope in https://github.com/erincatto/box2d/
3. Compare rope joint possibilities 

### Comparison
current libgdx box2d: https://github.com/libgdx/libgdx/blob/master/extensions/gdx-box2d/gdx-box2d/jni/Box2D/Rope/b2Rope.cpp
latest original box2d: https://github.com/erincatto/box2d/blob/master/src/rope/b2_rope.cpp

#### Paths on libgdx
libgdx/extensions/gdx-box2d/gdx-box2d-gwt/src/com/badlogic/gdx/physics/box2d/gwt/emu/com/badlogic/gdx/physics
/libgdx/extensions/gdx-box2d/gdx-box2d/jni/Box2D

crykn:
Duplicate of #5948.

ice1000:
https://github.com/erincatto/box2d/blob/main/CHANGELOG.md says rope has been removed, please use distance. So we should switch to distance joints?

ice1000:
Looks like they're different things...

ice1000:
Fixed in my own fork, please stay tuned for an uploaded version

https://github.com/ice1000/gdx-box2d

PR #7144, opened 2023-05-13
Box2d 2.4.1

### LINKED ISSUE #5948: Update box2d
The Box2D version included in LibGDX is from 2014. Box2D has received many updates since then: https://github.com/erincatto/box2d

Is it possible to update the sources and JNI?

Could someone guide me on how to do it?

ice1000:
This PR ports https://github.com/ice1000/gdx-box2d into this repository. I am unfamiliar with the project structure and this is the best I can do with just myself. Let me know if anything can be changed!

Fix #5948 and #6711 and maybe some others.

For GWT, a possible way of rescue is to use emscripten + JSNI feature of GWT (pretty close to gdx-jnigen), but I'm not sure how those things work in general, and I need people to teach me.

Feedbacks & testing are welcome! I've published my fork on maven central with a different groupId for testing.

/cc @MobiDevelop


ice1000:
Oh I made a breaking change: `QueryAABB` is renamed to `queryAABB`. I'll add a deprecation redirection soon but is that renaming okay?

rafaskb:
I tested @ice1000 's Box2D version and it seems to work well. I didn't test much because I was only checking if the update would fix some issues I'm having with Box2D (issues persisted unfortunately), but the updated Box2D worked well.

I had some trouble understanding which dependencies to pull, e.g. is there just one dependency or there's one for :core and another for :desktop/lwjgl3? And there are plenty of breaking API changes, but overall it's functional.

ice1000:
> which dependencies to pull

What do you mean by that? I'm unsure what is "pulling" a dependency

ice1000:
I think the tester depends on lwjgl3 and the core of box2d apis only depends on `:core`. If it's not the case, I'll fix that

PR #7147, opened 2023-05-17
Fixed setting Gradle javaVersion for Eclipse project file generation.

NathanSweet:
**Repro:**
```
gradlew -PjavaVersion=16 eclipse
```

**Expected:**
Generated Eclipse `.classpath` project files will use a Java 16 `JRE_CONTAINER`.

**Actual:**
The project files use Java 18.

**Proposal:**
The generated Eclipse project files appear to use `sourceCompatibility` or `targetCompatibility`, not `options.release`. Setting both should be fine since `options.release` will be preferred for compilation.


NathanSweet:
@PokeMMO Ah, in that case how about we keep the `if` but remove the `else` so `sourceCompatibility` and `targetCompatibility` are always set?

tommyettinger (dismissed):
The Java 9 or higher check looks good; I always use something similar when targeting Java 8 bytecode and building with a newer JDK.

A concern is that Java 20 (which has been out for a few months, but still isn't supported by Gradle or maybe even Kotlin) no longer allows targeting JDK 7 with `release`. People have already attempted to use libGDX with Java 20, which still doesn't work, and never will if their core targets JDK 7 and all they have is a Java 20 JDK. We can safely target Java 8 on LWJGL2, GWT (well, a large subset), and iOS (if we don't use any classes or methods added in Java 8, and stick to language level features at most), and we have to target at least 8 with LWJGL3. Android has so many versions that it's hard to know what language level is safe right now. It may be time to update to Java 8 in libGDX itself, depending on what Android permits.

PokeMMO (dismissed):

gdx/res/com/badlogic/gdx/utils/JsonReader.rl
tommyettinger, line 1: Err, why is there JSON code in this PR? The changes look fine, I just wasn't expecting them here.

gdx/res/com/badlogic/gdx/utils/JsonReader.rl
```diff
@@ -1,5 +1,5 @@
 // Do not edit this file! Generated by Ragel.
-// Ragel.exe -G2 -J -o JsonReader.java JsonReader.rl
```
tommyettinger, line 2: What did the `-G2` flag do before? It isn't needed now?

tommyettinger (approved):
Things still look merge-able, despite a mix of Gradle config and new JSON code. The usage of compileJava.options.release is still good, since it checks that we're building with Java 9+ first. I don't see anything blocking.

gdx/res/com/badlogic/gdx/utils/JsonReader.rl
NathanSweet: I accidentally pushed unrelated changes! Will fix.

gdx/res/com/badlogic/gdx/utils/JsonReader.rl
NathanSweet: `-G2` is for `Really fast goto-driven FSM` but only applies to C and D languages, it's never done anything for Java.

NathanSweet:
Thanks for checking things out! I reverted an unrelated commit, oopsie. Will squash the mess when merging.

PR #7128, opened 2023-04-23
optimize instanced rendering with renderables (using cache)

mgsx-dev:
Here is the optimal solution for #7127

PR #7149, opened 2023-05-18
Resolve NPE on null instance attributes

JamesTKhan:
Discovered from Discord chat that NPE is possible with changes from https://github.com/libgdx/libgdx/pull/7128 and replicated on local running Animation3DTest. Instanced VertexAttributes may be null but the code is not checking for null before calling methods (size()) on them. Can be replicated on any test that uses ModelBatch/BaseShader without instanced rendering.


Tom-Ski (approved):

tommyettinger (approved):
Certainly looks much safer than before.

PR #7150, opened 2023-05-20
Vector3 - Change all methods that modify coords to use set() only

GadWissberg:
Hello,
I had a tiny scenario where I wanted to make an immutable vector3 class which just extends Vector3 and forbids to modify its data by overriding the set(), but then I saw there are multiple methods that manually modify the coords. I thought it'd be easier just to override the set(), so I PR'd my suggestion.
Thanks in advance


gdx/src/com/badlogic/gdx/math/Vector3.java
```diff
@@ -558,10 +552,7 @@ public boolean hasOppositeDirection (Vector3 vector) {
 
 	@Override
 	public Vector3 lerp (final Vector3 target, float alpha) {
-		x += alpha * (target.x - x);
-		y += alpha * (target.y - y);
-		z += alpha * (target.z - z);
-		return this;
+		return set(this.x + alpha * (target.x - x), this.y + alpha * (target.y - y), this.z + alpha * (target.z - z));
```
tommyettinger, line 555: This line inconsistently uses `this.x` and `x`; it can just use `x` throughout.

tommyettinger (commented):
I really am unsure about this. The use-case is quite narrow; it avoids overriding 4 methods in your immutable Vector3, but that's the only advantage. Modern JVMs probably can inline set() pretty well, but especially on GWT and Android there could be overhead from an extra method call on common operations. The problem here, and in many places, is that it is very hard to accurately benchmark Android and GWT performance. We can be fairly certain this won't speed anything up, but the question is if it will slow anything down, by any amount, on any platform (and Vector3 is so heavily used that any performance drop will be noticeable by someone). I think you should just bite the tiny bullet and override these 4 methods plus set() in your immutable Vector3.

tommyettinger (commented):
I really am unsure about this. The use-case is quite narrow; it avoids overriding 4 methods in your immutable Vector3, but that's the only advantage. Modern JVMs probably can inline set() pretty well, but especially on GWT and Android there could be overhead from an extra method call on common operations. The problem here, and in many places, is that it is very hard to accurately benchmark Android and GWT performance. We can be fairly certain this won't speed anything up, but the question is if it will slow anything down, by any amount, on any platform (and Vector3 is so heavily used that any performance drop will be noticeable by someone). I think you should just bite the tiny bullet and override these 4 methods plus set() in your immutable Vector3.

gdx/src/com/badlogic/gdx/math/Vector3.java
GadWissberg: I initialy wrote it without `this` but thought maybe we want to keep the original convention of using it. I'll fix it

GadWissberg: Alright yes I guess it doesn't worth the risk of affecting performance

HydrolienF:
It make code more readable & more extensible. I always use thing like that in my own code to reduce error when updating one function and not the other one & size of my .java

But I get that performances may be reduce ☹️

It would be nice to have some actual data about it to choose whether to increase function calls or duplicate code.

tommyettinger:
Vector3 (and Vector2) are so heavily-used that even small performance changes would likely be noticeable by someone. This sort of change would probably be welcome in places that don't have operations happening potentially hundreds or thousands of times per-frame. I would really like to be able to look at some benchmarks and say, "Wow, all platforms we target now inline method calls with no overhead!" but I don't think we're there yet. There's also the other issue that it really isn't very much work this saves, and it caters to a use-case in one application. If we are catering to use-cases in one application, I have a whole list of things I want to get into libGDX:

 - Standard-range colors ideally should either be primitives or immutable objects. Immutable objects would be the most API-compatible, but aren't great for GC.
 - Json should be able to handle Maps with non-String keys (on the Java side) by converting the keys to JSON Strings.
 - Way too much stuff, throughout the library, is private or package-private. If API stability is the only thing holding these back from being at least protected, mark them with an annotation to say the API is internal and can change in any commit.
 - Java 8. Come on. Even without the library, there are useful language features that work cross-platform. We aren't going to be able to target Java 7 at all much longer if using current JDKs; Java 20 can't target bytecode compatible with Java 7.
   -  Seriously. Come on. Even GWT can target language level 11 if using a current GWT release (like 2.10.0, which works just fine with libGDX if the backend is built for it).
 - A pony.

noblemaster:
The code improvement makes a lot of sense. That said, method calls on Android are **very slow,** especially for methods that are used a lot. This probably applies to `Vector3` here? I have a matrix multiplication class, where and removed most method calls resulting in a lot higher performance. Code-wise it's not as nice, but it's a lot faster.

I vote to reject the pull request (not because it's bad, but because it can be slow).

crykn:
The majority of contributors seem to be in favour of closing the PR, so closing it is. Thanks @GadWissberg for taking the time to propose the change, though!

PR #7153, opened 2023-05-26
Remove Android Build Tools version check from gdx-setup

Frosty-J:
#7064 is causing problems for users on Discord, and even myself. If you don't have Build Tools 33.0.2 installed, gdx-setup tells you to update it. It doesn't say *how* to update it, just that it needs updating. For the record: Tools -> [if on IDEA, Android] -> SDK Manager -> SDK Tools -> check Show Package Details -> check or uncheck versions as desired (before this PR, 33.0.2 must be checked) -> OK.

As the Android plugin (for IntelliJ IDEA, or baked into Android Studio) will automatically install the correct Build Tools version (or prompt them to accept the licence if they haven't already done so) I see no need for this check. Frankly, it shouldn't even matter whether someone has 33.0.2 or 33.0.1 - any minor version will work 99% as well as any other.

Tested by:

1. Uninstalling my copy of Build Tools 33.0.2
2. Attempting to generate a project from the current gdx-setup (no cigar)
3. Generating a project using these new changes
4. Opening the project
5. Confirming that 33.0.2 has been installed

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/SettingsDialog.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/SettingsDialog.java
@@ -88,7 +88,7 @@ public class SettingsDialog extends JDialog {
 			public void actionPerformed (ActionEvent e) {
 				if (offlineBox.isSelected()) {
 					int value = JOptionPane.showConfirmDialog(null,
-						"You have selected offline mode. This requires you to have your dependencies already in your maven/gradle cache.\n\nThe setup will fail if you do not have the correct dependenices already.\n\nDo you want to continue?",
+						"You have selected offline mode. This requires you to have your dependencies already in your maven/gradle cache.\n\nThe setup will fail if you do not have the correct dependencies already.\n\nDo you want to continue?",
 						"Warning!", JOptionPane.YES_NO_OPTION);
 					if (value == 0) {
 						onOK();
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetup.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetup.java
@@ -74,12 +74,6 @@ public class GdxSetup {
 			} else {
 				DependencyBank.buildToolsVersion = newestLocalTool;
 			}
-		} else {
-			if (!versionsEqual(localToolVersion, targetToolVersion)) {
-				JOptionPane.showMessageDialog(null,
-					"Please update your Android SDK, you need build tools: " + DependencyBank.buildToolsVersion);
-				return false;
-			}
 		}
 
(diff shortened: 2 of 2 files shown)


intrigus:
So why exactly is this change necessary? I may be missing something, but what is the problem with libgdx telling people to update the sdk?
What if people don't use IntelliJ or Android Studio? In that case they may not have the correct version installed, right?

Frosty-J:
Because people don't know how to update the SDK

> Hi i'm trying to use the latest GdxSetup tool but it complains that i don't have SDK tools 33.0.0. However, only 34.0.0-rc2 is available to download in Android Studio.

> when I want to create a LibGDX project, I get the message: "Update your Android SDK, you need build tools 33.0.2". However, the SDK is displayed in the SDK Manager of Android Studio. But I don't see version 33.0.2 on the computer (see pictures).

> when I want to create a LibGDX project, I get the message: "Update your Android SDK, you need build tools 33.0.2". However, the SDK is displayed in the SDK Manager of Android Studio. But I don't see version 33.0.2 on the computer (see pictures).

If someone is developing without a compatible IDE, I'm not sure whether Gradle will install the correct version or just error. It'll be the same as if someone were to download a project from the internet, which may not be configured to use the same versions as they already have installed. Since gdx-setup doesn't offer an alternative solution it can handle such as using a different version, I don't think it preventing project generation achieves anything.

tommyettinger:
> What if people don't use IntelliJ or Android Studio? In that case they may not have the correct version installed, right?

If people don't use IntelliJ or Android Studio, they are either doing Android builds from the command line or not at all. Eclipse, NetBeans, VS Code, etc. are not supported by the Android tools and haven't been for a long time.

obigu (approved):

extensions/gdx-setup/src/com/badlogic/gdx/setup/SettingsDialog.java
```diff
@@ -88,7 +88,7 @@ public SettingsDialog (final SetupCheckBox gwtCheckBox) {
 			public void actionPerformed (ActionEvent e) {
 				if (offlineBox.isSelected()) {
 					int value = JOptionPane.showConfirmDialog(null,
-						"You have selected offline mode. This requires you to have your dependencies already in your maven/gradle cache.\n\nThe setup will fail if you do not have the correct dependenices already.\n\nDo you want to continue?",
+						"You have selected offline mode. This requires you to have your dependencies already in your maven/gradle cache.\n\nThe setup will fail if you do not have the correct dependencies already.\n\nDo you want to continue?",
```
tommyettinger, line 91: Nice stealth typo-fix.

tommyettinger (approved):
This looks good to me, and should help stem the tide of cries for help we get about this.

PR #7154, opened 2023-05-28
Added JsonReader#stop() to stop parsing.

NathanSweet:
Stopping parsing can be useful when overriding methods for event driven parsing. For example when you only care about a small part of a large amount of JSON.

Event driven parsing is done by overriding the protected methods and writing code to handle the events, rather than letting the default behavior build up an object graph:
```
new JsonReader() {
	protected void startObject (String name) {
	}

	protected void startArray (String name) {
	}

	protected void pop () {
	}

	protected void string (String name, String value) {
	}

	protected void number (String name, double value, String stringValue) {
	}

	protected void number (String name, long value, String stringValue) {
	}

	protected void bool (String name, boolean value) {
	}
}.parse(json);
```


gdx/src/com/badlogic/gdx/utils/JsonReader.java
```diff
@@ -657,11 +659,17 @@ private static byte[] init__json_eof_actions_0 () {
 	static final int json_en_array = 23;
 	static final int json_en_main = 1;
 
-	// line 343 "JsonReader.rl"
+	// line 348 "JsonReader.rl"
 
 	private final Array<JsonValue> elements = new Array(8);
 	private final Array<JsonValue> lastChild = new Array(8);
 	private JsonValue root, current;
+	private boolean stop;
```
tommyettinger, line 667: It looks like stop is private and the only way to set it in a class that extends JsonReader is to call super.stop() or super.start(). Is there any use case you can think of where users might want deeper control over the starting/stopping process, or is this fine as private and handled purely by JsonReader?

tommyettinger (changes requested):
This is interesting; I've never done event-based parsing with JSON before, but it seems like being able to stop parsing is a pretty important requirement for it. I'm wondering if there should be at least a getter to allow user code to tell if `stop` is currently, well, stopped.

gdx/src/com/badlogic/gdx/utils/JsonReader.java
NathanSweet: It's pretty deep in the parser. I suppose it's possible for the user to care if the parser has been stopped. The stop signal comes from the user currently and I can't see it needing to come from inside the parser, so the user should know they've called `stop()`.

tommyettinger (approved):

PR #7167, opened 2023-06-17
Revert a change which used to rely on eager evaluation

ice1000:
Based on https://github.com/libgdx/libgdx/pull/7134/files#r1232921055

/cc @tommyettinger


tommyettinger (approved):
Just tested this change in an otherwise-identical gdx-setup project, and now it works fine. Hooray! Approved, let's get this merged.

PR #7172, opened 2023-07-02
For GL30 compatibility, use indexed lsans-15.png .

tommyettinger:
I don't know the specifics for why GL30 has a hard time with the Grayscale+Alpha color mode for PNGs. But, it doesn't seem to load that correctly, and does load Indexed color mode PNGs just fine, so switching to a (smaller) indexed-mode file should allow the default font to work on all GL versions. There's a tiny bit of lost detail, but this takes advantage of PNG's ability to have different alpha levels for every color in a palette, so there should still be a soft, not jagged, edge.


obigu (dismissed):
70% file size improvement. Approved!

obigu (commented):
I've just noticed, do we need to update the ones in tests-android as well? (lsans-15 and lsans-15_00)

tommyettinger:
I believe the ones in the tests only need to be updated if they are grayscale+alpha and used in GL30 mode. I don't think that currently happens, but maybe?

obigu:
> I believe the ones in the tests only need to be updated if they are grayscale+alpha and used in GL30 mode. I don't think that currently happens, but maybe?

I think you can configure/choose the mode when running the Lwjgl3 tests. I don't know about the grayscale+alpha thing.

obigu (approved):
there were quite a few apparently, thanks!

crykn (approved):

tommyettinger:
Is it OK if I merge my own PR? This has had 2 approvals for 3 days now. If there are other issues anyone has, this would be a good time to bring them up!

Frosty-J:
I approve because it fixes a problem. It seems unnecessary to me to avoid RGB PNGs here (the size difference is insignificant in the grand scheme) but I'm not too concerned with losing a few shades. Kindles are 4-bit grey, and text looks perfect on them.

PR #7174, opened 2023-07-03
Add flag to have a nodes translation, scale and rotation blended with its animation

tylerhasman:
This fixes a serious problem with ```AnimationController ``` which prevents a developer from blending custom transformations to nodes while they are animating. 

For example, in the game I am currently making, this is a character which has body sliders applied to it: 
image

However, when a animation is playing, some limbs regrow back to their normal proportions: (see the front leg)
image

This change adds a backwards-compatible flag to the ```Node``` class. The flag defaults to false so all existing projects will be unaffected.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
@@ -49,6 +49,9 @@ public class Node {
 	 * {@link #calculateWorldTransform()} **/
 	public final Matrix4 globalTransform = new Matrix4();
 
+	/** if true, the translation, rotation and scale will be blended with this node when its animating **/
+	public boolean blendWithAnimation = false;
+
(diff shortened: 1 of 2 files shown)


mgsx-dev:
For 1001 reasons, we shouldn't merge this : 
* order of transforms is arbitrary
* some may want to blend more than two transforms
* the API (Node class) was not designed for that : multiple transforms shouldn't be a concern at Node level.
* it kinda still the concept of node transform for a totally different use
* it adds unnecessary transforms when blending is disabled
* it breaks most of documentation about it

In other words, it's a quick and dirty approach (mostly dirty) that you could keep in your own fork and use in your game.

For a more general solution, you could either : 
* modify your animations programmatically, based on original animations and some transforms on top
* create a custom animation controller (inspired by the libgdx one) supporting multiple animation layers

tylerhasman:
> * create a custom animation controller (inspired by the libgdx one) supporting multiple animation layers

I will give this a go, thank you!

PR #7176, opened 2023-07-03
One-line fix for Tiled failures on GWT.

tommyettinger:
It seems like Tiled now uses reflection for more types than just MapObject in the `com.badlogic.gdx.maps` package, so this preemptively adds all of that package to the reflection cache, like how `com.badlogic.gdx.maps.objects` and `com.badlogic.gdx.maps.tiled.objects` already are added wholesale. It's possible that a more fine-grained set of classes could be added instead of the whole package, but it would just break again on GWT if any code used a class from outside that set.

Without this change, users can work around the bug by adding this to a .gwt.xml file:
```xml
	<extend-configuration-property name="gdx.reflect.include" value="com.badlogic.gdx.maps" />
```

The TiledMapObjectLoadingTest and SuperKoalio tests currently fail/crash if run on GWT without this PR (at least those). They pass with this PR in.

### DIFF
--- a/backends/gdx-backends-gwt/res/com/badlogic/gwtref/GwtReflect.gwt.xml
+++ b/backends/gdx-backends-gwt/res/com/badlogic/gwtref/GwtReflect.gwt.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit trunk//EN" "http://www.gwtproject.org/doctype/2.8.0/gwt-module.dtd">
+<!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit trunk//EN" "https://www.gwtproject.org/doctype/2.8.2/gwt-module.dtd">
 <module>
 	<inherits name='com.google.gwt.user.User' />
 	<define-configuration-property name="gdx.reflect.include"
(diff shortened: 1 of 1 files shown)


JamesTKhan (dismissed):
Seems reasonable to me to add the package "wholesale" in this case.

backends/gdx-backends-gwt/res/com/badlogic/gwtref/GwtReflect.gwt.xml
```diff
@@ -49,7 +49,7 @@
 		value="com.badlogic.gdx.Net" />
 
 	<extend-configuration-property name="gdx.reflect.include"
-		value="com.badlogic.gdx.maps.MapObject" />
+		value="com.badlogic.gdx.maps" />
```
crykn, line 52: Does this include sub-packages (and thus we could remove lines 53 to 56)?

tommyettinger: I'm pretty sure it does, since a line much earlier adds `com.badlogic.gdx.scenes.scene2d` to the cache, but that package has nothing in it directly, only sub-packages. I'll commit the change after I make sure it still runs the tests.

tommyettinger:
This isn't a one-line fix anymore, but it's still quite small. I verified @crykn 's suggestion, and it turns out subpackages are included if a parent package is. That means the subpackages of com.badlogic.gdx.maps are all included now, which unfortunately includes renderers also. I'm pretty sure renderers do not need to be cached, so a more fine-grained inclusion may be warranted, even if it just lists the likely classes that might need to be added (Map, MapGroupLayer, MapLayer, MapLayers, MapObject, MapObjects, and MapProperties) and the two subpackages separately.

crykn (approved):

MobiDevelop (approved):

tommyettinger:
Woohoo, it can be merged. Does anyone have an idea what changed in Tiled to cause this to be required?

PR #7177, opened 2023-07-04
Lower GLIBC requirements by using chromium reversion script and patches

theofficialgman:
lowers GLIBC requirements that were bumped to up to GLIBC 2.31 as mentioned here https://github.com/libgdx/libgdx/pull/7005

previously up to 2.29 was found to be required on multiple of the libgdx binaries on multiple architectures on linux
```
./gdx/libs/linuxarm64/libgdxarm64.so: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./gdx/libs/linuxarm64/libgdxarm64.so)
```

see PR for more info about the same changes at the Jamepad repo https://github.com/libgdx/Jamepad/pull/22

successful actions run: https://github.com/theofficialgman/libgdx/actions/runs/5449736745/jobs/9914278499

@MrStahlfelge @Tom-Ski no more waiting, lets get this merged! no risk since you can always drop the changes (or @ me if you have issues).

PR #7178, opened 2023-07-06
Lower GLIBC requirements by using building linux natives on ubuntu 18.04 docker

theofficialgman:
alternative to https://github.com/libgdx/libgdx/pull/7177

https://github.com/theofficialgman/libgdx/actions/runs/5470509920

I personally prefer the GLIBC reversion script since it allows us to continue using github actions runners (and all the dependencies that are pre-installed). This PR has the same minimum GLIBC requirements (GLIBC 2.17)

I would attempt using musl to build the debs which would have an advantage of being able to statically link (no glibc) however its very difficult to compile projects using its (since everything including a projects dependencies need to be compiled with musl) (see https://github.com/desktop/dugite-native/pull/502#discussion_r1253754491 for my struggles with anothe project doing that)

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -133,7 +133,6 @@ jobs:
           zip \
           unzip \
           xz-utils \
-          openjdk-11-jdk-headless \
           maven \
           ant sudo locales
 
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -113,7 +113,36 @@ jobs:
 
   natives-linux:
     runs-on: ubuntu-20.04
+    container:
+      image: ubuntu:18.04
     steps:
+      - name: Install dependencies into minimal dockerfile
+        run: |
+          # ubuntu dockerfile is very minimal (only 122 packages are installed)
+          # need to install updated git (from official git ppa)
+          apt update
+          apt install -y software-properties-common
+          add-apt-repository ppa:git-core/ppa -y
+          # install dependencies expected by other steps
+          apt update
+          apt install -y git \
+          curl \
+          ca-certificates \
+          wget \
+          bzip2 \
+          zip \
+          unzip \
+          xz-utils \
+          openjdk-11-jdk-headless \
+          maven \
+          ant sudo locales
+
+          # set Locale to en_US.UTF-8 (avoids hang during compilation)
+          locale-gen en_US.UTF-8
+          echo "LANG=en_US.UTF-8" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/build-publish.yml
```diff
@@ -113,7 +113,36 @@ jobs:
...
+          apt update
+          apt install -y git \
+          curl \
+          ca-certificates \
+          wget \
+          bzip2 \
+          zip \
+          unzip \
+          xz-utils \
+          openjdk-11-jdk-headless \
```
Berstanio, line 136:
```suggestion
```

Berstanio (changes requested):

.github/workflows/build-publish.yml
SimonIT: We could also just use a Docker image with java pre-installed https://hub.docker.com/_/eclipse-temurin

Berstanio: Yes, but then it gets inconsistent with the rest of the gha, which in my opinion is better to avoid.

theofficialgman: comitted the suggestion. you can just squash while merging

Berstanio (approved):

.github/workflows/build-publish.yml
```diff
@@ -113,7 +113,35 @@ jobs:
 
   natives-linux:
     runs-on: ubuntu-20.04
+    container:
+      image: ubuntu:18.04
     steps:
+      - name: Install dependencies into minimal dockerfile
+        run: |
+          # ubuntu dockerfile is very minimal (only 122 packages are installed)
+          # need to install updated git (from official git ppa)
```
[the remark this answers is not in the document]
Berstanio: Because the regular git version is for some reason very old and doesn't support submodules through the gha checkout action.

theofficialgman:
@Berstanio no issues have been brought up. please merge this and https://github.com/libgdx/Jamepad/pull/23

Tom-Ski (approved):

obigu:
It has the neccessary approvals, merging.

00-Evan:
Thank you very much for merging this. Just as a note, a similar change will probably be needed for jamepad, box2d, bullet, and freetype native binaries.

theofficialgman:
> Thank you very much for merging this. Just as a note, a similar change will probably be needed for jamepad, box2d, bullet, and freetype native binaries.

jamepad is the only (that I am aware of) in development library that requires natives that is currently affected, there is a PR open (see two posts above). eg: box2d has natives but has not been built since 2020 (github actions 16.04 was used to build it)

PR #7184, opened 2023-07-12
Automatically switch audio device if it was changed on the OS

rafaskb:
This simple PR checks if the default audio device set by the OS was changed, and if so, switches the output device internally using the "auto" mode.
Or in other words, if you launch your game, realize the sound is coming out from your speakers instead of your headphones, and change the default device on the OS (e.g. Windows), libGDX will understand that and redirect the output to the device you selected there.

This PR is very simple and works on the assumption that the active device is the first one in the list returned by `getAvailableOutputDevices()`. It was tested on Windows, but it should work everywhere as long as that assumption is held true elsewhere.

https://github.com/libgdx/libgdx/assets/2331058/4f01f0cf-585b-4b7f-9162-89d02edc70d9

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALLwjgl3Audio.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALLwjgl3Audio.java
@@ -159,14 +159,8 @@ public class OpenALLwjgl3Audio implements Lwjgl3Audio {
 						List<String> currentDevicesList = new ArrayList<>(Arrays.asList(currentDevices));
 						currentDevicesList.removeAll(Arrays.asList(lastAvailableDevices));
 						// If a new device got added, re evaluate "auto" mode
-						if (currentDevicesList.size() != 0) {
+						if (!Arrays.equals(currentDevices, lastAvailableDevices)) {
 							switchOutputDevice(null);
-						}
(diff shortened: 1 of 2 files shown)


Berstanio (changes requested):
I think the better approach is to change this:
https://github.com/libgdx/libgdx/blob/75612dae1eeddc9611ed62366858ff1d0ac7898b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALLwjgl3Audio.java#L159-L164
to this:
```java
if (!Arrays.equals(currentDevices, lastAvailableDevices)) {
      switchOutputDevice(null);
}
```
Just checking for a changed length was a oversight by me.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALLwjgl3Audio.java
```diff
@@ -159,9 +159,10 @@ public void run () {
 						List<String> currentDevicesList = new ArrayList<>(Arrays.asList(currentDevices));
```
Berstanio, line 159: Can you remove the now unnecessary list allocation/operation?

rafaskb:
Good catch! I think now we should be good then 😁

Berstanio (approved):

crykn (commented):
@rafaskb if you could add a CHANGES entry, that'd be perfect!

crykn (dismissed):

PR #7191, opened 2023-07-20
Fix Ctrl keycodes

### LINKED ISSUE #7189: Wrong Ctrl keycodes on Android
#### Issue details
As seen in 
https://github.com/libgdx/libgdx/blob/75612dae1eeddc9611ed62366858ff1d0ac7898b/gdx/src/com/badlogic/gdx/Input.java#L180
``CONTROL_LEFT`` and ``CONTROL_RIGHT`` are defined as ``129`` and ``130``, respectively. While these values are correct on desktop, Android uses different values for the Control (Ctrl) keys for some reason. As we can see in https://developer.android.com/reference/android/view/KeyEvent, the values for ``KEYCODE_CTRL_LEFT`` and ``KEYCODE_CTRL_RIGHT`` are ``113`` and ``114``. For this reason, ``Gdx.input.isKeyPressed(Keys.CONTROL_LEFT)`` and ``Gdx.input.isKeyPressed(Keys.CONTROL_RIGHT)`` do not work correctly on Android.

#### Reproduction steps/code
A simple program that handles keyboard events by displaying the keycodes is enough. The problem can be noticed when connecting a USB keyboard to an Android phone (with the help of a USB to USB Mini adapter if necessary) and pressing any of the two Ctrl keys.

1.12.0

- [X] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

M374LX:
Fixes #7189.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -177,8 +177,8 @@ public interface Input {
 		public static final int META_SHIFT_ON = 1;
 		public static final int META_SHIFT_RIGHT_ON = 128;
 		public static final int META_SYM_ON = 4;
-		public static final int CONTROL_LEFT = 129;
-		public static final int CONTROL_RIGHT = 130;
+		public static final int CONTROL_LEFT = 113;
+		public static final int CONTROL_RIGHT = 114;
 		public static final int ESCAPE = 111;
 		public static final int END = 123;
 		public static final int INSERT = 124;


Frosty-J (changes requested):
This will break other platforms. At least desktop, where I can confirm the Ctrls are 129 and 130.

Frosty-J (approved):
Sorry, I was too quick to react. It's probably okay, so long as everyone's using the constants rather than hardcoding the values.

https://github.com/libgdx/libgdx/blob/75612da/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java#L589-L590

https://github.com/libgdx/libgdx/blob/75612da/backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/DefaultIOSInput.java#L1162-L1163

https://github.com/libgdx/libgdx/blob/75612da/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java#L859-L860

mgsx-dev:
> It's probably okay, so long as everyone's using the constants rather than hardcoding the values.

Could be a big problem for users that store key mapping in a file. Maybe having a mapping for these keys in DefaultAndroidInput would be better no?

Frosty-J:
Depends how they store their mapping. If it's just the raw values and not a human-readable format, it's a minor inconvenience. 

```java
// Psuedocode of sorts for updating outdated key mapping
if (userKey == 129) userKey = Input.Keys.CONTROL_LEFT;
if (userKey == 130) userKey = Input.Keys.CONTROL_RIGHT;
```

Remapping in `DefaultAndroidInput` or somewhere would indeed be an option. I'd rather libGDX stick to some sort of standard, as it seems unnecessarily complicated not to.

M374LX:
We could add a note about storing key mapping in a file to ``CHANGES``.

An alternative is to keep the current keycodes and add a check like this one to ``DefaultAndroidInput``'s ``onKey()`` method:

```Java
int keycode = e.getKeyCode();

if (keyCode == 113) {
    keyCode = Input.Keys.CONTROL_LEFT;
} else if (keyCode == 114) {
    keyCode = Input.Keys.CONTROL_RIGHT;
}
```

MrStahlfelge:
If someone maps the key values by configuration, it should be no problem as the real key values were saved.

MrStahlfelge (dismissed):

crykn (approved):

Berstanio:
> If someone maps the key values by configuration, it should be no problem as the real key values were saved.

I don't quite understand, how this does not break. If I write `129` as a key mapping in a config file and compare that "old" value for CONTROL_LEFT (129) to a key press event, which has the "new" value for CONTROL_LEFT (113), it will not work, right?
So I would need to write migration logic, as I understand it?

MrStahlfelge:
Usually you write custom key mapping configuration in a way that users press the keys they want. The bug here is that a constant in our framework has the wrong value. Independent from that, if a user presses a key, the system gives the real value and not a wrong value from our framework.
After we fix the wrong value in our framework, if a user presses a key, the system still gives the correct value. So no migration is needed.

Berstanio:
How I understand it is the following:
The user presses a key and the backend specific key code gets converted to a gdx keycode, which gets propagated to the listeners, like here:
https://github.com/libgdx/libgdx/blob/75612dae1eeddc9611ed62366858ff1d0ac7898b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java#L137-L138
I then receive the gdx keycode the user pressed for the custom mapping and write it to a file. After that I can compare the saved gdx keycode to the pressed gdx keycode. If now the pressed gdx keycode changes, this wont work anymore in my understanding.

I can test this pr someday later.

> If you really wrote the old value 129 in your mapping file, the mapping just did not work before the change and after the change. At least it did not work when the user pressed Ctrl key

I just tested it, mapping the Ctrl key works on desktop.

MrStahlfelge:
My fault, I only thought about Android and forgot that all other platforms use the constants for internal mapping. Sorry.

M374LX:
In fact, I did not think about the key mappings being saved in a file while preparing this PR.

On platforms other than Android, there might be problems when saving the old values (129 and 130) from an old libGDX version, upgrading libGDX, and then loading the values, which will no longer correspond to those of ``Keys.CONTROL_LEFT`` and ``Keys.CONTROL_RIGHT``.

Does anyone prefer the approach suggested in https://github.com/libgdx/libgdx/pull/7191/#issuecomment-1646330424?

PR #7196, opened 2023-07-25
Remove usage of deprecated tag on GL30

JamesTKhan:
Based on the linked issue, I have removed the use of deprecated tag on the useGL30 variables for Android, RoboVm, Metal-Angle, GWT. I also removed the experimental comment, but if any one feels some of these should still say experimental ( I don't know much about the IOS side) I can add that back in.

The original issue focused on Android, but I agree with MGSX that using deprecated is not appropriate even if we want to say it is experimental.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
@@ -91,11 +91,11 @@ public class AndroidApplicationConfiguration {
 	/** set this to true to enable Android 4.4 KitKat's 'Immersive mode' **/
 	public boolean useImmersiveMode = true;
 
-	/** Experimental, whether to enable OpenGL ES 3 if supported. If not supported it will fall-back to OpenGL ES 2.0. When GL ES
+	/** Whether to enable OpenGL ES 3.0 if supported. If not supported it will fall-back to OpenGL ES 2.0. When GL ES
 	 * 3* is enabled, {@link com.badlogic.gdx.Gdx#gl30} can be used to access its functionality. Requires at least Android 4.3 (API
 	 * level 18).
-	 * @deprecated this option is currently experimental and not yet fully supported, expect issues. */
-	@Deprecated public boolean useGL30 = false;
+	 */
+	public boolean useGL30 = false;
 
 	/** The maximum number of threads to use for network requests. Default is {@link Integer#MAX_VALUE}. */
 	public int maxNetThreads = Integer.MAX_VALUE;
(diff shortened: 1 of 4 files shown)


obigu:
Do we know why they were set to experimental in the first place? "Experimental" suggests it may be buggy, a use at your own risk warning. Has it been tested on Android and iOS to confirm it's stable now?

JamesTKhan:
> Do we know why they were set to experimental in the first place? "Experimental" suggests it may be buggy, a use at your own risk warning. Has it been tested on Android and iOS to confirm it's stable now?


I do not know the original reason for setting it as "Experimental". My assumption, at least for the Android side, was that not as many devices supported it at the time, but that could be incorrect assumption. I cannot speak for IOS or test it out as I do not own hardware for that.

I do have a 2017 model Note 8. I ran the GL30 tests on it, here are the results

- VBOWithVAOPerformanceTest - Pass
- MultipleRenderTargetTest - Pass
- TextureArrayTest - Pass
- GL30Texture3DTest - Pass
- InstancedRenderingTest - Pass
- ModelInstancedRenderingTest - Pass

- UniformBufferObjectsTest - Runs but only after applying updates for GLES 3.0 to Shader prefixes
- FloatTextureTest - Runs but only after applying updates for GLES 3.0 to Shader prefixes
- PixelBufferObjectTest - Runs but only after applying updates for GLES 3.0 to Shader prefixes ( looks a little funky though)

obigu:
If on Android we are making assumptions and need tweaks on several tests and on iOS it has not been tested I think it's prudent not to remove the "Experimental" warning. I would even make it more noticeable considering the Deprecated annotation will no longer "warn" about it (which I'm ok with).

Berstanio:
> iOS it has not been tested I think

I think at least MetalANGLE has been used with `useGL30`. MetalANGLE mentiones `90%+ of OpenGL ES 3.0 conformance tests passed`, but I honestly don't know whether this is a lot or not. Official ANGLE seems to has it completed, but the upgrade is not easy.

JamesTKhan:
> If on Android we are making assumptions and need tweaks on several tests and on iOS it has not been tested I think it's prudent not to remove the "Experimental" warning. I would even make it more noticeable considering the Deprecated annotation will no longer "warn" about it (which I'm ok with).
> 
> 

The tweaks are not a fault of the GL30 implementation, they are a fault of the tests just not being configured properly for it. 

That being the case I am fine with bringing the experimental message back in.

obigu:
> The tweaks are not a fault of the GL30 implementation, they are a fault of the tests just not being configured properly for it.

Ok, do you think they should be updated or somehow document how to run them on GL30?

JamesTKhan:
> 
> > The tweaks are not a fault of the GL30 implementation, they are a fault of the tests just not being configured properly for it.
> 
> Ok, do you think they should be updated or somehow document how to run them on GL30?
> 

Updated, I am going to do a separate PR for them.

crykn (approved):

PR #7200, opened 2023-07-30
Make `Button`#`setChecked(boolean isChecked, boolean fireEvent)` public

Frosty-J:
I need to be able to override this method for customising button behaviour when clicked.


raeleus:
I think exposing fireEvent is not good. Typically users will call setProgrammaticChanges(false) and do the weird stuff they need to do. This change bypasses that setting which will cause confusion. If you're trying to add some logic to prevent the button from being checked, the currently prescibed technique is to add it to a ButtonGroup and override ButtonGroup#canCheck()

PR #7207, opened 2023-08-12
align for TiledDrawable

TCreutzenberg:
This PR introduces align for TiledDrawable. Before it was always aligned bottom left.
Also works if the area to draw is smaller than the texture to draw.

To verify the result just run TiledDrawableTest: A+D keys change the scale, mouse movement will change the width+height of the area to render.

tiledDrawableTest01
tiledDrawableTest02


tommyettinger (approved):
Wow, this is a complete PR. There's what seems like a pretty hefty addition to TiledDrawable, plus you've already got tests, with data too! I ran the test locally, and it seems good; each alignment seems to be correct now. I'm somewhat curious why the code added in TiledDrawable is so much longer than what was there before, though this is probably just what happens when an extra feature gets implemented. Approved!

tommyettinger:
I requested a review from NathanSweet because I don't think anyone else knows Drawable code as well, and there might be some things that could be cleaner (that I would miss every time).

NathanSweet:
I haven't been able to check this out yet, sorry, I hope to soon.

tommyettinger (approved):
The latest edits are good, and don't change my approval. We don't have to update the code with merges every time because it will merge with a squash when the PR is accepted, but there was a change here, so it makes sense to do the merge when you did.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
```diff
@@ -43,50 +46,226 @@ public void draw (Batch batch, float x, float y, float width, float height) {
...
-			if (remainingY > 0) {
-				v = v2 - remainingY / (texture.getHeight() * scale);
-				batch.draw(texture, x, y, remainingX, remainingY, u, v2, u2, v);
+
+			// Left center partials
+			if (fullY == 0 && Align.isCenterVertical(align)) {
+				final float vOffset = 0.5f * (v2 - v) * (1f - (height / regionHeight));
+				final float leftCenterV = v2 - vOffset;
+				final float leftCenterV2 = v + vOffset;
+				for (int i = 0; i < 1; i++) {
```
NathanSweet, line 127: This is either a bug (loop can only execute once) or the loop can be removed. This occurs 4 times in this file.

NathanSweet (changes requested):
Overall it looks good! Just the one comment below to address before merge.

There's so many edge cases to handle. I wish there was some elegant way of doing it all without handling each edge case individually, but I'm not sure that's possible.

Inconsequential: I don't usually bother with `final` on locals, just a little more text to read. I usually prefer `else if` over `else { if`.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/TiledDrawable.java
tommyettinger: It seems to behave correctly, so it looks like it's just an extraneous loop around code that should be kept.

TCreutzenberg: Good catch, thank you! I removed the redundant loops.

NathanSweet: Sorry for the delay. I still see the loops? (`for (int i = 0; i < 1; i++) {`) Maybe you didn't push?

TCreutzenberg: Sorry for that! I just pushed again. Most likely something didn't work properly as we are moving houses at the moment and I'm working from another laptop. Github also wanted me to confirm strange stuff for access this time...

NathanSweet: Thanks! Sorry for the delay. For some reason I don't get notifications about this PR, despite being subscribed to all events.

PR #7213, opened 2023-08-19
Fix some casts in ParticleEditor and Flame.

tommyettinger:
This is just a small fix for what seems to be a crash in ParticleEditor; it looks like it affects Flame as well. The issue stems from:
 - We have a `NumericValue` that stores a `float`.
 - We get that float and assign it to a `JSpinner` with a `SpinnerNumberModel`.
   - Inside the `JSpinner`, it is promoted and assigned to either a `double` or a `Double`, I'm not sure which.
 - We later get a value from the `JSpinner`, which gets returned as a boxed `Double`.
 - We try to cast this to `Float` so it can be auto-unboxed.
   - This causes a crash! `Double` cannot be cast to `Float`, even though their primitive forms can.

In this PR, we cast to `Number` instead of `Float`, and call `.floatValue()` on it. This is a pattern already used heavily in Hiero.

There are no other changes in this PR. Happy reviewing!

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/NumericPanel.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/NumericPanel.java
@@ -59,7 +59,7 @@ class NumericPanel extends ParticleValuePanel<NumericValue> {
 		}
 		valueSpinner.addChangeListener(new ChangeListener() {
 			public void stateChanged (ChangeEvent event) {
-				NumericPanel.this.value.setValue((Float)valueSpinner.getValue());
+				NumericPanel.this.value.setValue(((Number)valueSpinner.getValue()).floatValue());
 			}
 		});
 	}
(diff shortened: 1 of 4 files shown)


crykn (approved):

JamesTKhan (approved):
Testing on Flame side, this does resolve the exceptions. 

I Found a new odd issue though, now that the spinner is working, if I change a spinner value, click on a different window, come back to Flame window and click the spinner arrows again, it has a chance of completely locking up, not exceptions, no nothing. Debugging did not give me much to go on. 

However, I rolled back libGDX to 1.11.0 locally, and I can replicate the same locking up issue on window switching so this PR is not to blame as it seems to be a separate existing issue so, approved.

tommyettinger:
> it seems to be a separate existing issue

Wow, this really is a buggy tool... and no one seems to be using it if these bugs never get reported...

raeleus:
> > it seems to be a separate existing issue
> 
> Wow, this really is a buggy tool... and no one seems to be using it if these bugs never get reported...

Indeed. All the preview options are broken too and noone has ever pointed it out.

PR #6821, opened 2022-03-15
Catch the cursor on Android

Frosty-J:
Have you ever wanted to play a first-person game using a mouse on Android? Yeah, me neither.

Test procedure:

* Connect a mouse and keyboard to an Android device on Android 8 or newer. I achieved this with an OTG cable and USB hub.
* Launch `InputTest`.
* See if everything seems sane before and after hitting `G`. Also after the app loses focus.

Obviously would be better to test with a real game, but it is what it is.

Resolves #6677.

PR #7215, opened 2023-08-22
Revert #6821

obigu:


Tom-Ski (approved):

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -933,8 +933,8 @@ public enum Orientation {
 		Landscape, Portrait
 	}
 
-	/** Only viable on desktop, GWT and Android 8+. Will confine the mouse cursor location to the window and hide the mouse cursor.
-	 * X and y coordinates are still reported as if the mouse was not catched.
+	/** Only viable on the desktop. Will confine the mouse cursor location to the window and hide the mouse cursor. X and y
```
tommyettinger, line 936: It looks like GWT hasn't changed here, but the mention of being viable on GWT was removed anyway. Was it incorrect before?

tommyettinger (approved):
I'm not sure about one comment, but it's not a big issue. Drag motions failing on Android is a bigger issue.

PR #7225, opened 2023-09-16
Add comment to Input.Keys documentation for ⌘ Command key

mwhiter:
I couldn't find an online reference for Mac-specific keycodes, so adding this comment to hopefully aid anyone who develops on MacOS for LibGDX in the future.

I'm aware this is probably a very niche group, but I think having some documentation would be helpful here anyway.


tommyettinger (approved):
Simple, informative, looks good to me. It also passes formatting checks it looks like, so there's nothing blocking it. Approved!

Issue #6901, opened 2022-06-15
Support Tiled "object" type MapProperty 


Faust90:
gdxVersion = '1.11.0'

Since Tiled 1.4 a Tiled Object can have custom properties of type "object", which can be used to reference another object inside that map.

It would be nice to have support of such feature in LibGDX Tiled libraries.

### STACKTRACE
Exception in thread "LWJGL Application" com.badlogic.gdx.utils.GdxRuntimeException: Wrong type given for property triggeredBy, given : object, supported : string, bool, int, float, color
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.castProperty(BaseTmxMapLoader.java:446)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadProperties(BaseTmxMapLoader.java:425)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:409)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:313)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObjectGroup(BaseTmxMapLoader.java:250)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadLayer(BaseTmxMapLoader.java:174)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadTiledMap(BaseTmxMapLoader.java:161)
	at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:81)
	at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:59)
	at com.faust.lhitgame.game.rooms.impl.FixedRoom.loadTiledMap(FixedRoom.java:52)

spencerflem:
just ran into this as well

JohnLavender474:
Hey, I'd like to take a try at this issue. Could someone assign me to it?

JohnLavender474:
Could you provide an example of how you're trying to use this? Are you wanting to access just the properties of another object?

Faust90:
[...]

PR #7229, opened 2023-09-20
6901 support tiled object type map property

JohnLavender474:
This PR is in response to issue [6901 Support Tiled "object" type MapProperty](https://github.com/libgdx/libgdx/issues/6901). The request is that the "object" property type in Tiled be supported in LibGDX. Here, my solution is to use the MapProperties of the object being pointed to as the value of the "object" type property. 

In the [BaseTmxMapLoader](https://github.com/JohnLavender474/LibGDX--Support-Tiled-object-type-MapProperty-6901/blob/6901_Support-Tiled-object-type-MapProperty/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java) class, the properties of objects are loaded as normal except for a few differences. 

The first difference is that mappings are added to an ObjectMap instance where the key is the id of the object and the value is the MapProperties of the object. 

The second difference is that when the type of a property is "object", then a Runnable instance is created where in the run method the MapProperties of the object being pointed to is added as a property. The Runnable is added to a queue, and the Runnables are not run until the end of the loadTiledMap method. Doing it this way ensures that all MapProperties are loaded before attempting to fetch them.

The [TiledMapObjectPropertyTest](https://github.com/JohnLavender474/LibGDX--Support-Tiled-object-type-MapProperty-6901/blob/6901_Support-Tiled-object-type-MapProperty/tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectPropertyTest.java) class tests that "object" properties are successfully loaded. The test includes ensuring that a circular graph does not lead to infinite recursion. Feel free to open [the tmx file](https://github.com/JohnLavender474/LibGDX--Support-Tiled-object-type-MapProperty-6901/blob/6901_Support-Tiled-object-type-MapProperty/tests/gdx-tests-android/assets/data/maps/tiled-objects/test-object-properties.tmx) to ensure proper testing.


JohnLavender474:
@SimonIT Thanks for looking over my PR. How many reviews should it receive before it's greenlit to merge into the master branch?

SimonIT:
Our rule of thumb is two reviews per PR. Good luck finding someone who reviews the PR. We are searching for some time now for someone who reviews tiled PRs

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectPropertyTest.java
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	@Override
+	public void dispose () {
+		map.dispose();
+		shapeRenderer.dispose();
+	}
+
+	private void test (MapProperties props, int idToObjProp, ObjectMap<Integer, MapProperties> idPropMap) {
```
tommyettinger, line 177: This would probably also need changing if you do switch the other `ObjectMap<Integer, MapProperties>` to use an IntMap.

tommyettinger (changes requested):
Looks pretty good; I have some minor suggestions, but I don't think anything should be hard to implement. I'm not really sure what our variable name code style rules are, other than I don't see underscore prefixes in many places in libGDX. I think switching to IntMap is only really something I will push for in BaseTmxMapLoader.java ; I think it's a good idea in the test but isn't necessary.

Yay, Queue! Another poorly-named class with a nice implementation! I just wanted to comment that I hadn't seen it used much, and I'm glad it is getting used now.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectPropertyTest.java
```diff
@@ -0,0 +1,190 @@
...
+			map = new TmxMapLoader().load("data/maps/tiled-objects/test-object-properties.tmx");
+			Gdx.app.log("TiledMapObjectPropertyTest", "Finish loading map!");
+
+			batch = new SpriteBatch();
+			shapeRenderer = new ShapeRenderer();
+			mapRenderer = new OrthogonalTiledMapRenderer(map);
+
+			MapObjects objects = map.getLayers().get("Objects").getObjects();
+
+			ObjectMap<Integer, MapProperties> idPropMap = new ObjectMap<>();
```
tommyettinger, line 68: I forgot to post that this is the other usage of ObjectMap with Integer keys, and could be an IntMap.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -430,23 +441,59 @@ protected void loadObject (TiledMap map, MapObjects objects, Element element, fl
...
+					/*
+					 * Wait until the end of the [loadTiledMap] method to fetch the map props for "object". This is because the map
+					 * properties of the object being pointed to might not have been loaded yet.
+					 */
+					try {
+						/*
+						 * Value should be the id of the object being pointed to. Nevertheless, the try-catch block safeguards in case
+						 * faulty data in inputted.
+						 */
+						final int _value = Integer.parseInt(value);
```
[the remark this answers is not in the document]
JohnLavender474: Good suggestion. I changed the name to id. And yeah, I'm glad to make use of the LibGDX collections! I recently wrote a simple game engine in Kotlin built on LibGDX. One thing on my ToDo list is to go back through that project and swap out all the Java collections with LibGDX ones. The optimization is definitely worth the effort!

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -64,6 +65,10 @@ public static class Parameters extends AssetLoaderParameters<TiledMap> {
 	protected int mapHeightInPixels;
 
 	protected TiledMap map;
+	/** MapProperty instances are mapped to the id of the respective tiled object */
+	protected ObjectMap<Integer, MapProperties> idToPropsMap;
```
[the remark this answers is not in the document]
JohnLavender474: Changed the usages of ObjectMap to IntMap. Again, good suggestion! Never hurts to get things as optimized as possible.

tommyettinger (approved):

JohnLavender474:
@tommyettinger Didn't mean to dismiss your review. I wanted to make a few more changes. I renamed the map in BaseTmxMapLoader to "objectIdToPropsMap" to make it clear that the map ONLY contains [id, MapProperties] entries for objects (and not tiles, animated tiles, or the Tiled map). Also, I cleaned up the logs in the TiledMapObjectPropertyTest file.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectPropertyTest.java
```diff
@@ -135,9 +133,11 @@ public void create () {
 						value = "props of object with id " + nestedProps.get("id", Integer.class);
 					}
 
-					System.out.println("\t" + key + " = " + value);
+					builder.append("\t").append(key).append(" = ").append(value).append("\n");
```
tommyettinger, line 136: This is definitely cleaner than mixing log() and println() calls.

It doesn't matter here at all, but when anyone else reviews this, I'm wondering when the StringBuilder class in libGDX is considered better than StringBuilder in the JDK. The JDK one has a number of intrinsics, if I remember it right, and also has better support from IDE tooling. The libGDX one supports equals() and hashCode().

Also, who keeps coming up with duplicate names? Array, Queue, and StringBuilder are all long-existing names from the JDK. At least StringBuilder actually correctly describes what the class is. `Array<T>` is much closer to an `ArrayList<T>` than a `T[]`, while `Queue` is just as much a stack as it is a queue, and can almost implement `Deque` from the JDK (for double-ended queues). Even AssetManager has some confusing code in AndroidFileHandle that uses Android's AssetManager class and not the libGDX one.

tommyettinger (dismissed):
Looks good; I like the more consistent logging (without println() mixed in). The new name is also better for the IntMap. Approved!

JohnLavender474:
@NathanSweet Whenever you have time, would you mind looking over this PR to see if it's good?

tommyettinger:
Looks like you need to run spotlessApply again; something changed in formatting somewhere and spotless doesn't like it.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapObjectPropertyTest.java
NathanSweet: StringBuilder collects chars and exposes the array, it has hascode, equals, and a couple utility methods. IIRC the JRE class stores bytes and doesn't expose anything, so you're allocating to get the chars/string out, which you need to do for hashcode or equals. Newer JVMs don't use StringBuilder at all for string concat, though there's a CLI option to enable it.

Names are hard and all the good names are taken! Short names are nice, but less descriptive.

NathanSweet:
Tiled is one of the areas I'm unfamiliar with, having never looked in there nor implemented an app using it. I can do a once over though! I'll dump some opinions or just notes, feel free to disagree, no worries.

* `objectIdToPropsMap` I'd drop the type in the name -> `objectIdToProps`. If `objectId` is not more descriptive in a Tiled context I'd use `idToProps`.
* I guess `runOnEndOfLoadTiledMap` and `objectIdToPropsMap` can be null at the end of `loadTiledMap`.
* Diamond operator on RHS doesn't do anything.
* Multiline comments sure take up a lot of lines when the start and end of the comment get their very own line. I'd just use `//` and let the IDE wrap it at 130 columns.
* @Override on JRE classes (Runnable) isn't useful, they are never going to make a breaking change.
* Queue is cool but you `addFirst` and `removeFirst` (LIFO), would be the same with Array `add` and `pop`. Could just iterate and null `runOnEndOfLoadTiledMap` afterward.
* The runnables seem OK. Another way could be to collect properties, name, id triplets but it'd be equivalent.
* Why catch and log? Why not catch and throw? If I need object properties and they fail, my app should probably die. Reacting to the log message is difficult.
* Log or exception could include the invalid value.
* MapObject vs MapProperties, I'm afraid I'm no help there.
* Storing ID to properties for everything may be a lot of storing. That's a little extra work and memory during loading only, seems pretty fine. I'm guessing there's no existing "get by ID" mechanism this could use?

Overall it seems good. The nulling and logging items may warrant revision.

tommyettinger:
> * Queue is cool but you `addFirst` and `removeFirst` (LIFO), would be the same with Array `add` and `pop`. Could just iterate and null `runOnEndOfLoadTiledMap` afterward.

`Array.add()` is the same as `addLast()`, no? Adding at the start of an `Array` means every existing element needs to get shifted over, but for a `Queue` it just changes where the head is.

The implementation for `Queue` is nice and I learned a lot reading it. I also figured out how to make a similar class (also a double-ended queue) have O(1) random access; I can't remember if `Queue` had this or had part of this, but [my ObjectDeque class](https://github.com/tommyettinger/jdkgdxds/blob/master/src/main/java/com/github/tommyettinger/ds/ObjectDeque.java#L1184-L1201) allows getting and setting elements by index in constant time. Removing and inserting in the middle are O(n), like `Array`.

JohnLavender474:
@NathanSweet @tommyettinger After thinking about it for a while, I decided that instead of fetching MapProperties, the "object" type should return MapObject instead. It make more sense that users would want to have direct access to the MapObject being pointed to. I changed the code in the BaseTmxMapLoader and included a public method for users to fetch a copy of the id-to-object map. And the test class has been changed too; it includes better logging and repeats the test 3 times to make sure that reusing the same TmxMapLoader instance doesn't present any problems. I edited the Tiled map to include two object layers to ensure that objects referencing each other across layers doesn't cause issues.

@NathanSweet I made most of the changes you suggested. Just a few minor quibbles: I consider it good practice to include the diamond operator and the @Override tag.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -88,6 +86,16 @@ public Array<AssetDescriptor> getDependencies (String fileName, FileHandle tmxFi
 		return getDependencyAssetDescriptors(tmxFile, textureParameter);
 	}
 
+	/** Gets a map of the object ids to the {@link MapObject} instances. The returned map is a copy of the internal map. Returns
+	 * null if {@link #loadTiledMap(FileHandle, Parameters, ImageResolver)} has not been called yet. The internal map is reset when
+	 * the load tiled map method is called.
+	 *
+	 * @return the map of the object ids to the {@link MapObject} instances, or null if the load tiled map method has not been
+	 *         called yet. */
+	public IntMap<MapObject> getIdToObject () {
```
SimonIT, line 94: We have a `@Null` annotation for nullable return values

JohnLavender474: Added the `@Null tag`

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -83,6 +85,16 @@ public Array<AssetDescriptor> getDependencies (String fileName, FileHandle tmxFi
...
 	}
 
+	/** Gets a map of the object ids to the {@link MapObject} instances. The returned map is a copy of the internal map. Returns
+	 * null if {@link #loadTiledMap(FileHandle, Parameters, ImageResolver)} has not been called yet. The internal map is reset when
+	 * the load tiled map method is called.
+	 *
+	 * @return the map of the object ids to the {@link MapObject} instances, or null if the load tiled map method has not been
+	 *         called yet. */
+	public @Null IntMap<MapObject> getIdToObject () {
+		return idToObject == null ? null : new IntMap<>(idToObject);
```
NathanSweet, line 95: Why a copy?

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -431,22 +450,44 @@ protected void loadObject (TiledMap map, MapObjects objects, Element element, fl
...
+				if (type != null && type.equals("object")) {
+					// Wait until the end of [loadTiledMap] to fetch the object
+					try {
+						// Value should be the id of the object being pointed to
+						final int id = Integer.parseInt(value);
+						// Create [Runnable] to fetch object and add it to props
+						Runnable fetch = new Runnable() {
+							@Override
+							public void run () {
+								MapObject props = idToObject.get(id);
```
NathanSweet, line 477: props -> object

NathanSweet (dismissed):
Just those comments, looks good to me.

NathanSweet:
> Array.add() is the same as addLast(), no? Adding at the start of an Array means every existing element needs to get shifted over, but for a Queue it just changes where the head is.

@tommyettinger Right, but this usage was Queue `addFirst` and `removeFirst`, so last-in-first-out. Array `add` and `pop` is also last-in-first-out, so functionally the same (and maybe the teeniest bit more familiar to most devs).

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
JohnLavender474: It returns a copy so that the internal state of the map can't be changed from the outside. The user should call this method only once each time the loadTiledMap method is called so that it doesn't incur much runtime penalty.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
JohnLavender474: Good catch

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
JohnLavender474: I don't have a strong opinion about this though if you think it should just return the internal map instead of a copy

JohnLavender474: I think you make a great argument for why it shouldn't return a copy. My thought was to protect the internal state, but after reading your explanation, I agree there's not much point since the map is reallocated every time the load tiled map method is called. I've changed it.

tommyettinger (dismissed):
I don't see anything wrong here, myself. The CHANGES file needs to have its conflict resolved, which should be possible to do in the web editor or your Git client of choice. I think we can get this merged before the next release (or very soon afterward if the release happens sooner than I'm thinking it will); this PR has been checked over lots already.

JohnLavender474:
> I don't see anything wrong here, myself. The CHANGES file needs to have its conflict resolved, which should be possible to do in the web editor or your Git client of choice. I think we can get this merged before the next release (or very soon afterward if the release happens sooner than I'm thinking it will); this PR has been checked over lots already.

@tommyettinger  CHANGES file should be resolved now. Awesome to here it might get merged! I'd like to make contributions to open-source projects something I do on a regular basis.

tommyettinger (approved):
Just reiterating my approval now that there are no conflicts.

PR #7233, opened 2023-09-23
"Dokumentation" Merge Rectangle

lukasb1b:
making rectangle merge more readable by adding this


Frosty-J:
That raises the issue of what else should be `this`'d, such as `merge(Vector2[] vecs)`. It matches the style of some of the other methods, so I'm for it even though I don't use `this` much myself.

lukasb1b:
Well your right but I just saw it specificly in that function. And I got confused as sometimes x and y are used for variable names. But here is a vector

crykn (approved):

PR #7234, opened 2023-09-23
Allow to disable SelectBoxScrollPane reset selection on exit behavior

fgnm:
There's a default behavior for `SelectBoxScrollPane` which is quite confusing on UX point of view. When de cursor exit from the scroll pane the selection is reset to the current value which create a bad UX, it's common to exit the scroll pane by mistake and this create confusions:

https://github.com/libgdx/libgdx/assets/5543339/eb58353d-d943-44f3-b936-fcb5d2cffc28

With this PR there's now an option to disable the default behavior:

https://github.com/libgdx/libgdx/assets/5543339/670741ec-9da7-4282-a114-e88bef936c1d

Similar to the behavior of any other software:

https://github.com/libgdx/libgdx/assets/5543339/2a0309c9-c472-4321-a78a-aa04845f0052

(Dunno why I can't record the cursor, btw it enter end exit from the scroll pane)

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SelectBox.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/SelectBox.java
@@ -426,6 +426,7 @@ public class SelectBox<T> extends Widget implements Disableable {
 		final List<T> list;
 		private InputListener hideListener;
 		private Actor previousScrollFocus;
+		boolean resetSelectionOnExit = true;
 
 		public SelectBoxScrollPane (final SelectBox<T> selectBox) {
 			super(null, selectBox.style.scrollStyle);
@@ -458,7 +459,7 @@ public class SelectBox<T> extends Widget implements Disableable {
 
 			addListener(new InputListener() {
 				public void exit (InputEvent event, float x, float y, int pointer, @Null Actor toActor) {
-					if (toActor == null || !isAscendantOf(toActor)) {
+					if ((toActor == null || !isAscendantOf(toActor)) && resetSelectionOnExit) {
 						T selected = selectBox.getSelected();
 						if (selected != null) list.selection.set(selected);
 					}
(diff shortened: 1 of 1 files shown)


fgnm:
I don't know why it was made in this way, but I didn't want to make it as a break change... There are years that SelectBox works in this way and no one never complained

crykn:
I agree with @Berstanio, I don't think we need to keep the old behaviour around. Maybe @NathanSweet has an opinion on this?

NathanSweet:
The premise for the proposed change is:

> is quite confusing on UX point of view. When the cursor exit from the scroll pane the selection is reset to the current value which create a bad UX, it's common to exit the scroll pane by mistake and this create confusions

This is not descriptive enough. Please explain why you feel it's confusing.

When a select box list is shown, the select box value is selected in the list by default. To highlight which item is hovered, the list selection is changed to the hovered item. If the mouse is moved off the list, what is the point of keeping the last hovered item selected? It was selected solely to indicate the hovered item, but the hover is gone. Keeping it selected doesn't provide any useful information. If it was kept, it should not be used to set the select box value if the select box is closed by clicking away. it would be used if enter is pressed, but I don't think it's common to use a combination of mouse and keyboard to choose an item.

Some select box implementations keep the last hovered item selected, but I don't see that as an improvement over the current implementation. Showing hover state via list selection is OK, but otherwise I prefer to see the select box value selected. While that value can be seen in the select box itself, it's often helpful to see it selected in the list. In many cases the list is sorted or otherwise items near the selected item are important. It would not be great to lose where the select box value is in the list because you accidentally hovered some other item briefly.

When using the keyboard then it makes sense to keep the selected list item something other than the select box value. scene2d.ui's select box allows keyboard input to change the value without needing to keep the hovered item selected after the hover is gone.

I think an argument can be made for either approach, but I prefer it as implemented.

If we want to discuss a more substantial improvement in this area, I would suggest having a separate list highlight for hover than for the select box value. That way we get both without compromises.

PR #7238, opened 2023-09-25
Add Bullet test demonstrating btHeightfieldTerrainShape

JamesTKhan:
I was contacted on YouTube for any examples of `btHeightfieldTerrainShape`. I noticed we do not have a test that covers it so I added a new Bullet test to 1. Provide example as a resource and 2. Better test coverage.

Tested Windows Desktop and Android (Note 8)

I have some strange text/font artifacts right now on Desktop on the bullet tests, but this problem is on the other tests too and not related to this PR.

https://github.com/libgdx/libgdx/assets/28971753/c4b6089c-4707-4636-8970-f3f59b11148c


crykn (approved):
Looks good to me!

PR #7242, opened 2023-09-30
Remove unnecessary Input hack

obigu:
Missed removing this hack from `onDreamingStopped()` when it was removed from `onPause()` here

https://github.com/libgdx/libgdx/pull/6871

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -46,7 +46,6 @@ import com.badlogic.gdx.backends.android.surfaceview.GLSurfaceView20;
 import com.badlogic.gdx.utils.Pool;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 /** An implementation of the {@link Input} interface for Android.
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
```diff
@@ -992,10 +991,6 @@ public void onDreamingStarted () {
 	@Override
 	public void onDreamingStopped () {
 		unregisterSensorListeners();
-		// erase pointer ids. this sucks donkeyballs...
-		Arrays.fill(realId, -1);
```
tommyettinger, line 996: Do we know why only `realId` and `touched` were filled with -1 and false, respectively? No other arrays were, which seems a little odd.

obigu: Not exactly, some of these hacks occured quite a long time ago (https://github.com/obigu/libgdx/commit/82681a44155e1db36c51714a7ce3e4f1f004385c) but my guess is that issues appeared due to no proper handling of touch cancel events and this kind of fixed it.

tommyettinger: It looks like the `realId` erasure is only for the Motorola Jordan? Yeah this should be fine to get rid of, I doubt any phones available in 2011 are still buying apps from the Play Store today.

tommyettinger (approved):
Since the original hacks seemed to be only in place for bugs in hardware from 12 or more years ago, we should be fine to remove them now that it's almost 2024.

PR #7243, opened 2023-10-01
feat: Change audio device on gwt

SimonIT:
One thing to notice is, that the `deviceId` is not very readable. But because of the interface we can just return a String. Also, the permission requested by `getUserMedia` is required. Else an empty array gets returned

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtAudio.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtAudio.java
@@ -24,12 +24,13 @@ import com.badlogic.gdx.backends.gwt.webaudio.WebAudioAPIManager;
 import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.GdxRuntimeException;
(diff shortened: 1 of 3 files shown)


crykn:
Hacktoberfest is long over, but better late than never: This works as expected on Windows! 

A couple things:
* > One thing to notice is, that the deviceId is not very readable.
   
   It might make sense to use `device.label` instead of `device.deviceId`?
* `navigator.mediaDevices` seems to include a `default` audio device from the get go, so we might make this line conditional on not being used with the web backend: https://github.com/libgdx/libgdx/blob/0a785d1f82426038a7e395269ca9e22152fb8269/tests/gdx-tests/src/com/badlogic/gdx/tests/AudioChangeDeviceTest.java#L29
* Also, these docs here need to be updated:
https://github.com/libgdx/libgdx/blob/0a785d1f82426038a7e395269ca9e22152fb8269/gdx/src/com/badlogic/gdx/Audio.java#L96-L98

SimonIT:
> It might make sense to use device.label instead of device.deviceId?

We can try it, but I don't know how unique the label is, so we might have multiple devices with the same label.

> navigator.mediaDevices seems to include a default audio device from the get go, so we might make this line conditional on not being used with the web backend:

I think we should filter the `default` device on gwt so we have it consistent across backends.

PR #7246, opened 2023-10-05
Upgrade libGDX to Gradle 8.4

obigu:
Upgraded libGDX project itself to Gradle ~~8.3~~ 8.4. It requires building with Java 17 and upgrading GWT plugin and AGP plugin.

Tested running the tests on Lwjgl3, Android and GWT.


tommyettinger:
Version 8.4 came out less than 24 hours ago -- should we update to that instead? It has partial support for Java 21 (read: very little support), and has some security fixes.

tommyettinger:
https://docs.gradle.org/8.4/release-notes.html

Important to note are 30% faster builds on Windows, the two security holes fixed, and maybe some other things.

tommyettinger:
0.8.4 seems to be more strict about some things that were warnings in earlier 0.8.x releases. I haven't used Gradle 0.8.4 with Spotless anywhere yet, and maybe Spotless can be updated (though I'm not sure that is what the issue is). There are some warnings about the build in the logs that aren't full-blown errors, but it looks like they're related. Probably fixing the errors will make the warnings go away, or if not, using the same process that fixed the errors might.

obigu:
May it be related to the `fetchNatives` tasks still using the old syntax to register? they were missed on https://github.com/libgdx/libgdx/pull/7134 apparently. That may give Gradle more control over when they get executed and fix it. I'm not sure how to test this stuff locally as executing `spotlessApply` alone doesn't fail. I'll try pushing the changes using the new syntax see what happens unless somebody has a better idea.

obigu:
The Spotless build issue `Reason: Task ':tests:gdx-tests-gwt:spotlessJava' uses this output of task ':tests:gdx-tests-gwt:warTemplate' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.` seems to disappear if we exclude output folders from being scanned by Spotless. Merging #7271 should hopefully fix it.

obigu:
The recent issues related to the non declared dependencies between `fetchNatives` and `Spotless` tasks such as

`Task ':spotlessJava' uses this output of task ':fetchExternalNatives' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.`

 were just the tip of the issue when trying to run both tasks in a single command (`./gradlew fetchNatives build`). I've managed to make the Spotless ones disappear by adding 
```
	tasks.named("spotlessGroovyGradle") {
		mustRunAfter(":fetchNatives")
	}

	tasks.named("spotlessJava") {
		mustRunAfter(":fetchNatives")
	}
```

but one fixed a new one appeared deeper down the line. 

Not to get in a rabbit hole with lots of dependencies to `fetchNatives` I've opted for the simple solution of executing two separate commands: `gradlew fetchNatives` and `gradlew build`. 

There may be a more elegant solution to ensure `fetchNatives` is always run first so that nothing breakes when executing it with other tasks and will happily apply it if somebody comes up with it but for the moment I think it's not unreasonable to agree `fetchNatives` task needs to be run on its own.

Tom-Ski:
I think this is an ok solution for now

Tom-Ski (approved):

PR #7247, opened 2023-10-05
Upgrade Gdx Setup to generate projects using Gradle 8.4

obigu:
Not to merge before #7246.

The only required change has been to remove the following line from the `html` module `build.gradle` as the property doesn't appear to exist on the latest GWT plugin version. 

`checkGwt.war = file("war")`

Would be great if somebody could confirm how important this line is.

**EDITED** 
Another change has been made to disable R8 full mode which is set to `true` by default on AGP 8 (was `false` https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes). Main issue with full mode is it removes implicit empty constructors which are quite commonly used by libGDX `ReflectionPool` (espcially Scene 2d classes). Should be possible to enable Full mode by adding some Proguard rules to keep the necessary empty constructors but it's not easy to make sure some will be missing and unexpected runtime errors occur. For the moment I think it's best to keep using old well tested Proguard rules by default.


tommyettinger:
It isn't important; it was needed to do some verification task in an older Gradle/GWT-plugin version, but isn't needed now. Liftoff hasn't included that line for a while; [it looks like I removed it 17 months ago](https://github.com/tommyettinger/gdx-liftoff/commit/6cb3abd02645cd8570a6fbbb495ed94b43cd7321).

SimonIT:
By the way, it seems that the [gwt-gradle-plugin](https://plugins.gradle.org/plugin/org.wisepersist.gwt) has changed its package to  `org.docstr`: [New PLugin Page](https://plugins.gradle.org/plugin/org.docstr.gwt) ([commit](https://github.com/jiakuan/gwt-gradle-plugin/commit/90e9f7b3eb381b367d2e30f598caa663ec1c04a3)). So there's already version [1.1.29](https://github.com/jiakuan/gwt-gradle-plugin/releases/tag/1.1.29)

tommyettinger:
I definitely agree about R8 full mode; it will be (or was just) set to disabled by default in new Liftoff projects because of how much trouble it causes during minimization. I still think we should update to Gradle 8.4, considering how infrequently gdx-setup updates.

CHANGES
```diff
@@ -2,6 +2,8 @@
 - [BREAKING CHANGE] GWT: Updated to 2.10.0. `com.google.jsinterop:jsinterop-annotations:2.0.2:sources` must be added as a dependency to your html project dependencies.
 - Fixed GlyphLayout for fixed width glyph offsets at the start and end of lines.
 - Fixed scene2d.ui layout for fractional positions and sizes.
+- libGDX is now built using Java 17 due to Gradle 8 requirements.
```
tommyettinger, line 5: I don't think I see this change in the list? Did a commit get lost?

obigu: heheh, I've added here but it should've been added to https://github.com/libgdx/libgdx/pull/7246

crykn (approved):

PR #7250, opened 2023-10-06
chore: Update to gwt 2.10.0

SimonIT:
I don't know why there's no dtd for 2.10.0, so I chose the dtd from 2.9.0


obigu:
Seems the 2.10.0 dtd may be deployed soon? https://github.com/gwtproject/gwt/issues/9770

tommyettinger (approved):
Looks good to me. We should definitely test this as a snapshot before releasing, since I have encountered lots of subtle quirks of GWT before (I'm sure many of us have).

tommyettinger (approved):
Still looks solid; the DTDs use HTTPS and point to 2.10.0, so that should be fine (it's what I've started changing my projects to use, and none of them have had issues). Approved!

crykn (approved):

tommyettinger:
It looks like there's no entry in the changelog. I think I'll merge this first, because it has two approvals, and encourage GWT users to test the latest SNAPSHOT release. If there are any critical issues, then fixing them may affect the changelog; if there aren't any issues, it should be easy to add it in (potentially including quick tips for updating, if we discover some are needed).

obigu:
Current Gdx Setup fails when executing `superDev` with several errors like this:
`[ERROR] Line 6669: No source code is available for type jsinterop.annotations.JsMethod; did you forget to inherit a required module?`
Maybe something missing in the Gdx Setup template app GWT config?

SimonIT:
Ahhh, probably because
```
"com.google.jsinterop:jsinterop-annotations:${versions.jsinteropAnnotations}:sources"
```
is missing

obigu:
> transitively

Not sure about that, I think we don't do it with any other sources. I'd just add the dep alongside the other sources deps.

Issue #7245, opened 2023-10-03
Error selecting angle renderer on desktop


jeffcragg:
Please ensure you have given all the following requested information in your report.

#### Issue details
Can't use angle renderer backend on Windows 11 with integrated plus dedicated GPU.

#### Reproduction steps/code
use this for Desktop config
config.setOpenGLEmulation(Lwjgl3ApplicationConfiguration.GLEmulation.ANGLE_GLES20, 0, 0);

#### Version of libGDX and/or relevant dependencies
Libgdx 1.12.1

#### Stacktrace
Caused by: java.lang.IllegalStateException: setFunctionMissingAddresses has been called already

#### Please select the affected platforms
Windows 11

PR #7251, opened 2023-10-06
Fix ANGLE GLES renderer on dekstop

Berstanio:
LWJGL 3.3.3 now disallows to call `ThreadLocalUtil#setFunctionMissingAddresses` multiple times.
This PR, not very elegant, fixes this issue by resetting the `FunctionMissingAddresses` after the `GLES` class set them in their static initializer. This way, when the `GL` class is initialized, it doesn't crash. 
This crash only happens with `Lwjgl3ApplicationConfiguration.GLEmulation.ANGLE_GLES20`
Related LWJGL3 issue: https://github.com/LWJGL/lwjgl3/issues/931

Fixes https://github.com/libgdx/libgdx/issues/7245

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -56,6 +56,7 @@ import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.ObjectMap;
 import com.badlogic.gdx.utils.SharedLibraryLoader;
 import org.lwjgl.system.Configuration;
+import org.lwjgl.system.ThreadLocalUtil;
 
 public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	private final Lwjgl3ApplicationConfiguration config;
@@ -562,6 +563,8 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 			try {
 				Class gles = Class.forName("org.lwjgl.opengles.GLES");
 				gles.getMethod("createCapabilities").invoke(gles);
+				// TODO: 06.10.23 Remove once https://github.com/LWJGL/lwjgl3/issues/931 is fixed
+				ThreadLocalUtil.setFunctionMissingAddresses(0);
 			} catch (Throwable e) {
 				throw new GdxRuntimeException("Couldn't initialize GLES", e);
 			}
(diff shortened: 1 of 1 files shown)


tommyettinger:
I can confirm that without this PR, the LWJGL3 tests currently fail when run with the `--angle` option.

```
$ java -jar gdx-tests-lwjgl3-1.12.1-SNAPSHOT.jar --angle
Exception in thread "main" java.lang.ExceptionInInitializerError                                      
        at org.lwjgl.opengl.GL11.<clinit>(GL11.java:38)                                               
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:467)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:448)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:162)      
        at com.badlogic.gdx.tests.lwjgl3.Lwjgl3TestStarter.main(Lwjgl3TestStarter.java:91)            
Caused by: java.lang.IllegalStateException: setFunctionMissingAddresses has been called already       
        at org.lwjgl.system.ThreadLocalUtil.setFunctionMissingAddresses(ThreadLocalUtil.java:202)     
        at org.lwjgl.opengl.GL.create(GL.java:182)                                                    
        at org.lwjgl.opengl.GL.create(GL.java:126)                                                    
        at org.lwjgl.opengl.GL.create(GL.java:112)                                                    
        at org.lwjgl.opengl.GL.<clinit>(GL.java:85)                                                   
        ... 5 more
```

I'll test with this PR in a minute, and edit this comment with what I find. I think this test failure is substantial, because no LWJGL3 tests pass with this option.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
```diff
@@ -562,6 +563,8 @@ static long createGlfwWindow (Lwjgl3ApplicationConfiguration config, long shared
 			try {
 				Class gles = Class.forName("org.lwjgl.opengles.GLES");
 				gles.getMethod("createCapabilities").invoke(gles);
+				// TODO: 06.10.23 Remove once https://github.com/LWJGL/lwjgl3/issues/931 is fixed
```
tommyettinger, line 566: This is very minor, but 06.10.23 is ambiguous and depends on culture-specific date formatting rules. I'm pretty sure you meant October 6, 2023, but my brain sees 06.10 and thinks June 10th. You could also just remove the date entirely, since the Git history should hold onto when this was entered.

tommyettinger (approved):
This is just a few lines changed, but it makes a big difference for the LWJGL3 tests with `--angle` enabled. Without this PR, no tests run or pass, and with this PR, it looks like all tests I tried work and pass. Big approval.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
Frosty-J: Agree. 2023-10-06 (ISO 8601) or 6 Oct 2023 would be clearer.

Berstanio: IntelliJ just autogenerated the date, I don't think there is any point in keeping it. I removed it

tommyettinger (approved):
Fully approved!

PR #7255, opened 2023-10-15
Prevent loss of precision in OrthographicCamera.update

armingensler:
Currently `position` and `direction` are first added and then in  `Matrix4.setToLookAt` subtracted again. This can lead to loss of floating-point precision if the the components of `position` is much larger than those of `direction`.


NathanSweet:
This looks like a good change, cheers! I modified it to use Matrix4#translate rather than a temp Matrix4.

PR #7258, opened 2023-10-19
Fix overlaps in the Rectangle Fixed #7257

### LINKED ISSUE #7257: Overlaps of Rectangle could not work properly when the width or height is negative
#### Issue details
the overlap works well when the width and height of the rectangle is positive,
but is doesn't work when we got the negative width and height.

#### Reproduction steps/code
like this code: 
```java
        Rectangle r1 = new Rectangle();
        r1.set(5,5,-10,-10);

        Rectangle r2 = new Rectangle();
        r2.set(0,0,10,10);

        System.out.println(r1.overlaps(r2));
        System.out.println(r2.overlaps(r1));

    }
in this way ,we can see the console print: false.
it seems like this method need to be fix: 
	public boolean overlaps (Rectangle r) {
		return x < r.x + r.width && x + width > r.x && y < r.y + r.height && y + height > r.y;
	}


```

1.10.0

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

voidvvv:
Fix overlaps in the Rectangle
check both Rectangle's side to determine which on is the left and witch one is the right, then this method could be used to negative width and height!
Fixed #7257

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Rectangle.java
+++ b/gdx/src/com/badlogic/gdx/math/Rectangle.java
@@ -217,16 +217,17 @@ public class Rectangle implements Serializable, Shape2D {
 
 	/** @param r the other {@link Rectangle}
 	 * @return whether this rectangle overlaps the other rectangle. */
-	public boolean overlaps (Rectangle r) {
+	public boolean overlaps(Rectangle r) {
 		float thisLeft = width >= 0 ? x : x + width;
-		float thisRight = width >= 0 ? x+width : x;
+		float thisRight = width >= 0 ? x + width : x;
 		float thisButton = height >= 0 ? y : y + height;
 		float thisTop = height >= 0 ? y + height : y;
 
 		float thatLeft = r.width >= 0 ? r.x : r.x + r.width;
-		float thatRight = r.width >= 0 ? r.x+r.width : r.x;
+		float thatRight = r.width >= 0 ? r.x + r.width : r.x;
 		float thatButton = r.height >= 0 ? r.y : r.y + r.height;
-		float thatTop = r.height >= 0 ?r. y + r.height : r.y;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/math/Rectangle.java
```diff
@@ -217,8 +217,18 @@ public boolean contains (Rectangle rectangle) {
 
 	/** @param r the other {@link Rectangle}
 	 * @return whether this rectangle overlaps the other rectangle. */
-	public boolean overlaps (Rectangle r) {
-		return x < r.x + r.width && x + width > r.x && y < r.y + r.height && y + height > r.y;
+	public boolean overlaps(Rectangle r) {
+		float thisLeft = width >= 0 ? x : x + width;
+		float thisRight = width >= 0 ? x + width : x;
+		float thisButton = height >= 0 ? y : y + height;
```
tommyettinger, line 223: I'm guessing you meant `thisBottom` instead of `thisButton`. This also applies to `thatButton` which should be `thatBottom`.

tommyettinger (commented):
What's the use case here? Is it possible to write it down on a sheet of paper with negative height? And put that in an envelope with negative width? I just don't see a point for supporting negative dimensions.

obigu:
> What's the use case here? Is it possible to write it down on a sheet of paper with negative height? And put that in an envelope with negative width? I just don't see a point for supporting negative dimensions.

For me it should behave as an empty `Rectangle` but it's `true` we don't have a contract regarding how it behaves with negative `width` and `height`. The `area()` method for instance would return a positive area if both width and height were negative. Same would apply to other geometry classes such `Circle` with negative radio.

I think this needs would need be tackled in a generic PR that would first require agreeing on how shapes with negative dimensions should behave.

PR #7259, opened 2023-10-20
Fixes #7256

### LINKED ISSUE #7256: Triple clicks on the screen disrupt the very state of pressing the screen
#### Issue details
_The problem arose when switching from the previous version of the library 1.11.0 to 1.12.0_

#### Reproduction steps/code
All you have to do is quickly press three fingers on the screen at the same time, and then libjx will record the extra press. More precisely, the fact that the finger was not removed.
And then if you move with one single finger, it will be considered that “pinching” is happening, and instead of moving, make a "zoom" here
image
And all because here the check is incorrectly verified, mistakenly believing that the finger was not released.
image
If you then press with two fingers at the same time, everything will return to normal.
We use libgdx in different modules, and each one has this problem, regardless of the screen.

_Version 1.12.0_

- [ ] Android

obigu:
Some custom system Android gestures such as touching the screen with 3 fingers to capure a screenshot can trigger an ACTION_CANCEL event. All pointers need to be reset in those cases to avoid an inconsistent state.

I haven't been able to test the PR because I don't have a device that has these kind of gestures so, even if the change is low risk, testing is needed to confirm the fix works.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@
 - API Addition: TiledDrawable: Align can be set to manipulate the alignment of the rendering (TiledDrawable#setAlign, TiledDrawable#getAlign)
 - API Addition: TiledDrawable#draw: Also available as a static function (with align) if you don't want to create an extra instance per texture region
 - Android: Removed mouse catching added on 1.12.0 due to unintended effects (see #7187).
+- Android: Fixed touch state inconsistency when touching screen with 3 fingers on some devices (see #7256)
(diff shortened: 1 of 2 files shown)


obigu:
User "@Ma" on Discord has been able to confirm the PR fixes the issue, it should be good to merge now.

tommyettinger (approved):
Looks pretty clean to me. If it works for someone who had experienced the issue before, that's as good as we can ask for, considering how hard this is to replicate. Approved.

PR #7260, opened 2023-10-20
Add Vector4 class and use it where fitting

tommyettinger:
There's a vec4 type in GLSL that I'm sure we're all familiar with, so it's a little strange that libGDX has Vector, Vector2, and Vector3, but as of right now, no Vector4. This tries to change that. Vector4 is rather similar to Vector3, except for a few methods that require a lot of extra effort due to how 4D space is (`setToRandomDirection()` and `isOnLine()`, mainly). Where ShaderProgram already allows setting a uniform with a Vector3 or Color, it now can take a Vector4 (which acts identically to Color, setting four float components).

Most of the work in Vector4.java was done by Antz and largely follows in the footsteps of Vector2 and Vector3. There's some small changes in other Vector-related files -- `idt()` was present in Vector3 and now Vector4, but not Vector2, so I added it there to match the others. It's a very simple method to add. I improved existing docs in a few places, mostly fixing typos and capitalization mistakes. I changed some C-style array declarations in Vector3 to use the more conventional Java-style array declarations; this was easy to do automatically and I feel it makes this (Java) code more legible.

The tests could be better; let me know if you think something should be tested that isn't now.

Thanks Antz, and thanks in advance to anyone reviewing this!


noblemaster:
Looks fine! 👍

noblemaster (approved):

PokeMMO (approved):

noblemaster:
I suggest, just merge it when you are ready! 👍

PR #7264, opened 2023-10-28
Change documentation for IdleFPS

Xyndra:


Frosty-J:
Not keen on the new wording, I'm afraid. Past-tense runnables is weird and "sleep time" doesn't convey it's a frequency like Hz or FPS.

crykn:
I agree with @Frosty-J, this change doesn't really improve the description.

PR #7271, opened 2023-11-07
Tighten Spotless rules to only include java src folders

obigu:
Prevents Spotless from scanning output folders which causes issues with Gradle 8 (see https://github.com/libgdx/libgdx/pull/7246).

The only folders I've found that contain java files not present in a `src` folder are test classes in `gdx/test` and `gdx-backend-robovm-metalangle/generator` class.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ allprojects {
 		lineEndings 'UNIX'
 
 		java {
-			target '**/*.java'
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
Looks good to me. We just need to remember to change this build.gradle section if we ever do include sources in another location. Approved.

PR #7275, opened 2023-11-08
feat: RISCV native build support

SimonIT:
~not tested yet, it needs LWJGL/lwjgl3#890~

Everything's off to merge🎉


build.gradle
```diff
@@ -19,10 +19,11 @@ buildscript {
 	repositories {
 		mavenCentral()
 		gradlePluginPortal()
+		maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
```
Tom-Ski, line 22: Can be removed

Tom-Ski: Ah we still need it for lwjgl3 release with that support. So probably can't be merged until then. Very cool though

SimonIT: lwjgl3 is released🎉

Berstanio (changes requested):
jnigen 2.5.1 is broken with things like android package, we should wait merging this until 2.5.2 is released

Tom-Ski:
@SimonIT we ready to merge this, jni gen 2.5.2 now pinned version for libgdx. Could you update this and we can test and get it in before relase

SimonIT:
@Tom-Ski All conflicts are resolved!

CHANGES
```diff
@@ -50,6 +50,7 @@
 - Android & iOS: Use empty Audio implementations on Android and iOS when audio is disabled
 - Adds status detection for an httpRequest
 - Fix ANGLE GLES renderer on dekstop (#7274)
+- Architecture support: Support for Linux RISC-V has been added. The gdx-xxx-natives.jar files now also contain native libraries for this architecture.
```
Berstanio, line 53: I would rather mention the gradle artifact here. That the `natives-desktop` now contains RISCV binaries.

Berstanio (approved):

Tom-Ski (approved):

badlogic:
@SimonIT @Tom-Ski @Berstanio this is super cool! Sadly, the upgrade to LWJGL 3.3.4 breaks GLFW on macOS ARM64. The libglfw.dylibs shipped by LWJGL 3.3.4 lack some symbols, which will make things explode.

I.e. running `Lwjgl3TestStarter` on macOS ARM64 results in:

```
Caused by: java.lang.NullPointerException: A required function is missing: glfwGetWindowTitle
```

(Sorry for the screenshot, I can't get IDEA to print out the full stack trace in the console, or libGDX is swallowing it).
Screenshot 2024-08-20 at 11 37 12

Can we revert this until LWJGL has fixed their end? I've reported it on their Discord but haven't gotten a response yet.

Berstanio:
Oh yeah, this is bad!
> Can we revert this until LWJGL has fixed their end?

This is all purely additional, so I think we can just revert, yes

badlogic:
The plot thickens. A standalone LWJGL 3.3.4 app works. My guess: our libgdx
shared library loader doesn't unpack the LWJGL 3.3.4 libglfwarm64.dylib,
but instead reuses the 3.3.3 dylib that was already on my system from
previous runs. That'd be not so good.

PR #7276, opened 2023-11-10
Fixed scene2d.ui layout for fractional positions and sizes.

NathanSweet:
When "rounding" for UI layout, floor x/y and ceil width/height.

For example if a label has a fractional width like 123.4, rounding a Container's width gives 123 and the label won't fit.

Table already does this, so I'm pretty sure it's the right thing to do in all other layouts. If others can please check this doesn't negatively affect their apps, that would be great!

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/HorizontalGroup.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/HorizontalGroup.java
@@ -124,8 +124,8 @@ public class HorizontalGroup extends WidgetGroup {
 		}
 		prefHeight += padTop + padBottom;
 		if (round) {
-			prefWidth = Math.round(prefWidth);
-			prefHeight = Math.round(prefHeight);
+			prefWidth = (float)Math.ceil(prefWidth);
+			prefHeight = (float)Math.ceil(prefHeight);
 		}
 	}
 
@@ -193,7 +193,7 @@ public class HorizontalGroup extends WidgetGroup {
 				y += (rowHeight - height) / 2;
 
 			if (round)
-				child.setBounds(Math.round(x), Math.round(y), Math.round(width), Math.round(height));
+				child.setBounds((float)Math.floor(x), (float)Math.floor(y), (float)Math.ceil(width), (float)Math.ceil(height));
 			else
 				child.setBounds(x, y, width, height);
 			x += width + space;
@@ -283,7 +283,7 @@ public class HorizontalGroup extends WidgetGroup {
 				y += (rowHeight - height) / 2;
 
 			if (round)
-				child.setBounds(Math.round(x), Math.round(y), Math.round(width), Math.round(height));
+				child.setBounds((float)Math.floor(x), (float)Math.floor(y), (float)Math.ceil(width), (float)Math.ceil(height));
 			else
 				child.setBounds(x, y, width, height);
 			x += width + space;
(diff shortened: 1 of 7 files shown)


tommyettinger:
Might be a slight challenge to test, but that should be do-able today... I think this is the type of issue that takes a variety of fonts and scenarios where they might or might not need rounding, to actually test completely.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/HorizontalGroup.java
```diff
@@ -124,8 +124,8 @@ private void computeSize () {
 		}
 		prefHeight += padTop + padBottom;
 		if (round) {
-			prefWidth = Math.round(prefWidth);
-			prefHeight = Math.round(prefHeight);
+			prefWidth = (float)Math.ceil(prefWidth);
+			prefHeight = (float)Math.ceil(prefHeight);
```
tommyettinger, line 128: We should be able to use `MathUtils.ceil()` here, and `MathUtils.floor()` for flooring. The range where those start to be inaccurate is only for floats that are both very large and negative, I think less than -16384 . It probably doesn't matter, since UI layout code isn't called thousands of times per frame. Just an option to consider if some of this PR actually is called very often.

tommyettinger (approved):
I'm not entirely certain under what conditions this fixes or even changes layout, but it hasn't broken anything in my limited testing, and so far looks identical. I might not have enabled rounding in the way this needs... It makes sense logically to fit a slightly smaller area in a slightly larger space, so I'm approving. It probably will cause a minor wrapping change in some projects, some of the time, but that's pretty much expected from any change like this.

NathanSweet:
Thanks for taking a look so quickly! Using MathUtils seems OK. Done, and changed Table to use it also.

Here's some repro code:
```
public class LayoutRoundingTest extends GdxTest {
	Skin skin;
	Stage stage;
	SpriteBatch batch;

	public void create () {
		batch = new SpriteBatch();
		stage = new Stage(new ScreenViewport());

		skin.getFont("default-font").getData().setScale(1.23f);

		Label label = new Label("This is some text.", skin);
		label.setEllipsis(true); // So we can easily see when the label doesn't fit.
		System.out.println(label.getPrefWidth()); // 127 normally * 1.23 font scale = 156.20999

		Container container = new Container(label);
		container.setPosition(100, 100);
		container.pack();
		stage.addActor(container);
	}

	public void render () {
		ScreenUtils.clear(0, 0, 0, 1);
		stage.draw();
	}

	public void resize (int width, int height) {
		batch.getProjectionMatrix().setToOrtho2D(0, 0, width, height);
	}
}
```
This test doesn't seem interesting enough to commit, but the key is to get a label that is > `x.0` and < `x.5` so rounding makes it too small. The ellipsis truncation visualizes when the Container sizes the label too small (it's how I came across it initially). Before PR:

After PR:


The problem doesn't occur with table, which already does this rounding. The other built-in layouts (Container and Horiz/VerticalGroup) aren't used as often as Table, so it probably rarely comes up.

The rounding in ProgressBar is difficult to test, but I think it's OK.

NathanSweet:
I changed my mind on MathUtils ceil/floor, makes me nervous to not be sure the values never exceed 16,384 and the code is called relatively infrequently.

tommyettinger:
Yeah, Math.ceil() and Math.floor() got much faster in relatively recent HotSpot JVM versions anyway. The limit is also -16384 on the negative side, and some very large number (over a million IIRC) on the positive side, if that helps. I seem to remember that the MathUtils docs might have had some small mistake on the upper bound, but the negative side seems pretty clear. Accuracy is the goal here, since this is a less-than-1-pixel adjustment, so Math.ceil/floor make sense.

PR #7277, opened 2023-11-11
Add instanced attributes flags to GLSL prefix

yahel-ck:
Adds `#define attrib_instancedFlag` for each instanced attribute, mostly to allow declaring an attribute instead of a uniform when the specific attribute is instanced (like I did with u_worlTransform for example).

I tested it with this demo that implements instancing `u_worldTransform`: https://github.com/yahel-ck/libgdx_instancing_demo
The demo is using gdx-gltf to render, but it uses the same underlying LibGDX classes. 
I tested it on VMs of iOS and Android, and on a physical Android device (Samsung S21).

It wouldn't take much work to implement the demo solely with LibGDX so let me know if you think it's necessary, or if there are any other tests I should make.
Thanks :)

### DIFF
--- a/gdx/res/com/badlogic/gdx/graphics/g3d/shaders/default.vertex.glsl
+++ b/gdx/res/com/badlogic/gdx/graphics/g3d/shaders/default.vertex.glsl
@@ -103,7 +103,11 @@ attribute vec2 a_boneWeight7;
 #endif
 #endif
 
+#ifdef u_worldTrans_instancedFlag
+attribute mat4 u_worldTrans;
+#else
 uniform mat4 u_worldTrans;
+#endif //u_worldTrans_instancedFlag
 
 #if defined(numBones)
 #if numBones > 0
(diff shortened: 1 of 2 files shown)


mgsx-dev (changes requested):
Hi, there are some things that won't work i think : 
* overriding u_worldTrans by instanced data is not a good idea : 
  * some may want to use both instance transform and modelInstance.transform together.
  * some other uniforms are computed based on renderable.worldTransform (eg. normalMatrix attribute).
* some changes are missing for the depth shader (used to render shadows)
* `#define worldTrans_instancedFlag`will be generated 4 times because it requires 4 attributes with the same name. And tbh, i'm not fan of generating these flags automatically (it could clash with custom shader implementations).

I played around with this topic some times ago and end up with a new attribute used together with u_worldTrans, something like this : 
```glsl
#ifdef INSTANCE_TRANSFORM
attribute mat4 i_transform;
#endif
```

yahel-ck:
Don't know why this says I request another review. No review is needed ATM, I'll unmark this as draft after addressing the requested changes in the depth shader

PR #7278, opened 2023-11-13
Fix Intersectot.intersectPolygons() corner case

### LINKED ISSUE #6986: Intersector.intersectPolygons issue with overlap polygon: polygons must contain at least 3 points.
#### Issue details
`Intersector.intersectPolygons` gives an unexpected error `polygons must contain at least 3 points.` when it is trying to generate the overlap polygon.

#### Reproduction steps/code
```java
import com.badlogic.gdx.*;
import com.badlogic.gdx.math.Intersector;
import com.badlogic.gdx.math.Polygon;

public class Barebones extends ApplicationAdapter {
    public void create () {
        Polygon p1 = new Polygon(new float[]{3018.6086f, 304.13184f, 2983.5085f, 263.56183f, 2903.8853f, 253.33517f, 2857.1885f, 283.38184f, 2863.5386f, 328.83517f, 2913.5886f, 357.0285f, 2981.1785f, 353.03516f});
        Polygon p2 = new Polygon(new float[]{2996.0f, 277.0f, 2996.0f, 278.0f, 2997.0f, 278.0f, 2997.0f, 277.0f});

        Polygon intersectionPolygon = new Polygon();
        if (Intersector.intersectPolygons(p1, p2, intersectionPolygon)) {
            System.out.println("Intersected!?");
        }
    }

    public void render () {
        // your code here
    }

        new LwjglApplication(new Barebones());
    }
}
```

1.11.0

obigu:
Fixes #6986

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -197,7 +197,8 @@ public final class Intersector {
 			floatArray2.addAll(floatArray);
 			floatArray.clear();
 		}
-		if (floatArray2.size != 0) {
+		// Check for 3 or more vertices needed due to floating point precision errors
+		if (floatArray2.size >= 6) {
 			if (overlap != null) {
 				if (overlap.getVertices().length == floatArray2.size)
 					System.arraycopy(floatArray2.items, 0, overlap.getVertices(), 0, floatArray2.size);
--- a/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
+++ b/gdx/test/com/badlogic/gdx/math/IntersectorTest.java
@@ -5,9 +5,7 @@ import com.badlogic.gdx.math.Intersector.SplitTriangle;
 
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class IntersectorTest {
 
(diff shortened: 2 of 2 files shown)


NathanSweet:
Sorry. Formatting can get fixed when this is merged!

obigu:
> This looks fine, but I think line 172 and 156 (I think, in the same function as the one changed) maybe should be checked too. If a malformed input got in somehow, then the check for `floatArray.get(floatArray2.size - 2)` could be a runtime exception.

https://github.com/libgdx/libgdx/blob/46fa173398ce38d0c9ee358cd67da782a5a95724/gdx/src/com/badlogic/gdx/math/Intersector.java#L156
https://github.com/libgdx/libgdx/blob/46fa173398ce38d0c9ee358cd67da782a5a95724/gdx/src/com/badlogic/gdx/math/Intersector.java#L172

No sure what you mean by checking them too. You mean checking for null arguments and well formed Polygon with minimum number of vertex (6+) on L156? If checks are added to 156, do we need to check anything else on L172?

tommyettinger:
I just mean you can change the `== 0` checks to `< 6` checks, I think. It probably doesn't matter; garbage in, garbage out.

obigu:
> I just mean you can change the `== 0` checks to `< 6` checks, I think. It probably doesn't matter; garbage in, garbage out.

I think it's not possible to build a `Polygon` with > 0 and < 6 vertices as both the constructors and `setVertices` make that check.

On the other hand, for proper validation we would need to check for clockwise order as well as well formed convex polygon (not sure what happens if for example all vertices are on the same line and if we can consider it an intersection in that case). I'm not sure adding those checks would be a good idea, I'd rather leave that responsibility to the caller of the method and javadocs. 

Independently of this, we could consider an improvement to `Polygon` class to allow setting a list of constraints to a `Polygon` (such as clockwise/counter-clockwise, convexity, etc...) that, when set, would ensure validity of that particular instance.

PR #7280, opened 2023-11-14
fix(gdx-setup): Add jsinterop-annotations as sources dependency

SimonIT:
Not tested, yet

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.12.2-SNAPSHOT]
+- [BREAKING CHANGE] GWT: Updated to 2.10.0. `com.google.jsinterop:jsinterop-annotations:2.0.2:sources` must be added as a dependency to your html project dependencies.
 - Fixed GlyphLayout for fixed width glyph offsets at the start and end of lines.
 - Fixed scene2d.ui layout for fractional positions and sizes.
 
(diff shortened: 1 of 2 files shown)


obigu:
Haven't tested either but I'm not so sure about this solution. Adding a transient dependency to `jsinterop-annotations:sources` by `gdx-backend-gwt:sources` creates a very weird and potentially inconsistent dependency tree. 

My understanding is that the GWT 2.10 library depends internally JSI interop annotations and to work with it users need to explicitely depend on `GWT library` iself and both the `GWT sources` and `JSI interops sources` in addition to both the `gdx-backends-gwt` library and sources.

 GWT 2.8.2 didn't require the `jsinterop-annotations:sources` dependency so, without this PR, users on older GWT 2.8.2 would need to add the `jsinterop-annotations sources` manually which is I understand is a bit inconvenient.

The problem is that it is GWT that has a dependency on JSI interop annotations, not the libGDX GWT backend. If a user changed the GWT dependency locally there could be a mismatch between the jsinterop-annotations library and its sources dependencies without realising.

If we are going down this path, why not add the `GWT sources` dependency as well to be set as a transient dependency of `gdx-backend-gwt:sources`?

I won't hold a strong position on this as I'm not a GWT user but I think we should consider if it's worth the risks and complexity to avoid users having to add a dependency to their projects.

obigu (dismissed):
Approving this simple approach. GWT is currently broken on Gdx Setup, we can discuss setting transient dependencies in the future.

@SimonIT Can you edit CHANGES and add a [BREAKING CHANGE] for users to know what to do when updating versions?

tommyettinger (approved):
It looks good. The Discord is collectively wondering why gdx-liftoff isn't even mentioned on the wiki, but even if it was recommended... people would still be using old versions of gdx-setup or projects made with them, and the breaking change note matters a lot for them. Approved!

PR #7286, opened 2023-11-24
Added antialiasing flag to the ScreenUtils.

GadWissberg:
I wanted to use the ScreenUtils clear() with applying antialiasing and found out it didn't support that. Thought it could be useful. I added in the javadocs that it is required in the class launcher to specify the number of samples (`config.numSamples = 2` - recommended for Android is 2, and 3 for desktop )
Thanks

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
+++ b/gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
@@ -59,19 +59,16 @@ public final class ScreenUtils {
 		clear(r, g, b, a, clearDepth, false);
 	}
 
-	/**
-	 * Clears the color buffers, optionally the depth buffer
-	 * and whether to apply antialiasing (requires to set number of samples in the launcher class).
+	/** Clears the color buffers, optionally the depth buffer and whether to apply antialiasing (requires to set number of samples
+	 * in the launcher class).
 	 *
-	 * @param clearDepth        Clears the depth buffer if true.
-	 * @param applyAntialiasing applies multi-sampling for antialiasing if true.
-	 */
-	public static void clear(float r, float g, float b, float a, boolean clearDepth, boolean applyAntialiasing) {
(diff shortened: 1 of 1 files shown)


Frosty-J:
Can you explain what adding `GL20.GL_COVERAGE_BUFFER_BIT_NV` to the mask does? I can't find out much about it beyond it being a GL extension, according to the constant's comment.

I've had no problem setting MSAA antialiasing (via the launcher config) without that. Regardless of Linux, Windows, Intel, Nvidia, Android, HTML5. An exception is the Android emulator I've just fired up, which isn't doing antialiasing regardless of `ScreenUtils.clear(Color.BLACK)` or your own method, but on a real device it's okay.

GadWissberg:
TBH, I haven't found info either (except for this answer https://stackoverflow.com/a/36663044) and thought it should be added since there are certain devices/os that requires that bit. If you say you had antialiasing on many cases regardless of that bit then I guess this PR can be closed.

crykn:
`GL_COVERAGE_BUFFER_BIT_NV` is added by the [NV_coverage_sample](https://registry.khronos.org/EGL/extensions/NV/EGL_NV_coverage_sample.txt) extension. In the usage example section of its spec they are clearing it every frame, so that seems to be necessary if coverage sampling is used. The same thing is recommended in the [libGDX docs](https://github.com/libgdx/libgdx/blob/5f2dae5f6879556f663774ba9d9ff439b5bae822/gdx/src/com/badlogic/gdx/Graphics.java#L106) and the commit that originally introduced this (a796233). 

At the moment, the only backend where `Gdx.graphics.getBufferFormat().coverageSampling` can return `true` is the Android one. All other backends just return `false`.

Regardless, this looks like a sensible addition. The only question is if we want have this option be configureable (as is now the case with the PR) or just enable it by default, since this seems to be expected in most cases. People that prefer not to clear the coverage buffer can just use `glClear` itself.

crykn (approved):

PR #7287, opened 2023-11-25
Added pauseWhenMinimized and pauseWhenLostFocus to Lwjgl3 backend

obigu:
PR does 2 things.

1. It fixes `LifecycleListener` `pause/resume` events not being fired on `Window` minimize/restore. Without this change, `Timer`s  are not being paused on windows minimization nor any other application custom registered `LifecycleListener` informed. This will make Lwjgl3 consistent with the other backends behaviour.

2. Adds a couple of flags to control behaviour on window minimizing/maximizing as well as losing focus/gaining focus. Default keeps old behaviour which is to pause on minimize and not to do anything on losing focus. It follows the same philosofy as the flags present in the old Lwjgl backend (`LwjglApplicationConfiguration`).

Quick disclaimer. My knowledge of the specifics of Lwjgl3 is limited, I've tested (and modified to also display lifecycle events) using the simple `Lwjgl3DebugStarter`.


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -185,6 +191,18 @@ public void setForegroundFPS (int fps) {
 		this.foregroundFPS = fps;
 	}
 
+	/** Sets whether to render and fire {@link LifecycleListener#pause()}/{@link LifecycleListener#resume()} events on when window
+	 * is minimized/restored. **/
```
crykn, line 195: That's not entirely true. `#render()` is still called even though the application is paused.

crykn (changes requested):

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
Frosty-J: What I do when making a desktop game is `Gdx.graphics.setContinuousRendering(false)` on pause and `true` on resume, as well as pausing any music, stopping or muting any sounds. This might not work for all games, e.g. if they have networking that relies on the game loop running, but it does feel like one of those things that shouldn't be my own responsibility to handle.

obigu: I see. To me not rendering while minimized shuld be the default behaviour even if there was a mechanism to disable it for those special cases you mention.

In any case, shall it say `/** Sets whether to pause the application {ApplicationListener@pause()}...` instead for example?

obigu: Javadoc has been changed

crykn (approved):

PR #7305, opened 2023-12-22
Adds status detection for an httpRequest

supermaximus80:
This commit adds the public boolean isHttpRequestPending (HttpRequest httpRequest) method to Gdx.net user interface. It allows end users to detect if an httpRequest is pending or already finished. It helpfull for creating advanced business logic.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidNet.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidNet.java
@@ -55,6 +55,11 @@ public class AndroidNet implements Net {
 	}
 
 	@Override
+	public boolean isHttpRequestPending (HttpRequest httpRequest) {
+		return netJavaImpl.isHttpRequestPending(httpRequest);
+	}
+
+	@Override
 	public ServerSocket newServerSocket (Protocol protocol, String hostname, int port, ServerSocketHints hints) {
 		return new NetJavaServerSocketImpl(protocol, hostname, port, hints);
 	}
--- a/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessNet.java
+++ b/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessNet.java
@@ -52,6 +52,11 @@ public class HeadlessNet implements Net {
 	}
 
 	@Override
+	public boolean isHttpRequestPending (HttpRequest httpRequest) {
+		return netJavaImpl.isHttpRequestPending(httpRequest);
+	}
+
+	@Override
 	public ServerSocket newServerSocket (Protocol protocol, String hostname, int port, ServerSocketHints hints) {
 		return new NetJavaServerSocketImpl(protocol, hostname, port, hints);
 	}
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglNet.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglNet.java
@@ -48,6 +48,11 @@ public class LwjglNet implements Net {
 	}
 
 	@Override
+	public boolean isHttpRequestPending (HttpRequest httpRequest) {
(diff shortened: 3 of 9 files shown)


tommyettinger (dismissed):
I don't personally do any networking code, but there's a good reason for this to exist. This was discussed on Discord first, and I felt I should mention that even though the new method is one line, it's one line that can't be run from code outside this package because it uses a package-private method. This avoids exposing the package-private code, which I am assuming is package-private for some reason, while still adding functionality that requires that code. Approved.

SimonIT:
Wouldn't it make sense to add this also to the `Net`-interface? In this state, it would only be usable on desktop with access to the https://github.com/libgdx/libgdx/blob/60f074309093046d7fd9dd3c895bba398b415797/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Net.java#L36 which is also only package-private

tommyettinger:
Ah, yeah since I haven't used the Net code, I didn't know how often you had access to an implementation, and I also didn't notice it was in lwjgl3... It should be in Net, though that requires every implementation of Net to either support it or just throw an UnsupportedOperationException.

SimonIT:
image
`NetJavaImpl` is used everywhere except GWT. My concern was that the method is not accessible as the instances of `NetJavaImpl` are package-private attributes

tommyettinger:
Then I suppose it just got much easier to implement this for all platforms!

supermaximus80:
@SimonIT You are right. isHttpRequestPending (HttpRequest httpRequest) method should be added to Net interface and overriden for all the platforms which use NetJavaImpl. Then PR should be updated. Should I do it myself?

tommyettinger:
That seems best, yes! I think it's fine if GWT throws an appropriate exception for this method, if it proves hard to implement it for some reason. Hopefully, it will be similar to the NetJavaImpl version.

supermaximus80:
@tommyettinger it was not hard to implement it for GWT. I updated PR. Could you please make a review.

tommyettinger (approved):
Looks great, now the pending check can be done on any platform without needing to access the internal implementation. I'm approving, but waiting for a maintainer who actually uses the networking code to look this over with a more experienced eye than mine.

supermaximus80:
@tommyettinger Ok.

NathanSweet:
Seems OK to me, though I also don't use libgdx Net.

supermaximus80:
@NathanSweet Why? Which library do you prefer for networking stuff?

NathanSweet:
@supermaximus80 I write mostly desktop software, so I just use the JDK.

crykn (approved):

crykn:
I don't think there are that many people around here that use gdx's Net class, so our three reviews will have to suffice.

Unrelated to this PR, but something I stumbled across while reviewing: the `removeFromConnectionsAndListeners` line here should probably be moved into the finally block below:
https://github.com/libgdx/libgdx/blob/f5404c9f9b66997c9854de9d7a17d0a9ec14e33c/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java#L226-L235

supermaximus80:
> I don't think there are that many people around here that use gdx's Net class, so our three reviews will have to suffice.
> 
> Unrelated to this PR, but something I stumbled across while reviewing: the `removeFromConnectionsAndListeners` line here should probably be moved into the finally block below:
> 
> https://github.com/libgdx/libgdx/blob/f5404c9f9b66997c9854de9d7a17d0a9ec14e33c/gdx/src/com/badlogic/gdx/net/NetJavaImpl.java#L226-L235

I found it very strange. Because modern games communicate with servers a lot. Even simple user registration requires server communication. Do libgdx devs use another library for networking stuff?
Regarding code block above.
You are right. In case of exception in listener.handleHttpResponse(clientResponse), httpRequest will never be removed from it's listeners. So removeFromConnectionsAndListeners(httpRequest); line has to be moved to finally block!

supermaximus80:
@crykn Do we need to create a bug for that? Or you can fix it by yourself?

crykn:
I created a separate PR for it over at #7312.

PR #7306, opened 2023-12-26
Added possibility to reset a particle effect without starting it

supermaximus80:
This commit adds the method `public void reset (boolean resetScaling, boolean start)` to particle effect which allows to reset it without starting. In my opinion it's very usefull method for constructing advanced business logic. Previously to reset a particle effect without starting it users had to dispose it, then to reload it again from disk.


obigu (commented):
I think it'd be preferrable to avoid repetition by making the other overloaded methods call the new ones.

supermaximus80:
@obigu I updated PR. Could you please take a look.

obigu:
> @obigu I updated PR. Could you please take a look.

Super picky but can also change `reset()` to call `reset(boolean, boolean) such as:
```
public void reset () {
	reset(true, true);
}
```

supermaximus80:
@obigu Done.

obigu (approved):
looks good to me, thanks!

supermaximus80:
@obigu Thank you.

PR #7310, opened 2023-12-27
Bump min Android API level to 19

### LINKED ISSUE #7307: SNAPSHOT build no longer works on Android 4.1
The next version of libGDX will be built with AGP 8 (#7246). A side effect of this is we will go from NDK 23 to 25, which drops support for Android Jelly Bean (API levels 16, 17, 18).

In my testing with the emulator, libGDX 1.12.2-SNAPSHOT crashes on API 16. Interestingly, it actually seems fine on API 18, but I haven't extensively tested (just because badlogic.jpg loads doesn't mean other code won't cause a crash).

As gdx-setup still supports API 14, we need to decide how to proceed. Possible options I see are:

1. Set `minSdkVersion` to 19 in gdx-setup. This is the easiest solution. Google pretty much killed Jelly Bean in 2021 and I'm not sure it can even access the Play Store anymore. It's certainly no longer listed in the version distribution. I don't think this is an unreasonable increase, though it does set a precedent for further increases in future (the latest NDK also drops support for KitKat, API 19 & 20...).
2. Specify `ndkVersion` so we can continue using an older NDK with newer AGP versions. 17.2.4988734 is the last version to support API 14, which would actually be a downgrade from what recent libGDX releases would've used. 23.2.8568313 would let us keep Jelly Bean support. I don't really see an immediate issue with this, but it's probably not a long-term solution.

References:

* https://developer.android.com/ndk/downloads/revision_history
* https://developer.android.com/studio/projects/install-ndk#default-ndk-per-agp

obigu:
Fixes #7307


Frosty-J (approved):
`@TargetAPI` is redundant if we're targeting a minimum of 19. It's doing no harm, of course - just an opportunity to remove more lines if you like seeing red.

Other instances: https://github.com/search?q=repo%3Alibgdx%2Flibgdx%20targetapi&type=code

Frosty-J:
de7e767 - Oh yeah, it should be marked as a breaking change, I figure, as some existing projects won't work unchanged.

Berstanio (approved):
I agree with the bump, api 19 should be more than reasonable.

MobiDevelop:
This feels reasonable at this point. AndroidX moved their minSdkVersion to SDK 19 in October because "According to Play Store check-in data, nearly all Android users have devices on API 19 or newer, so it’s no longer necessary to support legacy versions."

https://android-developers.googleblog.com/2023/10/androidx-minsdkversion-19.html

MobiDevelop (approved):

PR #7315, opened 2024-01-08
Update GdxSetupUI.java with a new error message and changed build gradle

LoopingMobiusStrip:
Changed error message

### DIFF
--- a/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
+++ b/extensions/gdx-setup/src/com/badlogic/gdx/setup/GdxSetupUI.java
@@ -86,7 +86,7 @@ import com.badlogic.gdx.setup.DependencyBank.ProjectDependency;
 import com.badlogic.gdx.setup.DependencyBank.ProjectType;
 import com.badlogic.gdx.setup.Executor.CharCallback;
 
-@SuppressWarnings("serial")
+//@SuppressWarnings("serial")
 public class GdxSetupUI extends JFrame {
 
 	// DependencyBank dependencyBank;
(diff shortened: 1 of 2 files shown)


crykn:
Looking at the regex in line 142 of `GdxSetupUI` (`[a-z][a-z0-9_]*(\\.[a-z0-9_]+)+[0-9a-z_]`), the requirements you listed aren't entirely correct:
- There need to be at least two segments
- Each segment may only contain the _lowercase_ letters `a` to `z`, numbers or an underscore
- The _first segment_ mustn't start with a number or an underscore

crykn (changes requested):
Could you remove the new line added to gdx-tools build.gradle file? And, personally, I'd remove the line breaks in the error message as well.

Frosty-J:
> I'd remove the line breaks in the error message as well.

Wouldn't it look ridiculous without line breaks in the error message? I'm okay with them, though I would remove the space from the start and end of each line.

screenshot

As I expressed on Discord, I would like rid of the space preceding lines 2 to 4 and following 1 to 3. So instead of...

```java
JOptionPane.showMessageDialog(this,
    "Invalid package name. Please check the following: \n Make sure there are at least 2 separate segments in the package name (1 or more dots). \n Ensure that the first segment does NOT start with a number or underscore. \n Ensure all characters are lowercase a-z, numbers, or an underscore.");
```

screenshot

...it could be:

```java
JOptionPane.showMessageDialog(this,
    "Invalid package name. Please check the following:\n" +
    "Make sure there are at least 2 separate segments in the package name (1 or more dots).\n" +
    "Ensure that the first segment does NOT start with a number or underscore.\n" +
    "Ensure all characters are lowercase a-z, numbers, or an underscore."
);
```

screenshot

tommyettinger:
I'm so glad no one reviews my contributions to Liftoff. I have nothing to add except that @Frosty-J 's text layout recommendation is a lot more well-thought-out, and I wouldn't go making it a one-line message. The build.gradle file shouldn't have an extra line added, though; that messes with git blame and history.

crykn (approved):

PR #7319, opened 2024-01-12
Fix formatting, Gradle build (fetchNatives)

tommyettinger:
The only changes here are that I applied formatting with Spotless, made a 2-line adjustment to the task ordering so processResources must run after fetchNatives, and also got LWJGL3 tests to be able to build a JAR again. The build seems to be broken in the current state without the task ordering changes, but it builds with them.


backends/gdx-backend-robovm-metalangle/src/com/badlogic/gdx/backends/iosrobovm/bindings/metalangle/MGLKViewControllerDelegate.java
```diff
@@ -27,7 +27,7 @@
 /*</javadoc>*/
 /*<annotations>*//*</annotations>*/
 /*<visibility>*/public/* </visibility> */ interface /* <name> */ MGLKViewControllerDelegate/* </name> */
-	/* <implements> */extends NSObjectProtocol/* </implements> */ {
+	/* <implements> */ extends NSObjectProtocol/* </implements> */ {
```
Berstanio, line 30: Are all these other formatting changes correct?

extensions/gdx-tools/build.gradle
```diff
@@ -112,6 +112,6 @@ tasks.register('buildRunnables') {
 	}
 }
 
-tasks.named('processResources').configure {
-    dependsOn ':fetchNatives'
+processResources.configure {
```
Berstanio, line 115: Does this really solve the underlying issue? 
If I remember correctly, there were way more issues with `fetchNatives`, because fetchNatives defines the whole project as a output.

backends/gdx-backend-robovm-metalangle/src/com/badlogic/gdx/backends/iosrobovm/bindings/metalangle/MGLKViewControllerDelegate.java
tommyettinger: I didn't add them, spotlessApply did. And that does look like a formatting violation of some kind.

extensions/gdx-tools/build.gradle
tommyettinger: Well, I read the documentation linked in the error message. The last part mentions ordering as a possible solution, and that is what this does. You can try building it yourself. It's possible something was cached on my side, though I ran `gradlew clean` before any building task.

tommyettinger: Discord said earlier that you have to run those as separate lines.

`./gradlew fetchNatives`

`./gradlew build`

Berstanio: Yes, I know, but then we also don't need the `mustRunAfter`, right?

tommyettinger: OK, the natives are also in tmp/ . I'm deleting that too now.

Berstanio: > However, using what's in the PR, I just hand-deleted natives.zip, ran gradlew clean, and then ran gradlew build (without specifying fetchNatives at all).

That is expected, since `gradlew clean` doesn't remove extracted natives, as you can see e.g. under `gdx/libs`. So this should work with or without this pr.
As a side note, `mustRunAfter` does not create any dependency between `fetchNatives` and the `processResources` task.

You could build the project before this pr and the commit too, you just needed to run `fetchNatives` seperatly, which you still need to do after the commit/pr in most cases.

backends/gdx-backend-robovm-metalangle/src/com/badlogic/gdx/backends/iosrobovm/bindings/metalangle/MGLKViewControllerDelegate.java
obigu: The Spotless formatter doesn't seem to care about these. If I run Spotless in master no changes are made but I've tested applying one of these changes manually and Spotless doesn't detect a format violation either. I'm not sure why Spotless complains in your environment, any chance you ran a different version of Spotless or formatter at some point in this branch?

extensions/gdx-tools/build.gradle
tommyettinger: OK, so we still need to get fetchNatives to run before anything else, and actually run... I'll try using both dependsOn and mustRunAfter, I guess. I can confirm that this can build but not run tests without fetching natives. Also, you can look at the GH Actions results to see that the build was actually failing, probably due to formatting issues.

Berstanio: > OK, so we still need to get fetchNatives to run before anything else, and actually run...

I don't think I agree, `fetchNatives` should not be run automatically, otherwise you wouldn't be able to build natives locally

> you can look at the GH Actions results to see that the build was actually failing, probably due to formatting issues.

Yeah, but as you said, that was only formatting, I was only talking about the added dependsOn/mustRunAfter

tommyettinger: So now I'm much more confused. Why did @badlogic make [this, most-recent commit](https://github.com/libgdx/libgdx/commit/fd535ab38900cd266631a7bbeab83d10c1bc5403) in the first place? What was broken before? We know he didn't run spotlessApply, and that's why it won't build currently.

backends/gdx-backend-robovm-metalangle/src/com/badlogic/gdx/backends/iosrobovm/bindings/metalangle/MGLKViewControllerDelegate.java
tommyettinger: I had tried (briefly) the latest version of Spotless to see if it would resolve various warning messages Gradle spits out at the start of each build. They look like:
```
Configuration 'spotless-1429330901' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless865455342' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless-641085672' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
```
The latest version didn't affect the warnings, and there wasn't any way I could check the dozens of files spotless always claims to have modified (but didn't, it was just line ending differences every time -- except apparently there are a few new things Spotless checks in the latest version).

extensions/gdx-tools/build.gradle
obigu: That fix doesn't work, it should be reverted. As explained here https://github.com/libgdx/libgdx/pull/7246#issuecomment-1807271546 and as @Berstanio points out there are many other places where you would need to add a dependency on `fetchNatives` because it defines the whole project as an output.

tommyettinger: Then should this PR just revert the change in the most recent commit?

obigu: I'd close this PR and let badlogic revert his commit.

PR #7336, opened 2024-02-10
Add Stage Viewport Anchoring

Cmrboy26:
A common issue I see online with LibGDX (and one I've experienced myself) is the problem of aligning stages to the edges of the screen. For example, some games may want a hot bar always "anchored" to the bottom of the screen or a sidebar anchored to the left or right. There are solutions for this problem in game engines like [Unity](https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/UIBasicLayout.html), [Unreal Engine](https://docs.unrealengine.com/4.26/en-US/InteractiveExperiences/UMG/UserGuide/Anchors/), and [Godot](https://docs.godotengine.org/en/stable/tutorials/ui/size_and_anchors.html).

The code is relatively simple, but adding this would allow for much more freedom when creating stages. These changes may not work properly with changing the zoom of the camera.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java
+++ b/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.utils.viewport;
 import com.badlogic.gdx.ApplicationListener;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Screen;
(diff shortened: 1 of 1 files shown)


Tom-Ski:
You can achieve this very easily with just Table. If you have a screen viewport on your stage, and it fills your screen, you can have a full screen table with bottom alignment to have its children 'stick' to the bottom of the screen. I'm not sure Stage really needs to have this, typical usage isn't to have many stages for different areas of your UI

PR #7343, opened 2024-02-18
Do not force-overwrite the LWJGL configuration option GLFW_LIBRARY_NAME

colorizenl:
Right now, GLFW_LIBRARY_NAME *always* gets overwritten. That's an issue for people using this property, as there is no way to get around this. The proposal is to only rewrite the configuration if no custom location was used, i.e. the property still has the default value of `null`.


colorizenl:
@crykn I *am* using it! I'm just circumventing the way it normally loads the `.dylib`, since the way this works no longer passes the Mac App Store review. 

However, you're right that https://github.com/libgdx/libgdx/pull/7361 removes the need for this, so I'll close this PR and will wait for that to be merged.

Issue #7345, opened 2024-02-19
Performance issues on Android when enabling config.useGL30


LobbyDivinus:
#### Issue details
I'm experiencing quite reduced performance on Android when enabling AndroidApplicationConfiguration.useGL30. It seems to be related to the amount of draw calls, regardless of whether I'm drawing meshes with custom shaders or just default sprites. I'd like to use GL30 functionality that's why I'd like to enable it. In the example below the performance difference on my old but highend Samsung device is 10 FPS vs 60 FPS.

#### Reproduction steps/code
In my project I experience this issue with 3D meshes and UI. However, I was able to reproduce it by just drawing 200 sprites (each with a separate draw call) in the template project.
```java
public class AndroidLauncher extends AndroidApplication {
	@Override
	protected void onCreate (Bundle savedInstanceState) {
		super.onCreate(savedInstanceState);
		AndroidApplicationConfiguration config = new AndroidApplicationConfiguration();
		config.useGL30 = true; // Toggle this to test difference
		initialize(new MyGdxGame(), config);
	}
}
```

```java
public class MyGdxGame extends ApplicationAdapter {
	SpriteBatch batch;
	Texture img;

	Stage stage;
	Label label;
	StringBuilder labelBuilder = new StringBuilder();
	
	@Override
	public void create () {
		batch = new SpriteBatch();
		img = new Texture("badlogic.jpg");

		stage = new Stage();
		label = new Label(" ", new Label.LabelStyle(new BitmapFont(), Color.WHITE));
		label.setFillParent(true);
		label.setAlignment(Align.topLeft);
		stage.addActor(label);
		stage.setViewport(new StretchViewport(400f, 300f));
	}

	@Override
	public void render () {
		ScreenUtils.clear(0, 0, 0, 1);
		int xmax = 400;
		int ymax = 300;
		int ctr = 200;

		batch.getProjectionMatrix().setToOrtho2D(0f, 0f, xmax, ymax);
		for (int i = 0; i < ctr; i++) {
			batch.begin();
			batch.draw(img, xmax * (0.5f + ((float) i / (ctr - 1) - 0.5f) * 0.8f) - 20f, 0.5f * ymax, 40f, 40);
			batch.end();
		}

		stage.getViewport().update(Gdx.graphics.getBackBufferWidth(), Gdx.graphics.getBackBufferHeight(), true);
		labelBuilder.clear();
		labelBuilder.append(Gdx.graphics.getFramesPerSecond());
[...]

PR #7346, opened 2024-02-21
Optimization for SpriteBatch when running non VertexArray VertexDataModes. (GL30 default)

Tom-Ski:
In flush of SpriteBatch, the index buffer gets a limit for the current size of the batch. It gets the indices buffer reference whilst marking the index buffer as dirty, so on bind of the mesh, the indices are uploaded again. Since SpriteBatch indices are static and never change, this change pre-uploads all the indices to the IndexBuffer at creation time, and skips any further updates at render.

We don't have a nice way to do this without making a bunch of new methods specifically for this case, so I'm just binding and unbinding the mesh to trigger the upload once before a render occurs.

Testers welcome! Seems to be no regressions for me in the test suite, but would be good if otherr can take a look on their systems and devices

Spritebatch defaults timings from SpriteBatchPerformanceTest
```
before change
gl20 0.225ms
gl30 0.280ms


after change
gl20 0.225ms
gl30 0.201ms 

```

Related issue
https://github.com/libgdx/libgdx/issues/7345


tests/gdx-tests/src/com/badlogic/gdx/tests/SpriteBatchPerformanceTest.java
```diff
@@ -0,0 +1,88 @@
...
+			counter.addValue(afterFlush - beforeFlush);
+
+		}
+
+		spriteBatch.end();
+
+		spriteBatch.begin();
+		stringBuilder.setLength(0);
+		stringBuilder.append("Mean Time ms: ");
+		stringBuilder.append(counter.getMean() / 1e6);
```
obigu, line 77: I'd add a conditional that shows "Please Wait..." on the text dependign if `counter.hasEnoughData()` returns `true` or not as it may take a while (over 5s on a Samsung S7) to populate depending on the device you run the test and I thought the test was not working.

tests/gdx-tests/src/com/badlogic/gdx/tests/SpriteBatchPerformanceTest.java
```diff
@@ -0,0 +1,88 @@
...
+		Gdx.gl20.glClear(GL20.GL_COLOR_BUFFER_BIT);
+
+		spriteBatch.begin();
+
+		// Accelerate the draws
+		for (int j = 0; j < 100; j++) {
+
+			// fill the batch
+			for (int i = 0; i < 8190; i++) {
+				spriteBatch.draw(texture, 0, 0, 1, 1);
```
obigu, line 59: Minor thing but maybe instead of drawing a single point we can change it to something like 20px * 20px just for the user to see something is happening.

obigu (dismissed):
I've tested this running the libGDX test suite on Android both GL20 and GL30 and on iOS. I've found no issues or differences between previous behaviour and with this change and for this reason I approve (the suggested changes would be nice to have but not required). I can't review from a functionality/correctness point of view even if they make sense to me.

EDIT: Sharing the results of the performance tests just for reference. On GL20 no changes observed. 
On Android Samsung 7 device **GL20**: 
- Mean Without changes: 1.7s
- Mean After changes: 1.7s

On Android Samsung 7 device **GL30**: 
- Mean Without changes: 1.2s
- Mean After changes: 0.6s

tests/gdx-tests/src/com/badlogic/gdx/tests/SpriteBatchPerformanceTest.java
mgsx-dev: @obigu  i think it would bias the test since it will overload fragment shader stage, considering there are about 819.000 drawings, it would have to render 20x20 pixels each, so ~327 Mega pixels which could be a lot compared to the actual amount (less than 1 Mega pixels). In this end it would defeat isolation of this test.

mgsx-dev: if we want to isolate a bit more, we could disable blending and use depth test (never pass) in order to fully bypass fragment stage.

NathanSweet:
I ran Spine with GL30 with and without these changes, all seems fine. I didn't measure performance differences.

I'm good with merging but I didn't press the button in case obigu's proposed changes will be added.

PR #7350, opened 2024-02-25
Make Screen and ApplicationListener actually Disposable

Anton-Samarskyi:
Both Screen and ApplicationListener interfaces contain dispose() method, but still don't implement Disposable interface. Why not?

### DIFF
--- a/gdx/src/com/badlogic/gdx/ApplicationListener.java
+++ b/gdx/src/com/badlogic/gdx/ApplicationListener.java
@@ -16,6 +16,8 @@
 
 package com.badlogic.gdx;
 
+import com.badlogic.gdx.utils.Disposable;
+
 /**
  * <p>
  * An <code>ApplicationListener</code> is called when the {@link Application} is created, resumed, rendering, paused or destroyed.
@@ -29,7 +31,7 @@ package com.badlogic.gdx;
  * </p>
  * 
  * @author mzechner */
-public interface ApplicationListener {
+public interface ApplicationListener implements Disposable {
 	/** Called when the {@link Application} is first created. */
 	public void create ();
 
--- a/gdx/src/com/badlogic/gdx/Screen.java
+++ b/gdx/src/com/badlogic/gdx/Screen.java
@@ -16,6 +16,8 @@
 
 package com.badlogic.gdx;
 
+import com.badlogic.gdx.utils.Disposable;
+
 /**
  * <p>
  * Represents one of many application screens, such as a main menu, a settings menu, the game screen and so on.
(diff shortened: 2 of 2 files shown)


Frosty-J:
Was wondering that myself until I saw why this PR's build failed. I guess interfaces can't implement interfaces.

gdx/src/com/badlogic/gdx/ApplicationListener.java
```diff
@@ -29,7 +31,7 @@
  * </p>
  * 
  * @author mzechner */
-public interface ApplicationListener {
+public interface ApplicationListener implements Disposable {
```
tommyettinger, line 34: You need to use `extends` for interfaces that "inherit" from another interface, not `implements`. That you didn't bother to build this, or even see if a red underline shows up in your IDE, does not instill confidence in this PR.

gdx/src/com/badlogic/gdx/Screen.java
```diff
@@ -24,7 +26,7 @@
  * Note that {@link #dispose()} is not called automatically.
```
tommyettinger, line 26: This comment isn't actually true in this PR's version.

tommyettinger (changes requested):
Obviously, this needs to be able to compile before it can be considered to be merged. Even then, I'd suspect that with as important a type as `ApplicationListener`, there's a reason why it might not implement `Disposable`, potentially related to the lifecycle and when things get disposed on various platforms. I haven't ever called `dispose()` on an ApplicationListener, and I suspect that it's the type of thing that would only be needed rarely (since usually if the application can be disposed, it has been closed and the OS is taking back all its used memory anyway).

gdx/src/com/badlogic/gdx/Screen.java
Anton-Samarskyi: It wouldn't be called automatically anyway. Even extending of Disposable interface won't make it automatic, the method has to be called from somewhere. This extension only provides the possibility to work with the Screen as with Disposable item.

gdx/src/com/badlogic/gdx/ApplicationListener.java
Anton-Samarskyi: This was a tiny change, I used simple text editor to do that. But forgot, that interfaces use "extends" keyword instead of "implements" for other interfaces.

gdx/src/com/badlogic/gdx/ApplicationListener.java
```diff
@@ -16,6 +16,8 @@
 
 package com.badlogic.gdx;
 
+import com.badlogic.gdx.utils.Disposable;
```
Frosty-J, line 19: This can be removed again.

I'm starting to think the `Disposable` `Screen` change makes sense. Doesn't really matter that it overrides the `Disposable` interface's `dispose()` - `Music` and probably other interfaces do that as well. Nice pull request.

crykn (approved):

tommyettinger (approved):

tommyettinger:
Looks fine now, Screen could probably benefit from being Disposable. Automatic builds pass, there are two approvals; I'll merge.

PR #7353, opened 2024-02-28
GdxSetup: Update generated main class to look more modern

Hyperion2400:
The project that is generated by GdxSetup looks quite dated with its bright red background and the pixelated Bad Logic Games logo.

I updated it to make it look more modern.

Before:
Screenshot 2024-02-28 041938

After:
Screenshot 2024-02-28 051631


Frosty-J:
Could define position on one line as `float imgPosX, imgPosY`, though I'm not saying you should. Since then we get into the weeds of "oh but what about access modifiers" and all that. Speaking of which...

If we're doing this, as much as I love how iconic the eye-bleeding red is, I'm really tempted to `ScreenViewport` it up to stay centred rather than stretch on resize. Even wrapping it as an  `Image` and tossing it in a `Stage` if we're feeling rambunctious, in which case an action can be applied to give it some movement.

If `MyGdxGame` extends `Game`, the code can go in a `ScreenAdapter`, giving the dispose method a purpose (since you conceivably might want to dispose a screen; never found reason to dispose at the application level since resources will be freed by the OS). You can see why I've never made a pull request for that myself - the scope just keeps increasing.

Hyperion2400:
I was thinking about doing something similiar as Phaser which has the logo bouncing around with a particle effect. 

sample1

You can just paste the hello world code into an html to see for youself: https://phaser.io/tutorials/getting-started-phaser3/part5

But that would require a bit more code so I opted for the simpler solution that is still a big improvement imo.

Frosty-J:
`Vector2` is safe to construct before `create()`.

Hyperion2400:
Sounds good. I changed it.

Frosty-J:
Can alternatively use a `Sprite` if storing position separately is too messy. This is not a suggestion, just a thought - I don't know how much you've libGDX'd before.

```java
// create
img = new Sprite(new Texture("logo_dark.png"));
img.setCenterX(Gdx.graphics.getWidth() / 2f);
img.setCenterY(Gdx.graphics.getHeight() / 2f);

// render
img.draw(batch);

// dispose (seems pointless)
img.getTexture().dispose();
```

Underscores in filenames can be a bit sketchy if they're ever run through Texture Packer, so often dashes are used instead. E.g. `bob_version_2` would by default become region `bob_version` with index `2`. I may or may not have been bitten by that the other day. This image is unaffected since `dark` is not a number. Though I realise this is the filename used for the website.

I should leave this place now, as I could go on indefinitely. It's such a difficult thing, knowing what's ideal to have in the main class. Something that says "it is working; welcome to libGDX."

crykn:
gdx-setup was retired in favour of [gdx-liftoff](https://github.com/libgdx/gdx-liftoff), so this PR is obsolete.

PR #7357, opened 2024-03-04
Fix incorrect glyph page index in FreeTypeFontGenerator

raeleus:
When using the BitmapFontWriter, fonts that require more than one texture page to fit all their glyphs fail. The resulting FNT file points to glyphs on the wrong page, resulting in a font that renders with odd looking characters. This occurs frequently with large fonts and with glyphs that could potentially land on the last row of the first page.

After some investigation, I discovered that it was a minor issue in FreeTypeFontGenerator causing this. It assigns the page index of each glyph to the last index of `packer.getPages()`. This is wrong because glyphs may still fit on the previous page depending on its dimensions.

The fix just applies the glyph hash code as the pixmap name to find the correct page. It's an inelegant solution, but does not require any API changes to adopt.

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -598,9 +598,9 @@ public class FreeTypeFontGenerator implements Disposable {
 			}
 		}
 
-		String pixmapName = Integer.toString(glyph.hashCode());
+		String pixmapName = glyph.hashCode() + "_" + glyph.id;
 		Rectangle rect = packer.pack(pixmapName, mainPixmap);
-		glyph.page = packer.getPages().indexOf(packer.getPage(pixmapName), true);
+		glyph.page = packer.getPageIndex(pixmapName);
 		glyph.srcX = (int)rect.x;
 		glyph.srcY = (int)rect.y;
 
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
@@ -598,8 +598,9 @@ public class FreeTypeFontGenerator implements Disposable {
 			}
 		}
 
-		Rectangle rect = packer.pack(mainPixmap);
-		glyph.page = packer.getPages().size - 1; // Glyph is always packed into the last page for now.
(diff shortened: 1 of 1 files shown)


Frosty-J (approved):
Looks good. This will alleviate the amount of untold suffering.

raeleus:
I had no idea it was such a pervasive problem! Thanks, James.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -598,8 +598,9 @@ public FreeTypeBitmapFontData generateData (FreeTypeFontParameter parameter, Fre
 			}
 		}
 
-		Rectangle rect = packer.pack(mainPixmap);
-		glyph.page = packer.getPages().size - 1; // Glyph is always packed into the last page for now.
+		String pixmapName = Integer.toString(glyph.hashCode());
```
tommyettinger, line 601: I think you should probably consider using this instead of a not-necessarily-unique hashCode():

```java
		String pixmapName = Integer.toString(glyph.id);
```

The ID should be unique in any font, barring some weirdness with alternate glyphs. The hashCode() is very likely to be unique, but it is the same as calling System.identityHashCode(), and that's neither as fast as it could be nor as unique as we want it to be.

tommyettinger (changes requested):
tl;dr hashCode() isn't guaranteed to be unique, so some Pixmap names might be duplicates using that. The glyph's id should be unique, if I understand it correctly.

raeleus:
I tested ID and it failed with my fonts. Several glyphs evaluate to 0. I did have the same concern about hash code collisions so I should be able to concatenate ID and hashcode. Hopefully that's enough unless you have any better ideas, @tommyettinger.

tommyettinger:
> I tested ID and it failed with my fonts. Several glyphs evaluate to 0. I did have the same concern about hash code collisions so I should be able to concatenate ID and hashcode. Hopefully that's enough unless you have any better ideas, @tommyettinger.

Sounds good to me! Make sure to separate the two strings with a space or underscore or something, because if the id is 0 that's colliding, then a hash of 12345 with an id of 10 and a hash of 123451 with an id of 0 might still collide (obviously extremely rarely) as 1234510. With an underscore, 12345_10 and 123451_0 wouldn't collide. I don't even know if PixmapPacker has problems with duplicate names; I know an atlas permits them for some reason.

tommyettinger (approved):

raeleus:
Thanks for the guidance, TEt. This one is ready for approval by another team member.

tommyettinger (approved):

PR #7358, opened 2024-03-04
Only allow scrolling of a scroll pane if it has scroll focus

raeleus:
This PR addresses the issue of a user using the scrollwheel inside a ScrollPane nested inside of another ScrollPane causing both to scroll at the same time. If you look at web browsers and other user interfaces, scrolling a nested scroll pane never scrolls the parent at the same time. Advanced layouts with lots of scrollable content will benefit from this.


NathanSweet:
I agree the currently scrolling behavior isn't good, only the inner scrollpane should scroll when it has focus.

Normally only the scroll focus actor gets the scroll event. When scrollpanes are nested, the event bubbles up to the parent and it reacts too. Your fix is fine, or it could be fixed using `event.cancel();` so the event doesn't reach the parent. The difference is minimal but I think I prefer that, so the logic is done with the event system and doesn't need to check external state.

raeleus:
Thank you so much!

Issue #7360, opened 2024-03-08
performance: unnecessary window.makeCurrent() in LWJGL3 backend


McMurat:
Please ensure you have given all the following requested information in your report.

#### Issue details
In the main loop of a Lwjgl3Application there is a loop which iterates over all Lwjgl3Windows and calls window.makeCurrent() which triggers an OpenGL context switch using glfwMakeContextCurrent(). This is unnecessary to call repeatedly when there is only one single window. Optimizing this will cause my application to run around 10% faster.

I suggest a simple change:

```java
                if (currentWindow != window) {
                    window.makeCurrent();
                    currentWindow = window;
                }
```

This is sufficient because currentWindow is null initially. So glfwMakeContextCurrent() will be called at least once.

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows
- [x] Linux
- [x] macOS

crykn:
Good catch! I think `currentWindow` would need to be set [here](https://github.com/libgdx/libgdx/blob/3d8cfb8ce3d86cdf0909ce298c38ecca53d50ea0/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L560) as well, so that the right context is active in the first render pass after a second window is created. Do you want to submit a PR for this?

McMurat:
> Good catch! I think `currentWindow` would need to be set [here](https://github.com/libgdx/libgdx/blob/3d8cfb8ce3d86cdf0909ce298c38ecca53d50ea0/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L560) as well, so that the right context is active in the first render pass after a second window is created. Do you want to submit a PR for this?

Yes, setting `currentWindow` to the created window and also calling `makeCurrent()` on the it [here](https://github.com/libgdx/libgdx/blob/3d8cfb8ce3d86cdf0909ce298c38ecca53d50ea0/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L464) is also necessary for multi-window support. The additional call to `makeCurrent()` is only needed to update Gdx.graphics, Gdx.input, Gdx.gl, ... for the new active window (currently this is not happening ant therefore a bug in the current implementation).

[...]

PR #7362, opened 2024-03-10
fix: avoid unnecessary window.makeCurrent() calls

McMurat:
This fixes the issue as discussed [here](https://github.com/libgdx/libgdx/issues/7360)

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -186,8 +186,10 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 			closedWindows.clear();
 			int targetFramerate = -2;
 			for (Lwjgl3Window window : windows) {
-				window.makeCurrent();
-				currentWindow = window;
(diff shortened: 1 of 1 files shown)


tommyettinger (commented):
I'm just commenting to try to bring some eyes to this; I don't personally feel like I can adequately review the code for correctness. I can review it for performance in a little while. I've had my suspicions about LWJGL3 not running as fast as it could for a while, but it's hard to compare different versions for speed. GL30 is much slower than GL20 on my machine, when using LWJGL3, which is odd, but I'm not sure if that's related to this.

casid:
I tried this patch with my game on windows and on mac.

It worked on both and subjectively the game felt a little smoother than before.

I have vsync activated and I'm profiling how long rendering the scene takes. Before the patch, on windows the rendering took a few us, now it takes roughly 16ms. Seems like this change made vsync work properly. On mac on the other hand no changes after the patch, rendering is still a few us.

tommyettinger:
@casid That seems odd that Mac still takes just a few microseconds -- if a frame took 1 millisecond, that would put you at about 1000 FPS, and if it's even less than 1ms/frame, the framerate would be... certainly it could be much higher than I can get with VSync off. Does it seem like your game is using more CPU or GPU time than expected on Mac?

Tom-Ski (approved):

PR #7365, opened 2024-03-13
Fix playback volume of mono sounds in the lwjgl3 backend

Hangman:
The position of sound sources was set to (0,0,1) in 3D space, resulting in sounds not playing at full volume. This led to varying levels of audio output volume compared to other backends (stereo sounds weren't affected).

This PR sets the position to (0,0,0). In order to not break panning, I had to change the way how 2D panning works.
If pan is set to 0 (center), sound output shouldn't differ from playing an unpanned sound. This was the case with the original panning calculation and a default source position of (0,0,1).

The new panning calculation is based on this [comment](https://github.com/kcat/openal-soft/issues/194#issuecomment-392520073) by kcat, the author of OpenAL Soft. If you read the comment, you'll notice a mention of a second variant in the last block. I've chosen this method because the sound output closely matches the output prior to this PR.

Visual comparison:
`play(volume, pitch, pan)`
The red line indicates the maximum level of output.
pr

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@
 - Android: Add configuration option to render under the cutout if available on the device.
 - Fix: Keep SelectBox popup from extending past right edge of stage.
 - Added Framebuffer multisample support (see GL31FrameBufferMultisampleTest.java for basic usage)
+- Fixed playback volume of mono sounds
 
 [1.12.1]
 - LWJGL3 Improvement: Audio device is automatically switched if it was changed in the operating system.
--- a/CHANGES
+++ b/CHANGES
@@ -9,7 +9,7 @@
 - Android: Add configuration option to render under the cutout if available on the device.
 - Fix: Keep SelectBox popup from extending past right edge of stage.
 - Added Framebuffer multisample support (see GL31FrameBufferMultisampleTest.java for basic usage)
-- Fixed playback volume of mono sounds
+- Fixed playback volume of mono sounds and improved panning quality
 
 [1.12.1]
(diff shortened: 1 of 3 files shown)


Frosty-J:
It's different but sadly still not quite right, in terms of both amplitude and frequency response.

Three-way comparison of a mono sound in this branch, a mono sound in 1.12.1, a stereo sound in this branch - in that order:

https://github.com/libgdx/libgdx/assets/60154347/00f8b8a0-a5d9-4d2c-94cc-e447ecf933ea

(FLAC in MP4 container due to GitHub file restrictions - player is muted by default for some reason)

The files I'm using for testing (mono version is louder): [boom-oggs.zip](https://github.com/libgdx/libgdx/files/14586341/boom-oggs.zip)

Hangman:
I also notice a slight difference. But for me it's minimal, barely audible, barely visible in the waveform. The difference in your mp4 is brutal. I honestly don't know what's going on.

Well, even if it doesn't completely solve the problem for you, this PR is also a prerequisite for panning stereo sources, as stereo channels only remain separated at position (0,0,0) (if not panned). If we want to enable that in the future.

Frosty-J:
Stereo panning would be nice, and I think this makes mono sounds slightly better. Doesn't make them any worse, at least, so I'm not against this. I'm sure panning stereo sources works on GWT; don't see any reason why it shouldn't work for Android either since it just adjusts the volume of the left and right channels independently. iOS I'm not so sure on, but it's sorta the ugly duckling anyway and not a reason to hold desktop back. I'll repeat the experiment if you get stereo panning working, to see if that screws it up as well or if mono sources are a special case.

PR #7366, opened 2024-03-19
Document units of Batch rotation

soundasleep:
In the Javadocs, it's not clear what the units of rotation are for Batch - are they degrees? radians?

After trial and error I found out they're degrees (e.g. 0...360), I think it'd be helpful to have this clearly documented in the API

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Batch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Batch.java
@@ -155,8 +155,7 @@ public interface Batch extends Disposable {
 	 * rectangle is offset by originX, originY relative to the origin. Scale specifies the scaling factor by which the rectangle
 	 * should be scaled around originX, originY. Rotation specifies the angle of counter clockwise rotation of the rectangle around
 	 * originX, originY. 
-	 * @param rotation rotation in degrees
-	 */
+	 * @param rotation rotation in degrees */
 	public void draw (TextureRegion region, float x, float y, float originX, float originY, float width, float height,
 		float scaleX, float scaleY, float rotation);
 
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/Batch.java
```diff
@@ -154,14 +154,16 @@ public void draw (Texture texture, float x, float y, float width, float height,
...
-	 * originX, originY. */
+	 * originX, originY. 
+	 * @param rotation rotation in degrees */
 	public void draw (TextureRegion region, float x, float y, float originX, float originY, float width, float height,
 		float scaleX, float scaleY, float rotation);
 
 	/** Draws a rectangle with the texture coordinates rotated 90 degrees. The bottom left corner at x,y and stretching the region
 	 * to cover the given width and height. The rectangle is offset by originX, originY relative to the origin. Scale specifies the
 	 * scaling factor by which the rectangle should be scaled around originX, originY. Rotation specifies the angle of counter
 	 * clockwise rotation of the rectangle around originX, originY.
```
tommyettinger, line 165: Not a hard requirement, but it would be nice if the descriptive text also mentioned that the angle is measured in degrees. This also applies to the other two changed sets of docs.

tommyettinger (approved):
Looks good to me, though I think it would be nice for the description to mention the degree measurement, as well as the parameter docs (it's not that important, though). Mostly this needs to build before it can merge; I think maybe I can trigger the `spotlessApply` task myself, but I don't know. I'm also just guessing that the build failed due to the Spotless plugin finding formatting issues, so running `spotlessApply` on your side would be easiest, but if it failed for some other reason, who knows.

soundasleep:
Thank you! Here I was thinking I'd save time by editing the code using the GitHub inline editor 😂 I'll get the build to pass eventually.

tommyettinger (approved):
It looks like this passes automatic build checks now, and the docs are just as improved now as when the PR was first submitted, so I'm approving again.

PR #7369, opened 2024-03-23
Added a convenient method to clamp the vector's components.

GadWissberg:
Hello,
I came across several times where I used the Vector3 to specify locations in my game's world and many times wanted to clamp the position to the map's size. So I thought instead of repeating it, it might be useful for other devs.
Thanks

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Vector3.java
+++ b/gdx/src/com/badlogic/gdx/math/Vector3.java
@@ -715,6 +715,20 @@ public class Vector3 implements Serializable, Vector<Vector3> {
 		return epsilonEquals(x, y, z, MathUtils.FLOAT_ROUNDING_ERROR);
 	}
 
+	/**
+	 * Clamps the vector's X, Y and Z with the given boundaries.
+	 *
+	 * @param min The minimum value to clamp the vector's components
+	 * @param max The maximum value to clamp the vector's components
+	 * @return this vector for chaining
+	 */
+	public Vector3 clampComponents(float min, float max) {
+		x = MathUtils.clamp(x, min, max);
+		y = MathUtils.clamp(y, min, max);
+		z = MathUtils.clamp(z, min, max);
+		return this;
+	}
+
 	@Override
 	public Vector3 setZero () {
 		this.x = 0;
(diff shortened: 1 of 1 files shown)


tommyettinger (changes requested):
This is definitely useful, though it would be best if it worked with Vector2 (and maybe Vector4) as well. I had to look at [the sources for Vector3.clamp](https://github.com/libgdx/libgdx/blob/d3ec4309792159ee97df8c1d87fcb8025e3ab683/gdx/src/com/badlogic/gdx/math/Vector3.java#L650-L659) to notice that the existing clamp() does not actually clamp, it scales to stay within a limit rather than truncating numbers that go above or below a max or min. Since the new method is here, maybe a different name is warranted for clampComponents(), but it's pretty descriptive as-is. If you think of a better name, go for it, but I'm fine approving once Vector2 gets this (and ideally Vector4), all of which should be extremely similar to this code.

Frosty-J:
In that case it could be added to the `Vector` interface and implemented in each respective class.

I'm unsure how robust this solution is, as you may not want your map size to be perfectly cubic forever and am generally struggling to come up with uses for clamping each component to the same range.

GadWissberg:
Taken your feedback and applied it by adding a generic unified method to the interface and also added a more flexible version to each Vector class.

tommyettinger:
> In that case it could be added to the `Vector` interface and implemented in each respective class.
> 
> I'm unsure how robust this solution is, as you may not want your map size to be perfectly cubic forever and am generally struggling to come up with uses for clamping each component to the same range.

Mainly, clamping to the -1 to 1 or the 0 to 1 range come to mind, since those can then be scaled out to any non-cubic range you want. I'm approving this now.

tommyettinger (approved):

obigu:
I like the added `clampComponents(float min, float max)` added to the `Vector` interface but I don't like the implementation specific overloaded methods. 

This signature hurts my eyes `public Vector4 clampComponents (float xMin, float xMax, float yMin, float yMax, float zMin, float zMax, float wMin,
		float wMax)`. I'd personally stick to the generic one and remove the others. I'm not dying on this hill though if others think it's worth having them all.

PR #7371, opened 2024-03-30
Adding iOS RoboVM AudioDevice implementation

auraxangelic:
Adding iOS RoboVM AudioDevice implementation. Adding custom bindings to the functions required to make this work.

I did some light testing and it works as far as I can tell. Please feel free to try it out yourself.

Small plug for my text-to-speech synthesizing project that uses the AudioDevice interface: https://github.com/auraxangelic/libgdx-tts You can use it to test this PR.

A lot was taken from the iOS moe implementation from Bernstanio. https://github.com/Berstanio/libgdx/blob/moe/1.12.1/backends/gdx-backend-moe/src/com/badlogic/gdx/backends/iosmoe/IOSAudioDevice.java


obigu:
I've tested using the libGDX test `AudioDeviceTest` which emits a tone and you can change the pan. It works on Simulator but on a device (iPhone 12 Pro iOS 17.4) the sound cracks a lot. I'll let @Berstanio doing a proper review :)

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/ALWrapper.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.robovm.apple.foundation.NSObject;
+import org.robovm.objc.ObjCRuntime;
+import org.robovm.objc.annotation.Method;
+import org.robovm.objc.annotation.NativeClass;
+import org.robovm.rt.bro.annotation.Library;
+import org.robovm.rt.bro.ptr.VoidPtr;
+
+/** @author Jile Gao */
+@Library(Library.INTERNAL)
```
Berstanio, line 12: This should be `@Library("ObjectAL")`, since ObjectAL is now linked dynamically.
But it is currently wrong in all ObjectAL bindings

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudioDevice.java
```diff
@@ -0,0 +1,125 @@
...
+
+	@Override
+	public void writeSamples (short[] samples, int offset, int numSamples) {
+		if (numSamples < 0) throw new IllegalArgumentException("numSamples cannot be < 0.");
+		ShortPtr shortPtr;
+		if (numSamples + tmpBuffer.position() >= minSize) {
+			shortPtr = Struct.allocate(ShortPtr.class, numSamples + tmpBuffer.position());
+
+			shortPtr.set(tmpBuffer.array(), 0, tmpBuffer.position());
+			shortPtr.set(samples, offset + tmpBuffer.position(), numSamples);
```
Berstanio, line 54: This is not correct.
`tmpBuffer.position()` is a destination offset and `offset` is a source offset.
I _think_ you need to do something like:
`shortPtr.next(tmpBuffer.position()).set(samples, offset, numSamples);`

Berstanio (commented):
Overall looks good, thank you for taking the time.
You can find a few further bug fixes that I forgot to upstream here: https://github.com/Berstanio/libgdx/commit/87268815762805b25ed4f79825c02e4adcd7de7a
The increase of the default buffer matches the LWJGL3 default and might address the audio stutter, that @obigu mentioned on the simulator.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/ALWrapper.java
auraxangelic: If it works with `Library.INTERNAL` I would prefer to use a different PR to change this string value.

auraxangelic:
Oof. I applied the changes and I think it's much worse. I'm going back to the drawing board.

Berstanio:
> I applied the changes and I think it's much worse.

What exactly? You had in the latest commit:
```java
		if (alSource.queueBuffer(buffer)) {
			alBuffers.add(buffer);
		}
		alBuffers.add(buffer);
```
, where the uncoditional `alBuffers.add(buffer);` should be removed. Was that maybe the root cause for the "much worse"?

auraxangelic:
> > I applied the changes and I think it's much worse.
> 
> What exactly? You had in the latest commit:
> 
> ```java
> 		if (alSource.queueBuffer(buffer)) {
> 			alBuffers.add(buffer);
> 		}
> 		alBuffers.add(buffer);
> ```
> 
> , where the uncoditional `alBuffers.add(buffer);` should be removed. Was that maybe the root cause for the "much worse"?

Nice catch! I'll try it and open the PR back up once everything works.

auraxangelic:
@Berstanio For more context, I'm still trying to get my text-to-speech library to work on iOS. It works fine on desktop and android. Currently on iOS it speaks the first phrase you give it and it speaks no other. With the latest change, it even plays the last sample written in a loop because no new samples have been written because my tts library is waiting for my libgdx game to tell to to speak something else. Until both of these issues are fixed, I'm going to continue cooking.

auraxangelic:
I think it's fixed. @obigu @Berstanio Please take a look when you get the chance.

auraxangelic:
I also added some typo fixes while I was at it.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudioDevice.java
```diff
@@ -0,0 +1,134 @@
...
+				ALBuffer alBuffer = alBuffers.get(j);
+				if (alSource.unqueueBuffer(alBuffer)) {
+					alBuffersFree.add(alBuffer);
+					alBuffers.remove(alBuffer);
+					j--;
+					toFree--;
+				}
+			}
+			try {
+				Thread.sleep(2);
```
Berstanio, line 78: If we managed to sucessfully unqueue a buffer, there is no need to sleep. The sleep is only needed, if there is no free buffer available and we need to wait until one gets free.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudioDevice.java
```diff
@@ -0,0 +1,134 @@
...
+		while (alBuffersFree.isEmpty()) {
+			int toFree = Math.min(alSource.buffersProcessed(), alBuffers.size());
+			for (int j = 0; j < toFree; j++) {
+				ALBuffer alBuffer = alBuffers.get(j);
+				if (alSource.unqueueBuffer(alBuffer)) {
+					alBuffersFree.add(alBuffer);
+					alBuffers.remove(alBuffer);
+					j--;
+					toFree--;
+				}
```
Berstanio, line 75: A bit nitpicky, so up to you whether you want to implement it, but:

> Removal of a given queue entry is not possible unless either the
source is stopped (in which case then entire queue is considered processed), or if the
queue entry has already been processed.
https://www.openal.org/documentation/openal-1.1-specification.pdf

So, if unqueuing fails, it means that the buffer we tried to unqueue is not processed yet, therefor every subsequent buffer is also not processed yet. Thus, we could bail out here and don't need to try to unqueue further.

auraxangelic:
@Berstanio Thank you. I've applied your 3 commented items and everything still seems to be working the way I expected.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudioDevice.java
```diff
@@ -0,0 +1,134 @@
...
+			shortPtr.next(tmpBuffer.position()).set(samples, offset, numSamples);
+			numSamples += tmpBuffer.position();
+			tmpBuffer.position(0);
+		} else {
+			tmpBuffer.put(samples, offset, numSamples);
+			return;
+		}
+
+		while (alBuffersFree.isEmpty()) {
+			int toFree = Math.min(alSource.buffersProcessed(), alBuffers.size());
```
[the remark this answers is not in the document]
auraxangelic: As I was cleaning up this PR trying to figure out why my audio was glitchy, I removed some things that weren't relevant to my specific use case. I've added this back in and everything still works.

backends/gdx-backend-robovm/src/com/badlogic/gdx/backends/iosrobovm/objectal/OALIOSAudioDevice.java
auraxangelic: Nice. I did not know this.

obigu:
Great, cracks have disappeared, merging!

Issue #6895, opened 2022-06-11
[iOS] OpenGL backend not working on arm64 simulator (silicon)


Berstanio:
The OpenGLES iOS backend is not working on arm64 simulators. The following message gets spammed:
`'cyclone' is not a recognized processor for this target (ignoring processor)`
and after the launch the App crashes (at least in my test app, I tried this one: https://github.com/Berstanio/chigginblazzder)
This is also not libGDX related, so unfortunately it looks like there is no way to fix this: https://developer.apple.com/forums/thread/658330

As a workaround either the MetalANGLE backend can be used or running the simulator with rosetta. 
This should be probably documented somewhere, but I wouldn't know where to do this best.

crykn:
The Importing & Running docs may be a good place to mention this: https://github.com/libgdx/libgdx.github.io/blob/dev/wiki/start/import-and-running.md#ios

PR #7372, opened 2024-03-30
Fix RoboVM backend gdx tests

obigu:
libGDX RoboVM tests `gdx-tests-iosrobovm` were not updated to use dynamic libraries after we migrated in  https://github.com/libgdx/libgdx/pull/7138 and didn't work, now they work.

### DIFF
--- a/tests/gdx-tests-iosrobovm/robovm.xml
+++ b/tests/gdx-tests-iosrobovm/robovm.xml
@@ -34,19 +34,20 @@
     <pattern>org.apache.harmony.security.provider.crypto.CryptoProvider</pattern>
   </forceLinkClasses>
   <libs>
-    <lib variant="device">../../gdx/libs/ios32/gdx.xcframework/ios-arm64_armv7/libgdx.a</lib>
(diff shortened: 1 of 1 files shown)


Berstanio (approved):
Works good!

On a related note, I think we could switch to the metalangle backend for the tests, otherwise the simulator is just broken on M1

obigu:
> Works good!
> 

Thanks for testing.

> On a related note, I think we could switch to the metalangle backend for the tests, otherwise the simulator is just broken on M1

As per our conversation on Discord (PM), regarding the move to metalangle I agree we should start considering seriously. Adding to the problem you refer to (https://github.com/libgdx/libgdx/issues/6895) current OpenGL backend limits apps to run at 60fps and lacks stuff such as VisionOS. If we add the risk of Apple introducing more bugs and not caring about them on openGL as has already occured on future versions the move to metalangle seems like a no brainer. I'll run some more tests on it and create a separate PR to make it default on Gdx tests as well as Gdx Setup.

PR #7374, opened 2024-03-31
Added new Color CLEAR_WHITE

raeleus:
When you are tinting a texture, you may want to lerp from white to completely transparent. Unfortunately, if you use Color.CLEAR, it will also darken the image as it turns transparent. That's because Color.CLEAR is defined as (0, 0, 0, 0). That has its uses, but in this case Color.CLEAR_WHITE is a better option. It's defined as (1, 1, 1, 0). The texture will not be darkened if users implement this new Color. This change will not impact any existing code.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -33,6 +33,7 @@ public class Color {
 	public static final float WHITE_FLOAT_BITS = WHITE.toFloatBits();
 
 	public static final Color CLEAR = new Color(0, 0, 0, 0);
+	public static final Color CLEAR_WHITE = new Color(1, 1, 1, 0);
 
 	public static final Color BLUE = new Color(0, 0, 1, 1);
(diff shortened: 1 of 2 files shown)


tommyettinger (approved):
I like this. It's up to you if you think `Colors.reset()` needs to be updated to include CLEAR_WHITE as well as CLEAR. `Colors` is where color markup in a BitmapFont looks up existing names of colors. Since I don't think color markup generally gets used as a (partial) tint on an existing color, it would be fine to omit it there, I think. Anyone who needs fully invisible text already has `CLEAR` available.

raeleus:
That's a good catch and I think it should be added to reset() for consistency. Thank you so much!

tommyettinger (approved):
Looks quite good to me! None of the colors have JavaDocs, but this particular color won't be immediately obvious why it exists to most users. Maybe a wiki section would be warranted later; there's a lot of gotchas about tinting that could be explained that way. Approved!

raeleus:
I'll make a mental note to make a wiki article on colors. That would make for a good subject. Muchas gracias.

PR #7379, opened 2024-04-08
Check for missing MetalANGLE changes on PR CI

Berstanio:
To not run into the situation where a PR is merged, that is missing generated MetalANGLE code for review, this PR adds a CI check to the PR workflow, to fail if there are outstanding MetalANGLE changes.

It might not be the most elegant solution, but it should work.
I tested it locally, but not on a GHA.


Tom-Ski (approved):

Tom-Ski:
Looks good, lets merge and keep an eye on it

PR #7380, opened 2024-04-10
Source-shade jnigen-loader into gdx

Geolykt:
This was done due to problems with JPMS as jnigen-loader shares the same package as some gdx classes.

Using traditional shading was not done as it would cause issues with the backends depending on jnigen-loader transitively (we cannot use the 'api' or 'compileOnlyApi' configurations as those wouldn't change the end result to other applications) and would also cause issues with the sources and javadoc jars not being the same as the actual sources

--

Somewhat blocking on https://github.com/libgdx/gdx-jnigen/pull/61, as jnigen-loader itself does not handle JPMS as intended at this point in time

### DIFF
--- a/gdx/build.gradle
+++ b/gdx/build.gradle
@@ -24,13 +24,27 @@ if (JavaVersion.current().isJava9Compatible()) {
 sourceCompatibility = versions.java
 targetCompatibility = versions.java
 
-sourceSets.main.java.srcDirs = ["src"]
+sourceSets.main.java.srcDirs = ["src", layout.buildDirectory.dir("jnigen")]
 sourceSets.main.resources.srcDirs = ["res"]
 sourceSets.test.java.srcDirs = ["test"]
+
+configurations {
+	jnigenLoader
+}
+
+tasks.register('unpackJnigenLoader', Copy) {
+	from {
+		configurations["jnigenLoader"].collect {
+			it.isDirectory() ? it : zipTree(it)
+		}
+	}
+	into layout.buildDirectory.dir("jnigen")
+}
 
 compileJava {
 	options.fork = true
 	options.incremental = true
+	dependsOn unpackJnigenLoader
 }
 
 jar {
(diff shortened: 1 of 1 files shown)


GustavBW:
Amazing job. Looking forward to this :)
(bump)

tommyettinger (commented):
It looks pretty clean and straightforward. There are no user-side changes to existing code needed because of this?

I'm not approving yet because I want to try the pull request out and make sure tests and such all work as intended.

Geolykt:
The behaviour of this PR is pretty much identical as if we were to bring all jnigen-loader classes into the main gdx project and drop the dependency on jnigen-loader. However, the advantage this PR has over just doing it manually is that jnigen-loader is still the only authoritative source - so no need to maintain basically two identical copies of the same software.
Over regular shading this has the advantage that the sources and javadocs match the old behaviour and the bytecode (which should ensure that OSSRH wouldn't reject the artifact on those grounds, although I don't think the semantics about sources and javadoc jars are that strong over there). 

GDX-Contributor-facing using regular shading would further have the disadvantage that is requires a larger changeset as the jnigen-loader classes would need to be defined as compileOnly (but not compileOnlyApi!) in gdx as well as all the backends making use of jnigen-loader transitively. 

NB: There are two edge-cases in my PR. One case is that IDEs might not necessarily run the unpackJnigenLoader task and thus you'd have a bunch of compilation warnings until you first build it (which then runs the task, thus making the sources available) - that being said I cannot think of any good solution to this problem that wouldn't include depending on the eclipse and IJ plugins. The other case is that if jnigen-loader were to ever remove source files it might be possible that these source files don't get deleted once the jnigen version is bumped - I'll try to address this issue soonish.

GustavBW:
> Wait, why'd you close this?

Because this issue was issued as package names between jnigen and gdx was incompatible. The root package in jnigen was renamed so this approach wasn't necessary.

PR #7381, opened 2024-04-11
Added a new interface for OpenGL objects and added a new constructor for ShaderProgram using it.

LunarOlympian:
Added a new interface which ShaderProgram implements. I also added a constructor to create a ShaderProgram using this interface. Two setters were also added to fill in missing info and allow full functionality.
This should allow shader program's functionality to be made without creating a program using OpenGL as is done with the current constructor setup.
The interface should be able to be implemented into more classes that use OpenGL, though ShaderProgram is the only one that currently implements it.

I have spent many hours I'm never getting back trying to find a way to add support for my own OpenGL shader implementation to work with ShaderProgram, and failed to find a particularly good solution. A short summary of the options I've tried, which all had their own issues:
- Extending the ShaderProgram class. It would've worked except for the constructor being called.
- Creating an unsafe instance of Shader program and using bytecode editors to add in a proxy to handle method calls.
- Reflections to create an instance of ShaderProgram with custom variable assignments.
- Directly editing the bytecode of a class instance and casting it to ShaderProgram.


crykn:
Honestly, this seems a little over-engineered for a very niche feature. Is there a reason why you are emplyoing a custom OpenGL shader implementation and still want to use ShaderProgram? 

> A short summary of the options I've tried, which all had their own issues:
> - Extending the ShaderProgram class. It would've worked except for the constructor being called.

Couldn't you simply pass two empty strings to the constructor? The shader compilation would silently fail, but you could manually set all the variables as you want.

PR #7388, opened 2024-04-21
Alternative to #7357 for setting glyph page index.

NathanSweet:
Sorry I'm late to #7357.

If `pack` returned a PixmapPackerRectangle, it could have a `Page page` field. Then we don't need a unique name, don't need to add to `rects` map, and page index lookup is `indexOf` pages rather than checking the name of all rects on all pages.

PixmapPackerRectangle is already public and exposed by PixmapPacker.Page#getRects, but all the fields are default access so it's not useful outside the package. All the fields could be public, as I'm guessing that was the intent of exposing the class.


raeleus (dismissed):
This is a much cleaner solution compared to my PR. I was worried about making API breaking changes, but for some reason I didn't think about doing it this way. Very clean. Thanks Nate!

tommyettinger (approved):
Looks good; cleaner than the approach using a hashCode() and faster than having to go through all Pixmaps that have been packed so far. I'm always in favor of making package-private code public, to boot! Approved.

I did have to run spotlessApply manually for some reason on this PR, since it wasn't building normally. The only problem was a single unused import of Rectangle in FreeTypeFontGenerator. Now it builds and I think it should be mergeable just fine.

PR #7391, opened 2024-04-24
gdx-ttf fonts bleed black color onto glyphs when they have shadows of any color

WinterAlexander:
**Problem**:

An oversight in the design of shadows in the gdx-ttf extension: When drawing fonts with shadows, it initialized an additional pixmap "shadowPixmap" without changing it's default fill color, leading it to be transparent black (0, 0, 0, 0). Then it renders the font on top which leads to bleeding the black color on the font

**Example**:

image
image

**Fix**:

Fill the shadow pixmap with the transparent color before using it, leaving transparent pixels to be the color choosen by the user.

image

**Workaround**:

While we wait for this to get merged, here's the fix in the form of a 3 classes: https://gist.github.com/WinterAlexander/5ec4bc65de744fdd90354013e5c0c0d7

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@
 - Android: Add configuration option to render under the cutout if available on the device.
 - Fix: Keep SelectBox popup from extending past right edge of stage.
 - Added Framebuffer multisample support (see GL31FrameBufferMultisampleTest.java for basic usage)
+- Fix: Fonts generated with gdx-ttf no longer bleed when drawn with a shadow
 
 [1.12.1]
 - LWJGL3 Improvement: Audio device is automatically switched if it was changed in the operating system.
--- a/CHANGES
+++ b/CHANGES
@@ -12,7 +12,7 @@
 - Android: Add configuration option to render under the cutout if available on the device.
 - Fix: Keep SelectBox popup from extending past right edge of stage.
 - Added Framebuffer multisample support (see GL31FrameBufferMultisampleTest.java for basic usage)
-- Fix: Fonts generated with gdx-ttf no longer bleed when drawn with a shadow
+- Fix: Fonts generated with gdx-freetype no longer bleed when drawn with a shadow
 
 [1.12.1]
 - LWJGL3 Improvement: Audio device is automatically switched if it was changed in the operating system.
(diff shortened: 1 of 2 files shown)


tommyettinger (approved):
This looks very reasonable, and the whole fix is just two lines! I looked through the process you went through to discover this in the scene2d channel on the Discord, and it seems very thorough. There's a very slight performance cost for filling a Pixmap, but since it should be done rather rarely, and certainly not per-frame, the quality consideration is more important. I can't think of any way that would meaningfully speed this up other than maybe checking if the transparent color is already (0,0,0,0), which would mean the fill isn't needed. Even that check is probably not that helpful for performance... I'm approving as-is, and I'd approve again if you want to add in the check for if the transparent color is already transparent black.

CHANGES
```diff
@@ -12,6 +12,7 @@
 - Android: Add configuration option to render under the cutout if available on the device.
 - Fix: Keep SelectBox popup from extending past right edge of stage.
 - Added Framebuffer multisample support (see GL31FrameBufferMultisampleTest.java for basic usage)
+- Fix: Fonts generated with gdx-ttf no longer bleed when drawn with a shadow
```
tommyettinger, line 15: I think it's called gdx-freetype, but I don't use it much so it might be called gdx-ttf in some places.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -527,7 +528,11 @@ public FreeTypeBitmapFontData generateData (FreeTypeFontParameter parameter, Fre
 				int mainW = mainPixmap.getWidth(), mainH = mainPixmap.getHeight();
 				int shadowOffsetX = Math.max(parameter.shadowOffsetX, 0), shadowOffsetY = Math.max(parameter.shadowOffsetY, 0);
 				int shadowW = mainW + Math.abs(parameter.shadowOffsetX), shadowH = mainH + Math.abs(parameter.shadowOffsetY);
-				Pixmap shadowPixmap = new Pixmap(shadowW, shadowH, mainPixmap.getFormat());
+				// use the Gdx2DPixmap constructor to avoid filling the pixmap twice
+				Pixmap shadowPixmap = new Pixmap(
+					new Gdx2DPixmap(shadowW, shadowH, Pixmap.Format.toGdx2DPixmapFormat(mainPixmap.getFormat())));
```
tommyettinger, line 533: I like this much better; it replaces one fill with transparent black with one fill by the intended color. I really doubt there's any measurable cost difference from before the PR, but the edge quality on the font should be quite a lot better now.

tommyettinger (approved):
Approved again! I wouldn't worry too much about changing "gdx-ttf" since it's not like there's some other extension that could refer to, heh.

tommyettinger (approved):
Good good, I'm sure gdx-freetype is happy to have its name used correctly!

NathanSweet:
Good find!

PR #7395, opened 2024-05-01
Correct grammatical error in the JavaDoc of the Application interface

SonicGDX:
Corrects "this facilities" to "these facilities" in Application.java


Frosty-J:
I approve of minor fixes.

This paragraph I'm not too fond of anyway. When it says you can query the operating system, I expect to be able to check if we're on Windows, macOS, ~~Solaris~~. But `getType()` isn't that granular and `UiUtils` is required instead.

PR #7399, opened 2024-05-17
Fix offsets in ConvexHull while unsorted being improperly accounted for

Geolykt:
When using ConvexHull with `sorted == false` and `offset != 0`, an ArrayIndexOutOfBoundsException used to occur due to `end` not being reset properly. This commit fixes this behaviour.

### DIFF
--- a/gdx/test/com/badlogic/gdx/math/ConvexHullTest.java
+++ b/gdx/test/com/badlogic/gdx/math/ConvexHullTest.java
@@ -1,0 +1,60 @@
+
+package com.badlogic.gdx.math;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.Arrays;
+
+import org.junit.Test;
+
+import com.badlogic.gdx.utils.FloatArray;
+
+public class ConvexHullTest {
+	@Test
+	public void testComputePolygon () {
+		ConvexHull convexHull = new ConvexHull();
+		float[] rawPolygon = new float[] {0, 0, 0, 1, 1, 1, 1, 0};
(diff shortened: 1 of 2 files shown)


gdx/test/com/badlogic/gdx/math/ConvexHullTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.Arrays;
+
+import org.junit.Test;
+
+import com.badlogic.gdx.utils.FloatArray;
+
+public class ConvexHullTest {
```
tommyettinger, line 14: Hooray, there's a test! The `assertArraySimilar()` seems a little strange, but I'm guessing that's required by the code we're testing here.

tommyettinger (approved):
Well, there's a test now, and there wasn't one before. I'll double-check ConvexHull and see if I spot anything else while we're at it. Approving in the meanwhile, since even just adding the test and checking some (maybe all!) of ConvexHull is worthwhile.

gdx/test/com/badlogic/gdx/math/ConvexHullTest.java
Geolykt: You are right - It is a little bit strange, but I wanted to conform to the specification of the computePolygon method, which are rather lax in defining what the first point is (aside that the first point is the last point, but that does not help much) - so just in case the algorithm is slightly altered and the first point might change we just want to check whether the vertices are still anti-clockwise, but not make more assumptions.

Of course, one could make more assumptions (after all, in case of changes we can just revise those assumptions no problem and I don't think the code will contain any kind of randomness, but I did not check that myself) - but I feel like that is a bit inappropriate to do within tests.

NathanSweet:
Good find, cheers!

PR #7401, opened 2024-05-20
Upgrade gha XCode to 14.3.1

SimonIT:
XCode 13.2 isn't available anymore

`macos-latest` is currently `macos-14` ([src](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) and the oldest XCode version there is 14.3.1 ([src](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#xcode))


Berstanio (approved):
I think we took care of all Xcode 14 changes (like removed armv7 and iOS 10) a while ago, so should be good!

Tom-Ski (approved):

PR #7403, opened 2024-05-27
Add FreeType Multiple Masters Variable Support

0880880:
This adds Multiple Masters Variable Support for Variable Fonts (https://freetype.org/freetype2/docs/reference/ft2-multiple_masters.html)
```java
FreeTypeFontGenerator.FreeTypeFontParameter parameter = new FreeTypeFontGenerator.FreeTypeFontParameter();
parameter.size = 32;
parameter.color.set(Color.BLACK);
parameter.variableDesignCoordinates = new int[] {axis.getMaximum()};
```

### DIFF
--- a/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
+++ b/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
@@ -37,6 +37,7 @@ public class FreeType {
 	// @off
 	/*JNI
 	#include <ft2build.h>
+	#include <freetype/ftmm.h>
 	#include FT_FREETYPE_H
 	#include FT_STROKER_H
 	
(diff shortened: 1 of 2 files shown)


tommyettinger:
I think this looks like it could be really great! From Discord: 
image

I'll review it later, though I don't use FreeType heavily at the moment. One potential issue is that this will be incompatible with the unofficial gwt-freetype solution people have been using -- why can't that be brought into the main repo, again?

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java
```diff
@@ -801,5 +836,35 @@ public static class FreeTypeFontParameter {
...
+
+	public class VariableAxis {
+		private int minimum;
+		private int maximum;
+		private int def;
+
+		public VariableAxis (FreeType.VarAxis varAxis) {
+			this.minimum = (int)varAxis.getMinimum() / 65536;
+			this.maximum = (int)varAxis.getMaximum() / 65536;
+			this.def = (int)varAxis.getDef() / 65536;
```
tommyettinger, line 853: I believe this is not the right behavior some of the time, but I'm not sure. On the docs, https://freetype.sourceforge.net/freetype2/docs/reference/ft2-multiple_masters.html#FT_Var_Axis , minimum and maximum are FT_FIXED type, roughly equivalent to a float I think here, but on FreeType's side they're a signed `long` (it's C, so size is compiler- and platform-dependent). If it really is the 32-bit representation of a 16.16 fixed-point number, then the sign bit matters for how division works vs. a bitwise shift. The minimum and maximum shouldn't ever be negative conceptually, but they're being converted from a `long` (in Java), so I don't know if there could be a sign bit set as a result of that (if minimum or maximum is a large-enough positive `long`, then casting to `int` can produce a negative `int`).

Long story short, if you want `this.minimum` and `this.maximum` to be non-negative `int`s in the 0-65535 range, I would use `this.minimum = (int)varAxis.getMinimum() >>> 16;`, but if negative min and max are fine, then this can stay as-is.

extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeType.java
```diff
@@ -341,6 +342,186 @@ public int getCharIndex(int charCode) {
...
+			super(address);
+		}
+
+		// No getCoords
+
+		private static native int getStrid(long varNamedStyle); /*
+			return (jint) (((FT_Var_Named_Style *) varNamedStyle)->strid);
+		*/
+
+		public int getStrid() {
```
tommyettinger, line 513: It might be worthwhile (if this faces the user at all) to just put a quick note with what the docs say for a parameter or just link to the FreeType docs at the top of the file, in a comment. https://freetype.sourceforge.net/freetype2/docs/reference/ft2-multiple_masters.html I did a double-take at "getStrid()" because I assumed it was a typo of "stride" -- I was surprised to learn that's the name used by FreeType, though the variable apparently doesn't matter to how FreeType works.

tommyettinger (changes requested):
I'm just hoping to validate if the behavior regarding minimum and maximum, plus any other FT_FIXED variables, is what you intend it to be. A quick link to the docs, anywhere in added classes in FreeType.java, would be nice, but not required. Feel free to @ mention me when you want me to review again; this looks really useful!

tommyettinger (approved):

PR #7404, opened 2024-05-30
Added a method to check if a sphere intersects a bounding box.

GadWissberg:
Hello, I found myself several times using the method for checking intersection between a sphere and a bounding box introduced in the following answer by the user `nerdinand`:
https://stackoverflow.com/a/15247348/3450476

I thought it'd be useful to include it in the Intersector class instead of going back to that answer each time.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -345,35 +345,33 @@ public final class Intersector {
 		return v2a.len2() <= circle.radius * circle.radius;
 	}
 
-	/**
-	 * Checks if a sphere intersects with a bounding box.
+	/** Checks if a sphere intersects with a bounding box.
 	 *
 	 * @param sphere the sphere to check for intersection
 	 * @param bounds the bounding box to check for intersection
-	 * @return true if the sphere intersects the bounding box, false otherwise
-	 */
-	public static boolean intersectSphereBounds(Sphere sphere, BoundingBox bounds) {
+	 * @return true if the sphere intersects the bounding box, false otherwise */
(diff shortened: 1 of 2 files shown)


SinfullySoul:
can OrientedBoundingBox and Sphere intersects method also be implement?

gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -348,6 +349,38 @@ public static boolean intersectSegmentCircle (Vector2 start, Vector2 end, Circle
...
+	 * @param bounds the bounding box to check for intersection
+	 * @return true if the sphere intersects the bounding box, false otherwise */
+	public static boolean intersectSphereBounds (Sphere sphere, BoundingBox bounds) {
+		float dmin = 0;
+
+		Vector3 boundsMin = bounds.getMin(v0);
+		Vector3 boundsMax = bounds.getMax(v1);
+
+		if (sphere.center.x < boundsMin.x) {
+			dmin += (float)Math.pow(sphere.center.x - boundsMin.x, 2);
```
tommyettinger, line 364: I'm not a big fan of using Math.pow() for extremely simple math, since pow() seems to be quite a bit slower than just a single multiply on a float variable. Intersection and collision detection code can be called potentially quite a lot of times per frame, so this is somewhere where the performance may matter. It's possible that `Math.pow(n, 2)` gets optimized to something simpler, but this still needs to convert n to double, and then convert that to float... I've heard that at least `Math.sqrt()` will operate differently if given a float and immediately cast to float, but I don't know about `pow()`.

Effectively what I'm suggesting is:
```java
		if (sphere.center.x < boundsMin.x) {
			float a = sphere.center.x - boundsMin.x;
			dmin += a * a;
		}
```

PR #7406, opened 2024-06-06
Deserialization of tiled maps which were changed from the code

MGDSStudio:
The class com.badlogic.gdx.maps.tiled.TiledMapTileLayer should have a default constructor. The programmer can serialize the whole tiled map in a single JSON-file but can not deserialize it back.


tommyettinger (changes requested):
This probably won't work as you hope. Because `cells` is never assigned with the zero-argument constructor (nor are width and height), and there isn't a getter to assign them, the only way deserialization could make a valid TiledMapTiledLayer would be to hack in values for those private fields using reflection. The better way would be for you to write a custom Json.Serializer for this type, which doesn't need a zero-argument constructor, or any changes to TiledMapTiledLayer.

PR #7410, opened 2024-06-26
JsonReader improvements, added JsonSkimmer, JsonString

NathanSweet:
* Minor JsonReader changes.
* Fixed logic error parsing multi line comments.
* Don't use an array of names. It has always been an array, but there doesn't appear to be any case that can have more than one name.
* Added JsonSkimmer. I use this when parsing large JSON where I only care about a couple values. It's nice not to parse every double and long just to discard them.
* Added JsonString, like JsonWriter but doesn't throw IOException.


gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+									else {
+										_trans += ((_mid - _keys) >> 1);
+										break _match;
+									}
+								}
+								_trans += _klen;
+							}
+						} while (false);
+
+						_trans = _json_indicies[_trans];
```
tommyettinger, line 176: Minor typo: `indicies` should be `indices` . This is only used internally, so it doesn't really matter.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+									if (data[p++] == '/') {
+										while (p != eof && data[p] != '\n')
+											p++;
+										p--;
+									} else {
+										while (p + 1 < eof && data[p] != '*' || data[p + 1] != '/')
+											p++;
+										p++;
+									}
+									if (debug) System.out.println("comment " + new String(data, start, p - start));
```
tommyettinger, line 284: Should we be using `Gdx.app.debug()` here, instead of `System.out.println()` ?

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+	static final int json_en_array = 23;
+	static final int json_en_main = 1;
+
+	// line 269 "JsonSkimmer.rl"
+
+	private boolean stop;
+
+	/** Causes parsing to stop after the current or next object, array, or value. */
+	public void stop () {
+		stop = true;
```
tommyettinger, line 530: How would one start parsing back up after it stops, if that makes sense at all? Would any starting-back-up make more sense as a new parsing attempt?

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	protected void startArray (@Null String name) {
+	}
+
+	protected void pop () {
+	}
+
+	protected void value (@Null String name, String value, boolean unquoted) {
+	}
```
tommyettinger, line 591: These seem to be intended to be implemented by subclasses as needed, but without any documentation, it's hard to know when these get called, what they do, or what they can be used to do. I'm guessing a name that is null indicates there was no name in that case.

tommyettinger (changes requested):
Wow, there's a lot here! I have mostly very insignificant complaints here, except about the documentation in JsonSkimmer. I wouldn't have much idea how to use the class if I had it stuck in front of me, and even the test, which is certainly helpful, raises questions about whether the typical usage is via a subclass and what the implemented methods do/when they get called. Since so much of these Json classes has been auto-generated and is basically unreadable to humans without a lot of effort, having human-authored docs helps indicate that a part of the file is intended to be read and used. The changes I'm requesting are just some quick docs that explain what the user-callable methods do and when to use them or override them. It would also be good, in my opinion,  for unescape() to be `protected` because of the possibility of unusual escaping situations, such as the octal escapes `\0`, `\00`, and `\000`. If users encounter that, they can handle it themselves then, and we don't need to guess at every unusual situation they could encounter.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+	/** Causes parsing to stop after the current or next object, array, or value. */
+	public void stop () {
+		stop = true;
+	}
+
+	public boolean isStopped () {
+		return stop;
+	}
+
+	private String unescape (String value) {
```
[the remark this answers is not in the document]
NathanSweet: That's never come up for JsonReader, but I don't mind it being protected.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
+
+// line 1 "JsonSkimmer.rl"
+// Do not edit this file! Generated by Ragel.
+// Ragel.exe -J -o ../../../../../src/com/badlogic/gdx/utils/JsonSkimmer.java JsonSkimmer.rl
```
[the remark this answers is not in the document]
NathanSweet: I've never heard of Ragel dropping Java support. I don't expect that will happen. I've added the version though.

Ragel has been a stable tool for a very long time. I used the latest stable version (6.10) which was published in 2009. There's an experimental 7.x version that is written in Colm (crazy stuff, [check out](https://github.com/adrian-thurston/colm/tree/master/grammar) the C++ or PCRE grammars). It's been in development a long time and presumably will not break older Ragel code.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
NathanSweet: Yeah that looks odd. It's generated though.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
NathanSweet: Probably! There's no `print` (without newline) version, which makes the `indent` method more annoying to write. There are many tests that uses sysout, so I'm inclined to leave it. It could be a separate initiative to change all those.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
NathanSweet: When you stop you lose all parser state machine state (the locals `cs`, `p`, etc). I can't think of a use case where you want to start/stop parsing. Usually you `stop()` when you've gotten the data you need and don't want to parse the rest at all.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
NathanSweet: I've added javadocs. They are for event based parsing where you handle JSON constructs as they are encountered.

I also changed to `push` with a `boolean object`. I'm torn which I like better. The single `push` method is shorter when you only count the level for both objects and arrays, which is common.

I added a real world usage example. It's maybe a bit of a verbose/nasty way to parse JSON, but it's pretty fast (I say having never written benchmarks) and beats trying to parse with regex.

NathanSweet:
@tommyettinger Thanks for taking a look! I think I addressed all the things.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
tommyettinger: Hooray docs! Hooray usage example! Definitely beats parsing JSON with regex.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
tommyettinger: Makes sense, seems fine then.

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
tommyettinger: Hm, it isn't in a test, but debug probably only gets turned on in tests, so I'm not too worried. It doesn't seem likely that people will be debugging this on platforms that have trouble with System.out .

gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -0,0 +1,592 @@
...
+
+	private static final byte _json_eof_actions[] = init__json_eof_actions_0();
+
+	static final int json_start = 1;
+	static final int json_first_final = 35;
+	static final int json_error = 0;
+
+	static final int json_en_object = 5;
+	static final int json_en_array = 23;
+	static final int json_en_main = 1;
```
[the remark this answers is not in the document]
tommyettinger: Ah, makes sense. It's fine I suppose; having to manually edit generated code would be worse.

tommyettinger (dismissed):

tommyettinger:
Tests currently fail (and it's not actually Spotless this time!) with:
```
/home/runner/work/libgdx/libgdx/tests/gdx-tests/src/com/badlogic/gdx/tests/JsonTest.java:223: error: local variable values is accessed from within inner class; needs to be declared final
					values.add(id);
					^
/home/runner/work/libgdx/libgdx/tests/gdx-tests/src/com/badlogic/gdx/tests/JsonTest.java:224: error: local variable values is accessed from within inner class; needs to be declared final
					values.add(watts);
					^
```

NathanSweet:
Oops, fixed! I'm ruined by modern Java.

tommyettinger:
That allows it to achieve a new failure!
```

> Task :tests:gdx-tests-gwt:compileGwt
   Tracing compile failure path for type 'com.badlogic.gdx.tests.JsonTest'
Error: ERROR] Errors in 'file:/home/runner/work/libgdx/libgdx/tests/gdx-tests/src/com/badlogic/gdx/tests/JsonTest.java'
Error:   [ERROR] Line 174: No source code is available for type com.badlogic.gdx.utils.JsonSkimmer; did you forget to inherit a required module?
Error: OR] Aborting compile due to errors in some input files

> Task :tests:gdx-tests-gwt:compileGwt FAILED

```

I suspect this can be solved by adding JsonSkimmer and JsonString to the .gwt.xml file that lists every class and whether it is excluded or not. I'm not sure why this approach of manually listing each file is used, instead of the `GwtIncompatible` annotation that GWT has supported for as long as I have used it. That approach would assume all classes are GWT-compatible unless marked with the `GwtIncompatible` annotation, which makes the GWT compiler skip that class, method, or whatever. The name `GwtIncompatible` is all that matters, it can be defined in libGDX code and no dependency on GWT is needed.

...But the easiest approach here is to add those Json-related files to the big .gwt.xml that lists JsonReader and similar files already.

NathanSweet:
GWT strikes again! I've done the needful, I think.

tommyettinger (approved):
All right, that looks like everything! GWT tests appear to pass, or at least they compile. Spotless isn't complaining at all. These seem like useful features, I think; I don't really know how much time difference there will be, but having this lets us test it in a benchmark. Approved!

SomeTroglodyte:
Any documentation or examples for `JsonSkimmer`?
https://libgdx.com/wiki/utils/reading-and-writing-json - nope

NathanSweet:
2 in JsonTest, otherwise feel free to add some. JsonSkimmer is a low level, minimal parser. JsonMatcher is likely a better choice for most needs.

Issue #7411, opened 2024-06-26
Petition for new API: Color.set(Color rgb, float a)


angelickite:
Hello.

I constantly find myself writing code like this:
```Java
color.set(DB32.black);
color.a = <desired alpha>;
```

Ideally libgdx would offer `Color.set(Color rgb, float a)`. Setting a color is generally a unit operation. In this case the unit operation needs to be split into two distinct lines like above. Doing it here and there is no biggy, but I encounter this operation constantly.

> You can use a constant

Constants like `Color myBlack = new Color(DB32.black.r, DB32.black.g, DB32.black.b, <desired alpha>);` do not help the issue. The alpha value often is dynamic in nature, e.g. an interpolated value.

> You can create your own subclass of Color

No! We know the mess that leads to. Also it doesn't solve the problem neatly as the libgdx API delivers `Color` not `T extends Color`.

> You can use your own Utility

Thats what I do. It does its job but a method calls turns into a function call. Nothing beats native support.

Wonder how everyone feels about this.

Thank you!

tommyettinger:
It's been merged! We can close now.

PR #7412, opened 2024-06-28
[GH-7411] New API: Color.set(Color rgb, float a)

angelickite:
My implementation of #7411

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -104,13 +104,27 @@ public class Color {
 
 	/** Sets this color to the given color.
 	 * 
-	 * @param color the Color */
+	 * @param color the Color
+   * @return this color. */
 	public Color set (Color color) {
 		this.r = color.r;
 		this.g = color.g;
 		this.b = color.b;
 		this.a = color.a;
 		return this;
+	}
+
+	/** Sets this color using the red, green and blue components of the provided Color and a deviating alpha value.
+	 * 
+	 * @param rgb the desired red, green and blue values (alpha of that Color is ignored)
+	 * @param float the desired alpha value
+	 * @return this color. */
+	public Color set (Color rgb, float alpha) {
+	  this.r = rgb.r;
+	  this.g = rgb.g;
+	  this.b = rgb.b;
+	  this.a = alpha;
+	  return this;
 	}
 
 	/** Multiplies the this color and the given color
--- a/gdx/src/com/badlogic/gdx/graphics/Color.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Color.java
@@ -16,6 +16,7 @@
 
 package com.badlogic.gdx.graphics;
 
+import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.utils.NumberUtils;
 
 /** A color class, holding the r, g, b and alpha component as floats in the range [0,1]. All methods perform clamping on the
@@ -105,7 +106,7 @@ public class Color {
 	/** Sets this color to the given color.
 	 * 
 	 * @param color the Color
-   * @return this color. */
+	 * @return this color. */
 	public Color set (Color color) {
 		this.r = color.r;
 		this.g = color.g;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -104,7 +104,8 @@ public Color (Color color) {
 
 	/** Sets this color to the given color.
 	 * 
-	 * @param color the Color */
+	 * @param color the Color
+   * @return this color. */
```
tommyettinger, line 108: Ah, my nemesis, indentation! I would run `gradlew spotlessApply` and commit the changed file. It may initially look like thousands of changed files, but it is actually only going to commit this one.

tommyettinger (changes requested):
I really like the idea here! It makes Color more usable, and it's such a commonly-used class that that is a nice quality-of-life improvement. You'll need to either run `gradlew spotlessApply` or manually correct the indentation for the build to pass. Once that's done, I see nothing this would possibly break, so merging should be easy.

gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -113,6 +114,19 @@ public Color set (Color color) {
...
+	/** Sets this color using the red, green and blue components of the provided Color and a deviating alpha value.
+	 * 
+	 * @param rgb the desired red, green and blue values (alpha of that Color is ignored)
+	 * @param float the desired alpha value
+	 * @return this color. */
+	public Color set (Color rgb, float alpha) {
+	  this.r = rgb.r;
+	  this.g = rgb.g;
+	  this.b = rgb.b;
+	  this.a = alpha;
```
mgsx-dev, line 126: i think that alpha should be clamped to be consistent with other setters.

mgsx-dev (changes requested):

gdx/src/com/badlogic/gdx/graphics/Color.java
tommyettinger: It could also call the existing set(float, float, float, float) method, which should clamp already.

mgsx-dev: i don't think so, there is no need to clamp rgb because it's supposd to be clamped and to be consistent with setColor(Color) which doesn't clamp components.

gdx/src/com/badlogic/gdx/graphics/Color.java
angelickite: Should be fixed. Took me a bit to figure out that Java 21 is not supported...

gdx/src/com/badlogic/gdx/graphics/Color.java
angelickite: Good catch. I've made use of `MathUtils.clamp`. Didn't see the immediate benefit to inline it given the powerhouse that is modern JVM, though I can be convinced otherwise.

angelickite: Not sure if the dependency is desired though, that is up to you guys.

mgsx-dev (approved):

tommyettinger (approved):

gdx/src/com/badlogic/gdx/graphics/Color.java
tommyettinger: In all of my recent code, I use `Math.min(Math.max(value, lowerBound), upperBound)` because I have noticed a measurable performance improvement on recent JVMs (16 and up, maybe slightly newer). That said, inlining is tricky, and if MathUtils.clamp() gets inlined but the core Math methods somehow do not, or if you're running on Android, iOS, or GWT, things are quite different. The clamp() approach could certainly be faster on Android, but I have no idea how to benchmark on that OS with the wide range of hardware and OS versions.

tommyettinger:
I just merged this, even though I had one small concern. I think this potential quibble is minor and unlikely, but an entry in the changelog might be desirable. It's a pretty small (but nice) API addition, but it is an API addition... I think it should be fine without it, myself. Unless CHANGES really needs another line.

PR #7415, opened 2024-07-08
Fix: Set alpha channel bits to 8 for Android backend

Raxorg:
- Changed the default alpha channel bits from 0 to 8 in the Android backend.
- This resolves rendering issues with textures and masks that require alpha blending.
- Ensures consistent behavior between Android and desktop backends.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplicationConfiguration.java
@@ -30,7 +30,7 @@ import com.badlogic.gdx.utils.GdxNativesLoader;
  * @author mzechner */
 public class AndroidApplicationConfiguration {
 	/** number of bits per color channel **/
-	public int r = 8, g = 8, b = 8, a = 0;
+	public int r = 8, g = 8, b = 8, a = 8;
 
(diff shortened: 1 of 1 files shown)


Tom-Ski:
wow!

obigu:
@Tom-Ski Sorry for comment after being closed but should this line be updated as well? https://github.com/libgdx/libgdx/blob/f3423c144644b6c6c205620b0a3fec13ad434741/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L97

Tom-Ski:
> @Tom-Ski Sorry for comment after being closed but should this line be updated as well? https://github.com/libgdx/libgdx/blob/f3423c144644b6c6c205620b0a3fec13ad434741/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L97

Yeah looks like it

Raxorg:
@obigu I implemented your suggestions regarding BufferFormat in #7641, let me know what you think

obigu:
> @obigu I implemented your suggestions regarding BufferFormat in #7641, let me know what you think

Cool, it's what I suggested but I can't review the PR from a functionality perspective as I don't know what I need to test to verify it works as intended. @Tom-Ski is probably more indicated for that.

PR #7416, opened 2024-07-08
Debug colors' access modifiers set to public

Raxorg:
Ensures consistent behavior across backends and fixes a (The field Box2DDebugRenderer.XXX is not visible) error thrown when changes to the colors are made in the core package and deployed to web


tommyettinger (approved):
I don't use Box2D myself, but I know how problematic it is when there's a mismatch between GWT visibility and visibility on other platforms. I don't see any reason these Color instances would need to be private in the emulated sources, so they should really match the main Box2D code. Approved.

PR #7421, opened 2024-07-19
Lessen device requirements for `vibrate(VibrationType)`

Frosty-J:
Cheaper devices such as the Galaxy A04s can utilise these predefined effects. They don't have to be limited to devices with amplitude control. `TICK`, `CLICK` and `HEAVY_CLICK` seem to be 10, 20, and 30ms respectively, though this may be vendor-specific.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidHaptics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidHaptics.java
@@ -60,7 +60,7 @@ public class AndroidHaptics {
 
 	@SuppressLint("MissingPermission")
 	public void vibrate (Input.VibrationType vibrationType) {
-		if (hapticsSupport) {
+		if (vibratorSupport) {
 			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
 				int vibrationEffect;
 				switch (vibrationType) {


obigu:
This was done on purpose to filter out non haptics devices which provide a very poor experience even if they do implement the API. We could expand the API to support those devices by adding a `fallback` boolean in a similar manner to `vibrate(int milliseconds, int intensity, boolean fallback)` with a `vibrate(VibrationType, boolean)` which would cover all cases without breaking existing code and functionality.

Frosty-J:
This is turning into quite the number of overloads. Would it be reasonable for the fallback to implement vibration on older versions of Android? If so, there's the question of durations; if not, less thought.

obigu:
> This is turning into quite the number of overloads. Would it be reasonable for the fallback to implement vibration on older versions of Android? If so, there's the question of durations; if not, less thought.

I think it would be reasonable, yeah. Because it's a fake implementation, I'd test it on an old device and try to make it feel as good as possible. I'd try setting the `LIGHT` to the shortest possible duration to make it feel really light and from there, as you suggest, try x2 and x3 (or even x4) for `MEDIUM` and `HEAVY`. The iOS implementation of this overloaded method would need to throw an `UnsupportedMethodException` (independently if fallback is set to true or false).

Frosty-J:
I've played around with some different devices now, half of which are too old to for libGDX (not that the cheapo vibration technology has changed much, but smartphones have gotten bigger). **25, 50, 100ms** is working pretty well, but I have no frame of reference to how this compares with a device with proper haptics. I think those are the values I will go with in my game, with a setting to double/halve/disable them. [MrStahlfelge as used a much wider range](https://github.com/MrStahlfelge/lightblocks/blob/master/core/src/de/golfgl/lightblocks/input/VibrationType.java), from 20ms all the way up to 300ms!

The oddball in my collection is the Galaxy S5. This can pretty much immediately turn on and off its vibration, unlike other devices that presumably have to spin up a weight. It's failing to vibrate reliably in LineageOS - its driver is terrible. The stock firmware had a setting to change the vibration intensity, but it never got updated to an Android version that exposes amplitude control to apps (and maybe it wouldn't be granular enough for that).

PR #7424, opened 2024-07-25
Added FPSLogger.setBound()

TCreutzenberg:
Added FPSLogger.setBound()

Useful when e.g. the game window is moved to another display with a different refresh rate and fps logging would return false results.


Frosty-J (approved):

gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
```diff
@@ -33,6 +33,10 @@ public FPSLogger () {
 
 	/** @param bound only logs when they frames per second are less than the bound */
 	public FPSLogger (int bound) {
+		setBound(bound);
+	}
+
+	public void setBound (int bound) {
 		this.bound = bound;
 		startTime = TimeUtils.nanoTime();
```
crykn, line 41: I think this line should be moved back into the constructor. There isn't really a reason to reset the start time after changing bounds.

TCreutzenberg: @crykn My reasoning behind it was this:

Gdx.graphics.getFramesPerSecond() actually counts the frames (at least in the lwjgl3 implementation).

If you change fps to something lower than before all is fine:
E.g. 120 FPS to 60 FPS (after 0.5 seconds of setting startTime):
Gdx.graphics.getFramesPerSecond() will either return 90:
90 > 60 so no log entry (which is correct)

If you change from lower to higher FPS though the FPSLogger will add a log about the FPS being too low:
E.g. 60 FPS to 120 FPS (after 0.5 seconds of setting startTime):
Gdx.graphics.getFramesPerSecond() will either return 90:
90 < 120 which will result in a log entry which is actually wrong as the FPS were changed manually

obigu: Not a fan of an overridable method call in the constructor either.

crykn: @TCreutzenberg: resetting the starting time in `FPSLogger#setBounds` still doesn't fix this completely, does it? The wrong result of `Gdx.graphics.getFramesPerSecond()` is caused by the timer in `Lwjgl3Graphics` and not by the timer in `FPSLogger`. E.g.:
- The game runs with 60 fps for 1,75 seconds 
- Now we change the screen, the refresh rate is different an we continue rendering with 120 fps. At the same time, we use `FPSLogger#setBounds(120)`.
- After 1 second (so 2,75 seconds after starting the game), FPSLogger is called and polls `Gdx.graphics.getFramesPerSecond()` which returns 75 (and not 120!). The reason for that is that the last time the fps value was updated was 2 seconds after application start and back then we rendered with 60 fps for 0,75 seconds (= 45 frames) and with 120 fps for 0,25 seconds (= 30 frames).

The ideal solution would be adapting the fps calculation in `Lwjgl3Graphics` in such a way that whenever `setForegroundFPS` is used or vsync & a window change happens, we'd reset `frameCounterStart`. Changing this is definitely outside the scope of this PR and I guess a partial fix is better than none, so I'm approving. Should anyone decide to tackle it in the future, this comment here should serve as a good starting point.

crykn (dismissed):

gdx/src/com/badlogic/gdx/graphics/FPSLogger.java
TCreutzenberg: @crykn Good point!

PR #7426, opened 2024-07-27
Remove gdx-setup

Frosty-J:
I take no pleasure in doing this, but if gdx-setup is to no longer be maintained and cannot be downloaded via the website, I feel it's better off removed from the repository by the next release. I have noticed all issues/PRs pertaining to it have been closed.

Obviously it will remain in the repository history and the sources can be accessed quite easily via the [tags](https://github.com/libgdx/libgdx/tags) we have. I've updated my shabby [gdx-setup archives](https://github.com/JavaCakeGames/gdx-setup-archive) with the latest JARs, for what good that does.

The website's versions page depends on gdx-setup's dependency bank, but that will have to change sooner or later anyway. I'm happy to disable its script until a better solution is found. Further discussion at https://github.com/libgdx/libgdx.github.io/issues/203.


tommyettinger (approved):
Looks good to me. I'm glad you found any usages in Gradle files, not just the code and resource files in gdx-setup's directory. The GH workflow change looks necessary here, and does save time and resources on S3, but will need approval. The one line added to CHANGES had me do a little jig. Approved.

Issue #7297, opened 2023-12-09
Misbehavior (with fix) on iOS when using FrameBuffer


b3nk4n:
#### Issue details

I've been facing an issue on iOS only when using `FrameBuffer`. When I ported my game [October Bro](https://play.google.com/store/apps/details?id=de.bsautermeister.jump) from Android/Desktop to iOS, I noticed that my game screen was just black (or just the screen-clear color I used).

After longer investigation and digging into the LibGdx code, I could figure out the problem, and come up with a workaround. And my hope is that this issue report turns into a simple fix PR (either by me or by you), so that my workaround will not be needed anymore.

#### Reproduction steps/code

Use the following code on iOS, and toggle the `FIX` flag.

```java
public class MyGdxGame extends ApplicationAdapter {

	private static final boolean FIX = true; // <-- please toggle me! :)

	SpriteBatch batch;
	Texture img;

	FrameBuffer frameBuffer;

	OrthographicCamera camera;

	boolean created = false;

	@Override
	public void create () {
		batch = new SpriteBatch();
		img = new Texture("badlogic.jpg");

		camera = new OrthographicCamera();

		created = true;
		frameBuffer = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false);
	}

	@Override
	public void render () {
		Gdx.gl.glClearColor(1f, 0f, 0f, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);

		IntBuffer oldFbo = BufferUtils.newIntBuffer(1);
		Gdx.gl.glGetIntegerv(GL_FRAMEBUFFER_BINDING, oldFbo);

		frameBuffer.begin();

		Gdx.gl.glClearColor(0f, 1f, 1f, 1);
		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);

		camera.setToOrtho(true, frameBuffer.getWidth(), frameBuffer.getHeight());
		camera.update();

		Gdx.gl.glClearColor(1.0f, 1.0f, 1.0f, 255 / 255.0f);

		batch.begin();
		batch.draw(img, 0, 0);
		batch.end();

		frameBuffer.end();

		if (FIX) {
			// use the previous back buffer index, instead of defaultFramebufferHandle from GLFrameBuffer#unbind()
			int fboIdx = oldFbo.get();
[...]

PR #7428, opened 2024-07-30
force a dummy render to make sure gl is initialized at the time of create

Tom-Ski:
addresses https://github.com/libgdx/libgdx/issues/7297

Simple dummy frame during init process. Tested this with the supplied test in the issue report


obigu (approved):
Tested on a couple of games both on device and simulator and can't see any regression related to black frames or delays in rendering. Haven't tested the actual frame buffer issue fix.

obigu (approved):

Berstanio (approved):

Issue #6687, opened 2021-10-25
confine mouse inside fullscreen app


judos:
#### Issue details

For my game I want map scrolling when the mouse touches the screen border (very common in RTS games, Starcraft, League of legends etc.). This requires however that the mouse does not leave the fullscreen app when there is a second monitor. So basically I need the mouse to stay inside or "confine in" (hope that's the correct term) the fullscreen app.

Is there a way to correctly achieve this? Right now I only see the following workarounds but neither works good enough in my opinion:

- `Gdx.input.isCursorCatched = true` useful for FPS Games, the mouse now has infinite space, and doesn't leave the app but also the cursor isn't shown anymore. In this case I would need to draw a mouse cursor myself. When I do this I however notice that the cursor lags behind the actual mouse significantly.

In the following pictures I did not set the cursorCatched flag, such that the actual cursor is visible. Also I drew two cursors, the current Gdx.input.x/y and the coordinates of the last frame (60 frames/s app, same as monitors).
![cursor pictures](https://user-images.githubusercontent.com/5525810/138718857-8935fa90-57a5-4449-9451-b31533a8696f.png)
The two cursor close to each other, are the custom ones I draw in the app.
![mouse-lag](https://user-images.githubusercontent.com/5525810/138719020-06c7d40d-4854-4767-8fde-6db491624700.gif)
This is recorded with ScreenToGif also at 60fps and then slowed down to 2fps to demonstrate the lag.

- `Gdx.input.setCursorPosition(...)` I could manually set the mouse coordinates to stay inside the screen, but mouse coordinates are no longer reported when the mouse is outside the fullscreen app.

- `java.awt.Robot.mouseMove(...) and MouseInfo.getPointerInfo().location` this seems to work a bit better regarding update speed and keeping the cursor inside the window. However I don't know how to get the appropriate screen coordinates where the window/fullscreen app is positioned in the multi-monitor setup.

#### Related Information (Fullscreen modes)
I only read in this [github ticket](https://github.com/libgdx/libgdx/issues/4785) that the difference between exclusive fullscreen and borderless window ("simulated" fullscreen) could also make a difference. Right now I'm just using this code to enter fullscreen:
```
val monitor = Lwjgl3ApplicationConfiguration.getPrimaryMonitor()
val currentMode = Lwjgl3ApplicationConfiguration.getDisplayMode(monitor)
Gdx.graphics.setFullscreenMode(currentMode)
```
[...]

PR #7430, opened 2024-08-03
Prevent mouse cursor from being hidden when catched using GLFW_CURSOR_CAPTURED

soundasleep:
I'm not sure if this is intentional but when calling `setCursorCatched()`, the cursor is hidden, requiring the user to implement laggy software cursors: https://github.com/libgdx/libgdx/issues/6687

However GLFW supports `GLFW_CURSOR_CAPTURED` which keeps the cursor captured, but not hidden. Is there a reason libgdx uses `GLFW_CURSOR_DISABLED` and not `GLFW_CURSOR_CAPTURED`?

A workaround is to extend your Lwjgl3Application like so:

```java
	private static class BetterLwjgl3Input extends DefaultLwjgl3Input {
		
		private final Lwjgl3Window window;

		public BetterLwjgl3Input(Lwjgl3Window window) {
			super(window);
			this.window = window;
		}

		@Override
		public void setCursorCatched (boolean catched) {
			// change from Default input: use GLFW_CURSOR_CAPTURED not GLFW_CURSOR_DISABLED
			// so that the captured cursor is not invisible
			GLFW.glfwSetInputMode(window.getWindowHandle(), GLFW.GLFW_CURSOR,
				catched ? GLFW.GLFW_CURSOR_CAPTURED : GLFW.GLFW_CURSOR_NORMAL);
		}

		@Override
		public boolean isCursorCatched () {
			// change from Default input: use GLFW_CURSOR_CAPTURED not GLFW_CURSOR_DISABLED
			// so that the captured cursor is not invisible
			return GLFW.glfwGetInputMode(window.getWindowHandle(), GLFW.GLFW_CURSOR) == GLFW.GLFW_CURSOR_CAPTURED;
		}
		
	}
```

This _may_ not work on Cocoa Mac OS X (I'm not sure if that's relevant): https://github.com/glfw/glfw/blob/master/src/cocoa_window.m#L1647

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java
@@ -326,12 +326,12 @@ public class DefaultLwjgl3Input extends AbstractInput implements Lwjgl3Input {
 	@Override
 	public void setCursorCatched (boolean catched) {
 		GLFW.glfwSetInputMode(window.getWindowHandle(), GLFW.GLFW_CURSOR,
-			catched ? GLFW.GLFW_CURSOR_DISABLED : GLFW.GLFW_CURSOR_NORMAL);
+			catched ? GLFW.GLFW_CURSOR_CAPTURED : GLFW.GLFW_CURSOR_NORMAL);
 	}
 
 	@Override
 	public boolean isCursorCatched () {
-		return GLFW.glfwGetInputMode(window.getWindowHandle(), GLFW.GLFW_CURSOR) == GLFW.GLFW_CURSOR_DISABLED;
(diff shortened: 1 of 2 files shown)


Frosty-J:
The idea of `setCursorCatched()` is the mouse cursor gets reset to the middle of the window each frame. You can then check the mouse deltas for use with games where you want the user to be able to move the mouse in a direction infinitely, such as for first-person shooters. Having it be visible while it's snapping back to centre would be unpleasant and it sounds like you've got confused as to the purpose of this method.

That said, it would be useful if we can confine the mouse cursor to the window when the cursor is "catched" as extremely fast movements can currently cause it to momentarily be visible outside the window boundaries.

soundasleep:
OK, that's good to know! I'm creating a 2D game so I don't need the cursor position to be reset.

Maybe a separate `setCursorCaptured()` would be better, although I'm sure an enum would be even betterer?

Frosty-J:
Sure, the enum idea went down well in the issue you linked.

Minikloon:
I'd also like this functionality in LibGDX, but I don't think it should be a breaking change like this PR.

PR #7433, opened 2024-08-09
Added comments about the order of logs 

TCreutzenberg:
Added comments to three of the logging `ApplicationLogger`s to clarify the order of the log messages.


Frosty-J:
Changing all instances to `System.out` will break error logging for developers who rely on the distinction. Does forcing a flush after each print not help?

TCreutzenberg:
Hm, actually didn't know there's a whole workflow around System.err. Learned something new! System.err can be redirected to a file, etc. cool!
So my solution obviously is not good, and it's also not printing the error stuff and red, missing the ANSI codes...
System.out.flush() / System.err.flush() does not seem to be very good for performance. MAybe another approach would be better?

1. Specifying the error-PrintStream? (Default to System.out and adding ANSI codes to print the error logs red?)
2. Specifying if logging should .flush() after each call having a small performance impact?

crykn (approved):

PR #7435, opened 2024-08-11
Update to GWT 2.11.0

tommyettinger:
This removes the need for any third-party backend to use GWT 2.11.0 . It updates the Maven coordinates to match the current ones for GWT itself (jsinterop didn't change). It updates the DTDs to match 2.11.0 and a working URL for that. It didn't need much else; some version strings needed changing.

The tests run, though they don't behave especially well (and I don't think this is related to 2.11.0). AudioChangeDeviceTest never stops beeping, even after closing the test, and opening it again makes it beep twice as much. The test list itself is over-sensitive to scrolling, and displays blank space at the bottom of the list instead of showing the last entries. Because the test list ScrollPane is essentially broken, I can't test anything after AudioChangeDeviceTest. Everything before that is fine, I guess.

Somewhat-related, this fixes a mysteriously missing stub method, `file()`, in GWT's emulated FileHandle class. It appears this was intended to be overridden in GwtFileHandle... but didn't actually exist to be overridden. With this change, the official libGDX has feature parity with the third-party backend Liftoff uses currently (and still will until there's a new stable release).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,5 @@
 [1.12.2-SNAPSHOT]
-- [BREAKING CHANGE] GWT: Updated to 2.10.0. `com.google.jsinterop:jsinterop-annotations:2.0.2:sources` must be added as a dependency to your html project dependencies.
+- [BREAKING CHANGE] GWT: Updated to 2.11.0. `com.google.jsinterop:jsinterop-annotations:2.0.2:sources` must be added as a dependency to your html project dependencies.
 - [BREAKING CHANGE] Android: Minimum API level is now level 19 (Android 4.4)
 - [BREAKING CHANGE] iOS: Increased min supported iOS version to 12.0. Update your Info.plist file if necessary.
 - [BREAKING CHANGE] Android, iOS: Exceptions occurring in Runnable tasks scheduled through Gdx.app.postRunnable() are no longer swallowed and will crash the app (add a protection if required).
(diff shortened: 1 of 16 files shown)


obigu (approved):
I've run the libGDX test suite and have found no issues. 

Considering GWT 2.11 has been already tested for a long time on Liftoff projects and changes are minimal on our side, I think it's safe to merge.

**On the test suite related issues:**

I can't reproduce the scrollpane issue which behaves as expected (tested on a Windows desktop with Google Chrome).

The `AudioChangeDeviceTest` doesn't stop beeping after being closed, I'll create a separate issue.

tommyettinger:
I'll test the LWJGL3 version of the tests; it may be an issue with my hardware scrolling too fast or too much, but it still shouldn't be able to scroll past the end of the list.

I just tested LWJGL3, and it behaves fine on my laptop. The GWT issue may be related to my browser (currently Vivaldi, I will probably switch to Brave if Vivaldi goes to Manifest v3). It may be related to the fact that the "scroll wheel" is a gesture on my touchpad. The same gesture works for LWJGL3, though. It might be that the GWT window is extremely scaled up compared to LWJGL3.

SimonIT (approved):
I tried it and everything seem to work.

I also have the scrolling issue:
libGDXGWTTestRunner-GoogleChrome2024-08-1813-53-22-ezgif com-optimize

But it is also on master so I don't think it's related to this update

Issue #7432, opened 2024-08-08
Support 'class' Tiled/Tmx MapProperty


silas-hw:
#### Issue details
LibGDX currently does not support the 'class' type for custom properties in Tiled maps (see [here](https://doc.mapeditor.org/en/stable/manual/custom-properties/#custom-types)). These are useful for having reusable sets of properties within Tiled. However, when trying to load a map with a class type property with `TmxMapLoader` the following error is produced:

```
Exception in thread "main" com.badlogic.gdx.utils.GdxRuntimeException: Wrong type given for property customProperty, given : class, supported : string, bool, int, float, color
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.castProperty(BaseTmxMapLoader.java:509)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadProperties(BaseTmxMapLoader.java:487)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:450)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObject(BaseTmxMapLoader.java:354)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadObjectGroup(BaseTmxMapLoader.java:287)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadLayer(BaseTmxMapLoader.java:211)
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadTiledMap(BaseTmxMapLoader.java:174)
	at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:81)
	at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:59)
	at com.mygdx.game.GameTest.create(GameTest.java:15)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.initializeListener(Lwjgl3Window.java:416)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:366)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:193)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:167)
[...]

PR #7437, opened 2024-08-11
Add support for 'class' type property of tmx files

silas-hw:
Solves #7432.

Adds a new condition in `BaseTmxMapLoader.loadProperties()` to check if the type of a property is 'class' and loads the property appropriately.

Tests are missing.


silas-hw:
Just noticed i forgot to close out the else if statement... my bad that's a really silly mistake

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -483,7 +483,18 @@ public void run () {
 						throw new GdxRuntimeException(
 							"Error parsing property [\" + name + \"] of type \"object\" with value: [" + value + "]", exception);
 					}
-				} else {
+				} else if (type != null && type.equals("class")) {
```
Quillraven, line 486: Small thing but you can simplify to "class".equals(type). No need for a null check.

PR #7438, opened 2024-08-11
Add support for 'class' type property in TMX files

silas-hw:
Solves #7432 

Adds a new condition in `BaseTmxMapLoader.loadProperties()` to check if the type of a *property* is 'class' and loads the property appropriately.

This isn't to be confused with the type/class of an *object*, for which there is already support for. Tiled allows specific properties of an object to be typed by a class, not just the object itself.

Tests are missing.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
@@ -494,7 +494,7 @@ public abstract class BaseTmxMapLoader<P extends BaseTmxMapLoader.Parameters> ex
 					loadProperties(classProperty, property.getChildByName("properties"));
 					properties.put(name, classProperty);
 
-				else {
+				} else {
 					Object castValue = castProperty(name, value, type);
 					properties.put(name, castValue);
 				}
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
@@ -483,7 +483,18 @@ public abstract class BaseTmxMapLoader<P extends BaseTmxMapLoader.Parameters> ex
 						throw new GdxRuntimeException(
 							"Error parsing property [\" + name + \"] of type \"object\" with value: [" + value + "]", exception);
 					}
-				} else {
+				} else if (type != null && type.equals("class")) {
+					// A 'class' property is a property which is itself a set of properties
+					MapProperties classProperty = new MapProperties();
+					String className = property.getAttribute("propertytype");
+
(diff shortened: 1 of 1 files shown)


lyze237:
There's already an open PR for that: #7007

Quillraven:
> There's already an open PR for that: #7007

I think your linked PR is not related to this issue.

Your linked PR is for the general object type in Tiled. One object can have one type. The "type" got changed to "class" at some point but I think that was reverted or can be changed in Tiled how you want to store it.


This PR however is related to object properties of type class. An object in Tiled can have multiple properties and one of the types is "class" which itself can reference other class properties.

So, imo, those two things are different and the separate PRs make sense.

lyze237:
Oh yeah my bad! PR seems good in that case 👍

Quillraven:
I personally would highly appreciate it because I have a custom loader right now just for this. Would be good if it works out of the box.

I try to find some time in the next two weeks to enhance this PR.

Regarding your comments:
- no need to support exported classes. This should be a separate PR but I guess it is different from game to game. I currently parse the project file to autogenerate classes in the core project but it is very specific to my game because those classes also get some utility functions auto created.
- I don't know if the type is that useful in the exported data. Since it anyway must be casted at runtime to the correct type. A user knows what kind of type he expects and I think the MapProperties class already supports this conversion/casting?
- I also wanted to have an extra loader Parameter to specify a factory method of a class to not create a generic MapProperties class and instead create a real class of the game. However, I think java.util.function is not allowed in LibGDX because it compiles against Java7. Therefore, it is impossible to do that and we can skip it.

BoBIsHere86:
> I personally would highly appreciate it because I have a custom loader right now just for this. Would be good if it works out of the box.
> 
> I try to find some time in the next two weeks to enhance this PR.
> 
> Regarding your comments:
> 
> * no need to support exported classes. This should be a separate PR but I guess it is different from game to game. I currently parse the project file to autogenerate classes in the core project but it is very specific to my game because those classes also get some utility functions auto created.
> * I don't know if the type is that useful in the exported data. Since it anyway must be casted at runtime to the correct type. A user knows what kind of type he expects and I think the MapProperties class already supports this conversion/casting?
> * I also wanted to have an extra loader Parameter to specify a factory method of a class to not create a generic MapProperties class and instead create a real class of the game. However, I think java.util.function is not allowed in LibGDX because it compiles against Java7. Therefore, it is impossible to do that and we can skip it.

These are all great points. I'm just the wrong person to look at this since I was unsure of the use cases for this and did not want to make any assumptions. I fully support adding every feature we can get in here. So feel free to a go at it. 👍

BoBIsHere86:
PR #7534 which was a replacement for this PR has now been merged.
Either @silas-hw or one of the admins can now close old PR out, so it doesn't add to any confusion.

PR #7443, opened 2024-08-18
Apply `transient` to Array's Iterable, Iterator

tommyettinger:
Various other iterators were made `transient` a while ago, but Array wasn't touched. `SortedIntList`, `Queue`, all `Set` types, and all `Map` types all use `transient` for their cached iterators, but for some reason `Array` doesn't. This matters when using default (typically reflection-based) serialization in something like [Fury](https://fury.apache.org/), or maybe also Kryo if it allows not writing a serializer. If the `ArrayIterable` and `PredicateIterable` are not marked `transient`, then they get written and form a cycle of references back to the original `Array`. Fury can handle this by tracking refs and by registering `ArrayIterable` and `ArrayIterator`, but this causes a significant increase to file size. Anything can handle this by registering a custom serializer, as well, but I think the cached variables that don't actually affect serialized state should not go in serialized data.

This also adds generics in places where they had been used inconsistently within `Array`, but it only touches iteration-related code.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -35,8 +35,8 @@ public class Array<T> implements Iterable<T> {
 	public int size;
 	public boolean ordered;
 
-	private ArrayIterable iterable;
-	private Predicate.PredicateIterable<T> predicateIterable;
+	private transient ArrayIterable<T> iterable;
+	private transient Predicate.PredicateIterable<T> predicateIterable;
 
 	/** Creates an ordered array with a capacity of 16. */
 	public Array () {
@@ -484,8 +484,8 @@ public class Array<T> implements Iterable<T> {
 	 * If {@link Collections#allocateIterators} is false, the same iterator instance is returned each time this method is called.
 	 * Use the {@link ArrayIterator} constructor for nested or multithreaded iteration. */
 	public ArrayIterator<T> iterator () {
-		if (Collections.allocateIterators) return new ArrayIterator(this, true);
-		if (iterable == null) iterable = new ArrayIterable(this);
+		if (Collections.allocateIterators) return new ArrayIterator<T>(this, true);
+		if (iterable == null) iterable = new ArrayIterable<T>(this);
 		return iterable.iterator();
 	}
 
@@ -667,7 +667,7 @@ public class Array<T> implements Iterable<T> {
 	static public class ArrayIterable<T> implements Iterable<T> {
 		private final Array<T> array;
 		private final boolean allowRemove;
-		private ArrayIterator iterator1, iterator2;
+		private transient ArrayIterator<T> iterator1, iterator2;
 
 // java.io.StringWriter lastAcquire = new java.io.StringWriter();
 
@@ -682,12 +682,12 @@ public class Array<T> implements Iterable<T> {
 
 		/** @see Collections#allocateIterators */
 		public ArrayIterator<T> iterator () {
-			if (Collections.allocateIterators) return new ArrayIterator(array, allowRemove);
+			if (Collections.allocateIterators) return new ArrayIterator<T>(array, allowRemove);
 // lastAcquire.getBuffer().setLength(0);
 // new Throwable().printStackTrace(new java.io.PrintWriter(lastAcquire));
 			if (iterator1 == null) {
-				iterator1 = new ArrayIterator(array, allowRemove);
-				iterator2 = new ArrayIterator(array, allowRemove);
+				iterator1 = new ArrayIterator<T>(array, allowRemove);
+				iterator2 = new ArrayIterator<T>(array, allowRemove);
 // iterator1.iterable = this;
 // iterator2.iterable = this;
 			}


NathanSweet:
I'm OK with marking the fields as transient. I would still write a custom serializer for core classes.

RHS generics (and the diamond operator) are rarely useful. They don't hurt, just adds a tiny bit of noise.

tommyettinger:
I didn't use the diamond operator. :) We still are in the stone age of Java 7 compatibility last I checked, and even that supports the diamond operator, but I know you have your issues with standard-practice generics in Java. I have to point out standard-practice, because the *warnings* that get issued by javac when omitting "rarely useful" generics are *more than a tiny bit of noise*.

idea64_8IYUwutpqk

How am I supposed to know which of these warnings are symptoms of a deeper problem? There's a lot of them, and this is just one inspection. The full list looks like this, without the few changes in this PR:

idea64_3vGLhIUXXD

And, with the changes back and applied:

idea64_Wp3td1ZDSJ

I didn't even change anything but the ArrayIterable/ArrayIterator generics, Xoppa already was (correctly) using generics in PredicateIterable code. I'd say 12 fewer warnings is generally a good thing.

I think almost everyone on the team tries to use generics in [the recommended way](https://docs.oracle.com/javase/tutorial/java/generics/why.html)...? Considering libGDX is used in university classes as part of teaching Java, I would think teaching Java that throws warnings left and right as part of accepted behavior would be odd, and not exactly conducive to finding a Java programming job when (or if) students graduate.

Tom-Ski (approved):

badlogic:
I'm OK with both transients and RHS diamond operator. Nate won't touch `Array.java` anyways :D

NathanSweet:
> warnings that get issued by javac when omitting "rarely useful" generics are more than a tiny bit of noise.

Those warnings are useless.

> How am I supposed to know which of these warnings are symptoms of a deeper problem?

My IDE has zero warnings for libgdx. I can see when something important needs fixing. We used to commit the IDE settings files so everyone has the same warnings setup. Now we generate the IDE files instead and that nicety was lost (while other niceties were gained).

> I think almost everyone on the team tries to use generics in [the recommended way](https://docs.oracle.com/javase/tutorial/java/generics/why.html)

It's not gospel to me. Their goal is to never be able to get a ClassCastException. That sounds nice, but needing a cast is relatively common and littering a codebase with suppress warning annotations, RHS generics, and diamond operators isn't worth it IMHO. That exception was never such a problem that it warrants the extra work needed to chase away every last warning, which is what you would need to do to achieve zero warnings. Better to turn them off, spend time on actually useful things, and enjoy cleaner code. Generics are great and useful in many cases, but taking it to the extreme is not. Students and jobs should understand this.

There are warnings that can be useful but not enabled by default, like synthetic access. Warnings should be configured at the project level and everyone should have zero warnings when they commit or pull.

> Nate won't touch Array.java anyways :D

Begone, demon! Back in your hole!

tommyettinger:
You're technically correct, there are (currently) no warnings, but only because the build fails due to Spotless errors.

```
12:06:10 AM: Executing 'build'...

Configuration 'spotless865455342' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless-641085672' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless-1429330901' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless865455342' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless-641085672' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
Configuration 'spotless-1429330901' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
Run with --info for a stacktrace.
> Task :assemble UP-TO-DATE
> Task :spotlessInternalRegisterDependencies
> Task :spotlessGroovyGradle
> Task :spotlessGroovyGradleCheck FAILED

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.4/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
3 actionable tasks: 3 executed

* What went wrong:
Execution failed for task ':spotlessGroovyGradleCheck'.
> The following files had format violations:
      build.gradle
          @@ -1,240 +1,240 @@
          -/*******************************************************************************\r\n
          - * Copyright 2011 See AUTHORS file.\r\n
          - *\r\n
          - * Licensed under the Apache License, Version 2.0 (the "License");\r\n
          - * you may not use this file except in compliance with the License.\r\n
          - * You may obtain a copy of the License at\r\n
          - *\r\n
          - *   http://www.apache.org/licenses/LICENSE-2.0\r\n
          - *\r\n
          - * Unless required by applicable law or agreed to in writing, software\r\n
          - * distributed under the License is distributed on an "AS IS" BASIS,\r\n
          - * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\r\n
          - * See the License for the specific language governing permissions and\r\n
          - * limitations under the License.\r\n
          - ******************************************************************************/\r\n
          -buildscript {\r\n
          -	apply from: "gradle/dependencies.gradle"\r\n
          -\r\n
          -	repositories {\r\n
          -		mavenCentral()\r\n
          -		gradlePluginPortal()\r\n
          -	}\r\n
          -	dependencies {\r\n
          -		classpath "com.diffplug.spotless:spotless-plugin-gradle:${versions.spotless}"\r\n
          -		classpath "com.badlogicgames.gdx:gdx-jnigen-gradle:2.5.2"\r\n
          -	}\r\n
          -}\r\n
          -\r\n
          -plugins {\r\n
          -	id "de.undercouch.download" version "5.0.1"\r\n
          -}\r\n
          -\r\n
          -apply from: "gradle/dependencies.gradle"\r\n
          -\r\n
          -ext {\r\n
          -	isReleaseBuild = {\r\n
          -		return project.hasProperty("RELEASE")\r\n
          -	}\r\n
          -\r\n
          -	getReleaseRepositoryUrl = {\r\n
          -		return project.hasProperty('RELEASE_REPOSITORY_URL') ? RELEASE_REPOSITORY_URL\r\n
          -				: "https://oss.sonatype.org/service/local/staging/deploy/maven2/"\r\n
          -	}\r\n
          -\r\n
          -	getSnapshotRepositoryUrl = {\r\n
          -		return project.hasProperty('SNAPSHOT_REPOSITORY_URL') ? SNAPSHOT_REPOSITORY_URL\r\n
          -				: "https://oss.sonatype.org/content/repositories/snapshots/"\r\n
          -	}\r\n
      ... (432 more lines that didn't fit)
  Run 'gradlew.bat :spotlessApply' to fix these violations.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 4s
12:06:15 AM: Execution finished 'build'.
```

After fixing those Spotless errors, the build of the most recent commit actually still fails -- and gets 119 warnings, mostly in documentation, to boot. To recap, the steps I took were to update to the current master (with no changes of my own), run `gradlew spotlessApply`, and run `gradlew build`. The large, but still incomplete, output is attached. The build stopped early because there's some broken Gradle configuration in the Android tests. So, yeah. You're only getting zero warnings because you aren't even building all of libGDX, or you aren't building it with Gradle.

[build_log.txt](https://github.com/user-attachments/files/16777874/build_log.txt)

If the automated builds aren't able to build libGDX (and right now even my whole-project builds that aren't automated aren't completing), that isn't just a warning, that's a legitimate problem.

The correct way to handle which warnings are officially ignored would be to use the `@SuppressWarnings` annotation that's part of the JDK. There are other tools available today to configure editors using stable, relatively-cross-platform plain-text files. I don't know the extent of what `.editorconfig` is able to do, but we aren't even using `.gitattributes` effectively to enforce line endings, last I checked.

Issue #6962, opened 2022-08-12
No way to retrieve the correct key letter from a keycode on non-QWERTY keyboards


Hangman:
At the moment, there's no way to get the correct String representation of a keycode, unless you're using a standard (US?) QWERTY keyboard. This is mainly an issue for displaying key bindings. The only exception is as far as I know the lwjgl2 backend where keycodes refer to logical keys instead of the physical position (correct me if I'm wrong and I will edit this).

### Example
Press Q on a QWERTY keyboard => `Input.Keys.toString(keycode)` will return "Q".
Press Q on a AZERTY keyboard => `Input.Keys.toString(keycode)` will return "A".

### Potential fix
Please read #6957 for a discussion about how this could be fixed.

mrlem:
Stumbled on this issue as well: this is actually a regression originating from the LWJGL3 backend: it used to work fine with the LWJGL2 backend. I guess that's linked to the behavior LWJGL3 has for glfwSetKeyCallback:

> The key functions deal with physical keys, with layout independent key tokens named after their values in the standard US keyboard layout

I get that for the specific case where you map WSAD keys for directions this is a feature, as it enforces the physical layout over the keymap. But for many very common situations, this breaks existing code:

Say you have an **A**ttack action you want to bind to the "A" key:

```kotlin
if (Gdx.input.isKeyJustPressed(Keys.A)) {
    /* do something */
}
```

is now broken depending on the keymap the users computer has. Or it's the new way it works, but then the doc could be clearer, as it's very misleading.

I wonder what's the suggested approach, now. The only one I found was to revert to an ```InputProcessor``` and override ```keyTyped(character: Char): Boolean```.

PR #7446, opened 2024-08-22
Localized key names for non-US keyboards

MonstrousSoftware:
Linked to issue #6962 and #6957 and recent discussions in the Discord help forum.

Since LWJGL3 the key codes (Input.Keys.*) always follow the US keyboard layout.  This is not accidental, but a deliberate choice by the underlying GLFW library. However, for presentation to the user of key names (e.g. in a tutorial or a key binding screen), it would be good to follow the user's regional keyboard setting. E.g. if the user has an AZERTY keyboard, the "WASD" keys map to "ZQSD".

Following the discussion in issue #6957 the conclusion was that Input.Keys.toString should be region aware, i.e. present the key name as the user sees it on their keyboard.

This only applies to the LWJGL3 backend. The web backend (GWT and TeaVM) use key codes that are localized.

The solution is to use ```glfwGetKeyName(int key, int scancode)``` in GLFW to provide the name of a printable key.  This also required a mapping function ``getGLFWKeyCode`` to reverse the mapping from gdx key code back to GLFW code.  For non-printable keys, this original switch of Input.Keys.toString is maintained. The Input class was modified to allow alternative key names, the DefaultLwjgl3Input class was modified to introduce a key mapping for printable keys.

Tested on Windows with LWJGL3 with a UK and German keyboard.

### DIFF
--- a/gdx/src/com/badlogic/gdx/Input.java
+++ b/gdx/src/com/badlogic/gdx/Input.java
@@ -273,11 +273,21 @@ public interface Input {
 
 		public static final int MAX_KEYCODE = 255;
 
+		public static class KeyNameMapper {
+			public String getKeyName (int keycode) {
+				return null;
+			}
+		}
+
+		public static KeyNameMapper keyNameMapper = new KeyNameMapper();
+
 		/** @return a human readable representation of the keycode. The returned value can be used in
 		 *         {@link Input.Keys#valueOf(String)} */
 		public static String toString (int keycode) {
 			if (keycode < 0) throw new IllegalArgumentException("keycode cannot be negative, keycode: " + keycode);
 			if (keycode > MAX_KEYCODE) throw new IllegalArgumentException("keycode cannot be greater than 255, keycode: " + keycode);
+			String name = keyNameMapper.getKeyName(keycode);
+			if (name != null) return name;
 			switch (keycode) {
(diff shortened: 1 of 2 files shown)


tommyettinger:
For some reason, Spotless is acting strange about files that weren't even changed here. Still, you can try what it says in the build failure message:
```
  Run './gradlew :backends:gdx-backend-lwjgl3:spotlessApply' to fix these violations.
```

gdx/src/com/badlogic/gdx/Input.java
```diff
@@ -633,7 +645,7 @@ public static int valueOf (String keyname) {
 		/** lazily intialized in {@link Keys#valueOf(String)} */
 		private static void initializeKeyNames () {
 			keyNames = new ObjectIntMap<String>();
-			for (int i = 0; i < 256; i++) {
+			for (int i = 255; i >= 0; i--) {
```
Tom-Ski, line 648: Whats the reason for this change?

MonstrousSoftware:
This is so that valueOf(“\”) will return the original BACKSLASH code value
instead of the new code which is also mapped to “\” (because that is what
is usually shown on that key). I think this is the first case where two key
codes have the same string representation and the reasoning is that the
lower value code will usually be the more stable one and higher value codes
will be more esoteric keys.

PR #7447, opened 2024-08-27
Add setter equivalent to Color constants: 

StartsMercury:
`Color.ORANGE` is to `mycolor.setToOrange()`

## Naming

The naming may be adjusted now, it's the only good initial format I thought it could have: `setTo<ColorInCamelCase>`.

### Adding, Updating, or Removing Color Definitions: missing the other equivalent

This could be mostly addressed by having constant definitions depend on their method counterparts: `Color ORANGE = new Color(...)` turns into `Color ORANGE = new Color().setToOrange()`. Though may not happen often, it may be more likely to miss a constant than method with this solution.

## Formatting

* Initially, the method and its body are formatted on one line


Frosty-J:
I'm not clear on the advantage of adding all these new methods over using `myColor.set(Color.ORANGE)`.

StartsMercury:
Oh, I forgot to mention the *why* I made this!

I'll add it later. So, the methods should be

Frosty-J:
All these assertions pass for me (with assertions enabled - verified using `assert false`).

StartsMercury:
(moved from last comment edit)

Okay. `setToOrange` prevents the possibility of `Color.ORANGE` being mutated somewhere else.

(addition)

Yeah, describing the scenario instead of adding it to my example scenario wasn't helpful at all I notced. Woops! I'll go add `c1.set(1.0, 2.0, 3.0, 4.0)`.

PokeMMO:
I'd love to see a proper encapsulation PR. Are method calls even expensive still?

StartsMercury:
I am not sure if method calls are not as expensive on the systems where they are. Though I am unsure if these kind of calls (`set(Color.RED)` and `setToRed()`) are always done each frame (probably not?).

PokeMMO:
Sorry for not being clear, I meant encapsulation method calls. For example making `r` private and require using `getR()` and `setR()` instead of directly changing the Color.

Doing so would allow us to create immutable colors.

Frosty-J:
I believe getters and setters will get inlined? During runtime, or possibly even earlier if ProGuard optimisations are enabled. But I can't speak with certainty, especially when it comes to iOS.

Frosty-J:
It's up to you, but `//@off` and `//@on` can be used to disable Spotless checks for a section of code (i.e. these methods that used to take a quarter of the lines that they do now) then re-enable them after.

tommyettinger:
All existing code that uses the r, g, b, and a properties would break if we forced those to change to getters. So much would break, inside libGDX and outside it, including code that wasn't mutating any Color instances, and so wasn't wrong. There's no automated way to fix it when people update. We would absolutely need to deprecate for at least a full minor version to give people a chance to prepare.

What I think is the better way, if you do decide to break everything, would be to use int or float primitives to represent RGBA colors, since those can be made final and thus fully immutable. You would use static methods on a class like Color to access the red, green, blue, or alpha segment of an int or float primitive representing an RGBA color. This is what I have been doing in [colorful-gdx](https://github.com/tommyettinger/colorful-gdx/blob/master/colorful/src/main/java/com/github/tommyettinger/colorful/rgb/Palette.java) and [SquidSquad](https://github.com/yellowstonegames/SquidSquad/blob/main/squidcore/src/main/java/com/github/yellowstonegames/core/DescriptiveColorRgb.java) for years now.

mgsx-dev:
Maybe there is another solution that minimize breaking. Something like that : 

```java
public class Color {
    public static class ConstantColor {
        public final float r,g,b,a;
        public ConstantColor(float r, float g, float b, float a){ ... }
    }
    public static final ConstantColor RED = new ConstantColor(1,0,0,1);
    ...
    public Color(ConstantColor color){ /* simply copy rgba fields */ }
    public void set(ConstantColor color){ /* simply copy rgba fields */ }
}
```

With this solution, a common code adaptation required would be things like : `Color color = Color.RED` which is already prone of issues. And the fix is easy : `Color color = new Color(Color.RED)`.

We could duplicate other Color methods like `add, sub, lerp, equals, mul` but i'm not sure it worth it.

For calls like : `spriteBatch.setColor, ScreenUtils.clear, shapeRenderer.setColor, and maybe few others` we could add methods supporting ContantColor. I don't think there are a lot (i didn't deeply checked).

Note that i found this issue in libgdx code recently : https://github.com/libgdx/libgdx/blob/master/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java#L759
and this one : https://github.com/libgdx/libgdx/blob/054b25b8ced6fa54ba0c8d959b6a08a8d54de9d9/gdx/src/com/badlogic/gdx/graphics/g2d/RepeatablePolygonSprite.java#L45

I'm not sure @PokeMMO solution (setters) is good : my concern is what setters do for these immutable colors ? 
* Do nothing ? that would be hard to spot why a color doesn't change when you change it.
* Throw exception ? You'd need good test coverage to spot them before production and spend time to find the source of the issue.

What do you think guys?

Frosty-J:
Subclass seems good. I'm not too concerned with your examples from the libGDX codebase since changing `FreeTypeFontParameter`'s `color` is a rare thing to do (tinting a font is effectively free; generating a new one is slow on mobile especially) and in `RepeatablePolygonSprite` it is private with a setter but no getter, though it wouldn't kill us to err on the side of caution and `new Color()` the FontParam.

Setters could log a non-fatal error with stacktrace if immutable. But to be honest I'm not that keen, the more I think about it. Oversimplistic example without delta, clamping, or `Interpolation` class: `col.a += .1f` would become `col.setA(col.getA() + .1f)`?

PokeMMO:
The more "correct" way to do immutability is to make all Colors immutable and return a copy for the `set` methods.

The middle ground fix, retaining some performance, but requiring an api change is `setX()` which modifies and returns `this` for mutable, and a copy for immutable.

Both of those are very breaking, and would require everyone to change the way they use Colors.

The yolo performance focused way, is to throw an exception/log a warning when trying to modify an immutable color, as @Frosty-J mentioned. This is what I would suggest, as if you do this, you're *already* shooting yourself in the foot. I'm sure a lot of us have accidentally modified a static Color and then had to figure out why something was inexplicably broken.

mgsx-dev:
@PokeMMO don't take it badly but i'm not convinced by the encapsulation solution. Any logging/throw exception/do nothing/copy behaviors are not good imho. 

To clarify my concerns above : Depending on game code complexity of branching and possible states you could have some rare cases where a Color object arrive into some mutating code. Could happen in dev (best case scenario), could happen in production (worse case scenario). Making game crashing for that reason would be worse than current state. (i guess some lucky games can live with the mistakes in their code).

The auto copy approach is maybe the better in the 4 possible behavior with still possible mistakes of generating garbage : eg.
```java
Color myColor = Color.WHITE; // at initialisation time
myColor.setR(time % 1f) // at every frames.
```

I prefer a solution that prevent user to make the mistake at compilation time like the one i suggested.

Also, in term of code readability, public fields are way better. As you noticed we'd need to do that for Vector2 and Vector3 as well.
(TBH if we make these fields private someday, i'll surely monkey patch these classes to keep the readability and avoid huge code migration in my stuff).

I'll stop here and read other team members thoughts on all that.

PR #7448, opened 2024-08-28
Color encapsulation

PokeMMO:
I took the 5 minutes to click the "encapsulate fields" button in idea and add new `ImmutableColor` class.

Currently there exists a major footgun, which is accidental mutation of the static `Color` fields such as `Color.WHITE`. Using encapsulation and a new `ImmutableColor` class which extends `Color` we can prevent this issue.

We have at least two ways to handle what should happen when an attempt to mutate an `ImmutableColor`.
1. Throw a `RuntimeException`. This is my preferred option. If you're accidentally doing this, you're going to run into all sorts of inconsistencies and problems until you track down where you did this.
2. Return a modified copy of the color. This would help for certain cases where chaining is used, but this would be far from a complete fix. See `ImmutableColorCpy`
3. Log a warning.


mgsx-dev:
see my comment here https://github.com/libgdx/libgdx/pull/7447#issuecomment-2315419923 (sorry about cross posting)

PokeMMO:
> see my comment here [#7447 (comment)](https://github.com/libgdx/libgdx/pull/7447#issuecomment-2315419923) (sorry about cross posting)

I don't hate your solution either, but there are a lot of places that accept `Color` as an argument that would need a new overload or garbage created.

gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -24,54 +24,60 @@
...
+	public static final Color MAROON = new ImmutableColor(0xb03060ff);
+
+	/** the red component **/
+	@Deprecated public float r;
+	/** the green component **/
+	@Deprecated public float g;
+	/** the blue component **/
+	@Deprecated public float b;
+	/** the alpha component **/
+	@Deprecated public float a;
```
tommyettinger, line 80: If a field is marked `@Deprecated`, it should also have `@deprecated` in its Javadocs, ideally with a replacement or reason why it is deprecated. From the documentation for the annotation:

```java
/**
...
 * It is strongly recommended that the reason for deprecating a program element
 * be explained in the documentation, using the {@code @deprecated}
 * javadoc tag. The documentation should also suggest and link to a
 * recommended replacement API, if applicable. A replacement API often
 * has subtly different semantics, so such issues should be discussed as
 * well.
 *
 * <p>It is recommended that a {@code since} value be provided with all newly
 * annotated program elements. Note that {@code since} cannot be mandatory,
 * as there are many existing annotations that lack this element value.
 *
 * <p>There is no defined order among annotation elements. As a matter of
 * style, the {@code since} element should be placed first.
 *
 * <p>The {@code @Deprecated} annotation should always be present if
 * the {@code @deprecated} javadoc tag is present, and vice-versa.
 */
```

gdx/src/com/badlogic/gdx/graphics/ImmutableColor.java
```diff
@@ -0,0 +1,124 @@
+
+package com.badlogic.gdx.graphics;
+
+/** A color class that prevents mutation to r/g/b/a values. */
+public class ImmutableColor extends Color {
```
tommyettinger, line 5: ImmutableColor should probably be `final` to ensure there isn't a mutable subclass created.

gdx/src/com/badlogic/gdx/graphics/ImmutableColorCpy.java
```diff
@@ -0,0 +1,124 @@
+
+package com.badlogic.gdx.graphics;
+
+/** A color class that prevents mutation to r/g/b/a values. */
+public class ImmutableColorCpy extends Color {
```
tommyettinger, line 5: The docs here should at least mention that this creates mutable copies of itself (with the `Color` class) when modified.

gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -358,10 +401,10 @@ public static Color valueOf (String hex) {
 	 * @see #toString() */
 	public static Color valueOf (String hex, Color color) {
 		hex = hex.charAt(0) == '#' ? hex.substring(1) : hex;
-		color.r = Integer.parseInt(hex.substring(0, 2), 16) / 255f;
-		color.g = Integer.parseInt(hex.substring(2, 4), 16) / 255f;
-		color.b = Integer.parseInt(hex.substring(4, 6), 16) / 255f;
-		color.a = hex.length() != 8 ? 1 : Integer.parseInt(hex.substring(6, 8), 16) / 255f;
+		color.setR(Integer.parseInt(hex.substring(0, 2), 16) / 255f);
```
tommyettinger, line 404: This and all Color-modifying static methods after this will be broken if given an `ImmutableColorCpy`, since the given object will create a modified copy and discard it once per channel. Then it will return the original, unmodified `ImmutableColorCpy`.

gdx/src/com/badlogic/gdx/graphics/Color.java
```diff
@@ -24,54 +24,60 @@
  * 
  * @author mzechner */
 public class Color {
-	public static final Color WHITE = new Color(1, 1, 1, 1);
-	public static final Color LIGHT_GRAY = new Color(0xbfbfbfff);
-	public static final Color GRAY = new Color(0x7f7f7fff);
-	public static final Color DARK_GRAY = new Color(0x3f3f3fff);
-	public static final Color BLACK = new Color(0, 0, 0, 1);
+	public static final Color WHITE = new ImmutableColor(1, 1, 1, 1);
```
tommyettinger, line 27: All Color constants should have the class `ImmutableColor` or be documented as immutable, otherwise users won't have any indication that they throw `Exception`s when modified.

tommyettinger (changes requested):
I don't like this one bit. Almost every mid-sized project updating libGDX would have to expect a wall of deprecation messages, none of which would be any different for code that modifies Color constants than from code which modifies Colors by the formerly-default field modification. I'm pretty sure the libGDX plugin for IDEA (by BlueBoxWare) warns on modifications to color constants currently, so this isn't really beneficial for anyone using that.

Personally, if I had a mid-size game I was developing, I just wouldn't update libGDX if it meant dealing with this. If I felt like drastic changes, I'd switch to Godot, which would mean a complete rewrite but with most of my tools written for me already.

Frosty-J:
> I'm pretty sure the libGDX plugin for IDEA (by BlueBoxWare) warns on modifications to color constants currently, so this isn't really beneficial for anyone using that.

It doesn't. I have it and it didn't emit any warnings during the testing at #7447. Even if I do something as obviously wrong as `Color.RED.set(Color.GREEN)`, it doesn't warn.

> Personally, if I had a mid-size game I was developing, I just wouldn't update libGDX if it meant dealing with this.

In how many places are you tinkering with colours? I would not expect the migration process to take more than a few minutes. Though it may be worth trying to write a Find & Replace regex if this were to go through.

Otherwise, I mostly agree with your review.

tommyettinger:
Interesting, I thought I remembered seeing a warning about that at some point. That would be a perfect place for that plugin to warn.

I'm curious what this regex would look like, considering it would need to handle changes to r, g, b, and a, as well as calls to anything that modifies a Color, potentially as a parameter. I'm guessing it would actually be quite a few regexes.

PokeMMO:
> Almost every mid-sized project updating libGDX would have to expect a wall of deprecation messages

The idea would be to protect these fields after a major update. Existing code bases should migrate to the getters and setters.

> Missing javadocs/missing final/etc

This PR is very rough to see if its even feasible to get this change in.

> ImmutableColorCpy

This is an alternative implementation, we should ideally pick one or the other.

gdx/src/com/badlogic/gdx/graphics/Color.java
PokeMMO: This is true, and its one reason why I prefer we don't do the `ImmutableColorCpy` route, and instead go the exception route.

angelickite:
I too have accidentaly changed the global `Color`-instances which led to a few moments of confusion. 
Nothing dramatic, but if (cheap) improvements are possible, why not save some developer hours.

If I were to attempt a change in this vein, I would be as explicit about is as possible and not beat around the bush:

- `ImmutableColor extends Color`: just no. that is barely a bandaid, which you can see through its ripple effect like having to resort to execution time exceptions.

- Deprecating the totally valid mutable r,g,b,a fields for the alternative of having to go through getters/setters is a big step backwards. It is rare to find code these days which treats the developer with some respect, this is not the way to go.

The core issue mainly stems from having a single "API" which covers multiple needs, like
  - being the reference values of the libGDX internal colors
  - bridging the gap between the libGDX API and user code (font markup)
  - providing a sort of default palette for convenience

Looking at it from the side, I would say that the Colors defined in `Color` are mainly about libGDX internals. Using these as a palette beyond quick debugging is just not recommendable, e.g. RED(255, 0, 0) rarely is good on the eyes.

I would either

- position these as such via documentation and offer access to a `GdxColorPalette` which can have all the immutable properties one desires, for example by only generating new Colors via static functions like `Color red = GdxColorPalette.red()`.
or
- break these values via renaming, removal or what else one desires. Then make the separation clear and let people discover their potential bugs.

IMO that would cover most needs and not beat around the bush.

There is at least one place in the code base that I am aware of (had some ouch with it...) where libGDX kind of runs into this situation itself: https://github.com/libgdx/libgdx/blob/f633eba7d013459033e489a488ed962e321942f1/extensions/gdx-freetype/src/com/badlogic/gdx/graphics/g2d/freetype/FreeTypeFontGenerator.java#L759 (I would expect a `Color.WHITE.cpy()` here)

Whether the breakage is worth it depends on the size and complexity of the usages in actual projects. I personally wouldn't mind the breakage in my own projects for the health of the framework in the long term, but I wouldn't argue against those saying its just not worth it either (given some clarification to the situation via the several documentation options). Just fiddling with the way Colors are structurally represented is not the way to go.

PR #7452, opened 2024-08-31
Introduce Labeler for Pull Request labeling

SimonIT:
We get a little lazy sometimes with the labels and the PRs, so I think this might come in handy.

I have no idea if this will actually work. So any input is appreciated

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,97 @@
+3D:
+- changed-files:
+  - any-glob-to-any-file:
+    - /**/g3d/**
+android:
+- changed-files:
+  - any-glob-to-any-file:
+    - /**/gdx-backend-android/**
+    - /**/gdx-tests-android/**
+audio:
+- changed-files:
+  - any-glob-to-any-file:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,97 @@
+3D:
+- changed-files:
+  - any-glob-to-any-file:
+    - /**/g3d/**
+android:
+- changed-files:
+  - any-glob-to-any-file:
+    - /**/gdx-backend-android/**
+    - /**/gdx-tests-android/**
```
Frosty-J, line 9: This would label changes to https://github.com/libgdx/libgdx/tree/master/tests/gdx-tests-android/assets/data as Android even if the pull requester hasn't touched Android? Since all test assets are in there. And possibly chaos if someone touches a large number of files. I like the idea.

SimonIT: Hmm, good point. We could explicitly exclude this path (if I'm not mistaken): `!/**/gdx-tests-android/assets/data/**`

Tom-Ski (approved):

PR #7453, opened 2024-09-02
Improve TexturePacker MaxRectsPacker pruning

lucas-viva:
The `TexturePacker` `MaxRectsPacker.MaxRects.pruneFreeList` method is O(n^3) in time, `n` being the number of `freeRectangles`.

`freeRectangles` size increases if you have many sprites with different sizes.

This PR changes the `pruneFreeList` method to use a O(n) approach.

Impirical tests, using 3000 random images generated by the code below showed the packer ran in around 60 seconds for the old method and in around 10 seconds for the new method.

```
  public static void generateRandomImages() throws Exception {
    Random random = new Random();
    for (int i = 0; i < 3000; i++) {
      BufferedImage image =
          new BufferedImage(
              32 + random.nextInt(32), 32 + random.nextInt(32), BufferedImage.TYPE_INT_ARGB);
      Graphics2D g2d = image.createGraphics();
      g2d.setColor(new Color(random.nextInt(256), random.nextInt(256), random.nextInt(256)));
      g2d.fillRect(0, 0, random.nextInt(32), random.nextInt(32));
      ImageIO.write(image, "png", new File("test/test" + i + ".png"));
    }
  }
```

For our game, it improved from 35~ minutes to 3~ minutes.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
@@ -25,6 +25,7 @@ import com.badlogic.gdx.tools.texturepacker.TexturePacker.ProgressListener;
 import com.badlogic.gdx.tools.texturepacker.TexturePacker.Rect;
 import com.badlogic.gdx.tools.texturepacker.TexturePacker.Settings;
 import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.IntSet;
 import com.badlogic.gdx.utils.Sort;
 
 /** Packs pages of images using the maximal rectangles bin packing algorithm by Jukka Jylänki. A brute force binary search is used
@@ -298,8 +299,9 @@ public class MaxRectsPacker implements Packer {
 	 * @author Nathan Sweet */
 	class MaxRects {
 		private int binWidth, binHeight;
-		private final Array<Rect> usedRectangles = new Array();
-		private final Array<Rect> freeRectangles = new Array();
+		private final Array<Rect> usedRectangles = new Array<>();
+		private final Array<Rect> freeRectangles = new Array<>();
+		private final Array<Rect> rectanglesToCheckWhenPruning = new Array<>();
 
 		public void init (int width, int height) {
 			binWidth = width;
@@ -696,6 +698,7 @@ public class MaxRectsPacker implements Packer {
 					Rect newNode = new Rect(freeNode);
 					newNode.height = usedNode.y - newNode.y;
 					freeRectangles.add(newNode);
+					rectanglesToCheckWhenPruning.add(newNode);
 				}
 
 				// New node at the bottom side of the used node.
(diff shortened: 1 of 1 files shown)


extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
```diff
@@ -129,7 +130,7 @@ private Page packPage (Array<Rect> inputRects) {
 					throw new RuntimeException("Image does not fit within max page width " + settings.maxWidth + paddingMessage + ": "
 						+ rect.name + " " + width + "x" + height);
 				}
-				if (height > maxHeight && (!settings.rotation || width > maxHeight)) {
+				if (height > maxHeight && width > maxHeight) {
```
tommyettinger, line 133: From what I can tell, you don't need to compare width and maxHeight unless rotation is true. I've never used rotation in texture packing, so  I don't know exactly how it works, but I suspect this line should be reverted unless there's some different behavior in this PR regarding pages with different height and width.

tommyettinger (changes requested):
This almost entirely looks good to me. I suspect the rotation code wasn't tested, because I don't even know how to use rotated `AtlasRegion`s, but I think the old code on the one line I mentioned was probably correct. If we can verify that the check between width and maxHeight is needed when rotation isn't enabled, then the new code can be kept.

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
lucas-kakele: Hey Tommy.

In line 120, there is a check for `settings.rotation` being true already.

Thus, if we are in line 132, it means `!settings.rotation` is always true.

Currently, the actual check is `height > maxHeight` because `(!settings.rotation || width > maxHeight)` is always true.

Should the check after `&&` be removed altogether then?

tommyettinger: I didn't see that part in the GitHub preview; it should be fine how you have it then. Approved!

tommyettinger (approved):

fgnm:
I don't use rotation too so can't test that part, but performances increases anyway. I have a test project for packing, time to repack all assets decreased from 88sec to 41sec, pretty nice!

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
```diff
@@ -129,7 +130,7 @@ private Page packPage (Array<Rect> inputRects) {
 					throw new RuntimeException("Image does not fit within max page width " + settings.maxWidth + paddingMessage + ": "
 						+ rect.name + " " + width + "x" + height);
 				}
-				if (height > maxHeight && (!settings.rotation || width > maxHeight)) {
```
NathanSweet, line 132: @lucas-viva Is there a reason to remove these checks?

Frosty-J: @NathanSweet This was explained in https://github.com/libgdx/libgdx/pull/7453#discussion_r1740999019

PR #7455, opened 2024-09-03
Fix when GLFW_LIBRARY_NAME.get() is null.

tommyettinger:
This is a simple fix for the line:

```java
if (!Configuration.GLFW_LIBRARY_NAME.get().equals("glfw_async")) {
```

When `Configuration.GLFW_LIBRARY_NAME.get()` returns `null`, this throws an NPE. Flipping the `.equals()` means the LHS will never be null, making it valid when `GLFW_LIBRARY_NAME` is not set.

This also runs Spotless.


TCreutzenberg:
(Thank you!)

SimonIT (approved):

PR #7457, opened 2024-09-07
Update actions

SimonIT:


tommyettinger (approved):
Looks good as far as I can tell. The GH Actions builds seem to be failing currently without the updated versions. I think this would be a good idea to at least investigate further, if we don't just merge this as-is.

.github/workflows/build-publish.yml
```diff
@@ -16,19 +16,19 @@ jobs:
   natives-ios:
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
Frosty-J, line 19: Berstanio was against this when I mentioned failing builds a couple months ago.

> No, v4 requires node 20, while v3 is fine with 16

But whether that's still the case and if we need old versions for anything, I do not know.

Tom-Ski:
Reverted this as breaks linux build. I think we can just update the ones failing? and leave others not updated?

PR #7462, opened 2024-09-21
Flame 3D particle editor UX improvement on bad user manipulations

WinterAlexander:
# Issue

As a new user to LibGDX's 3D environment and particles, I found myself struggling with the flame editor. When trying to load an atlas, I saw it didn't work but the software provided no feedback. So my first reflex was to run the flame editor with java in command line to get output and this is what I saw:

image

This was very confusing to me. 

# Solution

The goal of this PR is to enhance the UX of when trying to load an invalid file as atlas or texture. The most obvious mistake (which was mine) I think is to provide the PNG file instead of the ATLAS file for an atlas, so I made the atlas error message reflect that. There's also an error message when adding a bad texture file. This way we provide feedback to the user in case the operation they requested didn't work.

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/TextureLoaderPanel.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/TextureLoaderPanel.java
@@ -1,16 +1,23 @@
 
 package com.badlogic.gdx.tools.flame;
 
+import java.awt.GridBagConstraints;
+import java.awt.Insets;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.io.File;
+
+import javax.swing.DefaultComboBoxModel;
+import javax.swing.JButton;
+import javax.swing.JCheckBox;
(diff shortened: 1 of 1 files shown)


extensions/gdx-tools/src/com/badlogic/gdx/tools/flame/TextureLoaderPanel.java
```diff
@@ -56,6 +57,9 @@ public void actionPerformed (ActionEvent e) {
 					TextureAtlas atlas = editor.load(file.getAbsolutePath(), TextureAtlas.class, null, null);
 					if (atlas != null) {
 						editor.setAtlas(atlas);
+					} else {
+						JOptionPane.showMessageDialog(editor,
+							"Error loading atlas file. Make sure you must provide an atlas file, not an image file.");
```
tommyettinger, line 62: Nice feedback is always good. I haven't used Flame, but I like this.

tommyettinger (approved):
Looks like a solid small simple improvement. Approved.

PR #7467, opened 2024-09-28
Change visibility of PolygonSpriteBatch.switchTexture() to protected

TCreutzenberg:
Change visibility of PolygonSpriteBatch.switchTexture() to protected

Use case: On texture switch pass the new texture size to a shader

(SpriteBatch.switchTexture is already protected)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSpriteBatch.java
@@ -1310,7 +1310,7 @@ public class PolygonSpriteBatch implements PolygonBatch {
 		}
 	}
 
-	private void switchTexture (Texture texture) {
+	protected void switchTexture (Texture texture) {
 		flush();
 		lastTexture = texture;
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/PolygonSpriteBatch.java
```diff
@@ -1310,7 +1310,7 @@ protected void setupMatrices () {
 		}
 	}
 
-	private void switchTexture (Texture texture) {
+	protected void switchTexture (Texture texture) {
```
tommyettinger, line 1313: If this method is already protected in SpriteBatch, it makes sense to have the APIs be as similar as possible. Approved!

tommyettinger (approved):
Looks small and solid, and seems to pass Spotless without issue. Thanks for updating the changelog, too!

crykn:
@TCreutzenberg could you move the CHANGELOG entry to 1.13.1 please, then we can go ahead and merge this.

TCreutzenberg:
@crykn Done.

PR #7471, opened 2024-10-03
Fix floating point inaccuracy with touch position on iOS

Berstanio:
`loc.getX/Y()` returns a double, which currently gets eagerly casted to an `int`. The leads to loss of accuracy if `HdpiMode.Pixels` is enabled and the `int` location is multiplied by the `pixelsPerPoint` float again.

This pr fixes this by first multiplying the floating point numbers and than casting to an Integer.


tommyettinger (approved):
Math looks solid, though I can't meaningfully comment regarding anything about HDPI screens. If anyone needs to be convinced that the math is right, imagine a hypothetical screen, 1000x1000 physical pixels but only 10x10 logical pixels. Inputs to physical pixels 100,100 to 199,199 would (before this PR) all map to logical pixel `app.pixelsPerPoint,app.pixelsPerPoint,` never anything between 0 and that. That's because the cast to int forces locX and locY to 1,1, disregarding the smaller adjustments for values larger than 100,100. With this PR, the I-believe-intended behavior of the double getX/getY shines through, and all logical pixel values that can occur will be able to happen.

obigu (approved):

PR #7474, opened 2024-10-06
Implement new DefaultPool based on a supplier

Berstanio:
The current default pool is the `ReflectionPool`, which as the name suggests, uses reflection to create objects - specifically a no-arg constructor.
  
Because of proguard/r8 and other platforms with restrictions on reflection, it is in my opinion better to have a static alternative. It also allows to catch issues better, if e.g. no no-arg constructor is available.
  
With lambdas and functional interfaces, we in my opinion have a proper replacement.
  
I propose a `DefaultPool` which takes as an argument a simple supplier for the object. If a no-arg constructor (and java 8 syntax) is available , a new interface can be cleanly created with `new DefaultPool(MyClass::new)`.
The supplier is also flexible and can be used for any object creation method, even if no usable constructor is available.

The changes are pretty straight forward.


gdx/src/com/badlogic/gdx/utils/ReflectionPool.java
```diff
@@ -22,7 +22,12 @@
 
 /** Pool that creates new instances of a type using reflection. The type must have a zero argument constructor.
  * {@link Constructor#setAccessible(boolean)} will be used if the class and/or constructor is not visible.
+ * <p>
+ *
+ * The class is deprecated, please use {@link DefaultPool} instead.
```
tommyettinger, line 27: This is what the `deprecated` doclet is for, which is like `param`, `return` or `author` and has an at-sign before it. (I don't want to accidentally summon a user named author or something.)

gdx/src/com/badlogic/gdx/utils/DefaultPool.java
```diff
@@ -0,0 +1,30 @@
...
+		super(initialCapacity, max);
+		this.poolTypeSupplier = supplier;
+	}
+
+	@Override
+	protected T newObject () {
+		return poolTypeSupplier.get();
+	}
+
+	public interface PoolSupplier<T> {
```
tommyettinger, line 27: Just a little documentation would help here. This is the first recommended usage of Java 8's constructor reference syntax in all of libGDX, so showing how to use this might help.

gdx/src/com/badlogic/gdx/utils/DefaultPool.java
```diff
@@ -0,0 +1,30 @@
+
```
tommyettinger, line 1: I thought some automated thing took care of license headers, but maybe not... I think copying the existing license header from any file that has it is fine, if Spotless isn't handling it.

gdx/src/com/badlogic/gdx/utils/Pools.java
```diff
@@ -34,10 +37,37 @@ static public <T> Pool<T> get (Class<T> type, int max) {
...
 	}
 
+	/** Returns an existing pool or null for the specified type, stored in a Class to {@link Pool} map. */
+	static public <T> Pool<T> getOrNull (Class<T> type) {
+		return typePools.get(type);
+	}
+
+	/** Returns a new or existing pool for the specified type, stored in a Class to {@link Pool} map. A new pool will be of type
+	 * {@link DefaultPool} */
+	static public <T> Pool<T> getOrDefault (Class<T> type, PoolSupplier<T> poolTypeSupplier) {
```
tommyettinger, line 52: Something about this seems strange to me. This doesn't act like Java 8's Map.getOrDefault(), since it takes a supplier rather than a default return value (which would be a `Pool<T>`). I'm not sure if there's a better name, though... `getOrCompute()` would probably be close to the behavior, but I'm not sure...

gdx/src/com/badlogic/gdx/utils/ReflectionPool.java
```diff
@@ -22,7 +22,12 @@
...
+ *
  * @author Nathan Sweet */
+@Deprecated
```
tommyettinger, line 30: I'm not really convinced that ReflectionPool is actually obsolete or that it should be deleted in the future. The existing code that uses ReflectionPool is shorter and more convenient, even if its use of reflection internally causes problems sometimes.

tommyettinger (changes requested):
I like the idea and I think avoiding reflection where possible is good. I don't like deprecating code if there isn't a clearly-always-better alternative. I think if this PR didn't deprecate anything, I would have no substantive issues with it, just some requests to document or consider alternate names. I don't think this is currently passing automated build steps, I'd have to check that... Even if it did, I'm wary of deprecating code that works in 1.12.1 and will work in the next version, just with a big warning.

Berstanio:
> I don't like deprecating code if there isn't a clearly-always-better alternative.

I partially agree, I think a new API shouldn't be substantially worse in any case, but can be just equal in most cases.
And I don't think the `DefaultPool` is substantially worse in a way, do you have a specific use-case in mind that might have problems?

> I don't think this is currently passing automated build steps

GWT as usual!

gdx/src/com/badlogic/gdx/utils/Pools.java
```diff
@@ -16,13 +16,16 @@
...
+import com.badlogic.gdx.utils.DefaultPool.PoolSupplier;
+
+/** Stores a map of {@link Pool}s by type for convenient static access.
  * @author Nathan Sweet */
 public class Pools {
 	static private final ObjectMap<Class, Pool> typePools = new ObjectMap();
 
 	/** Returns a new or existing pool for the specified type, stored in a Class to {@link Pool} map. Note the max size is ignored
 	 * if this is not the first time this pool has been requested. */
+	@Deprecated
```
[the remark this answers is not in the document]
Berstanio: > For one thing, get(Class) is just an easier API to call.

I agree and I struggled a bit with the `Pools` class. On the one hand I wanted to keep the semantics of the old methods with a clear replacement, on the other hand it's not easily possible.
If I could re-desing the class I would split the logic up to a `Pools#create` and `Pools#get/obtain` methods. And before you can call `get/obtain` you need to create/set a Pool first one time. This would be in my opinion an okay migration requirement and should be easy enough when well documented.

> Would it be possible to have ReflectionPool actually be a DefaultPool that acquires a constructor via reflection?

Yes, but I don't really like the idea. Preferably we could offer a good enough `DefaultPool` API that makes the `ReflectionPool` uneccessary. 

> I don't think this should be merged without explaining in the docs what the deprecation is related to, and how to get past it.

Agreed, forgot to document it.

> Also, obtain() now calls what is deprecated API.

Yes, oversaw that.

gdx/src/com/badlogic/gdx/utils/DefaultPool.java
Berstanio: I can add it, but it's really just standard java syntax at this point.

gdx/src/com/badlogic/gdx/utils/Pools.java
Berstanio: Yeah, as mentioned in the other comment, I struggled with the method and method names. Unfortunatly, we can't seperate method signatures by generics, thats the reason of the weird name.
But yeah, I will further think on the `Pools` class and methods.

gdx/src/com/badlogic/gdx/utils/ReflectionPool.java
Berstanio: > that it should be deleted in the future.

I would keep it indefinitly, I just think discouraging it's use is a good idea, if the DefaultPools acts as a good replacement.

gdx/src/com/badlogic/gdx/utils/Pools.java
PokeMMO: For what it's worth, `computeIfAbsent` is the equivalent name.

gdx/src/com/badlogic/gdx/utils/Pools.java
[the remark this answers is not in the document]
PokeMMO: I'd go as far to say any gains from the "easier" usage api will be lost by increased difficulty adding GWT/R8/Proguard/NativeImage support later.

I would go as far to say ReflectionPool should be entirely removed.

PokeMMO (approved):
Outside of the minor issues tommyettinger addressed, this should be a great improvement.

I personally am all for deprecation (with the intent to remove) of ReflectionPool.

gdx/src/com/badlogic/gdx/utils/DefaultPool.java
tommyettinger: It is for normal Java development, but libGDX is still targeting Java 7 at this point in time.

tommyettinger:
These places would be using the deprecated Pools.get(Class) method:

It can be another PR, but should we think about getting them to use DefaultPool? I'm just a bit concerned about 1) user code suddenly displaying deprecation warnings on an update, even though the code is still functional and correct, and 2) having to actually get everyone using Java 8, finally. I don't think this PR even uses Java 8 syntax or features, it's just much easier to create the suppliers with Java 8.

PokeMMO:
We could replace `Pool#getOrDefault` with something like

```
static public <T> Pool<T> get (PoolSupplier<T> poolTypeSupplier) {
    T tmp = poolTypeSupplier.get();
    Class<T> type = (Class<T>)tmp.getClass();
    Pool<T> pool = typePools.get(type);
    if (pool == null) {
	pool = new DefaultPool<>(poolTypeSupplier);
	typePools.put(type, pool);
    }
    //optional?
    pool.free(tmp);
    return pool;
}
```

This is unlikely to be called enough that the performance of a single creation would matter.

tommyettinger:
It doesn't have to replace it, if for whatever reason users don't want to call a constructor once and free that created item once. That get(PoolSupplier) is a good idea to add, I think. It mirrors the constructor for DefaultPool, also taking one PoolSupplier arg.

Berstanio:
@PokeMMO @tommyettinger 
I updated the implementation, and I think the new design now matches ease-of-use with the old API. Feel free to comment your opinions on this.

If we agree on a design, I will do the CHANGES entry/proper javadoc/migrate libGDX internal uses.

gdx/src/com/badlogic/gdx/utils/ReflectionPool.java
tommyettinger: Upon revisiting,

`MyClass::new`

is shorter than

`MyClass.class`

So my claim that ReflectionPool's usage is shorter was incorrect!

tommyettinger:
So some of my review has been resolved; only docs are missing now. And my view hasn't changed, method docs are still important here, and the constructor reference syntax should at least be shown in use somewhere.

Berstanio:
Okay, so I did the changes/javadocs + example.
I wanted to also change the internal use, but I realised, that the code gets awful with inner classes, since method references are not available.
This PR probably makes mostly sense, if we also bump the target to java 8.

Berstanio:
Since we bumpet the target now, all libGDX internals use the new supplier based pool too!
This means, the PR can be reviewed again.

tommyettinger (changes requested):
Everything's good except for docs on DefaultPool and its nested PoolSupplier. Just one-liner method docs for anything that isn't totally obvious would be nice, and some kind of message for PoolSupplier (maybe mentioning why we don't use `Supplier<Pool>`) is IMO important.

tommyettinger (approved):
Docs are fine now, approved!

PokeMMO (approved):

PR #7484, opened 2024-10-22
Improve Android input keyboard with autocorrection

fgnm:
Especially with latest libGDX release, I've noticed that keyboard in android works pretty bad (both with stock keyboards and custom ones). The main reason is that in android backend the autocorrection is enabled by default. The current implementation of TextFields fits quite bad and the whole experience with text input is poor, here my tests:

- Fleksy Keyboard: it's not possible at all to write text.
- Samsung Keyboard: it's possible to write with autocorrection but only word by word, text is not typed in realtime and issues with blank spaces happens
- Google keyboard: Works and type in real time, but mess the text when autocorrection fix a word

In iOS backend, indeed, the default implementation of the keyboard is set to disable the autocorrection, which imo makes a lot of sense thinking on how input fields are needed for a game. However, as discussed in Discord, autocorrection might be useful in features like chats, so I've add a new `OnscreenKeyboardType` to enable autocorrection both in Android and in iOS, but leave it disabled in the default case.

Fixing this on project side is quite tricky because the static method used, a fix at libGDX level would be very appreciate, thanks :)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.13.1]
+- iOS: Update to MobiVM 2.3.22
 - Change visibility of PolygonSpriteBatch.switchTexture() to protected
 
 [1.13.0]
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -311,9 +311,6 @@ public class DefaultAndroidInput extends AbstractInput implements AndroidInput, 
 		case URI:
 			inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI;
 			break;
-		case Autocorrection:
-			inputType = InputType.TYPE_CLASS_TEXT;
-			break;
 		default:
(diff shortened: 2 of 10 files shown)


Berstanio:
I'm not entirely sold on the new enum value. Making Autocorrection/NoAutocorrection an input type makes it not mixable with others, e.g. e-mail which is a bit inconvinient. I think this would better off as a flag or something like that.

But I would be also fine just completly disabling autocorrect for the `setOnscreenKeyboardVisible` API with no flag to enable it. As you said, it's very broken there anyway and people that want the extra features should use the native input.

fgnm:
@Berstanio totally agree, thinking that we've the native input option autocorrection seems to be just useless.. After all any other use case (like chats, etc.) can be covered with native input and user experience will be surely better.

Berstanio (changes requested):
The native input API uses the same method to determine their inputType, so this would break this API.
I _think_ the correct way to only apply it to the `setOnscreenKeyboardVisible` API would be in `GLSurfaceView20#onCreateInputConnection` modifying the input type.

fgnm:
Should be fine now

Tom-Ski:
Looks ok to me. @Berstanio wanna sign off on it?

PR #7485, opened 2024-10-22
Migrate androidx API to internal Android 30+ one

Berstanio:
This PR removes the need for androidx on libGDX usual use. 
From looking through the source, androidx on sdk 30+ just redirected the API to the system API anyway, so this shouldn't really change anything.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/keyboardheight/AndroidRKeyboardHeightProvider.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/keyboardheight/AndroidRKeyboardHeightProvider.java
@@ -4,14 +4,16 @@ package com.badlogic.gdx.backends.android.keyboardheight;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.res.Configuration;
+import android.os.Build.VERSION_CODES;
 import android.view.View;
-import androidx.core.graphics.Insets;
-import androidx.core.view.OnApplyWindowInsetsListener;
-import androidx.core.view.ViewCompat;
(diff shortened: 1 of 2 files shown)


tommyettinger:
Thanks for this! The AndroidX thing has come up a few times already, so getting it back to not-strictly-needed would be welcome. I don't think I can review this reasonably because I don't know enough about Android or its internals.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/keyboardheight/AndroidRKeyboardHeightProvider.java
```diff
@@ -4,14 +4,16 @@
...
+import android.os.Build.VERSION_CODES;
 import android.view.View;
-import androidx.core.graphics.Insets;
-import androidx.core.view.OnApplyWindowInsetsListener;
-import androidx.core.view.ViewCompat;
-import androidx.core.view.WindowInsetsCompat;
+import android.view.View.OnApplyWindowInsetsListener;
+import android.view.WindowInsets;
+import android.graphics.Insets;
+import androidx.annotation.RequiresApi;
```
kevlahnota, line 12: you can replace `import androidx.annotation.RequiresApi` with `import android.annotation.RequiresApi;`

Roaddog56:
No thanks

obigu (approved):
Looks good to me

Berstanio:
After some discussion it was decided, that while we did not intended to add a androidx dependency, it is generally the right thing to do.
It is the recommended replacement for multiple deprecated API's and offers better compatibility accross API levels.
A change to make androidx an explicit dependency of libGDX will follow.

Issue #7480, opened 2024-10-18
Texfield Bug on GWT build with 1.13.0 release


BoBIsHere86:
Error Overview:
When using libGDX 1.13.0 and a gwt build and you have a scene2D textfield on screen. 
It causes repeated Errors in the console stating: "[.WebGL-000049340012A900] GL_INVALID_OPERATION: Insufficient buffer size"
Until eventually "WebGL: too many errors, no more errors will be reported to the console for this context."

*Side note, I did attempt pulling back the gwtFramework from 2.11.0  back to 2.10.0.  But, it had to effect.

I've tested this across 3 projects. Sometimes it happens on load. But it mainly seems to be related to when you focus on it or interact with the textField. Also possibly related to the cursor blinking. I will provide GIFs below to help understand.

I have included the console in each of the grabs so u can see whats going on.

Here are 2 back to back examples showing the error with the basic scene2D template from LiftOff (which i added a textfield to.
This is it running perfectly fine in 1.12.1
![1 12 1_texfield](https://github.com/user-attachments/assets/56907346-70b9-495e-ba6d-746416a66b2e)

After upgrading to 1.13.0, we can see the issue:
![1 13 0_texfield](https://github.com/user-attachments/assets/29cff2d5-f481-4966-9b54-c67399f322be)

This is a test using a jam project i made where i just slapped a textfield in the upper left corner. (this one only starts breaking when interacting with the textfield)
![jamTextfield](https://github.com/user-attachments/assets/a390a4d8-49f1-4a12-bc88-f4e7c94ea564)

This is another unrelated project which i updated to 1.13.0. One again only begins breaking when interacting with it.
![snekGame](https://github.com/user-attachments/assets/51a818cb-1ad3-416e-9d20-3bb4eebd19ec)

Reproduction:

The Easiest way to reproduce this is as follows.
Open up libGDX liftoff.(using version 1.12.1.17 at the moment)
Create a basic project with the HTML/GWT selected and the Scene2d Template.
When the project opens up, change some settings in order to update to 1.13.0.
In gradle.properties, change gdxVersion=1.12.1 to gdxVersion=1.13.0.
Then under the HTML project, in it's build.gradle.
You will see this:
```
dependencies {
  implementation "com.badlogicgames.gdx:gdx:$gdxVersion:sources"
  implementation "com.github.tommyettinger:gdx-backend-gwt:1.1210.1"
[...]

PR #7486, opened 2024-10-26
Change defaults of SpriteBatch to use VBO instead of VertexArray on gl2/gles2.0 

Tom-Ski:
Regression in gwt due to VertexArray/IndexArray not being implemented the same on all platforms.

After tests, I havent seen an issue in just moving from client side buffers (VA), to VBOs as the new default.

For all other buffer types we can still pre-upload indices, and save us some nice time during rendering.

Test results for lwjgl, lwjlg3, android and gwt below. IOS not tested yet.

Times here are all just relative to each other. Lower the better
```
android gl2
0.195 vertex array 
0.159 vbo

android gl3
0.171 vao
0.168 vbo

===============

lwjgl2 gl2
0.105 vertex array
0.88 vbo

lwjgl2 gl3
0.09 vao

===============

lwjgl3 gl2
0.105 vertex array
0.09 vbo

lwjgl3 gl3
0.08 vao

===============

gwt gl2
1.427 vertex array
1.309 vbo

gwt gl3
1.309 vbo
0.921 vao
```

https://github.com/libgdx/libgdx/issues/7480
https://github.com/libgdx/libgdx/commit/bfe255a2727377b910be20af48d40867c588a8a3

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
@@ -40,13 +40,10 @@ public class SpriteBatch implements Batch {
 	 *             used when gles 3 is not available, defaults to {@link VertexDataType#VertexArray}. */
 	@Deprecated public static VertexDataType defaultVertexDataType = VertexDataType.VertexBufferObject;
 
-	/**
-	 * Used to completely override the vertex type used by SpriteBatch. This is useful for picking a specific
-	 * vertex data type on construction of the sprite batch. Recommended to reset this back to defaultVertexDataType
-	 * Once the batch has been created with this flag
-	 */
-	@Deprecated
-	public static VertexDataType overrideVertexType = null;
+	/** Used to completely override the vertex type used by SpriteBatch. This is useful for picking a specific vertex data type on
+	 * construction of the sprite batch. Recommended to reset this back to defaultVertexDataType Once the batch has been created
+	 * with this flag */
+	@Deprecated public static VertexDataType overrideVertexType = null;
 
 	private VertexDataType currentDataType;
 
@@ -979,13 +976,12 @@ public class SpriteBatch implements Batch {
 		Mesh mesh = this.mesh;
 		mesh.setVertices(vertices, 0, idx);
 
-		//Only upload indices for the vertex array type
+		// Only upload indices for the vertex array type
 		if (currentDataType == VertexDataType.VertexArray) {
 			Buffer indicesBuffer = (Buffer)mesh.getIndicesBuffer(true);
 			indicesBuffer.position(0);
 			indicesBuffer.limit(count);
 		}
-
 
 		if (blendingDisabled) {
 			Gdx.gl.glDisable(GL20.GL_BLEND);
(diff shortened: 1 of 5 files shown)


BoBIsHere86:
Since I'm the jerk who submitted the issue. I pulled down the PR, built it locally, deployed and tested it out. Fixes the issue as initially reported. 
Awesome work!
fixTest

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
```diff
@@ -38,7 +38,14 @@
 public class SpriteBatch implements Batch {
 	/** @deprecated Do not use, this field is for testing only and is likely to be removed. Sets the {@link VertexDataType} to be
 	 *             used when gles 3 is not available, defaults to {@link VertexDataType#VertexArray}. */
-	@Deprecated public static VertexDataType defaultVertexDataType = VertexDataType.VertexArray;
+	@Deprecated public static VertexDataType defaultVertexDataType = VertexDataType.VertexBufferObject;
+
+	/** Used to completely override the vertex type used by SpriteBatch. This is useful for picking a specific vertex data type on
+	 * construction of the sprite batch. Recommended to reset this back to defaultVertexDataType Once the batch has been created
+	 * with this flag */
+	@Deprecated public static VertexDataType overrideVertexType = null;
```
tommyettinger, line 46: OK, another Deprecated field; I can sorta see why this exists now. Is there a plan to eventually remove these public deprecated fields?

tommyettinger (approved):
I'm not totally clear on how this is acting differently (and better than before) on GWT than it does on other platforms, but I'm glad it does. I would eventually like to know what the plan is for deprecation in libGDX in general -- do we intend to ever remove these deprecated fields?

gdx/src/com/badlogic/gdx/graphics/g2d/SpriteBatch.java
Tom-Ski: These are just for warnings mostly, as you should not touch these without understanding the consequences. There is no safe handling of what is compatible, no errors etc.

Probably would be better to have extra SpriteBatch constructors to create with specific VertexDataTypes, and then handle all incompatibilities there and give sensible warnings and exceptions when someone tries to create with something invalid.

Something that can be addressed in a separate PR as wanted this to be as noninvasive as possible.

Tom-Ski:
These deprecated fields can be removed with something like described in comments. Mostly they are used everywhere as warnings of not to use things. Its not often someone will be checking the javadocs for every method/field to see if its experimental or potentially a bit dangerous, but deprecated annotations make these very obvious.

As for the difference, Gwts implementation of VertexArray wasn't the same as other platforms. Which is why that bug cropped up.  Its not actually a client side vertex array. But should be working better now, because it no longer uses 'vertex arrays', and uses the vbo by default on gl2. With benefits from previous commits to optimize not updating indices, and in general a bit more of a performant approach by having these buffers mostly gpu side, we get a little boost too.

obigu (approved):
Tested on real projects on Android and iOS looks good. Also have run the SpriteBatchPerformanceTest on an Android 8 gl2 and results are Vertex Array 1.2s and VBO 0.65s

Issue #7488, opened 2024-10-28
TiledMap PR/Issue Cleanup Initiative


BoBIsHere86:
**Quick Overview:**
There are Currently 9 open Tiled Map PRs. 7 of them in a state of (what looks to be) completeness.
And 11 open TiledMap issues in the Tracker.

With the help and advice of everyone involved in the related PRs. I believed we could easily knock out 8 of the 9 PRs without issue and close out 6 issues (possibly more if some issues are outdated).
Plus, with the recent release of libGDX 1.13.0, I'd assume we would have PLENTY of time to get this done before the next release.

I know the TiledMap codebase has always been a point of contention and some people believe it should be refactored and split off into it's own library. (That's probably a VERY good idea) But For now, we should aim to incorporate the bug fixes and feature requests already in progress, to get us to a fairly feature complete state before any major restructuring.

**Why am I clogging up the tracker with a new Issue?**
I'm no authority on the TiledMap codebase. But I've had a bit of free time this week and looked through all of the PRs to see what state they were in and to my surprise most of them were pretty complete. But since the tiledmap codebase is pretty tightly coupled, changes in multiple PRs have a cascading effects on other PRs causing a whole chain of back and fourth with different contributors. Based on which get's merged first. And that can be a headache.

So I thought I'd provide 1 possible path to victory.
The idea here is that the contributors can make the necessary changes to support the PR that was merged before it.
I will make notes for Each Contributors PR if needed.

**Suggested PR Merge Order Steps:**
(Each step assumes the previous one has been merged to Master)

~~**First Merge:**~~ Merged AS of October 30th, 2024 👍 
~~**"FIX: Parsing imagelayer inside a group tag in TMX file"** - PR: #7135
Why?
Out of all the PRs this one is the most self contained and would have a smaller impact on any of the follow up PRs.
As well as fixing a long standing issue that's been around forever.~~

~~**Second Merge:**~~ Merged as of November 21st, 2024 👍 
~~**"Add a tiledmap loader for json files"** PR(#7505) (old: PR: #7386) - CLOSES Issue #7291 and #5707 (Duplicate issues)
A New PR(#7505) has been made to replace the old one one, since the original Author could no longer continue work on it.
This PR right here is the biggest decision I feel. As this PR has the largest impact on all the others. 
I believe it would be best to get it done first. Which allows the other contributors to properly add and test changes to their own PRs based around the new BaseTmjMapLoader.java~~

**Third Merge:**
[...]

Issue #7427, opened 2024-07-28
Tmx MapLoader crashes when using file properties and does not correctly load Text nodes


Quillraven:
Issue as disucced in #7291

Regarding the problem for text nodes: I only needed the `text` value in my game which I did not see anywhere in the LibGDX model classes. Therefore, I have written my own loader that simply calls `loadObject` and enhances the properties afterward:

```kotlin
class TiledLoader(fileHandleResolver: FileHandleResolver) : TmxMapLoader(fileHandleResolver) {

    // since LibGDX does not load the "text" information of a Text object in Tiled, we need to overrule
    // LibGDX's loadObject method and parse the text info ourselves and add it to the properties of the mapObject
    override fun loadObject(map: TiledMap, objects: MapObjects, element: XmlReader.Element, heightInPixels: Float) {
        super.loadObject(map, objects, element, heightInPixels)
        val lastObject = objects.last()
        element.getChildByName("text")?.text?.let { objTxt ->
            lastObject.properties.put("text", objTxt)
        }
    }
}
```

Text is just one of the fields of a text node in Tiled. There is also Font, Word Wrap and Alignment from what I can see:
![image](https://github.com/user-attachments/assets/24a683fa-5bb1-40ec-922e-5a3ab17f5536)

Would be of course beneficial to load all of them correctly.

----

The second issue is when using a file property in Tiled. That simply crashes LibGDX. Imo, the problematic method is in `BaseTmxMapLoader`:

```java
protected Object castProperty(String name, String value, String type) {
        if (type == null) {
            return value;
        } else if (type.equals("int")) {
            return Integer.valueOf(value);
        } else if (type.equals("float")) {
            return Float.valueOf(value);
        } else if (type.equals("bool")) {
            return Boolean.valueOf(value);
        } else if (type.equals("color")) {
            String opaqueColor = value.substring(3);
            String alpha = value.substring(1, 3);
            return Color.valueOf(opaqueColor + alpha);
        } else {
            throw new GdxRuntimeException("Wrong type given for property " + name + ", given : " + type + ", supported : string, bool, int, float, color");
        }
    }
```

"file" is not listed and therefore the GdxRuntimeException is thrown. Imo, file should be handled the same way as a string. No casting necessary. I don't know if a `FileHandle` makes sense. That might be difficult because Tiled stores the relative path to the map, if I remember correctly. That can be an arbitrary path outside the assets folder. I would leave that up to the user to handle it properly but at least give them the string.

[...]

PR #7386, opened 2024-04-20
Add a tiledmap loader for json files

Kevin-Garnier:
Add implementation from #7291

The code is based on what already exist for the xml loader.

I've alredy tested loading a map by launching the superkoalio game from the exemples and it worked fine.

I've also tested loading a custom Tileset from json and xml files.

PR #7498, opened 2024-11-02
Support for Text Objects in BaseTmxMapLoader + bugfix.

BoBIsHere86:
This PR is meant to address and close what was discussed in Issue #7427 

I will be adding this PR to the list of Tiled PRs I am keeping track of here #7488.
This is meant to be looked over and reviewed. But not merged until some of the more major additions are completed. Specifically #7386.

The intent of this PR is to finally add support for the Tiled Text Object into our BaseTmxMapLoader. It is the only object we do not currently support. This will allow users to load in data for any text objects they setup in their maps. This is not meant to render the text for the users as we do not render any objects. It's meant to just pass along the data for the user to implement in any way they see fit.

I have modified the old TiledMapObjectLoadingTest to include 3 text objects. You can see the comparison below:

textObjTest

Also, mentioned in the issue was a crash being caused by loading maps which contained objects which included a property of type "file".  This has been fixed in the castProperty() method.

All suggestions and comments welcomed!


Quillraven:
I will have a look at the code in the evening CET.

One question about the screenshot: Why is there a blue circle at the top next to the Polyline while in Tiled there is no object visible?

BoBIsHere86:
> I will have a look at the code in the evening CET.
> 
> One question about the screenshot: Why is there a blue circle at the top next to the Polyline while in Tiled there is no object visible?

That is a great question confused me as well when I first came across it. But the answer according to this piece of code in the test is
```
	// Test get objects by type (adding circle manually because it doesn't exists in Tiledmap editor)

		loadingStatus = "loading status:\n";
		MapLayer layer = map.getLayers().get("Objects");
		MapObjects mapObjects = layer.getObjects();

		mapObjects.add(new CircleMapObject(280, 400, 50));
```

So it's added in after the fact unrelated to the .tmx file. To "Test" Circle objects.

gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java
```diff
@@ -0,0 +1,269 @@
...
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.maps.objects;
+
+import com.badlogic.gdx.maps.MapObject;
+import com.badlogic.gdx.utils.Align;
+
+/** @brief Represents a text map object */
+public class TextMapObject extends MapObject {
```
Quillraven, line 23: I am not 100% sure but I think Text objects are always rectangles? If so, I'd vote to extend `RectangleMapObject`. One more reason is to still support LibKTX `shape` extension ([link](https://github.com/libktx/ktx/blob/master/tiled/src/main/kotlin/ktx/tiled/mapObjects.kt)) without the need to change anything there.

gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java
```diff
@@ -0,0 +1,269 @@
...
+	private float y = 0.0f;
+	private float width = 0.0f;
+	private float height = 0.0f;
+	private float rotation = 0.0f;
+	private String text;
+	// Alignment values as represented via libGDX Align
+	private int horizontalAlign = Align.left;
+	private int verticalAlign = Align.top;
+	// String alignment values
+	private String tiledHorizontalAlign;
```
Quillraven, line 35: I would just keep the String fields without the int fields because, as you pointed out, there is a  "justify" option for horizontal that LibGDX's `Align` does not have.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -400,6 +397,27 @@ protected void loadObject (TiledMap map, MapObjects objects, Element element, fl
...
+					textMapObject.setStrikeout(child.getIntAttribute("strikeout", 0) == 1);
+					textMapObject.setWrap(child.getIntAttribute("wrap", 0) == 1);
+					// When kerning is true, it won't be added as an attribute
+					textMapObject.setKerning(child.getIntAttribute("kerning", 1) == 1);
+					// Default color is #000000, not added as attribute
+					String textColor = child.getAttribute("color", "#000000");
+					// TODO:After PR #7076 is Merged. Update below section of code to use tiledColorToLibGDXColor().
+					String alpha = textColor.length() == 9 ? textColor.substring(1, 3) : "ff";
+					String color = textColor.length() == 9 ? textColor.substring(3) : textColor.substring(1);
+					textMapObject.setColor(Color.valueOf(color + alpha));
```
Quillraven, line 419: There is a very similar code in `castProperty` that converts the type color to a LibGDX `Color` instance. Is `tiledColorToLibGDXColor` the extracted method for this?

Quillraven (changes requested):

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
BoBIsHere86: So in my TintColor PR. I made a convenience method that will be used in all the areas where we take in tiled colors turn them into the libgdx color format.  That's just a note to myself to replace it after that one get's merged. And yes its similar to the code in castProperty(). (Which is replaced in the TintColor PR). 

Though interestingly  the Color code generated in the map files for the custom "Color" type property is always in the full AARRGGBB format. Unlike some instances, like font color or tint color where the color is passed in as RRGGBB unless you modify the alpha values then it passes it in as AARRGGBB.  Confusing, I know. lol

gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java
BoBIsHere86: So after I added in the option of also having the int values as well as the String. It was super convenient in the testing code being able to remove the giant alignment check before drawing it. I wonder if we should keep in on hand in case we ever add a justify?

gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java
BoBIsHere86: Can do! Though doing so, we immediately break the TiledMapObjectLoadingTest. Which is fine by me as long as it's ok for me to modify how we do the object checks.
Currently in the test they using instanceof. So the by extending from RectangleMapObject it get's picked up in there instead.

 ```
else if (mapObject instanceof ) {
				shapeRenderer.begin(ShapeRenderer.ShapeType.Filled);
				Rectangle rectangle = ((RectangleMapObject)mapObject).getRectangle();
				shapeRenderer.rect(rectangle.x, rectangle.y, rectangle.width, rectangle.height);
				shapeRenderer.end();
```

Also, I had no idea this LibKTX shape extension existed. I will take a look into it later. Also to note, there is another PR in the pipeline. Someone added a PointMapObject #6384. I'm assuming we may have to modify LibKTX shape extension as well to keep it all in sync?

Comment (Quillraven) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:23]: I see the point of having it separate. Then let's keep it separate to be consistent and not break an existing test. I can do the PR in LibKTX for the new object.

File: gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java (hunk 2 above)
Comment (Quillraven) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:35]: I don't have a strong opinion against it but the way I see it, users anyway have to do something with map objects because they are not rendered by the tiledmap renderer and most likely they have some game specific setup/parsing for map objects. 

E.g. in an ECS environment you will have an entity with a Text component or whatever, and you will then parse the TextObject and convert it to your own model. Doesn't matter if your parse the original strings or the int values. The string values though have the additional "justify" value which someone COULD do something with it.

But as mentioned, no problem if we keep it ;)

**Review by Quillraven — APPROVED** [review]

File: gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java (hunk 1 above)
Comment (BoBIsHere86) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:23]: Hmm, yea, looking through all the map objects even the Ellipse and Circle were made completely separated.
If we kept it as is, we could always have it maintain own Rectangle instance. Would there a need for that? Or should we just stick to having it keep track of all it's properties?

File: gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java (hunk 2 above)
Comment (BoBIsHere86) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:35]: Hmm... You are right. Maybe we should just try and follow the examples led by the other map objects. As far as I see there's no logic or utility methods in any of these. Circle just contains a Circle object, Polyline a Polyline and TextureMapObject contains a bunch of properties and a TextureRegion. No utility methods or logic. I guess maybe I'll go back to just storing the string values and not pollute it with all the extra stuff. Unless some other Tiledmap user chimes in and would like this. So I'll take my time with it.

File: gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java (hunk 1 above)
Comment (Quillraven) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:23]: Having its own rectangle instance with a getter would be great for the shape extension for LibKTX. I think it makes sense because an object is always related to a specific shape in Tiled like point, ellipse, rectangle or polygon.

Have a look at the RectangleMapObject and implement it in a similar way in the new text object.

Comment (BoBIsHere86) [reply gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java:23]: I refactored the TextMapObject a bit to be better in line with the other MapObjects. You're right in that all tiled objects do have a core "shape" to them and in the case of Text, that's obviously the object's bounds. So it makes sense it has it's own rectangle. 
Also, I removed all the Align.class related code. Convenient or not. Since it's not 1 to 1 with the Align.class. There is no point to force a relationship, especially since Tiled could easily change it at some point.
This class is now basically just a bunch of data representing the text object and it's properties. So it's up to the user to do with it as they need. 
Hope that's good. Let me know if I should alter anything else.

**Comment by Quillraven** [comment]
From my side it looks great and can be merged - thank you! Who has the authority to do that?

One last thing: a new MapObject might theoretically break existing code. Imagine if you have something like 

```
if(mapObj is RectangleMapObject) {

} else if(mapObj is PolygonMapObject) {

} /* other cases */
else {
  // only object left is e.g. EllipseMapObject
}
```

In those scenarios else will now be executed for the new TextMapObject as well. 

Maybe a note in the release notes doesn't hurt?

BoBIsHere86:
> From my side it looks great and can be merged - thank you! Who has the authority to do that?
> 
> One last thing: a new MapObject might theoretically break existing code. Imagine if you have something like
> 
> ```
> if(mapObj is RectangleMapObject) {
> 
> } else if(mapObj is PolygonMapObject) {
> 
> } /* other cases */
> else {
>   // only object left is e.g. EllipseMapObject
> }
> ```
> 
> In those scenarios else will now be executed for the new TextMapObject as well.
> 
> Maybe a note in the release notes doesn't hurt?

You are correct. We have a "slightly" similar issue with the PR for the new PointMapObject already. 
In BaseTmxMapLoader, the loadObject() basically ends up building a RectangleMapObject if nothing else was triggered.
So if anyone had been using "Point Objects" this whole time they were created as Rectangle. Once that PR is merged, those points are no longer Rectangles.
But the PR does have a Breaking Change note.

Luckily there is not any existing code in the codebase that would be negatively effected other than "any object not found is turned into a Rectangle Map Object"

My hope for this PR and all the others is that if we can pull this off and get everything merged, we can have a nice little section in the change notes for the next libGDX release specifically for he mass of features added for tiled compatibility. 


Oh and as for merging. I was really hoping to hear back on PR #7386 by now as it's one of the newer PRs. It's a huge feature and once it's merged everyone's changes would need to be replicated and tested there. I hope the original contributor hasn't abandoned us because of the lack of movement. haha

Quillraven:
> Oh and as for merging. I was really hoping to hear back on PR https://github.com/libgdx/libgdx/pull/7386 by now as it's one of the newer PRs. It's a huge feature and once it's merged everyone's changes would need to be replicated and tested there. I hope the original contributor hasn't abandoned us because of the lack of movement. haha

I remember that this was an issue for my parallax PR because like half a year or a year later someone mentioned an issue (I think the x movement was inverted) and shortly after it got merged. I think I did not even fix that in the end.

I don't know who is responsible for the "Tiled module" in LibGDX but I am happy to help out. I usually respond within a day or two. But I don't know how I get the rights to merge pull requests or how I can make myself trustworthy to whoever is in charge for that :D

BoBIsHere86:
> n fix that in the end.

Hah, yes. I remember that, I think I was the one who sent the PR to fix the invert issue. lol
Maybe the same can be done for the json loader since I dropped a PR with the completed code into his repo.

Anyways, thanks for all the help. Hopefully we can get this all squared away soon.

BoBIsHere86:
With the small refactor we did in the json loader PR finally merged. I went ahead and rebased this branch and updated the tmj loader to support the new TextMessageObject. I also updated the TiledMapJsonObjectLoadingTest test to include TextMapObjects. There should no longer be any conflicts when the time comes to eventually merge this.

BoBIsHere86:
I have rebased this PR against the latest codebase. Updating as needed against the 2 most recently merged Tiled PRs. There should now be no merge conflicts.
Also to note, the "+ bugfix" originally mentioned in the title of this PR has since been fixed in @Quillraven's #7534 pr.
I took in his changes.

I'm still waiting on the the PointMapObject PR getting merged. #7523 Once that is done I can update this and easily merge the changes to the TiledMapObjectLoadingTest to include the pointmapobject as well.

Other than that. Everything is good.

BoBIsHere86:
> @BoBIsHere86 thanks for the work on maps! I've merged your Point PR, and happy for this one to be included too. Just needs a conflict resolution and good to go

Awesome, thanks. I'll have this fixed up within the next hour or so. Cheers!

BoBIsHere86:
Alrighty,
Did a rebase against the master. Cleaned up the code, made sure the point map object properly integrated object into the map object Test with the new Text Object, in both the tmj and tmx fromat.
point_added

Also, I added a line to the CHANGES file since I always forget to do that.

Once this is merged I think there is only 1 more completed PR I will have to update against these latest changes, and this Tiled map journey can come to an end. Wooooo!

patton73:
I will open a new feature request to have public tileset loaders after all this work is merged. However great job!
Thanks a lot.

patton73:
Guys please merge this Pull request.

gdx/src/com/badlogic/gdx/maps/objects/TextMapObject.java
tommyettinger: There's a "Justify" PR for Labels that I've been meaning to look over, I think it's open now.

BoBIsHere86: Ah nice. Just took a look at the PR. Since it seems we will most likely have this as an option in the future. I think I will go and add these back in. 👍

PR #7499, opened 2024-11-02
XmlReader.getChildren() + XmlReader.replaceChild()

TCreutzenberg:
Until now XmlReader did not offer any possibility to e.g. replace a child or to access the children of an Element directly to perform actions there.
This PR introduces these two methods to manipulate xml files. (In my case I want to adjust my Tiled .tmx and .tsx files in a custom map packer)

- XmlReader.getChildren()
- XmlReader.replaceChild()

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/XmlReader.java
+++ b/gdx/src/com/badlogic/gdx/utils/XmlReader.java
@@ -507,6 +507,10 @@ public class XmlReader {
 			return children.size;
 		}
 
+		public Array<Element> getChildren () {
+			return children;
+		}
+
 		/** @throws GdxRuntimeException if the element has no children. */
 		public Element getChild (int index) {
 			if (children == null) throw new GdxRuntimeException("Element has no children: " + name);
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/utils/XmlReader.java
```diff
@@ -534,6 +538,14 @@ public void removeChild (Element child) {
 			if (children != null) children.removeValue(child, true);
 		}
 
+		public void replaceChild (Element child, Element replacement) {
+			if (children == null) throw new GdxRuntimeException("Element has no children: " + name);
+			final int index = children.indexOf(child, true);
+			if (index == -1) throw new GdxRuntimeException("Element does not contain child: " + child);
+			children.removeIndex(index);
+			children.insert(index, replacement);
```
tommyettinger, line 546: Is there any reason why this doesn't use `children.set(index, replacement);` ? Removing by index has to move later items toward the start, and then inserting it has to move them again toward the end. The appropriate method would be set(), I think, unless you want to only remove or only insert an item.

tommyettinger (changes requested):
This should be a quick change, but remove-then-insert is two O(n) operations, when that could be one O(1) with set().

SimonIT (changes requested):
I think the changes should also be added to gdx/res/com/badlogic/gdx/utils/XmlReader.rl

gdx/src/com/badlogic/gdx/utils/XmlReader.java
TCreutzenberg: You're absolutely right.

tommyettinger (approved):
Looks good to me!

TCreutzenberg:
All comments addressed.

SimonIT (approved):

PR #7500, opened 2024-11-04
Fix pauseWhenLostFocus not working as expected

obigu:
Recently added `pauseWhenLostFocus` config was not working. First of all code was not being executed unless a `WindowListener` was added by the app and even if it was a `resume()` was missing.


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
```diff
@@ -60,14 +60,15 @@ public void invoke (long windowHandle, final boolean focused) {
...
 						if (focused) {
 							if (config.pauseWhenLostFocus) {
 								synchronized (lifecycleListeners) {
 									for (LifecycleListener lifecycleListener : lifecycleListeners) {
 										lifecycleListener.resume();
 									}
 								}
+								listener.resume();
 							}
 							windowListener.focusGained();
```
crykn, line 73: This will throw an NPE if `windowListener` is null.

obigu: yeah, should be ok now

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Window.java
```diff
@@ -60,28 +60,31 @@ public void invoke (long windowHandle, final boolean focused) {
...
+							synchronized (lifecycleListeners) {
+								for (LifecycleListener lifecycleListener : lifecycleListeners) {
+									lifecycleListener.pause();
 								}
-								listener.pause();
 							}
+							listener.pause();
 						}
-						Lwjgl3Window.this.focused = focused;
+						Lwjgl3Window.this.focused = false;
```
crykn, line 87: I think you accidentally moved this line into the else block. It should also stay as `Lwjgl3Window.this.focused = focused;`.

obigu: You're right but I'm not sure it was correct before either. Before it was only set if `windowListener != null` which doesn't make sense to me either as the `focus` property wouldn't be updated in some cases right?

crykn: Yes, that seems to be wrong as well.

obigu: Ok, changed, unintendedly we have fixed fix another bug in master :)

crykn (approved):

PR #7507, opened 2024-11-10
add GL constants for MSAA

0XDE57:
-add MULTISAMPLE_EXT (0x809D)
-add SAMPLE_ALPHA_TO_ONE_EXT (0x809F)
-https://registry.khronos.org/OpenGL/extensions/EXT/EXT_multisample_compatibility.txt


0XDE57:
NOTE: this may be supported by desktop only.
web is maybe, android is maybe.
unlikely to work on iOS, still investigating.

PR #7516, opened 2024-11-19
A few modifications to LongArray and a test class for it

melvilpf:
I modified three methods of the LongArray class:
- lastIndexOf took a char argument when it was supposed to be a long since we are working with long arrays ;
- pop() and peekl() did not handle the case of empty arrays so I added this case.

I then implemented a test class for LongArray, testing all the methods I could (of course I can't test random() for example).

I also apologize, I did not want the iosgl20 and iosgl30 to be modified and couldn't remove the modifications.


gdx/src/com/badlogic/gdx/utils/LongArray.java
```diff
@@ -272,11 +272,13 @@ public boolean removeAll (LongArray array) {
 
 	/** Removes and returns the last item. */
 	public long pop () {
+		if (size == 0) throw new IllegalStateException("Array is empty.");
```
tommyettinger, line 275: Good catch here; calling pop() on an empty LongArray would produce a negative `size` and only then throw an Exception. Catching that would still mean dealing with negative size later!

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+		 */
+
+		int expectedHash = 205510429;
+		Assert.assertEquals(expectedHash, longArrayOrdered.hashCode());
+	}
+
+	/** Test of the indexOf() method */
+	@Test
+	public void indexOfTest () {
+		LongArray longArray1 = new LongArray(new long[] {1, 3, 4, 5, 6, 6, 3, 9});
```
tommyettinger, line 215: It's a little odd that the fix was for lastIndexOf() taking a char parameter, when it should have been long, but all of the tested values would still be valid before the fix. If some were negative or greater than 65535, then this would be a better test of the change that was just made.

gdx/src/com/badlogic/gdx/utils/LongArray.java
tommyettinger: One possible nitpick, now that I think about it, is that if the IllegalStateException is caught once, and pop() is called again, it will throw an IndexOutOfBoundsException just after the catch, but won't throw an IllegalStateException a second time, because size will be -1, and that won't be == 0. We might prefer size <= 0 because the size can be set externally, plus this method can make the size negative, and all sizes 0 or less are invalid here.

tommyettinger (changes requested):
The changes look mostly good to me, with some tiny issues. I think the pop() and peek() checks should see if `size <= 0` rather than `size == 0` because size can be assigned a negative value externally or by catching exceptions and letting size change in pop().

The test seems fine; it is only really testing smaller int values (within the range of shorts, it looks like), so the change in this PR to lastIndexOf() could have escaped notice, but maybe not... I don't have an issue with the tests right now, it's good to have them.

So, wrapping up, changing `size == 0` to `size <= 0` if both negative values and 0 are invalid would be good. This could be made in other places too, since it shouldn't break any valid operations, and will just  notice problems hopefully before they happen elsewhere.

melvilpf:
Hi, I've just made the changes requested. Also, I'm pretty new to GitHub and don't know if I should click on Update branch because I have a warning "This branch is out-of-date with the base branch".

melvilpf:
I will also maybe be testing shuffle and random with stats but tell me if that's relevant before I begin.

tommyettinger (approved):

tommyettinger:
I approved it as-is, it looks totally fine now! I don't feel like shuffle() and random() need to be tested, but if you do decide to test them, at the start of each test case, use `MathUtils.random.setSeed(12345);` (or some constant long seed other than 12345, if you find one you want to be reproducible). Using the same seed every time means there's no chance of really rare output sequences that appear to be statistically unlikely happening on one test run but not most test runs. It's a good idea to set the seed at the start of the test case, but run the shuffle() or random() methods many times, like 1000, and just make sure each item in a small array appears at least once in the first and at least once in the last position (for shuffle) or just that each item is returned at least once (for random). You shouldn't check that the output always is the same for a given seed, since future versions might change the algorithm MathUtils.random employs (it is currently a RandomXS128).

gdx/jni/iosgl/iosgl20.h
Frosty-J, line 1: If this is to be merged, a contributor must remember to fix these huge diffs first.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class LongArrayTest {
+	/** Test of the different adding methods */
+	@Test
+	public void addTest () {
+		/*
+		 * Test of the classic add(long value) method, it should be adding a number at the first available place.
```
Frosty-J, line 27: At risk of repeating what I said at https://github.com/libgdx/libgdx/pull/7514#discussion_r1845546463, the comments throughout this seem more plentiful and verbose than they need to be. It's irrelevant whether a method is classic or otherwise, and there's a lot of "Test of the ... method" in here stating the obvious. Since multiple tests are performed in this test, it's not unreasonable to break it up with comments, but many would argue that unit tests are supposed to be as small as possible. Since they are run automatically, it doesn't matter much if there's loads of them.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+		Assert.assertArrayEquals(new long[] {1, 2}, longArray2.toArray());
+
+		/*
+		 * Some test with the addAll(LongArray array) and addAll (long... array) methods. The second call should resize longArray3
+		 * to a size of 17.
+		 */
+		LongArray longArray3 = new LongArray();
+		longArray3.addAll(longArray2);
+		Assert.assertArrayEquals(longArray2.toArray(), longArray3.toArray());
+		longArray3.addAll(longArray1);
```
Frosty-J, line 47: I find this quite difficult to read and relying on the results of the previous tests weakens this. I would include `{3}` and `{1, 2}` directly in this test. obigu might agree? https://github.com/libgdx/libgdx/pull/7518#discussion_r1851716120

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+		/*
+		 * Expected hash calculation: 1. Start with h = 1 2. For each value in the array, compute (item ^ (item >>> 32)) and add it
+		 * to h as: h = h * 31 + (int)(item ^ (item >>> 32)) Detailed steps: - Iteration 1 (item = 10): h = 1 * 31 + 10 = 41 -
+		 * Iteration 2 (item = 20): h = 41 * 31 + 20 = 1 291 - Iteration 3 (item = 40): h = 1 291 * 31 + 40 = 40 061 - Iteration 4
+		 * (item = 50): h = 40 061 * 31 + 50 = 1 241 941 - Iteration 5 (item = 100): h = 1 241 941 * 31 + 100 = 38 500 271 -
+		 * Iteration 6 (item = 200): h = 38 500 271 * 31 + 200 = 1 193 508 601 - The next operation will overflow because h is an
+		 * int and can't be calculated if >= 2 147 483 647 (because negative numbers are defined by the first bit - The result of 1
+		 * 193 508 601 * 31 + 400 is 36 998 767 031 (1000 1001 1101 0100 1100 0110 0001 1011 0111) which is, truncated to 32 bits,
+		 * -1 655 938 633 (1001 1101 0100 1100 0110 0001 1011 0111) - Iteration 7 (item = 400): h = - 1 655 938 633 - Same here,
+		 * without redetailing the operation, we get, truncated to 32 bits, 205 510 429 - Iteration 8 (item = 500): h = 205 510 429
```
Frosty-J, line 203: I understand Spotless probably did this, but I find it very difficult to read as the `-` used for bullet points ends up looking like mathematical subtraction. This looks like a complete sentence on first read: `Start with h = 1 2.` You may wish to consider using `<ul>` and `<li>` to create a formatted list. If this was generated with AI and you haven't gone through it yourself, I urge you to just throw it away.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+	/** Test of the hashCode() method */
+	@Test
+	public void hashCodeTest () {
+		/*
+		 * Testing if the hash code of an ordered long array is in fact the same provided as the hashCode() method from Object.
+		 */
+		Object longArrayUnordered = new LongArray(false, 16);
+		((LongArray)longArrayUnordered).add(2, 1, -6, 2);
+		((LongArray)longArrayUnordered).add(5, 50, 34, 42);
+		Assert.assertEquals(longArrayUnordered.hashCode(), ((LongArray)longArrayUnordered).hashCode());
```
Frosty-J, line 188: This assertion is useless. `LongArray` implements `hashCode` - you are comparing the same result in each case. Try it - I replaced the `LongArray` implementation with `return 1` and this part still passed.

Frosty-J (changes requested):
I disagree with TEt. I thought he liked hash codes. My concerns are a mix of subjectivity and objectivity.

gdx/jni/iosgl/iosgl20.h
obigu: Ok, it was a mess but I think I've fixed it

Frosty-J: Thanks. @melvilpf you should `git pull` this before making any further changes, otherwise I don't think it'll be able to push them due to the rebase against `libgdx:master`. I could be saying some wrong things but you get the memo.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
tommyettinger: Honestly any test on hashCode() is probably pointless. I say get rid of the hash code test.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
tommyettinger: Also, for one thing testing hash codes this way doesn't actually provide any validation of any properties of a hashing function... for another thing, hashCode() impls in Java are usually quite simple, and that's fine for their typical usage in hash tables. The main requirement for a hash function to use in a hash table is just for the function to be fast to call (as in, a small bytecode size); most String keys aren't novel-length texts, and a LongArray used as a key in a HashMap probably won't have 1000+ items, so the actual performance of the function per item doesn't matter as much as if it can be inlined early, and the quality can be actually quite bad and still work fine in a hash table.

I think the whole hashCodeTest() should be removed.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+			Assert.fail();
+		} catch (IndexOutOfBoundsException e) {
+			// We should get here as we want to get an index that's out of bounds.
+		}
+	}
+
+	/** Test of the set() method */
+	@Test
+	public void setTest () {
+		LongArray longArray = new LongArray(new long[] {3, 4, 5, 7});
```
tommyettinger, line 79: All these usages of the LongArray constructor that takes a `long[]` could be a good amount cleaner by using `LongArray.with(3, 4, 5, 7)` . It isn't at all needed, but I feel like with() should at least be known about, and used when possible.

gdx/test/com/badlogic/gdx/utils/LongArrayTest.java
```diff
@@ -0,0 +1,425 @@
...
+		Assert.assertTrue(longArray1.equals(longArray4));
+
+		// Standard verification that two arrays with different content are not equal
+		longArray1.add(3);
+		Assert.assertFalse(longArray1.equals(longArray2));
+	}
+
+	/** Test of the toString() methods */
+	@Test
+	public void toStringTest () {
```
tommyettinger, line 417: I'm also reiterating Frosty-J's concern about the toString format being flexible to change; we might want to switch to `{}` instead of `[]` to work as an array literal, and append `L` to each number so this could go into source code. I don't think that's terribly likely, but it shouldn't be codified in a test.

tommyettinger (changes requested):
I'm mostly requesting that `hashCodeTest()` be removed, since I don't feel like the ways `hashCode()` can be tested actually help the class. I also feel like `toStringTest()` can be removed.

melvilpf:
Okay, so I'm getting rid of hashCodeTest() and should I also get rid of the toString() tests ? I will also be using at least once LongArray.with(...), is that it ?

tommyettinger:
> Okay, so I'm getting rid of hashCodeTest() and should I also get rid of the toString() tests ? I will also be using at least once LongArray.with(...), is that it ?

Sounds good to me. `LongArray.with()` is just a good API design IMO, one that the actual JDK didn't even have until Java 10 or 11 (where I think it's `List.of()` for some form of List implementation). I feel like using it to create a LongArray makes sense when you know the initial contents right away. Internally, it's essentially the same, a `long[]` is created automatically for `with()`, or explicitly with the constructor you have now, and it gets passed to the same constructor that takes a `long[]` . Using `with()` tends to be shorter and look cleaner, though it does hide that an array is being created.

It does make sense to at least lightly test `with()`, because some of these varargs methods have corner cases that should be checked against, like how you are allowed to do `LongArray.with(null)`, which is passing a `long[]` that is `null`, and it gets to the constructor before failing when it tries to get the `.length` of `null`. For object items, what should `Array.with()` (giving it no arguments) return? Currently, because it has no item to figure out what `T` is in `T...`, it returns an `Array<Object>` .

gdx/jni/iosgl/iosgl20.h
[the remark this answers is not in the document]
melvilpf: oh no did it ? ://
I git pulled though that's so weird, and it doesn't appear in my commits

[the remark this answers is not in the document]
obigu: My advice not to waste time with it is, close this PR, which is a mess, and create a new clean one with all the changes from comments already applied.

melvilpf:
Okay, I'm closing this pull request and opening a new one, thanks for the help

Issue #7115, opened 2023-04-03
Feature request: improve GWT reflection lib over Method Modifiers.


patton73:
This is a feature request regarding Method Modifiers. 
The class [Method in package](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/reflect/Method.java)  does not take care of Synthetic and Bridge methods. 
It should be added at least `boolean isSynthetic()` and `boolean isbridge()` methods to cover are possible Method's modifiers.
It would be even better if can be added a method `int getModifiers()` so that it would be simpler to cope with method's modifiers.

Please take in consideration to implement this feature in the near future.
Thanks in advance.

leschli:
is this still available?

PR #7517, opened 2024-11-20
Feature request: improve GWT reflection lib over Method Modifiers. #7115

Hugo801104:
The methods requested in the issue have been implemented


tommyettinger (changes requested):
This won't help on GWT one bit! The reason GWT doesn't support this is that https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gwtref/client/Method.java doesn't support it, and that is used by https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/com/badlogic/gdx/utils/reflect/Method.java , which is the actual Method implementation on GWT. The class you changed won't even be used by GWT, but if code tried to call any of the methods you added on GWT, it wouldn't compile because its Method version doesn't have that new code. I'll check and see if GWT even has existing code that can identify a synthetic or bridge method, though I doubt that. Those keywords might just do nothing on GWT, like strictfp.

tommyettinger:
Looks like synthetic isn't known to GWT normally, though it might be possible to add it during the non-GWT-based steps that generate the reflection cache. https://www.gwtproject.org/javadoc/latest/search.html?q=synthetic

Bridge doesn't fare any better. https://www.gwtproject.org/javadoc/latest/search.html?q=bridge It refers to a different concept in GWT.

PR #7519, opened 2024-11-20
fix duplicate points returned in polygon intersection

### LINKED ISSUE #7408: Duplicate points returned in polygon intersection
The following code intersects a square with a triangle, where one of the triangle edges passes directly over a corner of the square. The overlap polygon is returned with 5 vertices, one of them duplicated, whereas the expected overlap is a square with 4 vertices (the lower left sub-square of the original square). The code then continues to intersect the overlap with two triangles. Neither of these intersections return any overlap, though they clearly do overlap and the expected overlaps are two triangles. The erroneous overlap with the duplicate vertex causes further intersection issues.

```
Polygon p1 = new Polygon(new float[]{1,-1,2,-1,2,-2,1,-2});
Polygon p2 = new Polygon(new float[]{0.5f,-1.5f,1.5f,-1.5f,1.5f,-2.5f});
Polygon overlap = new Polygon();
Intersector.intersectPolygons(p1,p2,overlap);
System.out.println("Number of overlap vertices: "+overlap.getVertexCount());
tmpSB.setLength(0);
VectorMath.poly2String(overlap,tmpSB);
System.out.println(tmpSB.toString());
Polygon p3 = new Polygon(new float[]{1,-1,2,-1,1,-2});
Polygon p4 = new Polygon(new float[]{2,-1,2,-2,1,-2});
Polygon overlap2 = new Polygon();
Polygon overlap3 = new Polygon();
Intersector.intersectPolygons(p3,overlap,overlap2);
Intersector.intersectPolygons(p4,overlap,overlap3);
System.out.println("Number of overlap2 vertices: "+overlap2.getVertexCount());
System.out.println("Number of overlap3 vertices: "+overlap3.getVertexCount());
```

The output of the code fragment is shown below:

```
Number of overlap vertices: 5
 1.0,-2.0 1.0,-1.5 1.5,-1.5 1.5,-2.0 1.0,-2.0
Number of overlap2 vertices: 0
Number of overlap3 vertices: 0
```

jose-puente:
fix #7408

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/Intersector.java
+++ b/gdx/src/com/badlogic/gdx/math/Intersector.java
@@ -198,8 +198,8 @@ public final class Intersector {
 			floatArray.clear();
 		}
 		// Check for 3 or more vertices needed due to floating point precision errors
-		if (floatArray2.size >= 6 && floatArray2.get(0) == floatArray2.get(floatArray2.size - 2) && floatArray2.get(1) == floatArray2.get(floatArray2.size - 1))
-			floatArray2.setSize(floatArray2.size - 2);
+		if (floatArray2.size >= 6 && floatArray2.get(0) == floatArray2.get(floatArray2.size - 2)
+			&& floatArray2.get(1) == floatArray2.get(floatArray2.size - 1)) floatArray2.setSize(floatArray2.size - 2);
 		if (floatArray2.size >= 6) {
 			if (overlap != null) {
 				if (overlap.getVertices().length == floatArray2.size)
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/math/Intersector.java
```diff
@@ -198,6 +198,8 @@ public static boolean intersectPolygons (Polygon p1, Polygon p2, Polygon overlap
 			floatArray.clear();
 		}
 		// Check for 3 or more vertices needed due to floating point precision errors
```
obigu, line 200: I would leave this comment where it was before (over the line "if (floatArray2.size >= 6) {") and add a comment for this block here along the lines of "Ensure first and last point are different to protect agains floating point precision errors".

obigu (approved):
Looks good to me, thanks!

Not sure why it says there are 2 workflows that need approval though...

PR #7520, opened 2024-11-20
AlignTest / SortTest

JujuG02:
I made tests for the classes Align and Sort. 
For Align, I didn't make a test for toString, but I can do so if you think it's relevant!


Frosty-J:
Strange wrapping in this one. libGDX's source doesn't have an 80-column limit. I would either stick with the old `assertArrayEquals(new Integer[] {1, 1, 2, 3, 3, 4, 5, 5, 5, 6, 9}, array)` or move the `Integer[]` into its own variable instead of wrapping mid-array. There's an unusually high number of blank lines at the end of the classes, too (one would be normal for Unix-y reasons, which then gets hidden on GitHub.com). It's quite likely Spotless will change these things anyway - I haven't tried running it on this.

Some of the comments from #7514 still apply (copyright header capitalisation changed, verbose comments, to name a couple).

JujuG02:
I changed SortTest and made simpler comments on the AlignTest class.

gdx/test/com/badlogic/gdx/utils/SortTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertArrayEquals(new Integer[] {1, 1, 2, 3, 3, 4, 5, 5, 5, 6, 9}, array);
+    }
+
+    @Test
+    public void testSortArrayWithComparator () {
+        Integer[] array = {3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5};
+        Comparator<Integer> comparator = new Comparator<Integer>() {
+            @Override
+            public int compare (Integer o1, Integer o2) {
+                return o1.compareTo(o2);
```
Berstanio, line 52: If you want to test custom comparator usage, you would need to test with a non-default comparator.
Otherwise you don't test at all, whether a non-default constructor works and it is the same as `testSortArrayComparable`

Frosty-J:
Thinking about this some more, there is no test coverage for `Align.topLeft` and other combinations. Personally, I still don't see the value in writing unit tests for `Align`, as it is such a simple class, but it's just a thought.

gdx/test/com/badlogic/gdx/utils/SortTest.java
JujuG02: Okay ! I made custom comparators then

gdx/test/com/badlogic/gdx/utils/AlignTest.java
```diff
@@ -30,38 +30,71 @@
...
+        assertFalse(Align.isCenterHorizontal(Align.topRight));
+        assertFalse(Align.isCenterHorizontal(Align.bottomLeft));
+        assertFalse(Align.isCenterHorizontal(Align.bottomRight));
+    }
+
+    @Test
+    public void testCombinations () {
+        assertTrue(Align.isLeft(Align.bottomLeft));
+        assertTrue(Align.isRight(Align.bottomRight));
+        assertTrue(Align.isTop(Align.topRight));
```
Frosty-J, line 92: This test has an awful lot of overlap with the prior tests. `assertTrue(Align.isRight(Align.topRight))` etc. are duplicate lines.

gdx/src/com/badlogic/gdx/utils/Align.java
```diff
@@ -42,9 +42,7 @@ static public final boolean isTop (int align) {
 		return (align & top) != 0;
 	}
 
-	static public final boolean isBottom (int align) {
-		return (align & bottom) != 0;
-	}
+	static public final boolean isBottom (int align) { return (align & bottom) != 0; }
```
Frosty-J, line 45: I'd discourage pushing any changes to other classes unless there is something in them, however minor, that needs fixing. Spotless will revert this anyway (run the `gdx:spotlessApply` Gradle task if you want the CI/CD checks to pass) but it won't revert the removal of trailing whitespace in the copyright header, if I'm not mistaken. I don't know whether there's a consensus on if it should have the trailing whitespace or not, but I wouldn't have that as the sole change in any case (clutters the commit history for no gain).

gdx/test/com/badlogic/gdx/utils/AlignTest.java
JujuG02: I will delete the duplicate lines ! Thank you

gdx/src/com/badlogic/gdx/utils/Align.java
JujuG02: Sorry, I didn't mean to commit these changes and it wasn't intentional. I don't master Git yet.

Frosty-J: Don't worry about it too much. Sometimes I limit what I add to reduce the chances of accidentally including other stuff, e.g. `git add gdx/test` instead of `git add .`, sometimes I don't.

PR #7524, opened 2024-11-24
A few modifications to LongArray and a test class for it (clean)

melvilpf:
I modified three methods of the LongArray class:

- lastIndexOf took a char argument when it was supposed to be a long since we are working with long arrays ;
- pop() and peekl() did not handle the case of empty arrays so I added this case.
- I then implemented a test class for LongArray, testing all the methods I could (of course I can't test random() for example).

This is a pull request that I have closed and renewed so that it is cleaner.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/LongArray.java
+++ b/gdx/src/com/badlogic/gdx/utils/LongArray.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- * 
+ *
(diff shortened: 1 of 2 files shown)


Frosty-J:
Okay, I don't have any major concerns with this version.

tommyettinger (approved):
This also looks good to me. I am glad there's a check in pop() before it might set size to -1 otherwise. The lastIndexOf() change is excellent. I'm glad at least three tests check with() to make sure it isn't broken in some unexpected way. Approved!

Issue #3349, opened 2015-08-13
TilemapPacker doesnt work for some tilesets


Vichy97:
TilemapPacker always crashes when you try to pack a tileset that is made from a collections of images instead of one tileset image. It crashes at line 274. Normally I would use a single tileset image but it annoying when you have lots of odd shaped tiles and even when you do it still doesn't work. (wrong tiles are rendered most of the time)

MobiDevelop:
Don't use the TiledMapPacker for "Collection of images" tilesets, it wasnt designed to work with them. Just use TexturePacker and add an atlas property to your tmx.

PR #7525, opened 2024-11-25
TiledMapPacker Tool Enhancement, ImageLayer and Json map Support, + crashfix.

BoBIsHere86:
I know my PRs can get a bit long winded. But I'm hoping to build a case for this enhancement and see what you guys think. As well as describe in a bit more detail for those who are not familiar with Tiled.
Any advice, alterations, comments are welcome.

This will be broken up into 2 Parts.
**First, ImageLayer support:**
I've long ignored the AtlasTmxMapLoader as I never had any reason to use it, as well as not quite understanding it. Most of the maps I build consist of a single tileset anyways, so I left it at that.
Until I thought about ImageLayers, a very useful feature in tiled, allowing for an extremely simple way to setup nice looking backgrounds with parallax effects, img repeating, tinting, etc. But currently, rendering each individual imagelayer image causes a draw call during rendering. It's not that big of a deal, but still kind of annoying and it can add up fast.

Here is a gif of my example map, which uses 3 images layers to create a very simple effect. 2 layers in the background and 1 in the foreground. This results in 4 draw calls as you can see below.
normalImageLayerMap_small 
testmap_assets_s

Now imagin if we could pack those into the same atlas as our tilesets and drop it down to a single draw call for the entire map rendering call. Well, we don't need to imagine. Here it is:
atlasImageLayers_small
testmap_assets_packed_s

Ok so first, let's not get ahead of ourselves.
As of right now you actually can't even use imagelayers in ANY map being loaded using AtlasTmxMapLoader and I don't know how this has gone unnoticed for so long. Am I the only one who uses Image Layers in tiled? Or maybe no one uses the AtlasTmxMapLoader?
Anyways, the issue:
So if you use the TiledMapPacker to pack your map tilesets you now need to use the AtlasTmxMapLoader to use them.
Great, right? Well if your map happens to make use of imagelayers at all, it just crashes. See below:

> Exception in thread "main" java.lang.NullPointerException: Cannot invoke "com.badlogic.gdx.graphics.g2d.TextureRegion.getRegionHeight()" because "texture" is null
	at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadImageLayer(BaseTmxMapLoader.java:291)

Well, what's going on here?
So in the loadImageLayer() method of the BaseTmxMapLoader class, we call an ImageResolver to get the image.
image
The problem though is that when using AtlasTmxMapLoader, it has it's own special image resolver. Which is expecting a region name to check the atlas for. Sadly, when this runs instead we pass in something like "/(assets root)/some_background_img.png"
Which of course, then comes back as null, which then throws the exception on texture.getRegionHeight();

So how do we solve this? Well, the quickest way to just "make it not crash" would be to just drop an 'instanceof' check and ignore it when AtlasResolver's are used. But that would be pretty janky, right?
Or... Why not just add support for packing image layers into the atlas! So that's what I set about to do.

The last time the TiledMapPacker was touched was years ago, so it's pretty legacy. When making my changes I did my best to not mess with too much.
In it's currently implementation after the changes I have made to it. It will automatically detect if there is an image layer in the map file and pack it along with the tilesets, otherwise it packs the tilesets like normal.
If this is unacceptable I could always change it to make it one of the optional arguments instead. But if you are using image layers in your map and using the map TiledMapPacker, it just wont work anyways so I thought it might as well detect it automatically.
Some thoughts on what I did.
If you look through the code you may wonder why I made a generateUniqueImageName() method and why I modify the image names in the map files.
While testing this I came across certain outlier situations that I wanted to be prepared for. If a user had images in multiple folders/sub-folders, with the same exact name(for some reason) but different images. This would cause an issue when generating a region name for the atlas based on the image's filename.
Now there is still one more problem. Even if every image had it's own unique name. We would still run into the same crash as before. Why? Because when passing the handle.path() it also passes along the root assets folder in the string as well. So when it checks the atlas, it still won't be found and will return null.
So my thought's were to prepend an identifying string to packed image region names. This being "atlas_imagelayer".
So I then modified the AtlasTmxMapLoader to check for this string to determine we we are looking for an image that's been packed and if so strip out the folder structure leaving us with just the region name. If not, it processes it just like any other tileset region name.
Doing this keeps all of the changes localized to the AtlasTmxMapLoader and packer related classes without the possibility of messing up any other part of the tiledmap code.

I have also gone ahead and modified the tiled atlas test with a new map packed with imagelayers so you can toggle between the 2 maps. You can see the test below:
newtest_small

**Extra Note:**
You will notice I did add a small gradle task to build the TiledMapPacker into a runnable jar. This was the easiest way I found to build the code after I modified it. If this is not acceptable or un-needed. I can remove it. I did NOT however, add it to the buildRunnables section with dist3DParticles, distHiero, distTexturePacker. As I'm not sure if there would be consequences for that.

So all this being said, if this were implemented I would probably start using the AtlasTmxMapLoader from this point forward to save on those sweet, sweet draw calls.

Good idea, bad? Waste of time? Let me know.

**Second Part: JSON/TMJ Support**
With the advent of support for the .tmj format). I thought, why should .tmx files have all the fun? So I went ahead and added support with a new AtlasTmjMapLoader as well as modifying the TiledMapPacker to ingest and update .tmj files just like we do with .tmx files.

So this is where I may need some input. As it is now. It just works automatically, looking for either .tmj or .tmx files. If you would rather this be an explicit command line argument option. I can do that instead.
Through my testing so far everything seems to work as it should. But THERE is 1 issue that I came across. But I'm not sure if it's worth looking into since I don't know why a user would be doing this.
Lets say you have a directory like this which includes about 6 maps, 3 tmx and 3 tmj. If the tmx and tmj files have the same "name" (minus the extension) you run into an issue where the packer throws an exception because it uses the map name when its creating regions
image
You'd get an error like this:
image
One way to go about fixing this could be to check for duplicate names when handling tmx and tmj together and alter the name by incrementing it. But then we are now altering the user's file names when they could just alter them to what they like instead, we could provide a warning. Making sure not to have maps with the same filenames. Anyways in doing so you end up with a nice compact output for each of your maps like this.
allpacked

One last thing to note. When making the AtlasTmjMapLoader, I didn't want to make any breaking changes. So we end up with a bit of duplicate code between it and the the AtlasTmxMapLoader since has the protected interface AtlasResolver which has static classes which have public methods, so I wasn't sure what choice to make. I'd rather have duplicate code then refactor anything and break something for anyone who may be extending it.

Advice is welcome,
Also I am not some TexturePacker expert. I just tried my best to learn as I went along going through the TiledMapPacker. So If I made any obvious mistakes, let me know.

@tommyettinger I'm pinging you on this because I know that you maintain your own version of the TiledMapPacker. Knowing this I actually had a local copy of your repo and was making changes side by side. So if anyone actually thinks this PR is worth it, once we work the kinks out I can send a PR
That being said. It's not 1 to 1 yet. I had some problems working with the PixmapPacker, instead of the TexturePacker. The Texturepacker seems to handle duplicate images while the pixmappacker doesnt, and I wasnt too sure how to work with the different packing strategy's. While it does work, the end results was not very tightly packed.


tommyettinger:
I could probably update my split-off TiledMapPacker code; maybe it can use the "full" (desktop-only) TexturePacker code. Part of the reason I have a smaller split version of TiledMapPacker is so the JAR doesn't include all the sometimes-large files needed by the whole of gdx-tools. I don't know if TexturePacker is much size to add, but it does restrict the platforms to the ones with AWT.

I think your name unique-ifying code just needs to reuse the existing option in TexturePacker to not remove paths from filenames; you can't have two files with the same name in the same folder, barring case insensitivity weirdness.

If a user expects that having a .tmx file and a .tmj file having the same name without extension will act normally, when there's already the convenience feature where you don't specify an extension, they deserve whatever unpleasantness comes to them. The Tiled map format doesn't even need to have different extensions for .tmx and .tmj; there are ways in XML to indicate that the file uses a specific schema and encoding, which won't be present in any JSON file. If they had gone with one file extension, like .tmap, then this issue wouldn't even be possible.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
```diff
@@ -345,15 +440,68 @@ private void packTilesets (FileHandle inputDirHandle, Settings texturePackerSett
...
-		File relativeTilesetOutputDir = new File(tilesetOutputDir);
-		File outputDirTilesets = new File(relativeTilesetOutputDir.getCanonicalPath());
+	}
 
-		outputDirTilesets.mkdirs();
-		packer.pack(outputDirTilesets, this.settings.atlasOutputName + ".atlas");
+	// Method to generate a unique image name
+	private String generateUniqueImageName (String imageSource) {
+		String baseName = new FileHandle(imageSource).nameWithoutExtension();
+		String uniqueId = UUID.randomUUID().toString().replaceAll("-", "").substring(0, 8);
```
tommyettinger, line 448: As far as I understand, libGDX tries to avoid UUID wherever it can appear because creating a random UUID requires instantiating a SecureRandom instance at app startup, and that can be considerably slower than you might expect! UUID also isn't present on GWT, though that doesn't matter here, and I believe some security settings can prevent SecureRandom from being created at all. To me, the best solution is a `long` counter, maybe initialized to `System.nanoTime()` once and incremented after with `++counter;`, since that's actually more collision-proof than 8 digits of a UUID. `++counter;` can only collide if the initial value is somehow close in an earlier run and the earlier run's names are still in use, or if over 18 quintillion files are named, which just can't happen... A 64-bit secure-random value will collide with another such random value 50% of the time once about 4.2 billion files are hashed, but the odds are still non-zero even if only 2 files are hashed.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
```diff
@@ -20,7 +20,9 @@
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.UUID;
```
tommyettinger, line 23: If you do remove UUID usage in the other commented block, please also remove it here. Thanks!

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapAtlasAssetManagerTest.java
```diff
@@ -45,6 +47,8 @@ public class TiledMapAtlasAssetManagerTest extends GdxTest {
 	private SpriteBatch batch;
 	String errorMessage;
 	private String fileName = "data/maps/tiled-atlas-processed/test.tmx";
+	private String fileNameWithImageLayers = "data/maps/tiled-atlas-processed/test_w_imglayers.tmx";
```
tommyettinger, line 50: We have a test for .tmx files, that's great, but a lot of this PR handles .tmj files and there isn't a test for that, that I see so far? How hard would it be to make a .tmj version of the .tmx test?

tommyettinger (changes requested):
Mostly looks quite good! I think avoiding UUID would be a good idea for startup times, especially with gdx-tools often used from Gradle scripts that don't have time to JIT-compile, and the solution is pretty easy. The tests are great, but I do feel if it's possible to add a .tmj test, we should have that in to make sure the TMJ loading code works.

BoBIsHere86:
Thanks for taking the time to look at this @tommyettinger. I'll have a look at all the requested changes in a bit.

So the reason I suddenly updated this PR was because I was trying to help @Quillraven out. He was looking to see if we could get this compatible with the updates to his PR which added support for classes. #7534. So I made some new changes for him.

I kind of left this PR for dead once I saw that "TextureArraySpriteBatch" PR that's been around for quite while. I realized that might make this irrelevant. But after the discord discussion I thought. Well, maybe that "TextureArraySpriteBatch" PR never ends up getting finished. 
So I guess it's good to keep this updated I don't know, I guess it's helpful in the meantime. Sadly there's not much documentation on the AtlasTmxMapLoader . So maybe if we do update it I can add something to the wiki one day.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
BoBIsHere86: Thanks I had no idea about the UUID stuff. Honesty choosing UUID was just kind of overkill. If I remember correctly, I just needed a way to handle an outlier case where if someone happened to be using image files with the same name across different folders. I wanted to give them a short unique name to identify them in the atlas without resorting to adding long folder names and slashes.

I'll try and implement something from your suggestion

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapAtlasAssetManagerTest.java
BoBIsHere86: Not hard at all. I'll get back to ya when it's all done.

BoBIsHere86:
I just realized the NEW changes I originally meant to push for Quillraven. Were not committed when I did my push. So I basically just pushed a rebase.
What I MEANT to push was all these OTHER new additions that just popped up.

tommyettinger:
You've been very solid at getting your PRs in great condition so far, I have no reason to doubt that your changes won't get to a nice mergeable state here too! Sorry this PR in particular has taken over 4 months to make headway in terms of reviews and such...

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
```diff
@@ -85,17 +93,23 @@ public class TiledMapPacker {
 	private HashMap<String, IntArray> tilesetUsedIds = new HashMap<String, IntArray>();
 	private ObjectMap<String, TiledMapTileSet> tilesetsToPack;
 
+	private ObjectMap<String, Array<String>> imagesLayersToPack;
+	private ObjectMap<String, String> imageLayerSourceFiles;
+
 	static File inputDir;
```
Quillraven, line 99: I think that package private is a bad option for `inputDir` and `outputDir` because it is necessary information for the `processInputDir` method and cannot be adjusted from outside.

I get it that this tool is intended to be exported as a jar and then run from command line?

But I personally would prefer if it can also be run from a user's gradle task, which I was trying to do yesterday (or actually just be called from any arbitrary Java code).

For that, it would be great if we either make those fields public OR provide a a second `processInputDir` method that takes all necessary arguments of this tool, which is:
- inputDir
- outputDir
- projectFilePath

tommyettinger: You should be able to call the main() method with String arguments, the same as running the JAR. Is there something that can't be done using a call to the TiledMapPacker's main() from your Gradle task?

Quillraven: Hm I can't remember. I just remember I created PackerSettings and MapPackerSettings and then the packer instance itself. Calling "main" is weird imo but it should work, you are right.

Although I still think it is cleaner to provide a method that shows the necessary arguments of the tool instead of main and some string args where you need to lookup which index is which argument.

BoBIsHere86: So this kind of brings up a side discussion.
The last time there were any functional changes to this class were back in 2015. Whenever I work on PRs especially for very old code. I try my best not to make too many changes outside of what my goal is. This code's been like that for a decade, so I'd worry of breaking something for an existing user.
BUT since it does seem we are willing to move forward with some changes to this class. I would be more than happy to help implement other features or cleanup. If there's more requests for change. As long as it doesn't stray too far from the original goal of the class itself. Just let me know.

tommyettinger: I think making anything that can be public, public, and documenting it at least a little, is a great step over package-private user-facing code. I am *really* opposed to package-private code in a library that is meant to be reused. It seems to me that the only decent reason to make something package-private rather than actually private is to privilege libGDX code in the same package, which is antithetical to what a reusable library should do. The other reason isn't even decent, it's being too lazy to even document a `protected` field as "internal use" or even "internal use only, may be removed at any point", and instead making an undocumented package-private field. I feel like if there is that sort of documentation there, it says more than just the variable being package-private, since user code can still jump through hoops, create an equivalent package, and use it, without the same knowledge that it might be removed.

Tom-Ski:
Change path to 
```
		String path = "data/maps/";

```
And set working directory as `.........libgdx\tests\gdx-tests-android\assets`

BoBIsHere86:
Alright, some updates coming through.

*Test Related Changes*
So a lot of this focuses on updating the testing code to now support all of the latest changes to we've map to the TiledMapPacker as well as the tiledmaps in general.

Firstly the tiledmappacker updates included support for json map processing. So I made sure to add in new maps which use the AtlasTmjMapLoader. You'll find a new TiledMapJsonAtlasAssetManagerTest specifically meant to load those "pre-processed' maps from the tiled-atlas-processed folder.

I updated the TiledMapPackerTest with additional enum parameter options to support the new features.
There are now up to 4 different maps that can be generated after running TiledMapPackerTest. To show off more variety of map times the TiledMapPacker can process. As well as an option to generate one which supported the recently added scenario of a map which uses a tiled-project file for custom class loading.

I also made relevant changes to the TiledMapPackerTestRender class to view the new maps that are generated in the deleteMeFolder.

*More TiledMapPacker Changes*
Stopped using UUID for image name generation.
Added a check for an edge case where a user might have a folder with regular maps as well as maps that need a tiled-project file. If that folder is processed without passing the path argument for that map. 
This scenario popped up in the TiledMapPackerTest. We have 4 map files we want to process, only one requires a tiled-project file. But for demonstration purposes we want to run the normal default command. Instead of crashing out, we just skip processing maps which require those arguments if they are missing and make note of it in the log.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
BoBIsHere86: I took your advice and tried to implement it. A heck of a lot simpler then a UUID.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
BoBIsHere86: Removed.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapAtlasAssetManagerTest.java
BoBIsHere86: Bunch of new testing related stuff was added as well as a specific test for the Json maps.

extensions/gdx-tools/src/com/badlogic/gdx/tiledmappacker/TiledMapPacker.java
tommyettinger: Oops, I resolved the wrong one. I think public or protected should be used instead of package-private, and if it is the type of thing that is expected to change in public API for any reason, document it that it may change without warning.

BoBIsHere86: So after diving through the old commits it seems that the "static File inputDir;" and "static File outputDir;" were only just added as member variables in the very last commit (before the formatting). 
They did exist in very early versions as private but then removed entirely and then added again years later. There doesn't seem to be any explicit reason why they were made package private. So it might not of even had a purpose. I see nothing wrong with making them public if this could help add some easy functionality for an end user like QuillRaven had requested. I mean the whole point of this thing is for people to use it. So whatever makes that easier.

I won't be making any changes yet though, until I hear back from Quillraven on any other comments from the recent updates. So we can lump them all together.

Quillraven:
I tried it very quickly today and get a nullpointer with my example map. I think the problem is the objects tileset which doesn'T have a tilewidth property or something like that.

Sorry I don't have a lot of time :( I will attach the example files to investigate.

```
Exception in thread "LWJGL Application" java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "com.badlogic.gdx.maps.MapProperties.get(String, java.lang.Class)" is null
	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.packTilesets(TiledMapPacker.java:453)
	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.processSingleMap(TiledMapPacker.java:246)
	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.processInputDir(TiledMapPacker.java:166)
	at com.badlogic.gdx.tiledmappacker.TiledMapPacker$1.create(TiledMapPacker.java:843)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:150)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:127)
```

[maps.zip](https://github.com/user-attachments/files/19673313/maps.zip)

BoBIsHere86:
> I tried it very quickly today and get a nullpointer with my example map. I think the problem is the objects tileset which doesn'T have a tilewidth property or something like that.
> 
> Sorry I don't have a lot of time :( I will attach the example files to investigate.
> 
> ```
> Exception in thread "LWJGL Application" java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "com.badlogic.gdx.maps.MapProperties.get(String, java.lang.Class)" is null
> 	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.packTilesets(TiledMapPacker.java:453)
> 	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.processSingleMap(TiledMapPacker.java:246)
> 	at com.badlogic.gdx.tiledmappacker.TiledMapPacker.processInputDir(TiledMapPacker.java:166)
> 	at com.badlogic.gdx.tiledmappacker.TiledMapPacker$1.create(TiledMapPacker.java:843)
> 	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:150)
> 	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:127)
> ```
> 
> [maps.zip](https://github.com/user-attachments/files/19673313/maps.zip)

Thanks for dropping the map. I can use this as a guideline. I'll take a look and I'll see what it takes to get this up and running.

BoBIsHere86:
@Quillraven 
Ok, so the issue is because that map is using a "Collection of Images" style of tileset for one of its tilesets (the one with the 2 images in the objects folder). The TiledMapPacker was never meant to support those kind of tilesets. There's a github issue from 2015 asking about it #3349 with a comment from MobiDevelop stating so.
I guess I could potentially look into adding support for that.

One of the reasons it breaks now is because in our (for example) TmxMapLoader we have the addStaticTiles() method and it only populates the fields needed for the TiledMapPacker to work when its a regular singular image based tileset. 
If you took a look at the method you'd see:
props.put("imagesource", imageSource);
props.put("imagewidth", imageWidth);
props.put("imageheight", imageHeight);
props.put("tilewidth", tilewidth);
..etc
But not when we are loading maps with "Collection of Images" type tilesets.

I could potentially look into modifying our core tile map loaders in the case where we have a collection of images tileset we can populate a individual TiledMapTile tile's properties to have those as well. We could add an overloaded addStaticTiledMapTile() method which is in the BaseTileMapLoader for that specific scenario to use. Like:
protected void addStaticTiledMapTile (TiledMapTileSet tileSet, TextureRegion textureRegion, int tileId, float offsetX,
		 float offsetY, String imageSource) {
		  TiledMapTile tile = new StaticTiledMapTile(textureRegion);
		  tile.setId(tileId);
		  tile.setOffsetX(offsetX);
		  tile.setOffsetY(flipY ? -offsetY : offsetY);

		  // Put the tile-level properties here:
		  tile.getProperties().put("imagesource", imageSource);
		  tile.getProperties().put("imagewidth",  textureRegion.getRegionWidth());
		  tile.getProperties().put("imageheight", textureRegion.getRegionHeight());

		  tileSet.putTile(tileId, tile);
	 }

And then when it comes to TiledMapPacker packTilesets() method, We can check if the tileset level info is missing the "imageSource" property and if that's the case the tileset we are iterating over is a Collection of Images. And then we could process it differently using the info that would be saved in each specific tile, based on the changes to the maploader.
(Of course this is all in theory, not sure exactly how animated tiles might work  and some other things)

So yea, we'd have to make a decision as to weather we want to support collection of images or not. I'm not sure if adding those properties to a TiledMapTile would have any breaking effects on code yet.

But I also, I guess it would be a cool feature to finally have support for it. 
But I also think it's fair to just tell people. Hey we don't support that. Just keep all your images in a single tilesheet like a normal human!

Quillraven:
Thanks for having a look at it. To explain my use-case:

I use "Collection of Images" tilesets for my game objects like the player, enemies, NPCs, etc.. Together with the class/enum/properties feature of Tiled I have a nice visual editor to configure my game objects. I also use a specific convention. The image name is they key of the atlas of that game object. The atlas is usually configureable via a property.

I then use those tiles on object layers to place my game objects.

Afaik LibGDX doesn't render object layers which imo makes sense, because it is a game specific thing and is treated differently in every game.

Also, I pack those graphics separately in sometimes different atlases (game objects, props, SFX, ...).

----

So for me personally I don't need the functionality that it packs those images together into an atlas. **HOWEVER** I do see the benefit and if it will work, I also have a use-case where this would be great.

Therefore, if we can go for your Option 2, that would be great. Not sure if we do it in this PR or a separate one but imo it makes sense to have it all combined in this PR.

Another idea/additional feature might be to define exclusions for the parser. Like in my case if we could say "ignore objects.tsx" because it is anyway treated separately, it would also be fine.

BoBIsHere86:
> Thanks for having a look at it. To explain my use-case:
> 
> I use "Collection of Images" tilesets for my game objects like the player, enemies, NPCs, etc.. Together with the class/enum/properties feature of Tiled I have a nice visual editor to configure my game objects. I also use a specific convention. The image name is they key of the atlas of that game object. The atlas is usually configureable via a property.
> 
> I then use those tiles on object layers to place my game objects.
> 
> Afaik LibGDX doesn't render object layers which imo makes sense, because it is a game specific thing and is treated differently in every game.
> 
> Also, I pack those graphics separately in sometimes different atlases (game objects, props, SFX, ...).
> 
> So for me personally I don't need the functionality that it packs those images together into an atlas. **HOWEVER** I do see the benefit and if it will work, I also have a use-case where this would be great.
> 
> Therefore, if we can go for your Option 2, that would be great. Not sure if we do it in this PR or a separate one but imo it makes sense to have it all combined in this PR.
> 
> Another idea/additional feature might be to define exclusions for the parser. Like in my case if we could say "ignore objects.tsx" because it is anyway treated separately, it would also be fine.

Ok I tried some stuff out and I now understand why Collection of Images was not supported. What a headache. lol
So Option 2. Only half-worked. I was able to pack all the images properly. BUT then we end up with an issue in the atlas files. The files in the atlas are given region names based on the tileset they are in and then given an index. In the case of this test map you gave me we end up with "objects" as the region name. That's fine and all. 
But when we try to load and run the map with the Atlas Map Loader, it will break with a "Tileset atlasRegion not found: objects/fox"
This is because when the maps are loaded with the atlas loader it uses the image source as the name to look for in the atlas. And since in this case we have "objects/fox" and "objects/sign.png" they wont be found. 
At first I thought, oh, that's easy I'll just make the atlas names match the image source. BUT of course the tiles do NOT contain any of that information. So there's no way during processing to do that. Which leads to the only option now being to begin modifying the .tsx and .tsj files we process to alter the image source names the atlas names. This now starts to open up another can of worms. Now we are modifying the tileset files as well and then also a special case for when the tilesets are saved internally to to the .tmx and .tmj files. Yucko!

Ok, so now Option 1 is looking A LOT more reasonable to me. I actually think simply adding a alternate method for "addStaticTiledMapTile()" into BaseTiledMapLoader specifically for when we process "Collection of Images" to just save some metadata in the tiles property like so:
```
protected void addStaticTiledMapTileForCollections (TiledMapTileSet tileSet, TextureRegion textureRegion, int tileId, float offsetX,
		 float offsetY, String imageSource) {
		  TiledMapTile tile = new StaticTiledMapTile(textureRegion);
		  tile.setId(tileId);
		  tile.setOffsetX(offsetX);
		  tile.setOffsetY(flipY ? -offsetY : offsetY);

		  // Put the tile-level properties here:
		  tile.getProperties().put("imagesource", imageSource);
		  tile.getProperties().put("imagewidth",  textureRegion.getRegionWidth());
		  tile.getProperties().put("imageheight", textureRegion.getRegionHeight());

		  tileSet.putTile(tileId, tile);
	 }
```
Ends up being a heck of a lot less work, i can then use this info to process the individual tiles and then I wont even need to deal with the TextureRegion to Buffered image Conversion anymore. So I think that's what I'm going to tackle next and see how it goes.

Also, I will try to throw in a new command line option as well if it ends up being needed.

BoBIsHere86:
Updates coming through! Let's hope they are acceptable.

We now have the beginnings of support for packing tilesets made up of a "Collection of Images".
It will just work by default. If we would rather NOT pack those by default to keep functionality the same. I could always add a parameter so a user has to explicitly pass something like --pack-collection in order to pack them instead. But I don't see the harm in it just working.

So the way I got around making any changes to the core Tiled Map codebase was to just create 2 nested subclasses of the TmxMapLoader and TmjMapLoader called PackerTmxMapLoader and PackerTmjMapLoader and doing a simple override.
These are strictly meant to be used inside TiledMapPacker and nowhere else. I explained it's purpose in the large comment I left.
Doing it this way saved me from modifying 5 classes in the core libGDX library. Now its all localized within TiledMapPacker to add this support.

I also fixed a bug i didn't realize was in here. I forgot to recursively check for image layers in groups. But that's fixed now as well.

My next goal is to test some more scenarios with different maps and then I plan to work on would be building a new test map which specifically uses a collection of images as a tileset for packing testing. 

And if there are any other comments and suggestions as well. Oh, and I just realized I forgot to change those package private variables. I'll get those next commit.

@Quillraven 
I got your map packed and loaded (i added the fox and sign as to a tile layer, that's why you see them, for demo purposes)
test1

tommyettinger:
Wow, good job at making this worry-free for users! I always prefer a transparent solution if possible, one that acts like the way things are already documented, so this is great. I'll review as soon as you get any more commits you've mentioned out of the way, but this looks great and the tests showing everything correctly is an excellent thing!

Quillraven:
You are fast - wow!

I try to have a look over the weekend and this sounds like I can now try a second use-case I had in mind. 

I can attach the files then again. Might be useful as a test map because it will contain all basic things plus some extra special cases.

Again, thanks for the effort! 👍

BoBIsHere86:
> You are fast - wow!
> 
> I try to have a look over the weekend and this sounds like I can now try a second use-case I had in mind.
> 
> I can attach the files then again. Might be useful as a test map because it will contain all basic things plus some extra special cases.
> 
> Again, thanks for the effort! 👍

Yes, feel free to send me whatever maps you'd like, more test cases for me to find any weird stuff.
*Side note while testing I did find a bug related to the parsing of custom classes within .tsj files.
You end up with: Exception in thread "LWJGL Application" java.lang.IllegalArgumentException: Named value not found: propertytype
        at com.badlogic.gdx.utils.JsonValue.getString(JsonValue.java:684)
        at com.badlogic.gdx.maps.tiled.BaseTmjMapLoader.loadProperties(BaseTmjMapLoader.java:443)
Because in the .tsj files unlike in .tmj's tiled doesnt seem to add "propertytype" keyword in there.  But that's unrelated to this PR. We can look into a fix at another time for that.
Anyways, off to sleep.
Cheers all!

BoBIsHere86:
My latest update contains some fixes and a new test.
Animated tiles did not work if they were made from the newly added Collection of Images support. I have fixed that and we now have a new test map which shows a map using regular tileset, CoI tilset, and animations from CoI tiles.
collectionRender

I have also added in the request that Quillraven asked for a new processInputDir() with additional params to pass in.

I did have to make 1 change to the core tileset files in BaseTmjMapLoader but this was sort of a fix that had to happen anyway. I changed createAnimatedTile() from private to protected. To match the method in the BaseTmxMapLoader. It should never have been private. (Also, i needed to override it anyway).

As always I will continue to test until I am 100% satisfied on this. Any suggestions encouraged.

BoBIsHere86:
Quick follow up question about general use of this tool.
So I think the biggest issue that I faced when even trying to begin exploring it was getting it to run.
Other than some documentation in the comments for example here:
* Windows <br>
	   java -cp gdx.jar;gdx-natives.jar;gdx-backend-lwjgl.jar;gdx-backend-lwjgl-natives.jar;gdx-tiled-preprocessor.jar;extensions/gdx-tools/gdx-tools.jar
	    com.badlogic.gdx.tiledmappacker.TiledMapPacker inputDir [outputDir]  [--strip-unused] [--combine-tilesets] [-v]
	 * <br><br> </code>
	 
There was very little documentation on how to actually use this along with the Atlas Map Loaders.
Which was why I ended up just adding my own gradle task to build it real quick and do all my testing.
Ex. ./gradlew :extensions:gdx-tools:distTiledMapPacker
I found this to be a super easy way to just get the jar and run it for all my testing purposes.

Now that's all good for me since I had the repo downloaded, working and built. But what about the average user. It seems like they need to scour the discord or tutorials to figure out how to use gdx-tools and this TiledMapPacker.

But if we go to the https://libgdx.com/dev/tools/, we can see direct downloads for the 3D Particle Effects tool, the Hiero and Texture packer tools. Along with their own wiki articles telling us how to use them.

Would it be ok to ask that we could get a section like this for the TiledMapPacker? I would be more then willing to put together a wiki article detailing how to properly use it and get it running with an Atlas map. But then we would also need to provide the jar.
I'm no gradle/pipeline/build expert. But in the current build.gradle for under gdx-tools we have

tasks.register('buildRunnables') {
	dependsOn("build")
	dependsOn dist3DParticles, distHiero, distTexturePacker
	doLast {
		println "Building ye runnables"
	}
}

Would we be able to add my distTiledMapPacker to this? And make the jar available directly to users in the future?

Anyways, that's the last of my comments for now. Everything is pretty much at a stopping point for testing. Probably no more code changes going forward.

tommyettinger:
So part of the reason I made a standalone repo is so the users don't need all of gdx-tools as a dependency. It includes a strangely large amount of unrelated garbage files if all you use is one of the simple tools... Like the Flame particle editor has a bunch of 3D files that are totally irrelevant here. I'll definitely update to use this code as soon as I can; my standalone repo was last updated for 1.12.1, I think. It might make sense to move that and my several other standalone repos to the libGDX org, which I don't think I can do on my own.

tommyettinger (dismissed):
Wow, 18,557 lines added, this one got big! I figure most of that is the Tiled map files, but still... Thanks for all the hard work on this, I am definitely approving! I do want this in some final state so I can bring the code over to my standalone repo that doesn't directly use gdx-tools as a dep. I figure that may be now or very soon!

BoBIsHere86:
Heh, yea 18,000 lines of xml and json test maps. lol

Anyways, I've given one final update to clean up some comment's. Updates/typos. 
I tested the new map Quilraven gave me without issue.
As well as a very messy, unoptimized map I threw together to show off those whole point of this update.
With about 8 image layers for parallaxing. 3 separate tilesheets consisting of normal ones as well as "Collection of Image" based ones.
We can turn this hideous 14 Draw calls per frame down to a glorious 1!
finalTest

So that's about it for now, I think this PR is pretty much done.

This is a small recap of what this PR is adding:
libGDX TiledMapPacker now supports:
1. JSON TileMap format.
2. TiledMap Image Layer Support
3. Fixed a Crash which would happen if a user attempted to load an atlas map which contained image layers.
(Basically in the old TiledMapPacker you could never use image layers in your map.)
4. Full support for "Collection of Images" tilesets. (Just because we now support it doesn't mean it's recommended to use them though, it's just good to have the option)
5. Support for TiledMaps which load custom classes.
6. Addition of a new overloaded public method for processInputDir() with additional parameters.
7. 4 new test maps have been added per new feature. Specifically for the TiledMapPackerTest and TiledMapPackerTestRender.
These maps are meant to try and cover various scenarios. XML and JSON maps. Maps containing regular tilesets,"Collection of Images" tilesets, Animations and support for loading tiled project files.

One thing which could be a "nice to have" but could added in the future would be easier access to a jar build of the tool through the main libGDX website. I'm sure new users would appreciate it. But I'll leave that to guys up top to decide.

Cheers all!

tommyettinger (approved):
Approved again!

Quillraven (approved):

Quillraven:
Shall we merge it? 😉

Tom-Ski:
@BoBIsHere86 thanks for the contributions! Don't personally follow with the latest stuff here in maps as not currently using any of it. But happy to merge this as you've worked in this area consistently and becoming our Tiled guru. :)

No one else seems to have any objections here, so I think lets merge this and hopefully get some more testers that might not be active on the tracker

Tom-Ski:
With the standalone runnable, you can hook it up to the task like the others in gdx-tools, and then just hook it up with an entry on the site and that should be it.

BoBIsHere86:
Sounds good. At some point soon I'll try and put together a wiki article describing how to use the TiledMapPacker. Once I get that done I'll throw in a small PR to add the task like the other gdx-tools and we'll go from there.

Thanks for the input everyone.

PR #7526, opened 2024-11-25
Versions maintenance

obigu:
Upgraded Android target SDK to `35` which requires upgrading Gradle version. After the upgrades Android tests were not running due to a dependency collision (`Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules kotlin-stdlib-1.8.10 (org.jetbrains.kotlin:kotlin-stdlib:1.8.10) and kotlin-stdlib-jdk8-1.6.21 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.21)`). Fixed by adding `org.jetbrains.kotlin:kotlin-bom` dependency to tests.

Also upgraded Android min SDK to `21` because `19` is no longer supported by Android X. This allows removing the explicit dependency to `multiDex`.

@tommyettinger Liftoff will have to upgrade accordingly at some point.


tommyettinger (changes requested):
I'm just requesting we build with the most recent Gradle (8.11.1); it does give nicer error readouts, which might help a bit when people have trouble building. Other than that, this works with (the latest) IDEA and almost assuredly with Android Studio, so I'm very close to approving!

tommyettinger (approved):
I don't know really the first thing about multidex and what it requires, so if I don't have to add it as a dependency in Liftoff projects that should be good. Good catch on the audio version checks! Gradle 8.12 is working seamlessly for me, and it seems a little faster for me sometimes as well, compared to some earlier 8.x versions I encountered in other people's projects. I'm glad we're using it here.

I don't know what entrypoint.sh does, or if it was intentionally committed even. It's empty, so I don't see it doing any harm on GH Actions, but I'm not sure why it's in this PR?

I'm approving! It looks like you searched for the older version codes and found+fixed the only place in the code that it mattered in. I'll have to maybe see what an example project looks like with and/or without the multidex changes needed here. That might be in the Android docs; I haven't checked yet.

Frosty-J:
`entrypoint.sh` isn't empty, it just doesn't have any diffs. Its file permissions are changed in this PR, basically to remove the executable bit. Doesn't Docker need it to be executable to execute it? We shall see!

tommyettinger:
@Frosty-J Huh. The diff said "Empty file." and I assumed it wasn't lying.

vivaldi_zvukTv3aB8

obigu:
My bad, goopd catch, I messed up the merge and Windows did the rest. Should be fixed now!

tommyettinger (approved):
Repeating approval!

PR #7528, opened 2024-11-25
Use doubles in DelaunayTriangulator::circumCircle

### LINKED ISSUE #7487: Broken Edge cases in DelaunayTriangulator.computeTriangles()
Hello!

I found some broken edge cases in the Delaunay Triangulator: DelaunayTriangulator.computeTriangles()

#### Issue details
These coordinates (xy pair as csv):
281.99274,473.00427,481.97058,126.60693,881.9485,126.5686,1081.9927,472.9276,882.0813,819.3633,482.10352,819.4784

Should produce only 4 triangles. Produces 9, and they overlap.
Screenshot_20241027_020211

Here is same set of points that produces correct solution:
287.99274,475.00427,487.97058,128.60693,887.9485,128.5686,1087.9927,474.9276,888.0813,821.3633,488.10352,821.4784
This correctly produces only 4 triangles, that do not overlap.
Screenshot_20241027_020633

I have many other data that can cause this. It seems to occur when vertices line up on axis.
Sorted is false.

#### Reproduction steps/code
https://github.com/0XDE57/DelaunayBugRepro

Enter the following 

1.13.0 and all versions. bug has been for a long time.

#### Stacktrace
n/a

- [x] Android
- [x] iOS
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] macOS

jose-puente:
fix #7487

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
+++ b/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
@@ -84,17 +84,16 @@ public class DelaunayTriangulator {
 			if (value < ymin) ymin = value;
 			if (value > ymax) ymax = value;
 		}
-        float dx = xmax - xmin, dy = ymax - ymin;
+		float dx = xmax - xmin, dy = ymax - ymin;
 
-        // Setup the super triangle, which contains all points.
-        float[] superTriangle = this.superTriangle;
-        superTriangle[0] = xmin - dx;
-        superTriangle[1] = ymin - dy;
-        superTriangle[2] = xmax + 3 * dx;
-        superTriangle[3] = ymin - dy;
-        superTriangle[4] = xmin - dx;
-        superTriangle[5] = ymax + 3 * dy;
-
+		// Setup the super triangle, which contains all points.
+		float[] superTriangle = this.superTriangle;
+		superTriangle[0] = xmin - dx;
+		superTriangle[1] = ymin - dy;
+		superTriangle[2] = xmax + 3 * dx;
+		superTriangle[3] = ymin - dy;
+		superTriangle[4] = xmin - dx;
+		superTriangle[5] = ymax + 3 * dy;
 
 		IntArray edges = this.edges;
 		edges.ensureCapacity(count / 2);
--- a/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
+++ b/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
@@ -85,8 +85,8 @@ public class DelaunayTriangulator {
 			if (value > ymax) ymax = value;
 		}
 		// Setup the super triangle, which contains all points.
-		float deltax = 100 * Math.nextAfter(Math.max(Math.abs(xmin), Math.abs(xmax)), Float.POSITIVE_INFINITY);
-		float deltay = 100 * Math.nextAfter(Math.max(Math.abs(ymin), Math.abs(ymax)), Float.POSITIVE_INFINITY);
+		float deltax = 100 * Math.nextUp(Math.max(Math.abs(xmin), Math.abs(xmax)));
+		float deltay = 100 * Math.nextUp(Math.max(Math.abs(ymin), Math.abs(ymax)));
 		float[] superTriangle = this.superTriangle;
 		superTriangle[0] = xmin - deltax;
 		superTriangle[1] = ymin - deltay;
--- a/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
+++ b/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
@@ -84,16 +84,16 @@ public class DelaunayTriangulator {
 			if (value < ymin) ymin = value;
 			if (value > ymax) ymax = value;
 		}
-		float dx = xmax - xmin, dy = ymax - ymin;
-
 		// Setup the super triangle, which contains all points.
+		float deltax = 100 * Math.nextUp(Math.max(Math.abs(xmin), Math.abs(xmax)));
+		float deltay = 100 * Math.nextUp(Math.max(Math.abs(ymin), Math.abs(ymax)));
 		float[] superTriangle = this.superTriangle;
-		superTriangle[0] = xmin - dx;
-		superTriangle[1] = ymin - dy;
-		superTriangle[2] = xmax + 3 * dx;
-		superTriangle[3] = ymin - dy;
-		superTriangle[4] = xmin - dx;
-		superTriangle[5] = ymax + 3 * dy;
+		superTriangle[0] = xmin - deltax;
+		superTriangle[1] = ymin - deltay;
+		superTriangle[2] = xmax + deltax;
+		superTriangle[3] = ymin;
+		superTriangle[4] = xmin;
+		superTriangle[5] = ymax + deltay;
 
 		IntArray edges = this.edges;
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
```diff
@@ -85,17 +85,15 @@ public ShortArray computeTriangles (float[] points, int offset, int count, boole
...
 
 		// Setup the super triangle, which contains all points.
 		float[] superTriangle = this.superTriangle;
-		superTriangle[0] = xmid - dmax;
-		superTriangle[1] = ymid - dmax;
-		superTriangle[2] = xmid;
-		superTriangle[3] = ymid + dmax;
-		superTriangle[4] = xmid + dmax;
-		superTriangle[5] = ymid - dmax;
+		superTriangle[0] = xmin - dx;
```
Berstanio, line 91: I'm wondering, does choosing a smaller super triangle result in better precision?
If yes, I use this code in my triangulator:
```
Vertex v = new Vertex((minx - 1), (miny - height));
Vertex u = new Vertex((minx - 1), (miny + height * 2));
Vertex w = new Vertex((minx + width * 1.5 + 1), (miny + height * 0.5));
```
which results in an even smaller rectangle.
But not sure whether my implementation has some quirks, just throwing it in here.

jose-puente: @Berstanio The truth is that I really don't know what's the impact of the supertriangle.
I just guessed that in terms of performance the closer you get to the points cloud the better. But it's just a guess.
In terms of rounding errors it seems it needs to be big enough to avoid them

Regarding your triangle, it's not always smaller. It depends on the size of the cloud. For clouds with sizes smaller than one I guess it will be bigger

In any case, with the last commit I included a new way of computing the supertriangle that I think it's better, as close a s possible to the cloud but leaving enough room to possibly avoid round errors

Berstanio: > Regarding your triangle, it's not always smaller. It depends on the size of the cloud. For clouds with sizes smaller than one I guess it will be bigger

Interesting, good to know, thank you! I will revisit my code then. I'm using integer positions in my code, so maybe thats why it worked for me.

I will also look at your new supertriangle code!

Berstanio: Okay, I thought more about the supertriangle stuff.

I'm not sure whether the already code handles this case, but I think this might be problematic:
Consider only  three points, that are close to being co-linear. They will have a huge circumcircle, so that they might contain vertices of the super-triangle. This _might_ break the triangulator code, as triangles that shouldn't violate the delaunay condition might break it now.
The original issue also mentions near co-linear points, so it might be related.

If this is an issue, the proper way to fix it might be to just move the supertriangle vertices to infinity. And for the circumcircle check we could use the half-space then.

But again, I don't really understand the code very well and it might not be an issue here.

tommyettinger (approved):
I have some minor issues with the test, but they don't affect its accuracy, so I'm not bothered if this merges unchanged. (Specifically, `Collections.shuffle(Arrays.asList(points));` won't treat the primitive float array `points` as a generic Object array that it can use as varargs, and will instead make a List consisting only of a `float[]`. Shuffling a one-element List does nothing, so this won't crash; if it was even a two-element List, `Arrays.asList()` produces an unmodifiable List, and that would I think throw an Exception if shuffled. Long story short, you can and maybe should just delete all the `Collections.shuffle(Arrays.asList(points));` lines.)

Notably, the test here is passed every single time by the DelaunayTriangulator changes here, but the newer PR #7549 fails them sometimes. This PR should be preferred.

gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
Berstanio: @jose-puente 
I think my suspicion was correct. Take this test:
```
	public static class NearlyColinearTest extends DelaunayTriangulatorTest {
		@Parameters
		public static Object[] nearlyColinear () {
			float[] hs = new float[6];
			hs[0] = 1;
			hs[1] = 1;

			hs[2] = 1;
			hs[3] = 100;

			hs[4] = 10f;
			hs[5] = 200;
			return new Object[]{hs};
		}

		@Override
		protected int expectedTrianglesCount () {
			return 1;
		}
	}
```
This fails with your current super triangle, but passes with the old one. 
I think we should really move the super triangle to infinity.

NathanSweet: Cool, we should add the `nearlyColinear` test(s). It's great we have tests!

It'd probably be fine to make the super triangles 10% larger than the vertices' AABB. It should be large enough to not make nearly colinear triangles, but not so large that it reduces precision.

gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java
```diff
@@ -16,15 +16,12 @@
...
-import com.badlogic.gdx.utils.FloatArray;
-import com.badlogic.gdx.utils.IntArray;
-import com.badlogic.gdx.utils.ShortArray;
+import com.badlogic.gdx.utils.*;
 
 /** Delaunay triangulation. Adapted from Paul Bourke's triangulate: http://paulbourke.net/papers/triangulate/
  * @author Nathan Sweet */
 public class DelaunayTriangulator {
-	static private final float EPSILON = 0.000001f;
+	static private final float EPSILON = Float.MIN_VALUE;
```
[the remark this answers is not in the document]
0XDE57: What is the typical scale of polygons expected by this implementation? 

At what size is a polygon / triangle too big or too small?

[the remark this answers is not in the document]
NathanSweet: Sorry I missed these questions. Typical usage is probably related to pixel resolution, eg not larger than 8K and not extremely small. With doubles and the `1e-9` epsilon that was merged it's probably safe in the range of 10^-3 to 10^9 (`0.001` to `1,000,000,000`).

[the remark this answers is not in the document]
obigu: Looks like the PR is almost there. @NathanSweet Is setting EPSILON to `1e-9` the only thing preventing this from being merged?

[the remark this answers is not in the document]
Berstanio: > Is setting EPSILON to 1e-9 the only thing preventing this from being merged?

I think the super-triangle changes should be either reverted or the super-triangle moved to infinity (as-in, we assume the points of the super-triangle lie in infinity and consider half-spaces when doing circumcircle inclusion/exclusion). 
Reducing the super-triangle does increase numerical stability, but it decreases algorithmic-stability unfortunatly (I missed that earlier, when I suggested the reduction). Doing half-space calculations removes the precision issues introduced by super triangles, while keeping correct results.
I have an implementation of this for my own delauynay implementation, but I probably wont be able to adapt it to libGDX before christmas.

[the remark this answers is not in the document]
NathanSweet: Half-space checks sound good!

Berstanio:
Superseeded by https://github.com/libgdx/libgdx/pull/7800 based on this pr, thank you for the contribution!

PR #7530, opened 2024-11-26
Fix for classpath files on GWT.

tommyettinger:
This solves the problem in at least libGDX tests where the default font, lsans-15, is used by `new BitmapFont()` with no args. It also should fix some 3D code that uses default shaders from classpath files.

It's a one-line change! Four chars, really. I don't know precisely why this works, but it does. It makes sense that filePathOrig should come from orig, not dest, though, and now that is true.


BoBIsHere86:
Pulled your changes down and re-tested. All of the once broken gwt tests related to Bitmapfont() are now working, as well as a custom test I was running when I found it. Awesome.
👍

PR #7531, opened 2024-11-27
Javadoc

Yuri69420:
Pull request of javadocs I have added


Yuri69420:
I have added these to help our teacher what's happening, because we asked for their help and they were just lost. So I understand that less might be better but I believe that it is a bit more clearer with such comments.

gdx/src/com/badlogic/gdx/InputMultiplexer.java
```diff
@@ -55,24 +72,33 @@ public int size () {
 		return processors.size;
 	}
 
+	/** Empty the list of processors */
 	public void clear () {
 		processors.clear();
 	}
 
+	/** Empty the list of processors and implement a new list of processors
```
tommyettinger, line 80: OK... this would be a big review.

This line isn't even right. Like, the only part that's correct is that it empties the... well, not a list... of processors. And like the rest of the PR, there's no benefit to `@param processors`, we can see that immediately already. It might help to know it takes 0 or more InputProcessor items as varargs or in an array! Just knowing its name doesn't tell me that.

gdx/src/com/badlogic/gdx/InputMultiplexer.java
```diff
@@ -25,27 +25,44 @@
...
 
+	/** Add a processor to the list of processors at the end of the list
+	 * 
+	 * @param processor */
 	public void addProcessor (InputProcessor processor) {
 		if (processor == null) throw new NullPointerException("processor cannot be null");
 		processors.add(processor);
 	}
 
+	/** Remove a processor from the list of processors
```
tommyettinger, line 63: Would be nice to know it's removed by identity, not by value.

gdx/src/com/badlogic/gdx/AbstractInput.java
```diff
@@ -3,18 +3,36 @@
 
 import com.badlogic.gdx.utils.IntSet;
 
+/** Abstract implementation of the Input interface, providing functionalities for handling key presses, including checking if keys
+ * are currently pressed or have just been pressed. This class is intended to be extended by platform-specific input
+ * implementations. */
+
 public abstract class AbstractInput implements Input {
```
tommyettinger, line 10: If Input already has JavaDocs for a method, you generally shouldn't make duplicate JavaDocs in each subclass, unless something is different in that subclass that invalidates the parent docs.

gdx/src/com/badlogic/gdx/Version.java
```diff
@@ -46,20 +46,36 @@ public class Version {
 		}
 	}
 
+	/** @param major
+	 * @param minor
+	 * @param revision
+	 * @return true if the libGDX version is strictly higher than the current one, false else */
```
tommyettinger, line 52: This whole class of JavaDocs is just stating the obvious...

gdx/src/com/badlogic/gdx/AbstractGraphics.java
```diff
@@ -1,19 +1,29 @@
 
 package com.badlogic.gdx;
 
+/** Abstract implementation of the Graphics interface, providing common methods for graphics-related functionalities such as time
+ * and density calculations. */
 public abstract class AbstractGraphics implements Graphics {
 
+	/** Delegates to {@link #getDeltaTime()} for the actual time calculation.
```
Berstanio, line 8: If a method already has javadocs on the super method, it doesn't need new javadocs.

gdx/src/com/badlogic/gdx/InputMultiplexer.java
```diff
@@ -55,24 +72,33 @@ public int size () {
...
 		this.processors.clear();
 		this.processors.addAll(processors);
 	}
 
+	/** @return the list of processors */
 	public SnapshotArray<InputProcessor> getProcessors () {
 		return processors;
 	}
 
+	@Override
```
Berstanio, line 101: The added override annotations are generally fine to me

tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/AwtTestLWJGL.java
tommyettinger, line 1: I don't think any of the tests need JavaDocs. They aren't exposed to users to call... Who would be reading them? Why wouldn't they be reading the source?

gdx/src/com/badlogic/gdx/AbstractGraphics.java
```diff
@@ -1,19 +1,29 @@
 
 package com.badlogic.gdx;
 
+/** Abstract implementation of the Graphics interface, providing common methods for graphics-related functionalities such as time
+ * and density calculations. */
 public abstract class AbstractGraphics implements Graphics {
```
tommyettinger, line 6: Same issue as AbstractInput.

gdx/src/com/badlogic/gdx/Version.java
```diff
@@ -46,20 +46,36 @@ public class Version {
 		}
 	}
 
+	/** @param major
```
Berstanio, line 49: If their is no parameter description things like `@param major` are in my opinion pointless.
Just noting the return is imo fine here

tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/DragNDropTest.java
```diff
@@ -19,25 +19,36 @@
...
 		skin = new Skin(Gdx.files.internal("data/uiskin.json"));
+		// Set the input processor to the stage for handling input events
 		Gdx.input.setInputProcessor(stage);
+		// Initialize the root Table to hold file labels
 		root = new Table();
-		root.setFillParent(true);
-		root.align(Align.left | Align.top);
-		stage.addActor(root);
+		root.setFillParent(true); // Make the table fill the entire stage
+		root.align(Align.left | Align.top); // Align the table's contents to the top-left
```
Berstanio, line 45: There are a lot of comments like this, where the code is 100% self explanatory.

Berstanio (changes requested):
I think some of the javadoc additions are fine, also the override annotations.

However, javadoc on override methods is in my opinion counter-productive. If there is something special to note, it should be noted in the interface.

There are also a lot of inline comments that I would consider unnecessary

tommyettinger (changes requested):
Sorry to be blunt, but I don't like this at all. There's a lot of important things that are undocumented here, and a lot of unimportant things that are vaguely documented. Tests don't need docs, at all, unless they are part of a public API or you need to tell something to the reader. If the reader needs to know something in a test, they are probably reading the source.

I don't think your teacher will learn anything new about libGDX via these comments. I'm beginning to wonder if I could teach a libGDX class, given that I lack even a Bachelor's degree, but maybe know some things already... Not like a generic Bachelor's degree actually indicates much for job skills. Some might!

tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/AwtTestLWJGL.java
Yuri69420: I feel it's much easier to read the code and understand it with comments and javadocs instead of looking for the source

Frosty-J:
I expect Tommy will still be caught up on the "list" wording for the `InputProcessor`s and Berstanio on documenting overrides (anyone viewing tests should know what `ApplicationListener#resize` is... if they don't, it is easy to find out with any decent IDE).

I'm still not a fan of some of the wording. "This test uses libGDX"... well duh. "false else" is a strange order, probably best rewrite the whole comment.

PR #7532, opened 2024-11-27
SortTest class (clean)

JujuG02:
I made tests for the class Sort.


Frosty-J:
See line 492 in the output: `Run './gradlew :gdx:spotlessApply' to fix these violations.` This doesn't have to be done from the command line; you can use the Gradle tab in IntellIJ IDEA/Android Studio to run `spotlessApply`.

PR #7535, opened 2024-11-28
FloatCounter test and possible corrections

melvilpf:
I have worked on the com.badlogic.gdx.math FloatCounter class. I just wanted to add some tests to find out the limits of the counter and I am thus suggesting a few modifications to the class to avoid big errors with the large numbers (I think this could be interesting, since this class isn't meant for such usages, to prevent them from happening).

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/FloatCounter.java
+++ b/gdx/src/com/badlogic/gdx/math/FloatCounter.java
@@ -50,6 +50,15 @@ public class FloatCounter implements Poolable {
 	/** Add a value and update all fields.
 	 * @param value The value to add */
 	public void put (float value) {
+		if (Float.isNaN(value)) throw new RuntimeException("value is not a number");
+		/*
+		Suggestions that I make to avoid errors when putting big numbers :
+
+		else if (Float.isInfinite(value)) throw new RuntimeException("value is infinite");
+		else if (value==Float.MAX_VALUE||value==-Float.MAX_VALUE) throw new RuntimeException("value is too big and " +
+				"there are risks that the next results will be different than expected");
+		 */
+
 		latest = value;
 		total += value;
 		count++;
--- a/gdx/src/com/badlogic/gdx/math/FloatCounter.java
+++ b/gdx/src/com/badlogic/gdx/math/FloatCounter.java
@@ -52,11 +52,11 @@ public class FloatCounter implements Poolable {
 	public void put (float value) {
 		if (Float.isNaN(value)) throw new RuntimeException("value is not a number");
 		/*
-		Suggestions that I make to avoid errors when putting big numbers :
-
-		else if (Float.isInfinite(value)) throw new RuntimeException("value is infinite");
-		else if (value==Float.MAX_VALUE||value==-Float.MAX_VALUE) throw new RuntimeException("value is too big and " +
-				"there are risks that the next results will be different than expected");
+		 * Suggestions that I make to avoid errors when putting big numbers :
+		 * 
+		 * else if (Float.isInfinite(value)) throw new RuntimeException("value is infinite"); else if
+		 * (value==Float.MAX_VALUE||value==-Float.MAX_VALUE) throw new RuntimeException("value is too big and " +
+		 * "there are risks that the next results will be different than expected");
 		 */
 
 		latest = value;
--- a/gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
+++ b/gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
@@ -1,153 +1,152 @@
+
 package com.badlogic.gdx.math;
 
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.Assert;
 
-import static java.lang.Math.pow;
-
 public class FloatCounterTest {
-    private FloatCounter counter1;
-    private FloatCounter counter2;
+	private FloatCounter counter1;
+	private FloatCounter counter2;
 
-    @Before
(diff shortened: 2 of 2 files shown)


gdx/src/com/badlogic/gdx/math/FloatCounter.java
```diff
@@ -50,6 +50,15 @@ public FloatCounter (int windowSize) {
 	/** Add a value and update all fields.
 	 * @param value The value to add */
 	public void put (float value) {
+		if (Float.isNaN(value)) throw new RuntimeException("value is not a number");
+		/*
+		 * Suggestions that I make to avoid errors when putting big numbers :
+		 * 
+		 * else if (Float.isInfinite(value)) throw new RuntimeException("value is infinite"); else if
+		 * (value==Float.MAX_VALUE||value==-Float.MAX_VALUE) throw new RuntimeException("value is too big and " +
+		 * "there are risks that the next results will be different than expected");
```
Frosty-J, line 59: I do not like the idea of throwing an exception for crazy values. Someone might want to do that, or there might be some scenario where a game ends up at `MAX_VALUE` and now it'll crash (which is worse than just behaving incorrectly).

gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
```diff
@@ -0,0 +1,152 @@
...
+		counter1.reset();
+		counter1.put(Float.MIN_NORMAL);
+		Assert.assertEquals(1.17549435E-38f, counter1.latest, 0);
+		counter1.put(10000.85f);
+		Assert.assertEquals(10000.850000000000000000000000000000000000117549435f, counter1.total, 0);
+		counter1.put(75.15f);
+		counter1.put(7E18f);
+		Assert.assertEquals(1750000000000002519.00000000000000000000000000000000000002938735875f, counter1.average, 0);
+		counter1.put(5f);
+		Assert.assertEquals(1400000000000002013.80000000000000000000000000000000000002350988700f, counter1.mean.getMean(), 0);
```
Frosty-J, line 117: This looks like you've hardcoded values that will pass the test. Consider using `MathUtils#isEqual`.

gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
```diff
@@ -0,0 +1,152 @@
...
+		 * Because we put the max value and the total cannot be more than max value (except infinity), the new total will be 0
+		 * (should we prohibit MAX_VALUE and INFINITY to avoid that?)
+		 */
+		Assert.assertEquals(0f, counter2.total, 0);
+		counter2.put(Float.POSITIVE_INFINITY);
+		Assert.assertEquals(Float.POSITIVE_INFINITY, counter2.total, 0);
+		counter2.put(Float.NEGATIVE_INFINITY);
+		/*
+		 * After implementing this next test, I realized that putting POSITIVE_INFINITY then NEGATIVE_INFINITY, the total was not a
+		 * number (same as before, should this be prohibited ?)
```
Frosty-J, line 144: Consider marking your pull request as draft if it contains questions, since this doesn't look ready to be merged.

Frosty-J (changes requested):
Interesting observations.

gdx/src/com/badlogic/gdx/math/FloatCounter.java
melvilpf: should i keep the NaN exception though ?

gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
melvilpf: okay, will do !

gdx/test/com/badlogic/gdx/math/FloatCounterTest.java
melvilpf: also done thank you

gdx/src/com/badlogic/gdx/math/FloatCounter.java
melvilpf: or maybe I should avoid having both POSITIVE_INFINITY and - NEGATIVE_INFINITY ? that's the issue I tested where the total was then NaN

Frosty-J: I wouldn't throw an exception over `put`ting a NaN value for much the same reason. Maybe we should return early on NaN to avoid screwing up the total/average. I'd be interested to hear what someone else thinks.

Infinities and other extreme values will always cause problems.

Frosty-J: I haven't inspected it but I reckon NaN would break `WindowedMean`.

melvilpf: hey so should I keep FloatCounter as it is and rewrite the tests for it to take into account NaN etc ?

Frosty-J: If you want to. I'm still not that sure on any of this; would like to see if someone who actually uses `FloatCounter` has thoughts on the matter.

melvilpf: okay, I can still wait for someone to confirm how it works

NathanSweet: Imagine if we checked for odd float values for _every_ method that accepted floats -- it isn't reasonable to guard against unless there are some special requirements.

PR #7539, opened 2024-12-03
Add OpenGL ES 3.2 support for Android

AndEditor7:
OpenGL ES 3.2 support for Android is fully implemented. However, some libGDX tests for GL31+ are broken. 
It does support the OpenGL ES version fallback if it's not supported on the Android device.

This pull request is currently daft because I need feedback, and help fixing the libGDX GL31+ tests.

This pull request includes #7526 changes.


AndEditor7:
I just realized I forgot to remove some changes I made to get the libGDX project working on my end.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL32.java
```diff
@@ -0,0 +1,237 @@
...
+	public void glDebugMessageInsert (int source, int type, int id, int severity, String buf) {
+		GLES32.glDebugMessageInsert(source, type, id, severity, buf.length(), buf);
+	}
+
+	@Override
+	public void glDebugMessageCallback (DebugProc callback) {
+		GLES32.glDebugMessageCallback(new GLES32.DebugProc() {
+			@Override
+			public void onMessage (int source, int type, int id, int severity, String message) {
+				callback.onMessage(source, type, id, severity, message);
```
tommyettinger, line 44: `DebugProc callback` should be `final DebugProc callback` because libGDX continues to live in the Java 7 caves and refuses to move into the light of language level 8. But that's another issue, sort-of. This can be fixed here just by adding `final`.

tommyettinger (changes requested):
Ook ooook, me no like 7 cave. Ook. We go 8? Is good. For now. Add word.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGL32.java
AndEditor7: `DebugProc callback` is now final!

AndEditor7:
> Ook ooook, me no like 7 cave. Ook. We go 8? Is good. For now. Add word.

Yeah, I think libGDX can go with Java 8. This would mean bumping the minimal Android API version to 24 to support most Java 8 features. The current Android API is 19, but this pull request bumps the Android API to 21. It will only lose 2% of Android distribution if libGDX goes from 21 to 24.

tommyettinger (dismissed):

AndEditor7:
I fixed many GDX tests. The one test that runs but does not work is GL32AdvancedBlendingTest, I'm not sure why but I'll look into it later. The others do not work on my device because of limitations or not implemented.

Tomorrow, I'm going to work on adding the geometry shader test since it's a feature from OpenGL ES 3.2

PR #7540, opened 2024-12-04
Add unit tests for Polygon and update of GeometryUtils

Juliette-Rlt:
I have refactored the test class for Polygon to make it more useful.

While testing the value returned for the area of a concave polygon, I noticed that the value returned by the method polygonArea() could be negative. To address this, I modified the method to always return a positive value.


gdx/src/com/badlogic/gdx/math/GeometryUtils.java
```diff
@@ -222,7 +222,7 @@ static public float polygonArea (float[] polygon, int offset, int count) {
 			x1 = x2;
 			y1 = y2;
 		}
-		return area * 0.5f;
+		return Math.abs(area * 0.5f);
```
Frosty-J, line 225: If this was `Math.abs(area) * 0.5f`, it would better match `triangleArea()`. But I don't believe the order makes any real difference to anything.

gdx/test/com/badlogic/gdx/math/PolygonTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+
+public class PolygonTest {
+
+	@Test
+	public void testZeroRotation () {
+		float[] vertices = {0, 0, 3, 0, 3, 4};
+		Polygon polygon = new Polygon(vertices);
+		polygon.rotate(0);
+		assertArrayEquals("The polygons vertices don't correspond.", polygon.getTransformedVertices(), polygon.getVertices(), 1f);
```
Frosty-J, line 30: Should be `polygon's` since there is one polygon and it's possessive. Then again, the other tests that already exist in libGDX are a grammatical disaster and no-one really cares about that, since there's bigger things to worry about in life.

gdx/test/com/badlogic/gdx/math/PolygonTest.java
```diff
@@ -0,0 +1,57 @@
...
+		assertEquals("The area doesn't correspond.", expectedArea, concavePolygon.area(), 1f);
+	}
+
+	@Test
+	public void testTriangleArea () {
+		float[] triangleVertices = {0, 0, 2, 3, 4, 0};
+		Polygon triangle = new Polygon(triangleVertices);
+		float expectedArea = 6.0f;
+		assertEquals("The area doesn't correspond.", expectedArea, triangle.area(), 1f);
+
```
Frosty-J, line 55: Empty line, isn't needed.

Frosty-J (commented):
I see no real issues with this now that Spotless has been run on it. Only inconsequential nitpicks remain.

Tom-Ski (approved):

PR #7543, opened 2024-12-14
Add AndroidX Core dependency to Android backend

obigu:
https://github.com/libgdx/libgdx/pull/7004 introduced a dependency to `androidx.core` that was not added as a transitive dependency and required users to manually add it to their projects.

The version chosen of AndroidX Core is `1.13.1` which is the latest we can support with out `minSDK` of `19`. This version requires setting `targetSdk` to, at least, `34` which is something we had to do anyway because Google Play requires 34+ target sdk anyway.

The Android Tests module also requires Kotlin BOM dependency to handle a dependency collision that is well documented and should disappear when we upgrade to newer versions in the future.

Finally I've removed the explicit `BuildTools` version as it is automatically set by the AGP.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.13.1]
+- [BREAKING CHANGE] Android: Since 1.13.0 libGDX requires setting `android.useAndroidX=true` in your gradle.properties file. In 1.13.1 it is NO longer needed to define the `androidx.core:core` dependency in your Android module.
 - iOS: Update to MobiVM 2.3.22
 - iOS: Fixes Gdx.openURI() not working on iOS 18.1 Simulator.
 - Change visibility of PolygonSpriteBatch.switchTexture() to protected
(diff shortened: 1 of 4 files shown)


Frosty-J:
Would you consider targetSdk 35 or is there something blocking that change from happening? Android 15 has been out a few months now and it seems likely that the next libGDX version will be the last before Google requires it for new apps.

obigu:
> Would you consider targetSdk 35 or is there something blocking that change from happening? Android 15 has been out a few months now and it seems likely that the next libGDX version will be the last before Google requires it for new apps.

The idea with this PR is to make the minimum changes to fix 1.13.0. There's a separate versions upgrade PR (https://github.com/libgdx/libgdx/pull/7526) where the upgrades can be more radical. In any case, targetSDK is set by the user projects/Liftoff and what we set here doesn't make a big difference.

Berstanio (approved):

PR #7544, opened 2024-12-15
Use docker container action for building Linux natives in natives-linux job to fix issues with Node 20+

### LINKED ISSUE #7477: Fix natives-linux job to work with latest versions of actions and Node 20
The issue is, Upload-Artifact@v4 (and Download-Artifact@v4) require Node 20, which does not work with glibc 2.17. 

But v3 [is being deprecated on November 30th, 2024](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/) and workflows using it will fail after that date. This will break the docker container setup for building natives that was added in https://github.com/libgdx/libgdx/pull/7178, because we currently haven't found a way in which it could be updated to v4 while still using node 16 and glibc 2.17 for the container.

It would be great if we could find a workaround for this while being able to keep the current setup with glibc 2.17.

I discovered this impending issue while we were trying to fix node 20 being forced by default in GitHub Actions which is causing issues in the jamepad repo for a similar docker setup. https://github.com/libgdx/Jamepad/pull/23

SonicGDX:
Similar to https://github.com/libgdx/Jamepad/pull/34, this uses a dockerfile and script to build the linux natives in a Ubuntu 18.04 docker container instead of running the entire natives-linux job in the container. This prevents needing to run actions like checkout and upload-artifact in the container, meaning it can still use glibc 2.17 to build without any issues with Node. Hopefully this means it can continue to work with newer versions of checkout and upload-artifact for the forseeable future. I haven't updated those action versions in this PR, but they can be updated without any issues.

I've done this by moving the docker setup (that was introduced in #7178) into its own action (which is local to the repo) that runs a script in the container that builds the natives. How it works is that when the workflow uses this action and enters the docker container, the working directory is mounted and mapped from the runner onto the container. So the runner does all the GitHub actions to clone/setup the repo and then enters the container to build and pack the natives. After the container exits the changes persist on the runner and the runner uploads the artifacts.

See https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-docker-container-action for more info on docker container actions.

Tested working here: https://github.com/SonicGDX/libgdx/actions/runs/12339676708

Other minor changes:

- Used `apt-get` instead of `apt` because using `apt` in a script gives the warning message "apt does not have a stable CLI interface".
- Removed usages of `sudo` just to be consistent (I can't always use sudo because initially sudo isn't installed)
- Added `--no-daemon` to gradle command because only one gradle command is being run so the daemon isn't necessary.
- Added `--quiet` to apt-get usages to cut down unuseful lines in logs
- Moved around arguments to `apt-get` so they're always at the front
- removed setup-java and gradle-build-action usages in the natives-linux step because zulu installation is done in the container and gradle is downloaded when the wrapper is run. This of course has the downside that the gradle action cache can't be used, but running the gradle-build-action in the runner doesn't seem to work due to lock file issues.
- `gnupg` gets installed anyway as a dependency but I added it explicitly because it is needed for dearmoring the azul key

Having the action be in a separate actions folder may not be desirable, in which case I can probably move the action into the workflow folder somewhere but I'm not sure where.

Closes #7477

### DIFF
--- a/.github/actions/build-linux-natives/Dockerfile
+++ b/.github/actions/build-linux-natives/Dockerfile
@@ -1,0 +1,8 @@
+# Container image that runs your code
+FROM ubuntu:18.04
+
+# Copies your code file from your action repository to the filesystem path `/` of the container
+COPY entrypoint.sh /entrypoint.sh
+
+# Code file to execute when the docker container starts up (`entrypoint.sh`)
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/actions/build-linux-natives/action.yml
+++ b/.github/actions/build-linux-natives/action.yml
@@ -1,0 +1,6 @@
+# action.yml
+name: 'Build Linux Natives in Docker Container'
+description: 'Build Linux Natives in Ubuntu 18.04 docker container'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
--- a/.github/actions/build-linux-natives/entrypoint.sh
+++ b/.github/actions/build-linux-natives/entrypoint.sh
@@ -1,0 +1,44 @@
+#!/bin/sh -l
+
+# ubuntu dockerfile is very minimal (only 122 packages are installed)
+# need to install updated git (from official git ppa)
+apt-get -q update
+apt-get -yq install software-properties-common
+add-apt-repository ppa:git-core/ppa -y
+# install dependencies expected by other steps
+apt-get -q update
+apt-get -yq install git \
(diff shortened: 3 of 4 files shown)


Tom-Ski:
Looks good to me, i think we will have this as a temporary solution before we bump up our min ubuntu version. Lwjgl have announced they are doing the same, so there isn't a huge amount of point sticking with a workaround at that point

Issue #7537, opened 2024-11-30
NullPointerException on AndroidXKeyboardHeightProvider.start on version 1.13.0


DavidPDev:
#### Issue details
Hi, after updating from libGDX 1.12.1 to 1.13.0 and published a couple of games some errors have appeared in the Google Play Console and firebase crashlytics:
```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.setOnApplyWindowInsetsListener(android.view.View$OnApplyWindowInsetsListener)' on a null object reference
  at androidx.core.view.ViewCompat$Api21Impl.setOnApplyWindowInsetsListener (ViewCompat.java:4840)
  at androidx.core.view.ViewCompat.setOnApplyWindowInsetsListener (ViewCompat.java:2537)
  at com.badlogic.gdx.backends.android.keyboardheight.AndroidXKeyboardHeightProvider.start (AndroidXKeyboardHeightProvider.java:33)
  at com.badlogic.gdx.backends.android.AndroidApplication$2.run (AndroidApplication.java:298)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:233)
  at android.os.Looper.loop (Looper.java:344)
  at android.app.ActivityThread.main (ActivityThread.java:8248)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:589)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1071)
```

#### Reproduction steps/code
This is not happening to all users, just "some" of them, I can't identify a common pattern...
Happens to different device brands and OSs:
![image](https://github.com/user-attachments/assets/8ab85a51-253e-4105-9387-cd56937c8331)

#### Version of libGDX and/or relevant dependencies
> libGDX 1.13.0

Of course I have added on build.gradle file the line:
`implementation 'androidx.core:core:1.13.1' `

And the gradle.properties file has the line:
`android.useAndroidX=true`

#### Please select the affected platforms
- [X ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

Frosty-J:
I don't know why this might be happening. Hopefully #7485 can fix it, but it's hard to say if you don't own an afflicted device to test on.

[...]

PR #7553, opened 2024-12-27
Defer content view resolve for KeyboardHeightProvider to allow for late setContentView

Berstanio:
Fixes https://github.com/libgdx/libgdx/issues/7537

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
 - Added XmlReader.getChildren() and XmlReader.replaceChild()
 - LWJGL3: Fix pauseWhenLostFocus not working as expected
 - API Addition: Added FPSLogger#setBound
+- Android: Fix crash on startup if `setContentView` was manually called after `initializeForView`
 
 [1.13.0]
 - [BREAKING CHANGE] GWT: Updated to 2.11.0. `com.google.jsinterop:jsinterop-annotations:2.0.2:sources` must be added as a dependency to your html project dependencies.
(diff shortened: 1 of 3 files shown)


obigu (approved):

anhduyqh123:
NullPointerException on AndroidXKeyboardHeightProvider.close on version 1.13.1

Berstanio:
> NullPointerException on AndroidXKeyboardHeightProvider.close on version 1.13.1

Do you have a stacktrace? Is the crash common?

tommyettinger:
Looks an awful lot like a near-copy of the text for the PR (for 1.13.0) directly above it, @Berstanio . I'm thinking spam or other non-standard-human behavior. So I don't think it's real.

Berstanio:
> Looks an awful lot like a near-copy of the text for the PR (for 1.13.0) directly above it, @Berstanio . I'm thinking spam or other non-standard-human behavior. So I don't think it's real.

Yeah, not sure. The thing is, it is technically possible to NPE, if `onDestroy` is called before/without `onResume`, but not sure whether this can actually happen in the android lifecycle.

obigu:
> Yeah, not sure. The thing is, it is technically possible to NPE, if `onDestroy` is called before/without `onResume`, but not sure whether this can actually happen in the android lifecycle.

I think if the Activity forcibly finishes on `onCreate()` or `onStart()`, `onResume()` won't be called. Not a very usual scenario but probably worth adding a null check condition in `onDestroy()`.

anhduyqh123:
Screenshot 2025-01-18 170742
When I updated from version 1.13.0->1.13.1 I encountered a problem as shown in the attached image.

Issue #7552, opened 2024-12-27
Update LWJGL3 dependency to anything that passes AV.


tommyettinger:
#### Issue details
LWJGL 3.3.4, which libGDX 1.13.0 currently depends on, triggers several antivirus programs, including (intermittently) big ones such as Microsoft Defender. LWJGL 3.3.5 is out, though it still sometimes triggers even more AV programs (i.e. Google) on specific modules' native jars. It seems like unlike with 3.3.4, applications can bundle the files from inside 3.3.5's JARs in their own fat jar, and the fat jar won't trigger any AV programs. I really have no idea how commercial AV vendors stay in business, since the built-in Windows ones seem fine, but when I tried this on VirusTotal a week or two ago, some JARs failed Avast and AVG only. Now, the x64 Windows natives for lwjgl-stb trigger 11 different AV programs, including MS and Google's offerings. It only seems to be the JAR file that triggers it -- removing the .dll files and placing them in a .zip, or a larger .jar (which our fat JARs do), seems to eliminate issues.

#### Reproduction steps/code
Check VirusTotal for potential AV positives on whatever versions you are considering using in a release.

[Try to download gdx-liftoff-1.13.0.0.jar on Windows 11.](https://github.com/libgdx/gdx-liftoff/releases/download/v1.13.0.0/gdx-liftoff-1.13.0.0.jar) On my machine, it won't complete the download because Microsoft Defender intercepts it. This version (and only this version of Liftoff) uses LWJGL 3.3.4 .

[lwjgl-stb-3.3.5-natives-windows.jar has false positives even in the newest version.
](https://www.virustotal.com/gui/file/4cc056ced54c5d2882415529c4a528db83b61b157ecd01043f22233568419bfc)

As a possible solution, [This snapshot of Liftoff was built using LWJGL 3.3.5, and it seems to pass checks.](https://www.virustotal.com/gui/file/ad190d424dd3ff4cd19b9f0943c140c4da2b155dfc2e07a2990f6fab4b217dba) It looks like unpacking the native .JAR files may be sufficient to avoid false positives.

#### Version of libGDX and/or relevant dependencies
1.13.0, as well as the latest snapshot as of Dec 26, 2024 , depend on LWJGL 3.3.4 .

#### Stacktrace
```java
I would provide this if I could, but AV programs aren't especially chatty.
```

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows
[...]

PR #7555, opened 2024-12-31
Downgrade LWJGL 3 to 3.3.3

obigu:
Given the AV false positives of 3.3.4 (https://github.com/libgdx/libgdx/issues/7552) and some on 3.3.5 and the log spamming issue occurring on 3.3.5 (https://github.com/kcat/openal-soft/issues/1086) that won't be fixed until 3.3.6 best approach is to downgrade to stable 3.3.3 for libGDX 1.13.1 release until these problems are fixed.

RISC-V natives won't be shipped with libGDX, if there's any RISC-V user they can upgrade lwjgl and add the dependencies on their end.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@
 - Change visibility of PolygonSpriteBatch.switchTexture() to protected
 - Added XmlReader.getChildren() and XmlReader.replaceChild()
 - LWJGL3: Fix pauseWhenLostFocus not working as expected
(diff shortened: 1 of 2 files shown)


tommyettinger (approved):
Looks solid to me. The RISCV natives probably shouldn't be shipped anyway unless libGDX can also run on that arch; otherwise, they're just dead weight. I have no idea what RISCV devices even exist in ~~2024~~ 2025...

Frosty-J:
> The RISCV natives probably shouldn't be shipped anyway unless libGDX can also run on that arch

It does.

https://github.com/libgdx/gdx-jnigen/pull/48

Berstanio (approved):

PR #7557, opened 2025-01-01
API Addition: Added Array#replaceFirst and Array#replaceAll

TCreutzenberg:
API Addition: Added Array#replaceFirst and Array#replaceAll

These are quite handy and I already could replace one usage in the gdx code.


gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -192,6 +192,40 @@ public void swap (int first, int second) {
...
+			for (int i = 0, n = size; i < n; i++)
+				if (value.equals(items[i])) {
+					items[i] = replacement;
+					return true;
+				}
+		}
+		return false;
+	}
+
+	public void replaceAll (@Null T value, boolean identity, @Null T replacement) {
```
tommyettinger, line 218: I think there's a bit of a mismatch here in how replaceAll() returns void but replaceFirst() returns a usable boolean. Could replaceAll() return false if nothing was changed, or true if anything was replaced? Alternately, true if any value was found, even if the replacement didn't change it.

gdx/src/com/badlogic/gdx/utils/XmlReader.java
```diff
@@ -544,9 +544,9 @@ public void remove () {
 
 		public void replaceChild (Element child, Element replacement) {
 			if (children == null) throw new GdxRuntimeException("Element has no children: " + name);
-			final int index = children.indexOf(child, true);
-			if (index == -1) throw new GdxRuntimeException("Element does not contain child: " + child);
-			children.set(index, replacement);
+			if (!children.replaceFirst(child, true, replacement)) {
+				throw new GdxRuntimeException("Element '" + name + "' does not contain child: " + child);
```
tommyettinger, line 548: This looks good if you've tested it; I'm just assuming `name` is available at this point in the code. It looks like an improvement, if so!

tommyettinger (approved):
Looks pretty good to me. I'd encourage replaceAll() to return some info, like how replaceFirst() does, and like how several Collection methods in the JDK do. I'm approving and not requiring that change because I think this would help as-is, even if I think making replaceAll() return a boolean would be better.

gdx/src/com/badlogic/gdx/utils/XmlReader.java
SimonIT: These lines should probably be replaced as well...
https://github.com/libgdx/libgdx/blob/35afa7986e400f3e6e7b9b8dfb3d0fc3d74a4744/gdx/res/com/badlogic/gdx/utils/XmlReader.rl#L335-L337

TCreutzenberg: > I'm just assuming `name` is available at this point in the code.

It is used like this in several palces in the Element class and where I use it in my code it works.

TCreutzenberg: replaceAll now return the number of replacements done. This might be even more useful than just returning true or false?

tommyettinger: Looks good to me! I wonder why the JDK doesn't use that approach... I do like this better!

tommyettinger (approved):
Looks good; I'm glad the .rl file has been changed, too. It's hard to remember which files use those sometimes...

NathanSweet:
What do you guys think of `replaceFirst` -> `replace`? We have `removeValue` (and `removeAll`) which removes the first matching value and it doesn't specify "first" in its name. It's not exactly the same scenario, as we have `removeIndex` and needed "value" and "index" to make it clear what was being removed, eg if the values are Integer (an annoyance in the JDK). `replaceFirst` is nicely descriptive without being too long, while `removeFirstValue` would be unwieldy.

It looks good to me, I think we can merge as is. Possibly we could add these methods to other array classes later.

TCreutzenberg:
I'd (obviously) prefer `replaceFirst` as it instantly shows what's going on. But for consitency's sake I'd also be happy with just `replace`.

> Possibly we could add these methods to other array classes later.

That was my plan actually, I first wanted to see if you guys support the general idea of a replace-method.

PR #7560, opened 2025-01-05
Array classes: replaceFirst + replaceAll

TCreutzenberg:
Methods replaceFirst and replaceAll for each of these Array classes:
- BooleanArray
- ByteArray
- CharArray
- DelayedRemovalArray
- FloatArray
- IntArray
- LongArray
- ShortArray
- SnapshotArray


gdx/src/com/badlogic/gdx/utils/BooleanArray.java
```diff
@@ -170,6 +170,35 @@ public void swap (int first, int second) {
 		items[second] = firstValue;
 	}
 
+	/** Returns true if the specified value was replaced successfully with the replacement
+	 * @param value the boolean to be replaced
+	 * @param replacement the first value will be replaced by this replacement if found
+	 * @return if value was found and replaced */
+	public boolean replaceFirst (boolean value, boolean replacement) {
+		boolean[] items = this.items;
+		for (int i = 0, n = size; i < n; i++)
```
NathanSweet, line 179: A minor nitpick, but to match the style for readability no braces is OK only when the block fits on the same line or 1 additional line. If it's more lines, like this `if`, then it'd get braces.

TCreutzenberg: Fixed.

Issue #7561, opened 2025-01-06
Making FrameBuffer(s) nestable.


nulldoubt:
Framebuffers aren't nestable by default which causes way too much code clutter (many end() and begin() calls) to get that effect.

Why can't the `FrameBuffer` class just be nestable? By nestable I mean something like this:

```java

frameBufferA.begin();
frameBufferB.begin();
// draw
frameBufferB.end();
frameBufferC.begin();
// draw
frameBufferC.end();
frameBufferA.end(); // this captured the contents of both frameBufferB, and frameBufferC.

```

There are many working alternatives I found, including the [NestableFrameBuffer.java](https://github.com/crykn/guacamole/blob/master/gdx/src/main/java/de/damios/guacamole/gdx/graphics/NestableFrameBuffer.java) from the guacamole extension, which seems to work fine.

I'm opening this issue here to discuss whether I should create a PR for this or if it would be some sort of a breaking change.

Check out the [changes from my fork](https://github.com/libgdx/libgdx/compare/master...nulldoubt:libgdx:2df29dc42e42d63203fb198d8416be4c312fe982), there shouldn't really be any breaking changes though I'm not sure. The code wasn't tested yet.

Should probably work on GWT but I don't talk GWT and I'm unsure if that backend even supports the NIO Buffers...

Open for feedback!

crykn:
[A couple of years ago](https://github.com/libgdx/libgdx/pull/6343#issuecomment-757527119) I wanted to submit a PR for this, but apparently I never came around to that. Your code should work fine on GWT (as does NestableFrameBuffer); any special handling has become obsolete with #6175. However, IIRC the consensus seems to have been to make the nestability optional, since checking the currently bound framebuffer isn't free and could negatively affect performance.

streamsoftdev:
Is the render context the place to include state caching? E.g. I implemented my own `TextureBinder` to replace the default texture binder in the render context, and greatly reduced the number of texture binds since I could take into account more application specific details to consistently binding the same textures to the same places, not throwing away existing bound textures each frame, etc.

nulldoubt:
From my perspective a great gl-state caching layer would be similarly implemented as the `GLInterceptor` from the profiler in libGDX (basically, it hooks the Gdx.gl* instances), however, instead of intercepting all calls, it would basically just override the ones that can be cached and compare the previously cached value with the parameters. 

[...]

PR #7564, opened 2025-01-10
Made FrameBuffer nestable.

nulldoubt:
A framebuffer sequence may now be nested using the `FrameBuffer.begin()` and `FrameBuffer.end()` methods.

This means that the previously bound framebuffer, if available, will be automatically bound again.

See issue #7561 for more context.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/FrameBuffer.java
@@ -24,6 +24,10 @@ import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.Texture.TextureFilter;
 import com.badlogic.gdx.graphics.Texture.TextureWrap;
 
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.IntBuffer;
+
 /**
  * <p>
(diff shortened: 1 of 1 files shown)


mgsx-dev:
There are too many mistakes in this PR. (wrong class, duplicated code, unwanted synchronized code, many side effects that could break existing code, ...)

Also, as mentionned bt @crykn  in the issue : 
> However, IIRC the consensus seems to have been to make the nestability optional, since checking the currently bound framebuffer isn't free and could negatively affect performance.

Requesting to reject it.

nulldoubt:
Hey, thanks for your feedback!

Maybe I misunderstood the need.

I now see as well that there's unnecessary synchronization.

Also, just to make sure that I'm not mistaking it again, the class to modify should be the base class `GLFrameBuffer`, not the regular `FrameBuffer`, right?

And finally, this feature shouldn't be forced... It should be optional, so a toggling method, maybe called `setAutoRebind` or something similar, should be available, right?

Tom-Ski:
This seems to be something that should be handled externally in a util, rather than something that is baked deep into the framework internals. Much like how TextureBinder works, it is not enforced to be used.
```
NestedFrameBufferBinder {
begin(GLFramebuffer......)
end()

getCurrentBuffer()
getCurrentViewport

etc
} 
```

PR #7567, opened 2025-01-15
feat: Make reader in Json changeable

SimonIT:
Like the writer, the reader can now also be changed

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Json.java
+++ b/gdx/src/com/badlogic/gdx/utils/Json.java
@@ -46,6 +46,7 @@ public class Json {
 	static private final boolean debug = false;
 
 	private JsonWriter writer;
+	private JsonReader reader = new JsonReader();
 	private String typeName = "class";
 	private boolean usePrototypes = true;
 	private OutputType outputType;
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/utils/Json.java
```diff
@@ -46,6 +46,7 @@ public class Json {
 	static private final boolean debug = false;
 
 	private JsonWriter writer;
+	private JsonReader reader = new JsonReader();
```
Berstanio, line 49: A `JsonReader` seems to store some states when parsing, so I'm not sure whether having one `JsonReader` per `Json` class is non-breaking.

SimonIT: I thought so too. But then I ran the `JsonTest` and everything worked fine. I will check it more thoroughly (I also overlooked that there are attributes in the class because they are not at the top)

SimonIT: I added a test where the reader is reused. So I think it works. Or do you have any other ideas for tests?

NathanSweet: JsonReader fields are only used during `parse(...)`, it's safe to use the same reader instance multiple times.

PR #7569, opened 2025-01-18
Fix NPE on AndroidXKeyboardHeightProvider#close

Berstanio:
Simple possible fix for the issue mentioned here:
https://github.com/libgdx/libgdx/pull/7553#issuecomment-2599480691


obigu:
May the issue be here instead?

```
@Override
	protected void onDestroy () {
		super.onDestroy();
		keyboardHeightProvider.close();
	}
```

I think it's best practice to call onDestroy last 

```
@Override
	protected void onDestroy () {
                keyboardHeightProvider.close();
		super.onDestroy();
		
	}
```

Berstanio:
Good catch, maybe!

obigu:
Great, thx, do you mind adding a line to CHANGES and we can safely merge?

Berstanio:
Done!

PR #7570, opened 2025-01-19
feat: Deserialize char from number

SimonIT:
A char is just a number, so it makes sense to be able to deserialise from it


tommyettinger (approved):
Looks good, I think. As long as this doesn't interfere with Json's ability  to read in other types, like "90" as an int instead of the char 'Z', this looks very reasonable. Since short is already read in just fine, and uses nearly-identical code, I don't think that's a real concern.

Issue #7445, opened 2024-08-21
Lwjgl3: Not rendering since latest SNAPSHOT update


TCreutzenberg:
#### Issue details
I just started my app, gradle downloaded the latest 1.12.2-SNAPSHOT with these changes: https://github.com/libgdx/libgdx/commit/054b25b8ced6fa54ba0c8d959b6a08a8d54de9d9
Before this commit everything worked fine.

I debugged and found that my ApplicationListener only enters .render() once and then just hangs. The window stays black forever.

#### Version of libGDX and/or relevant dependencies
1.12.2-SNAPSHOT

#### Stacktrace
No stacktrace, no exception is thrown

#### Please select the affected platforms
- Windows
- (Other desktop platforms might be affected)

PR #7572, opened 2025-01-20
Use RTree for MaxRectsPacker overlap check

Frotty:
Following up on [#7453](https://github.com/libgdx/libgdx/pull/7453/files), I also wanted to improve the performance of MaxRectsPacker. In my case, the main bottleneck was the rectangle overlap check rather than the pruning logic addressed in the previous PR.

I started by trying a QuadTree-based approach, which gave a small improvement, but switching to an RTree-based solution significantly reduced my packing time from about 30 seconds down to 5 seconds in my project. I also noticed that RTrees performed better when fewer entries per node were allowed, so I ended up implementing a specialized RTree variant with a fixed number of entries per node.

I'm not entirely sure if it's worth adding this to the library, but it made a substantial difference in my particular use case. It would be great if @lucas-viva and @fgnm (or anyone else) could test it with their assets to see if these improvements hold more generally. Feedback would be greatly appreciated!

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java
@@ -797,20 +797,16 @@ public class MaxRectsPacker implements Packer {
 		ContactPointRule
 	};
 
-	/**
-	 * Specialized R-Tree with 2 entries per node.
-	 */
+	/** Specialized R-Tree with 2 entries per node. */
 	static class RTree {
 		private Node root;
 
-		public RTree() {
+		public RTree () {
 			this.root = new Node(true);
 		}
 
-		/**
-		 * Insert a rectangle into the R-tree.
-		 */
-		public void insert(Rect rect) {
+		/** Insert a rectangle into the R-tree. */
+		public void insert (Rect rect) {
 			Node leaf = chooseLeaf(root, rect);
 			addEntryToNode(leaf, rect);
 			leaf.mbr = combine(leaf.mbr, rect);
@@ -822,11 +818,9 @@ public class MaxRectsPacker implements Packer {
 			adjustTree(leaf);
 		}
 
-		/**
-		 * Retrieve (search) for all rectangles that might intersect the given query.
-		 * @return all intersecting rectangles
-		 */
-		public Array<Rect> retrieve(int queryX, int queryY, int queryWidth, int queryHeight) {
+		/** Retrieve (search) for all rectangles that might intersect the given query.
+		 * @return all intersecting rectangles */
+		public Array<Rect> retrieve (int queryX, int queryY, int queryWidth, int queryHeight) {
 			Array<Rect> results = new Array<>();
 			Rect query = new Rect(queryX, queryWidth, queryY, queryHeight);
 			search(root, query, results);
@@ -843,7 +837,7 @@ public class MaxRectsPacker implements Packer {
 			Node[] children = null;
 			int childCount = 0;
 
-			Node(boolean leaf) {
+			Node (boolean leaf) {
 				this.isLeaf = leaf;
 				if (!leaf) {
 					children = new Node[3];
@@ -851,7 +845,7 @@ public class MaxRectsPacker implements Packer {
 			}
 		}
 
-		private Node chooseLeaf(Node node, Rect rect) {
+		private Node chooseLeaf (Node node, Rect rect) {
 			if (node.isLeaf) {
 				return node;
 			}
@@ -872,12 +866,12 @@ public class MaxRectsPacker implements Packer {
 			return chooseLeaf(bestChild, rect);
 		}
 
-		private void addEntryToNode(Node node, Rect rect) {
+		private void addEntryToNode (Node node, Rect rect) {
 			node.entries[node.entryCount] = rect;
 			node.entryCount++;
 		}
 
-		private void addChildToNode(Node parent, Node child) {
+		private void addChildToNode (Node parent, Node child) {
 			parent.children[parent.childCount] = child;
 			parent.entries[parent.childCount] = child.mbr;
 			parent.childCount++;
@@ -885,7 +879,7 @@ public class MaxRectsPacker implements Packer {
 			parent.mbr = computeMBR(parent);
 		}
 
-		private void splitNode(Node node) {
+		private void splitNode (Node node) {
 			Node newNode = new Node(node.isLeaf);
 
 			Rect[] all = new Rect[3];
@@ -941,11 +935,11 @@ public class MaxRectsPacker implements Packer {
 			}
 		}
 
-		private int[] pickSeeds(Rect[] all) {
+		private int[] pickSeeds (Rect[] all) {
 			float maxWaste = -1;
 			int idx1 = 0, idx2 = 1;
 			for (int i = 0; i < 3; i++) {
-				for (int j = i+1; j < 3; j++) {
+				for (int j = i + 1; j < 3; j++) {
 					float areaCombined = area(combine(all[i], all[j]));
 					float areaI = area(all[i]);
 					float areaJ = area(all[j]);
@@ -957,10 +951,10 @@ public class MaxRectsPacker implements Packer {
 					}
 				}
 			}
-			return new int[] { idx1, idx2 };
+			return new int[] {idx1, idx2};
 		}
 
-		private void adjustTree(Node node) {
+		private void adjustTree (Node node) {
 			if (node == root) {
 				return;
 			}
@@ -974,15 +968,14 @@ public class MaxRectsPacker implements Packer {
 			}
 			parent.mbr = computeMBR(parent);
 
-			if ((parent.isLeaf && parent.entryCount > 2)
-				|| (!parent.isLeaf && parent.childCount > 2)) {
+			if ((parent.isLeaf && parent.entryCount > 2) || (!parent.isLeaf && parent.childCount > 2)) {
 				splitNode(parent);
 			}
 
 			adjustTree(parent);
 		}
 
-		private Node findParent(Node current, Node target) {
+		private Node findParent (Node current, Node target) {
 			if (current == target || current.isLeaf) {
 				return null;
 			}
@@ -998,7 +991,7 @@ public class MaxRectsPacker implements Packer {
 			return null;
 		}
 
-		private void search(Node node, Rect query, Array<Rect> results) {
+		private void search (Node node, Rect query, Array<Rect> results) {
 			if (!intersects(node.mbr, query)) {
 				return;
 			}
@@ -1019,18 +1012,17 @@ public class MaxRectsPacker implements Packer {
 
 		/* Helpers */
 
-		private static boolean intersects(Rect a, Rect b) {
+		private static boolean intersects (Rect a, Rect b) {
 			if (a == null || b == null) return false;
-			return !(a.x + a.width < b.x || a.x > b.x + b.width
-				|| a.y + a.height < b.y || a.y > b.y + b.height);
+			return !(a.x + a.width < b.x || a.x > b.x + b.width || a.y + a.height < b.y || a.y > b.y + b.height);
 		}
 
-		private static float area(Rect r) {
+		private static float area (Rect r) {
 			if (r == null) return 0;
-			return (float) r.width * (float) r.height;
+			return (float)r.width * (float)r.height;
 		}
 
-		private static Rect combine(Rect a, Rect b) {
+		private static Rect combine (Rect a, Rect b) {
 			if (a == null) return b;
 			if (b == null) return a;
 			int minX = Math.min(a.x, b.x);
@@ -1040,7 +1032,7 @@ public class MaxRectsPacker implements Packer {
 			return new Rect(minX, maxX - minX, minY, maxY - minY);
 		}
 
-		private static Rect computeMBR(Node node) {
+		private static Rect computeMBR (Node node) {
 			Rect m = null;
 			for (int i = 0; i < node.entryCount; i++) {
 				m = combine(m, node.entries[i]);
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
I just tested the changed MaxRectsPacker class in my standalone texture packer, https://github.com/tommyettinger/libgdx-texturepacker/commit/bd46582144ed08d9c43b3bfb451953d64bcdcc94 , and with a 4096x4096 atlas page using stripped whitespace, this is almost exactly 10x faster than the old one. WOW! Approved, for sure.

tommyettinger (changes requested):
I've investigated this pretty thoroughly now, and I do think I need to request some changes. Because this doesn't pack quite as tightly as before, either with fast mode on or off in the earlier code, it can't be a drop-in replacement for the existing algorithm. I think it would be reasonable to make a `faster` flag that switches to use this PR's algorithm, probably meant to be enabled in tandem with `fast`, though not required to be. [I made changes to do that here](https://github.com/tommyettinger/libgdx-texturepacker/commit/506fbb92a12fd88ca639e6bb15dbf0f75a064830#diff-407acda3873e3d2ad2098603138ebe702eca0e2762cf59d091ab53a19b29beac), and in some earlier commits from today. I also think there should be a `null` check around usage of `progress`, [in both MaxRectsPacker algorithms](https://github.com/tommyettinger/libgdx-texturepacker/blob/506fbb92a12fd88ca639e6bb15dbf0f75a064830/libgdx-texturepacker/src/main/java/com/badlogic/gdx/tools/texturepacker/MaxRectsPacker.java#L86-L89). I got an NPE in one test because that ProgressListener was null, and having a check fixed it.

Frotty:
@tommyettinger Thanks for the feedback and testing it rigorously❤️.
As we discussed this PR on discord, I want to add the following information that we discovered:
The gaps actually would be filled, if there would be no more space in the top left. This stems from the `fast` mode only comparing widths when sorting the candidates. This means a free, wider rectangle will always be preferred. Changing this to comparing area instead, will fill the gaps earlier, but not achieve higher density overall. Therefore, it is no more than an aesthetic problem.

tommyettinger (dismissed):

tommyettinger:
As long as the gaps will get filled in without needing more atlas pages, then the new very-very-fast mode should be totally fine! It probably doesn't need the faster mode either, with what we know now. Approved again!

tommyettinger (approved):
I like this. It could still use a null check on `progress`, but admittedly that isn't in the scope of this PR. Approved again!

lucas-viva:
Our total packing pipeline currently takes 4:15~ minutes. With these changes, it took 3:05~ minutes. Although sprites were rearranged a bit different, no extra pages were generated. Great job!

fgnm:
Had some time to test this PR and results are pretty good.. I'v tried with a test bench with 2692 images and a pretty fast machine (i9-13th):
- Base libGDX 1.13.1: 1:09 ~ minutes
- libGDX 1.13.2-SNAPSHOT: 45 ~ seconds
- libGDX 1.13.2-SNAPSHOT + this PR: 30 ~ seconds
- libGDX 1.13.2-SNAPSHOT + this PR + `fast=true` and `limitMemory=false`: 6 ~ seconds

100% approved... Aesthetics of the final atlas looks different, but as far as the result has no technical issues it's amazing as is, probably `limitMemory` should remain `true` by default, but why `fast` isn't enabled?

Frotty:
> Had some time to test this PR and results are pretty good.. I'v tried with a test bench with 2692 images and a pretty fast machine (i9-13th):
> 
> * Base libGDX 1.13.1: 1:09 ~ minutes
> * libGDX 1.13.2-SNAPSHOT: 45 ~ seconds
> * libGDX 1.13.2-SNAPSHOT + this PR: 30 ~ seconds
> * libGDX 1.13.2-SNAPSHOT + this PR + `fast=true` and `limitMemory=false`: 6 ~ seconds
> 
> 100% approved... Aesthetics of the final atlas looks different, but as far as the result has no technical issues it's amazing as is, probably `limitMemory` should remain `true` by default, but why `fast` isn't enabled?

Thanks for the detailed report 👍 
I agree. Even though `fast=true` packs slightly less dense, I would advocate for it to be default, as it speeds up the process tremendously with no real downsides - at least for development.
@tommyettinger Do you think changing this default is fine to add to this PR?

Tom-Ski:
I dont think the default should be a faster packing time over a higher quality packing. Considering the default now is the higher quality, changing this default is a breaking change, and quite a hidden one.
I think people should opt in to have a faster development cycle.

fgnm:
Ah got the point.. Didn't notice the number of pages produced and misunderstand a bit the `fast` option. My test has a lot of variation in size (from pixel art to high res assets) and the fast option produce one extra atlas page, which is not really great for production, so yes now I agree to keep things as they are. Anyway, this PR still brings improvements to the current TexturePacker, so I'd say it's fine as is

Frotty:
> I dont think the default should be a faster packing time over a higher quality packing. Considering the default now is the higher quality, changing this default is a breaking change, and quite a hidden one.
I think people should opt in to have a faster development cycle.

Although it's too late now, I'd generally prefer an opt-in 'high-quality production' mode rather than requiring an opt-in for faster development, as this aligns with many other frameworks, and development is a much more common use case. Most users won't even discover the fast option, but a dedicated production build—slower yet more optimized—is quite common. Personally, I've solved my own speed issues, but it's unfortunate that others might have a subpar experience due to default settings.

> As a side note, the tests above all have very similar sized assets. What happens to the quality of the packing when you have a lot of variation?

The RTree approach shouldn't reduce packing quality; it simply chooses fitting rectangles more quickly. Fast mode can be less dense depending on the input I would say, because it simply goes by largest width. I also explored changing this to rectangle area instead of width - it improved density with negligible performance impact - but then again for a dev mode that's not so relevant.

Frosty-J:
I suppose this is personal preference. I opt for efficient packing, with whitespace stripping and rotation. Though I do disable `limitMemory`, which doesn't seem important unless you're really starved for RAM. As my graphics don't change all that frequently, it's not a big deal if the atlas takes a bit longer to build. I can imagine people working in a team may need to build the atlas more frequently.

Tom-Ski:
> Although it's too late now, I'd generally prefer an opt-in 'high-quality production' mode rather than requiring an opt-in for faster development, as this aligns with many other frameworks, and development is a much more common use case. Most users won't even discover the fast option, but a dedicated production build—slower yet more optimized—is quite common. Personally, I've solved my own speed issues, but it's unfortunate that others might have a subpar experience due to default settings.
> 
Based on this, many people wouldn't find the optimized version if it were the other way around. Which is worse imo.

Frotty:
> Based on this, many people wouldn't find the optimized version if it were the other way around. Which is worse imo.

I think it's more common across the industry to use faster, iterative development modes—such as Unity’s Editor play mode, Unreal’s in-editor reload, or web dev’s hot module replacement—while deferring heavier optimizations to dedicated production builds. I suppose the different approach in libGDX aligns with its ethos of being a lean, unopinionated framework, granting developers full control but also requiring more manual effort.

> I suppose this is personal preference.

While personal preference does play a role, the choice between a faster, iterative packing process and a more thorough, time-consuming approach often depends on the specific data, project scope, and workflow. For example, as @tommyettinger showed, avoiding fast mode can increase processing times from under a minute to nearly an hour. In my own case, textures—sometimes just a single UI icon—are frequently added/updated, so switching to a faster packing method, combined with hash-based change detection, has reduced iteration times from minutes to mere seconds. This significantly boosts productivity and can be a major game changer.

Tom-Ski:
Yeah, if it was built in to some strict export pipeline, it would make sense. But as lots of people can use TP as they wish, and there is no standardized method of building a "production" build, it has many bad side effects changing defaults.

One less atlas page makes quite a lot of difference, especially if you don't have a single atlas, as that can potentially turn into many extra pages. And also when hosting downloadable content where you are paying for bandwidth. These things definitely matter

Frotty:
> Yeah, if it was built in to some strict export pipeline, it would make sense. But as lots of people can use TP as they wish, and there is no standardized method of building a "production" build, it has many bad side effects changing defaults.

Yep. As all things it has benefits and drawbacks. I do think a more streamlined pipeline for some of the deploy/packing tasks could be a neat addition, but now I have built this myself anyway.

> One less atlas page makes quite a lot of difference, especially if you don't have a single atlas, as that can potentially turn into many extra pages. And also when hosting downloadable content where you are paying for bandwidth. These things definitely matter

While having fewer atlas pages can matter in certain edge cases this should not have a significant impact for the majority of projects. In my experience (and with @tommyettinger’s data), no extra page was introduced in the first place, and even if one had been, it most likely would add less than 1 MB. Moreover, PNG compression can offset some of the extra space by taking advantage of gaps. I think it would be interesting to actually measure this with @fgnm 's data and see the increase in storage requirements.
Ultimately, these trade-offs depend on project needs and priorities. However, I think designing defaults around rare edge cases isn’t ideal. Developers with unique requirements are generally expected to invest in specialized solutions, while defaults should focus on the most common use case, **IMHO**.

Tom-Ski:
Png compression can help, but it still takes up more space in ram for no reason, which is also very critical on mobile. 

I'd agree with you if this wasn't an already well established default, and changing that causes a   regression.

Frotty:
> Png compression can help, but it still takes up more space in ram for no reason, which is also very critical on mobile.

Sure. I would probably assume there are bigger concerns regarding RAM than a few Megs of textures, but of course it's a "free" gain that should be taken.

> I'd agree with you if this wasn't an already well established default, and changing that causes a regression.

I agree that it's fine to keep it as is, but I think it's still valid to constructively discuss the status quo.
There are many possible ways to implement things without them being a breaking change, but still accomplishing the goal. E.g. something like `limitMemory` could be made obsolete by turning textures into weak references which are GC'd when the heap runs out and then re-loaded on demand.

Perhaps one very simple thing regarding fast mode could be adding `fast = true` to this code snippet instead of just in the description text (https://libgdx.com/wiki/tools/texture-packer#automatic-packing) and perhaps a log message if packing took a long time, that a fast mode exists as an alternative.
The snippet is still a bit suboptimal though, because it will slow down every application start, even if textures don't need to be repacked. Do you think adding the capability to only repack when folder contents changed is something belonging in the scope of libGDX?

In any case, this PR aims to improve performance in both modes without compromising texture density. If you’d like, I can set up a small test suite to better confirm these results.

Tom-Ski:
Unfortunately this is still in main repo, otherwise its easier to make a breaking change like this.

I think release notes, and changing docs is best thing to do here to advise and bring awareness for this for those that need to pack often where it becomes a problem. Log message also sounds good! For both modes. It can state which mode its in, and if its in 'non fast' it can suggest to use the other for development purposes if over some threshold maybe?

Regarding only packing if modified, processIfModified should be handling that case, or is there something it doesn't do?

Frotty:
> Unfortunately this is still in main repo, otherwise its easier to make a breaking change like this.
> 
> I think release notes, and changing docs is best thing to do here to advise and bring awareness for this for those that need to pack often where it becomes a problem. Log message also sounds good! For both modes. It can state which mode its in, and if its in 'non fast' it can suggest to use the other for development purposes if over some threshold maybe?
> 
> Regarding only packing if modified, processIfModified should be handling that case, or is there something it doesn't do?

Thanks, I will add changelog entry and some logging this week. `processIfModified` does indeed look like it might be sufficient. I guess I didn't find/try it before, maybe because it has no JavaDoc and is also not mentioned on https://libgdx.com/wiki/tools/texture-packer

Tom-Ski:
Yeah, there is no reason really not to encourage that if people want to pack on every run, since its most of the time not actually going to pack.

There is a small caveat, which is if you have assets in some kind of shared drive, or something with history, and you roll back the asset (reducing its last modified date), it wont detect it. But its a small quirk.

Tom-Ski:
Not specifically for this PR, but for the topic. Do you have the profiling breakdown for operations regarding packing? You say most of  the time spent and the focus here is on the packing logic itself, but what about the IO? Is it a considerable amount, or the packing logic just completely outscales it?

Frotty:
> Yeah, there is no reason really not to encourage that if people want to pack on every run, since its most of the time not actually going to pack.
> 
> There is a small caveat, if i remember correctly, which is if you have assets in some kind of shared drive, or something with history, and you roll back the asset (reducing its last modified date), it wont detect it. But its a small quirk.

Yeah, for my own implementation I went for hash-based change detection instead of just using time stamps to avoid such issues. Probably a rare edge case, as you mentioned, but trusting file system metadata can be a bit iffy. If it detects all other "changed" situation it should be ok - will test this a bit though.

> Not specifically for this PR, but for the topic. Do you have the profiling breakdown for operations regarding packing? You say most of the time spent and the focus here is on the packing logic itself, but what about the IO? Is it a considerable amount, or the packing logic just completely outscales it?

That depends on the input. In my case not, but for @tommyettinger 's data the File IO is now taking longer than the actual packing, especially with `limitMemory` enabled, because it has to load the hundreds of files two times. What do you think about logging the time for the different steps? I.e. load, pack, write.

NathanSweet:
I gave this a try on one of my larger atlases, as well as #7445:

* original: 58.2s
* prune: 62s
* rtree: 8.9s

I don't see improvements from prune/#7445 with these 731 images but rtree/this PR is a huge difference! Kudos! I'm using limit memory (very fast SSDs and CPU), fast is off.

Here are most of the packed images (a big image is cut off for privacy and there are 2 other pages with just a few images each):





I ran it with debug lines on, then filled the transparent area pink and background black so it's easier to see what is going on. Prune is pretty close, rtree is notably less tight. It may be aesthetic only and it still fit in the same number of pages, which is the crucial thing.

I packed the 772 Spine UI images with rtree + prune and they packed identically to original (did it 3 times to be sure) in ~0.4s less time.

This is a great change, but I wonder if it should be the `fast` mode since it (at least _appears_) to pack slightly less tightly?

Frotty:
@NathanSweet I haven't looked into this for a while because my libGDX project kinda went on hiatus. Would limiting it to fast mode be the easiest way to get this merged? I wanted to look into the suboptimal packing but haven't had the time.
Cheers.

NathanSweet:
Yeah I think having it be the `fast` mode is the safest way to merge.

PR #7576, opened 2025-01-22
Add Ellipse#toString()

soundasleep:
NOTE I haven't actually tested this at runtime (I don't have a working libgdx development environment at the moment), but this should be a pretty easy addition


Frosty-J:
Might wanna copy from another implementation such as:

https://github.com/libgdx/libgdx/blob/435da504c87b735778e89dca44ba1812fadffb55/gdx/src/com/badlogic/gdx/math/Rectangle.java#L372-L376

soundasleep:
Happy to change it from `x,y,w,h` to `[x,y,w,h]`, I originally copied the Circle toString

Frosty-J:
Ah. I don't know why they're different. Maybe it's fine how you've done it, then. The wonders of libGDX's shape classes!

Issue #597, opened 2013-09-14
[box2d] groupIndex on fixtureDef.filter not functioning correctly


badlogic:
_From [dasac...@gmail.com](https://code.google.com/u/105164384981209237729/) on June 08, 2010 06:19:24_

What steps will reproduce the problem? 1. Create two seperate bodies
2. Create fixtureDef's and set .filter.groupIndex = -8
3. createFixture's on each body
4. Watch them still collide What is the expected output? What do you see instead? setting the fixture's groupIndex for two bodies to the same negative number is suppose to prevent the bodies from colliding. Setting it for this jni wrapper appears to do nothing. What version of the product are you using? On what operating system? libgdx0.5

_Original issue: http://code.google.com/p/libgdx/issues/detail?id=6_

PR #7578, opened 2025-01-27
Add option to use Box2D native ContactFilter for performance optimization

obigu:
On Box2D `World` class we have this [FIXME](https://github.com/libgdx/libgdx/blob/435da504c87b735778e89dca44ba1812fadffb55/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java#L268-L271). It's been there for many years and I guess the idea was to make it an option to use the default native ContactFilter for performance reasons. 

Currently, every time the decision whether two fixtures are to collide or not, a java upcall to `contactFilter (long fixtureA, long fixtureB)` is made. If the user has defined a custom `ContactFilter` it will be executed, if not, the default Box2D logic based on category, mask and groups to determine if they should collide is run on Java side. 

Looking for the reason on why we are executing the default shouldCollide Box2D logic on the Java side I've found this https://github.com/libgdx/libgdx/issues/597 which seems to suggest a bug in Box2D at the time related to negative group ids (haven't tested or investgated this particular issue). I ignore the reason why this was not fixed on the Box2d library itself but at the time the solution was to always run the shouldCollide logic on Java and the FIXME was added. 

The problem with this approach is that Java calls from JNI are quite expensive and there are scenarios in which potentially thousands of calls are being made per jniStep causing a significant performance impact.

To give some numbers of a real world example, on a brick breaker game I've stressed test it to see the maximum number of concurrent balls while keeping 60fps.

**iOS iPhone 12 Pro**
Before: 2000 balls
After: 3500 balls

**Android Pixel 6**
Before: 1900 balls
After: 2900 balls

As you can see, on some scenarios the performance gains are Massive.

With this change it is possible to set the default native contact filter by calling `setContactFilter(null)`. This is the least breaking approach as it requires that the user explicitly makes the call. Ommiting it will default to previous Java default resolution (`contactFilter()`) but we can decide at some point if we want it to be default.

Acknowledgments to @Berstanio for suggesting this approach and support

---

**Important**: If this is merged it will require Android projects that use Box2D to add `boolean getUseDefaultContactFilter();` to Proguard to the folowing block:
```
-keepclassmembers class com.badlogic.gdx.physics.box2d.World {
   boolean contactFilter(long, long);
   boolean getUseDefaultContactFilter();
   void    beginContact(long);
   void    endContact(long);
   void    preSolve(long, long);
   void    postSolve(long, long);
   boolean reportFixture(long);
   float   reportRayFixture(long, float, float, float, float, float);
}
```
Liftoff affected as well @tommyettinger

### DIFF
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
@@ -99,7 +99,8 @@ public:
 	{
 		this->env = env;
 		this->obj = obj;
-		useDefaultContactFilter = env->CallBooleanMethod( obj, useDefaultContactFilterID );
+		if( useDefaultContactFilterID != 0 )
+			useDefaultContactFilter = env->CallBooleanMethod( obj, useDefaultContactFilterID );
 	}
 
 	virtual bool ShouldCollide(b2Fixture* fixtureA, b2Fixture* fixtureB)
--- a/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
+++ b/extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
@@ -59,6 +59,7 @@ public final class World implements Disposable {
 
 static jclass worldClass = 0;
 static jmethodID shouldCollideID = 0;
+static jmethodID useDefaultContactFilterID = 0;
 static jmethodID beginContactID = 0;
 static jmethodID endContactID = 0;
 static jmethodID preSolveID = 0;
@@ -91,17 +92,21 @@ class CustomContactFilter: public b2ContactFilter
 private:
 	JNIEnv* env;
 	jobject obj;
+	jboolean useDefaultContactFilter;
 
 public:
 	CustomContactFilter( JNIEnv* env, jobject obj )
 	{
 		this->env = env;
 		this->obj = obj;
+		useDefaultContactFilter = env->CallBooleanMethod( obj, useDefaultContactFilterID );
 	}
 
 	virtual bool ShouldCollide(b2Fixture* fixtureA, b2Fixture* fixtureB)
 	{
-		if( shouldCollideID != 0 )
+		if( useDefaultContactFilter == true )
+			return b2ContactFilter::ShouldCollide( fixtureA, fixtureB );
+		else if( shouldCollideID != 0 )
 			return env->CallBooleanMethod( obj, shouldCollideID, (jlong)fixtureA, (jlong)fixtureB );
 		else
 			return true;
@@ -246,6 +251,7 @@ b2ContactFilter defaultFilter;
 			reportFixtureID = env->GetMethodID(worldClass, "reportFixture", "(J)Z" );
 			reportRayFixtureID = env->GetMethodID(worldClass, "reportRayFixture", "(JFFFFF)F" );
 			shouldCollideID = env->GetMethodID( worldClass, "contactFilter", "(JJ)Z");
+			useDefaultContactFilterID = env->GetMethodID( worldClass, "getUseDefaultContactFilter", "()Z" );
 		}
 	
 		b2World* world = new b2World( b2Vec2( gravityX, gravityY ));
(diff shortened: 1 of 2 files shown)


extensions/gdx-box2d/gdx-box2d/src/com/badlogic/gdx/physics/box2d/World.java
```diff
@@ -91,17 +92,21 @@ class CustomContactFilter: public b2ContactFilter
...
 	JNIEnv* env;
 	jobject obj;
+	jboolean useDefaultContactFilter;
 
 public:
 	CustomContactFilter( JNIEnv* env, jobject obj )
 	{
 		this->env = env;
 		this->obj = obj;
+		useDefaultContactFilter = env->CallBooleanMethod( obj, useDefaultContactFilterID );
```
Berstanio, line 102: This needs a check whether `useDefaultContactFilterID != 0`.
If proguard/r8/native-image strips the method, this might lead to a segfault otherwise.

obigu: > This needs a check whether `useDefaultContactFilterID != 0`. If proguard/r8/native-image strips the method, this might lead to a segfault otherwise.

I've tested and it currently throws a Java `NoSuchMethodException` if Proguard doesn't add it. Won't adding that check swallow the error and the user won't notice as Java version would be executed in that case?

Berstanio: When is the error thrown? When getting the methodID or when trying to call the actual method?
Because I know native-image just segfaults if methods aren't registered for JNI.

obigu: I'd guess it's thrown when trying to execute the actual method as it's classic normal `NoSuchMethodException` Java Exception that specifies the method name in the stacktrace.

obigu:
```
 FATAL EXCEPTION: GLThread 138
                                                                                                    Process: com.xxxx.xxxx, PID: 16140
                                                                                                    java.lang.NoSuchMethodError: no non-static method "Lcom/badlogic/gdx/physics/box2d/World;.getUseDefaultContactFilter()Z"
                                                                                                    	at com.badlogic.gdx.physics.box2d.World.newWorld(Native Method)
                                                                                                    	at com.badlogic.gdx.physics.box2d.World.<init>(SourceFile:122)
                                                                                                    	at t6.g.f(SourceFile:11)
                                                                                                    	at c7.a.l(SourceFile:1)
                                                                                                    	at k7.b0.m(SourceFile:305)
                                                                                                    	at t6.q.a(SourceFile:135)
                                                                                                    	at k7.a.l(SourceFile:3)
                                                                                                    	at k7.r0.a(SourceFile:35)
                                                                                                    	at j7.j.l(SourceFile:11)
                                                                                                    	at k2.e.k(SourceFile:46)
                                                                                                    	at j7.j.k(SourceFile:1)
                                                                                                    	at h2.g.a(SourceFile:103)
                                                                                                    	at h2.h.h(SourceFile:97)
                                                                                                    	at g1.g.h(SourceFile:17)
                                                                                                    	at l1.s.f(SourceFile:196)
                                                                                                    	at l1.i.onDrawFrame(SourceFile:180)
```

obigu: It's ehen calling the Native one :D

Berstanio: >  It's when getting the method id :D

Ahh, I see! Then I think we could add a "0" check should be added, just to be safe in case the world is still getting used for some reason, but I have no strong opinion.

obigu: > Ahh, I see! Then I think we could add a "0" check should be added, just to be safe in case the world is still getting used for some reason, but I have no strong opinion.

Ok, I've added the check even if it should never reach that point with it being 0 as the informative Error will have been thrown before.

Berstanio (approved):

PR #7580, opened 2025-01-28
Add Lwjgl3ApplicationConfiguration.setMacOsX

damn:


Frosty-J:
Sounds like a sensible idea. Is `Configuration.GLFW_CHECK_THREAD0.set(false)` needed as well? Mario (libGDX creator) and Xyper (Minecraft modder) use it.

I'm not entirely sure on the setter name. If someone reads `config.setMacOsX(true)`, it's not self-explanatory what it does, and the Mac OS X name hasn't been used in many years. The oldest maintained version is macOS 13 - they've dropped the X/10 altogether.

`setMacOsXConfig()` doesn't seem worth splitting into a separate method, not when it's private. It doesn't save any lines in the constructor:

```java
public Lwjgl3Application (ApplicationListener listener, Lwjgl3ApplicationConfiguration config) {
  if (SharedLibraryLoader.os == Os.MacOsX) {
    setMacOsXConfig(config.macOsX);
  }
  // ...
}

private void setMacOsXConfig (boolean macOsX) {
  boolean useGlfwAsync = macOsX;
  if (useGlfwAsync) {
    Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
  }
}
```

vs

```java
public Lwjgl3Application (ApplicationListener listener, Lwjgl3ApplicationConfiguration config) {
  if (SharedLibraryLoader.os == Os.MacOsX && config.macOsX) {
    Configuration.GLFW_LIBRARY_NAME.set("glfw_async")
  }
  // ...
}
```

damn:
> Sounds like a sensible idea. Is `Configuration.GLFW_CHECK_THREAD0.set(false)` needed as well? Mario (libGDX creator) and Xyper (Minecraft modder) use it.

Works for me on MacOs Ventura without that line. I have also used another OS version where I forgot the name and it also works without.

> I'm not entirely sure on the setter name. If someone reads `config.setMacOsX(true)`, it's not self-explanatory what it does, and the Mac OS X name hasn't been used in many years. The oldest maintained version is macOS 13 - they've dropped the X/10 altogether.

Sure I will remove the 'X'. Although check out my other comment below which might make the configuration unnecessary.

> `setMacOsXConfig()` doesn't seem worth splitting into a separate method, not when it's private. It doesn't save any lines in the constructor:
> 
> ```java
> public Lwjgl3Application (ApplicationListener listener, Lwjgl3ApplicationConfiguration config) {
>   if (SharedLibraryLoader.os == Os.MacOsX) {
>     setMacOsXConfig(config.macOsX);
>   }
>   // ...
> }
> 
> private void setMacOsXConfig (boolean macOsX) {
>   boolean useGlfwAsync = macOsX;
>   if (useGlfwAsync) {
>     Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
>   }
> }
> ```
> 
> vs
> 
> ```java
> public Lwjgl3Application (ApplicationListener listener, Lwjgl3ApplicationConfiguration config) {
>   if (SharedLibraryLoader.os == Os.MacOsX && config.macOsX) {
>     Configuration.GLFW_LIBRARY_NAME.set("glfw_async")
>   }
>   // ...
> }
> ```

Initially I had also added another config option for setting the `java.awt.Taskbar` icon, which is the dock-icon, but it needs a higher level Java version than libgdx supports.

My idea was that the reader of the `Lwjgl3Application` constructor is not interested in any details of mac-os specific configuration - so that is why I have created a separate method. That way we can extend the mac-os configuration without touching the code in the constructor.

I have also thought that maybe the setting for `glfw_async` should be by default - that way users on mac do not need to configure anything at all ... we may even omit the whole config setting in this case? I am not knowledgable enough about GLFW to know if that is a good idea.

Frosty-J:
> Initially I had also added another config option for setting the `java.awt.Taskbar` icon, which is the dock-icon, but it needs a higher level Java version than libgdx supports.

Shame. I would've hoped for `Lwjgl3WindowConfiguration#setWindowIcon` to affect this.

> I have also thought that maybe the setting for `glfw_async` should be by default

This is something I'd need a big macOS user to weigh in on. Maybe it can break window resizing, but how and why I do not know. Any alternative to needing `-XstartOnFirstThread` I'd like very much.

tommyettinger:
I had tried enabling glfw_async not too long ago for Liftoff itself (not projects using it), and IIRC it led to lots of crashes when people resized the window. It could have been fixed somewhat since then, though the state of LWJGL3 recently doesn't give me much hope. The StartupHelper class that Liftoff distributes with every LWJGL3 project does help users quite a lot, though it interferes with debugging a little. And even debugging is handled by the Gradle `lwjgl3:run` task in generated projects automatically using `-XstartOnFirstThread` . So, I'd mostly like to know what this does differently from and better than StartupHelper...

Also, I really don't like it when PRs don't have a single word in the description. Even a sentence or two takes negligible time and effort, and the PR title doesn't explain anything. I'll look at the code in a minute, *but I really shouldn't have to*.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -234,6 +236,14 @@ public void enableGLDebugOutput (boolean enable, PrintStream debugOutputStream)
 		debugStream = debugOutputStream;
 	}
 
+	/** Configuration which will be set on if the {@link Os} is {@link Os/MacOsX}.
+	 *
+	 * @param useGlfwAsync Whether to use the experimental "glfw_async" library. This means you do not have to set the JVM argument
+	 *           "-XstartOnFirstThread". */
+	public void setMacOsX (boolean useGlfwAsync) {
```
tommyettinger, line 243: Am I understanding this right? It looks like:
```java
cfg.setMacOsX(true);
```
is exactly equivalent to:
```java
if (SharedLibraryLoader.os == Os.MacOsX) Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
```
And setting it to false does nothing. The equivalent (longer) line doesn't need changes to libGDX to test, so hopefully we can get some macOS testers to try it out.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -234,6 +236,14 @@ public void enableGLDebugOutput (boolean enable, PrintStream debugOutputStream)
 		debugStream = debugOutputStream;
 	}
 
+	/** Configuration which will be set on if the {@link Os} is {@link Os/MacOsX}.
```
tommyettinger, line 239: The second link is invalid; it should be `{@link Os#MacOsX}` . Don't ask me why the syntax is weird and different, I don't know!

tommyettinger (changes requested):
I only have one change I'm actually requesting, regarding invalid JavaDoc, but I'm still very much on the fence about anything relating to LWJGL3 on Mac and its first thread problem. As far as I'm concerned, the solution used in Liftoff... works fine? And I'd rather bring StartupHelper into the LWJGL3 backend than try something else that a) we are not sure if it works as well, and b) doesn't include the fixes that StartupHelper applies for Windows.

damn:
I have thought about this PR again and I am thinking to remove this commit and create a new branch with just one public static method added:

```java
	/**
	 * Whether to use the experimental "glfw_async" library.
   * This means you do not have to set the JVM argument "-XstartOnFirstThread" on mac os.
   * Use after checking with {@link SharedLibraryLoader} if {@link Os} is {@link Os/MacOs}.
   *
   * @see <a href= "https://javadoc.lwjgl.org/org/lwjgl/glfw/package-summary.html#using-glfw-on-macos-heading"> LWJGL GLFW Documentation</a>
	 */
	public static void setGlfwAsync() {
		 Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
	}

```

Reasoning is that lwjgl should be abstracted by this backend so the user should not call into lwjgl classes himself.

Also if for example `Lwjgl3ApplicationConfiguration.getMonitors` is called before the application is started, it will still fail if `xStartOnFirstThread` is not enabled because `getMonitors` initializes glfw.

damn:
I have been reading your comment again, I didn't know about any startup helper from liftoff.

That might be a good idea too for starting desktop lwjgl apps to include it here, I'll check it out.

Edit:
I think this is the code 
https://github.com/libgdx/gdx-liftoff/blob/master/src/main/java/gdx/liftoff/StartupHelper.java

crykn:
Hey, sorry for the late review, but I think this is generally a good addition. Three minor nitpicks:
- I'd change the method name to something more descriptive like "useGlfwAsync"
- I think the link in the method's javadoc should point to a libGDX specific ressource, i.e. https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else
- The test change isn't really needed, because the -XstartOnFirstThread argument is provided via Gradle already (and there are still some minor bugs with the experimental async library)

Edit: Another test change is warranted though at [AwtTestLWJGL.java#L62](https://github.com/libgdx/libgdx/blob/e063c6dc2cdc078de30c97f7d7aedd574b5efc35/tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/AwtTestLWJGL.java#L62C4-L62C49).

damn:
> Hey, sorry for the late review, but I think this is generally a good addition. Three minor nitpicks:
> 
>     * I'd change the method name to something more descriptive like "useGlfwAsync"
> 
>     * I think the link in the method's javadoc should point to a libGDX specific ressource, i.e. https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else
> 
>     * The test change isn't really needed, because the -XstartOnFirstThread argument is provided via Gradle already (and there are still some minor bugs with the experimental async library)
> 
> 
> Edit: Another test change is warranted though at [AwtTestLWJGL.java#L62](https://github.com/libgdx/libgdx/blob/e063c6dc2cdc078de30c97f7d7aedd574b5efc35/tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/AwtTestLWJGL.java#L62C4-L62C49).

I have made those changes and created a new PR with only one commit for a clean git history: https://github.com/libgdx/libgdx/pull/7659

PR #7581, opened 2025-01-31
Replacing the random class with an improved version

trietng:
I added a new class named `RandomXRSR128`. This class uses the `xoroshiro128++` algorithm instead of the older `xorshift128+` algorithm currently in use. `xoroshiro128++` has been [included](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/package-summary.html) in Java standard API since version 17.

According to the [authors](https://prng.di.unimi.it/), `xoroshiro128++` is better than their previous xorshift-based algorithms in both speed and statistical quality. I also modified the internal seed generation process (`setSeed`) to use SplitMix with a Phi offset after the first hashing function instead of the current method of double-hashing.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/RandomXRSR128.java
+++ b/gdx/src/com/badlogic/gdx/math/RandomXRSR128.java
@@ -1,0 +1,207 @@
+/*******************************************************************************
+ * Copyright 2011 See AUTHORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.math;
+
+import java.util.Random;
+
+/** This class implements the xoroshiro128++ pseudo random number generator. Adapted from
+ * Sebastiano Vigna's <a href="https://github.com/vigna/dsiutils">disutils</a>. Note that
+ * xoroshiro128++ is a part of the Java API from version 17 and above.
+ * <p>
+ * Instances of RandomXRSR128 are not thread-safe.
+ *
+ * @author trietng
+ */
+public class RandomXRSR128 extends Random {
+  /** Normalization constant for double. */
+  private static final double NORM_DOUBLE = 1.0 / (1L << 53);
+
+  /** Normalization constant for float. */
+  private static final float NORM_FLOAT = (float) (1.0 / (1L << 24));
+
+  /** 2<sup>64</sup> &middot; &phi;, &phi; = (&#x221A;5 &minus; 1)/2. */
+  private static final long PHI = 0x9E3779B97F4A7C15L;
+
+  /** The first half of the internal state of this pseudo-random number generator. */
+  private long seed0;
+
+  /** The second half of the internal state of this pseudo-random number generator. */
+  private long seed1;
+
+  /** Creates a new random number generator. This constructor sets the seed of the random number
+   * generator to a value very likely to be distinct from any other invocation of this constructor.
+   * <p>
+   * This implementation creates a {@link Random} instance to generate the initial seed. */
+  public RandomXRSR128 () {
+    setSeed(new Random().nextLong());
+  }
+
+  /** Creates a new random number generator using a single {@code long} seed.
+   * @param seed the initial seed */
+  public RandomXRSR128 (final long seed) {
+    setSeed(seed);
+  }
+
+  /** Creates a new random number generator using two {@code long} seeds.
+   * @param seed0 the first part of the initial seed
+   * @param seed1 the second part of the initial seed */
+  public RandomXRSR128 (final long seed0, final long seed1) {
+    setState(seed0, seed1);
+  }
+
+  /** Returns the next pseudo-random, uniformly distributed {@code long} value from this random
+   * number generator's sequence.
+   * <p>
+   * Subclasses should override this, as this is used by all other methods. */
+  @Override
+  public long nextLong () {
+    final long s0 = this.seed0;
+    long s1 = this.seed1;
+
+    final long result = Long.rotateLeft(s0 + s1, 17) + s0;
+
+    s1 ^= s0;
+    this.seed0 = Long.rotateLeft(s0, 49) ^ s1 ^ s1 << 21;
+    this.seed1 = Long.rotateLeft(s1, 28);
+    return result;
+  }
+
+  /** This protected method is final because, contrary to the superclass, it's not used anymore by
+   * the other methods. */
+  @Override
+  protected final int next (final int bits) {
+    return (int)(nextLong() & ((1L << bits) - 1));
+  }
+
+  /** Returns the next pseudo-random, uniformly distributed {@code int} value from this random
+   * number generator's sequence.
+   * <p>
+   * This implementation uses {@link #nextLong()} internally. */
+  @Override
+  public int nextInt () {
+    return (int)nextLong();
+  }
+
+  /** Returns a pseudo-random, uniformly distributed {@code int} value between 0 (inclusive) and
+   * the specified value (exclusive), drawn from this random number generator's sequence.
+   * <p>
+   * This implementation uses {@link #nextLong()} internally.
+   * @param n the positive bound on the random number to be returned.
+   * @return the next pseudo-random {@code int} value between {@code 0} (inclusive) and {@code n}
+   * (exclusive). */
+  @Override
+  public int nextInt (final int n) {
+    return (int)nextLong(n);
+  }
+
+  /** Returns a pseudorandom uniformly distributed {@code long} value between 0 (inclusive) and the
+   * specified value (exclusive), drawn from this random number generator's sequence. The algorithm
+   * used to generate the value guarantees that the result is uniform, provided that the sequence
+   * of 64-bit values produced by this generator is.
+   * @param n the positive bound on the random number to be returned.
+   * @return the next pseudorandom {@code long} value between {@code 0} (inclusive) and {@code n}
+   * (exclusive).
+   */
+  public long nextLong(final long n) {
+    if (n <= 0) throw new IllegalArgumentException("illegal bound " + n + " (must be positive)");
+    long t = nextLong();
+    final long nMinus1 = n - 1;
+    // Rejection-based algorithm to get uniform integers in the general case
+    for (long u = t >>> 1; u + nMinus1 - (t = u % n) < 0; u = nextLong() >>> 1);
+    return t;
+  }
+
+  /** Returns a pseudo-random, uniformly distributed {@code double} value between 0.0 and 1.0 from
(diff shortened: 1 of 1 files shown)


tommyettinger (changes requested):
So... this is a topic near and dear to me. I've actually suggested adding better RNGs than RandomXS128 (ones that pass statistical testing via PractRand) before, and was turned down enough times that I've just stopped asking. There are a few things to point out here, and none of them are obvious flaws -- some are subtle flaws, or easily fixed, though.

 - Long.rotateLeft() is incredibly slow on GWT. [The code is atrocious](https://github.com/gwtproject/gwt/blob/main/user/super/com/google/gwt/emul/java/lang/Long.java#L130-L149), and [the repo owner refuses to change it](https://github.com/gwtproject/gwt/issues/9563) (though it's in a milestone that keeps getting pushed back). It is much faster on GWT and about the same amount of code to replace `Long.rotateLeft(s0, 49)` with `(so << 49 | s0 >>> 64 - 49)`. The idea that Oracle invented the circular shift is completely bogus; they cannot possibly hold claims on something that was well-known in the 1980s, long before Java existed.
   - Also, on x64 platforms at least, `(x << n | x >>> 64 - n)` compiles to one ROL instruction, thanks to HotSpot.
 - `setSeed()` doesn't need to do three unary hashes (mixers); two calls to staffordMix13() will suffice as long as the inputs are different. The key thing is that as long as the unary hash is bijective (has one output for every input and vice versa), then running it on two different inputs will get two different outputs, one for each state. And that's important because the case where `seed0 == 0L && seed1 == 0L` is the only disallowed state.
 - That brings me to another point; `setState(long, long)` should check that both states are not 0, but if they are, it should set one to some different value, like `-1L`. If the generator is set into the 0,0 state, then every result it returns will be 0L, forever. Obviously we want to avoid this!
 - Xoroshiro128++ isn't a slouch when it comes to speed, but it isn't breaking any records. My current fastest algorithm I've written (AceRandom, [which is in my Juniper library for RNGs](https://github.com/tommyettinger/juniper/blob/main/src/main/java/com/github/tommyettinger/random/AceRandom.java#L400-L412)) is about twice as fast as relatives to Xoshiro and Xoroshiro that I've tested already. The performance is roughly comparable on JDK 8, but AceRandom benefits much more from optimizations in JDK 16 and later. AceRandom has 5 `long` states instead of the 2 here, but because it performs 5 operations that are independent of each other's progress, it takes full advantage of instruction-level parallelism (ILP) in modern processors. AceRandom can be handled with 5 operations all being processed simultaneously, with no data dependency between states. On the other hand, xoroshiro requires a minimum chain of 3 operations in sequence, assuming maximum ILP, and needs two of those to happen simultaneously to perform its best. The speed difference, in absolute terms, is tiny; each call takes under a nanosecond, but this is something that a specific group of RNG nuts like myself wages battles over intensely.

This is a good port of xoroshiro128++, and there's no way you could be expected to know about the Long.rotateLeft() issues on GWT (believe it or not, Long.rotateRight() is even worse!). Ideally, xoroshiro128++ code would have some form of its jump/leap code, but that stuff gets complex. [I do happen to have implemented that leap() code here for `xoroshiro128**`](https://github.com/tommyettinger/juniper/blob/main/src/main/java/com/github/tommyettinger/random/Xoroshiro128StarStarRandom.java#L241-L278), and the only difference for ++ is the last section after a comment.
I prefer `xoroshiro128**` to ++ because it's 2D-equidistributed (almost exactly), and that's one of the main things the xoshiro and xoroshiro family does that AceRandom doesn't do at all. That means that not just every output occurs equally often, but also that every pair of subsequent outputs occurs equally often. However, xoroshiro128++ is only 1D equidistributed (every individual output happens equally often, but not pairs). AceRandom isn't equidistributed at all because each cycle has unknown length (but guaranteed to be very long in the worst case, at least 2 to the 64 outputs before cycling). Some specific usages can expose worse flaws than what I have listed so far; for example, if you initialize a xoroshiro generator per-cell on a grid, and you use the cell's x for seed0 and y for seed1, there's an obvious grid pattern in the results that won't ever go away (there should be no pattern at all in an optimal generator, or more reasonably the pattern should dissipate quickly).

There's lots of tradeoffs with every generator. It's hard to pick just one as a recommendation, which is kind of why I have Juniper as a whole library for these things. What's more, we can't change the underlying algorithm used by `MathUtils.random` without changing the results of existing seeds, so this would have to be an additional option, not a replacement... and the relatively-lousy RandomXS128 would have to live on...

trietng:
Thank you, @tommyettinger, for giving me such a detailed review. It appears that you are well-versed in this topic!
 
> * Long.rotateLeft() is incredibly slow on GWT. [The code is atrocious](https://github.com/gwtproject/gwt/blob/main/user/super/com/google/gwt/emul/java/lang/Long.java#L130-L149), and [the repo owner refuses to change it](https://github.com/gwtproject/gwt/issues/9563) (though it's in a milestone that keeps getting pushed back). It is much faster on GWT and about the same amount of code to replace `Long.rotateLeft(s0, 49)` with `(so << 49 | s0 >>> 64 - 49)`. The idea that Oracle invented the circular shift is completely bogus; they cannot possibly hold claims on something that was well-known in the 1980s, long before Java existed.

Copyrighting such a fundamental algorithm like circular shifting is ridiculous. WTH?

> * `setSeed()` doesn't need to do three unary hashes (mixers); two calls to staffordMix13() will suffice as long as the inputs are different. The key thing is that as long as the unary hash is bijective (has one output for every input and vice versa), then running it on two different inputs will get two different outputs, one for each state. And that's important because the case where `seed0 == 0L && seed1 == 0L` is the only disallowed state.

I was using the `staffordMix13()` as a "purfier" function. Apparently it's how the authors of `xoroshiro` do it. But you aren't wrong.

> * That brings me to another point; `setState(long, long)` should check that both states are not 0, but if they are, it should set one to some different value, like `-1L`. If the generator is set into the 0,0 state, then every result it returns will be 0L, forever. Obviously we want to avoid this!

True, it's just that the original `RandomXS128` class didn't include the check so I thought there must be some reason they left it that way.

> I prefer `xoroshiro128**` to ++ because it's 2D-equidistributed (almost exactly), and that's one of the main things the xoshiro and xoroshiro family does that AceRandom doesn't do at all. That means that not just every output occurs equally often, but also that every pair of subsequent outputs occurs equally often. However, xoroshiro128++ is only 1D equidistributed (every individual output happens equally often, but not pairs). AceRandom isn't equidistributed at all because each cycle has unknown length (but guaranteed to be very long in the worst case, at least 2 to the 64 outputs before cycling). Some specific usages can expose worse flaws than what I have listed so far; for example, if you initialize a xoroshiro generator per-cell on a grid, and you use the cell's x for seed0 and y for seed1, there's an obvious grid pattern in the results that won't ever go away (there should be no pattern at all in an optimal generator, or more reasonably the pattern should dissipate quickly).

Can't disagree with you on that. There **is** an argument for `xoroshiro128**` over `xoroshiro128++`. Microsoft did after all pick `xoroshiro128**` for their newer .NET implementations, and .NET is Java on steroid. Still, quoting the author, `xoroshiro128++` is slightly faster than `xoroshiro128**`. We don't really need absolute optimality for a non-cryptograhically secured PRNG, do we?

tommyettinger:
C# and .NET 9 are indeed doing a very good job at optimization! It's pretty much impossible to compare benchmarks using different techniques, but I have [a benchmark of 17 C# generators doing one task](https://github.com/tommyettinger/ShaiRandom/blob/fda74af69ab59a3bc07f054b2099c4fdfa36779c/ShaiRandom.PerformanceTests/Benchmarks.cs#L583) and the unseeded one, surprisingly, is second-slowest. I think all the rest of the generators are taking so little time that there's no way to measure the effects of requesting a ulong with a bound vs. just requesting any possible ulong. The times were about the same for an unbounded ulong, which is the normal format these generators handle (like `xoshiro256**`, which is what Unseeded uses internally). I should note that the later 15 generators all seem to take either between 0.1 and 0.2 ns, or between 0.3 and 0.4 ns, for any task requested, so there's some imprecision when determining the actual speed of any generator, and the best and worst of the 15 vary between runs a lot.

The [Java benchmark here](https://github.com/tommyettinger/assorted-bumble-benchmarks/blob/main/src/main/java/net/adoptopenjdk/bumblebench/examples/AceRandomBench.java#L28-L32) uses BumbleBench, which doesn't test in the same way, and needs some assignment to take place so the whole loop doesn't get ignored. There are various benchmarks in that folder, all to some extent outdated or not equivalent to the C# ones, but if BumbleBench and Benchmark.NET had the same amount of overhead to a call to nextLong() or NextULong(), then the .NET version would be about twice as fast. I think BumbleBench adds some time to each call because it needs that assignment, so the difference is smaller than that.

But, back to Java land! The ++ and `**` generators replaced earlier xoroshiro generators, which replaced earlier xorshift+ generators (which libGDX still uses, but so do web browsers).  The xorshift+ and xoroshiro generators both fail a specific test for randomness almost immediately, a Binary Matrix Rank test. That is a test on the least significant bit of the returned numbers, typically, but they also fail on some other bits. I'm pretty sure the main flaw of the `**` generators is that if you multiply the results by one of... a lot.. of multipliers, then the modified output suddenly goes back to failing Binary Matrix Rank tests. I believe the last 7 or 8 bits of the multiplier have to match some specific odd number, and if they do it will not produce especially random output. The ++ generators don't have that flaw, but aren't as equidistributed either.

I've never written a generator with 2D or greater equidistribution. That is hard work, and it involves a lot of math on matrices. xoroshiro and xoshiro with the ** scrambler do that, though. It's also quite easy to change, if you want to, since what makes the result is usually just one line. In any case, all the generators that have been mentioned so far, from xoroshiro128++ to `xoshiro256**` to AceRandom and so on, are more resilient against flaws that would let people "cheat" at competitive games than the current RandomXS128. I'd say any `**` scramblers help the least against cheating, really irregular generators like AceRandom help more, and the best ones either truncate output (slowing down the generator but making the state impossible to guess for a long time), or mix multiple states to get one output. The simplest one of the last kind I have is [FlowRandom](https://github.com/tommyettinger/juniper/blob/main/src/main/java/com/github/tommyettinger/random/FlowRandom.java#L216-L222), which is the same type of algorithm as SplittableRandom (and that just uses staffordMix13() or something like it on a large-increment counter). FlowRandom is different because it uses two `long` states that share one cycle, with the same cycle length as SplittableRandom. Where SplittableRandom leaves one state unchanged and requires it to be an odd number (and some odd numbers do not work well with it; which ones aren't clear), FlowRandom changes both and merges 128 bits of changing state into 64 bits of output. That makes it harder to work back from an output to the original states. The ++ scramblers also do this, but AceRandom and the `**` scramblers don't. FlowRandom has 2 to the 64 possible cycles, each with 2 to the 64 outputs that are not equidistributed (per stream, anyway).  That last part may be useful in practice when trying to stop cheaters, because figuring out what stream you are on requires knowing both current states, and that's very hard to figure out given only output. If the stream is picked at random, then on average the results of a FlowRandom will not be biased toward high or low numbers, or toward any particular value, even though any individual stream will probably have a different sum over its entire cycle than a counter or an equidistributed generator. FlowRandom is about the same speed as SplittableRandom, which is comparable to xoroshiro generators in Java last I checked (it has been a while).

So I guess my point in all of this is either we figure out a goal for what each generator should be used for and add it in its own PR if we think that goal would help libGDX, or leave random number generation to a dedicated library like [juniper](https://github.com/tommyettinger/juniper) or [cringe](https://github.com/tommyettinger/cringe). The last one also has other commonly-used types of code related to RNGs, like noise functions. I refuse to make the README.md less cringey, however!

trietng:
I see. Good point with the noise function. Unity does have the noises function included.

PR #7583, opened 2025-02-01
Update LWJGL to 3.3.6 since AV problem has been fixed.

colorizenl:
https://github.com/libgdx/libgdx/pull/7555/files downgrades LWJGL from 3.3.5 to 3.3.3 to solve the issues with antivirus tools were misreporting problems with LWJGL's JAR files.

These problems have been fixed in LWJGL 3.3.6, as explained in https://github.com/LWJGL/lwjgl3/issues/1005, so I believe it is now safe for libGDX to once again use the latest version.


obigu:
3.3.6 still had the OpenAL issue. Should be fixed in 3.3.7.

PR #7584, opened 2025-02-01
Adding new a noise generation class 

trietng:
I added a new noise generator, [FastNoiseLite](https://github.com/Auburn/FastNoiseLite), to the `math` module. The imported library is nearly identical to the original Java port excepts for some refactoring and formatting changes. There is also one new method named `getNoise1D`, which is just a wrapper for `getNoise2D(x, 0.0f)`.

Some relevant information on `FastNoiseLite`:
* `FastNoiseLite` uses the `OpenSimplex2` as the default noise generation algorithm, which is pattern-free for *all* use cases.
* `FastNoiseLite` has been included in the Godot game engine since 2022.


gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
```diff
@@ -0,0 +1,2353 @@
...
+		0.4881873305f, -0.8727388672f, -0.3683854996f, -0.9296731273f, -0.9825390578f, 0.1860564427f, 0.81256471f, 0.5828709909f,
+		0.3196460933f, -0.9475370046f, 0.9570913859f, 0.2897862643f, -0.6876655497f, -0.7260276109f, -0.9988770922f, -0.047376731f,
+		-0.1250179027f, 0.992154486f, -0.8280133617f, 0.560708367f, 0.9324863769f, -0.3612051451f, 0.6394653183f, 0.7688199442f,
+		-0.01623847064f, -0.9998681473f, -0.9955014666f, -0.09474613458f, -0.81453315f, 0.580117012f, 0.4037327978f, -0.9148769469f,
+		0.9944263371f, 0.1054336766f, -0.1624711654f, 0.9867132919f, -0.9949487814f, -0.100383875f, -0.6995302564f, 0.7146029809f,
+		0.5263414922f, -0.85027327f, -0.5395221479f, 0.841971408f, 0.6579370318f, 0.7530729462f, 0.01426758847f, -0.9998982128f,
+		-0.6734383991f, 0.7392433447f, 0.639412098f, -0.7688642071f, 0.9211571421f, 0.3891908523f, -0.146637214f, -0.9891903394f,
+		-0.782318098f, 0.6228791163f, -0.5039610839f, -0.8637263605f, -0.7743120191f, -0.6328039957f,};
+
+	private static final float[] Gradients3D = {0, 1, 1, 0, 0, -1, 1, 0, 0, 1, -1, 0, 0, -1, -1, 0, 1, 0, 1, 0, -1, 0, 1, 0, 1, 0,
```
tommyettinger, line 435: Oh, I was confused about Gradients3D vs. RandVecs3D. The gradients are the centers of the 12 edges of a cube, repeated several times but not completely (the section `1, 0, 1, 0, -1, 0, 1, 0, 1, 0, -1, 0, -1, 0, -1, 0, 1, 1, 0, 0, -1, 1, 0, 0, 1, -1, 0, 0, -1, -1, 0, 0,` is repeated once less than the first four vertices). This is different from [what OpenSimplex2 has done for the last four years](https://github.com/KdotJPG/OpenSimplex2/blob/master/java/OpenSimplex2.java#L523-L581)! It's hard to tell what effect the gradient vectors have without seeing a preview with those very regular cube-edge vectors vs. well-distributed quasi-random 3D vectors vs. another polyhedron (I have used a rhombic triacontahedron's vertices before; that and a pentakis dodecahedron each have 32 vertices, so their vertices can be repeated without losing any to truncation). It's entirely possible there's no problem with these that humans can see. Analyzing noise in frequency space with an FFT is fun though, if you have an interesting definition of "fun!" That's where I'd expect to notice any quirks.

gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
```diff
@@ -0,0 +1,2353 @@
...
+	 * @param domainWarpAmp warp amplitude (default is {@code 1.0}) */
+	public void setDomainWarpAmp (final float domainWarpAmp) {
+		mDomainWarpAmp = domainWarpAmp;
+	}
+
+	/** 1D noise at given position using current settings. Equivalent to {@code getNoise1D(x, 0.0f)}.
+	 * @param x X coordinate
+	 * @return noise output bounded between -1...1 */
+	public float getNoise1D (float x) {
+		return getNoise2D(x, 0.0f);
```
tommyettinger, line 211: There will be noticeable visual artifacts if using this with one octave of Perlin noise, and probably some artifacts with more Perlin octaves. At every integer grid point in 2D, single-octave Perlin evaluates to 0, but in 1D this is much more noticeable because rotations can't hide how regular it is. I am not sure how this affects OpenSimplex2 other than the frequency will be different from the expected one for Perlin, and it probably won't be moving along a grid-aligned path.

tommyettinger: For 1D noise, I have generally had better luck using a dedicated algorithm. Some of these are really extremely simple! [This one isn't as simple](https://github.com/tommyettinger/juniper/blob/main/src/main/java/com/github/tommyettinger/random/LineWobble.java#L203-L238), but it has extraordinarily smooth changes in direction. It's related to ValueCubic here.

gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
```diff
@@ -0,0 +1,2353 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.math;
+
+// MIT License
```
tommyettinger, line 19: I'm pretty sure we're totally fine on licensing here as long as the MIT license and its authorship credit are intact. I also believe libGDX already uses MIT-licensed code somewhere, because so much open source code uses it.

gdx/res/com/badlogic/gdx.gwt.xml
```diff
@@ -354,6 +354,7 @@
 		<include name="math/DelaunayTriangulator.java"/>
 		<include name="math/EarClippingTriangulator.java"/>
 		<include name="math/Ellipse.java"/>
+		<include name="math/FastNoiseLite.java"/>
```
tommyettinger, line 357: Good catch on the .gwt.xml file! This is an easy file to miss when updating. This is good!

tommyettinger (commented):
As much as I have looked over this today, I don't feel like I can approve this or request specific changes yet. This is a **really** large file. I used to use a similarly massive file derived from the v1 of FastNoise, and extended in bits and pieces over the years. It was not a great way to work. My current approach for both of my active noise-related codebases ([SquidSquad](https://github.com/yellowstonegames/SquidSquad/tree/main/squidgrid) and [cringe](https://github.com/tommyettinger/cringe)) is to have a shared interface or abstract class that all noise algorithms can implement or extend (including user-added ones). In cringe, there's a good example in how the abstract class RawNoise provides a skeletal basis for single-octave noise, and the ContinuousNoise class wraps another RawNoise to provide multiple octaves, FBM/ridged/billow/domain-warp, and so on when noise is requested. Simplex noise has one class, Perlin noise has another class, OpenSimplex2 has two classes (the fast kind and the smooth kind), etc. SquidSquad uses a similar approach (using an INoise interface instead of RawNoise), and also has alternatives to ContinuousNoise, like one that produces radially symmetrical noise loops.

My noise code typically supports 1D through 6D noise, unless a particular algorithm has no working version for some dimensions. For example, OpenSimplex2 will probably never have 5D or 6D versions, because the math gets vastly more complex as dimension goes up; some generators are at least theoretically capable of producing noise in arbitrary dimensions without changing code, and though every type of noise gets slower as dimension goes up, they do so at different rates. All of my code for the last 4 years or so strongly emphasizes serialization, because I have seen how much of a problem this can be in other codebases. My noise code in particular has had some tricky changes needed to handle serializing generators wrapped by other generators; I know this works in SquidSquad, but I need to see if I made the changes to cringe as well. I can't say my approach is strictly better; it is probably slower on some tasks at the moment, and it doesn't offer identical features. But, it can be extended by users, each generator is in a manageable file, and it supports 4D, 5D, and 6D noise. I'd love to show off what 5D noise can do, but that's for another time.

Mostly, my concerns are about the implementation in libGDX being treated as official and so becoming more widely used, but also being impossible to extend and inflexible. Adding a small interface would help, but FastNoiseLite is organized as a very, very large API surface for one class. This is going to require some careful thinking about how we expect the class to be used, or not used in the case of its presence in MathUtils.

gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
trietng: Not sure if I can do any verification on this lol. I'm a CS major, not a math major (unfortunately). Like I mentioned in the PR comment, this is lifted straight from the original Java port of FastNoiseLite. Not a single number or math operation was changed.

I'm just interested in having noise generation in `libgdx` since I noticed that it was lacking compared to its counterparts like Godot or Unity. And I believe noise generation is commonly used in a lot of games that have procedural generation like Minecraft, etc.

gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
trietng: I believe this is also the approach taken by Godot. Using an entirely different algorithm for just `getNoise1D` would complicate the API interface, so it's this or removing the 1D generation entirely.

gdx/src/com/badlogic/gdx/math/FastNoiseLite.java
```diff
@@ -0,0 +1,2353 @@
...
+			zNSign = -zNSign;
+
+			seed += 1293373;
+		}
+
+		coord.x += vx * warpAmp;
+		coord.y += vy * warpAmp;
+		coord.z += vz * warpAmp;
+	}
+}
```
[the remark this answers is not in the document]
trietng: - 2353 lines of code might look like a lot but it is actually maybe like half of that. The code is just incredibly repeatitive and I don't think there is a way around it without breaking it into 10-20 more methods. 
- 4D noise is nice to have, although its usage might be niche though. Modifying `FastNoiseLite` to support 4D though is an entirely different story. I'm sure they (the original authors) didn't include it for a reason that's beyond me.
- I modified the code to be more in line with your suggestion. Now the `FastNoiseLite` class inherits from a `Noise` interface, which is intended for any user that wants to roll their own noise generator.

By the way another sampling technique that I've found to be very handy to have in any game development framework is Poisson disk sampling. I did play around with procedural generation and one of the steps, after using Simplex noise that is, does require the usage of this particular sampling algorithm. The step in question, of course, is spawn foilage on the generated terrain.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	public enum NoiseType {
+		OpenSimplex2, OpenSimplex2S, Cellular, Perlin, ValueCubic, Value
+	}
+
+	public enum RotationType3D {
+		None, ImproveXYPlanes, ImproveXZPlanes
+	}
+
+	public enum FractalType {
+		None, FBm, Ridged, PingPong, DomainWarpProgressive, DomainWarpIndependent
```
tommyettinger, line 56: Might as well do a little reviewing... The capitalization here seems weird, `FBm` isn't correct either for an acronym or a word. It's short for Fractal Brownian Motion, and other acronyms here use all capital letters for the starts of words, like `ImproveXYPlanes`. I'd suggest using either `FBM` or `Fbm` and sticking to it; I have no preference toward either but I don't think `FBm` is right at all.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+
+	private enum TransformType3D {
+		None, ImproveXYPlanes, ImproveXZPlanes, DefaultOpenSimplex2
+	}
+
+	/** Declaration of common constants */
+	private static final float SQRT3 = 1.7320508075688772935274463415059f;
+	private static final float R3 = 2.0f / 3.0f;
+
+	private int mSeed = 1337;
```
tommyettinger, line 79: As usual for libGDX, I think we should default to `protected` instead of `private` or especially package-private, and should be documenting fields so we understand them later. This is *especially true here*, since noise code is almost never easy to understand.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	}
+
+	/** Sets octave gain for all fractal noise types
+	 * @param gain octave gain (default is {@code 0.5}) */
+	public void setFractalGain (final float gain) {
+		mGain = gain;
+		calculateFractalBounding();
+	}
+
+	/** Sets octave weighting for all none DomainWarp fractal types. Keep between 0...1 to maintain -1...1 output bounding.
```
tommyettinger, line 166: Other than the typo here, where "none DomainWarp" should probably be "non-DomainWarp", I don't have the faintest idea what octave weighting is in this context. It isn't, as far as I know, a borderline-standard term like lacunarity or gain. The range restriction is definitely helpful to see documented, though.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	@Override
+	public float getNoise3D (float x, float y, float z) {
+		x *= mFrequency;
+		y *= mFrequency;
+		z *= mFrequency;
+
+		switch (mTransformType3D) {
+		case ImproveXYPlanes: {
+			float xy = x + y;
+			float s2 = xy * -(float)0.211324865405187;
```
tommyettinger, line 267: We should always be using a float literal instead of creating a double and casting it to float.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	private static float fastMin (final float a, final float b) {
+		return Math.min(a, b);
+	}
+
+	private static float fastMax (final float a, float b) {
+		return Math.max(a, b);
+	}
+
+	private static float fastAbs (final float f) {
+		return f < 0 ? -f : f;
```
tommyettinger, line 589: This is likely slower than Math.abs(), which is already defined for floats.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	}
+
+	private static float fastSqrt (final float f) {
+		return (float)Math.sqrt(f);
+	}
+
+	private static int fastFloor (final float f) {
```
tommyettinger, line 596: We already have fast floor, ceil, and round functions for floats defined in MathUtils, and they don't need branching.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+	private static float lerp (final float a, final float b, final float t) {
+		return a + t * (b - a);
+	}
+
+	private static float interpHermite (final float t) {
+		return t * t * (3 - 2 * t);
+	}
+
+	private static float interpQuintic (final float t) {
+		return t * t * t * (t * (t * 6 - 15) + 10);
```
tommyettinger, line 613: This interpolation can exceed the 0 to 1 range when passed certain less-than-1 floats. The lowest input that goes out-of-range is `interpQuintic(0.99535805f) == 1.0000001f`, which is surprisingly far from 1.0f. `interpQuintic(243f/244f) == 1.0000002f` is also true. I use `return t * t * t * (t * (t * 6f - 15f) + 9.999998f);` , myself, which won't exceed the boundary.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+
+	/** Simplex/OpenSimplex2 noise */
+	private float singleSimplex (final int seed, final float x, final float y) {
+		// 2D OpenSimplex2 case uses the same algorithm as ordinary Simplex.
+		final float G2 = (3 - SQRT3) / 6;
+
+		/*
+		 * --- Skew moved to switch statements before fractal evaluation --- final FNLfloat F2 = 0.5f * (SQRT3 - 1); FNLfloat s = (x
+		 * + y) * F2; x += s; y += s;
+		 */
```
tommyettinger, line 902: We definitely don't need this comment in... some language other than Java.

gdx/src/com/badlogic/gdx/math/noise/FastNoiseLite.java
```diff
@@ -0,0 +1,2375 @@
...
+			break;
+		case DefaultOpenSimplex2: {
+			float r = (xs + ys + zs) * R3; // Rotation, not skew
+			xs = r - xs;
+			ys = r - ys;
+			zs = r - zs;
+		}
+			break;
+		default:
+			break;
```
tommyettinger, line 1989: This default break isn't needed; both lines can be removed.

gdx/src/com/badlogic/gdx/math/noise/Noise.java
```diff
@@ -0,0 +1,16 @@
+
+package com.badlogic.gdx.math.noise;
+
+import com.badlogic.gdx.math.Vector3;
+
+public interface Noise {
```
tommyettinger, line 6: The noise interface would be nice but should probably have at least bare-minimum docs for the interface, if not the 5 methods.

tommyettinger (commented):
I'm not entirely sure if I want to approve a canonical noise for libGDX when none of the 3rd-party solutions seem to agree on... anything... but all do quite a lot more than this does. This can't do any 4D noise, which is useful both for 3D shapes that move over time, and 2D tiling textures using the `noise4d(sin(x), cos(x), sin(y), cos(y))` trick. FastNoiseLite is basically standard now for various game engines/frameworks, but I think it supports 4D noise also... There's a bunch of oddities in the port here, mostly minor. It gets hard to maintain another 2K lines of source code if they weren't written in Java in the first place, and no one exactly understands how they work.

OpenSimplex is an interesting algorithm, but its patent-free status hasn't mattered for several years now that Simplex noise has had its patent expire without renewal, and it's *considerably* slower and more complex in 3D than Simplex. Even the name is a misnomer in 3D; I don't think it involves a tetrahedron (the 3D simplex shape) at all...

I think maybe a dependency on Joise would make more sense for user projects than incorporating parts of OpenSimplexNoise. I have my own libraries that provide noise (cringe and SquidSquad), though they also do other things. There's also my older and worse make-some-noise file, which no one should be using now probably... It's a nightmare to navigate. I think there are other projects with other goals for noise, too. Cringe and Joise both allow serialization at arbitrary points, and cringe uses libGDX Json for that as an option. Cringe and SquidSquad support OpenSimplexNoise 2 (F and S variants), as well as standard Simplex noise. It seems like there's no common interface between all these libraries, though; they all provide different features and have different APIs.

PR #7585, opened 2025-02-03
Use `Locale.ROOT` instead of `ROOT_LOCALE`

Frosty-J:
Some small changes to the `I18NBundle` class and a trimming down of GWT's `Locale` documentation, because they are old.

Of course, there's more places than this where `String#isEmpty()` could be used. Search for `.length() == 0`, `.length() > 0`, `.equals("")`, and `"".equals` to find other instances. But this isn't very important to change, so I won't do so for unrelated classes in this pull request.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/I18NBundle.java
+++ b/gdx/src/com/badlogic/gdx/utils/I18NBundle.java
@@ -181,7 +181,7 @@ public class I18NBundle {
 					// Found the bundle for the only candidate locale
 					break;
 				}
-				if (isBaseBundle && baseBundle == null) {
+				if (baseBundle == null) {
 					// Store the base bundle and keep on processing the remaining fallback locales
 					baseBundle = bundle;
 				}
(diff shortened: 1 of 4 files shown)


gdx/src/com/badlogic/gdx/utils/I18NBundle.java
```diff
@@ -259,7 +259,7 @@ private static List<Locale> getCandidateLocales (Locale locale) {
 		String country = locale.getCountry();
 		String variant = locale.getVariant();
 
-		List<Locale> locales = new ArrayList<Locale>(4);
+		List<Locale> locales = new ArrayList<>(4);
```
tommyettinger, line 262: 🍿

tommyettinger (approved):
I like this. I haven't used I18N features much, but updating docs and using classes that are now guaranteed to be supported is helpful. My one nitpick is that the checks for emptyLanguage, emptyCountry, and emptyVariant made less checks before, since they stored the result of what now would be isEmpty(). It isn't a big deal, though. Approved!

Frosty-J:
Two extra length checks aren't going to increase your load times. But if you were to think it's more readable with those variables, I can bring them back.

tommyettinger:
It doesn't matter, leave them in so I don't need to review again!

PR #7586, opened 2025-02-03
Remove `touchSleepTime`

Frosty-J:
Given how long ago libGDX dropped Android Donut support, I think this can be removed from the app config without so much as a deprecation period. Nobody should still be using this, especially not setting it to 16ms on 90Hz+ devices. It's a hack for a problem long gone.

Once upon a time, I looked into this. It seemed Android 1.x would absolutely batter your app with touch events while the touchscreen was pressed, even if the user's finger wasn't moving, which was bad news for your single-core device without JIT. But I don't have a device old enough to test that on, and it's difficult to buy one for testing since many were updated to Android 2 and sellers don't always have pictures that confirm whether they've been updated. Not that there'd be any point in testing this now - I just find the idea of supporting beyond-ancient Android versions in some of my personal non-libGDX projects interesting.


obigu (approved):
Looks good to me

Issue #7582, opened 2025-01-31
Rare NPE in GlyphLayout


upost:
I have an Android app in the Play Store.
Rarely I see exceptions like this in the Google Play Console:

```
Exception java.lang.NullPointerException:
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:176)
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:95)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.computePrefSize (Label.java:162)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.scaleAndComputePrefSize (Label.java:147)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.getPrefWidth (Label.java:244)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.<init> (Label.java:75)
```
or:

```
(RuntimeException root cause)
Caused by java.lang.NullPointerException:
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.calculateWidths (GlyphLayout.java:280)
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:266)
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:95)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.computePrefSize (Label.java:162)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.scaleAndComputePrefSize (Label.java:147)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.getPrefWidth (Label.java:244)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.<init> (Label.java:75)
```
or, maybe related:

```
Exception java.lang.IndexOutOfBoundsException:
  at com.badlogic.gdx.utils.Array.get (Array.java:155)
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:215)
  at com.badlogic.gdx.graphics.g2d.GlyphLayout.setText (GlyphLayout.java:101)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.computePrefSize (Label.java:160)
  at com.badlogic.gdx.scenes.scene2d.ui.Label.scaleAndComputePrefSize (Label.java:147)
[...]

PR #7588, opened 2025-02-07
Make GlyphLayout's GlyphRun acquisition overrideable

LobbyDivinus:
This PR basically serves as a solution to https://github.com/libgdx/libgdx/issues/7582 by allowing the user to implement custom pooling mechanic that could for example allow for thread safe use

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
@@ -172,7 +172,7 @@ public class GlyphLayout implements Poolable {
 			runEnded:
 			{
 				// Store the run that has ended.
-				GlyphRun run = glyphRunPool.obtain();
+				GlyphRun run = obtainRun();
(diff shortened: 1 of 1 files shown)


Berstanio:
> This PR basically serves as a solution to https://github.com/libgdx/libgdx/issues/7582 by allowing the user to implement custom pooling mechanic that could for example allow for thread safe use

Just as a note, you can achieve similar already by calling `Pools#set` on startup.

LobbyDivinus:
True, that can be used to implement a thread safe pool for glyph runs. I don't think it can fix the access to static colorStack for fonts that use formatting though.

Berstanio:
> I don't think it can fix the access to static colorStack for fonts that use formatting though.

Thats true. From what I can see, it should be okay to just move the colorStack creation into the method or pool it. But others might know more about it.

Regarding the thread safe pool, I'm honestly in favor of making the standard libGDX pool thread safe. Having static methods not be thread-safe, especially such universal ones, seems like a big footgun to me.

PR #7590, opened 2025-02-15
Two GWT emu fixes re: long in DataInputStream.

tommyettinger:
The first change is just using the added-a-few-years-ago Numbers.longBitsToDouble(), which is faster, instead of GWT's slow Double.longBitsToDouble() . There's an extended discussion (with pictures!) of the second change here, https://tommyettinger.github.io/libGDX-GWT-checks/ . The reason the second change looks the way it does now is optimistically hoping that JS will be able to pipeline two non-dependent bit operations before it ORs them. This mostly affects UBJson, but it may affect other places that used DataInputSteam.readLong() on GWT. Long story short, an `L` was missing from a mask literal, and so it wasn't masking an int at all, letting it implicitly promote to long afterwards and sign-extending then if the int was negative.


tommyettinger:
This is such a minor thing to review that I'm adding it to milestone 1.13.2 . There's no reason UBJson needs to be broken on GWT in the next release.

obigu (approved):

PR #7592, opened 2025-02-20
Replace PixmapPackerRectangle Rectangle inheritance by Bounds

obigu:
`PixmapPackerRectangle` inherits from `Rectangle` which is a `Shape2D` from `math` package that represents a geometric rectangle and is not appropriate to use it to simply store x,y,width and height integer values of a box/rectangle bounds, and in this case even less to do it via inheritance instead of composition.

I'm not sure if creating a separate `Box` or `Bounds` class instead of the inner class would make sense as it's a pretty standard use case probably used in many games and some other libGDX classes such as `IOSScreenBounds` could benefit from it.


NathanSweet:
The static Bounds class seems fine to me. These bounds are int, other bounds are likely float, IntBounds/etc doesn't seem great.

PR #7597, opened 2025-03-10
Use official ANGLE for MetalANGLE

Berstanio:
The MetalANGLE repo, which we use, is currently unmaintained: https://github.com/kakashidinho/metalangle

I made a port of the MetalKit replacement and combined it with the official ANGLE sources: https://github.com/Berstanio/MetalANGLEKit

It is compatibility wise identical to MetalANGLE, the only difference is, that the OpenGLES compatibility backend is removed. This means, that macOS simulator/hackintosh wont be able to run the app in the iPhone simulator anymore.
However, for those the normal OpenGLES libGDX backend is still present.

The new MetalANGLEKit has been tested in two production apps so far.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,7 @@
 - API Deprecation: Deprecated ReflectionPool and the related Pools#get/Pools#obtain method. Please use the new DefaultPool with the new Pools#get/Pools#obtain methods. They offer more safety and can be used with the java 8 method reference syntax (e.g. Pools.get(MyClass::new))
 - API Deprecation: Deprecated constructors taking a "Class" for Queue/ArrayMap/Array. Please use the constructors utilising "ArraySupplier". They offer more safety and can be used with the java 8 method reference syntax (e.g. MyClass[]::new)
 - iOS: Update to MobiVM 2.3.23
+- iOS: The MetalANGLE backend has been replaced by MetalANGLEKit which uses a newer ANGLE revision
 
 [1.13.1]
 - [BREAKING CHANGE] Android: Since 1.13.0 libGDX requires setting `android.useAndroidX=true` in your gradle.properties file. In 1.13.1 it is NO longer needed to define the `androidx.core:core` dependency in your Android module.
--- a/backends/gdx-backend-robovm-metalangle/build.gradle
+++ b/backends/gdx-backend-robovm-metalangle/build.gradle
@@ -59,7 +59,7 @@ tasks.register('fetchMetalANGLEKit', Download) {
 	doFirst {
 		file("build/tmp").mkdirs();
 	}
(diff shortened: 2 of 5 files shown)


obigu:
This is a great contribution that not only improves mantainability and makes it more future proof but also:
- Unlocks fps to more than 60 fps which, for some reason, is still limited both on OpenGL and current MetalAngle
- Better performance. This would require more testing to 100% confirm but some tests run by @fgnm (check Discord) found there was a non negligible performance improvement vs current MetalAngle.
- Smaller binary sizes

Version 1.2 is currently in production in 3 of my apps and works perfectly.

As a side note to avoid possible misunderstandings, running the apps on a Mac on the iOS Simulator will work as usual. What will stop working, if I uderstand it correctly, is trying to run the iPhone Simulator on a Hackintosh which is an extremely niche use case.

obigu (dismissed):

fgnm:
As for my experience this is one of the best improvements in iOS. I usually blame myself if something doesn't run smooth as I want, but switching to the official ANGLE I've noticed a huge boost in general but mostly in shaders performances. 

My game was running between 30/40fps on an iPhone 7 with old MetalANGLE (with blur and some other post processing effects), with official ANGLE went to fixed 60fps. Even on high end iPhone, such as 15 Pro Max, game performances were not so good with MetalANGLE, and I'm not talking about the 120fps, sometimes even on that hardware I wasn't able to get 60fps! (now it works well at 120fps without major issues)

From my point of view there are only benefit with this PR and I can't see any reason why angle backend should support OpenGL ES as fallback anyway... There will be always the old gles backend for anyone want to use the simulator on a Hackintosh, and it's pretty easy to swap them.

The only thing I'd like to say is that we probably need a way to schedule ANGLE updates. Now MetalANGLEKit is based on a ANGLE version of 7 months ago (I can see more work was done on metal backend since then). Things get worse if you think that desktop version of ANGLE is stuck at 2023 :D bad thing is that ANGLE doesn't have a release channel.. So we should decide when update the git subproject.. Two times per year would be doable?

obigu (approved):
Should be good to merge now that the repo's been moved!

tommyettinger (approved):
Looks great, I love it when we can switch from unmaintained code to official code! This being used in production is great news. Is there anything I should change in the generated projects for Liftoff, like depend on the different backend that the tests will depend on when this is merged?

obigu:
> Looks great, I love it when we can switch from unmaintained code to official code! This being used in production is great news. Is there anything I should change in the generated projects for Liftoff, like depend on the different backend that the tests will depend on when this is merged?

yeah, once released the template will need to change the same way tests do in the PR. That is depending on `gdx-backend-robovm-metalangle` instead of `gdx-backend-robovm` and removing `<framework>OpenGLES</framework>` dependency. I can work on the PR when needed.

PR #7598, opened 2025-03-10
Hopefully fix labeler labeling everything as android

Berstanio:
>  If a base option is provided without a top-level key, then it will default to any.
https://github.com/actions/labeler

So, seems like it's just an "OR" currently and not an exclusion filter.


Tom-Ski (approved):

SimonIT:
Thanks!

PR #7600, opened 2025-03-10
Replace Android Vibrator deprecated methods

obigu:
Tested with `VibratorTest` on an Android 15 Pixel 6 device.


Frosty-J (approved):
On an Android 14 device with a basic vibration motor, this pull request makes no difference to the behaviour of VibrationTest. Feels good. Also tested on Lollipop emulator for the hell of it to unsurprising results.

Tom-Ski (approved):

PR #7601, opened 2025-03-10
Minor code improvements to DefaultAndroidInput

obigu:
This is a small experiment to test the viability of improving some of the Android backend codebase.

Android backend code is filled with warnings due to deprecations as well as code style/good practices warnings. Changing everything in one go is both risky and very hard to review (and thus unmergeable). The warnings affect both IDEs as well as build (Gradle) console outputs.

Let's try one class and one type of change at a time.

I've started with `DefaultAndroidInput`. Most changes are trivial and I expect to be viewed as positive by the great majority of contributors including adding the diamond operator (except from @NathanSweet) which is already present in this class (and most of libGDX code base except from Scene2d where there's a de facto veto).

If this "easy to review" PRs approach works I'll go through the main Android backend classes.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -17,6 +17,7 @@
 package com.badlogic.gdx.backends.android;
 
 import android.animation.Animator;
+import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.app.AlertDialog;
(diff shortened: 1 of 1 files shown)


Frosty-J (approved):
No issues with this. Was tempted to make similar changes myself when adding `PredictiveBackHandler`.

NathanSweet:
Here's my input:

* Diamond operator adds characters but otherwise makes no difference. It's fine, though I don't see the point.
* `final` on fields where possible is OK.
* Fixing comments, removing commented and old code, and `Arrays.fill` are great.
* For `@SuppressLint` and similar, if we truly never care about a warning we should turn that warning off rather than litter the code, otherwise it's fine.
* StringBuilder `append` change makes sense, though if the goal is efficiency we can pass a char instead a string containing one char.

Do with this feedback as you like -- I haven't been very involved in the Android backend and don't actively maintain it, so I don't mind if my usual obsessiveness isn't applied.

Berstanio (approved):

PR #7602, opened 2025-03-10
Bump project target to java 8

Berstanio:
Fixes https://github.com/libgdx/libgdx/issues/5487

All the discussion happened in the issue.
The main summary is, that new java 8 API's cannot be used. But it is still worth to bump the target, to be able to use java 8 syntax features and allow building with the newest JDK again.
An immidiate improvement to both performance and safety would be being able to merge and apply to the core repo: https://github.com/libgdx/libgdx/pull/7474#issuecomment-2710053369


PokeMMO (approved):

Tom-Ski (approved):

tommyettinger (approved):
Hallelujah!

PR #7605, opened 2025-03-13
Migrate away from ArrayReflection array instantiation

Berstanio:
Since we now target java 8, we can do further improvements by avoiding reflection.

This PR migrates array creation with `ArrayReflection(MyClass.class, 10)` to a functional interface like `MyClass[]::new`.
Avoiding reflection increases type-safety and makes other backends more reliable, as they often have limited reflection capabilities.

This PR migrates Queue/ArrayMap/Array for now, while deprecating the old methods. Some minor other reflection uses have been migrated too.
I also did some minor cleanups on things that I directly encountered, like duplicate generic specification and diamond operators.

### DIFF
--- a/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
+++ b/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
@@ -2,7 +2,6 @@
 package com.badlogic.gdx.math;
 
 import com.badlogic.gdx.utils.Array;
-import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
 /** This class represents a cumulative distribution. It can be used in scenarios where there are values with different
  * probabilities and it's required to pick one of those respecting the probability. For example one could represent the frequency
@@ -13,7 +12,7 @@ import com.badlogic.gdx.utils.reflect.ArrayReflection;
  * <a href="http://en.wikipedia.org/wiki/Cumulative_distribution_function">Wikipedia</a> for a detailed explanation.
  * @author Inferno */
 public class CumulativeDistribution<T> {
-	public class CumulativeValue {
+	public static class CumulativeValue<T> {
 		public T value;
 		public float frequency;
 		public float interval;
@@ -25,23 +24,20 @@ public class CumulativeDistribution<T> {
 		}
 	}
 
-	private Array<CumulativeValue> values;
+	private final Array<CumulativeValue<T>> values;
 
 	public CumulativeDistribution () {
-		@SuppressWarnings("unchecked")
-		// It is impossible to create a generic array without reflection (I think)
-		CumulativeValue[] tmp = (CumulativeValue[])ArrayReflection.newInstance(CumulativeValue.class, 10);
-		values = new Array<>(tmp);
+		values = new Array<CumulativeValue<T>>(false, 10, CumulativeValue[]::new);
 	}
 
 	/** Adds a value with a given interval size to the distribution */
 	public void add (T value, float intervalSize) {
-		values.add(new CumulativeValue(value, 0, intervalSize));
+		values.add(new CumulativeValue<>(value, 0, intervalSize));
 	}
 
 	/** Adds a value with interval size equal to zero to the distribution */
 	public void add (T value) {
-		values.add(new CumulativeValue(value, 0, 0));
+		values.add(new CumulativeValue<>(value, 0, 0));
 	}
 
 	/** Generate the cumulative distribution */
@@ -80,7 +76,7 @@ public class CumulativeDistribution<T> {
 	 * @param probability
 	 * @return the value whose interval contains the probability */
 	public T value (float probability) {
-		CumulativeValue value = null;
+		CumulativeValue<T> value = null;
 		int imax = values.size - 1, imin = 0, imid;
 		while (imin <= imax) {
 			imid = imin + ((imax - imin) / 2);
@@ -118,7 +114,7 @@ public class CumulativeDistribution<T> {
 
 	/** Set the interval size on the passed in object. The object must be present in the distribution. */
 	public void setInterval (T obj, float intervalSize) {
-		for (CumulativeValue value : values)
+		for (CumulativeValue<T> value : values)
 			if (value.value == obj) {
 				value.interval = intervalSize;
 				return;
--- a/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
+++ b/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
@@ -12,7 +12,7 @@ import com.badlogic.gdx.utils.Array;
  * <a href="http://en.wikipedia.org/wiki/Cumulative_distribution_function">Wikipedia</a> for a detailed explanation.
  * @author Inferno */
 public class CumulativeDistribution<T> {
-	public static class CumulativeValue<T> {
+	private static class CumulativeValue<T> {
 		public T value;
 		public float frequency;
 		public float interval;
--- a/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
+++ b/gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
@@ -2,6 +2,7 @@
 package com.badlogic.gdx.math;
 
 import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.reflect.ArrayReflection;
 
 /** This class represents a cumulative distribution. It can be used in scenarios where there are values with different
  * probabilities and it's required to pick one of those respecting the probability. For example one could represent the frequency
@@ -27,7 +28,10 @@ public class CumulativeDistribution<T> {
 	private Array<CumulativeValue> values;
 
 	public CumulativeDistribution () {
-		values = new Array<CumulativeValue>(false, 10, CumulativeValue.class);
+		@SuppressWarnings("unchecked")
+		// It is impossible to create a generic array without reflection (I think)
+		CumulativeValue[] tmp = (CumulativeValue[])ArrayReflection.newInstance(CumulativeValue.class, 10);
+		values = new Array<>(tmp);
 	}
 
 	/** Adds a value with a given interval size to the distribution */
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -52,8 +52,7 @@ public class Array<T> implements Iterable<T> {
 	 *           memory copy.
 	 * @param capacity Any elements added beyond this will cause the backing array to be grown. */
 	public Array (boolean ordered, int capacity) {
-		// noinspection unchecked
-		this(ordered, capacity, (ArraySupplier<T[]>)ArraySupplier.OBJECT_ARRAY_SUPPLIER);
+		this(ordered, capacity, ArraySupplier.object());
 	}
 
 	/** Creates a new array with {@link #items} with the specified supplier.
@@ -78,7 +77,6 @@ public class Array<T> implements Iterable<T> {
 	 * @deprecated Use {@link Array#Array(boolean, int, ArraySupplier)} instead */
 	@Deprecated
 	public Array (boolean ordered, int capacity, Class arrayType) {
-		// noinspection unchecked
 		this(ordered, capacity, size -> (T[])ArrayReflection.newInstance(arrayType, size));
 	}
 
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -52,20 +52,42 @@ public class Array<T> implements Iterable<T> {
 	 *           memory copy.
 	 * @param capacity Any elements added beyond this will cause the backing array to be grown. */
 	public Array (boolean ordered, int capacity) {
+        //noinspection unchecked
+        this(ordered, capacity, (ArraySupplier<T[]>)ArraySupplier.OBJECT_ARRAY_SUPPLIER);
+	}
+
+	/** Creates a new array with {@link #items} with the specified supplier.
+	 * @param ordered If false, methods that remove elements may change the order of other elements in the array, which avoids a
+	 *           memory copy.
+	 * @param capacity Any elements added beyond this will cause the backing array to be grown. */
+	public Array (boolean ordered, int capacity, ArraySupplier<T[]> arraySupplier) {
 		this.ordered = ordered;
-		items = (T[])new Object[capacity];
+		items = arraySupplier.get(capacity);
+	}
+
+	/** Creates an ordered array with {@link #items} with the specified supplier and a capacity of 16. */
+	public Array (ArraySupplier<T[]> arraySupplier) {
+		this(true, 16, arraySupplier);
 	}
 
 	/** Creates a new array with {@link #items} of the specified type.
 	 * @param ordered If false, methods that remove elements may change the order of other elements in the array, which avoids a
 	 *           memory copy.
-	 * @param capacity Any elements added beyond this will cause the backing array to be grown. */
+	 * @param capacity Any elements added beyond this will cause the backing array to be grown.
+	 *
+	 * @deprecated Use {@link Array#Array(boolean, int, ArraySupplier)} instead
+	 * */
+	@Deprecated
 	public Array (boolean ordered, int capacity, Class arrayType) {
-		this.ordered = ordered;
-		items = (T[])ArrayReflection.newInstance(arrayType, capacity);
+		// noinspection unchecked
+		this(ordered, capacity, size -> (T[])ArrayReflection.newInstance(arrayType, size));
 	}
 
-	/** Creates an ordered array with {@link #items} of the specified type and a capacity of 16. */
+	/** Creates an ordered array with {@link #items} of the specified type and a capacity of 16.
+	 *
+	 * @deprecated Use {@link Array#Array(ArraySupplier)} instead
+	 * */
+	@Deprecated
 	public Array (Class arrayType) {
 		this(true, 16, arrayType);
 	}
@@ -73,10 +95,10 @@ public class Array<T> implements Iterable<T> {
 	/** Creates a new array containing the elements in the specified array. The new array will have the same type of backing array
 	 * and will be ordered if the specified array is ordered. The capacity is set to the number of elements, so any subsequent
 	 * elements added will cause the backing array to be grown. */
-	public Array (Array<? extends T> array) {
-		this(array.ordered, array.size, array.items.getClass().getComponentType());
+	public Array (Array<T> array) {
+		items = Arrays.copyOf(array.items, array.size);
+		ordered = array.ordered;
 		size = array.size;
-		System.arraycopy(array.items, 0, items, 0, size);
 	}
 
 	/** Creates a new ordered array containing the elements in the specified array. The new array will have the same type of
@@ -91,9 +113,9 @@ public class Array<T> implements Iterable<T> {
 	 * @param ordered If false, methods that remove elements may change the order of other elements in the array, which avoids a
 	 *           memory copy. */
 	public Array (boolean ordered, T[] array, int start, int count) {
-		this(ordered, count, array.getClass().getComponentType());
+		items = Arrays.copyOfRange(array, start, start + count);
+		this.ordered = ordered;
 		size = count;
-		System.arraycopy(array, start, items, 0, size);
 	}
 
 	public void add (T value) {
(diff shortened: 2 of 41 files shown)


PokeMMO (approved):

PokeMMO:
Less reflection, more good.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
```diff
@@ -27,7 +28,10 @@ public CumulativeValue (T value, float frequency, float interval) {
 	private Array<CumulativeValue> values;
 
 	public CumulativeDistribution () {
-		values = new Array<CumulativeValue>(false, 10, CumulativeValue.class);
+		@SuppressWarnings("unchecked")
+		// It is impossible to create a generic array without reflection (I think)
+		CumulativeValue[] tmp = (CumulativeValue[])ArrayReflection.newInstance(CumulativeValue.class, 10);
+		values = new Array<>(tmp);
```
NathanSweet, line 34: Generics only matter on the LHS. I prefer to omit them from the RHS when possible, except when using `var`. I think it can be:
```
values = new Array(false, 10, CumulativeValue[]::new);
```

gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -52,31 +52,51 @@ public Array (int capacity) {
 	 *           memory copy.
 	 * @param capacity Any elements added beyond this will cause the backing array to be grown. */
 	public Array (boolean ordered, int capacity) {
+		// noinspection unchecked
```
NathanSweet, line 55: Does this comment add anything useful? It appear many times. I didn't flag them all but I think all can be removed.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
```diff
@@ -0,0 +1,12 @@
+
+package com.badlogic.gdx.utils;
+
+/** An interface that is used to create arrays. Even tho not annotated with "FunctionalInterface", it can act as one. <br>
+ * You can use a constructor reference or lambda as a ArraySupplier, such as with {@code MyClass[]::new} or
+ * {@code (size) -> new MyClass[size]}. */
+public interface ArraySupplier<T> {
+	/** A default array supplier that creates an Object[]. */
+	ArraySupplier<?> OBJECT_ARRAY_SUPPLIER = Object[]::new;
```
NathanSweet, line 9: I think I'd rather see `Object[]::new` inline everywhere than `ArraySupplier.OBJECT_ARRAY_SUPPLIER`.

Berstanio: The main issue was, that when doing that, you need to do:
```
(ArraySupplier<T[]>)(ArraySupplier<?>)Object[]::new
``` 
, which was in my opinion rather noisy.
But I don't have a strong opinion on that, I can move it to inline.

gdx/src/com/badlogic/gdx/utils/Array.java
Berstanio: It supresses warnings on a statement, but you are right, it is mainly an IntelliJ feature I think.
I can replace it with an Annotation of the constructor `@SuppressWarnings("unchecked")`

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
Berstanio: `values = new Array(false, 10, CumulativeValue[]::new);`
This doesn't work, because `CumulativeValue` has a generic field based on the outer class. And it is impossible (?) to create a generic element array.
Regarding the RHS generic, the java linter issues a warning when omiting it. So to keep the linter output cleaner/seeing actual warnings, I would think keeping the RHS generic is preferable.

gdx/src/com/badlogic/gdx/utils/Array.java
tommyettinger: It does add something useful. It has semantic meaning to modern IDEs, and prevents them from issuing a warning fpr one statement when unchecked code is encountered in that statement (including any of your idiosyncratic `new Array()` statements that omit the generics on the RHS). It avoids an annotation on the method or class that suppresses that warning. I know you've mentioned that you've configured your IDE to not warn on any libGDX code, but you haven't made your warning suppressions part of the repo, so they are only useful to you. This is useful to anyone who just checks out the repo.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
tommyettinger: Agreed, it's a lot shorter. As long as `Object[]::new` doesn't capture surrounding state, we should be fine just using the constructor reference.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
tommyettinger: Nope.
Doesn't work at all.

tommyettinger: There's also a "rawtypes" warning issued whenever you omit generics on the RHS. This is textbook Java and has been for almost as long as I have programming it, since I started with Java 6 (before the diamond operator).

gdx/src/com/badlogic/gdx/utils/Array.java
NathanSweet: Ah, I had had not seen this way of suppressing warnings. I am not a fan of jumping through hoops and littering code with warning suppression (whether it's this comment or `@SuppressWarnings`) when we don't actually care about the warning. Instead, we should turn off the warning for the entire codebase.

If you satisfy all `unchecked` warnings, you... can still get ClassCastException. It is just not a big problem in practice and the warning is not worth the trouble and clutter.

> I know you've mentioned that you've configured your IDE to not warn on any libGDX code

I don't know where you got that, but I use libgdx's standard Eclipse warnings, where only unimportant warnings are disabled. There are also warnings enabled that are not enabled by default.

> but you haven't made your warning suppressions part of the repo

Eh? https://github.com/libgdx/libgdx/blob/master/.settings/org.eclipse.jdt.libgdx.prefs

Maybe the problem is that you don't use Eclipse, or that IDEA doesn't have equivalent warnings configured, which IMO it should. The codebase has no warnings so if you see a warning in IDEA, that's one that should be disabled.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
tommyettinger: The extra noise would be an issue to the shortness, hm. Honestly, my only opposition to any chosen syntax is based on how much space any of these takes up, since we may see them a lot. The type inference with method and constructor references is something I don't have as much experience with, and it seems rather different from the language-level-7 techniques we were restricted to before.

How about `ArraySupplier.ANY` or `ArraySupplier.OBJECT`? They both say what it can be used for and are significantly less wordy.

gdx/src/com/badlogic/gdx/utils/Array.java
tommyettinger: I'll see if I can import the .settings preferences into IDEA... Most of these seem standardized enough, they usually interoperate. Of contributors from the last year until today, you're the only contributor who I know uses Eclipse (see https://github.com/libgdx/libgdx/graphs/contributors?from=3%2F9%2F2024 ) and if no one else has imported those preferences into their (JetBrains) IDE, no one else has seen a codebase with no warnings.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
NathanSweet: Try:
```
values = new Array(false, 10, CumulativeDistribution.CumulativeValue[]::new);
```
`rawtypes` warnings are uninteresting.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
NathanSweet: It could be `(ArraySupplier)Object[]::new`. Raw type usage discards generics safety checks. If we can agree on disabling the right warnings (libgdx has been doing for ~a decade) then we can use this straightforward approach when necessary without warnings, `@SuppressWarnings`, arcane casts, extra fields, etc.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
tommyettinger: The error can be fixed by making CumulativeValue a static class, with a type parameter for its T value that should just be the same as the CumulativeDistribution's T. I have the changes and can commit them if you want.

gdx/src/com/badlogic/gdx/utils/Array.java
tommyettinger: Haven't yet found a way to import the .settings preferences, but I have found a few non-existing annotations mentioned in those preferences. I'd say I personally disagree with about a quarter to a half of the settings it's been told to ignore. In any case, this preferences file is not easy to get into IDEA, and I can see why.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
Berstanio: > The error can be fixed by making CumulativeValue a static class, with a type parameter for its T value that should just be the same as the CumulativeDistribution's T. I have the changes and can commit them if you want.

I implemented that. Since `CumulativeValue` was never exposed outside `CumulativeDistribution`, the change should be non-breaking

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
Berstanio: How about this?
```java
public interface ArraySupplier<T> {
	/** A default array supplier that creates an Object[]. */
	ArraySupplier<?> ANY = Object[]::new;

	/** Returns a default array supplier that creates an Object[]. */
	@SuppressWarnings("unchecked")
	static <T> ArraySupplier<T[]> object () {
		return (ArraySupplier<T[]>)ANY;
	}

	T get (int size);
}
```

Then we can just do `this(initialSize, ArraySupplier.object());`. No warnings that need surpression and looks not bad imo.

gdx/src/com/badlogic/gdx/utils/Array.java
NathanSweet: I doubt you can even tell which settings are non-default. There are probably ~3 warnings to ignore.

Some annotations shouldn't be in the Eclipse settings, but also don't matter.

I don't use IDEA, so I won't configure its warnings. You can either continue to have warnings in IDEA or ignore the unimportant warnings (and optionally enable a few warnings disabled by default) and make your IDEA settings available to others, in whatever way makes sense. What I don't want is tons of comments and annotations for useless warnings.

gdx/src/com/badlogic/gdx/math/CumulativeDistribution.java
NathanSweet: `CumulativeDistribution.CumulativeValue[]::new` compiled with no other changes (before changing to a static class). Static is better though, as it doesn't make any use of the enclosing instance. CumulativeValue should probably be private.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
NathanSweet: `(ArraySupplier<T[]>)(ArraySupplier<?>)` via a field and method reads OK as `ArraySupplier.object()`, but I contend that `(ArraySupplier)Object[]::new` is simplest and the code is clearer. No warnings with the settings required for libgdx to have no warnings. No suppression is needed. Raw types are a tool for exactly this situation, where generics get in the way.

Berstanio: > No warnings with the settings required for libgdx to have no warnings. No suppression is needed. Raw types are a tool for exactly this situation, where generics get in the way.

The issue is, that IntelliJ by default issues the warnings. And also javac issues linter warnings for them.
We can look into ways, to disable these warnings for IntelliJ, eclipse and javac, but I think this is out of scope for this PR.
Regardless, I personally like `ArraySupplier.object()` to, because it imo falls in line with other java 8 added API style (e.g. `Function.identity()` and similar).
So, if you are fine with `ArraySupplier.object()`, I would personally prefer that.

NathanSweet: IMHO when defaults aren't great, we should configure our tools appropriately.

I'm OK with `ArraySupplier.object()` if you feel strongly about it. The call looks nice, though slightly less clear what it does and doesn't handle eg `(ArraySupplier)Object[][]::new`.

The implementation could still use raw types instead of `(ArraySupplier<T[]>)(ArraySupplier<?>)` shenanigans. We already use raw types, there is no point in make the code less clear to avoid them.
```
static <T> ArraySupplier<T[]> object () {
	return (ArraySupplier)Object[]::new;
}
```

PokeMMO: Is this code even used enough to justify the concerns? Nobody cares about a slightly more verbose call that isn't often used.

People do care about having to configure their ide/javac to ignore warnings though.

PokeMMO: >Raw types are a tool for exactly this situation, where generics get in the way.

No, they exist for [legacy code](https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.8).

>Raw types can be thought of as wildcards whose type rules are deliberately unsound, to accommodate interaction with legacy code.

NathanSweet: > No, they exist for [legacy code](https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.8).

Why they exist doesn't make them any less of a tool for exactly this situation.

@Berstanio I'm not sure what situations those are, but possibly a raw type cast would be applicable there.

gdx/src/com/badlogic/gdx/utils/Array.java
tommyettinger: Was that a challenge I hear? I'm um, *rather certain* this one's non-default:

> `org.eclipse.jdt.core.compiler.annotation.nullable.secondary=com.esotericsoftware.spine.launcher.utils.Null,com.esotericsoftware.spine.server.utils.Null,com.esotericsoftware.spine.editor.utils.Null`
Really great practice there, polluting the settings with classes that libGDX can't legally depend on! I don't have a license to Spine, nor do I really want one!

> `org.eclipse.jdt.core.compiler.annotation.nonnull=com.badlogic.gdx.utils.Null.NonNull`
> `org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=com.badlogic.gdx.utils.Null.NonNullByDefault`

These don't even exist, so... Pretty sure they're both non-default and non-valid.

> `org.eclipse.jdt.core.compiler.problem.autoboxing=ignore`
This one seems useful to not ignore, since libGDX goes to pretty extensive lengths to avoid boxing.

> `org.eclipse.jdt.core.compiler.problem.deprecation=ignore`
I'd be surprised if Eclipse was not actively discouraging deprecated code, but then again, it's Eclipse, so who knows.

> `org.eclipse.jdt.core.compiler.problem.emptyStatement=ignore`
I don't know why this would be set to ignore; empty statements can absolutely be a source of bugs.

```java
if(x < 0);
    throw new GdxRuntimeException("No negative numbers please!");
```

> `org.eclipse.jdt.core.compiler.problem.missingHashCodeMethod=ignore`

This should probably not be ignored because it can lead to subtle bugs, and a warning would at least point that out.

> `org.eclipse.jdt.core.compiler.problem.unusedImport=ignore`
> `org.eclipse.jdt.core.compiler.problem.unusedLocal=ignore`
> `org.eclipse.jdt.core.compiler.problem.unusedObjectAllocation=ignore`

All of these seem perfectly reasonable to see warnings for, it's probably helpful to see what code doesn't need to be present!

Pretty sure that's more than 3 non-default settings!

As for your last comment, [this annotation isn't IDE-specific, it's been a part of Java since 1.5](https://docs.oracle.com/javase/8/docs/api/java/lang/SuppressWarnings.html).

Frosty-J: @tommyettinger The Spine thing has already been removed in bab07573cd63c3a426d0079e86345cd2eec2818d. I didn't check the rest.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
tommyettinger: > The use of raw types is allowed only as a concession to compatibility of legacy code. The use of raw types in code written after the introduction of generics into the Java programming language is strongly discouraged. It is possible that future versions of the Java programming language will disallow the use of raw types.

Not a tool, a concession. It literally says "strongly discouraged" and also that future versions can potentially disallow the use of raw types, which doesn't sound like a very good tool.

gdx/src/com/badlogic/gdx/utils/Array.java
NathanSweet: libgdx's warning settings have long been established. They are chosen to have zero warnings for the way libgdx code is written. If any warnings appear, they should be fixed. This is more useful than having a long list of warnings where the majority are unimportant. It's not about which warnings "seem helpful". All warnings are helpful in some scenarios, but enabling them all is not.

If you turn on a warning and it occurs hundreds times, that is not a useful warning to have enabled. For example, enabling `autoboxing` gives 393 warnings for libgdx source. An argument could be made to fix all occurrences so a warning can be enabled, but that's probably not the right thing to do when there are many. Warnings should help find problems, not cause busywork.

A middle ground is to set less important warnings to `info` instead of `ignored`, then exclude `info` from the `Problems` pane. Code underlines and line number annotations [still appear](http://n4te.com/x/1564-bmhF.png). In some cases `info` is unnecessary, eg autoboxing is already [highlighted red](http://n4te.com/x/1563-R90J.png). Even `info` should not be overused as too many underlines makes them less useful. `info` is good for null analysis. That can be very useful for a lot of code, though it's sometimes wrong.

> As for your last comment, [this annotation isn't IDE-specific, it's been a part of Java since 1.5](https://docs.oracle.com/javase/8/docs/api/java/lang/SuppressWarnings.html).

You've misinterpreted what I wrote. `@SuppressWarnings` should be avoided, but can be acceptable if the warning is enabled in libgdx's settings. IDE-specific suppression comments are not acceptable.

> direct commit

Doesn't matter much since I heard I'm the only one using Eclipse.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
NathanSweet: Why they exist doesn't make them any less of a tool for exactly this situation. You guys not read good? 😛

Despite the JLS's portents, it is unlikely to be removed. In the meantime, it will be enjoyed. I refuse to live in fear!

Berstanio: > @Berstanio I'm not sure what situations those are, but possibly a raw type cast would be applicable there.

Yes, raw-types would work, but they issue a warning (especially in my own code base) and `ArraySupplier.object()` and `ArraySupplier.ANY` is just nicer to work with then casts, especially in my own code bases.
So, as they have use for internal libGDX and also potentionally user applications, I think it's a convinient addition.

gdx/src/com/badlogic/gdx/utils/Array.java
Frosty-J: Yeah I doubt you're the only one who uses Eclipse. mgsx does or did and there's likely to be others I don't know about. But I think it's important to have an agreed-upon way of doing these things and that any warning customisation needs to be reflected upon the majority of users (i.e. in IntelliJ/AS and CI/CD) instead of the minority.

For example, the linter that ensures each pull request has the code formatted just the way it likes it... as much as I often disagree with its decisions, I can't deny it is useful at fostering a relatively consistent codebase. It is the authority, albeit one I can override.

NathanSweet: @Frosty-J Agreed, that would be great. Contributors seeing zero warnings before their edits and CI failure for any enabled warnings would help PRs. Unnecessary `@SuppressWarnings` could be removed. Null analysis at `info` level with our `@Null` is minor but nice to have ([described here](https://github.com/libgdx/libgdx/commit/bab07573cd63c3a426d0079e86345cd2eec2818d#r153780535)).

tommyettinger: I seem to recall mgsx switching from Eclipse to IDEA or AS a little while ago, but I'd have to ask to be sure. There aren't that many contributors here... https://github.com/libgdx/libgdx/graphs/contributors?from=3%2F9%2F2024

There's also as far as I can tell zero mention of that .settings file in any documentation for libGDX (wiki, [the instructions for building from source](https://libgdx.com/dev/from-source/), docs for the repo itself), so it will only affect people who either have it automatically enabled (I don't know if Eclipse does this) or have somehow gone looking for it. I see it's mentioned [in build.gradle here](https://github.com/libgdx/libgdx/blob/bab07573cd63c3a426d0079e86345cd2eec2818d/build.gradle#L155-L163), but I don't know if that applies when loading libGDX as a Gradle project or if it only works when Eclipse dotfiles are created by the `eclipse` task.

gdx/src/com/badlogic/gdx/utils/ArraySupplier.java
Berstanio: So, what is the state of this? Is this mergable as-is? @NathanSweet

NathanSweet: I prefer the cast, but I don't mind the static `object()`. Merged now!

> but they issue a warning

They do not if you setup your warnings as intended for the libgdx codebase. In the future I would like to remove unnecessary SuppressWarnings.

PR #7607, opened 2025-03-13
Migrate set/getStyle calls in `Skin#setEnabled` to an interface

Berstanio:
This PR replaces the resolution of the `getStyle` and `setStyle` methods in `Skin#setEnabled` with a `Styleable` interface.
The reflection based approach does not work well with shrinker/dead code elimination and is more error prone when it comes to type safety. It also runs risks of duplicate `get/setStyle` methods, especially with subclassing.

The interface added fixes these issue, while retaining full backwards-compatibility. Therefore, this change is non-breaking.
A deprecation warning is issued to users, if the interface is not present and `Skin#setEnabled` is called.

@NathanSweet pinging you as the original author

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
@@ -415,23 +415,28 @@ public class Skin implements Disposable {
 	}
 
 	/** Sets the style on the actor to disabled or enabled. This is done by appending "-disabled" to the style name when enabled is
+	 * false, and removing "-disabled" from the style name when enabled is true. If the style was not found in the skin, an
+	 * exception is thrown. */
+	public <V> void setEnabled (Styleable<V> styleable, boolean enabled) {
+		V style = styleable.getStyle();
+
+		// noinspection unchecked
+		Class<V> styleClass = (Class<V>)style.getClass();
+
+		String name = find(style);
+		if (name == null) return;
+		name = name.replace("-disabled", "") + (enabled ? "" : "-disabled");
+		style = get(name, styleClass);
+
+		styleable.setStyle(style);
+	}
+
+	/** Sets the style on the actor to disabled or enabled. This is done by appending "-disabled" to the style name when enabled is
 	 * false, and removing "-disabled" from the style name when enabled is true. A method named "getStyle" is called the actor via
-	 * reflection and the name of that style is found in the skin. If the actor doesn't have a "getStyle" method or the style was
-	 * not found in the skin, no exception is thrown and the actor is left unchanged. */
-	public void setEnabled (Actor actor, boolean enabled) {
-		if (actor instanceof Styleable) {
-			// noinspection unchecked
-			Styleable<Object> styleable = (Styleable<Object>)actor;
-			Object style = styleable.getStyle();
-
-			String name = find(style);
-			if (name == null) return;
-			name = name.replace("-disabled", "") + (enabled ? "" : "-disabled");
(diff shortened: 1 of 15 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
```diff
@@ -440,6 +453,9 @@ public void setEnabled (Actor actor, boolean enabled) {
 			method.invoke(actor, style);
 		} catch (Exception ignored) {
 		}
+
+		Gdx.app.error("DEPRECATED",
+			"Setting style on Actor " + actor.getClass().getName() + ", which does not implement 'Styleable' interface");
```
NathanSweet, line 458: It's more of a warning than an error. I don't believe we do this kind of deprecated logging anywhere else.

We could deprecate `setEnabled(Actor, boolean)` and add `setEnabled(Styleable, boolean)`, which would have to cast to Actor. Callers would have to cast to remove ambiguity, eg `setEnabled((Styleable)moo, false)` which is a bummer. To improve that we could rename `setEnabled(Actor, boolean)` to `setEnabledReflection(Actor, boolean)` and deprecate it. Then we could use the fancy `<T extends Actor & Styleable> void setEnabled (T, boolean)`.

None of the options are amazing, but I think I like the logging the least.

NathanSweet (commented):
This adds a bit of noise solely for Skin `setEnabled`, but avoiding reflection is good. Maybe it is useful for other things that need to juggle styles. If the deprecation can be improved I'm OK with merging.

Alternatively we could remove (deprecate first) Skin `setEnabled`. I don't use it in any of my projects and haven't for many years.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Skin.java
Berstanio: Thank you for the review!
I'm very indifferent about what deprecation method to use. I choose the logging, cause it avoided the ambigious method call problem, but I'm fine with the method overload too!

> To improve that we could rename setEnabled(Actor, boolean) to setEnabledReflection(Actor, boolean) and deprecate it.

Regarding this tho, I think this will break third-party libs that are compiled against an older libGDX version? However, if this method is rarely used anyway, it might be fine.

mgsx-dev: Considering that migrating user code is simply to add `implements Styleable<MyWidget.Style>` to their custom widgets, for me it's a minor breaking change. So we could simply ditch the old code, no?

tommyettinger: I'm fully in favor of adding an `implements` clause to all my widgets in TextraTypist; it shouldn't be hard to do. It will mean that it will be briefly broken with nightly builds of libGDX and the next stable release until I can release TextraTypist with an update, so a stop-gap in the middle might help.

NathanSweet: Given there's not a super clean way to deprecate this one, I'm OK with a little breakage.

Tom-Ski: Fine with breakage. Also fine with having deprecation migration hints as a fallback for a single release version, to be removed in next release.

Berstanio: I implemented the breaking change with `setEnabledReflection`

PR #7608, opened 2025-03-15
Remove reflection use for lwjgl3 sound/audio creation

Berstanio:
This is a simple change that avoids unnecessary reflection for creating the lwjgl3 sound classes.
Since lwjgl3 requires java 8+, we can safely use the BiFunction here.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - [BREAKING CHANGE] Android: Proguard config line `boolean getUseDefaultContactFilter();` needs to be added. See https://github.com/libgdx/libgdx/pull/7578.
 - [BREAKING CHANGE] Android: Removed `AndroidApplicationConfiguration#touchSleepTime`.
 - [BREAKING CHANGE] Skin `setEnabled` now only works with actors that implement `Styleable`. Use `setEnabledReflection` for old behavior or add `Styleable` to your actors.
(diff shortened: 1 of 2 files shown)


Frosty-J (approved):
Go for it, Boss. Backend-specific changes are as safe as it gets.

PokeMMO:
This is a public function, so likely needs a CHANGES entry. Other than that, should be easy to merge.

Berstanio:
> This is a public function, so likely needs a CHANGES entry. Other than that, should be easy to merge.

Ahh, didn't realised. If it's public, rather deprecating the old one first? Not sure, whether any library actually uses this function.

SomeTroglodyte:
So how would "To migrate just replace `registerMusic("myMusic", MyMusic.class);` with `registerMusic("myMusic", MyMusic::new);`" translate into kotlin?

```kotlin
    class AacMusic(audio: OpenALLwjgl3Audio, file: FileHandle) : OpenALMusic(audio, file) {
        override fun read(buffer: ByteArray?): Int {
            //...
            return 0
        }
        override fun reset() {
            //...
        }
    }
    fun registerCodecs(audio: OpenALLwjgl3Audio) {
        audio.registerMusic("m4a", AacMusic::new)  // unresolvable reference
    }
```

Can you confim that in that construed example
```
        audio.registerMusic("m4a", { a, f -> AacMusic(a, f) })
```
would be the correct replacement? Just curious, we didn't _actually_ implement that, this PR is just invalidating a comment.

PokeMMO:
Kotlin does method references differently, I believe you can simply do 

```
audio.registerMusic("m4a", ::AacMusic)
```

SomeTroglodyte:
> I believe you can simply do

Ooooh, learned something, thanks! Mouse-over confirms:
image
Readability is another thing - matter of taste.

PR #7610, opened 2025-03-19
scene2d: Solve NPEs when clicking with both mouse buttons

yairm210:
As reported [here](https://github.com/yairm210/Unciv/issues/12922), if users click with left AND right mouse buttons, and then release, there will be **two** touchUp events that will be sent. 
The first one will set actor to null; The second will cause a null pointer exception trying to run actor.stageToLocalCoordinates
This resolves the issue


obigu:
I've tried reproducing with Scene2dTest where the "Fancy Background" button has a, `ActorGestureListener` but haven't been able to make it crash. Can you please provide an SSCCE (https://libgdx.com/wiki/articles/getting-help#executable-example-code) to be able to reproduce?

yairm210:
I find I am unable to reproduce this simply 🤔 
I find myself confused, the code really shouldn't allow this sort of crash
Since it seems this isn't a general problem and is specific to my setup somehow, I'll handle this in my codebase

PR #7612, opened 2025-03-23
Adding initial character to iOS invisible text delegate.

auraxangelic:
When setText is called in TextField, the iOS delegate is not notified that there are characters in the text field. Adding an initial character to invisible delegate to accept backspaces.

To reproduce issue:
1. Create new project and create a text field and add to stage.
2. Call setText("asdf") on the text field.
3. Create IPA and launch on iOS.
4. Give focus to text field and move cursor to the end of the text. Try to backspace. No characters are removed.


MrStahlfelge:
Valid use case, but does it work when backspace is hit twice?

auraxangelic:
> Valid use case, but does it work when backspace is hit twice?

From my testing, it does work. The setText("") function doesn't update the invisible text delegate, but the result of the backspace does and updates the invisible text delegate correctly.

MrStahlfelge (dismissed):

auraxangelic:
@Berstanio @obigu Can we merge this for now with the fix? Anyone can always go back and find another solution later.

obigu:
> @Berstanio @obigu Can we merge this for now with the fix? We can always go back and find another solution later.

Sure, can you please merge master to make sure it builds and add a line to CHANGES before?

PR #7614, opened 2025-03-26
Versions Maintenance

obigu:
~~Three~~ Two small maintenance changes.

1. Fixed wrong min Android SDK protection.
2. AGP upgrade to `8.7.3` which is already well tested and supported by IDEs
~~3. Updated to Gradle `8.12.1` as per Gradle docs recommendations (I'm aware 8.13.0 exists but that that would require testing libGDX tests suites on all platforms run correctly which can be done in a separate PR).~~


tommyettinger:
8.12.1 has known bugs with GWT. I can find sources if you want, but we encountered them in Liftoff before the most recent version, and they required some incredibly nasty workarounds involving System.exit() to actually terminate Gradle.

tommyettinger:
https://github.com/gradle/gradle/issues/32213 Big No to 8.12.1, I am not willing to deal with that bug again. Either downgrade to 8.12 if you're that opposed to 8.13, or use the current version.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.12-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.12.1-bin.zip
```
tommyettinger, line 3: Please show me where the current Gradle docs recommend using 8.12.1 ? I don't believe this myself.

tommyettinger (changes requested):
Press X to doubt.

gradle/wrapper/gradle-wrapper.properties
Frosty-J: They write this for just about every patch:

> We recommend users upgrade to 8.12.1 instead of 8.12.

https://docs.gradle.org/8.12.1/release-notes.html

obigu: It's great that Gradle releases a patch version supposedly to fix stuff, breaks things, doesn't release a new patch 8.12.2 to fix it and recommends upgrading to it.

https://docs.gradle.org/8.12.1/release-notes.html
**We recommend users upgrade to 8.12.1 instead of 8.12.**

It's ok, I'll rollback this change, I didn't expect things to break on a minor update released a couple of months ago.  Nothing agains 8.13.0 in particular, as I've said, it should be properly tested (and now I see even patch versions should be tested).

tommyettinger: Yeah, the 8.12.1 break really surprised me too! AFAICT it mostly affects Gretty, but I haven't found an alternative for running GWT superDev... Gretty is used a lot by GWT-related code it seems.

tommyettinger (approved):
All good now!

PR #7616, opened 2025-03-29
Remove reflection use in ParallelArray

Berstanio:
This PR finishes of, what was started in https://github.com/libgdx/libgdx/pull/7605
The only downside I can see is, that an additional object allocation happens, when a `ChannelDescriptor` is allocated. However, this should not be a real issue I think.


tommyettinger (approved):
Looks clean, simple, straightforward, and similar to the other ArraySupplier changes. I'm a little surprised to see ParallelArray in the 3D particles package, and I didn't even know it existed before. I'm not sure if anyone actually uses 3D particles, especially since the Flame editor doesn't seem to be working much of the time... (Even more so on Macs...)

Tom-Ski (approved):

PR #7619, opened 2025-03-30
Add Drawable default interface methods

NathanSweet:
These methods are just for convenience. There's a very small risk that someone's Drawable already has methods with these names (they'd have to not extend BaseDrawable).

Do all our targets handle default interface methods correctly? I would assume so, but these are the first in libgdx.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Drawable.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Drawable.java
@@ -42,6 +42,21 @@ public interface Drawable {
 
 	public void setBottomHeight (float bottomHeight);
 
+	default public void setPadding (float topHeight, float leftWidth, float bottomHeight, float rightWidth) {
+		setTopHeight(topHeight);
+		setLeftWidth(leftWidth);
+		setBottomHeight(bottomHeight);
+		setRightWidth(rightWidth);
+	}
+
+	default public void setPadding (float padding) {
+		setPadding(padding, padding, padding, padding);
+	}
+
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Drawable.java
tommyettinger, line 1: I like these being here for sure! They are simple and straightforward, they follow directly from how padding is already implemented, and users who already override them will I believe have their class implementations take precedence over interface ones anyway. I'm tentatively approving but I'll test on the platforms I can soon! Someone else will need to test RoboVM and hopefully MOE, but I'm happy to check GWT and TeaVM in detail.

tommyettinger (approved):
Tentatively approving, but I can only test some platforms, and I will do so when time permits. The code looks like a nice solid usage of default methods.

NathanSweet:
We could merge and if some platforms explode, we'll find out. Maybe give it another week or so.

PokeMMO (approved):

tommyettinger:
Guess we're on Team Potential Explosion now... This should make it easier to test, I guess?

PokeMMO:
Yolo, it should either explode immediately (gwt or ios) or users can fix their projects fairly easily.

tommyettinger:
So I think the plan is to encourage at least someone who has a multi-platform app to try to update to the -SNAPSHOT dep and test it on as many platforms as they can easily...

obigu:
The bomb has been defused. Quick test on RoboVM project using latest SNAPSHOT seems to be working.

Issue #6638, opened 2021-08-27
`setBackgroundFPS` has no LWJGL3 equivalent, despite `setIdleFPS()` sounding like one


Frosty-J:
Note that this issue has been completely rewritten to save face. GitHub keeps an edit history anyway.

#### Issue details

LWJGL3's `setIdleFPS()` is not equivalent to LWJGL2's `backgroundFPS`. Likewise, LWJGL2 has no equivalent for `setIdleFPS()`. Not only is this confusing - particularly when getting an existing project running on a different backend - it's also inconvenient if someone wants their LWJGL3 game to run at a lower frame rate when unfocused, such as to save resources.

https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglApplicationConfiguration.java#L89-L91

https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java#L172-L175

#### Reproduction steps/code

At first glance you might think these two snippets do the same thing, but they don't!

```java
config.foregroundFPS = 30;
config.backgroundFPS = 15;
```

```java
config.setForegroundFPS(30);
config.setIdleFPS(15);
```

#### Version of libGDX and/or relevant dependencies
1.10.1-SNAPSHOT

#### Stacktrace
N/A

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows
- [X] Linux
- [X] MacOS

calvertdw:
I always thought I would eventually have to implement a "pause" or some custom logic to get this to work, but being built in would be nice.

Betalord:
I use a temporary solution like this:
```java
config.setWindowListener(new Lwjgl3WindowAdapter() {
            @Override
            public void focusLost() {
                Gdx.graphics.setForegroundFPS(60);
            }
            
            @Override
            public void focusGained() {
                Gdx.graphics.setForegroundFPS(0);
            }
});
```
Which limits FPS in background to 60 and does not limit it in foreground.

shabanshala:
Hello. I am working with my group to get this issue resolved. Would I be able to get this issue assigned to me?

-Cheers

[...]

PR #7621, opened 2025-04-01
Added BackgroundFPS equivalent functionality to LWJGL3

bermuta19:
https://github.com/libgdx/libgdx/issues/6638
LWJGL3 has no equivalent function to LWJGL2's BackgroundFPS, despite setIdleFPS() sounding like one. Added functionality to implement equivalent system to BackgroundFPS.

### DIFF
--- a/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessApplication.java
+++ b/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessApplication.java
@@ -70,6 +70,7 @@ public class HeadlessApplication implements Application {
 		// they are only implemented as mock objects
 		this.graphics = new MockGraphics();
 		this.graphics.setForegroundFPS(config.updatesPerSecond);
+		this.graphics.setBackgroundFPS(config.updatesPerSecond);
 		this.audio = new MockAudio();
 		this.input = new MockInput();
 
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -176,7 +176,11 @@ public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 					window.makeCurrent();
 					currentWindow = window;
 				}
-				if (targetFramerate == -2) targetFramerate = window.getConfig().foregroundFPS;
+				if (targetFramerate == -2 && window.isFocused()) targetFramerate = window.getConfig().foregroundFPS;
+				synchronized (lifecycleListeners) {
+					haveWindowsRendered |= window.update();
+				}
+				if (targetFramerate == -2 && !window.isFocused()) targetFramerate = window.getConfig().backgroundFPS;
 				synchronized (lifecycleListeners) {
 					haveWindowsRendered |= window.update();
 				}
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -67,7 +67,9 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 
 	int idleFPS = 60;
 	int foregroundFPS = 0;
+	int backgroundFPS = 0;
 
+	boolean activateBackgroundFPS = false;
 	boolean pauseWhenMinimized = true;
 	boolean pauseWhenLostFocus = false;
 
@@ -104,6 +106,7 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		transparentFramebuffer = config.transparentFramebuffer;
 		idleFPS = config.idleFPS;
 		foregroundFPS = config.foregroundFPS;
+		backgroundFPS = config.backgroundFPS;
 		pauseWhenMinimized = config.pauseWhenMinimized;
 		pauseWhenLostFocus = config.pauseWhenLostFocus;
 		preferencesDirectory = config.preferencesDirectory;
@@ -190,6 +193,12 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 	 * 0. */
 	public void setForegroundFPS (int fps) {
 		this.foregroundFPS = fps;
+	}
+
+	/** Sets the target framerate for the application. The CPU sleeps as needed. Must be positive. Use 0 to never sleep. Default is
+	 * 0. */
+	public void setBackgroundFPS (int fps) {
+		this.backgroundFPS = fps;
 	}
 
 	/** Sets whether to pause the application {@link ApplicationListener#pause()} and fire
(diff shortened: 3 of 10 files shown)


backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/HeadlessApplication.java
```diff
@@ -70,6 +70,7 @@ public HeadlessApplication (ApplicationListener listener, HeadlessApplicationCon
 		// they are only implemented as mock objects
 		this.graphics = new MockGraphics();
 		this.graphics.setForegroundFPS(config.updatesPerSecond);
+		this.graphics.setBackgroundFPS(config.updatesPerSecond);
```
tommyettinger, line 73: Headless apps don't have a concept of foreground or background, so I don't see any point in adding setBackgroundFPS() to anything headless...

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
```diff
@@ -176,7 +176,11 @@ protected void loop () {
 					window.makeCurrent();
 					currentWindow = window;
 				}
-				if (targetFramerate == -2) targetFramerate = window.getConfig().foregroundFPS;
+				if (targetFramerate == -2 && window.isFocused()) targetFramerate = window.getConfig().foregroundFPS;
+				synchronized (lifecycleListeners) {
+					haveWindowsRendered |= window.update();
+				}
+				if (targetFramerate == -2 && !window.isFocused()) targetFramerate = window.getConfig().backgroundFPS;
```
tommyettinger, line 183: This looks odd to me; why is this a separate if statement, where it is after the window updates, instead of an else block that runs when the window isn't focused?

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -67,7 +67,9 @@ public enum GLEmulation {
 
 	int idleFPS = 60;
 	int foregroundFPS = 0;
+	int backgroundFPS = 0;
 
+	boolean activateBackgroundFPS = false;
```
tommyettinger, line 72: This also looks odd and different from how the rest of the Lwjgl3 FPS config works. Normally backgroundFPS would be initialized to some meaningful but invalid "sentinel" number like -1 or -2, and if it still has that value later, then that means it was never set. If it was set, such as to 0 to have unlimited background FPS, 60 to have reasonably high background FPS, or 5 to have very low-power-usage background FPS, then you know it was intentional because it isn't that negative sentinel value.

tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/Lwjgl3TestStarter.java
```diff
@@ -78,6 +78,7 @@ public static void main (String[] argv) {
 			if (SharedLibraryLoader.os == Os.MacOsX) {
```
tommyettinger, line 78: Only testing this on MacOS? Could other OSes test this, too?

tommyettinger (changes requested):
I have some minor, mostly style suggestions here, but I do think the last change listed should expand to be used on more OSes (even if they don't need the foreground FPS set, setting the background FPS should at least get tested). The `activateBackgroundFPS` change isn't... necessary, it just helps the class be more consistent. The actual meat of the change, in Lwjgl3Application, looks quite strange to me, and I'm not sure if it's correct or why it is what it is. If there's an explanation and it tests correctly, I'm fine with it staying the same, though, I just need to know it works. Headless apps don't need to set background FPS because they don't create a window.

tests/gdx-tests-lwjgl3/src/com/badlogic/gdx/tests/lwjgl3/Lwjgl3TestStarter.java
Frosty-J: It's an odd place to put this, in the test starter window for ANGLE macOS. Non-continuous rendering is probably the better way to reduce its background usage, but it's not something I'm terribly concerned about given its low impact and the short amount of time it is typically run for.

isaak1294: Hey, thanks for the feedback this was the first issue I worked on. I will can implement the other comments, but I wasn't sure what to do for tests. I just added to the only test that I could find that referenced FPS. Is there documentation on testing standards? Should I add additional tests for this issue?

Frosty-J: If you want to create a test for it, you can make a class in https://github.com/libgdx/libgdx/tree/master/tests/gdx-tests/src/com/badlogic/gdx/tests that extends `GdxTest` and add it to the list at https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/utils/GdxTests.java#L110.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtGraphics.java
```diff
@@ -576,6 +576,10 @@ public void setVSync (boolean vsync) {
 	public void setForegroundFPS (int fps) {
 	}
 
+	@Override
+	public void setBackgroundFPS (int fps) {
+	}
```
Frosty-J, line 581: Not sure why this was added here but not in `AndroidGraphics` and `IOSGraphics` (which also have empty overrides). I don't think it's important but just don't see reason for GWT to get special treatment.

Frosty-J (commented):
I haven't tested it but it looks better!

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
```diff
@@ -176,7 +176,11 @@ protected void loop () {
 					window.makeCurrent();
 					currentWindow = window;
 				}
-				if (targetFramerate == -2) targetFramerate = window.getConfig().foregroundFPS;
+				if (targetFramerate == -2 && window.isFocused())
+					targetFramerate = window.getConfig().foregroundFPS;
+				else {
+					if (targetFramerate == -2 && !window.isFocused()) targetFramerate = window.getConfig().backgroundFPS;
```
crykn, line 182: I don't think that works quite as intended in multi-window environments. If the first windows isn't focused that causes all windows to be rendered with the background fps rate.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Graphics.java
```diff
@@ -513,6 +513,14 @@ public void setVSync (boolean vsync) {
 	@Override
 	public void setForegroundFPS (int fps) {
 		getWindow().getConfig().foregroundFPS = fps;
+		if (getWindow().getConfig().backgroundFPS < 0) {
+			getWindow().getConfig().backgroundFPS = fps;
+		}
```
crykn, line 518: Is there a reason for this? The old LWJGL 2 backend didn't have this, see [LwjglGraphics.java#L681-L683](https://github.com/libgdx/libgdx/blob/e063c6dc2cdc078de30c97f7d7aedd574b5efc35/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java#L681-L683).

Issue #7577, opened 2025-01-25
Bug: when an actor is removed, toActor is wrong in exit() callback


sulai:
The two calls to `fireExit()` in `Stage.actorRemoved()` should be replaced by `fireEnterAndExit()`.

- **Obervation**: As it is now, InputListener.exit() has `toActor` set to the removed actor.
- **Expected**: toActor should be set to the actor behind the removed actor.

NathanSweet:
The expected behavior sounds right, particularly given the `InputListener#exit` parameter name `toActor`. I don't have time at the moment to write a test to see the wrong behavior, change it, and verify the change. A test and/or PR is welcome!

PR #7622, opened 2025-04-01
Replaced fireExit() with fireEnterAndExit() inside actorRemoved() inLWJGL3 

shabanshala:
Issue #7577 : Bug : when an actor is removed, toActor is wrong in exit() callback

Replaced fireExit() with fireEnterAndExit() in Stage.actorRemoved() to ensure the correct toActor is set in the exit() event when an actor is removed from the stage.

NOTE: When I saved the new code, I believe all the formatting changed, hence why there were so many affected files and additions/deletions.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -386,7 +386,7 @@ public class Array<T> implements Iterable<T> {
 		size = n - count;
 	}
 
-	/** Removes from this array all of elements contained in the specified array.
+	/** Removes from this array the first instance of each element contained in the specified array.
 	 * @param identity True to use ==, false to use .equals().
 	 * @return true if this array was modified. */
 	public boolean removeAll (Array<? extends T> array, boolean identity) {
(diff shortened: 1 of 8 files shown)


tommyettinger (changes requested):
There is no way in Heaven, Earth or Hell this will get merged as-is. Look at the files you didn't just change, but deleted! The whole freetype submodule is gone, unrelated Bullet JNI files are missing, all Eclipse settings are deleted... You should have only changed two lines, one method each, in Stage.java , and nothing else should be affected at all. Especially nothing should be deleted! You should probably make sure you're fully up-to-date and able to run tests *before* you make the two changes for this PR, commit them, and make the PR.

gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
Frosty-J, line 1: Is this the changes? You're probably best off reverting that commit and trying again. If for whatever reason you can't avoid changing the line endings or whatever's going on here, only `git add` the files you've made changes to.

NathanSweet:
Please note that we can't merge even a correct PR without an [SSCCE](https://sscce.org/) showing the wrong behavior before the fix.

PR #7623, opened 2025-04-01
Replaced fireExit() with fireEnterAndExit() inside actorRemoved() inLWJGL3

shabanshala:
Issue #7577 

Replaced fireExit() with fireEnterAndExit() in Stage.actorRemoved() to ensure the correct toActor is set in the exit() event when an actor is removed from the stage.


NathanSweet:
https://github.com/libgdx/libgdx/pull/7622#issuecomment-2770514561

PR #7625, opened 2025-04-09
Add JsonValue copy constructor and setChild methods

NathanSweet:
I found these necessary when converting JSON to different JSON.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/JsonValue.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonValue.java
@@ -83,6 +83,28 @@ public class JsonValue implements Iterable<JsonValue> {
 		set(value);
 	}
 
+	/** Creates a deep copy of the specific value, except {@link #parent()}, {@link #next()}, and {@link #prev()} are not null. */
+	public JsonValue (JsonValue value) {
+		this(value, false);
+	}
+
+	private JsonValue (JsonValue value, boolean siblings) {
+		type = value.type;
+		stringValue = value.stringValue;
+		doubleValue = value.doubleValue;
+		longValue = value.longValue;
+		name = value.name;
+		if (value.child != null) {
+			child = new JsonValue(value.child, true);
+			child.parent = this;
+		}
+		if (siblings && value.next != null) {
+			next = new JsonValue(value.next, true);
+			next.prev = this;
+		}
+		size = value.size;
+	}
+
 	/** Returns the child at the specified index. This requires walking the linked list to the specified entry, see
 	 * {@link JsonValue} for how to iterate efficiently.
 	 * @return May be null. */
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/utils/JsonValue.java
```diff
@@ -83,6 +83,26 @@ public JsonValue (boolean value) {
 		set(value);
 	}
 
+	/** Creates a deep copy of the specific value, except {@link #parent()}, {@link #next()}, and {@link #prev()} are not null. */
+	public JsonValue (JsonValue value) {
+		this(value, null);
+	}
+
+	private JsonValue (JsonValue value, @Null JsonValue parent) {
```
tommyettinger, line 91: Why is the copy constructor with a parent private? Is this just a really narrow use case? Also, if value.child == value, this will infinitely recurse until it StackOverflowErrors. I believe this can also happen for the single-argument copy constructor.

tommyettinger (changes requested):
I think this is mostly correct, there's just a corner case that worries me a bit that could maybe occur with programmatically-created JSON. If a value has its child assigned to itself (same reference), then copying the child will recurse until there's a StackOverflowError, if I read this correctly. Obviously coders who want the app to run correctly will try to avoid this, but I worry about malicious, incompetent, or "vibe" coders somehow crafting a worst-case JsonValue programmatically. Checking for self-identical children is something I've become unfortunately aware that I have to do in some of my code, largely thanks to the ancient Apache Harmony test suite for JDK data structures.

gdx/src/com/badlogic/gdx/utils/JsonValue.java
NathanSweet: It's likely only needed for the copy constructor implementation. We want to copy a node and all its children. For the first node we don't want to set or copy parent, next, or prev but for everything below it we do. I can't imagine needing that elsewhere.

NathanSweet:
> If a value has its child assigned to itself (same reference), then copying the child will recurse until there's a StackOverflowError

That's true. It won't happen from parsing JSON but could happen by writing bad code. JsonValue doesn't try to guard against this, it'll just happily loop forever via iterator or `for` loop. There are many references in the doubly linked lists for hierarchy and siblings, exhaustive checking would be a pain. We could do a basic check, like throw if `value == this` in `setChild`, but that only catches the simplest of issues. In practice I haven't found it hard to avoid, so I feel like it's OK having it as a "don't do that".

gdx/src/com/badlogic/gdx/utils/JsonValue.java
tommyettinger: Cool, that seems reasonable then.

tommyettinger (approved):
Fair enough, footguns will exist regardless I suppose!

Issue #7227, opened 2023-09-19
Input codepoints bigger than 2 bytes (e.g. emojis) get casted into single 2 byte char


LobbyDivinus:
#### Issue details
I noticed in my game that certain characters (i.e. emojis like 🏗️) get not reported correctly to the application via InputProcessor.keyTyped.

What happens in detail:
1. Let's type the emoji 🏗️ which consists of the codepoint 0x1F3D7 (that is, more than two bytes so it will have to be represented using multiple chars)
2. The GLFWCharCallback in https://github.com/libgdx/libgdx/blob/bb799a9cd0a92cadb0425acd6064654cd7a4d6c8/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/DefaultLwjgl3Input.java#L51C14-L59C4 receives the correct (4 byte) integer codepoint value.
3. However, the callback casts the codepoint into a (2 byte) and only puts that into the eventQueue for the application to consume. In my example that is 0xF3D7 which is not a valid character.

A solution may be to put the upper 2 bytes of the integer codepoint into the eventQueue, too, if they are non zero.

#### Reproduction steps/code
See description above. All you need is to do is to check the values the keyTyped(char character) method in a set input processor gets called with when entering 🏗️ or similar codepoints.

As this issue relies on actual key input I don't how tests can be written for it.

#### Version of libGDX and/or relevant dependencies
ligGDX 1.12.0 with lwjgl3 backend.

#### Please select the affected platforms
- [X] Windows
Probably all platforms but I tested on Windows only.

LobbyDivinus:
Apparently [Character.highSurrogate](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#highSurrogate-int-) and [Character.lowSurrogate](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#lowSurrogate-int-) can be used to convert an integer codepoint to a char surrogate pair that would be compatible with String codepoint methods.

tommyettinger:
[...]

PR #7628, opened 2025-04-19
Make BitmapFont understand high+low surrogate pairs (#7227)

JamesXNelson:
I didn't add extra tests for BitmapFontWriter, but it looked like a pretty straight-forward refactor to me.

Let me know if you want to merge this, and I'll sign the CLA / etc etc.


gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -817,6 +863,26 @@ public boolean hasGlyph (char ch) {
...
+				return null;
+			} else if (Character.isLowSurrogate(ch)) {
+				if (highSurrogate == null) {
+					assert false : "No surrogate found for 0x" + Integer.toHexString(ch);
+					return null;
+				}
+				final Glyph[] page = surrogates == null ? null : surrogates[highSurrogate - HIGH_SURROGATE_MIN];
+				highSurrogate = null;
+				if (page == null) {
+					// allow people who choose to run with asserts enabled get notified when characters don't work
```
NathanSweet, line 877: No need for this comment.

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -817,6 +863,26 @@ public boolean hasGlyph (char ch) {
...
+					assert false : "No glyph found in " + name + " for character '" + highSurrogate + ch + "'";
+					return null;
+				}
+				return page[ch - LOW_SURROGATE_MIN];
+			}
+			assert highSurrogate == null : "Saw high surrogate " + Integer.toHexString(highSurrogate) + " but no low surrogate at character '" + ch + "'";
+            //noinspection DataFlowIssue
```
NathanSweet, line 884: Please remove IDE-specific warning comment.

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
```diff
@@ -844,6 +910,9 @@ public void getGlyphs (GlyphRun run, CharSequence str, int start, int end, Glyph
 				if (ch == '\r') continue; // Ignore.
 				Glyph glyph = getGlyph(ch);
 				if (glyph == null) {
+					if (Character.isHighSurrogate(ch)) {
+						continue;
+					}
```
NathanSweet, line 915: Minor but ideally brace style is consistent within a file, in this case "no brace for single line statement".

extensions/gdx-tools/src/com/badlogic/gdx/tools/bmfont/BitmapFontWriter.java
```diff
@@ -224,14 +224,9 @@ public static void writeFont (BitmapFontData fontData, String[] pageRefs, FileHa
...
-
-			for (int j = 0; j < fontData.glyphs[i].length; j++) {
-				if (fontData.glyphs[i][j] != null) {
-					glyphs.add(fontData.glyphs[i][j]);
-				}
-			}
+		extractGlyphs(fontData.glyphs, glyphs);
+		if (fontData.surrogates != null) {
+			extractGlyphs(fontData.surrogates, glyphs);
 		}
```
NathanSweet, line 230: We could `glyphs.shrink();` here.

tests/gdx-tests/src/com/badlogic/gdx/tests/BitmapFontUnicodeTest.java
```diff
@@ -0,0 +1,199 @@
...
+		Colors.put("PERU", Color.valueOf("CD853F"));
+		renderer = new ShapeRenderer();
+		renderer.setProjectionMatrix(spriteBatch.getProjectionMatrix());
+		stage = new Stage(new ScreenViewport());
+
+		Skin skin = new Skin(Gdx.files.internal("data/noto-sans-symbol2.json"));
+
+		BitmapFont labelFont = skin.get("noto-sans-symbol2", BitmapFont.class);
+
+		// Notice that the last [] has been deliberately added to test the effect of excessive pop operations.
```
NathanSweet, line 62: This comment doesn't seem to match the code.

NathanSweet (commented):
Made a few comments but overall it looks good to me!

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
tommyettinger: The warning is IDE-specific, so naturally the comment suppressing it is IDE-specific. Eclipse apparently doesn't understand this DataFlowIssue or doesn't warn about it? Are you telling people to configure their IDE, which is not the same as yours, the exact way you do again?

Please *don't* remove this warning suppression comment. It doesn't interfere with reading the code at all and makes a warning that, in this context, is probably unavoidable not affect the build output. It does make it clear in the source that at least some IDEs can notice a data flow issue here, which if I recall correctly affects how HotSpot is able to optimize exceedingly complex blocks of code.

Frosty-J: There are no usages of `noinspection` in libGDX. What are the chances this can be the first?

SimonIT:
Do you mind merging the master branch into your branch so that the CI can run?

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
NathanSweet: @tommyettinger I am very strongly against littering libgdx source with IDE-specific warning suppression or with `@SuppressWarnings` for warnings that should be disabled instead. PRs with such should not be merged. See my [explanation](https://github.com/libgdx/libgdx/pull/7605#discussion_r1996548911) for warnings in libgdx.

tommyettinger: Well, all I have to say is, "Tough luck pal, no man is an island."
Vote screenshot
We held a vote on this, all the people who are actually involved in the libGDX community, on Discord, where we actually help people who have problems with the library. Where you aren't present. And so have never helped a single person in that community. Literally *no one* whose opinion matters to the current community agrees with you. You linked to an explanation, which we are all tired of, and I had actually replied to that "explanation" but you never responded. ***We are all sick of having this argument over and over in every PR.*** You need to admit that your opinion is shared only by **you**. It might look like Frosty-J is agreeing with you, but even he voted in favor of using SuppressWarnings (note that he's James in the screenshot). He uses Android Studio, anyway, so he definitely doesn't have the legendary no-warning state.

This PR should be merged with the warning suppressed, since that is what 9 people who are actively part of the community agree on, one says it doesn't matter to him, and only 1 person disagrees.

Frosty-J: That's not for you to say, whether I agree with Nate. javac and `@SupressWarnings` is different from `//noinspection`, which is only used by IntelliJ IDEA, Android Studio, and maybe other JetBrains stuff like Fleet. Personally, I would use it, but we have to consider longevity. A decade ago, Eclipse was the most popular IDE. Who's to say JetBrains/AS won't get dethroned over the next 10 years?

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
Frosty-J: Assertions aren't used much in libGDX. The `Gdx.app` logging might be more appropriate?

gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
tommyettinger: I'm in favor of SuppressWarnings when needed myself, because it's part of the JDK, but the best option would be to fix the situation that causes the DataFlowIssue in the first place. Looking at https://www.jetbrains.com/help/inspectopedia/DataFlowIssue.html , we can see that DataFlowIssue tends to occur when a block of code is redundant, which this next line actually is when assertions are enabled. The assertion before this line would end the program if highSurrogate was not already null and assertions were enabled, so assigning null to highSurrogate again is unnecessary. **You should be able to delete both the contested noinspection comment and the next line `highSurrogate = null;` .**

There, wasn't that useful information that IDEA gave us? If we had just deleted the comment, we wouldn't have noticed its nice helpful warning was given for good reason. If the assertion can be incorrect in a valid program, then it should be removed instead, and the next line kept intact. In that case, the noinspection comment should be shown as no longer necessary, and it can be removed.

tommyettinger (changes requested):
Indeed, IDEA is a quality product made by competent developers! I'm glad it gave such useful information to us as PR reviewers.

JamesXNelson:
@NathanSweet sorry for the excessive delay.  I got this PR revived if you'd like to take a peek.

tommyettinger (approved):

PR #7629, opened 2025-04-19
ci: Update to Ubuntu 22.04 runner

SimonIT:
Ubuntu 20.04 runners aren't available anymore

actions/runner-images#11101


Berstanio (approved):

tommyettinger (approved):
I barely touch GH Actions, but from what I understand, the 20.04 runners aren't even going to be available soon, so this isn't an "if" but a "when" we merge this, right?

Frosty-J:
They haven't been available since 2025-04-15.

PR #7632, opened 2025-04-26
[scene2d.ui] HorizontalGroup and VerticalGroup set transform false by default.

NathanSweet:
HorizontalGroup and VerticalGroup are intended for UI and should have transform set to false, like Table, Container, and Stack. It was just never noticed until now!

This is a breaking change, since someone could be relying on transform being true, though that is probably unlikely.


Tom-Ski (approved):

NathanSweet:
Ha, I punished myself with this one, had a toolbar that needed it. Hope it doesn't get anyone else!

PR #7637, opened 2025-04-30
Update spotless gradle plugin version to let libgdx compila with java 21

patton73:
Update works and libgdx can compile with java 21 without problems.


tommyettinger (dismissed):
It's interesting that running Spotless 7 caught some style violations that had lingered for... a while, in publish.gradle. That seems to also be the reason why some robovm-metalangle files changed. Since Spotless appears to have run just fine with the newer version, and does so without needing com.sun APIs removed in Java 21 or maybe earlier, this seems like a good thing to update! Approved.

tommyettinger (dismissed):
I think updating the PR is usually not preferred because it adds changes unrelated to it, BUT, this is a Spotless PR so if there were changes in 1.13.5 that Spotless 7.0.3 won't tolerate, we do need to have Spotless run on them, too. Approved, again!

patton73:
Just for curiosity? who is authorized to merge it?

tommyettinger:
We try to wait for two people to review any PR so a second set of eyes can catch issues the first set didn't. Once anyone else can review and approve, there's various contributors who can merge (like me!).

tommyettinger:
Ah, yeah, same author. There must be some cross-pollination of changes from the other PR.

patton73:
Oh. Very strange i used different branches to make the pull requests. I try to remove those commits from this branch, i am not completely sure how to do this.

patton73:
It should be polished.

SimonIT (approved):

PR #7638, opened 2025-05-02
Add the possibility to load a standalone tileset file

patton73:
Now the method loadTileSet is public and can be used outside the loading on complete tiled map. All relative test still run without problems.


patton73:
@BoBIsHere86 @Quillraven You could be interestedn in this one.

Quillraven:
For me personally this is a cool addition, since I work with "collection of images" tilset to configure my game related objects like NPCs, items, enemies, props, ...

Currently, if I need to load an item (either in game or for a test), I need to load a map that contains the "game objects tileset" to get access to it. At least if I want to stick to LibGDX features.

Having it as a separate functionality to load my game objects tileset will be great.

The only thing I wonder is, if it makes sense to have it as a separate loader as well ? Because as is, if you have an `AssetManager` and just want to load a tileset you cannot do it the normal way via `load`/`unload` right? Not sure if we need/want that.

I won't need it because that object loading is usually a one-time process when the game is starting up, parsing everything and converting it into my game related structures.

patton73:
Well basically i was thinking about generating random maps using same tileset of some scenario package. The isea was to have a scenario package where there are some Tiled's static handmap maps with their external tileset files that can be reused/reloaded also to generate a random map in the game. 

Dunno if its better to have a separate Tileset Loader. However to me is the same.  If we can extract that code we can also refactor completely since right now there are methods with tons of parameters (very ugly). I would as first step accept this kind of refactoring and then perhaps take the code out. As I said to me is the same.

Quillraven:
Then let's keep it simple in this PR with the changes you did. Separate loader/refactoring in a separate PR.

Quillraven (approved):

Quillraven:
Time to mergy merge? 😉 Would be cool to have it in the upcoming release

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
```diff
@@ -690,98 +693,94 @@ else if (compression.equals("zlib"))
...
-
-			JsonValue offset = element.get("tileoffset");
-			int offsetX = 0;
-			int offsetY = 0;
-			if (offset != null) {
-				offsetX = offset.getInt("x", 0);
-				offsetY = offset.getInt("y", 0);
-			}
-			TiledMapTileSet tileSet = new TiledMapTileSet();
+	public TiledMapTileSet loadTileSet (JsonValue rawElement, FileHandle tmjFile, ImageResolver imageResolver) {
```
tommyettinger, line 696: Normally docs on this would be important, but I'm inclined to be lenient because it's an AssetManager Loader and those seem to be less critical for docs. I think anyone who needs to be really digging into AssetManager problems should be looking at the source code anyway. I'm a little unclear why the JsonValue is `raw` and what that means here. I think I could probably figure it out by how it's used, assuming it's used somewhere in the code, though it doesn't look like there's a new test in this PR.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
```diff
@@ -690,98 +693,94 @@ else if (compression.equals("zlib"))
...
+		}
+		for (AnimatedTiledMapTile animatedTile : animatedTiles) {
+			tileSet.putTile(animatedTile.getId(), animatedTile);
+		}
 
-			map.getTileSets().addTileSet(tileSet);
+		return tileSet;
+	}
 
+	private JsonValue resolveTilesetElement (JsonValue element, FileHandle tmjFile) {
```
tommyettinger, line 755: I feel like with almost everything else here `protected` or `public`, adding new `private` methods is something to be careful with. If they will never need customization, or if users can be expected to write their own resolution methods, this is probably fine, but I think `protected` would likely be better for future usage in unforeseen circumstances. Not a dealbreaker.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
```diff
@@ -690,98 +693,94 @@ else if (compression.equals("zlib"))
...
+
+		return getRelativeFileHandle(base, imageSource);
 	}
 
 	protected abstract void addStaticTiles (FileHandle tmjFile, ImageResolver imageResolver, TiledMapTileSet tileSet,
 		JsonValue element, JsonValue tiles, String name, int firstgid, int tilewidth, int tileheight, int spacing, int margin,
-		String source, int offsetX, int offsetY, String imageSource, int imageWidth, int imageHeight, FileHandle image);
+		int offsetX, int offsetY, String imageSource, int imageWidth, int imageHeight, FileHandle image);
 
 	private void addTileProperties (TiledMapTile tile, JsonValue tileElement) {
```
tommyettinger, line 785: It isn't in your code, but addTileProperties is private here and protected for BaseTmxMapLoader. That's odd.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmxMapLoader.java
```diff
@@ -735,96 +738,94 @@ else if (compression.equals("zlib"))
...
-					imageHeight = imageElement.getIntAttribute("height", 0);
-					image = getRelativeFileHandle(tmxFile, imageSource);
-				}
-			}
-			String name = element.get("name", null);
-			int tilewidth = element.getIntAttribute("tilewidth", 0);
-			int tileheight = element.getIntAttribute("tileheight", 0);
-			int spacing = element.getIntAttribute("spacing", 0);
-			int margin = element.getIntAttribute("margin", 0);
+	public TiledMapTileSet loadTileSet (Element tilesetElementRaw, FileHandle tmxFile, ImageResolver imageResolver) {
```
tommyettinger, line 741: `loadTileSet` is inconsistent with the capitalization of other overloads, which use `loadTileset`. I think Tileset is correct here, if TMJ-format files call them tilesets as one word.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
tommyettinger: Also, the capitalization is different here too, `TileSet` should be `Tileset` to match the rest of the code. I'm not sure if this is a safe change to make, though, because it was TileSet before in this method only. Ugh... We should keep this as it is unless we know nothing relies on the old name...

tommyettinger (approved):
I'm not happy with the API this has to maintain from before, mainly because `TileSet` is not the same as `Tileset`, but the changes here look fine to me. More docs would be nice in a perfect world, but TmjMapLoader and TmxMapLoader have sufficient docs on their loading methods, so I'm not too worried about the newly-public BaseTm?MapLoader classes not having docs. How they are used is also clear enough. This seems like a good set of changes in practice; approved.

patton73:
Do you want me to fix those issues you found?

tommyettinger:
> Do you want me to fix those issues you found?

None were sufficiently problematic that I couldn't approve as-is! There isn't exactly a rush on merging this because 1.14.0 just came out less than 24 hours ago, and there isn't even an announcement yet. It will hopefully be merged before the next release of libGDX (1.14.1 or 1.15.0?); I see no reason why it can't be by then... I kinda just need to know if Quillraven's review with read-only permissions is considered as "counting as a review" by the rest of the team, though I think it's in-depth enough for sure.

Urban-Elf (approved):

PR #7648, opened 2025-05-23
fix: Modify Pools api to prevent android/desugar issues.

PokeMMO:
As discussed in Discord, there are currently  issues using the Supplier API on Android and maybe other desugar platforms. This is a lackluster revert of the changes while  maintaining the lack of reflection for all libgdx internal Pools.

The get/obtain methods will revert to using the .class style requests.

Optional warnings/throw can be enabled to help prevent reflection issues.

Users who use the `Pools` api should pre-populate the Pool by using `Pools#set`.


gdx/src/com/badlogic/gdx/utils/Pools.java
```diff
@@ -16,104 +16,126 @@
...
 
 /** Stores a map of {@link Pool}s by type for convenient static access.
  * @author Nathan Sweet */
 public class Pools {
 	static private final ObjectMap<Class<?>, Pool<?>> typePools = new ObjectMap<>();
-	static private final ObjectMap<Class<?>, Pool<?>> supplierPoolsCache = new ObjectMap<>();
+	static public boolean WARN_ON_REFLECTION_POOL_CREATION = true;
+	static public boolean THROW_ON_REFLECTION_POOL_CREATION = false;
+
+	static {
```
tommyettinger, line 39: I uh... don't like this at all. It breaks any kind of separation of concerns, since any change in libGDX that uses Pools needs to also change the source of Pools. Why on Earth can't we create a DefaultPool in the file where one is used? If we did that, we could also choose whether or not for it to be static, in case someone potentially doesn't want threading to break due to a static pool. I don't know if we have any benchmarks for recent Android (post-Dalvik, since our MinSDK now means we won't ever run on a Dalvik-using OS) that can verify if pooling has a performance benefit, or if it's just breaking threading for no current reason. I don't have any clue how people benchmark on Android, either.

gdx/src/com/badlogic/gdx/utils/Pools.java
```diff
@@ -16,104 +16,126 @@
...
+	 * is 100. */
 	static public <T> Pool<T> get (Class<T> type) {
 		return get(type, 100);
 	}
 
-	/** Returns a new or existing pool for the specified type, stored in a Class to {@link Pool} map. Note the max size is ignored
-	 * if this is not the first time this pool has been requested. This method may create a temporary object once, but the
-	 * allocation will probably be eliminated on most platforms. <br>
-	 * Usage can use java 8 method references: {@code Pools.get(MyClass::new, max)} */
-	static public <T> Pool<T> get (PoolSupplier<T> poolTypeSupplier, int max) {
```
tommyettinger, line 54: It may have only been in for one version, but this was still the only non-deprecated API to use Pools in 1.13.5. I think it's a bad idea to remove it wholesale, without warning.

PokeMMO: Sure, I can readd it with a Deprecation warning, I just figure it would be simple to migrate back for the few who already have migrated theirs. Since they already did it once I mean.

gdx/src/com/badlogic/gdx/utils/Pools.java
PokeMMO: As I said, we could do this in the individual files, like the `InputEvent` example in the original topic.

Unfortunately, we can't really create a `DefaultPool` in all of the cases. There's at least one api which accepts a `<T extends Action>` which we can't hold the unlimited potential `Pool`s locally. 

https://github.com/libgdx/libgdx/blob/c9638eabb78290bdfcd18b38570f4f91343f8a99/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/Actions.java#L46


I personally, do manage all of the Pools in my own codebase myself, but the `Pools` api is public. If we want this continue to work, I don't see a better option compatible with old Android and no-reflection usage.

obigu: Following these premises, shouldn't we at least replace internal usages to preferred `DefaultPool` where possible leaving `Actions` to be the exception?

gdx/src/com/badlogic/gdx/utils/Pools.java
obigu: I wouldn't readd with deprecation tbh (as long as we release in a short period of time of course), for the few users who may have made the change it should be simple to migrate back.

gdx/src/com/badlogic/gdx/utils/Pools.java
Tom-Ski: Yes, i think those are probably out of scope for the reason of the PR

gdx/src/com/badlogic/gdx/graphics/g2d/GlyphLayout.java
```diff
@@ -45,7 +45,7 @@
  * @author Alexander Dorokhov
  * @author Thomas Creutzenberg */
 public class GlyphLayout implements Poolable {
-	static private final Pool<GlyphRun> glyphRunPool = Pools.get(GlyphRun::new);
+	static private final Pool<GlyphRun> glyphRunPool = Pools.get(GlyphRun.class);
```
Berstanio, line 48: Maybe this is intended for a follow-up pr, but places like this could be easily replaced with a `new DefaultPool(GlyphRun::new)`, right?

PokeMMO: Yes, as mentioned in the other comment thread on this PR it would be better to do separately.

https://github.com/libgdx/libgdx/pull/7648#discussion_r2104114250

PokeMMO:
`Pools#set` calls have been moved to each Pooled class's static block. This should resolve any treeshaking concerns.

PokeMMO:
Moving Pools#set calls to individual classes doesn't actually work. The static blocks are not called before the pool is attempted to be used.

Berstanio (approved):

PR #7650, opened 2025-05-29
Update freetype to v2.13.3

PokeMMO:
Compiles but untested


Tom-Ski:
Probably should merge so can test easier without having to build on all platforms locally?

PR #7652, opened 2025-05-31
fix wrong x, y layer offset in HexagonalTiledMapRenderer

willie-deev:
the wrong layer offset made wrong tiles rendered

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/renderers/HexagonalTiledMapRenderer.java
@@ -128,13 +128,13 @@ public class HexagonalTiledMapRenderer extends BatchTiledMapRenderer {
 			final float tileWidthUpperCorner = (layerTileWidth + layerHexLength) / 2;
 			final float layerTileHeight50 = layerTileHeight * 0.50f;
 
-			final int row1 = Math.max(0, (int)((viewBounds.y - layerTileHeight50 - layerOffsetX) / layerTileHeight));
+			final int row1 = Math.max(0, (int)((viewBounds.y - layerTileHeight50 - layerOffsetY) / layerTileHeight));
 			final int row2 = Math.min(layerHeight,
-				(int)((viewBounds.y + viewBounds.height + layerTileHeight - layerOffsetX) / layerTileHeight));
+				(int)((viewBounds.y + viewBounds.height + layerTileHeight - layerOffsetY) / layerTileHeight));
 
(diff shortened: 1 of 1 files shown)


tommyettinger:
Can you show any tests as evidence that the behavior was wrong before? That changed code was stable for about 8.5 years with no complaints ([it was last changed here by @piotr-j in 2016](https://github.com/libgdx/libgdx/commit/f35c50b5f1d7967548dd30d078dfb326c2b9f6e9)), and tests were passing since @lyze237 fixed hexagonal rendering a few years ago. It's entirely possible the offsets were never tested before, but for over 8 years of usage to pass without any reports of the offsets being wrong at all seems... strange. Maybe offsets are just a rarely-used feature.

willie-deev:
i think maybe it's really a rarely used feature

this is a video about it
https://streamable.com/o3lwoj
and the repo of the testing project
https://github.com/willie-deev/hexagon_example

lyze237:
What I did was change the stagger index, I don't think I ever used layer index. So good catch! :) 
I'll have a look at this PR in a bit.

BoBIsHere86:
Wow, this one was hard to see.
So using the tests we currently have I honestly couldn't see a difference between any of the hexmaps in the original code vs the new code. That is until I realized I was looking for the wrong thing. The tests do an extremely poor job of showcasing this very specific issue because of how it's related to the camera.

In order to spot what was up. I went into one of our Hexagonal test maps and I ramped up the Y offset of one of the layers. To -304 or so, from -24.
And I now see where it's breaking.
I have made 2 versions of a gif of myself moving the Hex map around. First in the Current code and Second in the Newly fixed pr code.
In the current code with the altered Yoffset of around -304. Pay attention to the Dark Green and Orange hexs and when they stop being rendered as the camera moves DOWN, and also when it moves to the RIGHT. They stop being rendered waaaay before it comes close the the camera edge, which is represented by that very hard to see thin red rectangle. You will also notice they are rendered out way further then they should be when going UP or to the RIGHT.
OffsetCurrent



Now here's a gif running that same exact map file with the Fix in place. As you can see now the hexes all render up until the edge and disappear where they would be expected to.
OffsetPR


Other than that I did not see any issues popping up when running the tests in this PR.

So just dropping off my 2 cents. Hope it helps.
Cheers!

BoBIsHere86:
Just popping into the Tiled related PRs before the next big release. Hope we can clean these up.
I believe this was a great catch, and no reason it should not be fixed.
Since we always like to get multiple eyes on things, I am tagging @Quillraven who is a Tiled user to see what he thinks.

Quillraven:
this one looks fine to me and I actually already had a look a couple of months ago where I upvoted your respone :)

Issue #7651, opened 2025-05-30
[Tiled] Class support


Quillraven:
I talked to @BoBIsHere86 in discord and we agreed that it would be cool to also support the "Class" functionality of Tiled.
It is similar to "class properties" that we have added in a previous PR.

You can assign a class to an object/tile in Tiled (see screenshot). It then gets the properties that you have defined for that class in your Tiled project.
All of that is working at the moment unless you choose the "default value" of a property and don't override it.

Like in the screenshot below "atlasAsset" has the default value "OBJECTS" which you see in Tiled but when you load it in LibGDX the object/tile doesn't have an atlasAsset property at all.

So when an object/tile has a class, we need to go through all properties of that class and if it is not overruled, then load the default value which is specified in the project file (which we already parse in LibGDX with the loaders).

![Image](https://github.com/user-attachments/assets/deae11eb-ae2e-40d2-981a-7f59233d2bba)

PR #7653, opened 2025-06-02
[Tiled] Class support 

Quillraven:
Related issue: #7651

This PR supports "default value" loading for tiles/objects in Tiled that use a project class. In Tiled, if you don't overrule the default value of a project class property then it is still shown in a "greyed out" text in the editor:
image

LibGDX currently doesn't load such properties. This PR changes that so that the `MapProperties` of a tile/object correctly contain the default values.

----

I added one additional very minor change that annoyed me personally as well, which is the `protected` scope of the `BatchTiledMapRenderer.renderMapLayer` method. If you have a more advanced rendering process where you try to batch together as much things as possible, you currently have troubles doing that because the public render methods of tiled map render instances internally call batch begin/end. 

However, at least I, have my own batch processing and I call batch begin/end myself. My current workaround is to extend the BatchTiledMapRenderer and provide a public renderMapLayer method that calls the protected one. This is a little bit annoying and this change prevents that.

----

@BoBIsHere86: whenever you have some free time, please test this stuff with your things as well.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTiledMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTiledMapLoader.java
@@ -315,7 +315,7 @@ public abstract class BaseTiledMapLoader<P extends BaseTiledMapLoader.Parameters
 		return color + alpha;
 	}
 
-	protected void loadMapPropertiesClassDefaults(String className, MapProperties mapProperties) {
+	protected void loadMapPropertiesClassDefaults (String className, MapProperties mapProperties) {
 		if (className == null || !projectClassInfo.containsKey(className)) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


BoBIsHere86:
These are my favorite kind of changes. The kind that barley touch much existing code. 😄
I ran through all my normal tests, including the atlas with properties file test, as well as the GWT tests. Specifically the properties test.
Everything looks and loads as it should.
Also, I am totally fine with the protected to public change. This makes sense to me as it gives the option to anyone who finds themselves looking for the need to do the same.

👍

Quillraven:
Thanks for having a look!

@obigu @tommyettinger: I think one of you merged our last Tiled PRs? Whenever you feel comfortable please go ahead 😉

Quillraven:
@Tom-Ski / @SimonIT : I saw you merged some previous tiled PRs. Can you please merge this one to?

patton73:
Ehm guys @Tom-Ski / @SimonIT there are a lot of Tiled's PRs to merge. Can you please have a look at this bug? https://github.com/libgdx/libgdx/issues/7488 and follow "tilemap" label to merge verified prs? Thanks.

PR #7655, opened 2025-06-17
increment version

Berstanio:
Was probably forgotten during last release


Quillraven:
@Berstanio badges in the Readme are also outdated. Do you know how to fix it? It is not just simple entering a string 😅

PR #7657, opened 2025-06-23
Fixed a GWT incompatibility issue in the AtlasTmxMapLoader and AtlasTmjMapLoader + new GWT specific Test

BoBIsHere86:
Thanks to the latest libGDX jam I just realized a bug in relation to loading AtlasTmxMapLoader with ImageLayers in GWT.

The fool who added the feature to support imagelayers in atlas maps stupidly used FileHandle as a shortcut way to parse file path strings... *cough* *cough*

Anyway this caused an issue in GWT so I fixed it with a simple string parse instead, so all is well.
But also leads into another discussion about the lack of tests in GWT for tiled stuff. Sadly most of our tiledmaps are saved using compression options. So when the existing TiledMapAtlasAssetManagerTest runs in GWT tests it just says. "ERROR (OK if running in GWT): "
Which may have led to me missing this error in the first place.

Well I've gone ahead and fixed that. I've created a brand new GWT only version of that specific test. With matching tiled files but this time using the uncompressed base64 format. 
So now we actually have a working test for the Atlas rendering in GWT.
I also snuck in one other test that I noticed used an uncompressed tiled format.
GwtTiledAtlasTest

I think my next goal for Tiled might be to begin adding more GWT compatible tests. But it's a fairly manual process to convert the maps to keep them all in sync. So it might take some time.

Anyways this is a pretty small, but much needed fix to the newly added Atlas stuff. So If there's anything wrong here please let me know.


gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmjMapLoader.java
```diff
@@ -245,8 +245,11 @@ protected void setTextureFilters (Texture.TextureFilter min, Texture.TextureFilt
 	 * @return The name of the region to pass into an atlas */
 	private static String parseRegionName (String name) {
 		if (name.contains("atlas_imagelayer")) {
-			// Extract the name of region from path
-			return new FileHandle(name).name();
+			// Find the last '/' in the path
+			int lastSlash = name.lastIndexOf('/');
```
tommyettinger, line 249: I'll have to check further, but I'm not sure either of these will work if the region is a TextureRegion loaded from a TextureAtlas rather than a file. The reason the old way failed was because `new FileHandle(String)` shouldn't really be used on any platform to my knowledge... I always use `Gdx.files.internal()` or some other way of creating the appropriate FileHandle type, and it isn't clear here if `internal` would even be appropriate in this code always.

In the case of the new code, a TextureAtlas can contain names of regions with slashes (and maybe also backslashes in rare cases, I don't know), and just the section after the slash won't be enough to locate the right TextureRegion. If "region" means something different here, feel free to disregard all of this.

tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapGWTAtlasAssetManagerTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+	private TiledMap map;
+	private TiledMapRenderer renderer;
+	private OrthographicCamera camera;
+	private OrthoCamController cameraController;
+	private AssetManager assetManager;
+	private BitmapFont font;
+	private SpriteBatch batch;
+	String errorMessage;
+	private String fileName = "data/maps/tiled-atlas-processed/test_gwt_uncompressed.tmx";
```
tommyettinger, line 49: More GWT-compatible tests are always good!

gdx/src/com/badlogic/gdx/maps/tiled/AtlasTmjMapLoader.java
BoBIsHere86: Ah I can probably help clarify a bit. All this imagelayer handling was a recent addition to the map atlas stuff(added with the recent updates), since it was originally never supported. If you had any image layers in your atlas map. It just crashed on load. 
Anyway, when getImage(String name), is called the tiledmap code passes in the path to the image so for example it would be /assets/maps/backgrounds/CoolBackground2.png. But since the image was now in an atlas this wouldn't work.

Now, when we process CoolBackground2.png. We change the name a bit to make it unique while also signifying it's a processed atlas image layer, so in the texture atlas it looks something like atlas_imagelayer_CoolBackground2_433236548469800x
So now when that  getImage(String name) gets called libgdx is passing in "/assets/maps/backgrounds/atlas_imagelayer_CoolBackground2_433236548469800x" now.
So I just needed a way to get that last portion "atlas_imagelayer_CoolBackground2_433236548469800" to reference the region. Which is also what parse region means in that context. Im just trying to get the region name from that filepath.

I didn't think it would be possible for any other  "/"s to end up in that string we parse outside of the folder location. Since  we manually process and put the filename of the image into the region ourselves and the filenames wouldnt be able to have any /s in them i assumed. And this code change is restricted to only affect the atlas loaders specifically imagelayer processing. and nothing else.

Hope that helps.

Tom-Ski (approved):

PR #7658, opened 2025-06-24
[Tiled] Optimization for #7653

Quillraven:
This PR fixes a NullpointerException introduced in #7653.
It happens when you use class properties or classes of an object and don't set a `projectFilePath` parameter.

I fixed that area now and log a warning once if that happens. Imo it should not throw an error because maybe a user has some custom code already in place that handles something for him and he doesn't want to use the new built-in functionality.

I used `Gdx.app.log`. I hope this is allowed? Or is there a better option to log a warning?

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/BaseTiledMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/BaseTiledMapLoader.java
@@ -318,10 +318,8 @@ public abstract class BaseTiledMapLoader<P extends BaseTiledMapLoader.Parameters
 
 	protected void loadMapPropertiesClassDefaults (String className, MapProperties mapProperties) {
 		if (projectClassInfo == null) {
-			Gdx.app.log("TiledMapLoader",
-					"WARN: There is at least one property of type class or an object with a class defined. " +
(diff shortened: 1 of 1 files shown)


Frosty-J:
> Or is there a better option to log a warning?

You can use `Gdx.app.error()` - this doesn't have to throw an exception or be fatal. You be the judge of whether it's important enough for this instead of debug level.

Quillraven:
> > Or is there a better option to log a warning?
> 
> You can use `Gdx.app.error()` - this doesn't have to throw an exception or be fatal. You be the judge of whether it's important enough for this instead of debug level.

I think an error is also to much. I saw that "info" is the default log level of an Lwjgl3Application which means the current WARN info should show up.

But I don't have a strong opinion. I can also change it to an error log.

Quillraven:
@SimonIT can you please merge this PR before the "patch release" for Android will be made? I don't know when that will be, I just know it is "soon" and would be great if this optimization is part of the next release.

tommyettinger:
@Quillraven It's in the next milestone, that should mean the next release will have it merged if it doesn't stop the build or something.

Quillraven:
Cool thank you 😊

PR #7659, opened 2025-06-25
Add Lwjgl3ApplicationConfiguration.useGlfwAsync option

damn:
Merged all commits from previous PR (https://github.com/libgdx/libgdx/pull/7580) into one. See discussion there.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -26,6 +26,7 @@ import org.lwjgl.PointerBuffer;
 import org.lwjgl.glfw.GLFW;
 import org.lwjgl.glfw.GLFWVidMode;
 import org.lwjgl.glfw.GLFWVidMode.Buffer;
+import org.lwjgl.system.Configuration;
 
 import com.badlogic.gdx.Audio;
 import com.badlogic.gdx.Files;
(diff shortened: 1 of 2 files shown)


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -234,6 +235,15 @@ public void enableGLDebugOutput (boolean enable, PrintStream debugOutputStream)
 		debugStream = debugOutputStream;
 	}
 
+	/** Whether to use the "glfw_async" library.
+	 *
+	 * This means you do not have to set the JVM argument "-XstartOnFirstThread"
+	 *
+	 * @see <a href= "https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else"> Documentation</a> */
+	public static void useGlfwAsync () {
+		Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
```
tommyettinger: I gave this PR a test run and calling `Lwjgl3ApplicationConfiguration.useGlfwAsync()` causes a crash at startup on Windows 11. Since this only is helpful when called on MacOS (and is harmful everywhere else), shouldn't it have a check so it only runs on that OS? Like:
```java
if (SharedLibraryLoader.os == Os.MacOsX)
  Configuration.GLFW_LIBRARY_NAME.set("glfw_async");
```

tommyettinger (changes requested):
I think the OS check should be inside useGlfwAsync() to reduce clutter in the Launcher class. This is basically useless without a check somewhere to ensure it only runs on MacOS, unless you are only distributing your desktop apps to Macs and doing all development on Macs. Which would be silly to use libGDX for.

damn:
If we add the mac os check we would have to change the name right?
Something like 'set mac os config'?

crykn:
> This is basically useless without a check somewhere to ensure it only runs on MacOS

I don't think it's completely useless, since it still abstracts the GLFW classes. But including a macOS check makes sense to further reduce the code needed to use the custom GLFW library.

> If we add the mac os check we would have to change the name right?

I'd keep the name `useGlfwAsync` and mention in the javadocs that it only does something on macOS.

tommyettinger (approved):
Looks good to me now; approved!

Issue #7663, opened 2025-07-09
IME and navigation bar showing up when using OTG keyboard in Android


lucas-viva:
Hello!

When using an OTG keyboard in Android, the system IME and navigation bar are showing up after pressing keys (such as letter 'A').

<img width="1659" height="760" alt="Image" src="https://github.com/user-attachments/assets/57761f89-ec67-4f5b-802d-43a709a9f849" />

The side effects of this is that players cannot play in Android using keyboards anymore.

This can be reproduced by creating a new Android project and using libGDX version `1.13.1` or newer.

OTG keyboard works in version `1.12.1`.

I'm currently testing a few things to check what is causing this. I will keep you posted on any updates.

Thanks!

lucas-viva:
Findings so far:

https://github.com/libgdx/libgdx/blob/b3989cb385ab2f308756e6735eb90eb41cfec558/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/surfaceview/GLSurfaceView20.java#L83

For `1.12.1`, `GLSurfaceView20.onCreateInputConnection` method, when modifying `outAttrs.inputType`, it would set it as `InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`, which would not trigger IME.

https://github.com/libgdx/libgdx/blob/7bfcd983a25b1943935fbc0d2aed052bf5c49cd3/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/surfaceview/GLSurfaceView20.java#L82

For `1.13.1` on, `InputType.TYPE_CLASS_TEXT` (plus more options) is returned, which triggers IME.

lucas-viva:
I have few context on the Android classes, but it looks like a possible fix is:

1- `GLSurfaceView20.onscreenKeyboardType` should be null if onscreen keyboard is not visible 
2- `DefaultAndroidInput.getAndroidInputType` should return `InputType.TYPE_NULL`  if that is the case

PR #7665, opened 2025-07-10
Fix OTG keyboards popping. #7663

lucas-viva:
This change fixes issue #7663.

1- `DefaultAndroidInput` has a new protected method, `getInputType`, so users can extend the class and override the conversion behavior if desired (removed the static method).

2- `GlSurfaceView20` no longer depends on `DefaultAndroidInput` (they were depending on each other).

3- `DefaultAndroidInput.setOnscreenKeyboardVisible` was refactored to make the code cleaner.

4- `DefaultAndroidInput.onKeyboardHeightChanged` will call `setOnscreenKeyboardVisible` to properly reset `GlSurfaceView20.inputType` (this was the main issue).

5- `Input` had some whitespaces stripped.


Berstanio:
`OnscreenKeyboardType` is also used in GWT and iOS, both would need to support this new type to properly. You could also just make `onscreenKeyboardType` nullable and handle it through that.

Also, `setOnscreenKeyboardVisible` doesn't have to be called, when the IME closes (keyboards can be closed by the system too), leading to not resetting the field in `GLSurfaceView20`. This is an issue in the previous code too tho.

Regarding a fix, I think the best way is to treat `onCreateInputConnection` as consuming the `onscreenKeyboardType`.
This breaks `if (view.inputType != inputType) {` check ofc. I think we might get away by always calling `restartInput`, tho that calls `onCreateInputConnection` always, so that needs to be considered when consuming.
All this options are not great tho, if you have a better idea, feel free to propose it.

 Regarding the PR in general a few nitpicks:
- I don't think creating a new class just for one method is great
- `doSetOnscreenKeyboardVisible` making a new method is kinda pointless, since it is neither huge nor used in multiple places

Berstanio:
We can also go back to just `TYPE_TEXT_VARIATION_VISIBLE_PASSWORD` for `defaultDisableAutocorrection` imo.
If that doesn't mess with the IME, as it reads from the original issue?

lucas-viva:
Thanks for your review @Berstanio.

Based on your feedback, I changed the code again. The final changes now are:

1- `DefaultAndroidInput` has a new protected method, `getAndroidInputType` so users can extend the class and override the conversion behavior if desired. The static method was removed since we don't need it for the item 2 below.

2- `GlSurfaceView20` no longer depends on `DefaultAndroidInput` (they were depending on each other).

3 - We are now using `onKeyboardHeightChanged` to check if the keyboard was closed and calling `setOnscreenKeyboardVisible` if needed. This should reset the IME input type to `TYPE_NULL` correctly now.

Let me know what you think!

Berstanio:
1. and 2. look fine to me!

> 3 - We are now using onKeyboardHeightChanged to check if the keyboard was closed and calling setOnscreenKeyboardVisible if needed. This should reset the IME input type to TYPE_NULL correctly now.

Unfortunatly, this does not work for floating keyboards, as they always report a height of "0".
See what I did here: https://github.com/libgdx/libgdx/blob/3e015ac8e6e373e657744114976c6b68c4003379/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java#L661-L665

There is, to my knowledge, no universal way to detect when an Onscreenkeyboard is closed. So, instead of calling `setOnscreenKeyboardVisible(false);`, which will try to close the keyboard, you can try to just reset the `inputType` to `TYPE_NULL`, when you get a `height` of `0`. We might get away with this for floating keyboards too, depending on the timing/order of method executions. However, this would need to be tested.

lucas-viva:
Thanks for the information!

Argh Android ;(

So, if I understood correctly, `AndroidXKeyboardHeightProvider` (Android 23+) can reliably tell us whether the keyboard is opened through `windowInsets.isVisible(WindowInsetsCompat.Type.ime())`, even for floating keyboards.

And `StandardKeyboardHeightProvider` (Android 21+, libGDX min version) cannot do for floating keyboards.

Could we:

1- Introduce a new method to `KeyboardHeightObserver` called `onKeyboardVisibilityChanged` and implement it for `AndroidXKeyboardHeightProvider`.

2- Do not implement it for `StandardKeyboardHeightProvider` and this will effectively maintain the previous behavior.

3- Eventually, when libGDX min version is 23+, we can get rid of `StandardKeyboardHeightProvider`.

Berstanio:
> So, if I understood correctly, AndroidXKeyboardHeightProvider (Android 23+) can reliably tell us whether the keyboard is opened through windowInsets.isVisible(WindowInsetsCompat.Type.ime()), even for floating keyboards.

Unfortunatly, not. Maybe let me explain the code snipped linked a bit:
`!isStandardHeightProvider` the check is there, because the `StandardKeyboardHeightProvider` is just to buggy to rely on `height = 0` in any case.
`keyboardHeightProvider.getKeyboardLandscapeHeight() != 0 || keyboardHeightProvider.getKeyboardPortraitHeight() != 0`
This check checks, whether  the keyboardHeightProvider _ever_ encountered a different height then "0". If it ever encountered any real height, we assume we don't deal with a floating keyboard.
That the `windowInsets` don't report a height on floating keyboards makes sense if you think about it, because if the keyboard floats there is no "inset" to exist at any side to account for.
The main issue is here, that Android does not provide a way to detect, when the keyboard is opened/closed.

Regarding `StandardKeyboardHeightProvider` vs `AndroidXKeyboardHeightProvider`, the code comments and official docs lie a bit here. While the official android docs _claim_ it works pre android 30, it just simply doesn't. So we use `AndroidX` provider for 30+, and Standard for 21-30. 

So yeah, your proposed solution unfortunatly doesn't work :/

Best idea I would have are either:
> you can try to just reset the inputType to TYPE_NULL, when you get a height of 0. We might get away with this for floating keyboards too, depending on the timing/order of method executions. However, this would need to be tested.

or

> We can also go back to just TYPE_TEXT_VARIATION_VISIBLE_PASSWORD for defaultDisableAutocorrection imo.

lucas-viva:
Thanks for the follow up Berstanio!

I tested `AndroidXKeyboardHeightProvider` which uses the window insets call for GBoard and Samsung Keyboard and:

1- the height for floating keyboards was indeed incorrect, but:
2- `windowInsets.isVisible(WindowInsetsCompat.Type.ime())` returned correctly

So the proposed solution would work for Android 30+, using `AndroidXKeyboardHeightProvider`?

About the other ideas:

> reset the inputType to TYPE_NULL

I tested it and it won't call `onCreateInputConnection` in `GLSurfaceView20` again, not resetting the IME, causing the OTG keyboard bug.

> go back to just TYPE_TEXT_VARIATION_VISIBLE_PASSWORD

That works! If we can't agree in another viable solution, we can use this and document the code with all the details.

Berstanio:
> windowInsets.isVisible(WindowInsetsCompat.Type.ime()) returned correctly

That great news, seems like I did something wrong when I tested it last time. I get these results now too!
So yes, the proposed solution should work for android 30+.
I'm currently in the middle of rewriting the API anyway (https://github.com/libgdx/libgdx/pull/7670), I will incorperate the new insights into it (probably an extra `explicitClosed` flag to the observer), and then the fix can be implemented ontop of it way easier!

lucas-viva:
Awesome! I will wait the API rewrite then. After it is implemented, we can resume this PR. Thanks!

Berstanio:
> Awesome! I will wait the API rewrite then. After it is implemented, we can resume this PR. Thanks!

@lucas-viva Sorry it took so long, it's merged now.
The `KeyboardHeightObserver` has now two extra methods `onKeyboardShow/onKeyboardHide`

lucas-viva:
Thank you @Berstanio ! I will be able to work on this on Q1/2026!

lucas-viva:
I updated the PR title and description to reflect the current changes. The code now uses the new API to check if the keyboard was closed.

Berstanio:
> I updated the PR title and description to reflect the current changes. The code now uses the new API to check if the keyboard was closed.

Thank you, I will try to test it soon!

PR #7666, opened 2025-07-10
fix: Move Pools registration to individual classes for better initialization

badlogic:
Problem: unnecessary warnings for internal libGDX classes in `Pools#get()`, static block in `Pools` that will pull in various classes and disallow ProGuard/r8/GraalVM/RoboVM tree shaking.

Refactors the Pools class to use lazy initialization with static blocks in each poolable class instead of a central static block. This approach:
- Fixes Android/desugar compatibility issues
- Ensures proper class initialization order
- Allows ProGuard/R8/GraalVM to strip unused classes
- Prevents pulling in unused classes during startup
- Improves error messages for missing pool registrations

Each poolable class now registers itself via static block with Pools.set(ClassName::new), ensuring the pool is available before use.

See also https://github.com/libgdx/libgdx/pull/7648

### DIFF
--- a/gdx/src/com/badlogic/gdx/Net.java
+++ b/gdx/src/com/badlogic/gdx/Net.java
@@ -33,6 +33,7 @@ import com.badlogic.gdx.net.SocketHints;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.Null;
 import com.badlogic.gdx.utils.Pool.Poolable;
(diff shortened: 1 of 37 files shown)


Berstanio:
Closing this as discussed on discord, because `MyClass.class` doesn't necessarly run the static init block. So moving registration to the necessary classes is not feasable unfortunatly, even tho it would be prefarable.

Issue #7599, opened 2025-03-10
Re-Evaluate what WindowInsets to consider in `AndroidXKeyboardHeightProvider`


Berstanio:
As reported on discord, `WindowInsets.Type.systemGestures()` might lead to weird resizing of the text field.
https://discord.com/channels/348229412858101762/1347021765926457435/1347021765926457435

Furtheremore, a method like `AndroidApplication#setKeyboardHeightProvider` should be added.

Issue #7613, opened 2025-03-24
NativeInputTest issues


obigu:
Testing the new Native Input API running `NativeInpuitTest` on iOS I've run into some issues some of which may affect the API itself.

**Issue 1**
1. Click on "Placeholder" text field to open keyboard
2. Touch anywhere on the screen -> App crashes

**Issue 2**
1. Click on "Open TextInput" button
2. Click again on "Open TextInput" button. -> App crashes

I understand it should not be possible to open the keyboard twice but I'd expect it not to crash if called twice.

**Issue 3**
Keyboard doesn't open on Simulator (is this expected?)

**Issue 4 (Minor)**
Part of the test UI is displayed under the Insets (iPhone notch)

PR #7670, opened 2025-07-25
Native input API rewrite

Berstanio:
This is a cleanup PR, that fixes multiple minor issues and makes the API more robust/easier to work with using close callbacks.
The main improvements are in the Android backend, cause Androids IME is way harder to work with.
See CHANGES entry for a detailed list of changes.

I also changed the constructor of `DefaultAndroidInput` to take a View, cause thats the only parameter it ever takes. No idea why it was ever `Object`.
I also changed the orientation of the android tests to `sensorLandscape`, so that issues with rotating the device can be tested.

This addresses also Issue 1 and 2 of https://github.com/libgdx/libgdx/issues/7613.
Fixes https://github.com/libgdx/libgdx/issues/7599

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,5 @@
 [1.14.1]
 - [BREAKING CHANGE] iOS: Removed IOSAudio#willTerminate to be replaced by new Audio#dispose.
-[1.13.6]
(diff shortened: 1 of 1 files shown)


obigu (approved):
Seems like a good improvement. Tested using the NativeInputTest.

PR #7671, opened 2025-07-26
scene2d native input integration

Berstanio:
This PR is based on https://github.com/libgdx/libgdx/pull/7670.
The commit to review is https://github.com/libgdx/libgdx/commit/4f75a18e25b12426a0d9e63ff5c187feeafe1739.
I'm opening this as a draft for initial feedback.

This PR elevates the new API introduced in https://github.com/libgdx/libgdx/pull/7004 to make it easily accessible for use in scene2d.
This PR adds new fields to `TextField`, to make use of all features provided by the new API. Specifically `autocompleteOptions`, `keyboardType` and `preventAutoCorrection`.
The API supports placeholder, the `messageText` was used for that now. 
The `OnscreenKeyboard` method was splitted in two (close/show). This is not strictly necessary, but imo leads to a cleaner API.
The default keyboard used can be toggled globally via `TextField#DEFAULT_ONSCREEN_KEYBOARD`. Note, that `Input#setOnscreenKeyboardVisible` and `Input#openTextInputField` are mutually exclusive and should not be used together.

@NathanSweet fyi, because it is scene2d


NathanSweet:
TextField changes look good! 👍

PR #7673, opened 2025-07-27
Added glBlendFuncSeparate support in RenderContext

MORIMORI0317:
We have made changes to extend the RenderContext setBlending so that RGB and alpha factors can be specified individually.
In addition, the setBlending function, which specifies the same factor for RGB and Alpha as before, has been retained for compatibility.

This change overlaps with part of #7095, but since there has been no response for a long time, I created this PR.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderContext.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/utils/RenderContext.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright 2011 See AUTHORS file.
- *
+ * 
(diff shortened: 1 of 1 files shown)


MORIMORI0317:
Sorry, I thought there was no problem because the LWJGL3 tests ran successfully.
Fixed a format violation that occurred during build.

dar-dev:
Thanks for adding `glBlendFuncSeparate` support! 🙌  

I just wanted to point out that in my earlier PR (#7095) I had also added `glBlendEquationSeparate` support, as well as two bug fixes. These parts weren’t covered here and the bugs are still present in master.

PR #7674, opened 2025-07-27
Update StringBuilder to apache.commons.text.

NathanSweet:
libgdx's StringBuilder was based on Apache Harmony in 2012 (funny as Harmony was retired in 2011). It's been changed and added to and works OK, but could benefit from an update. This PR updates it to [TextStringBuilder](https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/TextStringBuilder.java) from apache.commons.text.

That class has been massaged a little to fit libgdx. It had some suboptimal ideas about methods silently accepting null and doing nothing. Tokenizer and StringMatcher were removed. We don't need customizing the string appended for null or newline.

Most importantly, methods were added so we don't lose anything. It should all work as ye olde StringBuilder did, with some risk that things are subtly different or, as ever, that there are bugs. There aren't tests. YOLO or add at least some basic tests before merging? FWIW, a couple apps using it have not had issues so far.

The benefits are nicer code, javadocs, @Null annotations (for clearer documentation), and a better API with more convenience. From the class javadoc:

> The main differences from StringBuffer/StringBuilder are:
> * Not final or synchronized
> * Direct access to the char[] and length
> * Appending int and long does not allocate
> * Compare against strings without allocating
> * Implements equals and hashCode
> * char[] grows by 50% rather than doubling
> * Additional builder-style methods:
>   * appendWithSeparators - adds an array of values, with a separator
>   * appendPadding - adds padding characters
>   * appendFixedLength - adds a fixed width field to the builder
>   * toCharArray/getChars - simpler ways to get a range of the character array
>   * delete/replace/replace(char,String) - delete or replace chars and strings
>   * leftString/rightString/midString - substring without exceptions
>   * contains/equals/equalsIgnoreCase/etc - string methods
>   * size/clear/isEmpty/notEmpty - collections style API methods
>   * codePointAt/codePointCount/reverse/etc - Unicode surrogate aware methods
>   * toStringAndClear/etc - general convenience
>   * Readable/Appendable/CharBuffer methods
> * Views:
>   * reader() - uses the internal buffer as the source of a Reader
>   * writer() - allows a Writer to write directly to the internal buffer

If we want this update, it may be a good time to consider renaming the class. The JDK class with the same name makes it difficult to tell which is being used and easy to accidentally use the wrong one. TextStringBuilder is too long. TextBuilder? TextBuffer? It's basically a CharArray, combine into that?

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/CharArray.java
+++ b/gdx/src/com/badlogic/gdx/utils/CharArray.java
@@ -2239,7 +2239,7 @@ public class CharArray implements CharSequence, Appendable {
 		public long skip (long n) {
 			if (pos + n > size) n = size - pos;
 			if (n < 0) return 0;
-			pos = Math.addExact(pos, Math.toIntExact(n));
+			pos += (int)n;
 			return n;
 		}
 	}


tommyettinger:
I like the idea in concept! The current build errors are all GWT stuff:
```
Error: ERROR] Errors in 'file:/home/runner/work/libgdx/libgdx/gdx/src/com/badlogic/gdx/utils/StringBuilder.java'
Error:   [ERROR] Line 365: The method format(String, Object[]) is undefined for the type String
Error:   [ERROR] Line 1607: The method compareUnsigned(int, int) is undefined for the type Integer
Error:   [ERROR] Line 1911: The method toUnsignedString(int, int) in the type Integer is not applicable for the arguments (int)
Error:   [ERROR] Line 1606: The method compareUnsigned(int, int) is undefined for the type Integer
```
You are far from the first, and the last, person to try to use String.format() on GWT!!! It is a strange omission from the API, though somewhat understandable on GWT's part because the implementation is apparently huge on the JDK side.  I believe any "unsigned" APIs were added in Java 8, and so won't work on RoboVM either.

The downsides to this are that if we don't need the features added here, then StringBuilder in the JDK benefits from a number of intrinsic methods that userspace code simply cannot compete with. You can see which intrinsics are present here: https://www.chriswhocodes.com/hotspot_intrinsics_openjdk7.html and in this screenshot:
vivaldi_CAaQtfH584
Changing the name would be a significant benefit because it could clarify that we are using the libGDX version because we benefit from its features, and not the JDK version which needs fewer features but is likely faster on the listed methods (on desktop JDKs, and likely also Android because it also has intrinsics).

NathanSweet:
Good point on intrinsics, noted in javadoc. Looks like usage in `gdx` can be removed. Some of it is probably accidental, especially in the same package.

Build is good now. Unless someone hates ramming it all into CharArray, I guess I'll do that.

tommyettinger:
> Build is good now. Unless someone hates ramming it all into CharArray, I guess I'll do that.

I actually really like the idea of it being a super-augmented version of CharArray; it should make CharArray much more useful and avoids splitting up the functionality into two incompatible classes that both are fundamentally sequences of `char`. It would be great if `CharArray` implements `CharSequence` going ahead, since, well, it is a char sequence anyway, so the small (I think) interface would make sense to implement.

NathanSweet:
OK, I jammed all that shit into CharArray and made it semi-nice. I did some basic poke-it-with-a-stick but wouldn't say I have tons of confidence in it, even though there are some tests.

StringBuilder is gone, classes using it changed to JDK. App code may need to change slightly, as CharArray has `size` while the old StringBuilder had `length`.

tommyettinger:
[We jamming!](https://www.youtube.com/watch?v=oqVy6eRXc7Q) I'll be jamming too in [jdkgdxds](https://github.com/tommyettinger/jdkgdxds/blob/master/src/main/java/com/github/tommyettinger/ds/CharList.java), implementing CharSequence for my CharList that's a fairly close copy of CharArray (with changes to use common interfaces). I'll try to follow your lead on CharArray getting features from StringBuilder; there should be a number of String-centric optimizations and features possible that aren't the case for things like an IntArray/IntList, or other primitive types.

gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -408,58 +481,1797 @@ public char[] toArray () {
...
+	 * @return true if the CharArrays contain the same characters in the same order */
+	public boolean equalsIgnoreCase (CharArray other) {
+		if (this == other) return true;
+		if (other == null) return false;
+		int length = this.size;
+		if (length != other.size) return false;
+		char[] chars = this.items, chars2 = other.items;
+		for (int i = 0; i < length; i++) {
+			char c = chars[i];
+			char upper = Character.toUpperCase(chars2[i]);
```
tommyettinger, line 2118: Just letting you know this method doesn't work on GWT except for ASCII letters! I have a few workarounds for this. RegExodus stores both [to-upper and to-lower versions](https://github.com/tommyettinger/RegExodus/blob/master/src/main/java/regexodus/Category.java#L775-L787) of the entire Unicode Basic Multilingual Plane, while jdkgdxds only stores [toUpper data](https://github.com/tommyettinger/jdkgdxds/blob/master/src/main/java/com/github/tommyettinger/ds/Casing.java). Only toUpper is used there because with the sole exceptions of the Georgian language and, the old classic exceptional case, Turkish with its dotless and dotted `i`, comparing the toUpper versions of two chars is case-insensitive. At least, for the invariant locale.

Neither approach is fantastic, really; it requires a 128KB array in memory for just to-upper-case data, unless you use a compressed data structure like RegExodus does for some other data... which is significantly slower at runtime to look up. Using multiple StringBuilder instances in Casing is a requirement because R8 and ProGuard both try to form one giant String from pure `"abc" + "def"` concatenations, and that giant String is too large for the JDK to permit in source code, leading to a strange crash during compilation.

gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -408,58 +481,1797 @@ public char[] toArray () {
...
+		}
+
+		public void reset () {
+			pos = mark;
+		}
+
+		public long skip (long n) {
+			if (pos + n > size) n = size - pos;
+			if (n < 0) return 0;
+			pos = Math.addExact(pos, Math.toIntExact(n));
```
tommyettinger: I'm pretty sure RoboVM won't have access to addExact or toIntExact because they were added in Java 8.

gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -408,58 +481,1797 @@ public char[] toArray () {
...
+		return -1;
+	}
+
+	public int indexOfIgnoreCase (String str, int start) {
+		if (start < 0) start = 0;
+		int length = str.length();
+		if (length == 0) return start < size || start == 0 ? start : size;
+		int maxIndex = size - length;
+		if (start > maxIndex) return -1;
+		char firstUpper = Character.toUpperCase(str.charAt(0));
```
tommyettinger, line 1500: This method is a great addition, but it also is limited to ASCII on GWT. 😢

gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -408,58 +481,1797 @@ public char[] toArray () {
...
+		return append(value, minLength, '0');
+	}
+
+	/** Appends the string representation of the specified {@code long} value. The {@code long} value is converted to chars without
+	 * memory allocation.
+	 * @param minLength the minimum number of characters to add
+	 * @param prefix the character to use as prefix */
+	public CharArray append (long value, int minLength, char prefix) {
+		if (value == Long.MIN_VALUE) {
+			append("-9223372036854775808");
```
tommyettinger, line 649: Good catch for the one long that stays negative both before and after applying unary `-`!

gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -408,58 +481,1797 @@ public char[] toArray () {
...
+		return append(value, minLength, '0');
+	}
+
+	/** Appends the string representation of the specified {@code int} value. The {@code int} value is converted to chars without
+	 * memory allocation.
+	 * @param minLength the minimum number of characters to add
+	 * @param prefix the character to use as prefix */
+	public CharArray append (int value, final int minLength, final char prefix) {
+		if (value == Integer.MIN_VALUE) {
+			append("-2147483648");
```
tommyettinger, line 605: This check might not work as expected on GWT, which implements the `int` type using a JavaScript `Number`, which is internally a double-precision IEEE float with extra behavior to convert to a 32-bit signed int when most bitwise operations are used on it, or a 32-bit unsigned int in the sole case of `(value >>> 0)`. Checking this is kind-of a pain, but you can easily make an int on GWT exceed its normal range with `int tooLarge = Integer.MAX_VALUE + 1;` (it won't overflow). If you do `value |= 0;`, it will be forced back to the normal world of 32-bit signed integers. JS devs have also been known to use `~~value` for this purpose.

gdx/src/com/badlogic/gdx/utils/CharArray.java
NathanSweet: Ehhh yikes. I guess such a fix would be in the GWT backend, for all Character.toUpperCase/etc usage.

gdx/src/com/badlogic/gdx/utils/CharArray.java
NathanSweet: Ah yeah I meant to nuke it. The file is too long, my eyes cross. It's already bounded by an int, no idea what they were thinking.

gdx/src/com/badlogic/gdx/utils/CharArray.java
NathanSweet: We can compare differently, but I'm not setup with GWT to fix and test it.

We also have `== Integer.MIN_VALUE` for scene2d.InputEvent `isTouchFocusCancel`.

NathanSweet:
Re: jdkgdxds, cool, glad to see things getting used. It being a drop-in replacement for StringBuilder makes sense, even with some gotchas. Eg the existing CharArray `reverse()` differs from SB, whose `reverse()` does CharArray's `reverseCodePoints()`. Might be a few other differences, but for the most part it works like SB.

The most likely reason to use CharArray is to append with a separator or for comparison without toString allocation. scene2d.ui.Label `setText` uses it for that, the only place in libgdx CharArray is used.

Probably the most compelling reason to merge this PR is to get rid of the accidental usage of the old StringBuilder. We can give it more time before merging for more input.

tommyettinger (approved):
With the "Exact" Java 8 APIs gone, my only serious problem is gone. Now we just need to figure out what the options are for GWT...

NathanSweet:
Character `toUpperCase` and friends for GWT seems separate from this PR. Ideally it's fixed in the backend, ie `backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/emu/java/lang/Character.java` or however that black magic works.

NathanSweet:
Would like to merge as it's used in 7675.

PR #7675, opened 2025-07-28
Added JsonMatcher, extracts values with pattern matching

NathanSweet:
This PR greatly improves JsonSkimmer (from #7410) by using JsonToken instead of allocating strings for every name and value. That allows code to check if it actually cares before creating strings, the goal being efficient JSON parsing particularly when you care about only a small portion. JsonToken also unescapes and decodes to JSON data types (when you want it), whereas before you got only strings and as they appeared in the data. Still no streaming though.

This PR also adds JsonMatcher, making it easier to pull values out of JSON than using JsonSkimmer directly. Patterns describe the values to extract, sort of like jq, JSONpath, JSONata, and hundreds of others, though this is kept simple (330 LOC). It adds minimal overhead to JsonSkimmer. One neat feature is parsing automatically stops as soon as all values are captured (depending on the pattern). Filtering is done in code, not the DSL. It has a gajillion tests.

From class javadoc:

> ### Pattern syntax
> 
> - `/` Path separator.
> - `*` Matches any single object or array.
> - `@` Matches any single object or array, calls process() after each object or array.
> - `**` Matches zero or more objects or arrays.
> - `@@` Matches zero or more objects or arrays, calls process() after any capture.
> - `[name]` Captures field "name" as a single value.
> - `[name[]]` Captures field "name" into an array.
> - `[name,id,title]` Captures multiple fields.
> - `[*]` Captures first object, array, or field.
> - `[*[]]` Captures all objects, arrays, or fields into an array.
> - `[@]` Captures each object, array, or field, calls process() after each.
> - `[@name]` Captures field "name", calls process() right away.
> 
> ### Examples
> 
> ```
> {users:[{name:nate},{name:iva}]}
> ```
> Process each user name: `users/@[name]`
> 1. The JSON root is matched implicitly.
> 2. users matches the `users:[` array. The field name and value are matched together.
> 3. `@` matches each object in the array. process() will be called at the end of each object.
> 4. `[name]` captures the name field and value.
> 5. Result: process() is called with `{name=nate}` and again with `{name=iva}`.
> ---
> ```
> {config:{port:8081}}
> ```
>
> Get the first port from config found at any depth: `**/config[port]`
> Result: process() is called with: `{port=8081}`
>
> --- 
>
> ```
> {services:[{status:up},{status:down},{status:failed}]}
> ```
> Get all service statuses in an array: `services/*[status[]]`
> Result: process() is called with: `{status=[up, down, failed]}`
>
> ---
> ```
> {items:[{id:123,type:cookies},{id:456,type:cake}]}
> ```
> Process each id and name: `items/@[id,type]`
> Result: process() is called with: `{id=123,type=cookies}` and `{id=456,type=cake}`
> 
> ### Arrays
> 
> Arrays are captured as-is:
> - `data[items]` with `{data:{items:[1,2,3]}}` gives: `{items=[1,2,3]}`
> 
> To collect multiple values into an array, capture with `[]` after the name:
> - `*[id]` with `{first:{id:1},second:{id:2}}` gives: `{id=1}` (parsing stops after first match)
> - `*[id[]]` gives: `{id=[1, 2]}` (all matches in an array)
> 
> ### Processing behavior
> 
> - Calling reject() prevents any further matching at this level or deeper, making for easy filtering.
> - Calling stop() prevents further matching and stops parsing.
> - Without `@` (in any form), `[*]`, or `[]` parsing stops once all specified values are captured.
> - With `@` process() is called after the object or array that `@` matched.
> - With multiple `@` process() is called for each capture, except for captures after the last `@`.
> - With `@@` process() is called for each capture.
> - With `[@]` process() is called for each object, array, or field.
> - When capturing a whole object or array, other patterns will not match inside it.
> - The key for unnamed values (eg array elements) is empty string ("").
> - Value types are true, false, Long, Double, Array, ObjectMap, or null.


gdx/src/com/badlogic/gdx/utils/JsonSkimmer.java
```diff
@@ -534,61 +542,158 @@ public boolean isStopped () {
...
+				if (length == 4) {
+					if (equals("true")) return Boolean.TRUE;
+					if (equals("null")) return null;
+				} else if (length == 5 && equals("false")) //
+					return Boolean.FALSE;
+				boolean couldBeDouble = false, couldBeLong = true;
+				char[] chars = this.chars;
+				for (int i = start, n = i + length; i < n; i++) {
+					switch (chars[i]) {
+					case '0':
```
tommyettinger: I'm not sure what number literal syntax variants libGDX JSON supports, but if 0xCAFEBEEF hex literals or 0x1.8p-20 hex floats are supported (this one is unlikely in JSON, but I think parseDouble() can actually handle it), those use more characters, and the `_` underscore separator inside numerical literals is also used not-infrequently in Java code, especially for large numbers. I don't know if JSON handles it, though I don't really think it's necessary to handle in most cases.

tommyettinger (approved):
This looks potentially very useful for JSON-heavy projects. I have my doubts people will actually bother to learn a regex-like syntax to handle JsonMatcher and will probably use heavier existing approaches more often than not, but this seems good to have available. Is the syntax similar to other existing JSON-matching DSLs? I figure familiarity with those is the main reason why people would decide to use this class. I'm approving because there's lots of tests that seem pretty much solid, and hopefully someone will find this useful!

NathanSweet:
I wasn't happy with some aspects so I did a big rewrite. It's pretty flippin' great now! See new class javadoc below. The basic match syntax is as simple as it gets. The value capture and processing are annotations to that, the rest are details.

I think it will be useful for a lot of JSON processing, not just the high performance subset extraction use case. Much of JSON processing is navigating the structure to dig out values and this does that nicely.

One feature it lacks is matching the root, which is not often needed but sometimes useful. I'll add that and let it stew a bit, see how using it my real apps goes.

None of the DSLs I've reviewed were appropriate to copy for this. They vary in complexity but all do too much IMHO. They start by putting filtering in the DSL, that leads to functions, and it explodes from there. [jq](https://github.com/jqlang/jq/wiki/Advanced-Topics) has a whole custom programming language with bytecode interpreter, module system, control flow with generators & backtracking, TCO, SQL syntax, etc. [JSONiq](https://en.wikipedia.org/wiki/JSONiq) and others are similarly bonkers. Others are simpler but not simple like I want, below. To be fair, I want simple and lightweight single pass Java API while other tools are intended or appropriate for other jobs, like CLI, transforms, etc.

Re: number formats, currently unquoted values that look like numbers are decoded as Long or Double, the rest are String. If other formats like `0xFF` or `1_000` would do that, it could make for surprise numbers where a string was expected. There's always potential for surprise when parsing JSON, but this may go too far. It'd make more sense if we didn't allow unquoted strings (which I love in my config files).

---

### Match
Match objects, arrays, or fields in the JSON:
- `/` Path separator.
- `name` Matches field "name".
- `a,b,c` Matches fields "a", "b", or "c".
- `*` Matches one object, array, or field.
- `**` Matches zero or more objects, arrays, or fields.

### Capture
Surround any match with parenthesis to capture that value in a map.
- `(name)` Captures field "name".
- `(a),b,(c,d)` Captures multiple fields.
- `(*)` or `(**)` Captures wildcard matches.

### Process
Use `@` after a match to control when processors receive captured values.
- `devices/(id,name)` Without `@` values are processed once at the end and only the first values matched are captured.
- `devices/(id@,name@)` Process "id" and "name" as soon as each is captured.
- `devices@/(id,name)` Process "id" and "name" as a pair for each device.
- `*@/(id,name)` Any match can be annotated to process.
- `**@/(id,name)` For `**@` processing happens as soon as each value is captured.

### Arrays
Arrays are captured as-is:
- `data/(items)` with `{data:{items:[1,2,3]}}` gives: `{items=[1,2,3]}`

Use `[]` in a capture to collect it into an array rather than as a single value.
- `*/(id)` with `{first:{id:1},second:{id:2}}` gives: `{id=1}` (parsing stops after first match)
- `*/(id[])` gives: `{id=[1, 2]}` (all matches collected in an array)

### Escaping
Use special characters `/,*@()[]',\` by surrounding them with single quotes.
- `email/'dog@example.com'` Uses "dog@example.com" as a literal string.
- `words/'can\'t'` Escape single quote with backslash.
- `files/'C:\\path'` Escape backslash with double backslash.

### Examples
`{users:[{name:nate},{name:iva}]}`  
Process each user name: `users@/(name)`
1. The JSON root is matched implicitly.
2. `users` matches the `users:[` array. The field name and value are matched together.
3. `@` processes captured values at the end of each array item.
4. `(name)` captures the "name" field and value.
5. Result: processors are called with `{name=nate}` and again with `{name=iva}`.

`{config:{port:8081}}`  
Get the first port from "config" found at any depth: `**/config/(port)`  
Result: processors are called with: `{port=8081}`

`{services:[{status:up},{status:down},{status:failed}]}`  
Get all service statuses in an array: `services/*/(status[])`  
Result: processors are called with: `{status=[up, down, failed]}`

`{items:[{id:123,type:cookies},{id:456,type:cake}]}`  
Process each "id" and "type" pair: `items@/(id,type)`  
Result: processors are called with: `{id=123,type=cookies}` and `{id=456,type=cake}`

### Notes
- reject() prevents further matching at this level or deeper, useful for filtering.
- clear() discards unprocessed captured values.
- end() prevents further matching and ends parsing. stop() does the same but also clears.
- Without `*`, `**`, or `[]` parsing ends once all specified values are captured.
- All captures after `**@` are processed right away.
- The key for unnamed values (eg array items) is empty string ("").
- Value types are true, false, Long, Double, Array, ObjectMap, or null.

tommyettinger:
The main things I notice are relative to regex syntax, there are some missing concepts. I'm still not entirely sure how this would be used or what it would end up being used for in a project, so I may be mixed up here! So far, I don't see any explicit way to mark a part of a match as optional (such as how in `foo,( ?)bar`, a single space is optional after the comma, and the space or empty string can be put in the match using `$1`). Being able to match a specific number of items in an array (minimum, maximum, and/or exact match) seems useful, and overlaps with the optional syntax (which is just minimum 0, maximum 1, pretty much). There is a way to match 0 or more, but not to limit the number of matches that can soak up unless reject() is somehow used.

To me, using `*` and `**` to mean "any single item" and then "0 or more items" seems a little strange, but I'm used to those being the regexes `.` and `.*`. Thinking about it, using `.` has a math-y origin, probably, and in plain English I would use `?` as a placeholder for an unknown value, like how it is used as a wildcard in Java generics. Particularly because there doesn't need to be a dedicated syntax for a single optional item if there exists syntax for min and max counts, using `?` might make sense here for an unknown or "don't care what it is" value. That's just a suggestion if you think it might help readability, especially in the case where two match-any items might follow each other without a separator (I don't know where this might happen).

As a suggestion for min and max item counts, the capture syntax `*/(option[0,1])` could gather at most one match for "option" in an array, or `*/(vector[2,4])` could gather 2, 3, or 4 items matching "vector" into an array, but not if only one item matches (since there's a Vector2, Vector3, and Vector4 here, but there isn't a Vector1 class!). An exact match for an array size would just have one number, like `*/(triple[3])` . That could either match the first three and discard later array items, or require exactly three items to match; I am not sure if both options would be beneficial to have.

I think these might be useful features, but they aren't necessary in the initial PR at all. I think merging this at any time when we know it passes tests on all platforms will be fine; it won't break anything already present, as far as I can tell, unless it prevents compilation somewhere.

NathanSweet:
`*` and `**` are common for globbing in Unix shells, .gitignore, etc. My [old globbing lib](https://github.com/EsotericSoftware/wildcard) is similar in that it only traverses folders that could possibly match.

Most often the JSON keys you care about are known and you want the values for particular keys. I think it's uncommon to filter on the keys themselves, eg with regex `abcd?` or `abc.+`. If you needed that, you could capture all the keys `object/(*)` and write code to process them. This would also bring in the values though. A way to capture only keys would be nice. Maybe empty name in a capture `object/()`, not sure.

Filtering on values is much more common, but that's not in the DSL either. It could be convenient for simple logic, but complicates syntax only to fall apart for any nontrivial filtering. Instead code to filter is like:

```java
JsonMatcher matcher = new JsonMatcher();
matcher.addPattern("*/(type@)", map -> {
	if (!map.get("type", "").equals("ENCHARGE")) matcher.rejectAll();
});
matcher.addPattern("*/(serialNumber[])", map -> {
	var serials = (Array)map.get("serialNumber");
	System.out.println(serials);
});
matcher.parse(json);
```

Anything beyond matching the structure/field names is code in the processor as a design choice. Compare to the usual approach: parse the whole JSON to JsonValue or maps/arrays, then dig through them to find the values you want. That's more and uglier code and less efficient. Automatic mapping of JSON values to objects is less ugly but less efficient and more brittle/magical. JsonMatcher is efficient and gets rid of a good amount of ugly, or at least that's the hope. I'll use it more in various real code and see how it goes.

Array indices are sort of structure, but I expect the use case is rare. How often do you care about a subset of items? Usually it's all, so you'd just capture and go. For anything else, you'd write code. Eg capture all, process or ignore some, reject when done.

tommyettinger:
I like the '' escape, isn't that also used by libGDX I18N code? It's good that it's coherent!

Yeah, the familiarity with glob syntax is good, and I don't think two "match any single item" could ever make sense next to each other without separators anyway.

Delegating to processors as a standard practice, but also having processors be simple enough to be a quick lambda, also looks good. I'm hoping you can get a version you're happy with into libGDX as an early draft, and revise it in the future as features seem useful to add. Maybe documenting this as an "incubating" API would make sense if some parts shouldn't be set in stone yet.

NathanSweet:
Went to use it and realized providing JsonValues would be sooo much better. What a rabbit hole 🐰 and had to fix all the tests, but it's pretty awesome now.

When there is one capture, you get the value directly (not wrapped in a map/object). Multiple captures give the root value in an object. This is based on the pattern, so how the root value is given doesn't change with different data.

It's nice that when you get eg a JsonValue representing a string, it has the `name` field (without needing to be wrapped), giving processing some context beyond which pattern matched.

Last thing it needs is to improve the O(n^2) append (`addChild`) to the JsonValue linked lists:

1. Could add a `tail` field. +4-8 bytes/JsonValue 😢 and there's already 4 fields just for structure. If not, could track the last child externally:
2. Identity map, JsonValue -> last child. The JSON is usually matched in chunks, so < ~100 values. Have to clear the map after each processing.
3. Track the last child for the object currently being collected with a stack, [like JsonReader](https://github.com/libgdx/libgdx/blob/master/gdx/res/com/badlogic/gdx/utils/JsonReader.rl#L351).
4. A half-assed solution that only tracks the last child for the top of the stack (`JsonMatcher.Pattern#stack`). Helps for objects/arrays of values, doesn't help when there are nested objects/arrays.

Any thoughts? 1 (`tail` field) or 3 (last child stack) seem OK to me.

The linked list pain comes up only when assembling JsonValues, done almost exclusively by parsers. Only once I used JsonValue to assemble a JSON document, because I needed to parse it, munge some values, and write it again.

There is also the inefficient JsonValue `setChild` for replace/add that is harder to improve. It happens in JsonMatcher but doesn't come up as often as appending: once per capture when there are > 1 captures. There's often only 1-4 captures, rarely > ~16. Append is more important (and easier to improve) and happens for every field/value for every captured object or array.

tommyettinger:
Am I reading JsonValue.addChild() wrong? https://github.com/libgdx/libgdx/blob/ed6955792360a1ae957594b9af908faee4cd7e23/gdx/src/com/badlogic/gdx/utils/JsonValue.java#L963-L981 It looks like it's at most O(n), and shortcuts often when it encounters a valid child. If the concern is calling `addChild()` more than once, maybe a dedicated `addChildren()` method could be made that constructs the linked list of children separately and splices it in? Is the quadratic time method somewhere else?

NathanSweet:
It never shortcuts, just forwards to the tail and appends. n^2 comes from calling `addChild` for each child.

It's true that `addChildren` would be more efficient, though we'd need to buffer until we have all the children. JSON has nested items so we'd need an `Array<JsonValue>` at each depth. Or it could be a single array if we also have an `IntArray` to track the start of each depth.

Option 3 above also needs a `Array<JsonValue>` but it only has to store 1 value for each depth (last child).

Option 1 doesn't need any external tracking, which is nice, but adds a `tail` field to every JsonValue. That's clean and memory is abundant. On the other hand, this is only needed during assembly of a JsonValue structure (usually during parsing, sometimes for parse > modify > write).

Another thought is build the linked list backward (append at head), then reverse it afterward. Reverse it for all objects and arrays in the whole graph is a lot though.

Go with `tail`?

NathanSweet:
I've never actually used the JsonValue `prev` field or `prev()`. It was added from the start just because [cJSON](https://github.com/DaveGamble/cJSON/blob/master/cJSON.h#L107) has it. Typically a value is reached with `next`, so `prev` isn't needed. Theoretically it could be useful in some situation, but I would guess no one uses it.

Think we should get rid of `prev`? Adding `tail` would put us back to 4 structural fields. `tail` is a breaking change, since `child`/`next` manipulation needs to also change `tail`, so other breaking changes at the same time could make sense. Removing `prev` would certainly draw attention to code that needs to set `tail`. EDIT: Ah `remove()` would need to walk all the children. It currently doesn't because we have `prev`. While seldom used, that might be compelling enough to keep it. If you iterate to find the one to remove, it's a little surprising that removal is O(n) (and a regression to change from O(1)). `remove(int)` and `remove(String)` being O(n) is less surprising.

NathanSweet:
I think it's pretty good now. Real code processing HTTP responses from a photovoltaic system looks like:
```
controllerParser = new JsonMatcher();
controllerParser.addPattern("*/(type@)", value -> {
	if (value.asString().equals("ENPOWER"))
		controllerParser.reject(1);
	else if (value.asString().equals("ENCHARGE"))
		controllerParser.reject(2);
	else
		throw error("Unexpected type: " + value);
});
controllerParser.addPattern("*/devices/*@/(serial_num,last_rpt_date,communicating,percentFull,temperature,maxCellTemp)",
	value -> {
		// Battery data (ENCHARGE).
		BatteryDevice device = BatteryDevice.devices.get(value.getString("serial_num"));
		if (!value.getBoolean("communicating"))
			device.lastReportTime = 1; // Status.unreachable
		else
			device.lastReportTime = value.getLong("last_rpt_date") * 1000;
		device.status();
		device.level = value.getInt("percentFull") / 100f;
		device.overallC = value.getInt("temperature");
		device.maxCellC = value.getInt("maxCellTemp");
	});
controllerParser.addPattern("*/devices/*@/(last_rpt_date,communicating,temperature,mains_oper_state)", value -> {
	// Controller data (ENPOWER).
	if (!value.getBoolean("communicating"))
		lastReportTime = 1; // Status.unreachable
	else
		lastReportTime = value.getLong("last_rpt_date") * 1000;
	controllerC = value.getInt("temperature");
	grid = value.getString("mains_oper_state").equals("closed");
});

batteryParser = new JsonMatcher();
batteryParser.addPattern("devices/*@/(serial_num,real_power_mw)", value -> {
	// Battery watts.
	BatteryDevice device = BatteryDevice.devices.get(value.getString("serial_num"));
	device.watts = value.getInt("real_power_mw") / 1000; // Doesn't seem to report < 1W.
});

solarPanelParser = new JsonMatcher();
solarPanelParser.addPattern("*@/(serialNumber,lastReportDate,lastReportWatts)", value -> {
	// Solar panel data.
	SolarPanelDevice device = SolarPanelDevice.devices.get(value.getString("serialNumber"));
	device.lastReportTime = value.getLong("lastReportDate") * 1000;
	device.watts = value.getInt("lastReportWatts");
});

meterParser = new JsonMatcher();
meterParser.addPattern("*/(eid@)", value -> {
	if (value.asString().equals(solarProductionMeter))
		meterParser.reject(2);
	else if (value.asString().equals(netConsumptionMeter))
		meterParser.reject(1);
	else
		meterParser.rejectAll();
});
meterParser.addPattern("*@/(activePower,actEnergyDlvd)", value -> {
	// Solar production meter.
	solarWatts = value.getFloat("activePower");
	producedTotalWh = value.getDouble("actEnergyDlvd");
});
meterParser.addPattern("*@/(activePower,actEnergyDlvd,actEnergyRcvd)", value -> {
	// Net consumption meter.
	gridWatts = value.getFloat("activePower");
	importedTotalWh = value.getDouble("actEnergyDlvd");
	exportedTotalWh = value.getDouble("actEnergyRcvd");
});
```

PR #7678, opened 2025-08-01
Nuke Pools class

Berstanio:
As discussed at length in various places, we have multiple problems with the current `Pools` class:
https://github.com/libgdx/libgdx/pull/7666
https://github.com/libgdx/libgdx/pull/7648
https://github.com/libgdx/libgdx/pull/7474

The main issue is, that we can't make it play nice with static analysis tools.

This PR deprecates the `Pools` class and schedules it for removal.
The PR introduces a new, smaller class `PoolManager`, that can be used to manage multiple pools together.

### DIFF
--- a/gdx/src/com/badlogic/gdx/net/HttpRequestBuilder.java
+++ b/gdx/src/com/badlogic/gdx/net/HttpRequestBuilder.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import com.badlogic.gdx.Net.HttpRequest;
 import com.badlogic.gdx.utils.Base64Coder;
 import com.badlogic.gdx.utils.Json;
-import com.badlogic.gdx.utils.Pools;
 
 /** A builder for {@link HttpRequest}s.
  * 
@@ -50,7 +49,7 @@ public class HttpRequestBuilder {
 			throw new IllegalStateException("A new request has already been started. Call HttpRequestBuilder.build() first.");
 		}
 
-		httpRequest = Pools.obtain(HttpRequest.class);
+		httpRequest = new HttpRequest();
 		httpRequest.setTimeOut(defaultTimeout);
 		return this;
 	}
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
@@ -21,19 +21,21 @@ import static com.badlogic.gdx.utils.Align.*;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.g2d.Batch;
+import com.badlogic.gdx.graphics.g2d.GlyphLayout;
 import com.badlogic.gdx.graphics.glutils.ShapeRenderer;
 import com.badlogic.gdx.graphics.glutils.ShapeRenderer.ShapeType;
 import com.badlogic.gdx.math.MathUtils;
 import com.badlogic.gdx.math.Rectangle;
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.scenes.scene2d.utils.ActorGestureListener;
+import com.badlogic.gdx.scenes.scene2d.utils.ChangeListener.ChangeEvent;
 import com.badlogic.gdx.scenes.scene2d.utils.ClickListener;
 import com.badlogic.gdx.scenes.scene2d.utils.ScissorStack;
 import com.badlogic.gdx.utils.Align;
 import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.DelayedRemovalArray;
 import com.badlogic.gdx.utils.Null;
-import com.badlogic.gdx.utils.Pools;
+import com.badlogic.gdx.utils.PoolManager;
 import com.badlogic.gdx.utils.reflect.ClassReflection;
 
 /** 2D scene graph node. An actor has a position, rectangular size, origin, scale, rotation, Z index, and color. The position
@@ -71,6 +73,8 @@ public class Actor {
 	float rotation;
 	final Color color = new Color(1, 1, 1, 1);
 	private @Null Object userObject;
+
+	protected PoolManager poolManager = new PoolManager(Rectangle::new, Array::new, GlyphLayout::new, ChangeEvent::new);
 
 	/** Draws the actor. The batch is configured to draw in the parent's coordinate system.
 	 * {@link Batch#draw(com.badlogic.gdx.graphics.g2d.TextureRegion, float, float, float, float, float, float, float, float, float)
@@ -128,7 +132,7 @@ public class Actor {
 		event.setTarget(this);
 
 		// Collect ascendants so event propagation is unaffected by hierarchy changes.
-		Array<Group> ascendants = Pools.obtain(Array.class);
+		Array<Group> ascendants = poolManager.obtain(Array.class);
 		Group parent = this.parent;
 		while (parent != null) {
 			ascendants.add(parent);
@@ -162,7 +166,7 @@ public class Actor {
 			return event.isCancelled();
 		} finally {
 			ascendants.clear();
-			Pools.free(ascendants);
+			poolManager.free(ascendants);
 		}
 	}
 
@@ -834,16 +838,16 @@ public class Actor {
 		tableBounds.y = y;
 		tableBounds.width = width;
 		tableBounds.height = height;
-		Rectangle scissorBounds = Pools.obtain(Rectangle.class);
+		Rectangle scissorBounds = poolManager.obtain(Rectangle.class);
 		stage.calculateScissors(tableBounds, scissorBounds);
 		if (ScissorStack.pushScissors(scissorBounds)) return true;
-		Pools.free(scissorBounds);
+		poolManager.free(scissorBounds);
 		return false;
 	}
 
 	/** Ends clipping begun by {@link #clipBegin(float, float, float, float)}. */
 	public void clipEnd () {
-		Pools.free(ScissorStack.popScissors());
+		poolManager.free(ScissorStack.popScissors());
 	}
 
 	/** Transforms the specified point in screen coordinates to the actor's local coordinate system.
@@ -989,5 +993,9 @@ public class Actor {
 			if (dotIndex != -1) name = name.substring(dotIndex + 1);
 		}
 		return name;
+	}
+
+	public PoolManager getPoolManager () {
+		return poolManager;
 	}
 }
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
@@ -42,7 +42,7 @@ import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.Disposable;
 import com.badlogic.gdx.utils.Null;
 import com.badlogic.gdx.utils.Pool.Poolable;
-import com.badlogic.gdx.utils.Pools;
+import com.badlogic.gdx.utils.PoolManager;
 import com.badlogic.gdx.utils.Scaling;
 import com.badlogic.gdx.utils.SnapshotArray;
 import com.badlogic.gdx.utils.viewport.ScalingViewport;
@@ -67,6 +67,7 @@ public class Stage extends InputAdapter implements Disposable {
 	/** True if any actor has ever had debug enabled. */
 	static boolean debug;
 
+	protected PoolManager poolManager = new PoolManager(InputEvent::new, FocusEvent::new, TouchFocus::new);
 	private Viewport viewport;
 	private final Batch batch;
 	private boolean ownsBatch;
@@ -219,7 +220,7 @@ public class Stage extends InputAdapter implements Disposable {
 
 		// Exit overLast.
 		if (overLast != null) {
-			InputEvent event = Pools.obtain(InputEvent.class);
+			InputEvent event = poolManager.obtain(InputEvent.class);
 			event.setType(InputEvent.Type.exit);
 			event.setStage(this);
 			event.setStageX(tempCoords.x);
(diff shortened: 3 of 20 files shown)


gdx/src/com/badlogic/gdx/utils/PoolManager.java
```diff
@@ -0,0 +1,92 @@
...
+		for (Pool<?> pool : pools) {
+			addPool(pool);
+		}
+	}
+
+	/** Registers a new pool with the given supplier. Will throw an exception, if a pool for the same class is already registered.
+	 * This can be used like `PoolManager#addPoll(MyClass::new);` */
+	public <T> void addPool (PoolSupplier<T> poolSupplier) {
+		Class<T> clazz = (Class<T>)poolSupplier.get().getClass();
+		if (typePools.containsKey(clazz)) {
```
NathanSweet: `put` returns the old key, could throw if not null. That changes the behavior since the PoolSupplier is stored by that time, but it's probably not a recoverable error.

gdx/src/com/badlogic/gdx/net/HttpRequestBuilder.java
```diff
@@ -50,7 +49,7 @@ public HttpRequestBuilder newRequest () {
 			throw new IllegalStateException("A new request has already been started. Call HttpRequestBuilder.build() first.");
 		}
 
-		httpRequest = Pools.obtain(HttpRequest.class);
+		httpRequest = new HttpRequest();
```
NathanSweet, line 52: Giving up here? ;)

gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
```diff
@@ -55,6 +57,9 @@
  * @author mzechner
  * @author Nathan Sweet */
 public class Actor {
+
+	static public PoolManager POOL_MANAGER = new PoolManager(Rectangle::new, Array::new, GlyphLayout::new, ChangeEvent::new);
```
NathanSweet: Can we name this `POOLS`? 🙏

gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java
```diff
@@ -67,6 +67,7 @@ public class Stage extends InputAdapter implements Disposable {
 	/** True if any actor has ever had debug enabled. */
 	static boolean debug;
 
+	protected PoolManager poolManager = new PoolManager(InputEvent::new, FocusEvent::new, TouchFocus::new);
```
NathanSweet: I'd also like to name this `pools`.

gdx/src/com/badlogic/gdx/scenes/scene2d/actions/Actions.java
```diff
@@ -22,17 +22,54 @@
...
 import com.badlogic.gdx.utils.Null;
 import com.badlogic.gdx.utils.Pool;
-import com.badlogic.gdx.utils.Pools;
+import com.badlogic.gdx.utils.PoolManager;
 
 /** Static convenience methods for using pooled actions, intended for static import.
  * @author Nathan Sweet */
 public class Actions {
 
+	public static final PoolManager ACTION_POOL_MANAGER = new PoolManager();
```
NathanSweet: This one is public and sometimes use with `import static` so `ACTION_` makes sense, but can it be `ACTION_POOLS`?

Generic suffixes like "controller" or "manager" don't add much IMHO. `PoolManager` makes sense for the class name and differentiates it from other classes, but fields of type `PoolManager` don't necessarily need "manager" in their names.

gdx/src/com/badlogic/gdx/net/HttpRequestBuilder.java
Berstanio: I totally forgot to mention it in the PR description, but I'm not sure whether pooling makes sense here?
The object is never freed, the object doesn't have a clear lifecycle and the allocation seems not really happen often?
It also makes the code not thread safe, without it needing to.

Berstanio:
> In this PR, within libgdx, Actor and BitmapFontCache have separate GlyphLayout pools. Would it make sense for GlyphLayout to expose a static pool? Should that be taken further to have each type pooled within libgdx expose a static pool? App code could use those rather than manage separate pools.

I think in this specific case, it is fine to keep those seperated. `BitmapFontCache` benefits of having the flushable pool.

But in general, I don't like having pooling instances based on pooled `Class`, and not where it is used. It ties every pool interaction for that specific class to one thread. I know, that libGDX is not thread safe, but having "Class" Pools creates an unexpected interaction between otherwise fully unrelated code.

gdx/src/com/badlogic/gdx/net/HttpRequestBuilder.java
NathanSweet: True, seems OK. Certainly better than the memory leak from never freeing! If no one ever noticed that, it probably doesn't need pooling.

NathanSweet:
FlushablePool vs an Array to store the objects to free (as before this PR) is pretty much the same.

(FlushablePool is a hilarious class name 🚽😄)

We're already tied to one thread and aaaaall kinds of things break if other threads are used without care, so while I get what you are saying, it doesn't bother me too much that otherwise unrelated code shares a pool. Shared or not you avoid allocation, which is the main goal. Separate pools have isolation but scale memory usage by O(n).

Berstanio:
> FlushablePool vs an Array to store the objects to free (as before this PR) is pretty much the same.

Fair, yes!

> We're already tied to one thread and aaaaall kinds of things break if other threads are used without care, so while I get what you are saying, it doesn't bother me too much that otherwise unrelated code shares a pool.

Yeah, my point is, "used without care" becomes more difficult with shared pools. I actually got bitten by a shared pool once with threads, which lead to a very rare issue and evil to debug, because the pool usage was deep in the callstack. Thats why I'm a bit reluctant against this I think xD

Regarding, the memory complexity, you are right that it is linear. But, I think in the usual case, every pool has < 5 objects and we have < 3 pools per type, so memory pressure is neglectable here I think.

NathanSweet:
Debug trauma is real! I can appreciate shared pools causing surprise threading problems, though the lesson is probably more about careful multithreading than avoiding shared pools. You can still get bit in myriad ways. Thread checks are the only real protection.

I'm OK with not sharing pools if that's what you guys want. We still avoid allocation and it's unlikely to be enough memory to matter much. I personally write desktop software anyway!

For a data point, I ran Spine, clicked around doing various actions for a minute, then dumped `Pools.typePools`:
```
for (Iterator<Entry<Class<?>, Pool<?>>> iter = typePools.entries().iterator(); iter.hasNext();) {
	Entry<Class<?>, Pool<?>> entry = iter.next();
	Pool pool = entry.value;
	if (pool.peak > 0) System.out.println(entry.key.getSimpleName() + ": " + pool.peak);
}
```
```
PaneEvent: 1
AlphaAction: 6
MoveByAction: 5
DelayAction: 3
FocusEvent: 1
GlyphRun: 85
RemoveActorAction: 2
ChangeEvent: 3
TouchFocus: 7
InputEvent: 2
SizeToAction: 100
Rectangle: 4
Array: 3
ScaleToAction: 12
ParallelAction: 100
GlyphLayout: 99
RepeatAction: 1
RemoveListenerAction: 1
VisibleAction: 1
MoveToAction: 100
SequenceAction: 6
```
EDIT: The 100s looked suspicious so I dug around and found a bunch of things queuing up actions when only really one should. Did it again after fixing that:
```
FocusEvent: 1
RepeatAction: 1
ScaleToAction: 12
SequenceAction: 7
TouchFocus: 7
AlphaAction: 8
VisibleAction: 1
PaneEvent: 1
RemoveActorAction: 5
MoveByAction: 5
ChangeEvent: 3
RemoveListenerAction: 2
Rectangle: 4
Array: 3
InputEvent: 2
GlyphLayout: 78
ParallelAction: 6
MoveToAction: 2
DelayAction: 4
GlyphRun: 43
ColorAction: 2
SizeToAction: 2
```
GlyphLayout and GlyphRun are pooled pretty heavily.

gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
```diff
@@ -55,6 +57,9 @@
  * @author mzechner
  * @author Nathan Sweet */
 public class Actor {
+
+	static public PoolManager POOLS = new PoolManager(Rectangle::new, Array::new, GlyphLayout::new, ChangeEvent::new);
```
tommyettinger, line 61: This POOLS is `static public`, there's a non-static `protected` pools in Stage, Actions has `public static final` ACTION_POOLS... what's the reason for all of these being so different? I've seen lots of issues pop up from people using static pools in text-related code before, and some of our thorniest and hardest-to-evaluate bugs in the libGDX issue tracker seem to be related to unexpected threading breaking GlyphLayout or GlyphRun pooling in particular.

NathanSweet: I think the idea is to compartmentalize the pools rather than have them centralized. For me, central makes sense (see above), whether static or another way of centralizing them.

Actions is holder for static methods, not really anywhere else to put it.

There is often a single Stage in an app, though sometimes two can make sense, in which case not sharing the pools is a little suboptimal. They'd be better static. 😁

If the types of pools on Actor are only pooled by Actor, it seems OK to keep them there, though GlyphLayout at least is used elsewhere.

Using libgdx in a thread unsafe way is going to cause pain, if not in pools then in other ways. Undefined behavior can give you infinite loops and other catastrophes. Non-static pools or not pooling at all can't save us from that, unless pooling happens to be the last thing preventing something from being thread safe. Most commonly that's not the case, eg the most common thing is probably setting Label text from the wrong thread.

If we want something to make wrong thread bugs less painful we could have thread checks. Turned off with a final field they are compiled away. It's easy to turn on when running from sauce, for everyone else we could provide debug and non-debug libgdx JARs.

gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
```diff
@@ -128,7 +133,7 @@ public boolean fire (Event event) {
 		event.setTarget(this);
 
 		// Collect ascendants so event propagation is unaffected by hierarchy changes.
-		Array<Group> ascendants = Pools.obtain(Array.class);
+		Array<Group> ascendants = POOLS.obtain(Array.class);
```
JsonMack, line 136: Could/Should rename to "pools" to let people know that the underlying implementation isn't immutable.

Berstanio: I'm sorry, I don't quite get what you mean.
Could you elaborate a bit more?

gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
Berstanio: > If we want something to make wrong thread bugs less painful we could have thread checks. Turned off with a final field they are compiled away. It's easy to turn on when running from sauce, for everyone else we could provide debug and non-debug libgdx JARs.

I think we can make it final, but have it dependent on some property? _Should_ be fine performance wise I think, even tho not compiled away immidiatly by javac?

gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java
JsonMack: No need to apologize, it's a good question. I'm sure you can find a better standard doc somewhere, but this one below is good. It's a small gripe, and not really a big deal, it's just a good practice to follow. 

When you see something static and final and it's lowercase, it's a pretty good indicator that it's **mutable**. Alternatively, if something is all uppercase and both static-final, it's a good indicator it's **immutable**.

> Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects. Examples include primitives, strings, immutable value classes, and anything set to null. If any of the instance's observable state can change, it is not a constant. Merely intending to never mutate the object is not enough. 

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

tommyettinger: Already libGDX violates this in [quite](https://github.com/libgdx/libgdx/blob/2aeb6c0fc0d190191289ac71a38c386305eabb4a/gdx/src/com/badlogic/gdx/graphics/VertexAttributes.java#L34-L42) [a](https://github.com/libgdx/libgdx/blob/5368811323eb4ac9c9167a73fc6910e42fcb425d/gdx/src/com/badlogic/gdx/math/Vector2.java#L29-L31) [few](https://github.com/libgdx/libgdx/blob/79cf00af53b7f38667291fbacf544d3074a811bd/gdx/src/com/badlogic/gdx/utils/Scaling.java#L32) [places](https://github.com/libgdx/libgdx/blob/d95a70ce23bc1056619c94b6497a1221cc18b103/gdx/src/com/badlogic/gdx/graphics/Color.java#L27-L31), so it isn't really a rule followed in this codebase. We aren't Google and that's IMO a good thing.

JsonMack: Haha well said, that's a fair point.

Berstanio: Ahh, now I understand, thank you!
I usually do upper-case for static variables, to indicate they are static and shared between objects. The field is not final, so that users can potentially swap out the implementation with another one.
Even if the `POOLS` field changes mid-execution multiple times, it shouldn't be of any real issue for anything.

NathanSweet: Interesting, I've never heard of static final case denoting immutability. Common practice is all static final fields get all caps (enum values too), like C/C++. Personally I find the shouting ugly and usually use camel case.

Tom-Ski (dismissed):

tommyettinger:
I fixed the conflict in CHANGES... and added a bunch of extra conflicts because CHANGES was an absolute mess. There was a duplicate line in 1.13.5's breaking changes (duplicate except for a typo at the end), grammatical issues were all over, and the whole set of recent entries didn't do anything in a standard way. In particular, the change for this PR was marked as a breaking change, but it's actually an API Deprecation if I'm not mistaken, which is a different label that was used a bunch in 1.13.5 .

That could have been an extra PR, but this couldn't be merged with a CHANGES conflict anyway, and if we're planning a new release soon, a little tidying up is just good for presentation. I understand that CHANGES is an abbreviated format, but we point people to CHANGES every time there's a release, and I think having it be easily understood is important.

@Berstanio Feel free to revert my commit editing CHANGES if you want, I just want to see this merged.

PR #7679, opened 2025-08-02
Multisample FBO to support OpenGL ES 3.0

WinterAlexander:
For some reason, the multi-sample FBO functionality was limiting itself to only be available with GL ES 3.1, even though the only features it needs to use are available in GL30 (GL ES 3.0). This limits the availability of the feature unnecessarily, and so we can make it use GL ES 3.0 for convenience and compatibility with a higher range of devices. 

Thank you to @Tom-Ski for the help on discord, see channel #libgdx-specific

Additionally, in a second commit (which can be removed if you disagree), I expose publicly fields which are currently very annoying to access when building a subclass to FrameBuffer. See my workaround below as a prime example of why these needs to be made available. If public fields are an issue, we can change to getters.

cc @mgsx-dev 

If for some reason this hasn't been merged yet and you want a drag drop workaround for your project, here is one https://gist.github.com/WinterAlexander/9d321557bb7a98a34f3e78352ab8ed45 (using ReflectionUtil from https://github.com/WinterAlexander/gdx-utils/blob/master/src/main/java/com/winteralexander/gdx/utils/ReflectionUtil.java)

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/GLFrameBuffer.java
@@ -446,7 +446,7 @@ public abstract class GLFrameBuffer<T extends GLTexture> implements Disposable {
 		Gdx.gl.glBindFramebuffer(GL30.GL_READ_FRAMEBUFFER, framebufferHandle);
 		Gdx.gl.glBindFramebuffer(GL30.GL_DRAW_FRAMEBUFFER, destination.framebufferHandle);
 
-		if ((copyBits & GL20.GL_COLOR_BUFFER_BIT) == GL20.GL_COLOR_BUFFER_BIT){
+		if ((copyBits & GL20.GL_COLOR_BUFFER_BIT) == GL20.GL_COLOR_BUFFER_BIT) {
 			int totalColorAttachments = 0;
 			for (FrameBufferTextureAttachmentSpec textureAttachmentSpec : destination.bufferBuilder.textureAttachmentSpecs) {
 				if (textureAttachmentSpec.isColorTexture()) {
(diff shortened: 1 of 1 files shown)


Tom-Ski (approved):

Tom-Ski:
Looks good to me, only tested android and desktop so far.
Will wait for @mgsx-dev to weigh in and see if there was some reason for 3.1 in first place.

mgsx-dev:
TBH i don't remember why i thought it was a GLES 3.1 only feature, maybe i made confusion between textureStorageMultisample (GLES 3.1) and renderBufferStorageMultisample (GLES 3.0). Sorry about that.

Minor change request, you could also update these 2 related tests :
https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/gles31/GL31FrameBufferMultisampleTest.java#L32
https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/gles31/GL31FrameBufferMultisampleMRTTest.java#L35

* moving them to gles3 package
* renaming them with GL30 prefix
* and make them available for tests : `@GdxTestConfig(requireGL30 = true)`

Thank you for your contribution :+1:

WinterAlexander:
Thank you for the feedback @mgsx-dev! Changes applied and also updated the changelog.

Tom-Ski:
Just need to test on all platforms then!

WinterAlexander:
@Tom-Ski Is there easy steps to produce runnable tests for HTML, Android? I could test those 2 myself, as well as desktop Linux for LWJGL 2 and 3

Tom-Ski:
https://libgdx.com/dev/from-source Guide on running is here

WinterAlexander:
Desktop LWJGL 3
image

Desktop LWJGL 2
image
image


GWT
image
image

Perhaps the MRT test is broken on GWT? I had to add both to `GwtTestWrapper` as they were not already there

Android
Both tests crashes on my phone, but also a lot other tests crash on my Phone. I have a Pixel 6a. A GL32 works though, so my phone should support GL ES 3.0. Not sure if there's a way to get the logs?

Tom-Ski:
@WinterAlexander Pushed a change here for webgl to work.

For android/ios, are you enabling gl3 in the launchers? GdxTestActivity needs to have it enabled in the config.
But also could be shader failing to compile on those devices, which should be ok now.

WinterAlexander:
Thank you @Tom-Ski, I can confirm your fix fixed GWT for me and Android works after setting useGL30 = true in GdxTestActivity

PR #7683, opened 2025-08-03
JsonValue#last added for O(1) append.

NathanSweet:
Currently JsonReader keeps track of the last child at every depth because append is O(n). Instead of all code assembling a JsonValue graph needing to work around that, it's nicer to handle it inside the API -- at the cost of another field. Some discussion [here](https://github.com/libgdx/libgdx/pull/7675#issuecomment-3146609660).

Also JsonValue `get` is no longer case insensitive (use getIgnoreCase()). It was odd as JSON is case sensitive and JsonValue uses `equals` elsewhere.
```
json.addChild("key", value);
json.get("KEY"); // returns key, wut
json.addChild("KEY", value); // doesn't overwrite key
json.get("key"); // random what you get
```

Tests pass. Not sure why UBJsonTest isn't in the test chooser (GWT?) but I ran it manually (should really be a unit test, JsonTest too).


tommyettinger (dismissed):
This looks good to me, though it would be nice if we could benchmark this. I imagine avoiding case-insensitive comparisons will be faster in general, not to mention more correct! Using more memory to improve speed always is a finicky tradeoff, but it seems worthwhile here. It would be nice to know how much more speed we get... Although a benchmark would ideally test the extra `last` field in isolation, with both using case-sensitive get() if that's used at all.

tommyettinger:
Oh, and UBJSON didn't always work on GWT in 1.13.1, but it should now. I believe the issue was with `long` in UBJSON, but may have affected other types as well. It's been a while since my PR was accepted, it changed GWT's DataInputStream implementation: https://github.com/libgdx/libgdx/pull/7590

NathanSweet:
Creating our own benchmark is a bit of a pain to get right. The best way would probably be to find some existing Java JSON benchmark project and add libgdx to it. This PR should be slightly better than JsonReader's last child stack, but nice that isn't needed anymore.

It'd be a good separate PR to enable tests for GWT if possible and move UI tests to unit tests where possible. It's easy to run all unit tests, while UI tests rarely get run.

tommyettinger:
I found a benchmark! https://github.com/tommyettinger/java-json-benchmark/tree/add-libgdx
So far I am testing against libGDX 1.13.1 and I can try to test against this PR if it gets merged and built in a snapshot, or if I build it locally.

```
GUIDE:
Library/Mode type
                                   Throughput mode
                                          Ran 20 times
                                              Score, higher is better
                                                             Error, affects confidence in score
                                                                         Operations per second
```

```
Benchmark                           Mode  Cnt        Score        Error  Units
Serialization.dsljson              thrpt   20  4994667.080 ±  26894.738  ops/s
Serialization.dsljson_reflection   thrpt   20  2690022.345 ± 110263.410  ops/s
Serialization.fastjson             thrpt   20  7053188.116 ±  46371.992  ops/s
Serialization.fastjson_features    thrpt   20  6340939.926 ± 486899.585  ops/s
Serialization.gson                 thrpt   20  1766968.262 ±  74281.894  ops/s
Serialization.jackson              thrpt   20  3118974.122 ±  64659.052  ops/s
Serialization.jackson_afterburner  thrpt   20  3380996.347 ±  35456.985  ops/s
Serialization.jackson_blackbird    thrpt   20  3137537.719 ±  96531.010  ops/s
Serialization.libgdx               thrpt   20   511014.613 ±   1938.024  ops/s
```

I can see if java-json-benchmark has any pure-Java no-Unsafe libraries I can test against, but against these big names... things don't look great.

NathanSweet:
Cool, I figured that should exist. It'll be interesting to see how libgdx compares, especially to itself (even though JSON is the wrong format for high performance).

Your branch tests `new Json()` for their `databind` benchmark. The Json class was created solely for convenience and has lots of reflection on top of JsonReader. I'm not super interested in putting in effort to make that faster, unless there is low hanging fruit.

More interesting would be only JsonReader (full DOM) and JsonSkimmer/JsonMatcher (selective DOM). Both of these would be basically testing the Ragel generated parser + JsonValue assembly (would show the difference with the new `last` field).

The benefits of JsonSkimmer won't be apparent. We'd have to "cheat" the test by parsing but not creating objects for some of the data. That would represent what JsonSkimmer is for and why it exists, but would be unfair to the other benchmarks who create all the objects. That's common in benchmarks, eg lazy deserialization. Charts alone are misleading and really need an accompanying explanation of important differences. Charts also don't capture how nice/ugly an API is, but at least that's easy to see looking at the benchmark code.

Damn, now I'm curious. Down the rabbit hole we goooooooooooooo 🕳️

Their `Serialization` methods all return Object: some return String, some ByteArrayOutputStream (why bytes for a character-based format). It's not good that the benchmarks are doing different work. Even returning a stream is a big advantage! To use the data you need to copy the bytes using `toString` or `toByteArray()`, but that is not captured in the benchmark. Meanwhile those that return String have paid the costs. Your Json test does objects > String > bytes which is doubly unfair. Some tests get byte[] input for free while Json has to copy String > char[]. A new Json is created each time while others avoid reallocating parsers with ThreadLocal.

I think this makes sense:
* Databinding: `Json`
* Streaming deserialization: `JsonReader`
* Streaming serialization: `JsonWriter`, `JsonString`, `JsonValue`

The project doesn't support such asymmetry but I hacked it in:
https://github.com/NathanSweet/java-json-benchmark/tree/add-libgdx
I also changed serialization to String for all benchmarks, using the various APIs to serialize to String (didn't just add a bytes > string step). Except for dsljson, logansquare, and moshi who have APIs so bad I couldn't easily figure out how to serialize to a String (they get what they deserve).

```
ser --apis stream --libs jackson,gson,fastjson,libgdx
Benchmark                         Mode  Cnt        Score         Error  Units
Serialization.gson               thrpt   20  4140162.229 ± 1289405.740  ops/s
Serialization.jackson            thrpt   20  9596064.326 ±  106769.855  ops/s
Serialization.libgdx_JsonString  thrpt   20  1856878.098 ±   26094.107  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1161879.610 ±    2765.685  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  1809218.656 ±   35102.634  ops/s
```
* JsonValue is doing objects > JsonValue > string, so it'll be slower.
* JsonString uses StringBuilder, JsonWriter uses a Writer, otherwise they are the same. They've never been benchmarked before.
* Jackson (didn't look at the others) uses the same approach with an almost identical API. Likely we are slower because:
  * Jackson writes to char[] rather than StringBuilder.
  * Using an `Array<JsonObject>` stack, we allocate just to store 2 booleans. Could be more efficient, eg IntArray with 2 bits. I've done that in this PR, ez.
  * OutputType quoteValue/Name aren't optimized. I improved them a little in this PR.
  * JsonString is 118 LOC while Jackson is... more. Reasonable things we can do to make it better are cool, but it should stay simple as an easy, fast enough built-in utility.
* 5.0x libgdx's 1M ops/sec for minor details, we're measuring something already very fast.
* Can't fully reproduce the above with this PR until merge. Quick test shows JsonString changes are 23.5% faster than before (Jackson still 4.2x).

```
deser --apis stream --libs jackson,gson,fastjson,libgdx
Benchmark                           Mode  Cnt        Score        Error  Units
Deserialization.gson               thrpt   20  1944272.360 ±  15076.934  ops/s
Deserialization.jackson            thrpt   20  6195549.226 ± 250229.263  ops/s
Deserialization.libgdx_JsonReader  thrpt   20  3656443.164 ± 152005.395  ops/s
```
* OK perf, not bad for 410 LOC!
* Quick test shows this PR (removes stack from JsonReader) is just 4.7% faster.
* I put in only JsonReader since 1) JsonMatcher isn't merged yet, 2) JsonSkimmer is too much of a PITA to use directly, and 3) performance should be comparable to JsonReader when no data is skipped, which is what would be fair if they were added.

```
ser --apis databind --libs jackson,gson,fastjson,libgdx
Benchmark                           Mode  Cnt         Score        Error  Units
Serialization.fastjson             thrpt   20  12436915.972 ± 959998.501  ops/s
Serialization.fastjson_features    thrpt   20  15402120.517 ± 498336.255  ops/s
Serialization.gson                 thrpt   20   5021531.585 ±  12116.517  ops/s
Serialization.jackson              thrpt   20  10408988.895 ± 196914.095  ops/s
Serialization.jackson_afterburner  thrpt   20  11455508.231 ± 366570.719  ops/s
Serialization.jackson_blackbird    thrpt   20  10837417.939 ±  73867.267  ops/s
Serialization.libgdx_Json          thrpt   20   1518021.645 ±   6462.528  ops/s
```
```
deser --apis databind --libs jackson,gson,fastjson,libgdx
Benchmark                             Mode  Cnt        Score        Error  Units
Deserialization.fastjson             thrpt   20  7345077.631 ± 115918.039  ops/s
Deserialization.fastjson_features    thrpt   20  8069633.142 ± 153015.910  ops/s
Deserialization.gson                 thrpt   20  1960689.874 ±  27124.649  ops/s
Deserialization.jackson              thrpt   20  5344424.880 ± 240437.574  ops/s
Deserialization.jackson_afterburner  thrpt   20  6297457.793 ± 188526.053  ops/s
Deserialization.jackson_blackbird    thrpt   20  4942411.392 ± 645946.941  ops/s
Deserialization.libgdx_Json          thrpt   20  3304291.309 ±  65806.493  ops/s
```
* I didn't look into `Json` performance but it's mildly interesting to compare:
  * Jackson is at 6.8x ser and 1.6x deser.
  * gson beats us 3.3x for ser but we beat it 1.7x for deser.
* The fixed Json test shows worse performance on my computer than the old test on yours, weird.

tommyettinger:
Woo, I'm glad the benchmark can be useful! Mostly I had to allocate some form of Writer to match most of the other APIs using an OutputStream as their target; libGDX's API is an outlier there. Targeting String output seems fair, since JSON is a String format anyway. 1M ops/second is essentially measuring how many small JSON rows it can serialize or deserialize, and at least for most libGDX purposes, a million rows seems rare. TextraTypist does use libGDX JsonReader to parse the output of msdf-atlas-gen, which can be relatively large JSON files (not a million chars even, but still thousands of `{}` objects). An example JSON font is attached; it might be a nice piece of real-world example data.

[Gentium-standard.json](https://github.com/user-attachments/files/21587802/Gentium-standard.json)

Also, I think all of your tests have an extra digit over mine on a laptop, so your fixed serialize test is actually 3x faster on your machine than the 1.13.1 Json serialize on mine.

NathanSweet:
It's hard to use different data in the benchmarks, except with databinding. Could still be interesting to measure:

~32ms to parse `Gentium-standard.json` with JsonReader. Nothing stands out as being inefficient. It's mostly the parsing, the rest is assembling the JsonValues.

~20ms (1.6x) to parse it with JsonMatcher using pattern `""`, which parses the whole document to JsonValues. It's faster because it only provides JsonValues with types object, array, null, boolean, or String. Numbers result in a string JsonValue, deferring parsing until `asFloat`/etc is called to get a number. The downside/feature to that is a malformed number won't explode until then. Also you lose the fact that the value was a number in the data. I'll add that to the javadoc.

> your fixed serialize test is actually 3x faster on your machine than the 1.13.1 Json serialize on mine.

Aye, but Jackson on your 🥔 is 6.1x faster than libgdx's `Json`, while on mine it's 6.85x. Lots of variables involved, but I still expected the fixed test to do better relatively (`Json` reuse, no string > bytes, etc).

I don't see any more easy performance wins, so I think it's ready to merge. Then we can bench using a nightly and compare with previous runs above.

NathanSweet:
Do we have a nightly SNAPSHOT build somewhere? 🤔 
https://repo.maven.apache.org/maven2/com/badlogicgames/gdx/gdx/

Tom-Ski:
> Do we have a nightly SNAPSHOT build somewhere? 🤔 https://repo.maven.apache.org/maven2/com/badlogicgames/gdx/gdx/

https://central.sonatype.com/repository/maven-snapshots/
Is the repo url for snapshots. The browser ui is broken at the moment apparently, so you cant manually navigate to search, which is really annoying.

NathanSweet:
Thanks @Tom-Ski! That worked after an initial screw up.

NathanSweet:
1.13.1:
```
Benchmark                         Mode  Cnt        Score         Error  Units
Serialization.jackson            thrpt   20  9596064.326 ±  106769.855  ops/s
Serialization.libgdx_JsonString  thrpt   20  1856878.098 ±   26094.107  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1161879.610 ±    2765.685  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  1809218.656 ±   35102.634  ops/s
```

1.13.6-SNAPSHOT (20250805):
```
Benchmark                         Mode  Cnt        Score         Error  Units
Serialization.jackson            thrpt   20  9596064.326 ±  106769.855  ops/s
Serialization.libgdx_JsonString  thrpt   20  2940658.957 ±   99017.003  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1916192.351 ±   23665.428  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  3038512.046 ±   11718.800  ops/s
```
Cool, it's 1.58x faster. Now Jackson wins by 3.26x (instead of 5.0x). I fixed a bug in the libgdx stream test, so part of this difference could be that (oh well).

I'll make a new PR that removes boxing and reduces conditionals. It's not really important but it's kind of fun, and seems to be one of the last reasonable things to do.

tommyettinger:
Woohoo, all around faster Json! That benchmark shows no degradation and just solid improvement, I'm happy for sure!

PR #7684, opened 2025-08-06
JsonWriter/String/Matcher improvements

NathanSweet:
JsonWriter/JsonString performance improvements (follow up to #7683):
* Methods to avoid boxing and conditionals.
* Use stack only for depths < current.
* Avoid comma check after name for set().

JsonMatcher changes from IRL usage:
* Added parseRoot() to return root JsonValue, else it's awkward to use.
* start() for init before parsing.
* Patterns constructor, convenience.
* Javadoc.

---

Can be used with `var` this way:
```
var matcher = new JsonMatcher("(message,access_token)") {
   String error, accessToken;
   protected void process (JsonValue value) {
      error = value.getString("message", null);
      accessToken = value.getString("access_token", null);
   }
};
matcher.parse(json);
if (matcher.error != null) fail(matcher.error);
return matcher.accessToken;
```

---

For another example, I get 120KB of JSON from a weather web service. I start parsing at the part I want, grabbing the first 24 hours out of 200+, then stop before parsing fails due to starting in the middle:
```
String data = ...
int start = data.indexOf("\"hourly\":");
if (start == -1) throw error("Hourly data not found.");
start += 9;
data.getChars(start, start + 11000, chars, 0);
hourlyForecast.parse(chars, 0, chars.length);
...
private final char[] chars = new char[11000];
private final JsonMatcher hourlyForecast = new JsonMatcher("*@/(precip,precip_probability)") {
   int count;
   protected void process (JsonValue value) {
      float rain = value.getFloat("precip"); // mm per hour.
      float chance = value.getInt("precip_probability") / 100f;
      ...
      if (++count == 24) {
         count = 0;
         stop();
      }
   }
};
```

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
@@ -23,7 +23,8 @@ import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.utils.JsonValue.ValueType;
 import com.badlogic.gdx.utils.JsonWriter.OutputType;
 
-/** Efficient JSON parser that does minimal parsing to extract values matching specified patterns using a single pass.
+/** Efficient JSON parser that extracts values matching specified patterns using a single pass. It conveniently collects values
+ * for processing in batches or at once. Values not collected have minimal overhead, ideal for processing a subset of large JSON.
  * 
  * <h4>Match</h4> Match objects, arrays, or field names (not field values) in the JSON:
  * <ul>
@@ -102,10 +103,11 @@ import com.badlogic.gdx.utils.JsonWriter.OutputType;
  * 
  * <h4>Behavior notes</h4>
  * <ul>
- * <li>reject() prevents further matching at this level or deeper, useful for filtering.
+ * <li>When multiple patterns are added they match in parallel, each with independent capture and processing.
+ * <li>reject() prevents further matching at this level or deeper, useful for filtering. A pattern can reject a different pattern.
  * <li>clear() discards unprocessed captured values.
- * <li>end() prevents further matching and ends parsing. stop() does the same but also clears.
  * <li>start() can be overidden to perform setup just before parsing.
+ * <li>end() prevents further matching and ends parsing. stop() does the same but also clears.
  * <li>If not using {@code *@}, {@code **@}, or {@code []} parsing ends once all specified values are captured.
  * <li>A single capture before processing provides the value directly, multiple captures provide an object.
  * <li>parseRoot(), no patterns, or a {@code ""} pattern captures the entire JSON document.
@@ -161,6 +163,7 @@ public class JsonMatcher extends JsonSkimmer {
 	public JsonMatcher () {
 	}
 
+	/** Adds all of the specified patterns. */
 	public JsonMatcher (String... patterns) {
 		for (String pattern : patterns)
 			addPattern(pattern);
--- a/gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
@@ -16,6 +16,10 @@
 
 package com.badlogic.gdx.utils;
 
+import java.io.InputStream;
+import java.io.Reader;
+
+import com.badlogic.gdx.files.FileHandle;
 import com.badlogic.gdx.utils.JsonValue.ValueType;
 import com.badlogic.gdx.utils.JsonWriter.OutputType;
 
@@ -101,9 +105,10 @@ import com.badlogic.gdx.utils.JsonWriter.OutputType;
  * <li>reject() prevents further matching at this level or deeper, useful for filtering.
  * <li>clear() discards unprocessed captured values.
  * <li>end() prevents further matching and ends parsing. stop() does the same but also clears.
+ * <li>start() can be overidden to perform setup just before parsing.
  * <li>If not using {@code *@}, {@code **@}, or {@code []} parsing ends once all specified values are captured.
  * <li>A single capture before processing provides the value directly, multiple captures provide an object.
- * <li>A {@code ""} pattern captures the entire JSON document.
+ * <li>parseRoot(), no patterns, or a {@code ""} pattern captures the entire JSON document.
  * </ul>
  * 
  * <h4>JsonValues</h4>
@@ -113,6 +118,23 @@ import com.badlogic.gdx.utils.JsonWriter.OutputType;
  * <li>JsonValue types are object, array, null, boolean, or String. Numbers are provided as String, deferring number parsing until
  * the JsonValue is converted to a number (eg {@link JsonValue#asFloat()}).
  * </ul>
+ * 
+ * <h4>var keyword</h4> With Java 10+ {@code var} enables access to anonymous class fields, making value collection easier:
+ * 
+ * <pre>
+ * var matcher = new JsonMatcher("(message,access_token)") {
+ * 	String error, accessToken;
+ * 
+ * 	protected void process (JsonValue value) {
+ * 		error = value.getString("message", null);
+ * 		accessToken = value.getString("access_token", null);
+ * 	}
+ * };
+ * matcher.parse(json);
+ * if (matcher.error != null) fail(matcher.error);
+ * return matcher.accessToken;
+ * </pre>
+ * 
  * @author Nathan Sweet */
 public class JsonMatcher extends JsonSkimmer {
 	static private final boolean debug = false;
@@ -126,7 +148,7 @@ public class JsonMatcher extends JsonSkimmer {
 	static final int single  = 0b100000; // @on
 
 	Processor processor;
-	Pattern[] patterns = new Pattern[0];
+	Pattern[] patterns = new Pattern[0], original, all;
 	int total;
 	private boolean rejected;
 	boolean stoppable = true;
@@ -135,6 +157,14 @@ public class JsonMatcher extends JsonSkimmer {
 	char[] chars;
 	final IntArray path = new IntArray();
 	Pattern processPattern;
+
+	public JsonMatcher () {
+	}
+
+	public JsonMatcher (String... patterns) {
+		for (String pattern : patterns)
+			addPattern(pattern);
+	}
 
 	/** This processor is invoked for all pattern matches, after per pattern processors and before {@link #process(JsonValue)}. */
 	public void setProcessor (@Null Processor processor) {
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
```diff
@@ -16,6 +16,10 @@
 
 package com.badlogic.gdx.utils;
 
+import java.io.InputStream;
```
tommyettinger, line 19: I see InputStream is used somewhere here; would writing to an OutputStream be a reasonable feature to add? It would make testing libGDX Json implementations against others a little easier (as in, we could PR that JVM Json benchmark to include libGDX 1.13.6). If OutputStream support is a hard feature to add, I wouldn't worry about it, though. It might be nice for some usages libGDX doesn't have internally, but users may benefit from.

gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
```diff
@@ -136,6 +158,14 @@ public class JsonMatcher extends JsonSkimmer {
 	final IntArray path = new IntArray();
 	Pattern processPattern;
 
+	public JsonMatcher () {
+	}
+
+	public JsonMatcher (String... patterns) {
```
tommyettinger, line 167: Maybe some docs here would be nice? I'm guessing multiple patterns are all tried as if they were separated by OR, so any matches get yielded, but I could also see this producing each pattern's output in turn.

tommyettinger (approved):
Other than maybe some docs being helpful (and the one place that stuck out might already be documented in the class JavaDocs), this looks solid. I especially like all the primitive value() methods in JsonWriter; those will be handy.

gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
NathanSweet: The InputStream is for input, to convert to chars (by the super class).

Usually you dig data out of the matches and stick it in records/objects or otherwise do things with it. I'm not sure piping all the matches to an OutputStream is common enough to warrant built-in support. If you did want to transform the data I think it's easy enough:
```
JsonMatcher matcher = new JsonMatcher();
matcher.addPattern("devices/*@", value -> value.toJson(OutputType.json, writer));
matcher.parse(json);
```

gdx/src/com/badlogic/gdx/utils/JsonMatcher.java
NathanSweet: It just adds all the patterns. I'll add to class docs, "When multiple patterns are added they match in parallel, each with independent capture and processing." Also, about `reject()`, "A pattern can reject a different pattern."

NathanSweet:
1.13.6-SNAPSHOT (20250805):
```
Benchmark                         Mode  Cnt        Score         Error  Units
Serialization.gson               thrpt   20  4140162.229 ± 1289405.740  ops/s
Serialization.jackson            thrpt   20  9596064.326 ±  106769.855  ops/s
Serialization.libgdx_JsonString  thrpt   20  2940658.957 ±   99017.003  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1916192.351 ±   23665.428  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  3038512.046 ±   11718.800  ops/s
```

1.13.6-SNAPSHOT (20250806) with this PR:
```
Benchmark                         Mode  Cnt        Score        Error  Units
Serialization.gson               thrpt   20  4140162.229 ± 1289405.740  ops/s
Serialization.jackson            thrpt   20  9596064.326 ±  106769.855  ops/s
Serialization.libgdx_JsonString  thrpt   20  5522136.958 ±  143411.698  ops/s
Serialization.libgdx_JsonValue   thrpt   20  2650720.617 ±   29119.050  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  5896188.937 ±  460728.437  ops/s
```
1.8x faster than before this PR, though the benchmark has some variation below.

Jackson is now 1.7x JsonString, probably because we support 3 formats and our OutputType quoting does a string allocation instead of writing the quotes directly to the StringBuilder/Writer: `'"' + string + '"'`. Avoiding that isn't very elegant so I think it's fine to leave like this.

Some anomalies: JsonWriter shouldn't be faster than JsonString (but has larger error range) and JsonValue didn't change but the score did (wut). Ran it all again to be sure:
```
ser --apis stream --libs libgdx,jackson,gson
Benchmark                         Mode  Cnt        Score        Error  Units
Serialization.gson               thrpt   20  6086083.604 ± 244361.403  ops/s
Serialization.jackson            thrpt   20  9906212.544 ±  15149.837  ops/s
Serialization.libgdx_JsonString  thrpt   20  5016828.168 ± 169945.325  ops/s
Serialization.libgdx_JsonValue   thrpt   20  2816135.057 ±  68762.870  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  5995487.078 ± 305092.763  ops/s

deser --apis stream --libs libgdx,jackson,gson
Benchmark                           Mode  Cnt        Score        Error  Units
Deserialization.gson               thrpt   20  1913556.244 ±   6610.469  ops/s
Deserialization.jackson            thrpt   20  6535725.986 ± 409016.901  ops/s
Deserialization.libgdx_JsonReader  thrpt   20  3625959.439 ±   9606.868  ops/s

ser --apis databind --libs libgdx,jackson,gson,fastjson
Benchmark                           Mode  Cnt         Score        Error  Units
Serialization.fastjson             thrpt   20  12328006.886 ± 400454.154  ops/s
Serialization.fastjson_features    thrpt   20  14891028.324 ± 399570.868  ops/s
Serialization.gson                 thrpt   20   3959212.274 ±  24236.480  ops/s
Serialization.jackson              thrpt   20  10561399.685 ±  91478.599  ops/s
Serialization.jackson_afterburner  thrpt   20  11856468.778 ±  74816.440  ops/s
Serialization.jackson_blackbird    thrpt   20  11053023.058 ± 190752.002  ops/s
Serialization.libgdx_Json          thrpt   20   3962384.688 ±  73667.265  ops/s

deser --apis databind --libs libgdx,jackson,gson,fastjson
Benchmark                             Mode  Cnt        Score        Error  Units
Deserialization.fastjson             thrpt   20  8782236.597 ± 314024.258  ops/s
Deserialization.fastjson_features    thrpt   20  8172197.488 ± 105518.430  ops/s
Deserialization.gson                 thrpt   20  1947468.899 ±   7779.230  ops/s
Deserialization.jackson              thrpt   20  5795394.423 ± 202042.724  ops/s
Deserialization.jackson_afterburner  thrpt   20  6545159.424 ± 357838.643  ops/s
Deserialization.jackson_blackbird    thrpt   20  6897461.790 ± 202712.242  ops/s
Deserialization.libgdx_Json          thrpt   20  3337827.880 ±  46103.957  ops/s
```

Has been JDK 19 so far, here's JDK 24:
```
ser --apis stream --libs libgdx,jackson,gson
Benchmark                         Mode  Cnt        Score        Error  Units
Serialization.gson               thrpt   20  6109705.528 ±  19749.528  ops/s
Serialization.jackson            thrpt   20  9855181.920 ± 399467.793  ops/s
Serialization.libgdx_JsonString  thrpt   20  4639294.901 ± 389510.707  ops/s
Serialization.libgdx_JsonValue   thrpt   20  2902835.155 ±   9129.966  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  5274055.024 ±  16985.391  ops/s
```

Welp, that's probably enough dorking with this! We've got respectable performance given the tradeoffs.

AFAIK no other lib has efficient skipping with patterns like JsonMatcher. Jackson has streaming and filtering, but doing it manually is tedious (plus the enterprisey API). [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) is close but only identifies a value in the JSON, no wildcards. [JSON Path](https://datatracker.ietf.org/doc/html/rfc9535) is closer but I've only seen it applied to an in-memory DOM and I'm not a fan of filtering in the DSL.

tommyettinger:
I found there's some substantial speedup possible on Java 24 by using the ZGC garbage collector. These use the latest snapshot build; I'm not sure if this PR ( #7684 ) is included yet.

Without ZGC, on Java 24 HotSpot:

```
./run.ps1 ser --apis stream --libs "libgdx,jackson,gson"
...
Benchmark                         Mode  Cnt        Score        Error  Units
Serialization.gson               thrpt   20   764023.069 ±  84439.362  ops/s
Serialization.jackson            thrpt   20  3287666.185 ±  17215.528  ops/s
Serialization.libgdx_JsonString  thrpt   20  2055947.615 ± 241755.356  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1576455.340 ±  20383.485  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  1242307.989 ±  35994.488  ops/s
```
With ZGC enabled, on the same Java 24 HotSpot JVM:
```
./run.ps1 ser --apis stream --libs "libgdx,jackson,gson"
...
Benchmark                         Mode  Cnt        Score        Error  Units
Serialization.gson               thrpt   20   951513.589 ±  28294.372  ops/s
Serialization.jackson            thrpt   20  3245205.875 ±  18333.969  ops/s
Serialization.libgdx_JsonString  thrpt   20  2117777.458 ± 129965.207  ops/s
Serialization.libgdx_JsonValue   thrpt   20  1570682.857 ±  11517.930  ops/s
Serialization.libgdx_JsonWriter  thrpt   20  1278273.285 ± 182119.899  ops/s
```

However, that speedup mostly applies to GSON! That seems to suggest libGDX Json is already producing very very little garbage, since even the generational ZGC doesn't have a significant impact on the libGDX benchmarks. That seems good, especially considering Android, GWT, etc. as targets and their less-than-amazing GCs.

NathanSweet:
1.13.6-SNAPSHOT for 20250806 should have this PR:
```
>gradlew dependencyInsight --dependency gdx --configuration compileClasspath | grep 202508
com.badlogicgames.gdx:gdx:1.13.6-SNAPSHOT:20250806.063703-17
```

Your numbers look good, if some high error variation. Jackson at ~1.5-1.6x is about what I saw. Still confused why I consistently see JsonWriter faster than JsonString.

Interesting about ZGC and gson. We allocate for quotes, mentioned above. StringBuilder may allocate for some primitives, but I'd hope not on desktop. IIRC long ago someone contributed the now CharArray int and long append methods, probably because Android was allocating. I think those two are the only allocation possibilities for JsonString.

tommyettinger:
Yep, that matches the version I have:
`com.badlogicgames.gdx:gdx:1.13.6-SNAPSHOT:20250806.063703-17`

This is definitely a good change; hopefully everything else in the next release will go smoothly (fingers crossed!) and TextraTypist can benefit from the faster Json parsing. Right now having to parse all JSON fonts provided by the library takes maybe 10-15 seconds at startup, which is a lot for a test that I sometimes have to run several times in a row to see small changes. I wonder if there are other ways I can adjust the data itself; it stores a whole lot of JSON Number types as large double-precision decimals, when it really only needs the precision of a Java float... And there could be a common divisor for all of them anyway, which would make them representable as small-ish integers if I just divide when I load it. In any case, the low-to-no garbage production probably helps quite a lot on non-desktop JVMs, and even if we could find any other JSON parser that works on Android, RoboVM, and GWT, it probably won't perform as well on those platforms as what we have at this point.

Issue #7686, opened 2025-08-07
[Bug] Confusing playMode replacement in 2D animation getKeyFrame Method


ZCHawk:
In 'com.badlogic.gdx.graphics.g2d.Animation' 
there is a method called `getKeyFrame` with an additional looping param, I think the parameter looping can override the Animation playMode temporarily when it get its keyFrames, the current code are following:

```java
public T getKeyFrame (float stateTime, boolean looping) {
		// we set the play mode by overriding the previous mode based on looping
		// parameter value
		PlayMode oldPlayMode = playMode;
		if (looping && (playMode == PlayMode.NORMAL || playMode == PlayMode.REVERSED)) {
			if (playMode == PlayMode.NORMAL)
				playMode = PlayMode.LOOP;
			else
				playMode = PlayMode.LOOP_REVERSED;
		} else if (!looping && !(playMode == PlayMode.NORMAL || playMode == PlayMode.REVERSED)) {
			if (playMode == PlayMode.LOOP_REVERSED)
				playMode = PlayMode.REVERSED;
			else
				playMode = PlayMode.LOOP;
		}

		T frame = getKeyFrame(stateTime);
		playMode = oldPlayMode;
		return frame;
}
```

In Addition, the current PlayMode enum has six values where the code is 
```java
public enum PlayMode {
    NORMAL, REVERSED, LOOP, LOOP_REVERSED, LOOP_PINGPONG, LOOP_RANDOM,
}
```

In the first if part, it says when `looping=true` and `playMode` is `NORMAL ` or `REVERSED`, the playMode is changed to corresponding loop mode `LOOP` AND `LOOP_REVERSED`.

Next, in the second `else if` part, when `looping=false` and `playMode` is `LOOP_REVERSED`, the playMode will be replaced to `PlayMode.REVERSED`, which also makes sense, but it turns the other three loop `PlayMode` into `LOOP`, which confuses me.

In my Opinion, all the 4 loop modes shall keep the Animation playing forever. the other 2 non-loop modes will stop the Animation and offer the last frame. But when you are in `LOOP_PINGPONG` or `LOOP_RANDOM` mode, and you want to a non-loop output, I'm not quite sure what kind of output you are looking for, may be no changes shall be made? throw an Exception seems unfriendly.

So perhaps the second part can be written like this: 
```java
else if (!looping && !(playMode == PlayMode.NORMAL || playMode == PlayMode.REVERSED)) {
[...]

PR #7688, opened 2025-08-09
fix play mode replacement, 

ZCHawk:
when looping=false and PlayMode=loop, temporarily replace mode to normal.
#7686

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
@@ -80,6 +80,7 @@ public class Animation<T> {
 
 	/** Returns a frame based on the so called state time. This is the amount of seconds an object has spent in the state this
 	 * Animation instance represents, e.g. running, jumping and so on. The mode specifies whether the animation is looping or not.
+	 *  But if the animation is set to LOOP_PINGPONG and LOOP_RANDOM mode, non-loop param takes no effect.
 	 * 
 	 * @param stateTime the time spent in the state represented by this animation.
 	 * @param looping whether the animation is looping or not.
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
```diff
@@ -80,6 +80,7 @@ public Animation (float frameDuration, T... keyFrames) {
 
 	/** Returns a frame based on the so called state time. This is the amount of seconds an object has spent in the state this
 	 * Animation instance represents, e.g. running, jumping and so on. The mode specifies whether the animation is looping or not.
+	 * But if the animation is set to LOOP_PINGPONG and LOOP_RANDOM mode, non-loop param takes no effect.
```
tommyettinger, line 83: I have some minor comment nitpicks. It should be:

`LOOP_PINGPONG or LOOP_RANDOM`

and the verb in `param takes no effect` should be changed to `param has no effect` .

It's clear enough already that it can be merged now, but if you want to polish it before merging that would be nice.

gdx/src/com/badlogic/gdx/graphics/g2d/Animation.java
```diff
@@ -94,10 +95,12 @@ public T getKeyFrame (float stateTime, boolean looping) {
...
 		} else if (!looping && !(playMode == PlayMode.NORMAL || playMode == PlayMode.REVERSED)) {
-			if (playMode == PlayMode.LOOP_REVERSED)
+			if (playMode == PlayMode.LOOP_REVERSED) {
 				playMode = PlayMode.REVERSED;
-			else
-				playMode = PlayMode.LOOP;
+			} else if (playMode == PlayMode.LOOP) {
+				playMode = PlayMode.NORMAL;
+			}
+			// non-loop param takes no effect on LOOP_PINGPONG and LOOP_RANDOM mode
```
tommyettinger, line 103: This comment isn't strictly necessary, but I think that if you want this here at all, it should be `if the loop param is false, LOOP_PINGPONG and LOOP_RANDOM playModes won't change` . Again, a very minor nitpick, especially since users usually won't be reading this.

PR #7694, opened 2025-08-27
Support for integer arrays `glUniform_iv` functions in ShaderProgram

WinterAlexander:
Adding convenience functions in `ShaderProgram` to input integer arrays into a shader program without needing to call OpenGL directly. Directly matches the already present `glUniform_fv` functions.

Important use case: array of texture sent to the shader, for example with cascaded shadow maps

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@
 - API Addition: Added JsonValue#toJson that takes a Writer.
 - API Addition: Added getProgrammaticChangeEvents() to scene2d.ui actors that have setProgrammaticChangeEvents.
 - API Addition: Added JsonMatcher, extracts values with pattern matching.
+- API Addition: Added setUniform_iv functions in ShaderProgram to set uniform integer arrays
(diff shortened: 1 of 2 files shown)


MonstrousSoftware:
Beware that using an index into an array  causes issues for some GPU drivers (shader compile error).  It was implemented, also for use case of cascaded shadow maps, in gdx-gltf but in the end it had to be reverted to separate variables for each cascade to be compatible with all drivers.
 
https://github.com/mgsx-dev/gdx-gltf/pull/127

WinterAlexander:
> Beware that using an index into an array causes issues for some GPU drivers (shader compile error). It was implemented, also for use case of cascaded shadow maps, in gdx-gltf but in the end it had to be reverted to separate variables for each cascade to be compatible with all drivers.
> 
> [mgsx-dev/gdx-gltf#127](https://github.com/mgsx-dev/gdx-gltf/pull/127)

Thank you for pointing this out. That being said, the ability to pass an array of int (not just for sampler2D) should be supported without issue

PR #7533, opened 2024-11-27
Added Template Object Support for TiledMaps. Replacement for PR (#5938)

BoBIsHere86:
(This should be the very last TiledMap related PR in relation to the Cleanup Initiative [#7488](https://github.com/libgdx/libgdx/issues/7488))

This PR is a replacement for the Oldest(4 years old) TileMap PR currently on the list. #5938. Closes https://github.com/libgdx/libgdx/issues/5102

Unlike the other replacement PRs. I had to start this one from scratch since so much of the map codebase had changed since then. So anyone who has the spare time, feel free to take a look and provide comment's where you see fit.

I have gone ahead and added a new Test as well. You can even toggle between both the tmj and tmx loader to make sure everything is the same.
When a template object is used and a user changes the value of a property. That property's new value should take precedence. Overriding the default value.
The test is meant to show the default template object with it's properties and another template object with overridden property values, to make sure they are being parsed properly.
You can see below,
templateTest_es

And that's pretty much it, this should probably be one of the final PRs merged when the time comes as I will probably have to update it to make sure it's supports the new PointObject and TextObjects we have added in the other PRs.

PR #7698, opened 2025-09-05
Small but Important fix to the TmjMapLoader.

BoBIsHere86:
While checking out the leftover Tiled PRs, after rebasing I noticed the TMJ map tests were failing.

This was because of a change to the JsonValue class, recently on August 8th.
A new method, JsonValue.getIgnoreCase()  was added and the case insensitivity from .get() was removed.

In the TmjMapLoader we were mistakenly using "tileSets" instead of the proper "tilesets".

But since the original JsonValue.get() was always comparing with equalsIgnoreCase. It just "worked" anyways.

But it was wrong to be checking against that string, it should be looking for "tilesets". So I have fixed it.

If you are wondering why I did not instead change it to use the newly added JsonValue.getIgnoreCase() method. It is because the tmx loader IS case sensitive so we should keep functionality the same.

That is all. But this should most definitely be merged before the next release or the TMJ maps will begin breaking.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
@@ -109,7 +109,7 @@ public class TmjMapLoader extends BaseTmjMapLoader<BaseTmjMapLoader.Parameters> 
 		Array<FileHandle> fileHandles = new Array<>();
 
 		// TileSet descriptors
-		for (JsonValue tileSet : root.get("tileSets")) {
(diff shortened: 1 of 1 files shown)


BoBIsHere86:
Just a small note. In order to get the Template tests to run i actually had to add this fix in the recently merged https://github.com/libgdx/libgdx/pull/7533.

But I didn't realize that PR would get merged so fast. And this bug was important and wanted to document it. So technically, this issue is now fixed in the current master and we wont need this to be merged anymore since the fix is in place. But it's also good to have a PR mentioning why the change happened.
So do with this whatever you like. Close it or give it a redundant merge. Etc..

Issue #7626, opened 2025-04-16
ANR/Crash on Android due to infinite wait() in AndroidGraphics.destroy()


Diman96:
We are experiencing a critical ANR crash when our application (using `AndroidFragmentApplication`) is paused. The crash occurs because the main (UI) thread is blocked indefinitely while waiting in `AndroidGraphics.destroy()`. Specifically, the crash stack trace points to an infinite wait at line 431 of `AndroidGraphics.java`:

```
  at java.lang.Object.wait (Native method)
  at java.lang.Object.wait (Object.java:405)
  at java.lang.Object.wait (Object.java:543)
  at com.badlogic.gdx.backends.android.AndroidGraphics.destroy (AndroidGraphics.java:431)
  at com.badlogic.gdx.backends.android.AndroidFragmentApplication.onPause (AndroidFragmentApplication.java:197)
  at com.spaceinc.covaragers.fragments.MapFragment.onPause (MapFragment.java:47)
  at androidx.fragment.app.Fragment.performPause (Fragment.java:3314)
  at androidx.fragment.app.FragmentStateManager.pause (FragmentStateManager.java:631)
  at androidx.fragment.app.FragmentStateManager.moveToExpectedState (FragmentStateManager.java:293)
  at androidx.fragment.app.FragmentStore.moveToExpectedState (FragmentStore.java:113)
  at androidx.fragment.app.FragmentManager.moveToState (FragmentManager.java:1435)
  at androidx.fragment.app.FragmentManager.dispatchStateChange (FragmentManager.java:2979)
  at androidx.fragment.app.FragmentManager.dispatchPause (FragmentManager.java:2915)
  at androidx.fragment.app.FragmentController.dispatchPause (FragmentController.java:296)
  at androidx.fragment.app.FragmentActivity.onPause (FragmentActivity.java:285)
...
```

It appears that during the pause process, the `destroy()` method calls `synch.wait()` without any timeout. As I assuming this can lead to a deadlock if the corresponding notify is never invoked, which in turn causes an ANR.

**What We Tried:**
We attempted to adapt the fix discussed in [issue #7342](https://github.com/libgdx/libgdx/issues/7342). However, on some devices, that solution led to even more serious crashes, so we are unable to use it in our project.

**Steps to Reproduce:**
[...]

PR #7699, opened 2025-09-07
Fix ANR/crash in AndroidGraphics.destroy() due to infinite wait()

palatin:
### Summary
This PR resolves an ANR/crash in `AndroidGraphics.destroy()` ([#7626](https://github.com/libgdx/libgdx/issues/7626)) caused by an infinite `wait()` when `onDrawFrame()` is never invoked

### Technical Details
- The fix follows the same approach already used in the `pause()` method: it manually enqueues a call to `onDrawFrame`.
- This ensures that `destroy()` can complete without blocking indefinitely.
- It applies the same logic introduced in [PR #6745](https://github.com/libgdx/libgdx/pull/6745), but that PR was closed without being merged.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -428,6 +428,20 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 			running = false;
 			destroy = true;
 
+            view.queueEvent(new Runnable() {
+                @Override
+                public void run () {
+                    if (!destroy) {
+                        // destroy event already picked up by onDrawFrame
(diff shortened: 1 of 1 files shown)


bitdream:
When  will it be merged? I can see this bug in my console log too.

bitdream:
> Apparently I reviewed these changes in an earlier PR and approved them. That PR couldn't be merged, so the review went away, but I'll still approve this one... even though I really don't do much Android stuff at all, and don't know if this will work or if it is still needed.

Does aprroved mean it will be merged into the master brach so we can get the update?
I see the info showed is "can be merged" not "has been merged" so it has not been merged yet.

obigu (approved):
The issue seems to only affect `AndroidFragment` based apps and the fix has been confirmed to work. Should be harmless for non Fragment apps.

Issue #7573, opened 2025-01-20
TextureArray constructor that takes an Array of textures.


JDSweet:
It would be convenient if the TextureArray class had a constructor that takes an array of textures, instead of an array of filehandles.

mgsx-dev:
That's not how it works, Texture and TextureArray are different type of texture GPU side and you can't convert from one to the other GPU side.
If you need to use Pixmaps instead of FileHandles, you can use FileTextureArrayData when creating your TextureArray.

JDSweet:
@mgsx-dev I'd appreciate having a chance to respond before the issue is closed. You can only create a FileTextureArrayData instance with an array of FileHandles. 

I wasn't referring to the Texture class, I was trying to refer to images conceptually (so, Pixmaps). And my point was, being able to create a TextureArray with a list of procedurally generated Pixmaps (without having to first write them to the filesystem and reload them) is useful functionality.

mgsx-dev:
my bad, it missing a constructor `FileTextureArrayData (Pixmap.Format format, boolean useMipMaps, TextureData[] data)` i think you can PR that.

streamsoftdev:
I made my own `PixmapTextureArrayData` class that implements my own `AdvancedTextureArrayData` interface that extends the existing `TextureArrayData` interface. This allows me to construct a texture array either from an array of textures in memory, or as GL only texture arrays.

PR #7706, opened 2025-09-21
Add a constructor taking a TextureData array to FileTextureArrayData

armingensler:
This PR just adds a new constructor to `FileTextureArrayData` which takes a `TextureData[]` instead of file handles.

It also adds a check to `FileTextureArrayData.prepare()` to make it work with already prepared texture data instances, like `PixmapTextureData`.

I'm not completely sure if `FileTextureArrayData` is the right place to add this, because I'm actually using it for `PixmapTextureData`. But it seems to be an easy extensions and is already mentioned in (#7573).


mgsx-dev (approved):
LGTM

PR #7707, opened 2025-09-29
Move ScreenUtils/clear implementation to AbstractGraphics

damn:
I realised that the ScreenUtils/clear method does not give me the possibility to pass a Graphics instance myself. (I prefer to not use global state and write libgdx via clojure, a functional programming language).

While adding an alternative method with the Graphics parameter, I realised that the method only works on the Graphics object, so it actually kind of belongs in AbstractGraphics.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
+++ b/gdx/src/com/badlogic/gdx/utils/ScreenUtils.java
@@ -66,7 +66,7 @@ public final class ScreenUtils {
 	 * @param clearDepth Clears the depth buffer if true.
 	 * @param applyAntialiasing applies multi-sampling for antialiasing if true. */
 	public static void clear (float r, float g, float b, float a, boolean clearDepth, boolean applyAntialiasing) {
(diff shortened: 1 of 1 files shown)


damn:
To add some more background: I have been using a context map in my game instead of relying on global state.

This PR is experimental and more of an idea, I am aware the changes are quite inconsequential.

But if it gets approved, more PRs like this may follow. Long term I think it would be great to give the libgdx library user the option him or herself how to handle state and to set the com.badlogic.gdx.Gdx globals or not.

To achieve this just requires alternative parameter versions of methods or constructors which rely on Gdx, I believe.

crykn:
Hm, I'm not sure about this change and I guess some others aren't either, or they would have commented by now. You are stating yourself that libGDX is currently relying on using a global state, so IMO we should either change that as a whole (which won't happen as it would be quite a big and disruptive change) or don't do it at all.

mgsx-dev:
i agree with @crykn and i think that if you wanted to track down OpenGL state, you should do that in your own GL20/GL30/GL31/GL32 implementations, the same way it's done by GLProfiler (eg. GL20Interceptor) which would make this change unnecessary.

PR #7709, opened 2025-10-01
Android Maintenance

obigu:
- Removed legacy style support since current min SDK is 21.
- Upgraded AGP to 8.11.2. Reason not to update to 8.12.3 which is already supported by major IDEs is both lack of extensive testing (in personal libGDX projects) and when tested I've run across OutOfMemory issues during App Bundle building which require some further investigation.


tommyettinger (approved):
Thanks for thoroughly testing the AGP version. We definitely want to avoid running out of memory if at all possible! I don't really do Android development, so my review isn't exactly deep or meaningful, but I'm hoping anyone else who does use Android more can take a look.

PR #7710, opened 2025-10-01
Upgrade Gradle to 8.14.3 properly

obigu:
Somehow I've messed up the previous update...


tommyettinger (approved):
Looks correct for the result of running `gradlew wrapper` when the JAR was out of date before. We should go by what the machine says for whether this is valid or not, of course!

PR #7711, opened 2025-10-03
Clean-up in CharArray.

tommyettinger:
There's some small behavior fixes in CharArray, but I tried to stay as true as possible to the other existing Array types when there was any question. `CharArray.random()` shouldn't thrown an exception if the CharArray is empty because no other Array types throw exceptions in that situation -- they all return 0 or something similar. `appendAll(T...)` should be final to avoid "heap pollution," though I really don't know what heap pollution is or does (it seems bad, and final methods that take generic varargs avoid it, but non-final, non-static, non-private ones can't). `equals()` had a redundant check for size that could be elided, and there was an explicit call to Long.toString that didn't need to be written out (it was already in a String concatenation). There's various doc cleanup, such as for broken links in JavaDocs.

That's about it. I like the new changes in CharArray! Almost all of these changes have been applied to CharList in my jdkgdxds library already, except the IO-related ones involving Reader or Writer. I may add those later... This was a good inspiration to get CharList working as an Appendable CharSequence! I also made a lot of my existing code append easily to Appendable CharSequence types, such as StringBuilder, CharList, and now CharArray.


gdx/src/com/badlogic/gdx/utils/CharArray.java
```diff
@@ -815,8 +815,8 @@ public CharArray appendAll (Iterator<?> iter) {
 
 	/** Appends each item in an array to this CharArray without any separators. Each object is appended using
 	 * {@link #append(Object)}. */
-	public <T> CharArray appendAll (T... array) {
-		if (array.length > 0) {
+	public final <T> CharArray appendAll (T... array) {
+		if (array != null) {
```
NathanSweet, line 819: Null should probably crash naturally. It's an error unless we think there is some reason passing null should do nothing. If we keep the null check, the parameter should have javadoc or `@Null`.

tommyettinger: This PR was merged over a month and a half ago.

Issue #7717, opened 2025-10-30
[Tiled] tsx file not loading with a default class attribute


Quillraven:
Hi,

I updated LibGDX in one of my [projects](https://github.com/Quillraven/Masamune) to 1.14 and now the loading of menu.tmx fails (I guess all of them fail but menu.tmx is the first one that is loaded and used).

The problem is the `loadMapPropertiesClassDefaults` method of the `BaseTiledMapLoader` class:
```
// Properties of the type 'class' are already correctly loaded before in loadProperties and can
// therefore be ignored here. They are already excluded via the 'containsKey' check above.
String value = classMember.defaultValue.asString();
if ("object".equals(classMember.type)) {
    loadObjectProperty(mapProperties, propName, value);
} else {
    loadBasicProperty(mapProperties, propName, value, classMember.type);
}
```

According to the comment, class properties are already loaded before but in my scenario this is not the case. The classMember is the "stats" class which is an empty JsonValue of {} and which then cannot be converted to a string via the `asString()` call.

<img width="1380" height="936" alt="Image" src="https://github.com/user-attachments/assets/b05d751b-3873-4810-a99d-3a8f0ea94f49" />

The issue comes from the objects.tsx file that I use to define objects for the game (see screenshot below). The ItemObject "terealis_plant.png" uses the project class `ItemObject` which has a `stats` property that is of type `class`. This specific item is a Quest item and has no stats. That's why we get the empty Json Object in the LibGDX parser.

<img width="1568" height="1099" alt="Image" src="https://github.com/user-attachments/assets/9db7a132-2c60-43a0-bcee-ce23f1d4883a" />

I will attach the entire maps asset folder containing the project file + maps + tilesets. Loading the menu.tmx file should reproduce the problem.

@BoBIsHere86: I am pretty sure that the code is from me but you might have a better understanding why this is happening and how to resolve it. I am unfortunately quite busy at the moment and will not have time to look at it. Maybe in December.

I hope this is not a big issue for others. For me it is rather simple to work around it. I just need to provide a value to one of the `stats` members and then it seems to work again.

[maps.zip](https://github.com/user-attachments/files/23227268/maps.zip)

tommyettinger:
I'd think the empty JSON value should probably be converted to a String, likely an empty String, when asString() is called... This worked in 1.13.1 or 1.13.5, so we need to figure out what changed, I think.

Quillraven:
[...]

PR #7718, opened 2025-10-31
fix nested class property defaults loading

Quillraven:
fix for #7717.

@BoBIsHere86 please play around with it as well whenever you have time.

The problem was with nested classes. My example was assigning a "project class" that has a class property itself to a tile but not specifying any of the values of the class property.

We then got an empty Json object in `loadMapPropertiesClassDefaults` and the `classMember.defaultValue.asString()` then failed.

The comment that class properties were already loaded before was wrong. It is true for flat class properties but it is not the case for nested classes (or at least not in that specific scenario).

We now handle this case with the `"class".equals(classMember.type)` branch. I also enhanced the test and tried it in GWT and worked on my machine. I was also able to reproduce the issue of my project with this test initially.

The new test object looks like this (Class is assigned to a project class called "testClassNested" which has another class property called "classClass"):
image

I also played around with it some more by adding more nesting and overriding certain values on different levels. For me it looked fine but would be great if someone else also plays around a little bit with it.

----

Iirc this method and the problem was introduced by me. Sorry for that! :(

### DIFF
--- a/tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapPropertiesTest.java
+++ b/tests/gdx-tests/src/com/badlogic/gdx/tests/TiledMapPropertiesTest.java
@@ -202,18 +202,18 @@ public class TiledMapPropertiesTest extends GdxTest {
 		objProps.remove("rotation");
 		verifyProperty("classObjProps", expectedProps, objProps);
 
-        // verify that class with just a string also works
-        mapObj = tiledMap.getLayers().get("object layer").getObjects().get("Tile Object No Class");
-        objProps = mapObj.getProperties();
-        expectedProps = new MapProperties();
-        expectedProps.put("type", "NonExistingClass");
-        objProps.remove("x");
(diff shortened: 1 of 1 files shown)


tommyettinger (dismissed):
Change seems good. Test looks solid. Are we sure that the old case, where the property was actually a String, still works here in case old maps actually relied on that?

Quillraven:
Good input. I think it should work because the type should be string instead of class but I will add another object to the properties test map to validate it and to not break it in the future ;)

BoBIsHere86:
I should have some free time this weekend. I'll give this a quick once over and a test, then report back any findings.

BoBIsHere86:
Had a moment to finally take a look. This seems pretty solid, as well as the tests. 
The fix was pretty isolated to the newer class related features. So I wasn't too worried about this messing with anything else. I did a double check and went through all the tiled tests locally, just incase. All ran as predicted. 
I think this is a nice clean fix.  👍

PR #7720, opened 2025-11-05
Make PoolManager compatible with r8 full mode

Berstanio:
r8 full mode does horizontal class merging, if no explicit reference to a class exists. This is problematic, because the PoolManager uses classes as keys. With tree shaking, this can lead to situations, where registering a pool fails, because the reference to the class was stripped, but the registration of the pool still exists.


obigu (approved):
Tested with adapted libGDX Tests app and it seems to work.

PR #7722, opened 2025-11-12
Add return values to InputProcessor#removeProcessor

soundasleep:
If someone tries to remove an `InputProcessor` from an `InputMultiplexer` that is not actually added as a processor, it will silently do nothing.

This change (not tested) adds two return values to the two different `InputMultiplexer#removeProcessor()` methods: one returning the boolean, and one returning the processor removed at the given index.

Also adds a couple of JavaDocs.


tommyettinger (approved):
Looks good from here; the returned values can safely be ignored by existing code that expects them to return `void`, and this should help avoid any accidental multiple-removals by index.

PR #7724, opened 2025-11-19
Ensure space and u007F are also in Hiero's Extended character set.

tommyettinger:
The absence of space, in particular, is quite bad, since every font has the width of the space char as part of its design, and both ASCII and NEHE modes included space already. I'm not sure exactly why NEHE includes u007F, since it's a reserved code point after the end of normal ASCII chars (after `~`), but many fonts do have some glyph shown when it is entered. Extended should, logically, be the extended version of NEHE, which itself is extended ASCII, and shouldn't omit anything from either.

I found this was an issue due to a font, generated by Hiero and submitted as part of https://github.com/tommyettinger/textratypist/pull/34 , missing the vital character with id=32, the space. TextraTypist can load BitmapFont objects without a space, technically, but it can't load a .fnt without one at this point in time. It struck me as odd right from the start, because I've never seen any usable .ttf or .otf file without a space char. I think I've seen only one .ttf without one because it only included the single character 'a', as a test or demo.


tommyettinger:
TextraTypist can now load fonts without a space, but this is still a weird omission for just the largest of the three character sets to be missing ' '.

PR #7725, opened 2025-11-21
Move dispose() method to Audio interface

obigu:
All backend `Audio` interfaces add `dispose()`  except from gwt. This PR adds it to the general  interface. Other than simplifying, it allows easier integration to 3rd party libs such as gdx-miniaudio that want to provide an `Audio` implementation.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.14.1]
+- [BREAKING CHANGE] iOS: Removed IOSAudio#willTerminate to be replaced by new Audio#dispose.
 
 [1.14.0]
 - [BREAKING CHANGE] API: Pools changes in 1.13.5 have been partially reverted. Pools#get/Pools#obtain method return to requiring a Class parameter. Reflection is avoided by using pre-populated Pools for libgdx classes. See https://github.com/libgdx/libgdx/pull/7648 and https://github.com/libgdx/libgdx/pull/7678


gdx/src/com/badlogic/gdx/Audio.java
```diff
@@ -100,4 +100,7 @@ public interface Audio {
 	 *
 	 * @return A array of available output devices */
 	public String[] getAvailableOutputDevices ();
+
```
TCreutzenberg, line 104: Just out of curiosity, would there be any advantage/disadvantage of extending Audio with Disposable?
I.e. "Audio extends Disposable"?

fgnm:
I'm 100% for this, as otherwise it's not possible to create a full drop-in alternative to the whole GDX audio stuff (users should dispose the custom audio by themself in the right moment), but I also agree with @TCreutzenberg , we have the `Disposable` interface so makes sense for the `Audio` interface just extends `Disposable` without introduce its own method.

gdx/src/com/badlogic/gdx/Audio.java
obigu: For a small interface and for just one method I personally prefer to see all the methods of the interface in one place when I inspect it, but it's a personal taste. I'll extend Disposable that seems to be the preferred option.

PR #7729, opened 2025-11-30
Handle Android Audio lifecycle listeners on main/UI thread

obigu:
For some reason `AndroidAudio` lifecycle related callbacks are called asynchronously from the GLThread through `LifecycleListeners` (see `AndroidGraphics`) instead of main thread. 

This has a several issues.
1. There are no guarantees of order of execution or even that the execution will occur between Android Activity (or equivalent) lifecycle events (`onResume()`, `onPause()` and `onDestroy()`) and GLThread `LifecycleListener` (and also Android `ApplicationListener`) `resume()`, `pause()` and `dispose()` calls.
2. . It makes 3rd party `Audio` implementations much harder as it doesn't allow reusing the standard Android lifecycle events.

I've created this as a Draft because some code I've removed I don't know why it was there in the first place but when testing it using the libGDX test suite it appears to work as expected.

Let me know if I'm missing something important here, otherwise imo this makes much more sense.


fgnm:
Uhm I'm wondering why this was made in the first place and why it was needed the hasFocus listener... However, `pause` and `resume` should also be moved to the `Audio` interface imo, otherwise you still cannot drop-in replace the default Audio object with a 3rd party one

obigu:
> Uhm I'm wondering why this was made in the first place and why it was needed the hasFocus listener... However, `pause` and `resume` should also be moved to the `Audio` interface imo, otherwise you still cannot drop-in replace the default Audio object with a 3rd party one

Yeah, was thinking it might be a separate PR but might as well do it in this one.

Issue #7734, opened 2025-12-04
XmlReader.Element parent is immutable after construction, making DOM node movement impossible without rebuilding subtrees


LyricSeraph:
### Problem

In [XmlReader.Element](https://github.com/libgdx/libgdx/blob/323c1515ddf3855b6706caf47050acd9d5f7a057/gdx/src/com/badlogic/gdx/utils/XmlReader.java#L460), the _**parent**_ field is assigned only in the constructor and **never updated** afterwards.

This makes it impossible to correctly move an existing node in the DOM tree. Once an Element is constructed, its parent relationship is effectively immutable.

For example, removing a child from one parent and adding it to another does not update the parent field at all.
The logical structure becomes inconsistent.

### Impact

This design makes DOM-style tree operations impossible:

- Moving nodes between parents
- Reorganizing a subtree
- Any operation that assumes getParent() reflects the actual tree

If a subtree needs to be moved, the only way today is to:

- Create a new Element
- Rebuild the entire subtree
- Copy attributes, children and text manually

This is very costly and surprising for something that otherwise looks like a DOM-style structure.

### Expected Behavior

When performing tree mutation operations, the parent field should stay consistent:

- When a child is added, its parent should be reassigned
- When a child is replaced, the new node should update its parent
- When a child is removed, its parent should be cleared or updated

Otherwise, getParent() does not reflect the current structure and cannot be relied upon.

### Workaround (current)

Although this can be worked around at the moment, by not using the parent field at all and instead treating the DOM as a pure downward structure. But it makes code unnecessarily verbose and indirect.

Another possible workaround is to modify the parent field via reflection.
However, on modern JREs (Java 9+ with module system), this is not practical in real-world projects, which requires to control JVM startup options.

crykn:
Fixed by #7735.

PR #7735, opened 2025-12-04
Update XmlReader.Element's parent field when element is added/removed…

LyricSeraph:
… ([#7734](https://github.com/libgdx/libgdx/issues/7734))

As expained in the issue.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,8 +6,14 @@
                         - `NativeInputConfiguration#maxTextLength` is not nullable anymore. "-1" should be used for unset.
                         - `TextInputWrapper#setText/setPosition` have been removed and replaced with `writeResults`
                         - `TextInputWrapper#shouldClose` has been removed and should now be handled over the close callback
+- [BREAKING CHANGE] API: `TextField.OnscreenKeyboard` has been refactored. `OnscreenKeyboard#show(boolean)` has been split of in `OnscreenKeyboard#show(TextField)` and `OnscreenKeyboard#close()`.
 - API Addition: `Input#KeyboardHeightObserver` was extended by `onKeyboardShow` and `onKeyboardHide`. See javadocs for more info.
 - API Addition: `Input#isTextInputFieldOpened` has been added, to check, whether an InputField opened by `openTextInputField` is open
+- API Addition: Added support for the native input API in scene2d. It can be enabled via `TextField#DEFAULT_ONSCREEN_KEYBOARD = new NativeOnscreenKeyboard()` on startup. This only affects mobile platforms.
+- API Addition: Added multiple NativeInput exclusive options to `TextField`
+                - `setAutocompleteOptions` will set a list of autocomplete options to the user on mobile
+                - `setKeyboardType` will set the keyboard type shown to users
+                - `preventAutoCorrection` will disable auto correct for the textfield
 - Android: The KeyboardHeightObserver will not report duplicate events anymore
 - Android: The KeyboardHeightObserver will now only consider `systemBars`, `ime`, `displayCutout`, `mandatorySystemGestures` for insets.
 - Android: Added missing soundId based API implementations to AsynchronousSound.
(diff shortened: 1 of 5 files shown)


crykn (dismissed):
I don't use XmlReader, but that looks like a sensible change.

SimonIT:
I think it would be good if the ragel file is changed accordingly

LyricSeraph:
> I think it would be good if the ragel file is changed accordingly

Oh, I was not aware of these related contents. Can someone explain to me how to due with it? It looks like those files were generated by program.

tommyettinger:
Ragel is a code generator that takes in the .rl file as input and generates a (typically very complicated) pure-Java file that can be used as a parser. Any files that have a comment at the top that say "generated by Ragel" can't be modified by hand, since the Ragel output would replace your changes. Someone would need to make the changes in the Ragel file, which has a lot of pretty normal Java sources when it isn't in the parser section.

The one changed file starts with:
```java
// line 1 "XmlReader.rl"
// Do not edit this file! Generated by Ragel.
// Ragel.exe -G2 -J -o XmlReader.java XmlReader.rl
```
But the changes look like they should be outside the parser messy zone, and in normal Java template-y stuff in the XmlReader.rl file.

gdx/src/com/badlogic/gdx/utils/XmlReader.java
```diff
@@ -520,6 +520,7 @@ public Element getChild (int index) {
 		public void addChild (Element element) {
 			if (children == null) children = new Array(8);
 			children.add(element);
+			if (element != null) element.parent = this;
```
NathanSweet: Don't need a null check here. Parameters should be assumed not null unless it has javadoc or `@Null`.

This method should have had a null check and throw, but with your change (after the null check is removed) the method will crash if an null parameter is given.

gdx/src/com/badlogic/gdx/utils/XmlReader.java
```diff
@@ -531,21 +532,30 @@ public void setText (String text) {
...
 			parent.removeChild(this);
+			parent = null;
 		}
 
 		public void replaceChild (Element child, Element replacement) {
 			if (children == null) throw new GdxRuntimeException("Element has no children: " + name);
 			if (!children.replaceFirst(child, true, replacement)) {
 				throw new GdxRuntimeException("Element '" + name + "' does not contain child: " + child);
+			} else {
+				if (child != null) child.parent = null;
```
NathanSweet: Don't need this null check either. This method should check its args at the top to ensure a crash when null is passed.
```
if (child == null) throw new IllegalArgumentException("child cannot be null.");
if (replacement == null) throw new IllegalArgumentException("replacement cannot be null.");
```

NathanSweet (changes requested):
Looks good after the 2 changes mentioned! However this is a generated file. Please change [XmlReader.rl](https://github.com/libgdx/libgdx/blob/master/gdx/res/com/badlogic/gdx/utils/XmlReader.rl) instead, then run Ragel using the command at the top of that file. Or make the exact changes to both the `.rl` and `.java` file if that's easier.

LyricSeraph:
> Looks good after the 2 changes mentioned! However this is a generated file. Please change [XmlReader.rl](https://github.com/libgdx/libgdx/blob/master/gdx/res/com/badlogic/gdx/utils/XmlReader.rl) instead, then run Ragel using the command at the top of that file. Or make the exact changes to both the `.rl` and `.java` file if that's easier.


I've done the corresponding modifications on the ragel file. And then run the ragel command to generate the java file.

But I'm a bit scare of the new changes...
It produces much more changes on the format. I don't know if it has something to do with my ragel version.

I'm using the ragel installed from pacman on Arch:
```
Ragel State Machine Compiler version 6.10 March 2017
Copyright (c) 2001-2009 by Adrian Thurston
```

gdx/src/com/badlogic/gdx/utils/XmlReader.java
LyricSeraph: Removed, and add a null check at the beginning.

gdx/src/com/badlogic/gdx/utils/XmlReader.java
LyricSeraph: Get it, and I found that I forgot to update `replacement.parent = child.parent`. Both added.

LyricSeraph:
<s>
sto
The build failed as expected due to the formatting issue. :(

Are there any tools to format the file?
</s>

I've fixed the formating issue by running the spotless gradle task.

LyricSeraph:
As the PR is merged, should I close the related issue? Or wait until the release of the new verson?

Issue #7740, opened 2025-12-05
Hiero freezes and consumes all RAM if glyphs don't fit on page


Frosty-J:
Reproduction steps:

1. In Rendering -> Glyph cache, reduce the page size, such as to 128x128.
2. Set the font size to something large, such as 128.

<img width="502" height="95" alt="Image" src="https://github.com/user-attachments/assets/880fe583-b753-4d2c-820c-3ff1abcfac58" />

tommyettinger:
I think I'll look into this, it looks like it's some kind of infinite loop when it can't place even one glyph.

PR #7741, opened 2025-12-06
Hiero infinite loop fix

tommyettinger:
This solves #7740 by avoiding creating any additional pages when no glyphs could be placed at all (generally due to the page size being tiny and the font size being huge).

It also disables the splash screen, because that seems to crash when launching from IDEA (including debugging, which this needed a lot of to solve) without some extra configuration to let the splash screen be available in the IDE's working directory. The non-functional splash screen is also pretty dated as a thing for GUI apps in 2025... If it was loading an update or something, that would be fine, but it just isn't doing anything for 2 seconds.

I didn't delete the Splash class (even though it isn't used) in case there's some sentimental value to it; if we want to keep it, I just need to uncomment two lines.


crykn (approved):

Frosty-J:
Agree with getting rid of splash screen, though I would remove the comment about it - launching from IDEA with it seems fine (maybe OS-dependent or some rubbish), and asking there if it adds anything is snarky. It appears at the same time as the application itself, on my end, so doesn't even act as an "it's loading, I promise" indicator. Nice showcase of Papyrus, however.

Can confirm this PR fixes my stupidity. It leaves some blank pages but the alternatives aren't much better when the user is asking for the impossible.

tommyettinger:
I'll get rid of the splash screen commented code, Splash class, and image then.

Issue #7726, opened 2025-11-23
[Tiled] Wrong GIDs in Tiled template object instances


damnedpie:
There's an issue with how Templates are currently processed.

A Tilemap can embed links to multiple Tilesets; it will refer to their particular Tiles via GIDs. GIDs can't repeat throughout a single given Map, so they will get rebound by the Map like this:

```xml
 <tileset firstgid="1" source="../../tilesets/auto_block.tsx"/>
 <tileset firstgid="65" source="../../tilesets/dungeon.tsx"/>
 <tileset firstgid="170" source="../../tilesets/dungeon_decorations.tsx"/>
 <tileset firstgid="238" source="../../tilesets/dungeon_animations.tsx"/>
 <tileset firstgid="255" source="../../tilesets/lights.tsx"/>
```

A Template, however, will not do such rebinding. It would just point to a Tileset and specify the GID for the texture in that particular Tileset.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<template>
 <tileset firstgid="1" source="../../tilesets/dungeon_decorations.tsx"/>
 <object name="CageSkeleton" gid="47" width="207" height="358"/>
</template>
```

A TMX map won't specify any GIDs for template-instantiated objects unless there are any flipping overrides on those instances, so usually those instances will look like this in a TMX (no GIDs, everything relies purely on info stored in a Template):

```xml
 <objectgroup id="14" name="DecoStatic">
  <object id="52" template="../../templates/dungeon/flaming_torch.tx" x="481" y="264.667"/>
  <object id="69" template="../../templates/dungeon/puddle.tx" x="700.667" y="686"/>
  <object id="70" template="../../templates/dungeon/column_inwall.tx" x="640" y="384"/>
  <object id="72" template="../../templates/dungeon/puddle.tx" x="896" y="704"/>
 </objectgroup>
 <objectgroup id="20" name="DecoYSort">
  <object id="64" template="../../templates/dungeon/cage_skeleton.tx" x="79" y="532"/>
 </objectgroup>
```

So, when BaseTmxMapLoader#mergeParentElementWithTemplate() happens, the GID shifting is not accounted for, and if the TMX had no data for GID for a particular template instance (which will usually be the case), the GID is going to be relative **to the Tileset** the Template refers to, **not the map**.

[...]

PR #7742, opened 2025-12-06
Tiled Map Template Fixes,  Proper handling of GIDs from multiple Tilesets. Fix for issue #7726

BoBIsHere86:
There was an issue with Templates not taking into account the proper GIDs when multiple tilesets were used. #7726 

Implemented the fixes suggested by @damnedpie, who submitted the issue. And also provided most of the suggested code to fix the issue. Many thanks! (if only we had more issue's like this)

It took me some time to replicate this issue using our own template object test. I found out that the issue could be replicated when bringing in a second tileset in to the template. So I was able to successfully break the test and implemented a fix to resolve it.
So our template test has been updated to pull in in the grass tileset for the sole purpose of having a texture load from a different tileset which would need the proper GID.
image

While doing this fix I also had to make sure to keep both the Tmj and Tmx map loaders to keep them in sync. As both the Tmx and Tmj loader ran into this issue during testing. So I updated both tmx and tmj template test maps as well as made the same changes to the BaseTmjMapLoader and BaseTmxMapLoader.

@damnedpie, I'm not sure if you are able to pull down these changes, build and test for yourself. But it would be very helpful to get feedback. To make sure your original use case works. And not just the one I cobbled together in the test.


damnedpie:
@BoBIsHere86 Hey and thanks a lot for bringing this fix into life! I would gladly test these changes in my current project but I am not very sure on how to do it properly. I mean, I would need to clone the libGDX repo and somehow make my project depend on it instead of official repositories, right? Is there any guide for me that explains how to do it? :D

I can be reached in TG if anything @damnedpie

crykn:
There is a guide on working with the libGDX source code on the website: https://libgdx.com/dev/from-source/

damnedpie:
@BoBIsHere86 Hey, cloned libGDX with this PR and can confirm this fix works as intended in my project with numerous Tiled templates!

BoBIsHere86:
> @BoBIsHere86 Hey, cloned libGDX with this PR and can confirm this fix works as intended in my project with numerous Tiled templates!

Awesome, thanks for taking the time to confirm.
 
I added one small update to the test maps, to keep the properties in sync. But other than that. I think this is good on my end.
😄

BoBIsHere86:
Just popping in so this bug PR doesn't get forgotten about. I think this is pretty much ready to go. The original user who submitted the issue tested and confirmed it as being fixed with the changes made here. 

I will ping @Quillraven just so another Tiled user can give a stamp of approval if needed.

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
```diff
@@ -65,6 +65,7 @@ public abstract class BaseTmjMapLoader<P extends BaseTiledMapLoader.Parameters>
 	protected JsonValue root;
 
 	protected ObjectMap<String, JsonValue> templateCache;
+	protected ObjectMap<String, Integer> tilesetPathsToGIDs;
```
tommyettinger: It's a minor improvement, but this could probably use `com.badlogic.gdx.utils.ObjectIntMap<String>` instead of `ObjectMap<String, Integer>`, because the values are never null. That would avoid some boxing and GC, though since it's only on load it likely doesn't matter at all. It might help a bit with any IDE annoyances about values being able to be null if they're Integer, but not if they are int, if that matters to you.

tommyettinger (dismissed):
It looks like good work was put in here; I haven't tested it but it sounds like other people have had good results from their tests. My one comment is really just a suggestion for something you could do if you wanted to, not a requirement; anything happening only on load doesn't really need to worry about GC.

Quillraven:
Sorry for late response. Will have a look over the weekend!

edit: I personally don't work with templates but the fix looks fine to me (codewise).

Also, not sure if related, but I needed a "tile by id" method in my last project and had to consider the tileset gid myself. Is this the intended way? Or is there a built-in LibGDX way to get the correct id automatically?

Here is my code:

```kotlin
fun tileById(tilesetName: String, id: Int): TiledMapTile {
        val tileSet = currentMap.tileSets.single { it.name == tilesetName }
        val tiledId = tileSet.property<Int>("firstgid") + id

        return tileSet.getTile(tiledId) ?: gdxError("There is no tile with id $id in tileset $tilesetName")
    }
```

gdx/src/com/badlogic/gdx/maps/tiled/BaseTmjMapLoader.java
BoBIsHere86: This is a good enough reason for me to change it. Though small, there is no reason NOT to implement this suggestion as it is nothing but a positive change. I have gone ahead and switched over to ObjectIntMap. Reran the template test just incase. All is well. 👍

BoBIsHere86:
> Sorry for late response. Will have a look over the weekend!
> 
> edit: I personally don't work with templates but the fix looks fine to me (codewise).
> 
> Also, not sure if related, but I needed a "tile by id" method in my last project and had to consider the tileset gid myself. Is this the intended way? Or is there a built-in LibGDX way to get the correct id automatically?
> 
> Here is my code:
> 
> ```kotlin
> fun tileById(tilesetName: String, id: Int): TiledMapTile {
>         val tileSet = currentMap.tileSets.single { it.name == tilesetName }
>         val tiledId = tileSet.property<Int>("firstgid") + id
> 
>         return tileSet.getTile(tiledId) ?: gdxError("There is no tile with id $id in tileset $tilesetName")
>     }
> ```

Sorry, I did not see your edit!

Sadly, yeah, I think that's basically what we're forced to do. There's no libGDX helper method for that which I recall. 😞

BoBIsHere86:
A recent Tiled commit change had made this PR unable to be auto merge,
The changes had refactored a few methods so I had to go in and manually merge the changes after a rebase.

All is completed and I have rerun the tests and everything is back to being ready again.

PR #7743, opened 2025-12-09
Calculate cachedOrientation for android KeyboardHeightObserver on startup

Berstanio:
Small followup fix for https://github.com/libgdx/libgdx/pull/7670

This ensures, that the first recognised keyboard change is really a change in values, and not some android internal relayout.


obigu (approved):
Tested on Android Gdx Tests

Issue #7745, opened 2025-12-27
TextField's undo function (Ctrl + Z) does not fire ChangeEvent


Hyperion2400:
### Describe the bug

The TextField's undo function does not fire a change event. It's clear when comparing the undo case with paste and cut where fireChangeEvent is explicitly handled.

https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L1048

It seems like an easy fix but it's a core component and the code in that class is a bit convoluted (undo calls setText() which calls paste() and in some cases changeText() 🤷), so I don't feel confident doing it myself as someone who hasn't worked on the project at all yet.

### Reproduction

You can reproduce this using gdx-liftoff by triggering a validation error in the first panel and then undoing the change.
The validation error will not be reset. It's not a gdx-liftoff bug. Pasting and cutting trigger the ChangeEvent just fine.

No validation error:
<img width="455" height="194" alt="Image" src="https://github.com/user-attachments/assets/0e53e7fc-4d15-45ee-a6a8-bc05ff9ffb0d" />

Remove project name:
<img width="459" height="192" alt="Image" src="https://github.com/user-attachments/assets/5adffbca-8f6c-42ee-84eb-99e45dc3a789" />

Ctrl + Z:
<img width="455" height="180" alt="Image" src="https://github.com/user-attachments/assets/8021bfb1-7808-4014-83a3-ddbc183ddb37" />

### Stack trace

```shell

```

### Affected platforms

- [x] Android
- [x] iOS
- [x] HTML/GWT
- [x] Windows
- [x] Linux
- [x] macOS

bluemeaford:
Thanks — this seems like a genuine event consistency issue: paste/cut fire ChangeEvent, but undo changes the text without firing it.  ￼
If maintainers agree undo should behave like other user edits, I can open a PR that ensures a ChangeEvent is fired when undo results in an actual text change, plus a small regression test.

PR #7746, opened 2025-12-30
Fix TextField and TextArea undo do not fire ChangeEvent

Billadom0123:
A tiny fix for issue [#7745](https://github.com/libgdx/libgdx/issues/7745#issue-3764677374)

Current undo shortcut ( CTRL + Z ) in TextField and TextArea only calls `setText()` function, and will not fire any ChangeEvent.

This commit fixes this by using `changeText()` function, which will fire a ChangeEvent. Concerning different ways to execute undo behavior, this commit also wraps it into a method `undo()` and gives a boolean param `fireChangeEvent`, just like what `paste()` and `delete()` do.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -19,6 +19,7 @@ package com.badlogic.gdx.scenes.scene2d.ui;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Input;
 import com.badlogic.gdx.Input.Keys;
+import com.badlogic.gdx.Input.OnscreenKeyboardType;
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.g2d.Batch;
 import com.badlogic.gdx.graphics.g2d.BitmapFont;
@@ -111,7 +112,7 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 	protected int visibleTextStart, visibleTextEnd;
 	private int maxLength;
 	private String[] autocompleteOptions;
-	private Input.OnscreenKeyboardType keyboardType = Input.OnscreenKeyboardType.Default;
+	private OnscreenKeyboardType keyboardType = OnscreenKeyboardType.Default;
 	private boolean preventAutoCorrection;
 
 	boolean focused;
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -542,6 +542,16 @@ int delete (boolean fireChangeEvent) {
 		return minIndex;
 	}
 
+	void undo (boolean fireChangeEvent) {
```
tommyettinger, line 562: Is there a reason this isn't public? I would think users might want to perform `undo()` programatically sometimes. I think using package-private by default is a mistake, anyway, for both Java and libGDX, since it doesn't actually provide any kind of prevention from use, it just makes it more cumbersome to use and means people think they don't have to document code...

tommyettinger (commented):
The changes seem good, I just would think undo() should probably be public to allow users to programatically undo changes, such as if another keyboard shortcut is wanted, or a button for controller usage.

gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
tommyettinger: Note that for cut() and paste(), there's a public version as well as the package-private implementation: https://github.com/libgdx/libgdx/blob/545467102fd3782cf11ad0c6ef25567fe0f6c2ad/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L484-L496

Billadom0123:
Thank you for the review! I've updated the PR to make a public version of `undo()`. While implementing this, I actually looked into the original code and find some cursor boundary problem. This was also fixed in my latest commit.

While mirroring the `cut()` and `paste()` pattern, I discovered a potential state desynchronization situation in the existing implementation: when a `ChangeEvent` is cancelled, the text reverts but the cursor still moves, causing a cursor **drift**.

Is this a bug or an intended behavior?

tommyettinger (dismissed):
Looks good to me! The cursor drift seems hard to avoid without tracking the cursor state just before the text last changed. Clamping, as you have it, seems like the simplest and sanest option to avoid the cursor going out of bounds.

Billadom0123:
It seems like the MetalANGLE check failed due to a 403 Forbidden error while resolving dependencies (Gradle 8.14.3), is it a network issue or my code problem?

Billadom0123:
> Looks good to me! The cursor drift seems hard to avoid without tracking the cursor state just before the text last changed. Clamping, as you have it, seems like the simplest and sanest option to avoid the cursor going out of bounds.

yeah, I made a small fix to the `undo()` method, but `cut()` and `paste()` also have the "cursor drift" problem. I can open a issue for this problem if you are interested

tommyettinger:
> It seems like the MetalANGLE check failed due to a 403 Forbidden error while resolving dependencies (Gradle 8.14.3), is it a network issue or my code problem?

Sounds like it might be a network issue on Maven Central's end, or maybe the server that supplies the natives, nothing has changed recently with the dependencies for ANGLE or MetalANGLE to my knowledge...

EDIT: Yeah, it was some brief blip. I re-ran the build and nothing was forbidden this time.

obigu:
Looks like a good change, @NathanSweet want to take a look before merging?

NathanSweet:
Thanks @obigu. It looks good! We could clamp the cursor for cut and paste in this PR.

obigu:
> Thanks @obigu. It looks good! We could clamp the cursor for cut and paste in this PR.

Cool, would you push those changes directly or give @Billadom0123 some time see if he finds a moment to do it? Otherwise I'd merge as it is and that can be added on a separate PR.

NathanSweet:
Hacked on it a bit, see commit message. Seems good to merge to me, but ya'll may want to review.

Billadom0123:
Oops. I forgot to check my mails, sorry to be late.

> Cool, would you push those changes directly or give @Billadom0123 some time see if he finds a moment to do it? Otherwise I'd merge as it is and that can be added on a separate PR.

I fixed that in #7747. And I am not pretty sure if the logic problem in #7747 is fixed in this PR...
If it is fixed, then #7747 can be closed.

PR #7747, opened 2026-01-03
Fix TextField and TextArea cursor drift and unreachable logic on event cancellation

Billadom0123:
This issue was identified during the implementation of the `undo()` feature in #7746.

Currently, `TextField#cut()`, `TextField#paste()`, and `TextField#delete()` update the cursor and selection positions before ensuring the text change is actually accepted. If a `ChangeListener` cancels the event, the text reverts to its original value, but the cursor remains at the "drifted" position.

This PR ensures that internal state updates (cursor, selection) only occur if `changeText()` returns true.

Changes are below:
`delete(boolean)`: Modified to return the current cursor position if the change is cancelled, and skip `clearSelection()`.
`cut()`: Inherits the fix from the updated `delete()` logic.
`paste(String, boolean)`: Added a check for the `changeText()` return value before updating the cursor position.

This PR also fixed a logical bug in:
https://github.com/libgdx/libgdx/blob/9f5ddf1dae56cabc9c6c8feb62eea7bed7f0c91c/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L1210-L1216

where the `else if` branch will never be reached. (because `changeText()` returns false and `text` reverts to `oldText`, thus `.equals()` always returns **true**) 

I have verified this fix basically using a custom GdxTest. When a listener calls `event.cancel()`, the text, cursor, and selection now remain perfectly synchronized at their original positions.

Here's the test code:
<details>
<summary>TextAreaTest4.java</summary>

``` java
// Imports omitted...

public class TextAreaTest4 extends GdxTest {
    private Stage stage;
    private Skin skin;
    private boolean cancelled = false;

    @Override
    public void create () {
        stage = new Stage();

        final TextArea textArea = new TextArea("Line 1\nLine 2\n0123456789\nABCDEFGHIJ", skin);
        textArea.setBounds(20, 20, 300, 200);

        textArea.addListener(new ChangeListener() {
            @Override
            public void changed(ChangeEvent event, Actor actor) {
                if (cancelled) {
                    event.cancel();
                    Gdx.app.log("Test", "ChangeEvent Cancelled.");
                }
            }
        });

        Table table = new Table();
        table.setFillParent(true);
        table.top().right().pad(20);

        final CheckBox cancelToggle = new CheckBox(" Cancel Event", skin);
        cancelToggle.addListener(new ChangeListener() {
            @Override
            public void changed(ChangeEvent event, Actor actor) {
                cancelled = cancelToggle.isChecked();
            }
        });

        final Label statusLabel = new Label("", skin);

        table.add(cancelToggle).left().row();
        table.add(new Label("--- Status ---", skin)).padTop(10).row();
        table.add(statusLabel).left();

        stage.addActor(textArea);
        stage.addActor(table);

        stage.getRoot().addCaptureListener(new InputListener() {
            @Override
            public boolean mouseMoved(InputEvent event, float x, float y) {
                updateStatus(textArea, statusLabel);
                return false;
            }
            @Override
            public boolean keyTyped(InputEvent event, char character) {
                updateStatus(textArea, statusLabel);
                return false;
            }
        });
    }

    private void updateStatus(TextArea textArea, Label label) {
        String info = "Cursor: " + textArea.getCursorPosition() + "\n"
                + "Selection: " + textArea.getSelectionStart() + "\n"
                + "Text Len: " + textArea.getText().length();
        label.setText(info);
    }

    @Override
    public void render () {
        ScreenUtils.clear(0.2f, 0.2f, 0.2f, 1);
        stage.act(Gdx.graphics.getDeltaTime());
        stage.draw();
    }

    @Override
    public void resize (int width, int height) {
    }

    @Override
    public void dispose () {
        stage.dispose();
        skin.dispose();
    }
}
```
</details>

Note: The `undo()` method is not included in this PR because the fix was already integrated into #7746.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -705,6 +705,7 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 		text = "";
 		paste(str, false);
 		if (programmaticChangeEvents) changeText(oldText, text);
+		undo.clear();
 		cursor = 0;
 	}
 
@@ -1362,21 +1363,13 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 		/** Saves the current state for redo, moves back, and returns the undo target state. */
 		UndoState undo (String text, int cursor, boolean hasSelection, int selectionStart) {
 			storeCurrent(text, cursor, hasSelection, selectionStart);
-			return states.get(index--);
+			return states.get(--index);
 		}
 
 		/** Saves the current state for undo, moves forward, and returns the redo target state. */
 		UndoState redo (String text, int cursor, boolean hasSelection, int selectionStart) {
 			storeCurrent(text, cursor, hasSelection, selectionStart);
 			return states.get(++index);
-		}
-
-		void cancelUndo () {
-			index++;
-		}
-
-		void cancelRedo () {
-			index--;
 		}
 
 		private void storeCurrent (String text, int cursor, boolean hasSelection, int selectionStart) {
@@ -1391,6 +1384,20 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 			state.cursor = cursor;
 			state.hasSelection = hasSelection;
 			state.selectionStart = selectionStart;
+		}
+
+		void cancelUndo () {
+			index++;
+			if (index == states.size - 1) states.removeLast();
+		}
+
+		void cancelRedo () {
+			index--;
+		}
+
+		void clear () {
+			states.clear();
+			index = 0;
 		}
 	}
 


Billadom0123:
Seems that logical bug was brought in Commit https://github.com/libgdx/libgdx/commit/2616ac7a5d66d6f2cf06a2ae9330eaf1f237f04e , which tried fixing a rollback problem. But sadly it's just made the code unreachable to achieve it.
In the latest Commit I updated it to check `.equals() || changeText()` in the `if` branch, ensuring the cursor would rollback only when the text are not same but `changeText()` still returns `false`. So even type a identical character with selection, the cursor still goes as intended.

Billadom0123:
> I added the test you wrote, and it seems to work just fine for cut, copy and paste. As expected, it is waiting on that other PR to work for undo as well. The tests have a little weird behavior because the status label doesn't update when the cursor moves, but does when the mouse moves. That's fine with me, though.

Sorry for that because the main part of the test code was generated by AI and it doesn't always handle with the intricacies of the render loop.

In the latest commit I change it to put the update logic in `render()`, so the text would automatically update.

NathanSweet:
Sorry for the conflicts, we tried to squeeze in these additional fixes into #7746. It's a bit trickier than it looks.

Testing this PR, it has these issues that don't occur in HEAD:
* Undo and redo resets cursor position to 0 (uncancelled).
* Undo and redo still occur when cancelled.

Using tests from this PR with HEAD (after #7746) the issues I found are:
* When cancelled, typing or backspace still moves the cursor.
* When cancelled, the selection is lost by all actions that would have modified the text: typing, cut, delete, paste, undo, redo, backspace.

Would you like to sync this branch with HEAD and fix these?

Billadom0123:
> Sorry for the conflicts, we tried to squeeze in these additional fixes into #7746. It's a bit trickier than it looks.
> 
> Testing this PR, it has these issues that don't occur in HEAD:
> 
> * Undo and redo resets cursor position to 0 (uncancelled).
> * Undo and redo still occur when cancelled.
> 
> Using tests from this PR with HEAD (after #7746) the issues I found are:
> 
> * When cancelled, typing or backspace still moves the cursor.
> * When cancelled, the selection is lost by all actions that would have modified the text: typing, cut, delete, paste, undo, redo, backspace.
> 
> Would you like to sync this branch with HEAD and fix these?

Thank you @NathanSweet for all of the tests!

In the latest commit I fixed those bugs and tested, and no weird behavior was found now.
This commit also fixes one bug where `cut()` , `delete()`, `paste()` do not set `undoText` properly.

But here is still one problem.
> * When cancelled, the selection is lost by all actions that would have modified the text: typing, cut, delete, paste, undo, redo, backspace.

For this bug I do not fix the **typing** and **backspace** behavior, because that logic is in `keyTyped()`, and in that method the code calls `delete(False)`.

https://github.com/libgdx/libgdx/blob/7d5b1c95954536aeb2f31ecaf00b6426475939da/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L1221-L1223

https://github.com/libgdx/libgdx/blob/7d5b1c95954536aeb2f31ecaf00b6426475939da/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L1236-L1238

Even though in the last the code calls `changeText()` and the text is reverted, the selection always being cleared.

I am not pretty sure if it would be OK if just change these two `False`es to `True`s...

NathanSweet:
Your changes are nicer, kudos!

`delete(false)` -> `delete(true)` would give 2 change events (delete + text change), not good.

I added some fixes, can you please try it?
* Paste needs to advance cursor.
* Undo/redo keeps cursor position. My fix works but is complex, alternative is an `undoCursor` field.
* Restore selection/undo on cancelled key type.

Billadom0123:
Seems everything works fine. Thanks!

> * Undo/redo keeps cursor position. My fix works but is complex, alternative is an `undoCursor` field.

the newest commit fixed one condition: when `undo()` a delete operation, the cursor do not roll back as intended

NathanSweet:
What are repro steps for the problem the last commit wants to fix? It breaks this:

1. `abcd|efgh`
2. Perform undo or redo so `cdef` are deleted.

Without last commit: `ab|gh` (correct)
With last commit: `|abgh` (incorrect)

Billadom0123:
> What are repro steps for the problem the last commit wants to fix? It breaks this:
> 
> 1. `abcd|efgh`
> 2. Perform undo or redo so `cdef` are deleted.
> 
> Without last commit: `ab|gh` (correct) With last commit: `|abgh` (incorrect)

Sorry I am nor pretty sure how to actually delete `cdef` when the cursor is between `d` and `e` with `undo` and `redo`, could you please give me a repro step, thanks.

but if you use `abcdefgh|` and delete to `abcd|`
    without last commit it would be `abcd|efgh`
    with the last commit it would be `abcdefgh|`


EDIT:
hmm, seems the latest commit brings another bug
when text changed and moved the cursor then perform `undo`, the cursor do not really go back.

> Without last commit: `ab|gh` (correct) With last commit: `|abgh` (incorrect)

I guess that's the bug you want to mean.
Maybe still need some fix...

Billadom0123:
Add `undoCursor` now, and basic tests were OK.

NathanSweet:
1. `abcdefgh` select `cdef` and delete, then undo.
2. Place cursor at `abcd|efgh`.
3. Redo the delete to remove `cdef`.

> but if you use abcdefgh| and delete to abcd|
> without last commit it would be abcd|efgh

By "delete" you must mean "backspace to delete and then undo". I agree that behavior is not quite right. I thought maybe it could be good enough without a field, but now I agree a field is the only way to restore exactly what was there.

`undoCursor = maxIndex;` looks like a bug, should be `undoCursor = to;`. I added also saving the selection, and fixed a bug where pressing delete on an empty text field nuked the undo state.

What a pain in the ass, and it's not pretty. It should work right at least, please confirm.

To improve it further we could store multiple undo states, but maybe it's ugly enough for now. ;)

Billadom0123:
Tested, and seems everything works pretty fine now, thanks!

> To improve it further we could store multiple undo states, but maybe it's ugly enough for now. ;)

Would it be suitable to wrap those `undo` related fields into a class?
as it would make the code much cleaner, and would be much easier to do if we want multi-level undo in the future.

NathanSweet:
Yes, but where does it stop!? 😭

Please give that a try.

Billadom0123:
OH ITS MULTILEVEL UNDO!

> Yes, but where does it stop!? 😭

I guess we may squeeze too many things in this PR now ;)
Anyway no bugs have been found in the latest commits.
Thanks a lot @NathanSweet !

PR #7504, opened 2024-11-07
fix: emulating backspace keypress on DefaultGwtInput

itsFufie:
Fixes #7219 by properly emulating the BACKSPACE key press following the same method used for TAB key press

Issue #7219, opened 2023-08-29
Why the BACKSPACE key is treated differently in libGDX/GWT than in libGDX/LWJGL ? 


PoulpoGaz:
#### Issue details
In libGDX, I use `Gdx.input.isKeyPressed(Input.Keys.BACKSPACE)` to know when the use push BACKSPACE.
But, results are different depending on the backend. In libgGDX/LWJGL, when I press BACKSPACE, the above expression returns true but in libgGDX/GWT, it always returns false. After searching in code, I found the following lines in libGDX: [DefaultGwtInput line 699](https://github.com/libgdx/libgdx/blob/6fbf3abe28dcfa55b81a9c3dd2817d4f5f3c96ad/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java#L699) where I can clearly see that BACKSPACE key is treated differently and that `Gdx.input.isKeyPressed(Input.Keys.BACKSPACE)` will never returns true. In libGDX/LWJGL there is no such difference.
Why the BACKSPACE key is treated differently in libgGDX/GWT than in libGDX/LWJGL ? 

#### Reproduction steps/code
[test.tar.gz](https://github.com/libgdx/libgdx/files/12466637/test.tar.gz)
Download and untar `test.tar.gz`.
The program is just a black window. An `InputProcessor` is attached to Gdx.input. When BACKSPACE is pressed, an event is triggered, InputProcessor#keyDown is called and I print '[DEBUG] DOWN'. Then, the render method is called multiple times, resulting in printing multiple times `[DEBUG] PRESSED` (using Gdx.input.isKeyPresssed). Finally, when BACKSPACE is released,  InputProcessor#keyUp is called and I print '[DEBUG] UP'. This is the expected output.

On a desktop, the output is:
```
[DEBUG] DOWN
[DEBUG] PRESSED       (repeated multiple times)
[DEBUG] UP
```
which is expected. 
With the HTML version the output is:
```
[DEBUG] DOWN           (repeated multiple times)
[DEBUG] UP
```

#### Version of libGDX and/or relevant dependencies
libGDX version 1.12.0
GWT version 2.10.0

#### Please select the affected platforms
- [ ] Android
- [ ] iOS
- [x] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

crykn:
[...]

PR #7748, opened 2026-01-05
A draft fix for backspace and tab keypress on DefaultGwtInput

Billadom0123:
After checking #7219 and #7504 , seems the most urgent need of the problem is to make gwt's `TAB` and `BACKSPACE` response totally the same as the one in lwjgl/lwjgl3. But this is hard to tackle because most of the browsers won't send `keypressed` event for functional keys like `TAB` and `BACKSPACE`.

This PR tries to fix this by listening the `keydown` event and use `e.repeat` to distinguish whether it's the initial press or the auto-repeat. If this is a "initial press" we do nothing as a normal `keydown` event; if this is a "auto-repeat", we dispatch a `keypress` event for the key instead. So that `TAB` and `BACKSPACE` could also be captured as usual keys.

I have basically verified it on my pc, and it works well. These two keys are captured just like those in lwjgl3.

Existing problems/questions:
- Apparantly this solution breaks the original logic so they may need more tests.
- though `repeat` property is commonly used in nowaday browsers, there are still edge cases in legacy embedded webviews or older mobile browsers where this property might be missing. I just skipped the change so I don't know how those will behave. Maybe need further changes.
- need some tests in different devices and envs (Sorry i still don't know how to test these on a Android device like in #7504 )

Request Changes are welcome.

### DIFF
--- a/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
+++ b/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
@@ -788,35 +788,15 @@ public class DefaultGwtInput extends AbstractInput implements GwtInput {
 						processor.keyDown(code);
 					}
 				}
-// if (code == Keys.BACKSPACE) {
-// if (processor != null) {
-// processor.keyDown(code);
-// processor.keyTyped('\b');
-// }
-// } else {
-// if (!pressedKeys[code]) {
-// pressedKeySet.add(code);
-// pressedKeyCount++;
-// pressedKeys[code] = true;
-// keyJustPressed = true;
-// justPressedKeys[code] = true;
-// if (processor != null) {
-// processor.keyDown(code);
-// }
-// }
-// }
 			}
 
 			if (e.getType().equals("keypress")) {
 				// Gdx.app.log("DefaultGwtInput", "keypress");
 				char c = (char)e.getCharCode();
-				// usually, browsers don't send a keypress event for tab, so we emulate it in
-				// keyup event. Just in case this changes in the future, we sort this out here
-				// to avoid sending the event twice.
+				// usually, browsers don't send a keypress event for tab and backspace.
+				// but the capture listener creates those events, so here just handle
+				// them as normal keys.
 				if (processor != null) processor.keyTyped(c);
-// if (c != '\t') {
-// if (processor != null) processor.keyTyped(c);
-// }
 			}
 
 			if (e.getType().equals("keyup")) {
@@ -825,11 +805,6 @@ public class DefaultGwtInput extends AbstractInput implements GwtInput {
 				if (isCatchKey(code)) {
 					e.preventDefault();
 				}
-// if (processor != null && code == Keys.TAB) {
-// // js does not raise keypress event for tab, so emulate this here for
-// // platform-independant behaviour
-// processor.keyTyped('\t');
-// }
 				if (pressedKeys[code]) {
 					pressedKeySet.remove(code);
 					pressedKeyCount--;


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```diff
@@ -731,23 +778,33 @@ private void handleEvent (NativeEvent e) {
...
-						pressedKeyCount++;
-						pressedKeys[code] = true;
-						keyJustPressed = true;
-						justPressedKeys[code] = true;
-						if (processor != null) {
-							processor.keyDown(code);
-						}
 					}
 				}
+// if (code == Keys.BACKSPACE) {
```
tommyettinger: All of the commented code is in Git history, it shouldn't be present in the source code as a comment if it doesn't help understand the current code.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```diff
@@ -756,9 +813,10 @@ private void handleEvent (NativeEvent e) {
 				// usually, browsers don't send a keypress event for tab, so we emulate it in
 				// keyup event. Just in case this changes in the future, we sort this out here
 				// to avoid sending the event twice.
-				if (c != '\t') {
-					if (processor != null) processor.keyTyped(c);
-				}
+				if (processor != null) processor.keyTyped(c);
+// if (c != '\t') {
```
tommyettinger: This commented code should be removed too, BUT the documentation comment above it is now inaccurate -- that should be updated to no longer mention special treatment for tab.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/DefaultGwtInput.java
```diff
@@ -767,11 +825,11 @@ private void handleEvent (NativeEvent e) {
 				if (isCatchKey(code)) {
 					e.preventDefault();
 				}
-				if (processor != null && code == Keys.TAB) {
-					// js does not raise keypress event for tab, so emulate this here for
-					// platform-independant behaviour
-					processor.keyTyped('\t');
-				}
+// if (processor != null && code == Keys.TAB) {
```
tommyettinger: This commented code can just be removed.

tommyettinger (changes requested):
I'm not entirely familiar with the problem this solves, but tab and backspace being different in a browser setting isn't that surprising. Tab can switch focus to the browser window itself for accessibility reasons, and backspace at least sometimes is used to go to the previous page, though Chrome hasn't used it for that in a while by default.

As long as tab and backspace don't suddenly change behavior for users who are used to, say, tab being used to select links for visually impaired users, it's probably an improvement to have them using similar code to the rest of the key events.

It's considered sloppy practice to leave in commented code in a Git repo. (Even though I do it all the time! Do as I say, not as I do!) The Git history tracks what the full code was at the time the commented code would have been used, and `git blame` can get to the commit where a change was made or before it. Code that isn't currently relevant should just get removed from the file.

Billadom0123:
> As long as tab and backspace don't suddenly change behavior for users who are used to, say, tab being used to select links for visually impaired users, it's probably an improvement to have them using similar code to the rest of the key events.

Yeah in the code I just changed the sent `keydown` event for `TAB` and `BACKSPACE`, and won't change the original browser behavior 

> It's considered sloppy practice to leave in commented code in a Git repo. (Even though I do it all the time! Do as I say, not as I do!) The Git history tracks what the full code was at the time the commented code would have been used, and `git blame` can get to the commit where a change was made or before it. Code that isn't currently relevant should just get removed from the file.

I commented but not delete those code mainly because they may still be used at **edge cases** (as I said before), and I'm still not sure if they should be totally deleted. Considering what you said, I think they can be added back if they're still needed. I'll delete them for now.

Issue #7727, opened 2025-11-25
Dispose not called on Android backend


obigu:
On Android, `ApplicationListener#dispose` is not called when closing/killing the app from the Recents Screen. It is called if using the old school Back button though.

This can be a problem if users add critical resource freeing login on `dipose` trusting it will always be called.

On Android, the `ApllicationListener` lifecycle calls are managed by the GLThread on `AndroidGraphics` and are not direct calls from `Activity` `onResume`, `onPause` and `onDestroy`. This is required not to dispose OpenGL context stuff on the main thread (instead of the GLTrhead).

Best workaround to relase critical resources on Android is to implement `onDestroy()` (or `onStop()` in your `AndroidApplication` and dispose them there (bear in mind in theory the OS can potentially skip `onDestroy` under extreme conditions) .

PR #7749, opened 2026-01-06
Fix ApplicationListener#dispose call on Android backend

Kasl0:
Fix for issue [#7727](https://github.com/libgdx/libgdx/issues/7727).

I basically moved
```
graphics.clearManagedCaches();
graphics.destroy();
```
block from `AndroidApplication#onPause()` to `AndroidApplication#onDestroy()` method.

As mentioned in the issue description `AndroidGraphics#destroy()` manages `ApllicationListener` lifecycle calls, but it also manages `LifecycleListener` calls. There is no need to change logic of that, but I believe `AndroidGraphics#destroy()` call is placed in a wrong method. To make it work like described in the issue there is a need to initiate disposing in `AndroidApplication#onDestroy()`. Also we have to remove `graphics.destroy()` from `AndroidApplication#onPause()`, because that method would be called twice when pressing "Back" key.

The solution was tested on the demo application on Android emulator and `ApplicationListener#dispose()` is now executed both after pressing the “Back” button and after closing/killing the app from the Recents Screen.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidApplication.java
@@ -248,11 +248,6 @@ public class AndroidApplication extends Activity implements AndroidApplicationBa
 
 		input.onPause();
 
-		if (isFinishing()) {
-			graphics.clearManagedCaches();
-			graphics.destroy();
-		}
-
 		AndroidGraphics.enforceContinuousRendering = isContinuousEnforced;
 		graphics.setContinuousRendering(isContinuous);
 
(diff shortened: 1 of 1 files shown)


obigu:
`onDestroy()` may not only be executed much later in the background but it may not even be executed at all, which is a change of behaviour that has the potential to break stuff. The change makes sense considering modern Android devices don't suffer from the memory issues old devices had but I don't think we can be sure changing this creates memory leaks or other issues due to the OpenGL context not having been destroyed on some devices.

This PR touches a very time-tested part of the libGDX lifecycle handling that was done like this for a reason and we can't change it without a lot of testing in real production scenarios with different devices which is always hard.

Related to this PR, even if, imo, libGDX Android backend lifecycle design has problems and would benefit from a refactor that makes `ApplicationListener` methods to be executed on Android `Activity` lifecycle methods (not just `onDestroy()` but all of them) and not linked to the internal GL Context management, I don't think it can be done without breaking backwards compatibility.

PR #7752, opened 2026-01-28
fix: prevent delta time from being negative

PokeMMO:
* some multicore devices have unstable clocks leading to inconsistent nanoTime values
* currentTimeMillis could change with clock synchronization
* LwjglGraphics is now more consistent with Lwjgl3Graphics

### DIFF
--- a/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/graphics/MockGraphics.java
+++ b/backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/graphics/MockGraphics.java
@@ -259,6 +259,7 @@ public class MockGraphics extends AbstractGraphics {
 
 	public void updateTime () {
 		long time = System.nanoTime();
+		if (lastTime == -1 || lastTime > time) lastTime = time;
 		deltaTime = (time - lastTime) / 1000000000.0f;
 		lastTime = time;
 
--- a/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
+++ b/backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
@@ -208,13 +208,14 @@ public class LwjglGraphics extends AbstractGraphics {
 	}
 
 	void updateTime () {
-		long time;
+		long time = System.nanoTime();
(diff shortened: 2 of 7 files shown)


tommyettinger (approved):
Looks good from here! GWT can use System.nanoTime() now, but it doesn't really need to. I suspect it's not nearly as accurate as nanosecond time on a desktop JVM, anyway; browsers intentionally skew timing results slightly to defeat timing attacks.

backends/gdx-backend-lwjgl/src/com/badlogic/gdx/backends/lwjgl/LwjglGraphics.java
```diff
@@ -208,13 +208,14 @@ public int getFramesPerSecond () {
 	}
 
 	void updateTime () {
-		long time;
+		long time = System.nanoTime();
+		if (lastTime == -1 || lastTime > time) lastTime = time;
```
crykn, line 212: The same is true here.

backends/gdx-backend-headless/src/com/badlogic/gdx/backends/headless/mock/graphics/MockGraphics.java
```diff
@@ -259,6 +259,7 @@ public boolean isFullscreen () {
 
 	public void updateTime () {
 		long time = System.nanoTime();
+		if (lastTime == -1 || lastTime > time) lastTime = time;
```
crykn, line 262: Unlike in Lwjgl3Graphics, `lastTime` here isn't initialized as -1, so we don't really need the check here.

PokeMMO:
> Generally looks good except two minor nitpicks, see the comments.
> 
> Also a question: wouldn't it be better to set `time` to `lastTime`, instead of setting `lastTime` to `time`, because the latter leads to double counting? (Except in the cases, where `lastTime` is -1, obviously).

One reason I didn't do it that way is that it could never recover if there is a one off highly out of range value. This is more applicable to currentTimeMillis, where a time sync would cause the game to freeze for the duration of the clock drift difference.

This issue I've observed is mostly resolving very small negative deltas in nanoTime() so the missed delta time doesn't have any realistic effect.

crykn:
Ah, yes, that's true! Then we should probably keep it the way it is 👍

PR #7753, opened 2026-02-01
Fix deprecated Android API usage in AndroidGraphics

### LINKED ISSUE #7066: AndroidGraphics.java use deprecated apis

njg7194:
## Summary
This PR fixes the deprecated Android API usage reported in #7066.

## Changes
- Replace `WindowManager.getDefaultDisplay().getMetrics()` with `context.getResources().getDisplayMetrics()` for getting display metrics (PPI, density)
- Replace `Display.getWidth()/getHeight()` with `WindowManager.getCurrentWindowMetrics().getBounds()` for API 30+ while maintaining backward compatibility for older API levels

## Deprecated APIs Fixed
- [WindowManager.getDefaultDisplay()](https://developer.android.com/reference/android/view/WindowManager#getDefaultDisplay()) - deprecated in API level 30
- [Display.getMetrics()](https://developer.android.com/reference/android/view/Display#getMetrics(android.util.DisplayMetrics)) - deprecated in API level 30

## Testing
Builds successfully and maintains backward compatibility with older Android versions through API level checks.

Fixes #7066

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -25,6 +25,7 @@ import android.opengl.GLSurfaceView.Renderer;
 import android.os.Build;
 import android.util.DisplayMetrics;
 import android.view.Display;
+import android.view.WindowMetrics;
 import android.view.DisplayCutout;
 import android.view.View;
 import android.view.WindowManager.LayoutParams;
@@ -151,8 +152,7 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 	}
 
 	protected void updatePpi () {
-		DisplayMetrics metrics = new DisplayMetrics();
-		app.getWindowManager().getDefaultDisplay().getMetrics(metrics);
+		DisplayMetrics metrics = app.getContext().getResources().getDisplayMetrics();
 
 		ppiX = metrics.xdpi;
 		ppiY = metrics.ydpi;
@@ -337,9 +337,15 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 
 		logManagedCachesStatus();
 
-		Display display = app.getWindowManager().getDefaultDisplay();
-		this.width = display.getWidth();
-		this.height = display.getHeight();
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+			WindowMetrics windowMetrics = app.getWindowManager().getCurrentWindowMetrics();
+			this.width = windowMetrics.getBounds().width();
+			this.height = windowMetrics.getBounds().height();
+		} else {
+			Display display = app.getWindowManager().getDefaultDisplay();
+			this.width = display.getWidth();
+			this.height = display.getHeight();
+		}
 		this.lastFrameTime = System.nanoTime();
 
 		gl.glViewport(0, 0, this.width, this.height);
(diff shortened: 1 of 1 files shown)


backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
```diff
@@ -820,3 +826,6 @@ public AndroidMonitor (int virtualX, int virtualY, String name) {
 		}
 	}
 }
+
+
+---SHA: 742f5b361f127b486b978c6bb132d65443e0f19a
```
tommyettinger, line 831: Pretty sure this wasn't intentional, but it makes me doubt it when you say this passes tests... this prevents compilation.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
```diff
@@ -337,9 +337,15 @@ public void onSurfaceCreated (javax.microedition.khronos.opengles.GL10 gl, EGLCo
 
 		logManagedCachesStatus();
 
-		Display display = app.getWindowManager().getDefaultDisplay();
-		this.width = display.getWidth();
-		this.height = display.getHeight();
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+			WindowMetrics windowMetrics = app.getWindowManager().getCurrentWindowMetrics();
+			this.width = windowMetrics.getBounds().width();
+			this.height = windowMetrics.getBounds().height();
```
Frosty-J, line 343: I think this is good, but we should be cautious because it yields different results. This includes cutout area, whereas the older code... well I don't know what the older code includes. On my device, I get 1459x720 for `Display` and 1600x720 for `WindowMetrics`. If I perform `Gdx.graphics.getWidth()`, I get 1600 with `config.renderUnderCutout = true` and 1555 with it false.

To be honest, I don't understand why libGDX sets width and height here at all - I can force them to silly values and that has no bearing on the tests.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
```diff
@@ -151,8 +152,7 @@ protected EGLConfigChooser getEglConfigChooser () {
 	}
 
 	protected void updatePpi () {
-		DisplayMetrics metrics = new DisplayMetrics();
-		app.getWindowManager().getDefaultDisplay().getMetrics(metrics);
+		DisplayMetrics metrics = app.getContext().getResources().getDisplayMetrics();
```
Frosty-J, line 155: This is good. It's unclear why libGDX ever used `getDefaultDisplay()`. Note that it is also used in `StandardKeyboardHeightProvider` and `DefaultAndroidInput`.

backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
obigu: I also think it's good even if I don't know if the slight differences in behaviour may affect some random device on some old Android version with some configuration.
Since the edge-to-edge enforcing (https://developer.android.com/develop/ui/views/layout/edge-to-edge), at least when running on Android 15+ up, users should expect to get maximum rendering area and adapt UI to insets manually using provided API, which from my understanding is consistent with this change.

PR #7758, opened 2026-02-17
model/Node id null/uninitialized change to empty string

pietru2004:
The method `getNode` crashes if Node id is not initialized or null.
I fought it might be good idea to have node id be empty string by default so if a another pice of code wants to do node.id.equals() it won't crash https://github.com/pietru2004/libgdx/blob/98f51139ceac623c70ecffd67ee704329a4363b4/gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java#L298

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
@@ -29,7 +29,7 @@ import com.badlogic.gdx.utils.GdxRuntimeException;
  * child nodes. Optionally a node can specify a {@link MeshPart} and a {@link Material} to be applied to the mesh part.
  * @author badlogic */
 public class Node {
-	/** the id, depending on use might be empty string or not unique string or unique string **/
+	/** The node's id; must not be null, and defaults to an empty String **/
 	public String id = "";
 	/** Whether this node should inherit the transformation of its parent node, defaults to true. When this flag is false the value
 	 * of {@link #globalTransform} will be the same as the value of {@link #localTransform} causing the transform to be independent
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
```diff
@@ -30,7 +30,7 @@
  * @author badlogic */
 public class Node {
 	/** the id, may be null, FIXME is this unique? **/
```
tommyettinger: If we're calling .equals on id, the above comment is incorrect, and id must *not* be null. Other than that, the PR looks fine to me!

tommyettinger (dismissed):
I'm encouraging you to update the comment just above the changed line so no one mistakenly thinks that id can be null. Other than that, good catch!

gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
```diff
@@ -29,8 +29,8 @@
  * child nodes. Optionally a node can specify a {@link MeshPart} and a {@link Material} to be applied to the mesh part.
  * @author badlogic */
 public class Node {
-	/** the id, may be null, FIXME is this unique? **/
-	public String id;
+	/** the id, depending on use might be empty string or not unique string or unique string **/
```
tommyettinger: The comment is now more confusing than before, heh... How about:
```java
/** The node's id; must not be null, and defaults to an empty String. **/
```

tommyettinger (changes requested):
Sorry to keep going back and forth on this, but the comment got a bit more confusing once the uniqueness part of the FIXME message was worked in. I don't think the FIXME message mattered...? I think we can just omit the part about uniqueness, since it seemed to not matter since this code was written.

gdx/src/com/badlogic/gdx/graphics/g3d/model/Node.java
pietru2004: Sounds good to me (Node ID PR)

tommyettinger (approved):
Looks good! Approved.

pietru2004:
updated via git rebase, and now I wonder is it better to do merge or rebase

PR #7759, opened 2026-02-17
ObjLoader multimaterial with one object

pietru2004:
Blender by default exports only one `o` mention per object, so if object has multiple materials, each time use mtl is used the previous material for that object will get overriden.

Also I have found this which I think is related to this topic.
https://dirsig.cis.rit.edu/docs/new/obj.html#_assigning_materials

This issue can be overcomed by using blender option `export_material_groups` but by default it is not checked.

This could be enchanced by making it set group as `objectName_material` which would kinda result to it working as blender option `export_material_groups`, but `objectName` would need to be stored somewhere.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
@@ -183,7 +183,7 @@ public class ObjLoader extends ModelLoader<ObjLoader.ObjLoaderParameters> {
 						activeGroup = setActiveGroup("default");
 					else {
 						String mat_name = tokens[1].replace('.', '_');
-						activeGroup = setActiveGroup(activeObject+"_"+mat_name);
+						activeGroup = setActiveGroup(activeObject + "_" + mat_name);
 						activeGroup.materialName = mat_name;
 					}
 				}
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
@@ -181,14 +181,14 @@ public class ObjLoader extends ModelLoader<ObjLoader.ObjLoaderParameters> {
 					if (tokens.length == 1)
 						activeGroup = setActiveGroup("default");
 					else {
-                        String mat_name = tokens[1].replace('.', '_');
-                        if (activeGroup.materialName.equals("default"))
-                            activeGroup.materialName = tokens[1].replace('.', '_');
-                        else {
-                            activeGroup = setActiveGroup(mat_name);
-                            activeGroup.materialName=mat_name;
-                        }
-                    }
+						String mat_name = tokens[1].replace('.', '_');
+						if (activeGroup.materialName.equals("default"))
+							activeGroup.materialName = tokens[1].replace('.', '_');
+						else {
+							activeGroup = setActiveGroup(mat_name);
+							activeGroup.materialName = mat_name;
+						}
+					}
 				}
 			}
 			reader.close();
--- a/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
@@ -114,6 +114,7 @@ public class ObjLoader extends ModelLoader<ObjLoader.ObjLoaderParameters> {
 
 		// Create a "default" Group and set it as the active group, in case
 		// there are no groups or objects defined in the OBJ file.
+		String activeObject = "default";
 		Group activeGroup = new Group("default");
 		groups.add(activeGroup);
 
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/g3d/loader/ObjLoader.java
```diff
@@ -179,9 +179,16 @@ protected ModelData loadModelData (FileHandle file, boolean flipV) {
 					mtl.load(file.parent().child(tokens[1]));
 				} else if (tokens[0].equals("usemtl")) {
 					if (tokens.length == 1)
-						activeGroup.materialName = "default";
-					else
-						activeGroup.materialName = tokens[1].replace('.', '_');
+						activeGroup = setActiveGroup("default");
+					else {
+						String mat_name = tokens[1].replace('.', '_');
```
noblemaster: Why the replacement? Why is `.` in the material name a problem?

Also,  `mat_name` is not camel-case.

noblemaster:
I also find the title "ObjLoader multimaterial with one object" and description of this merge request quite incoherent? That said, it's not my area of expertise? Is this AI slop? 

The other 2 merge requests are dubious also:  https://github.com/libgdx/libgdx/pull/7760 https://github.com/libgdx/libgdx/pull/7758

tommyettinger:
#7758 is a good change; `getNode()` really does call methods on a field that is not necessarily ever initialized, so is sometimes `null`. I think that had come up once before, but I can't find where...

#7760 is a bad change as it stands now, since it's a breaking change for no real reason, but if setBackBufferConfig() is brought back, it isn't breaking anymore, and is a nice QoL feature.

As for this one, I don't know a thing about the 3D model code, and it isn't my area of expertise either.

pietru2004:
I didn't use AI for those or to name those I was just not sure how to name stuff.

pietru2004:
discord msgs related to this https://discord.com/channels/348229412858101762/405518143818301443/1459616772071358677

pietru2004:
also here is a example where current obj loader would make error
image

Basicaly it would set active group as `Wall`
and when `usemtl` would be called 2 times and after second call it would change material name of that group instead adding new one

pietru2004:
changed to draft since I need to check something...

pietru2004:
Note: got confused why I was not seing material colors, long story short I was using modelbatch with not default shader and got confused

MonstrousSoftware:
As I understand it (see the Grouping section in [Paul Bourke](https://paulbourke.net/dataformats/obj/) ), `usemtl` is meant to affect the faces that follow it, not to set the material of the current group or object in the OBJ file, which
is a misunderstanding or shortcut in the original `ObjLoader` code.  The line ordering in an OBJ file is important.
 
So in a way each `usemtl` line should start a kind of subgroup corresponding to the concept of a LibGDX NodePart which links a MeshPart to a Material. (In practice, the ObjLoader creates a new Mesh, Node, NodePart and MeshPart for each group).

This PR seeks to address this by creating a new Group whenever `usemtl` is encountered. (Such a Group then no longer corresponds to what is called group in the OBJ file, but what I called a subgroup above).

There are still some issues in the PR however:

- `usemtl` is optional, if it does not appear a default white material should be used.  In the PR it however assumed there is always a `usemtl` statement 
otherwise no Group is created. In my view there should still be a call of setActiveGroup when processing a 'o' or 'g' as a fallback in case no material is specified. Empty groups will be discarded later anyway.

- if `usemtl` is used without a material name it is taken to use the default material for the next faces. 
In the PR, it (re-)activates a group called "default" which may already exist and the following faces will then be appended to that group. Instead it should create a new Group with some unique id and set its material name to "default".

pietru2004:
@MonstrousSoftware I assume `usemtl` as used since it is the way blender exported it...

Also the way ObjLoader works, the original from what I can see is it just overrides active groups materialName.

So what I did is I tried to fix it without making a too big change. Basicaly for each material use 

Can I correctly assume that activeGroup should hold list of materials that were added to it or more like `HashMap<MaterialNameString,ArrayList<FacesIntigers>>` or `HashMap<MaterialNameString,FacesData>` ?

MonstrousSoftware:
Blender is not the only source of OBJ files. Try to follow the file format specification (such as it is) and to anticipate edge cases (zero materials, 1 material, multiple materials, multiple groups, multiple objects, etc.).  It would also be useful to collect some test samples of OBJ files from different sources.

The original (which has big warnings in the comments on being non-compliant) indeed overrides the group's material name and this caused the issue you detected if an object has multiple materials.  

I think your approach is valid to create a new Group every time the material changes, the object changes or the (OBJ) group changes, so on every 'o' line, 'g' line or 'usemtl' line.  Groups without faces are skipped when generating nodes. The activeGroup would have 1 material, possibly the default material.

pietru2004:
Made a new test model... 3 diffrent objects, first having 1 material, second having 2 and third having 3
[test_boxes.zip](https://github.com/user-attachments/files/31355144/test_boxes.zip)

pietru2004:
I think now it should be better...
If model would have 2 diffrent materials it will create 2 ModelNodePart under model node...
Or it should... from testing I made till this point it did work...

MonstrousSoftware:
I've reviewed the code and tested it with some sample models and it seems to work fine.

Minor niggles on coding style consistency to match the existing code base: 
- line 185: the else clause doesn't need braces (the if clause doesn't have them either).
- variable `node_parts` should probably be in camelCase, i.e. `nodeParts`.

pietru2004:
at line 185 I added those braces since atleast when I was testing the code the java seamed to skip/ignore contents of that else... Mayby I have borked java or system specific bug IDK...
as for node_parts... done... ranamed var to nodeParts, althrou now I kinda wonder would it make it throu if I renamed it to ICBM... just for fun

PR #7760, opened 2026-02-17
Update Lwjgl3ApplicationConfiguration.java

pietru2004:
I think it might be good to split inLwjgl3ApplicationConfiguration the setBackBufferConfig method into setRGBA, setDepth, setStencil, setSample and add getters for RGBA, depth, stencil and sample.

I have in mind: What if you want to change one of those and not all of those at once ?

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
@@ -212,11 +212,6 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		this.samples = samples;
 	}
 
-	/** @return array with bit depths of red, green, blue, alpha, depth buffer, stencil buffer and number of MSAA samples */
-	public int[] getBackBufferConfig () {
-		return new int[] {this.r, this.g, this.b, this.a, this.depth, this.stencil, this.samples};
-	}
-
 	/** Set transparent window hint. Results may vary on different OS and GPUs. Usage with the ANGLE backend is less consistent.
 	 * @param transparentFramebuffer */
 	public void setTransparentFramebuffer (boolean transparentFramebuffer) {
@@ -228,20 +223,10 @@ public class Lwjgl3ApplicationConfiguration extends Lwjgl3WindowConfiguration {
 		this.idleFPS = fps;
 	}
 
-	/** @return polling rate during idle time in non-continuous rendering mode. */
-	public int getIdleFPS () {
-		return this.idleFPS;
-	}
-
 	/** Sets the target framerate for the application. The CPU sleeps as needed. Must be positive. Use 0 to never sleep. Default is
 	 * 0. */
 	public void setForegroundFPS (int fps) {
 		this.foregroundFPS = fps;
-	}
-
-	/** @return foreground framerate of the application. */
-	public int getForegroundFPS () {
-		return this.foregroundFPS;
 	}
 
 	/** Sets whether to pause the application {@link ApplicationListener#pause()} and fire


backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -159,25 +159,75 @@ public void setOpenGLEmulation (GLEmulation glVersion, int gles3MajorVersion, in
...
 	 * @param r red bits (default 8)
 	 * @param g green bits (default 8)
 	 * @param b blue bits (default 8)
-	 * @param a alpha bits (default 8)
-	 * @param depth depth bits (default 16)
-	 * @param stencil stencil bits (default 0)
-	 * @param samples MSAA samples (default 0) */
-	public void setBackBufferConfig (int r, int g, int b, int a, int depth, int stencil, int samples) {
+	 * @param a alpha bits (default 8) */
+	public void setRGBABits (int r, int g, int b, int a) {
```
tommyettinger, line 187: This would be a significant breaking change because the original, compatible `setBackBufferConfig(int, int, int, int, int, int, int)` method is removed here. It should be kept in the code, even if some users want to update only some fields. Making sure existing code doesn't break out of nowhere is important.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -159,25 +159,75 @@ public void setOpenGLEmulation (GLEmulation glVersion, int gles3MajorVersion, in
...
+	 * 
+	 * @param stencil stencil bits (default 0) */
+	public void setStencilBits (int stencil) {
 		this.stencil = stencil;
+	}
+
+	/** Sets the bit of multi-sampling.
+	 * 
+	 * @param samples MSAA samples (default 0) */
+	public void setSamplesBits (int samples) {
```
mgsx-dev: samples are not bits

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -159,25 +159,75 @@ public void setOpenGLEmulation (GLEmulation glVersion, int gles3MajorVersion, in
...
 		this.samples = samples;
 	}
 
+	/** @return bit of color depth r */
```
mgsx-dev: and it's not bit but a number of bits... are you sure what you do?

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
mgsx-dev: please review the whole file when apply changes

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
mgsx-dev: please make getter/setter documentation consistent

pietru2004: so originaly in setBackBufferConfig it was written as red/green/blue/alpha bits (I might have missed to write s after `bit` in setters) 
also for getters I was not 100% sure how I should write what it returns

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
pietru2004: re added `setBackBufferConfig(int, int, int, int, int, int, int)` method

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
Frosty-J: @pietru2004 I don't think getters are necessary here. Bit depths don't get changed once the application is already running, and the configuration isn't normally passed anywhere else, so I'm not sure what the use case for getters would be. Other fields such as fps lack respective getters. That's not to say getters would do any harm, just it would save the trouble of getting `getRGBA_RBit()` etc. to agreeable standards to not have getters.

mgsx-dev: i meant, getter says ''bit depth of...'' and setter says ''bits of....''.
i think bit depth for all these fields (except samples) is the right word

pietru2004: @mgsx-dev do you think it is good now ?

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -178,6 +178,60 @@ public void setBackBufferConfig (int r, int g, int b, int a, int depth, int sten
...
+	/** @return number of stencil buffer bits */
+	public int getStencilBits (int stencil) {
+		return this.stencil;
+	}
+
+	/** @return number of MSAA samples */
+	public int getSamplesBits (int samples) {
+		return this.samples;
+	}
+
```
tommyettinger: I agree with @Frosty-J that these getters can just be removed. They can't be accessed at runtime, unless the configuration was for some reason kept in the main ApplicationListener, which would force it to be Lwjgl3-only... Having setters is good, but we don't need these getters.

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -178,6 +178,60 @@ public void setBackBufferConfig (int r, int g, int b, int a, int depth, int sten
 		this.samples = samples;
 	}
 
+	/** Sets the bit depth of the color.
```
mgsx-dev: color buffer

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -178,6 +178,60 @@ public void setBackBufferConfig (int r, int g, int b, int a, int depth, int sten
...
+	 * @param b blue bits (default 8)
+	 * @param a alpha bits (default 8) */
+	public void setRGBABits (int r, int g, int b, int a) {
+		this.r = r;
+		this.g = g;
+		this.b = b;
+		this.a = a;
+	}
+
+	/** Sets the bit depth of depth.
```
mgsx-dev: depth buffer

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
```diff
@@ -178,6 +178,60 @@ public void setBackBufferConfig (int r, int g, int b, int a, int depth, int sten
...
+	 * 
+	 * @param depth depth bits (default 16) */
```
mgsx-dev, line 196: bit depth of depth buffer (to be consistent)

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
mgsx-dev: i agree :-)

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
pietru2004: should be correct now.. . I hope

backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java
pietru2004: can I just keep those as 1 method called `getBackBufferConfig()` ? I feel like there could be that 1 rare case where somebody might want to get that data...

PR #7762, opened 2026-02-21
Fix #7761 (Maven Central/Snapshot badges).

### LINKED ISSUE #7761: README badges not updating since Central Portal migration
Image

tommyettinger:
This updates the shields.io badges using a similar strategy to the one used by other Maven Central-targeting projects. Because the Sonatype Snapshots shield doesn't currently work, we use the Maven Metadata shield to get the latest snapshot version. The Snapshot link also didn't go to any version newer than 1.13.2-SNAPSHOT, which is I suppose the last one that went to oss.sonatype.org . I managed to figure this out in part by looking at how [kotest](https://github.com/kotest/kotest/blob/master/README.md?plain=1) does it in their README.md, but I didn't do things the same way exactly.

This looks like this now: 
Preview of new Shields

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 
 [![GitHub Actions Build Status](https://img.shields.io/github/actions/workflow/status/libgdx/libgdx/build-publish.yml?branch=master&label=GitHub%20Actions)](https://github.com/libgdx/libgdx/actions?query=workflow%3A%22Build+and+Publish%22)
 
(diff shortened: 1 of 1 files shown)


Frosty-J:
https://search.maven.org/artifact/com.badlogicgames.gdx/gdx could be changed to https://central.sonatype.com/artifact/com.badlogicgames.gdx/gdx or is it more robust to have it redirect?

The XML in the SNAPSHOT link isn't terrbily human-friendly, but looks like the best we have. Wowy fast PR.

tommyettinger:
> ...is it more robust to have it redirect?

I'm not sure, but it seems to work. Maven Central hasn't had an actually "robust" search since the change to the new Central Portal...

PR #7763, opened 2026-02-21
Make ModelBuilder fields and related classes protected for better backend extensibility

xpenatan:
Extending LibGDX for new graphics APIs (e.g. WebGPU) is difficult because many important fields in these classes are private. This forces large code duplication or hacks.

In the WebGPU backend, most of ModelBuilder and related code had to be copied just to adjust for webgpu.
Proposal: Change fields from private → protected.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/Mesh.java
+++ b/gdx/src/com/badlogic/gdx/graphics/Mesh.java
@@ -76,13 +76,13 @@ public class Mesh implements Disposable {
 	/** list of all meshes **/
 	static final Map<Application, Array<Mesh>> meshes = new HashMap<Application, Array<Mesh>>();
 
-	final VertexData vertices;
-	final IndexData indices;
-	boolean autoBind = true;
-	final boolean isVertexArray;
+	protected final VertexData vertices;
+	protected final IndexData indices;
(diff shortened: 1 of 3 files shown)


gdx/src/com/badlogic/gdx/graphics/Mesh.java
```diff
@@ -76,13 +76,13 @@ public enum VertexDataType {
 	/** list of all meshes **/
 	static final Map<Application, Array<Mesh>> meshes = new HashMap<Application, Array<Mesh>>();
```
tommyettinger, line 77: Any reason this one should stay package-private, while the rest are protected? I'm pretty firmly against package-private visibility in an open source library, since it doesn't actually hide or protect anything, it just makes it harder to use in your own code outside the library. And, I suppose, it allows docs to be omitted, but `meshes` already has a Javadoc line.

tommyettinger (approved):
I'm all in favor of being able to use libGDX as a library and not needing to copy-paste files out of it. When files must be copy-pasted, and there isn't any kind of common interface like Batch, there's no compatibility with the original class... Approved; even with the lack of documentation on protected fields, having them available to users who may need them is tremendously better than private fields.

PR #7764, opened 2026-02-22
Json can't read back CharArray in some cases.

tommyettinger:
Any CharArray containing the chars `'}'` or any digits from `'0'` to `'9'` won't be written correctly for round-trip reads by the minimal OutputType. The reads of JSON produced by libGDX 1.14.0 from any CharArray containing these chars will result in a SerializationException being thrown. I added several automated tests for this, and I believe I have it fixed now in `JsonWriter`. I know `Json` can now round-trip write and read CharArray instances containing these chars, but I can't vouch for `JsonString`, since I've never used it.

The fix considers the initial closing curly brace `'}'` invalid as a bare String even in minimal `OutputType`, where before, it was bare in minimal but quoted in other `OutputType`s. Digit chars presented a different problem; they are written as bare Strings, but are read back as JSON Number types, which get converted to `char` by their codepoint (as if being cast from an `int`). This means that the digit `'0'` would be written as the bare String `0` and read back as the same value as `'\u0000'`, not `'0'`. Similarly, the digit `'7'` is read as a tab char, `'\u0007'`. This fixes the case where the initial char in a String is a digit by quoting that String, so it won't be read as a Number.

There are better fixes available for CharArray, but I don't know of any for `char`s in general. A `CharArray` implements `CharSequence`, and has a constructor available that creates a `CharArray` from a `String`. `CharArray` could implement `Json.Serializable` and write itself as an object containing a String, and could read from that type of object to create itself from that String. In the tests, this goes from a JSON String with length=313 (using minimal mode, but quoting digits and the closing curly braces in addition to the chars it quoted before; this also writes every char in the CharArray's capacity, even ones greater than its size) to a JSON String with length=99 (though this isn't wrapped in an object). This isn't implemented here because I seem to recall `Json.Serializable` being discouraged from use, for unclear reasons. Json could also opt to serialize CharArray or CharSequence types in general as Strings with small wrapping objects.

This probably changes some output's ability to round-trip serialize and deserialize with the same exact JSON String in transit, if it used minimal mode before. It should only add quotes around questionably-bare Strings, so it shouldn't ever produce invalid minimal-mode JSON, though it certainly does without this PR. Other OutputTypes are unaffected.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/JsonValue.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonValue.java
@@ -373,11 +373,11 @@ public class JsonValue implements Iterable<JsonValue> {
 		case stringValue:
 			return stringValue.length() == 0 ? 0 : stringValue.charAt(0);
 		case doubleValue:
-			return (char)(doubleValue + '0');
+			return (char)(doubleValue);
 		case longValue:
-			return (char)(longValue + '0');
+			return (char)(longValue);
 		case booleanValue:
-			return longValue != 0 ? '1' : '0';
+			return longValue != 0 ? '\u0001' : '\u0000';
 		}
 		throw new IllegalStateException("Value cannot be converted to char: " + type);
 	}
@@ -636,7 +636,7 @@ public class JsonValue implements Iterable<JsonValue> {
 				v = (char)value.longValue;
 				break;
 			case booleanValue:
-				v = value.longValue != 0 ? (char)1 : 0;
+				v = value.longValue != 0 ? '\u0001' : '\u0000';
 				break;
 			default:
 				throw new IllegalStateException("Value cannot be converted to char: " + value.type);
--- a/gdx/src/com/badlogic/gdx/utils/JsonWriter.java
+++ b/gdx/src/com/badlogic/gdx/utils/JsonWriter.java
@@ -313,7 +313,8 @@ public class JsonWriter extends Writer {
 			if (this == OutputType.minimal && !string.equals("true") && !string.equals("false") && !string.equals("null")
 				&& !string.contains("//") && !string.contains("/*")) {
 				int length = string.length();
-				if (length > 0 && string.charAt(length - 1) != ' ' && minimalValuePattern.matcher(string).matches()) return string;
+				if (length > 0 && string.charAt(length - 1) != ' ' && !(value instanceof Character)
+					&& minimalValuePattern.matcher(string).matches()) return string;
 			}
 			return quote ? escapeQuote(string) : '"' + string + '"';
 		}
--- a/gdx/test/com/badlogic/gdx/utils/JsonTest.java
+++ b/gdx/test/com/badlogic/gdx/utils/JsonTest.java
@@ -25,11 +25,8 @@ public class JsonTest {
 	@Test
 	public void testCharFromNumber () {
 		Json json = new Json();
-		// Behavior here changed because minimal syntax doesn't distinguish between a bare String containing a digit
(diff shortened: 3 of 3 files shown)


tommyettinger:
So there are JsonMatcher tests that fail here. I don't know if they're using arrays of Strings or arrays of Numbers, but the numbers are getting quoted.

tommyettinger:
All tests pass now, but this had to change the behavior of asChar() when called on a long or double JsonValue. Because a bare String in minimal OutputType is indistinguishable from a JSON Number, we need to make sure a single-digit String and a single-digit long are parsed identically. That means JSON `"0"` and minimal JSON `0` both need to parse to the Java char `'0'`, which the latter didn't before. The major difference here is that any long or double JsonValue that was using asChar() before will now return what it would if it were a String, but only when it's a valid single digit. Otherwise, it returns a garbage char. Most longs and doubles don't really make sense as chars anyway; what would the char value of `-100000000000000L` or `1.2E100` be?

NathanSweet:
There are 2 issues here.

The change to quote values starting with `{` makes sense. I updated the javadoc and added final.

The change for JsonValue `asChar` is questionable. `asCharArray` would need to match, but it breaks the previous contract, eg 65 -> `A`. It treats the value as a digit index `value + '0'`, which only makes sense for digits. Maybe the thought was only JsonValues from JsonReader that look like numbers will be of type `doubleValue` or `longValue` and take this code path, but JsonValue could come from anywhere `new JsonValue(65).asChar() == 'q'`. It's surprising to convert a number to char in this way.

To be clear, this is primarily what we want to fix:
```
Json json = new Json();
String text = json.toJson('2');
Character out = json.fromJson(char.class, text);
System.out.println(text); // 2
System.out.println(out); // (char)2
```

The minimal format's quote omission loses whether a char is a string or a number. For a digit char, when reading it back we get what looks like a number. When interpreting that number as a char, the result is unexpected.

If we didn't lose quotes for a char it would read back as a string, then `asChar` would give the original char. I propose keeping `asChar` as it was and adding `&& !(value instanceof Character)` to Json.OutputType `quoteValue`:
```
if (length > 0 && string.charAt(length - 1) != ' ' && minimalValuePattern.matcher(string).matches()
	&& !(value instanceof Character)) return string;
```

With that `testCharFromNumber` needs `"\"9\""` to pass. It's fair to expect the right serialized form.

Then CharArray looks OK:
```
Json json = new Json();
CharArray in = CharArray.with('1','2','3');
String text = json.toJson(in);
CharArray out = json.fromJson(CharArray.class, text);
System.out.println(in.toString(",")); // 1,2,3
System.out.println(text); // {items:["1","2","3"],size:3}
System.out.println(out.toString(",")); // 1,2,3
```

There is no serializer for CharArray, so the unnecessary nesting and `size` isn't great, but that is to be expected. Every type being serialized needs to be considered and serializers provided if necessary. CharArray could serialize as a string, eg `"123"` above, but it seems uncommon to serialize char. We probably don't need to provide that OOTB and users should be able to do it themselves. We already don't try to cover everything.

Technically non-digit chars don't need quotes. I don't really mind either way, I don't think I've ever serialized a char.

tommyettinger:
> With that `testCharFromNumber` needs `"\"9\""` to pass. It's fair to expect the right serialized form.

Wait, then what's that testing then? That test was meant to check chars from numbers, no? And "9" is a String.

I'm fine with the instanceof Character test for quoting values. I think that's the only change needed, then?

NathanSweet:
> Wait, then what's that testing then?

The new behavior of digit char (as a string) to char.

> That test was meant to check chars from numbers, no?

True. The 90 -> Z test is reasonable behavior.

Looks good!

PR #7770, opened 2026-03-25
Added LongSet.

NathanSweet:
Identical to IntSet except for `hashCode`, which accumulates then folds into an int.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/LongSet.java
+++ b/gdx/src/com/badlogic/gdx/utils/LongSet.java
@@ -56,7 +56,7 @@ public class LongSet {
 	protected int shift;
 
 	/** A bitmask used to confine hashcodes to the size of the table. Must be all 1 bits in its low positions, ie a power of two
-	 * minus 1. If {@link #place(long)} is overriden, this can be used instead of {@link #shift} to isolate usable bits of a
+	 * minus 1. If {@link #place(long)} is overridden, this can be used instead of {@link #shift} to isolate usable bits of a
 	 * hash. */
 	protected int mask;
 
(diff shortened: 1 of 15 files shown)


gdx/src/com/badlogic/gdx/utils/LongSet.java
```diff
@@ -0,0 +1,442 @@
...
+			return iterator1;
+		}
+		iterator2.reset();
+		iterator2.valid = true;
+		iterator1.valid = false;
+		return iterator2;
+	}
+
+	static public LongSet with (long... array) {
+		LongSet set = new LongSet();
```
tommyettinger, line 361: The LongSet constructor can take an initialCapacity of `array.length` to avoid the possibility of resizing during this call.

tommyettinger (commented):
Looks good to me, I only had the one comment. But it looks like this was already merged!

NathanSweet:
Good find, I put in initial size for all 3 sets.

mgsx-dev:
@NathanSweet Hi, it seams like some of these changes breaking some libraries : 
`"java.lang.NoSuchMethodError: 'void com.badlogic.gdx.utils.ObjectSet.addAll(com.badlogic.gdx.utils.ObjectSet)'"`.

for instance, with my gdx-gltf library i don't extend ObjectSet anywhere but i use it with no special things (the regular way).

When i use libgdx 1.14.1 in a project, togeter with gdx-gltf (compiled against a prior version of libgdx), it crashes at runtime with the error above. It sounds like an unwanted breaking change.

So we have 2 options : 
* have every libraries (including user libraries) to upgrade to libgdx 1.14.1 : not the best option imho.
* revert signature changes for these methods : do we really need them?

What do you think?

NathanSweet:
@mgsx-dev Sorry for the trouble. It's binary incompatibility from changing existing `addAll` method return values. Libraries would need to be recompiled with the latest.

The new method signature is useful, but if we don't want the breakage I'm OK going back. IntSet, ObjectSet, OrderedSet would need their `addAll` methods changed to void. Since it's been released, changing it again will require things compiled with 1.14.1 to be recompiled. Is that what we want to do?

mgsx-dev:
@NathanSweet  pros and cons that come up in my mind for a revert : 

pros:
* it was released only 10 days ago, so imho, it's resonnable to do it now.
* addAll returning boolean doesn't add so much value, it could be useful sometimes but easily done by comparing size before and after in client code (people needed it probably did it like that).
* i picked up random repo in the gdx team (ashley and gdx-ai), they're using ObjectSet but fortunately don't calling addAll methods (lucky libs :D ), i didn't checked all other libs, but it would be a PITA to have to upgrade those only for that.
* recompiling libs to 1.14.1 also need to release with java 8 source compatibility (since 1.13.5) which might have no impact on code i guess, but it requires time to test them properly since it crashed at runtime, not at compile time.
* it wasn't marked as breaking change and appears to be an unwanted breaking. So it'd be preferable to have this kind of changes in a major release (if we really want it).

cons:
* better code consistency (the reason why you did it i guess).
* dev projects (games or libs) who already switched to 1.14.1 would have to upgrade to 1.14.2

I probably missed some pros and cons though. But if it was my take, i'd revert.

Issue #7754, opened 2026-02-02
GLSurfaceView20 has password flag set by default, preventing screen recording on some devices


LobbyDivinus:
### Describe the bug

I had this issue for a long time where some players report they cannot take screenshots or record videos in my game TheoTown.

I tried a lot of stuff but to no avail. Until today, Vivo (the smartphone brand) reached out to me about an issue their users have with my game on their phones:
> We detected that the custom View com.badlogic.gdx.backends.android.surfaceview.GLSurfaceView20has its inputType attribute set to isPassword.
>               This attribute is typically set for EditTextViews, and we hope it can be removed. Vivo's security mechanism prevents screen recording from the window containing a View with the isPassword attribute set, 
>               thus preventing screen recording from being performed on the game page.

Some investigation later this flag seems to be set by
https://github.com/libgdx/libgdx/blob/9f5ddf1dae56cabc9c6c8feb62eea7bed7f0c91c/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/surfaceview/GLSurfaceView20.java#L82
which contains the flag from here:
https://github.com/libgdx/libgdx/blob/9f5ddf1dae56cabc9c6c8feb62eea7bed7f0c91c/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java#L331
(InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD includes InputType.TYPE_TEXT_VARIATION_PASSWORD)
I don't know why this flag is set here, but if it fixes the issue it's a no-brainer for me to remove it.

My temporary solution until an official fix is available:
```java
    @Override
    protected AndroidGraphics createGraphics(AndroidApplicationConfiguration config) {
        return new AndroidGraphics(this, config,
                config.resolutionStrategy == null ? new FillResolutionStrategy() : config.resolutionStrategy) {
            @Override
            protected GLSurfaceView20 createGLSurfaceView(AndroidApplicationBase application, ResolutionStrategy resolutionStrategy) {
                if (!checkGL20()) throw new GdxRuntimeException("libGDX requires OpenGL ES 2.0");

                GLSurfaceView.EGLConfigChooser configChooser = getEglConfigChooser();
                GLSurfaceView20 view = new GLSurfaceView20(application.getContext(), resolutionStrategy, config.useGL30 ? 3 : 2) {
                    @Override
                    public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
[...]

PR #7771, opened 2026-03-25
Fix password flag on GLSurfaceView causing screen recording issues

LobbyDivinus:
A fix for #7754 

I am assuming TYPE_TEXT_VARIATION_VISIBLE_PASSWORD was used to prevent auto correction. Since it causes screen recording issues on specific vendors I changed it to use the subset TYPE_TEXT_VARIATION_URI instead which does not include TYPE_TEXT_VARIATION_PASSWORD. In my testing it makes no difference, but due to lack of documentation it'd be great if some more people could test it.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java
@@ -328,7 +328,7 @@ public class DefaultAndroidInput extends AbstractInput implements AndroidInput, 
 		default:
 			if (defaultDisableAutocorrection) {
 				inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS
-					| InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
+					| InputType.TYPE_TEXT_VARIATION_URI;
 			} else {
 				inputType = InputType.TYPE_CLASS_TEXT;
 			}


Frosty-J:
Perhaps on smartphones it makes little difference other than toggling the visibility of the number row, but on tablets this often adds a `.com` key or similar (on old Kindle Fires in UK locale, this was `.co.uk`). You can use `OnScreenKeyboardTest` to cycle through them and type stuff - I did try the default option as well, which appeared to behave the same as the Password and URI types respectively (depending on whether your change had been applied).

<table>
<tr>
<th></th>
<th width="50%">Password</th>
<th width="50%">URI</th>
</tr>
<tr>
<th>Gboard</th>
<td>Screenshot_20260325-171121</td>
<td>Screenshot_20260325-171130</td>
</tr>
<tr>
<th>Fire Standard Keyboard</th>
<td>Screenshot_20260325-171211</td>
<td>Screenshot_20260325-171219</td>
</tr>
</table>

Probably not the end of the world. Neither is perfect.

tommyettinger:
Just needs spotless applied for the build to complete, probably: `./gradlew :backends:gdx-backend-android:spotlessApply`

LobbyDivinus:
Based on this, I ran some tests using `OnscreenKeyboardTest` on a tablet with Gboard. The results are as follows:

`TYPE_CLASS_TEXT | TYPE_TEXT_FLAG_NO_SUGGESTIONS | TYPE_TEXT_VARIATION_VISIBLE_PASSWORD` (the current default)
current

`TYPE_CLASS_TEXT | TYPE_TEXT_FLAG_NO_SUGGESTIONS`
nopw
Despite the flags this shows input suggestions.

`TYPE_CLASS_TEXT | TYPE_TEXT_FLAG_NO_SUGGESTIONS | TYPE_TEXT_VARIATION_URI`
uri
Despite the flags this shows input suggestions. Additionally, this one shows a ".com" button.

From these tests, none of the alternatives fully replicate the current behavior of completely suppressing suggestions. At the same time, the current configuration (VISIBLE_PASSWORD) appears to cause issues with screen recording on some devices.

This suggests that any change here would involve a trade-off between preserving the existing “no suggestions” behavior and avoiding the side effects caused by the password variation.

Given that this affects user-visible behavior and device compatibility, I’m unsure which direction is preferable and would appreciate guidance on how to proceed.

Berstanio:
> From these tests, none of the alternatives fully replicate the current behavior of completely suppressing suggestions. At the same time, the current configuration (VISIBLE_PASSWORD) appears to cause issues with screen recording on some devices.
> 
> This suggests that any change here would involve a trade-off between preserving the existing “no suggestions” behavior and avoiding the side effects caused by the password variation.
> 
> Given that this affects user-visible behavior and device compatibility, I’m unsure which direction is preferable and would appreciate guidance on how to proceed.

The sad reality is, that getting IME's to work across all android vendors is borderline impossible. So whatever we do, it will be inconsistent across devices (see https://github.com/libgdx/libgdx/issues/7663 for more fun).
That said, the reason why we want to supress suggestions is, because they don't properly work. They only work for the new Native Input API (`openTextInputField`). 
In my opinion not being able to take screenshots is worse then having non-functional suggestions, so I would lean towards accepting this PR if we are sure, we can't fully block suggestions without breaking screenshots.

LobbyDivinus:
Small update: I contacted vivo again about this issue, and they told me that they have addressed it on their side.

They said the issue was mainly reported on Android 15 and 16 devices, and that a patch for those devices was released in April. From their side, no further changes are needed at this time.

So this issue may already be resolved on updated vivo devices. I’m not sure whether there are still devices without the relevant update, or whether other vendors could show similar problems. Given the lack of further negative feedback on this matter, I’ll close this PR for now. If the problem resurfaces on devices that did not receive a fix, this can always be reconsidered.

Issue #7303, opened 2023-12-19
setScale() is broken in AtlasSprite and ParticleEffect


raeleus:
#### Issue details
A common problem when dealing with physics is that the simulation often uses meters instead of pixels as a unit of measurement. That means you should use an appropriate Viewport with world units and scale your sprites down by a Pixels Per Meter ratio. In Sprite, this would be sprite.setScale(1 / PPM).

That usually works, however the issue is that this does not work for sprites created from a TextureAtlas and the atlas was set to stripwhitespace. AtlasSprite handles most of the adjustments necessary for other transformations, but Sprite#getVertices() is not overridden. This method handles the localized position of the Sprite after scaling. So when the user scales down the sprite, the Sprite is drawn in the incorrect position.

Upon further investigation, I found that there is a similar problem with ParticleEffect. It turns out that ParticleEffect does not implement AtlasSprite at all, causing even more issues. This one is a little more complicated because Particle extends Sprite directly and is tightly paired with the functionality of ParticleEmitter. Note that the problems with ParticleEffect persist even if you just set the scale in the particle effect file.

In the end, you could just say that you shouldn't use the stripwhitespace option (and potentially rotate too). However, that is not documented and seems like an important oversight.

#### Reproduction steps/code
Create a texture atlas with the stripwhitespacex and stripwhitespacey options activated. Make sure to also use images with a lot of whitespace to make a noticeable difference.  You may use code like this to reproduce:

```java
package com.ray3k.testparticle;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.g2d.ParticleEffect;
import com.badlogic.gdx.graphics.g2d.Sprite;
import com.badlogic.gdx.graphics.g2d.SpriteBatch;
import com.badlogic.gdx.graphics.g2d.TextureAtlas;
import com.badlogic.gdx.math.Vector2;
import com.badlogic.gdx.utils.ScreenUtils;
import com.badlogic.gdx.utils.viewport.FitViewport;
import com.badlogic.gdx.utils.viewport.Viewport;

public class Core extends ApplicationAdapter {
    private SpriteBatch batch;
    private ParticleEffect particleEffect;
    private TextureAtlas textureAtlas;
    private Viewport viewport;
    private static final float PPM = 20f;
    private static final Vector2 vector2 = new Vector2();

    private Sprite sprite;

    @Override
[...]

PR #7774, opened 2026-04-02
Fix offset for particles and images with texture atlas with stripwhitespace enabled

brendenobrien:
Fix for #7303

When using a TextureAtlas with whitespace stripping enabled, AtlasSprite.getWidth() and AtlasSprite.getHeight() return the original image dimensions, not the packed region size. Multiple functions were using these methods where the packed dimensions were needed, causing incorrect positioning.

Changes
- Sprite: added getPackedWidth()/getPackedHeight() methods that return the actual packed region size. for regular sprites this is identical to getWidth()/getHeight(); AtlasSprite overrides them to return the true packed dimensions.

- TextureAtlas.AtlasSprite: overrides getPackedWidth()/getPackedHeight() to return the packed size. also fixes setCenter, setCenterX, setCenterY, and setOriginBasedPosition to correctly use the full original dimensions when computing position.

- ParticleEmitter: switched from getWidth()/getHeight() to getPackedWidth()/getPackedHeight() when computing particle scale, size, origin, and frame transitions, fixing unwanted positional offsets when using whitespace-stripped atlas sprites.

Verification

We used the initial texture files included in the issue and created our own as well to test the behavior of the positioning and scaling functions with stripwhitespace enabled and disabled. With the changes we made to the classes both atlases with stripwhitespace enabled and disabled looked identical upon rendering

With stripwhitespace enabled
Screenshot 2026-04-01 184651

With stripwhitespace disabled
Screenshot 2026-04-01 184719

The particles are hovering over the cursor in both cases, the cursor is just not visible when using PrtSc


gdx/src/com/badlogic/gdx/graphics/g2d/TextureAtlas.java
```diff
@@ -786,6 +806,16 @@ public float getHeight () {
 			return super.getHeight() / region.getRotatedPackedHeight() * region.originalHeight;
 		}
 
+		@Override
+		public float getPackedWidth () {
+			return super.getWidth();
```
tommyettinger: This technically could just return `width`, which is package-private in Sprite, but this almost certainly will get inlined on desktop to return the same thing anyway. Using the call to `super.getWidth()` is more flexible in code that's copying AtlasSprite into another package, so this comment would only matter if some platforms see performance loss from frequently calling `super.getWidth()` instead of just returning the field. There probably is no real difference, I just wanted to point out that if there is any performance loss, this would be a quick thing to change.

gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java
```diff
@@ -643,9 +643,9 @@ private boolean updateParticle (Particle particle, float delta, int deltaMillis)
 				float prevSpriteWidth = particle.getWidth();
 				float prevSpriteHeight = particle.getHeight();
 				particle.setRegion(sprite);
-				particle.setSize(sprite.getWidth(), sprite.getHeight());
-				particle.setOrigin(sprite.getOriginX(), sprite.getOriginY());
-				particle.translate((prevSpriteWidth - sprite.getWidth()) * 0.5f, (prevSpriteHeight - sprite.getHeight()) * 0.5f);
+				particle.setSize(sprite.getPackedWidth(), sprite.getPackedHeight());
+				particle.setOrigin(sprite.getPackedWidth() / 2, sprite.getPackedHeight() / 2);
+				particle.translate((prevSpriteWidth - sprite.getPackedWidth()) * 0.5f, (prevSpriteHeight - sprite.getPackedHeight()) * 0.5f);
```
tommyettinger: These lines, and some below this, are inconsistent about multiplying by 0.5f or dividing by 2. I'd personally multiply by 0.5f for all cases where a float is divided by 2, but compilers often make that change themselves anyway, where possible. The existing code multiplied by 0.5f, so I'd go with what it does. This is just a suggestion, it won't block my review.

tommyettinger (changes requested):
This is a really clean solution to what seemed like a thorny problem. Being able to access packed height in Sprite but have it only behave differently in AtlasSprite is a good approach, and probably has no performance penalty. The behavior changes in AtlasSprite look like good fixes also, but because users are likely working around the incorrect centered positions in existing code, the changes to how AtlasSprite positioning works should be documented with a line in the CHANGES file.

Updating CHANGES is the only thing I think must be done here, though I suggest using `* 0.5f` instead of `/ 2` in a few places for consistency with the existing code. That said, this is something that could affect a lot of people using AtlasSprite. I don't know how many people are using origins with stripped-whitespace AtlasSprites right now, especially since they appear incorrect without this (at least in particle code).

gdx/src/com/badlogic/gdx/graphics/g2d/Sprite.java
```diff
@@ -612,6 +612,20 @@ public float getHeight () {
...
+		return width;
+	}
+
+	/** @return the height of the packed region of the sprite. For sprites created from a {@link TextureAtlas} with whitespace
+	 *         stripping, this is the height of the region without the stripped whitespace, as opposed to {@link #getHeight()} which
+	 *         returns the height of the original image. */
+	public float getPackedHeight () {
+		return height;
+	}
+
```
NathanSweet, line 628: So far Sprite doesn't know anything about packed or unpacked. This change leaks it up a layer. That could be OK if special casing for AtlasSprite is annoying and using the packed size is done often enough, just want to make sure it's done consciously. AtlasRegion doesn't get similar getPackedWidth/Height methods?

Where I special case AtlasRegion (eg [1](https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/attachments/RegionAttachment.java#L102), [2](https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/attachments/MeshAttachment.java#L111)) I need more than the packed size, so doesn't change.

gdx/src/com/badlogic/gdx/graphics/g2d/ParticleEmitter.java
tommyettinger: Looks like you fixed this! I'll mark this as resolved.

tommyettinger (dismissed):
My changes requested have been all addressed! Approved.

tommyettinger:
It does need `./gradlew :gdx:spotlessApply` run to allow the build to pass, but other than that I like these changes.

tommyettinger (dismissed):
Spotless ran, so this builds now. Approved again!

obigu:
I think this is a good change. I've come across this issue in the past and decided not to trim atlases with whitespace. Any chance you can resolve the CHANGES conflict? (I can do it myself but I see the PR is on `master` , it's usually better if you work on a separate branch for the PR)

gdx/src/com/badlogic/gdx/graphics/g2d/TextureAtlas.java
```diff
@@ -756,6 +756,26 @@ public void rotate90 (boolean clockwise) {
...
+		}
+
+		@Override
+		public void setCenterY (float y) {
+			setY(y - getHeight() * 0.5f);
+		}
+
+		@Override
+		public void setOriginBasedPosition (float x, float y) {
+			setPosition(x - getOriginX(), y - getOriginY());
```
obigu: I think this is redundant as adjustments to region offsets are cancelled on `setPosition` and `getOrigin` methods resulting on the same implementation of `super.setOriginBasedPosition()`.

obigu (changes requested):
Changes are good but I think 3 method implementations are redundant and can be removed. Other than that, once conflicts in CHANGES are fixed we can merge.

gdx/src/com/badlogic/gdx/graphics/g2d/TextureAtlas.java
```diff
@@ -786,6 +806,16 @@ public float getHeight () {
...
 
+		@Override
+		public float getPackedWidth () {
+			return width;
+		}
+
+		@Override
+		public float getPackedHeight () {
+			return height;
+		}
```
obigu: These 2 are also redundant, super.getPackedWidth() and super.getPackedHeight() are same, can be removed

tommyettinger (approved):
The changes look fine, approved again!

obigu (approved):
Thanks!

Issue #7775, opened 2026-04-03
Random crashes on Back gesture or button on Android 36


obigu:
### Describe the bug

I'm seeing random crashes on devices running Android 36 after updating to target SDK 36. This looks related to the change in behaviour of Back mechanism with the deprecation of the old key based `onBackPressed()` and the introduction of Predictive Back gestures.

libGDX added support to predictive back gestures here https://github.com/libgdx/libgdx/pull/7429.

My guess the problem is related to the following code in `PredictiveBackHandler` here

```
private final OnBackInvokedCallback callback = new OnBackInvokedCallback() {
			@Override
			public void onBackInvoked () {
				if (processor != null) {
					processor.keyDown(Keys.BACK);
				}
			}
		};
```

which executes the back related callback immediately instead of processing it like it was before in `DefaultAndroidInput` though `processEvents()` which was executed at a specific time in the lifecycle.

### version

1.14.0

### Stack trace

```shell

```

### Affected platforms

- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

PR #7429, opened 2024-08-01
Add support for predictive back gestures

Frosty-J:
It's what it sounds like: https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture

So for users using gestures, going back in your app is much the same as without support, except now they can see where they'll end up if they try to go back from your main menu. Unfortunately, I don't know of a way to test if a user has these gestures enabled, such as to only show an "are you sure you want to quit" prompt to users that don't have it (though as of Android 12 the default behaviour is to hide an app on back press instead of close it, so it's not as if they'd be subjected to load times by accidentally exiting).

I also removed the old deprecated methods for setting and catching the back and menu keys. I haven't even marked it as a breaking change because they've been deprecated for long enough that it really shouldn't come as a surprise when it stops working. There isn't strictly a need to remove them, but it cleans up the codebase a little.

To-do (after merge and next libGDX release):

* Add manifest attribute to gdx-liftoff.
* Update https://libgdx.com/wiki/input/back-and-menu-key-catching, even if just to remove the bit about the menu key (killed off in 2011 - you'll struggle to find a device still compatible with libGDX that has a menu key, though some may still be able to emulate one by holding down a capacitive key).

PR #7776, opened 2026-04-03
Android and iOS versions maintenance

obigu:
Updated some Android and iOS backends related versions that are battle tested. 

As a highlight it seems we don't need to declare the `kotlin-bom` dependency anymore.


gradle/dependencies.gradle
```diff
@@ -33,13 +33,12 @@ versions.lwjgl3 = "3.3.3"
 versions.jlayer = "1.0.1-gdx"
 versions.jorbis = "0.0.17"
 versions.junit = "4.13.2"
-versions.androidPlugin = "8.11.2"
-versions.androidCompileSdk = 35
-versions.androidTargetSdk = 35
+versions.androidPlugin = "8.13.2"
+versions.androidCompileSdk = 36
```
tommyettinger, line 37: Should we wait to update these two versions to 36 until after https://github.com/libgdx/libgdx/issues/7775 is figured out? Is updating this part of fixing that?

obigu: It's independent of https://github.com/libgdx/libgdx/issues/7775 and it doesn't have an impact. On the one hand these versions impact only Gdx Tests. They also serve the purpose (in theory) to inform about the tested versions we recommend users and should match the ones on Liftoff. 

Thing is, since `android:enableOnBackInvokedCallback="true"` was added to the `AndroidManifest.xml` as recommended in https://github.com/libgdx/libgdx/pull/7429 both in Gdx Tests and Liftoff if I'm not mistaken, the issue will also occur even when targetting Android 35.

I'm almost sure the https://github.com/libgdx/libgdx/issues/7775 fix I'm testing works, just waiting for more users in production to be sure, will probably PR either later today or tomorrow. In the meantime I will add some more info in the issue description of the affected devices/configs/versions.

PR #7779, opened 2026-04-09
Add Justify text option to Label

StartsMercury:
Justify option for Label and getter, setter. Label layout() is updated to handle justify through GlyphLayout. (Default is disabled, none)

Preview of modified LabelTest: 15143e6.

JustifyLabel


tommyettinger (approved):
Looks good to me. We have justified text now, so we just need to be able to use it. Approved.

Issue #7594, opened 2025-03-02
Dragging Slider inside ScrollPane moves the ScrollPane


nirslsk:
In my android libgdx game i have a growing vertical series of widgets inside a ScrollPane where each widget contains a Slider. My problem was that dragging any of the sliders would also move the scrollpane up and down in response to unintentional vertical movement. After some investigation is appears that the scrollpane correctly ignores touchDragged events while the slider is dragged, but that it still (I think) receives drag event through the gestures system. Calling scrollPane.setFlickScroll(false) in the slider's touchDown handler and then setFlickScroll(true) in the touchUp handler works perfectly. Tangentially, using setScrollingDisabled instead of setFlickScroll in the touchDown and touchUp handlers makes the scrollpane jump to the top when the slider is touched if the scrollpane isn't already at the top. Not sure if the ScrollPane responding to gesture events is a bug or intended behavior.

nulldoubt:
I know the exact behavior you're talking about and can assure that it's really bad.

One nice workaround was to experiment with the `setCancelTouchFocus` method in the `ScrollPane` class, and just to validate: No this **doesn't** fix the ugly effect but it **minimizes** it.

nirslsk:
The setFlickScroll method works perfectly for me so far, for what it's worth.

PR #7781, opened 2026-04-13
Add field to prevent touch propagation in drag-based actors [VISUAL E.G.]

Urban-Elf:
Some Scene2d actors that implement drag functionality (`TextField`/`TextArea` for selecting, `Slider` for value adjustment, and `Touchpad` for, well, touchpadding), don't really behave properly while in the context of a `ScrollPane`.  By that I mean the drag gets interrupted by the `ScrollPane`'s own `ClickListener`, which causes the event it receives to be `stop()`-ed.

This PR adds a `preventTouchPropagation` field in said drag-capable actors.  When enabled, `event.stop()` is called within their own internal `ClickListener#touchDown`, preventing aforementioned issue.

Resolves:
- #7594.

https://github.com/user-attachments/assets/17c784a8-87ed-4f4a-ae1e-e2269c0fcb62

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Slider.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/Slider.java
@@ -73,8 +73,7 @@ public class Slider extends ProgressBar {
 				if (draggingPointer != -1) return false;
 				draggingPointer = pointer;
 				calculatePositionAndValue(x, y);
-				if (preventTouchPropagation)
-					event.stop();
+				if (preventTouchPropagation) event.stop();
 				return true;
 			}
 
@@ -244,14 +243,13 @@ public class Slider extends ProgressBar {
 		setValue(min + (max - min) * visualInterpolationInverse.apply(percent));
 	}
 
-	public boolean isPreventTouchPropagation() {
+	public boolean isPreventTouchPropagation () {
 		return preventTouchPropagation;
 	}
 
-	/** Prevents internal touchDown events from propagating to parental hierarchy.
-	 * This stops parents from cancelling them (as in the case of {@link ScrollPane},
-	 * which would otherwise lead to unexpected behavior. */
-	public void setPreventTouchPropagation(boolean preventTouchPropagation) {
+	/** Prevents internal touchDown events from propagating to parental hierarchy. This stops parents from cancelling them (as in
+	 * the case of {@link ScrollPane}, which would otherwise lead to unexpected behavior. */
+	public void setPreventTouchPropagation (boolean preventTouchPropagation) {
 		this.preventTouchPropagation = preventTouchPropagation;
 	}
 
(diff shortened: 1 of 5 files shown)


Urban-Elf:
Currently `preventTouchPropagation` is set to false, but unless forcibly enabling it would cause some kind of regression (I'm not very educated at all in the event side of Scene2d), I think its application could be considered broad enough to warrant having it set to true as a default.  Just an idea, though.

tommyettinger:
I haven't used drag and drop at all, so I'll ping @NathanSweet as the expert on the topic to see if there are any unexpected regressions involved in this.

NathanSweet:
[ScrollPaneTest](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneTest.java) has a slider in a scrollpane. It works by adding a listener to the slider that does `event.stop()`, so it doesn't reach the scrollpane. Is adding a listener sufficient? Seems nicer as configuration (if maybe not terribly intuitive) and not a feature that every UI control needs.

Urban-Elf:
> [ScrollPaneTest](https://github.com/libgdx/libgdx/blob/master/tests/gdx-tests/src/com/badlogic/gdx/tests/ScrollPaneTest.java) has a slider in a scrollpane. It works by adding a listener to the slider that does `event.stop()`, so it doesn't reach the scrollpane. Is adding a listener sufficient? Seems nicer as configuration (if maybe not terribly intuitive) and not a feature that every UI control needs.

That's actually a great approach, I never considered that. I would say so, yeah. Perhaps some documentation could be added somewhere, or maybe a special listener packaged with scene2d? I think this issue is common enough that newer devs would benefit from it. As far as myself, I had gone as far as copying the Widget files into my own project just to make that change.

NathanSweet:
I agree, it's easily missed. I propose adding this to ScrollPane class javadoc:
```
 * <p>
 * If the scroll pane contains widgets that respond to drag, a listener that calls {@link Event#stop()} in
 * {@link InputListener#touchDown(InputEvent, float, float, int, int) touchDown} can be added to the widgets to prevent the scroll
 * pane from panning when {@link #setScrollBarTouch(boolean)} is true.
```

Urban-Elf:
> I agree, it's easily missed. I propose adding this to ScrollPane class javadoc:
> ```
>  * <p>
>  * If the scroll pane contains widgets that respond to drag, a listener that calls {@link Event#stop()} in
>  * {@link InputListener#touchDown(InputEvent, float, float, int, int) touchDown} can be added to the widgets to prevent the scroll
>  * pane from panning when {@link #setScrollBarTouch(boolean)} is true.
> ```

Great! Should I close this PR now, or are you suggesting editing this branch to make that change? I'm unsure about what the standard practice is.

NathanSweet:
I prefer whatever is easiest but still gives others a chance to participate. Generally it's good to get more than one set of eyes on a change. Since you like the proposed javadoc I'll throw it in there and we can close this issue.

PR #7782, opened 2026-04-16
Fix NullPointerException in AndroidGraphics.updateSafeAreaInsets when WindowInsets is null

### LINKED ISSUE #7772: Crash in com.badlogic.gdx.backends.android.AndroidGraphics.updateSafeAreaInsets
### Describe the bug

Hi, I am facing this issue since a lot of time ago in all of my games.
Not happening always, just on some "maybe edge" cases...

Image

A null check on the result of getRootWindowInsets() before calling .getDisplayCutout() could fix it,

### Reproduction

_No response_

### version

1.14.0 (also on previous versions, but not identified since when)

### Stack trace

```shell
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.DisplayCutout android.view.WindowInsets.getDisplayCutout()' on a null object reference
  at com.badlogic.gdx.backends.android.AndroidGraphics.updateSafeAreaInsets (AndroidGraphics.java:664)
  at com.badlogic.gdx.backends.android.AndroidGraphics.onSurfaceCreated (AndroidGraphics.java:329)
  at android.opengl.GLSurfaceView$GLThread.guardedRun (GLSurfaceView.java:1545)
  at android.opengl.GLSurfaceView$GLThread.run (GLSurfaceView.java:1276)
```

### Affected platforms

- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [ ] macOS

mritunjayyy07:
Fixes #7772 
Changes:

- Added null checks for Window, DecorView, and WindowInsets before accessing DisplayCutout
- Replaced direct chained calls with safe intermediate variables
- Ensured DisplayCutout is accessed only when WindowInsets is available

Reason:

View.getRootWindowInsets() may return null during early lifecycle stages (e.g., when called from onSurfaceCreated). 
The previous implementation directly accessed getDisplayCutout() without null checks, leading to a NullPointerException.

Risk:
Low — changes are defensive and preserve existing behavior. If WindowInsets are not available, safe area insets default to 0 as before.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java
@@ -676,9 +676,11 @@ public class AndroidGraphics extends AbstractGraphics implements Renderer {
 
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
 			try {
-				WindowInsets windowInsets = app.getApplicationWindow().getDecorView().getRootWindowInsets();
-				if (windowInsets == null) return;
-				DisplayCutout displayCutout = windowInsets.getDisplayCutout();
+				View decorView = app.getApplicationWindow().getDecorView();
+				android.view.WindowInsets insets = decorView.getRootWindowInsets();
+				if (insets == null) return;
+
+				DisplayCutout displayCutout = insets.getDisplayCutout();
 				if (displayCutout != null) {
 					safeInsetRight = displayCutout.getSafeInsetRight();
 					safeInsetBottom = displayCutout.getSafeInsetBottom();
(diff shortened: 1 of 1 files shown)


obigu (changes requested):
Adding a null check here is ok but the reported problem is `WindowInsets` being null in some edge case (the view has not been yet attached). Let's just add that null check not to add unnecessary noise.

mritunjayyy07:
Thanks for the feedback! I've updated the PR to only check for null WindowInsets as suggested.

obigu (dismissed):
LGTM now, thanks!

It looks like it's not properly formatted, I wonder why Spotless has not run automatically.

mritunjayyy07:
Thanks! I've applied Spotless formatting.

obigu:
I see why it was not running Spotless, your changes don't compile. Good practice is to build and test locally before pushing.

mritunjayyy07:
I am so sorry, I am new to contributing to Open Source. 
I will build and test from now onwards

obigu:
> I am so sorry, I am new to contributing to Open Source. I will build and test from now onwards

No worries, feel free to commit the changes when tested.

mritunjayyy07:
I have updated and compiled the code, please verify on your end as well
thank you

PR #7786, opened 2026-04-25
Fix NPE in AndroidDaydream.onConfigurationChanged when input is null

### LINKED ISSUE #6633: NullPointerException in AndroidDaydream.onConfigurationChanged
[This line](https://github.com/libgdx/libgdx/blob/af899350378fc0908eb26867b8cc44fd3cfdce7c/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidDaydream.java#L299) is throwing NPEs for null `input` in my Google Play crash reports on Android 11 Samsung devices only. It is very frequent.

I don't have a device with Android 11 to test with. But if no one else is able to fix this, I'm open to doing 

```
//pseudo code
if (input != null) 
    input.setKeyboardAvailable(keyboardAvailable)
else
    postRunnable(() -> if (input != null) input.setKeyboardAvailable(keyboardAvailable) )
```
to give it an opportunity to still set it if this is some kind of lifecycle mis-ordering.

Vikashksingh1308:
Fixes #6633

## Problem
onConfigurationChanged fires before input is initialized on Android 11 
Samsung devices, causing a frequent NullPointerException.

## Fix
Added a null check for input before calling setKeyboardAvailable(). If input is null, the call is deferred via postRunnable() so it still executes once input is initialized.

### DIFF
--- a/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidDaydream.java
+++ b/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidDaydream.java
@@ -282,7 +282,17 @@ public class AndroidDaydream extends DreamService implements AndroidApplicationB
 		super.onConfigurationChanged(config);
 		boolean keyboardAvailable = false;
 		if (config.hardKeyboardHidden == Configuration.HARDKEYBOARDHIDDEN_NO) keyboardAvailable = true;
-		input.setKeyboardAvailable(keyboardAvailable);
+		final boolean finalKeyboardAvailable = keyboardAvailable;
+		if (input != null) {
+			input.setKeyboardAvailable(keyboardAvailable);
+		} else {
+			postRunnable(new Runnable() {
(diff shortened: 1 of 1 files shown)


obigu:
This seems to be a duplicate of https://github.com/libgdx/libgdx/pull/6906. I think the reason explained not to add the deferred call in a `postRunnable` on the othe PR is reasonable, I really would avoid the hack unless really necessary. @Vikashksingh1308 would just merging #6906 be enough for your use case? If that's the case we can merge that one and close this one.

Vikashksingh1308:
Thanks for the review @obigu ! That makes sense. I added the `postRunnable` thinking it would be safer to ensure `setKeyboardAvailable()` still ran, but you're right that the keyboard state gets set correctly later in `initialize()` anyway and for a screensaver it's not critical. #6906 would fully resolve my use case. Happy to close this PR in favor of merging that one. Thanks again!!

PR #7787, opened 2026-04-26
Json#ignoreUnknownField() parameters changed to provide more information

NathanSweet:
Json `ignoreUnknownField()` used to take `Class type, String fieldName`. This PR changes it to pass `Object object, JsonValue value`, providing the exact same values as the old parameters by using `object.getClass()` and `value.name`. This gives a lot more information about unknown fields (which I need in app code to more easily preserve that information).

The downside is it's a protected method, so public API and a breaking change. Subclasses that override the old method need to change. If they use `@Override` it's obvious, if not it's silent. I doubt a whole lot of code overrides this method though.


tommyettinger (approved):
I'm a reasonably heavy user of libGDX Json, and I didn't even know this function existed. There are only 3 `protected` methods in `Json`, and 24 usages of the `private` keyword in code, so I'm not sure how often subclasses exist of it. I'd expect it to be more common to copy the whole class to avoid duplicating the `private` stuff.

The change in `Skin` is the only mildly concerning part, because subclasses of `Skin` exist all over. The FreeType config for Skins in Stripe is one example, as is my FWSkin in TextraTypist. I don't think either would be affected significantly, or maybe at all.

I think this is probably fine to merge, and having more info to decide how to handle a potentially-ignored field would be good.

NathanSweet:
The Skin change is in its internal Json instance. Someone overriding `getJsonLoader` might do similar with `ignoreUnknownField` overriding, so that is the most likely place this change impacts. The fix is easy enough so let's merge and revisit if someone has concerns.

The functionality it provides is neat. It lets me collect (in app code) JSON for unused fields when reading. When writing later I append the extra JSON so data isn't lost. This way settings JSON can evolve with backward and lossless forward compatibility.

PR #7788, opened 2026-04-26
Add IdentitySet

NathanSweet:
Not to bloat our utils/collections too much further, but having IdentitySet would be nice.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/IdentitySet.java
+++ b/gdx/src/com/badlogic/gdx/utils/IdentitySet.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright 2026 See AUTHORS file.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+package com.badlogic.gdx.utils;
+
+/** An unordered set that uses identity comparison for its keys. Null keys are not allowed. No allocation is done except when
+ * growing the table size.
+ * <p>
+ * This class performs fast contains and remove (typically O(1), worst case O(n) but that is rare in practice). Add may be
+ * slightly slower, depending on hash collisions. Hashcodes are rehashed to reduce collisions and the need to resize. Load factors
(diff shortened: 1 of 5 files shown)


gdx/src/com/badlogic/gdx/utils/IdentitySet.java
```diff
@@ -0,0 +1,88 @@
...
+		super(initialCapacity, loadFactor);
+	}
+
+	/** Creates a new set identical to the specified set. */
+	public IdentitySet (IdentitySet<? extends T> set) {
+		super(set);
+	}
+
+	protected int place (T item) {
+		return (int)(System.identityHashCode(item) * 0x9E3779B97F4A7C15L >>> shift);
```
tommyettinger: There's no need to multiply by the long hex constant and no need to cast to int here; identityHashCode() is already very random as-is and won't collide often at all unmodified. The shift will work fine on an int as well as on a long due to how shifts work in Java (only the bits that can matter are used, so a shift between 32 and 63 is the same as that shift minus 32 if used on an int).

Not involving any long math is definitely preferable on GWT, where it is slow.

gdx/src/com/badlogic/gdx/utils/IdentitySet.java
```diff
@@ -0,0 +1,88 @@
...
+ * <p>
+ * This class performs fast contains and remove (typically O(1), worst case O(n) but that is rare in practice). Add may be
+ * slightly slower, depending on hash collisions. Hashcodes are rehashed to reduce collisions and the need to resize. Load factors
+ * greater than 0.91 greatly increase the chances to resize to the next higher POT size.
+ * <p>
+ * Unordered sets and maps are not designed to provide especially fast iteration. Iteration is faster with OrderedSet and
+ * OrderedMap.
+ * <p>
+ * This implementation uses linear probing with the backward shift algorithm for removal. Hashcodes are rehashed using Fibonacci
+ * hashing, instead of the more common power-of-two mask, to better distribute poor hashCodes (see <a href=
```
tommyettinger: If the multiplication by 0x9E3779B9... is removed, this won't be using Fibonacci hashing, and that section of the docs can be removed. Using identityHashCode() as-is is actually better most of the time for hash collisions, since it doesn't have the same problem case where Fibonacci numbers returned by hashCode() can collide more often.

tommyettinger (changes requested):
I really like the idea of an IdentitySet; it's so strange that the JDK doesn't have one! There's some quick speedups possible here because System.identityHashCode() is already very well-mixed and doesn't benefit from additional mixing.

gdx/src/com/badlogic/gdx/utils/IdentitySet.java
NathanSweet: Great catch! Removed from both IdentitySet and IdentityMap. Good to go now?

tommyettinger (approved):
Good to go! Let's merge!

gdx/src/com/badlogic/gdx/utils/IdentitySet.java
tommyettinger: Yep! Good to go!

PR #7790, opened 2026-04-28
TextField, enforce cursor/selection invariants on text change.

NathanSweet:
TextField requires `cursor` and `selectionStart` to be valid indices into `text`, and if `hasSelection` the two must differ. This PR enforces those invariants when text changes.

bdbcd7b24 in 2021 was an incomplete attempt at keeping the state consistent but I see crash reports like:
```
java.lang.StringIndexOutOfBoundsException: begin 0, end 4, length 0
   at java.lang.String.checkBoundsBeginEnd(String.java)
   at java.lang.String.substring(String.java)
   at scene2d.ui.TextField.copy(TextField.java:481)
   at scene2d.ui.TextField$TextFieldClickListener.keyDown(TextField.java:1091)
```


tommyettinger (approved):
Seems like a reasonable set of extra precautions. You could use Math.min(), though, which gets optimized on recent JDKs (11 and up, maybe some earlier too), but it probably doesn't matter.

NathanSweet:
The readability with `>` is slightly better than `min`, shouldn't matter.

PR #7791, opened 2026-04-29
Deserialized Strings are interned

Ambeco:
This slows deserialization slightly, but reduces memory used by deserialized objects, which can be dramatic.

Specifically, when deserializing an Unciv saved game after the game ends, I'm seeing about a full Megabyte of wasted memory from duplicate strings in heap dumps

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Json.java
+++ b/gdx/src/com/badlogic/gdx/utils/Json.java
@@ -1141,7 +1141,7 @@ public class Json {
 				if (type == int.class || type == Integer.class) return (T)(Integer)jsonData.asInt();
 				if (type == long.class || type == Long.class) return (T)(Long)jsonData.asLong();
 				if (type == double.class || type == Double.class) return (T)(Double)jsonData.asDouble();
-				if (type == String.class) return (T)jsonData.asString();
+				if (type == String.class) return (T)jsonData.asString().intern();
 				if (type == short.class || type == Short.class) return (T)(Short)jsonData.asShort();
 				if (type == byte.class || type == Byte.class) return (T)(Byte)jsonData.asByte();
 				if (type == char.class || type == Character.class) return (T)(Character)jsonData.asChar();
@@ -1160,7 +1160,7 @@ public class Json {
 
 		if (jsonData.isString()) {
 			String string = jsonData.asString();
-			if (type == null || type == String.class) return (T)string;
+			if (type == null || type == String.class) return (T)string.intern();
 			try {
 				if (type == int.class || type == Integer.class) return (T)Integer.valueOf(string);
 				if (type == float.class || type == Float.class) return (T)Float.valueOf(string);


tommyettinger:
1MB of memory is not a dramatic difference in the year 2026, on any hardware or platform that runs a JVM. That's as much memory as a single 512x512 RGBA8888 texture takes (in VRAM). A typical 1080p background takes 8x more VRAM, or 6x if it's RGB888. A brief .wav sound effect will take more than 1MB of memory to load. The 1MB of Strings will also be garbage collected if all references to those Strings are gone. Each call to intern() takes time, which didn't have to be spent before, on every String read in. It's approximately comparable to adding a String into a HashSet, regardless of whether it's been interned already or not. I'd need to benchmark in a real-world application to see how much time the calls to intern() take, but it's comparing against 0ns without intern() calls. On Android in particular, we don't even know how ART handles the String constant pool, and it could change in the next version of Android without any warning. On GWT and TeaVM, intern() I believe is a no-op (GWT source: https://github.com/gwtproject/gwt/blob/main/user/super/com/google/gwt/emul/java/lang/String.java#L490-L492 ), so it wouldn't help or hurt at all there, other than adding an extraneous method call. 

Most advice I can find on the internet, mostly Stack Overflow at this point, is quite dated regarding String.intern(). Lots of things are stuck in the Java 6 advice where intern() must be avoided because there's a small limit on how many Strings can actually be interned. That said, we don't really know how Android handles this, and it may still have that Java-6-era limit in some versions. If it does, it would be really bad to add extra calls to intern(). If it doesn't, it's still slower for a very small reduction in memory usage, which should get GCed anyway.

I'm not in favor of adding extra work to serialization when the only gain we can find is 1MB less garbage that will likely get collected anyway. If you can show that multiple loads are using more memory that isn't getting GCed, then maybe it's worth it? I'd think that if it reaches 50MB without getting GCed in a real world application, that would be a possible problem.

NathanSweet:
We definitely don't want to intern strings for everyone. It could probably be done in app code by subclassing Json, something like: if `readValue` would return String then intern it.

Some unsolicited advice:
* If you can suffer event-based parsing (doubt), JsonSkimmer is most efficient.
* If you're processing a small part of a lot of JSON, use JsonMatcher. It's cool!
* If you want the entire JSON document as a JsonValue graph, using JsonMatcher can still make sense. It provides JsonValues as object, array, null, boolean, or String. Number parsing is deferred until explicitly converted (eg JsonValue `asFloat()`). If you don't need ALL numbers in the JSON parsed, this can save a lot of work.
* If you want JSON mapped to Java classes, use Json.
* If memory or performance is important, don't use JSON for your data format. I recommend Kryo.

Ambeco:
I did see if there was an easy way to wrap, but didn't see anything
immediately obvious that worked with String.class specifically, and
*speculated* that deserialization was already slow enough that the addition
of interning strings wouldnt matter.

That's fine, I'll look deeper about finding a way to wrap string
deserialization.

Ambeco:
Looking closer, it does indeed appear as if I can simply use `classToSerializer` in my client code to intern for our project. Thank you for looking at the PR, and for your patience!

PR #7792, opened 2026-05-03
Use IdentityMap for Skin resources.

NathanSweet:
IdentityMap is ever so slightly better for Skin `resources`, since the keys are Class.


tommyettinger (approved):
Looks good to me.

PR #7795, opened 2026-05-21
Fixed stuck over state when ClickListener is cancelled

NathanSweet:
When ClickListener is cancelled during press/drag, enter/exit is ignored until touch up. This means the value of `over` (usually true) is stuck until next enter.

With this fix we set `over` to false on cancel and maintain the real over state separately in `cancelledOver`. On touch up we restore the real `over` state and everything resumes working as normal, no stuck/lost over state.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+[1.14.2]
+- Fixed stuck over state when ClickListener is cancelled.
+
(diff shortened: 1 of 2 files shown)


NathanSweet:
Tested in prod, seems good.

PR #7796, opened 2026-05-22
Sort improvements

NathanSweet:
* Reset minGallop to not be affected by previous sorts.
* ComparableTimSort uses 4 run mergeCollapse like TimSort.
* Clear refs in try/finally using Arrays.fill.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
+++ b/gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
@@ -13,6 +13,8 @@
 
 package com.badlogic.gdx.utils;
 
+import java.util.Arrays;
+
 /** This is a near duplicate of {@link TimSort}, modified for use with arrays of objects that implement {@link Comparable},
  * instead of using explicit comparators.
  * 
@@ -39,9 +41,9 @@ class ComparableTimSort {
 	/** When we get into galloping mode, we stay there until both runs win less often than MIN_GALLOP consecutive times. */
 	private static final int MIN_GALLOP = 7;
 
-	/** This controls when we get *into* galloping mode. It is initialized to MIN_GALLOP. The mergeLo and mergeHi methods nudge it
-	 * higher for random data, and lower for highly structured data. */
-	private int minGallop = MIN_GALLOP;
+	/** This controls when we get *into* galloping mode. It is reset to MIN_GALLOP before each sort. The mergeLo and mergeHi
+	 * methods nudge it higher for random data, and lower for highly structured data. */
+	private int minGallop;
 
 	/** Maximum initial size of tmp array, which is used for merging. The array can grow to accommodate demand.
 	 * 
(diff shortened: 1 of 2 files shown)


gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
```diff
@@ -39,9 +41,9 @@ class ComparableTimSort {
 	/** When we get into galloping mode, we stay there until both runs win less often than MIN_GALLOP consecutive times. */
 	private static final int MIN_GALLOP = 7;
 
-	/** This controls when we get *into* galloping mode. It is initialized to MIN_GALLOP. The mergeLo and mergeHi methods nudge it
-	 * higher for random data, and lower for highly structured data. */
-	private int minGallop = MIN_GALLOP;
+	/** This controls when we get *into* galloping mode. It is reset to MIN_GALLOP before each sort. The mergeLo and mergeHi
```
tommyettinger, line 44: Good catch, this is mutable state and does need resetting, unlike in the Android version that creates new TimSort instances all the time.

gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
```diff
@@ -87,47 +89,48 @@ public void doSort (Object[] a, int lo, int hi) {
...
+				nRemaining -= runLen;
+			} while (nRemaining != 0);
+
+			// Merge all remaining runs to complete sort
+			if (DEBUG) assert lo == hi;
+			mergeForceCollapse();
+			if (DEBUG) assert stackSize == 1;
+		} finally {
+			this.a = null;
+			Arrays.fill(tmp, 0, tmpCount, null);
```
tommyettinger, line 124: Very good stuff. I'm glad we can use Arrays.fill(), and I hope it is used in other places where we need to clear references in bulk, too.

gdx/src/com/badlogic/gdx/utils/ComparableTimSort.java
```diff
@@ -337,21 +340,25 @@ private void pushRun (int runBase, int runLen) {
...
 	/** Examines the stack of runs waiting to be merged and merges adjacent runs until the stack invariants are reestablished:
 	 * 
-	 * 1. runLen[i - 3] > runLen[i - 2] + runLen[i - 1] 2. runLen[i - 2] > runLen[i - 1]
+	 * 1. runLen[n - 2] > runLen[n - 1] + runLen[n] 2. runLen[n - 1] > runLen[n]
+	 * 
+	 * where n is the index of the last run in runLen.
+	 * 
+	 * This method has been formally verified to be correct after checking the last 4 runs. Checking for 3 runs results in an
+	 * exception for large arrays. (Source:
+	 * http://envisage-project.eu/proving-android-java-and-python-sorting-algorithm-is-broken-and-how-to-fix-it/)
```
tommyettinger, line 349: Woah, we were missing this change? I thought it was in the Android version we got this from... I guess only ComparableTimSort was missing it. Not like the huge-array situation affects games a whole lot, but it's still important to be correct.

tommyettinger (approved):
Looks like solid changes to me! Nothing seems to break here, just some good fixes. Approved!

PR #7800, opened 2026-05-24
Use Shewchucks exact predicates and infinite super-triangle DelaunayTriangulator

### LINKED ISSUE #7487: Broken Edge cases in DelaunayTriangulator.computeTriangles()

Berstanio:
Based on the work of @jose-puente and fixes https://github.com/libgdx/libgdx/issues/7487.
The PR consists of two parts:
- Porting of [Shewchuk Exact Predicates](https://www.cs.cmu.edu/~quake/robust.html) to java based of the JS implementation https://github.com/mourner/robust-predicates. The java port does not follow the JS based coordinate system conventions, but the original Shewchuk's one which align with libGDX defaults. Meaning, y-up in 2d and right-handed in 3d. The port has been assisted by generative AI.
- The delaunay triangulation now utelises the exact incircle test. The finite super-triangle has also been replaced with ghost vertices at infinity. The incircle tests for ghost-vertices collaps to half-space/orientation tests. It's the canonical super-triangle implementation that avoids interference of the super-triangle with the triangulation.

A `ShewchukExactPredicate` object is not thread-safe in itself. `ShewchukExactPredicate` will allocate scratch buffer on demand, which will not be deallocated. At worst, if all code-pathes are excersised, ~280KB of buffers need allocation. Tho worst offender by far is the insphere test that allocates itself 260KB.

For colinear and other difficult inputs the master version does not produce correct results, so performance is not comparable in those cases. The PR baseline also fails on nearly colinear triangles, and only this implementation produces a valid triangulation there.

```
Input: uniform  
┌──────┬─────────────────┬─────────────────┬────────────┐
│ size │  pr7528 µs/op   │    new µs/op    │ new/pr7528 │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 100  │ 41.43 ±  4.92   │ 47.10 ±  1.79   │ 1.14×      │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 1000 │ 2747.78 ± 28.02 │ 2740.51 ±  9.45 │ 1.00×      │
└──────┴─────────────────┴─────────────────┴────────────┘

Input: grid
┌──────┬─────────────────┬─────────────────┬────────────┐
│ size │  pr7528 µs/op   │    new µs/op    │ new/pr7528 │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 100  │ 42.72 ±  3.10   │ 58.58 ±  1.14   │ 1.37×      │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 1000 │ 2825.87 ± 27.10 │ 2894.07 ±  6.86 │ 1.02×      │
└──────┴─────────────────┴─────────────────┴────────────┘

Input: clustered
┌──────┬─────────────────┬─────────────────┬────────────┐
│ size │  pr7528 µs/op   │    new µs/op    │ new/pr7528 │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 100  │ 52.39 ± 10.72   │ 43.48 ±  1.94   │ 0.83×      │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 1000 │ 2487.11 ± 23.15 │ 2432.94 ±  3.99 │ 0.98×      │
└──────┴─────────────────┴─────────────────┴────────────┘

Input: cocircular
┌──────┬─────────────────┬─────────────────┬────────────┐
│ size │  pr7528 µs/op   │    new µs/op    │ new/pr7528 │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 100  │ 77.38 ±  0.85   │ 105.85 ±  0.55  │ 1.37×      │
├──────┼─────────────────┼─────────────────┼────────────┤
│ 1000 │ 6711.50 ± 21.12 │ 9295.21 ± 72.90 │ 1.38×      │
└──────┴─────────────────┴─────────────────┴────────────┘
```

For standard point distributions (uniform and clustered) performance is comparable with the PR baseline. Clustered is actually ~17% faster than the pr baseline for n=100, while uniform is ~14% slower at that size. 
On a n=100 grid performance degrades by ~37% compared to the original PR. For n=1000 the difference is negligible. On fully cocircular inputs the new implementation is consistently ~38% slower than the PR baseline.

So the conclusion is, the shewchuk implementation performs worse on specifically constructed difficult inputs. This comes as no surprise. What the implementation wins is guaranteed correctness for all inputs and performance improvements in clustered configurations. Performance differences usually vanish as n increases. The exception is when all points lie on a single circle, which should basically never happen anyway, evident by the master version not working correctly on those in the first place.

@NathanSweet fyi, since you reviewed the original PR too.


tommyettinger (commented):
This is a lot of work! Thanks for putting in so much effort.

I'm somewhat concerned about the conversions from float to double, which I'm not entirely sure are always accurate. The Shewchuk algorithm looks like it's correct for doubles, but I don't know if that means we can assume it's also correct for float inputs.

I'm not really concerned about the amount of memory this allocates because it only can do it at most once, and it's a comparable amount of memory to one 256x256 RGBA8888 Pixmap.

The test... I'm not very confident in. The way it uses java.util.Random never generates subnormal floats or doubles, if I'm reading it correctly, and those are the values that are especially likely to have problem behavior. java.util.Random is a truncated linear congruential generator, which suffers from hyperplane correlation issues (Marsaglia's Theorem), but I don't know how badly it affects the generation for the methods we're using and the order we're calling them in. If it's fine, we can get a lot of points without seeing issues. If it's bad, that means every (x,y,z) triple we get will fall on one of a finite number of planes in 3D space, and will never produce a point not on one of those planes. If the only points that cause problems are not on the generatable planes, then we won't see problems *until user data comes in*. The test for ShewchukExactPredicates against known results is good, but it applies to double inputs if I'm not mistaken, and we are always casting float inputs to double when we go through DelaunayTriangulator.

A possible solution to generate a much wider range of float inputs is to a) use libGDX's RandomXS128, which doesn't have hyperplane correlation issues, and b) use it to generate floats by their bits directly. You can get a very-non-uniformly-distributed float, from the full range of normal and subnormal floats between 0.0f inclusive and 2.0f exclusive, with `NumberUtils.intBitsToFloat(randomXS128.nextInt() & 0x3FFFFFFF)` . For a signed float between -2f and 2f, both exclusive, use `NumberUtils.intBitsToFloat(randomXS128.nextInt() & 0xBFFFFFFF)` These snippets have ranges of just over 1 billion and just over 2 billion possible float results; both can produce 0f, and the second can also produce -0.0f, but neither can produce infinite or NaN results. Most values they produce will be extremely small! java.uti.Random can produce 0.0 and `(float)pow(2.0, -24.0)`, but nothing in between those two. The snippets I gave can produce both 0.0 and the float closest to 0.0, Float.MIN_VALUE, each very rarely (the first snippet produces Float.MIN_VALUE 1 out of 2 to the 30 times). We probably need to test primarily or exclusively with float values, because we have no way of entering doubles into DelaunayTriangulator, if I'm not mistaken.

OK, this is way too much text for a comment! This is good work, I'm just not sure it is as exact as the math would be for only using doubles.

tommyettinger (changes requested):
This currently fails a test when building because ShewchukExactPredicates.java isn't known to GWT, even though it's valid code for GWT to be able to use. The current output includes:

```
> Task :tests:gdx-tests-gwt:compileGwt
   Tracing compile failure path for type 'com.badlogic.gdx.math.DelaunayTriangulator'
Error: ERROR] Errors in 'file:/home/runner/work/libgdx/libgdx/gdx/src/com/badlogic/gdx/math/DelaunayTriangulator.java'
Error:   [ERROR] Line 35: No source code is available for type com.badlogic.gdx.math.ShewchukExactPredicates; did you forget to inherit a required module?
Error: OR] Aborting compile due to errors in some input files
```

This can be fixed by adding this line to `gdx/res/com/badlogic/gdx.gwt.xml`, at or around line 380 (alphabetized): 

```xml
		<include name="math/ShewchukExactPredicates.java"/>
```

Berstanio:
> I'm somewhat concerned about the conversions from float to double, which I'm not entirely sure are always accurate. The Shewchuk algorithm looks like it's correct for doubles, but I don't know if that means we can assume it's also correct for float inputs.

Good callout, I checked the spec:, §5.1.2 of JLS covers it:
> A widening primitive conversion does not lose information about the overall magnitude of a numeric value in the following cases, where the numeric value is preserved exactly: from float to double in a strictfp expression

So it is technically not guarenteed pre java 17, but realistically I think all hardware we care about guarentees it (correct me if I'm wrong). We could add strictfp, but not sure if its really needed. Source:  https://docs.oracle.com/javase/specs/jls/se10/html/jls-5.html

> The test... I'm not very confident in. The way it uses java.util.Random never generates subnormal floats or doubles, if I'm reading it correctly, and those are the values that are especially likely to have problem behavior. java.util.Random is a truncated linear congruential generator, which suffers from hyperplane correlation issues (Marsaglia's Theorem), but I don't know how badly it affects the generation for the methods we're using and the order we're calling them in.

Here I was, thinking random should be random enough. Thank you for the review, I added new tests based on your suggestions!

gdx/test/com/badlogic/gdx/math/ShewchukExactPredicatesTest.java
```diff
@@ -0,0 +1,230 @@
...
+			double result = p.orient3d(ax, ay, az, bx, by, bz, cx, cy, cz, dx, dy, dz);
+			assertEquals("orient3d fixture: " + result, -sign, (int)Math.signum(result));
+			assertEquals("symmetry", (int)Math.signum(result),
+				(int)Math.signum(p.orient3d(dx, dy, dz, bx, by, bz, ax, ay, az, cx, cy, cz)));
+		}
+		// 1000 hard fixtures
+
+		double tol = 5.0e-14;
+		for (int i = 0; i < 1000; i++) {
+			double ax = 0.5 + tol * Math.random();
```
tommyettinger: Math.random() uses one ThreadLocalRandom(), which is the same as a java.util.Random in algorithm. If there was hyperplane correlation before, there still is here. It's probably not a problem; LCGs are supposed to be better than the (terrible) MCGs that were used in the 1970s when Marsaglia found issues with them. Truncated LCGs, better still. So we're probably fine, and 1000 points is probably not enough to notice the absence of some small areas.

tommyettinger: Whoops, was reading the wrong file. I was wondering why it looked so different... The checks against known conditions are probably sufficient on their own, and the random checks just add certainty.

gdx/test/com/badlogic/gdx/math/DelaunayTriangulatorTest.java
```diff
@@ -0,0 +1,545 @@
...
+
+	private static float nextFloat (float max) {
+		return nextSign() * nextPositiveFloat(max);
+	}
+
+	private static float nextPositiveFloat (float max) {
+		return rnd.nextFloat() * max;
+	}
+
+	private static float nextFloatBits () {
```
tommyettinger, line 112: Looks good to me. Just keep in mind that while this can produce floats in the `[1.0, 2.0)` range, it does so quite rarely (I believe 1/128 floats should be 1.0 or higher), so if anything depends on that happening often enough, it might need more trials.

tommyettinger (dismissed):
If the tests keep passing, and strictfp isn't needed on modern hardware, I have no more concerns about the math using double cast from float. I'm a little worried about how we could possibly fix any bugs in the 1000+ line Shewchuk file, but the existing Delaunay one wasn't simple either, and this at least has a paper and existing implementation we can double-check with.

gdx/test/com/badlogic/gdx/math/DelaunayTriangulatorTest.java
Berstanio: Ahh, I see! I ran all the tests 10 million times on my local machine without test failure, so I think it _should_ be good then.

Berstanio:
> and this at least has a paper and existing implementation we can double-check with.

Yes, it has actually multiple papers with multiple implementations across various languages, with also active ongoing research: https://arxiv.org/abs/2208.00497. The JS version this port is based of has according to NPM ~18 Million weekly downloads (https://www.npmjs.com/package/robust-predicates), so it's also very battle tested.

NathanSweet:
mourner/robust-predicates is Unlicense, so fine to put under Apache 2.0.

DelaunayTriangulator returns CW triangles, this PR breaks that. Can you add winding tests? We should be able to produce CW directly by making the ghost CW and test incircle so negative is inside.
```
triangles.add(end);
triangles.add(end + 4);
triangles.add(end + 2);
```

Random tests make pass/fail random. Fuzz testing is good but we should also use fixed seeds. Any time fuzz fails, add it to a list of seeds (or record input).

Can we remove ShewchukExactPredicates `instance`? DelaunayTriangulator uses its own instance and app code can manage others. If we want to support sharing instances DelaunayTriangulator could have a constructor param.

`verticie` is an interesting word, but should be `vertices`. ;) (or less commonly `vertexes` which would make logical sense but sense is disallowed in English)

DelaunayTriangulator class javadoc still says "Adapted from Paul Bourke's triangulate", add your `@author`.

ShewchukExactPredicates is complex enough it's nearly opaque for the time I have. Making reliable decisions is great. Being slow in some scenarios is a minor bummer, but correctness is likely worth it.

Memory usage is probably a lot less than mentioned since it allocates lazily and we only use the 2D parts. FWIW, more was ported than we need for Delaunay. We could call it Predicates2D, omit `orient3d`/etc, possibly make it private, the benefit being just less code to maintain.

If looking for more performance, triangles left of the sweep line still pay for `incircle`. If we had a rightmost X per tri then `if (x > right) return COMPLETE`. Use a margin since false positive makes everything wrong. False negative means we pay for `incircle` same as now.

tommyettinger:
Good catch on the typo; singular is vertex, plural is vertices, and English doesn't play by any consistent rules! There's inconsistent casing in ShewchukExactPredicates both for API and internal fields, but if it's consistent with the ported JS it actually makes it easier to read. I don't have a problem with `alllowercase` methods if the names are the same as the original library, and the original library is linked. If it's simple enough, I'm in favor of dropping `instance` and having one possibly-never-constructed ShewchukExactPredicates in the DelaunayTriangulator, since this is quite unlikely to be used anywhere else. The ShewchukExactPredicates class can also be `final`, though it probably doesn't matter.

Berstanio:
> DelaunayTriangulator returns CW triangles, this PR breaks that. Can you add winding tests? We should be able to produce CW directly by making the ghost CW and test incircle so negative is inside.

Ah good catch, somehow thought it should be CCW. Fixed it.

> Random tests make pass/fail random. Fuzz testing is good but we should also use fixed seeds. Any time fuzz fails, add it to a list of seeds (or record input).

Fixed, also for the Shechuk tests.

> Can we remove ShewchukExactPredicates instance? DelaunayTriangulator uses its own instance and app code can manage others. If we want to support sharing instances DelaunayTriangulator could have a constructor param.

I felt like its libGDX idiomatic, but I removed it. Probably not worth to support sharing instances between triangulator instances.

> verticie is an interesting word, but should be vertices. ;) (or less commonly vertexes which would make logical sense but sense is disallowed in English)

Vertex weird word, fixed it, ty!

> DelaunayTriangulator class javadoc still says "Adapted from Paul Bourke's triangulate", add your @author.

Added my author, but keept the "Adapted from" unchanged as it's still a Bowyer-Watson at its core.

> We could call it Predicates2D, omit orient3d/etc, possibly make it private, the benefit being just less code to maintain.

I was debating it, but I would personally think keeping the 3d predicates can be useful in the context of libGDX. afaik exact predicates are more relevant in 3d due to higher degree and more degenerate cases. Briefly skimming libGDX I think e.g. `Intersector#splitTriangle` and `Plane#testPoint` could benefit from exact 3d predicates, if there ever are correctness issues. Or if anyone ever wants to write a 3d delaunay triangulator (not me).
There is also no other implementation of the exact predicates in java that I could find. 


> If looking for more performance, triangles left of the sweep line still pay for incircle. If we had a rightmost X per tri then if (x > right) return COMPLETE. Use a margin since false positive makes everything wrong. False negative means we pay for incircle same as now.

Good idea, I did a quick test and I measured perf increases by ~5% for n = 100, and ~2% for n = 1000. It's measurable, but not sure if worth to do in this PR.

> There's inconsistent casing in ShewchukExactPredicates both for API and internal fields, but if it's consistent with the ported JS it actually makes it easier to read. I don't have a problem with alllowercase methods if the names are the same as the original library, and the original library is linked

Yeah, I tried to follow the original lib as closely as possible.

NathanSweet:
Higher initial allocation would avoid growth copies (tris = point count * 6):
```
triangles.ensureCapacity(count * 3);
edges.ensureCapacity(count);
```

`if (completeArray[completeIndex]) continue;` If many completed triangles are checked could remove them from the unordered `triangles` and not need to iterate/check them at all.

Those may be a little off topic, maybe you don't want to bother. PR looks good as is!

Berstanio:
> Higher initial allocation would avoid growth copies (tris = point count * 6):
> 
> ```
> triangles.ensureCapacity(count * 3);
> edges.ensureCapacity(count);
> ```
> 
> `if (completeArray[completeIndex]) continue;` If many completed triangles are checked could remove them from the unordered `triangles` and not need to iterate/check them at all.

I implemented and pushed it, and yes, the speedup is impressive! It's a 2x speedup on regular input.
```md
Input: uniform
┌──────┬───────────────────┬───────────────────┬──────────────┐
│ size │  preCache µs/op   │     new µs/op     │ new/preCache │
├──────┼───────────────────┼───────────────────┼──────────────┤
│  100 │   75.44 ±   51.66 │   49.09 ±    2.95 │ 0.65×        │
├──────┼───────────────────┼───────────────────┼──────────────┤
│ 1000 │ 4519.52 ± 1589.39 │ 2155.95 ±   14.48 │ 0.48×        │
└──────┴───────────────────┴───────────────────┴──────────────┘

Input: grid
┌──────┬──────────────────┬──────────────────┬──────────────┐
│ size │  preCache µs/op  │    new µs/op     │ new/preCache │
├──────┼──────────────────┼──────────────────┼──────────────┤
│  100 │   98.74 ±  33.42 │   77.33 ±   3.63 │ 0.78×        │
├──────┼──────────────────┼──────────────────┼──────────────┤
│ 1000 │ 4877.86 ± 143.40 │ 2658.67 ±  95.35 │ 0.55×        │
└──────┴──────────────────┴──────────────────┴──────────────┘

Input: clustered
┌──────┬──────────────────┬──────────────────┬──────────────┐
│ size │  preCache µs/op  │    new µs/op     │ new/preCache │
├──────┼──────────────────┼──────────────────┼──────────────┤
│  100 │   73.76 ±  25.48 │   46.18 ±   0.81 │ 0.63×        │
├──────┼──────────────────┼──────────────────┼──────────────┤
│ 1000 │ 3999.16 ± 440.09 │ 1779.48 ±  28.21 │ 0.44×        │
└──────┴──────────────────┴──────────────────┴──────────────┘

Input: cocircular
┌──────┬───────────────────┬───────────────────┬──────────────┐
│ size │  preCache µs/op   │     new µs/op     │ new/preCache │
├──────┼───────────────────┼───────────────────┼──────────────┤
│  100 │   166.76 ±  66.83 │   147.21 ±   5.61 │ 0.88×        │
├──────┼───────────────────┼───────────────────┼──────────────┤
│ 1000 │ 14531.26 ± 435.50 │ 12771.28 ± 616.11 │ 0.88×        │
└──────┴───────────────────┴───────────────────┴──────────────┘

Input: nearlyCollinear
┌──────┬───────────────────┬───────────────────┬──────────────┐
│ size │  preCache µs/op   │     new µs/op     │ new/preCache │
├──────┼───────────────────┼───────────────────┼──────────────┤
│  100 │   163.54 ±   5.58 │   140.86 ±   6.41 │ 0.86×        │
├──────┼───────────────────┼───────────────────┼──────────────┤
│ 1000 │ 15507.27 ± 511.92 │ 13330.79 ± 356.98 │ 0.86×        │
└──────┴───────────────────┴───────────────────┴──────────────┘
```

> Those may be a little off topic, maybe you don't want to bother. PR looks good as is!

The perf optimisation was really worth it, but I think I would personally would want to stop here. I'm a bit exhausted working on the triangulator, getting it to this point was a bumpy road! 2x perf improvement for the average case + guarenteed correctness on all inputs is a good stop point for me for now I think! So in my opinion it can be merged, if the latest changes look good too!

tommyettinger (dismissed):
If it's faster, it's correct, and it's "maintainable enough," I'm all in favor of a merge! Maybe a CHANGES entry could be added; though it isn't a breaking change, it's a significant improvement to correctness and performance. It might also help people be aware during profiling that a small spike in double array allocations is actually a good thing here, since it means we're using a correct algorithm!

CHANGES
```diff
@@ -1,5 +1,8 @@
 [1.14.2]
 - Fixed stuck over state when ClickListener is cancelled.
+- Added ShewchukExactPredicates, for floating-error resistent orient2d/incircle/orient3d/insphere tests
```
tommyettinger: Hehe, "resistent" should be "resistant" 🙂

tommyettinger: Other than that I'm all in favor of a merge, still!

NathanSweet:
Now we're cooking, good job! I did a quick pass, minor edits.

0xDE57:
Wow. Thanks all for the great work on resolving this! 

I look forward to abusing it. :)

PR #7802, opened 2026-05-27
Update ubuntu 18 to 26 in Dockerfile

Project516:
ubuntu 18.04 is EOL


Frosty-J:
glibc version is tied to the OS. We are using 18.04 on purpose so that users don't require the latest version to run libGDX.

PR #7803, opened 2026-05-27
BitmapFontCache.clear() should reset glyphCount.

### LINKED ISSUE #7793: BitmapFontCache.setColors(Color tint,  int start,  int end) only works one time
### Describe the bug

After calling setText() and setColors() the colors get correctly set - but only once. 
When calling setText() again, and then setColors(Color, int, int), the color is the base color.

note this only happens when using setColors() with a start and end argument. The variant where only a color is passed in for the entire text works.

### Reproduction

String text1 = "abc";
String text2 = "def";

font_cache.setText(text1, 100, 100);
font_cache.setColors(Color.RED, 0, 1);
font_cache.setColors(Color.RED, 1, text1.length());

// all is well

font_cache.setText(text2, 100, 100);
font_cache.setColors(Color.RED, 0, 1);
font_cache.setColors(Color.BLACK, 1, text2.length());

// colors not set

font_cache.setColors(Color.RED);

// color set, but for the entire text

### version

1.14

### Stack trace

```shell

```

### Affected platforms

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [x] Windows
- [ ] Linux
- [ ] macOS

tommyettinger:
tl;dr This is a one-line change that fixes a missing reset of mutable state in BitmapFontCache.

Without this change, every time the BitmapFontCache is cleared and text is added again, the items in pageGlyphIndices get larger every time, which breaks some color handling at the very least. You can reproduce the bug by changing line 98 of BitmapFontTest from `true` to `false`; without this PR and with that line change, the cached multi-color text doesn't render at all. With this PR and that line change, it renders correctly. This fixes #7793 .

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFontCache.java
@@ -310,6 +310,7 @@ public class BitmapFontCache {
 		y = 0;
 		pooledLayouts.flush();
 		layouts.clear();
+		glyphCount = 0;
 		for (int i = 0, n = idx.length; i < n; i++) {
 			if (pageGlyphIndices != null) pageGlyphIndices[i].clear();
 			idx[i] = 0;


tommyettinger:
This one was tricky to find because the bug affecting #7793 was actually not in setColors(), but in clear()! Without this change, the pageGlyphIndices values just go up and up, and never end up referring to valid indices after the first frame or so. The test would have caught it, but it had an `if(true)` block that always ran a different type of test, and if that is changed to `if(false)`, the other test has noticeable problems.

PR #7805, opened 2026-05-28
Revert to set addAll returning void.

NathanSweet:
Reverts unwanted binary incompatibility from #7770.

Discussion: https://github.com/libgdx/libgdx/pull/7770#issuecomment-4565195699

@mgsx-dev

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/ObjectSet.java
+++ b/gdx/src/com/badlogic/gdx/utils/ObjectSet.java
@@ -138,29 +138,27 @@ public class ObjectSet<T> implements Iterable<T> {
 		return true;
 	}
 
-	public boolean addAll (Array<? extends T> array) {
-		return addAll(array.items, 0, array.size);
+	public void addAll (Array<? extends T> array) {
+		addAll(array.items, 0, array.size);
 	}
 
-	public boolean addAll (Array<? extends T> array, int offset, int length) {
+	public void addAll (Array<? extends T> array, int offset, int length) {
 		if (offset + length > array.size)
 			throw new IllegalArgumentException("offset + length must be <= size: " + offset + " + " + length + " <= " + array.size);
-		return addAll(array.items, offset, length);
+		addAll(array.items, offset, length);
 	}
 
-	public boolean addAll (T... array) {
-		return addAll(array, 0, array.length);
+	public void addAll (T... array) {
+		addAll(array, 0, array.length);
 	}
 
-	public boolean addAll (T[] array, int offset, int length) {
+	public void addAll (T[] array, int offset, int length) {
 		ensureCapacity(length);
-		int oldSize = size;
 		for (int i = offset, n = i + length; i < n; i++)
 			add(array[i]);
-		return oldSize != size;
 	}
 
-	public boolean addAll (ObjectSet<T> set) {
+	public void addAll (ObjectSet<T> set) {
 		ensureCapacity(set.size);
 		int oldSize = size;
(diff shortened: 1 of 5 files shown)


gdx/src/com/badlogic/gdx/utils/ObjectSet.java
```diff
@@ -138,37 +138,34 @@ public boolean add (T key) {
...
+	public void addAll (Array<? extends T> array, int offset, int length) {
 		if (offset + length > array.size)
 			throw new IllegalArgumentException("offset + length must be <= size: " + offset + " + " + length + " <= " + array.size);
-		return addAll(array.items, offset, length);
+		addAll(array.items, offset, length);
 	}
 
-	public boolean addAll (T... array) {
-		return addAll(array, 0, array.length);
+	public void addAll (T... array) {
```
mgsx-dev: This one wasn't modified in #7770 it was already returning boolean

gdx/src/com/badlogic/gdx/utils/ObjectSet.java
```diff
@@ -138,37 +138,34 @@ public boolean add (T key) {
...
+		addAll(array, 0, array.length);
 	}
 
-	public boolean addAll (T[] array, int offset, int length) {
+	public void addAll (T[] array, int offset, int length) {
```
mgsx-dev: and this one neither

mgsx-dev (changes requested):
I double checked, there are 2 methods to not change in order to keep signatures as it was before. Also, maybe adding something in the CHANGES file would be great : explaining the revert (which is a breaking change regarding 1.14.1) could be useful for people already switched to 1.14.1.

Thank you for the quick revert. I think that's the right decision.

tommyettinger:
Yeah, this bit me in TextraTypist when updating libGDX to 1.14.1 broke TextraTypist's usage of OrderedSet.addAll(). There was also a breaking change in TextFieldListener, which affected tests at least, so maybe I would have had to update anyway.

tommyettinger:
This may also need reverting, not sure when it was last in a release though... https://github.com/libgdx/libgdx/commit/4e54c0a351d4f009a42973d70fdf164128b97c2b

Frosty-J:
> not sure when it was last in a release though

1.14.1. This has been noted in https://github.com/crykn/libgdx-screenmanager/issues/32. While [CHANGES](https://github.com/libgdx/libgdx/blob/master/CHANGES) has more API changes, I think those two are the most likely to break things and the others are pretty safe.

Berstanio:
> This may also need reverting, not sure when it was last in a release though... https://github.com/libgdx/libgdx/commit/4e54c0a351d4f009a42973d70fdf164128b97c2b

~~Are libraries known to use it? I think changing signatures is more fine, if we expect only direct user code to utilise it (not sure if thats the case here).~~

EDIT: Has been made obsolete by the previous reply

NathanSweet:
Should be good now.

Not sure if the similar breaking changes in InputMultiplexer should revert too.

mgsx-dev (dismissed):
@NathanSweet as you can see in linked issues InputMultiplexer changes already break  one of @crykn library. And as i can see, these changes doesn't add so much value (similar to addAll), so same pros and cons i mentionned earlier applies.  I think we should revert these too (probably in another PR).

NathanSweet:
Each PR adds overhead, I'd rather do it here. I guess this is good to merge then.

Should we do a 1.14.2 release soon to minimize user impact?

PR #7808, opened 2026-05-31
Fix lwjgl3 OGG memory leak

### LINKED ISSUE #7712: LWJGL3 ogg loader in 1.12.1+ leaks native memory
Steps to reproduce:
1. Setup sample project e.g. with gdx-liftoff
2. Replace core application with
```java
package com.example;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.assets.AssetDescriptor;
import com.badlogic.gdx.assets.AssetManager;
import com.badlogic.gdx.audio.Sound;

public class Main extends ApplicationAdapter {
    @Override
    public void create() {
        var assetManager = new AssetManager();
        var descriptor = new AssetDescriptor<>("Example.ogg", Sound.class);
        int i = 0;
        // load and unload Example.ogg infinitely
        while(true) {
            if(i++ % 100 == 0) {
                System.out.println(Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory());
            }
            assetManager.load(descriptor);
            assetManager.finishLoading();
            assetManager.unload(descriptor.fileName);
        }
    }
}
```
3. Add some sound as Example.ogg (I used https://commons.wikimedia.org/wiki/File:Example.ogg)
4. Run with -Xmx64M

With libgdx 1.12.1/1.13.5, the printed (JVM heap) memory usage stays bounded but the process memory usage grows unbounded (10GB+ after a minute).
With 1.12.0, process memory usage stays low (~150MB).

Based on the version I suspect this is related to #7140.

Berstanio:
Fixes https://github.com/libgdx/libgdx/issues/7712

Free pattern derived from lwgjl3 docs:
```java
    /**
     * Decode an entire file and output the data interleaved into a {@code malloc()ed} buffer stored in {@code *output}. When you're done with it, just
     * {@link LibCStdlib#free} the pointer returned in {@code *output}.
     *
     * @param filename    the file name
     * @param channels    returns the number of channels
     * @param sample_rate returns the sample rate
     *
     * @return the number of samples decoded, or -1 if the file could not be opened or was not an ogg vorbis file
     */
    @Nullable
    @NativeType("int")
    public static ShortBuffer stb_vorbis_decode_filename(@NativeType("char const *") CharSequence filename, @NativeType("int *") IntBuffer channels, @NativeType("int *") IntBuffer sample_rate) 
[...]

    /**
     * In-memory version of {@link #stb_vorbis_decode_filename decode_filename}.
     *
     * @param mem         the data to decode
     * @param channels    returns the number of channels
     * @param sample_rate returns the sample rate
     */
    @Nullable
    @NativeType("int")
    public static ShortBuffer stb_vorbis_decode_memory(@NativeType("unsigned char const *") ByteBuffer mem, @NativeType("int *") IntBuffer channels, @NativeType("int *") IntBuffer sample_rate) {
```
Verified it does not leak memory enymore.

The allocated DirectByteBuffer has been replaced with an explicit lifecycle tracked.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/Ogg.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/Ogg.java
@@ -17,7 +17,6 @@
 package com.badlogic.gdx.backends.lwjgl3.audio;
 
 import com.badlogic.gdx.files.FileHandle;
-import com.badlogic.gdx.utils.BufferUtils;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.StreamUtils;
 import org.lwjgl.stb.STBVorbis;
@@ -82,8 +81,7 @@ public class Ogg {
 					final IntBuffer sampleRateBuffer = stack.mallocInt(1);
 
 					// decode
-					final ShortBuffer decodedData = STBVorbis.stb_vorbis_decode_memory(encodedData, channelsBuffer,
-							sampleRateBuffer);
+					final ShortBuffer decodedData = STBVorbis.stb_vorbis_decode_memory(encodedData, channelsBuffer, sampleRateBuffer);
 					if (decodedData == null) {
 						throw new GdxRuntimeException("Error decoding OGG file: " + file);
 					}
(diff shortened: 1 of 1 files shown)


tommyettinger (approved):
If you verified it doesn't leak memory and it's the same pattern LWJGL3 recommends, I see no reason not to use this! Approved.

PR #7810, opened 2026-06-01
Change default BULLET char from Windows-1252 to Unicode.

### LINKED ISSUE #7809: Legacy Windows codepage use in TextField
### Describe the bug

https://github.com/libgdx/libgdx/blob/04644580eb78b98f64c0edcb333d34a6d631f51e/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L74

Umm, don't all Java platforms use Unicode nowadays? That's a bullet in Windows-1252, but a "Start of Guarded Area" control in Unicode. Point in fact, that renders as missing glyph indicator when the app uses a dynamic font from glyphs rendered via AWT (or via `android.graphics.Canvas.drawText`), like Unicv does.

This should be U+2022.

### Reproduction

<details><summary>Screenshots</summary>

Not doing a scce, but here's Unciv **with** `setPasswordCharacter('\u2022')`:
Image

... and without:
Image

...same on LineageOS 19:
Image

</details>

### version

1.14.0

### Stack trace

```shell

```

### Affected platforms

- [x] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [x] Linux
- [ ] macOS

tommyettinger:
This changes the default char used in password fields from code point 149 or 0x0095, which is not a displayable Unicode char, to Unicode code point 0x2022 . This also documents that a zero-arg constructor BitmapFont can't display that char by default, and how to make it display correctly in that common case. I tried to avoid using any non-ASCII chars in the Javadocs, because those can sometimes display incorrectly if users have their fonts set up a certain way.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -71,7 +71,11 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 	static protected final char NEWLINE = '\n';
 	static protected final char TAB = '\t';
 	static protected final char DELETE = 127;
-	static protected final char BULLET = 149;
+	/** This is the default char used in password fields, a Unicode bullet at code point 0x2022. If you need the
+	 * Windows-1252 codepage for legacy support, you can call {@link #setPasswordCharacter(char)} with the char at code
+	 * point 0x0095. Note that the default BitmapFont constructed with {@link BitmapFont#BitmapFont()} does not contain
+	 * any bullet char, so if you use that, you should call {@code field.setPasswordCharacter('*');} instead. */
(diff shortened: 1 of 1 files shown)


Frosty-J:
Looks good to me, tested on Windows host. Doesn't look like anything else is depending on `TextField.BULLET`.

I do find libGDX's default rather peculiar... you can see in `UITest` that it has to set an asterisk as the default character, since there is no bullet in the skin. And the fallback to not hiding characters at all seems not ideal. I don't see why it doesn't use asterisks from the get-go, then anyone who goes through the trouble of adding a bullet character to their font can set that as the password character.

https://github.com/libgdx/libgdx/blob/58f823d26d54ee3405ddc3792530d79470c07822/tests/gdx-tests/src/com/badlogic/gdx/tests/UITest.java#L138-L141

tommyettinger:
It is odd, yeah, but I didn't want to change behavior without a good reason.

PR #7814, opened 2026-06-03
Add putMissing() to maps.

NathanSweet:
I needed LongMap `putMissing()` to do this in 1 lookup:
```
if (!map.containsKey(key)) map.put(key, value); // 2 lookups when missing.
map.putMissing(key, value); // Same with 1 lookup.
```
If you need the existing value and it could be null:
```
existing = map.containsKey(key) ? map.get(key) : map.put(key, value); // 2 lookups.
existing = map.putMissing(key, value); // Same with 1 lookup.
```
It's useful for the other maps, so here we are.

Open to other naming. JDK naming for similar is `putIfAbsent`. Having `If` in a method name isn't great IMHO and the semantics differ: the JDK method stores the value when a key exists storing null. We allow null values and treat them as any other, so `putMissing` is based on whether the key exists, regardless of whether the current value is null.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/IntIntMap.java
+++ b/gdx/src/com/badlogic/gdx/utils/IntIntMap.java
@@ -179,6 +179,42 @@ public class IntIntMap implements Iterable<IntIntMap.Entry> {
 		return defaultValue;
 	}
 
+	public void putMissing (int key, int value) {
+		if (key == 0) {
+			if (!hasZeroValue) {
+				zeroValue = value;
+				hasZeroValue = true;
+				size++;
+			}
+			return;
+		}
+		int i = locateKey(key);
+		if (i >= 0) return; // Existing key was found.
+		i = -(i + 1); // Empty space was found.
+		keyTable[i] = key;
+		valueTable[i] = value;
+		if (++size >= threshold) resize(keyTable.length << 1);
+	}
+
+	public int putMissing (int key, int value, int defaultValue) {
+		if (key == 0) {
+			if (!hasZeroValue) {
+				zeroValue = value;
+				hasZeroValue = true;
+				size++;
+				return defaultValue;
+			}
+			return zeroValue;
+		}
+		int i = locateKey(key);
+		if (i >= 0) return valueTable[i]; // Existing key was found.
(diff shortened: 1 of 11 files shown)


gdx/src/com/badlogic/gdx/utils/IntIntMap.java
```diff
@@ -179,6 +179,42 @@ public int put (int key, int value, int defaultValue) {
...
+		}
+		int i = locateKey(key);
+		if (i >= 0) return; // Existing key was found.
+		i = -(i + 1); // Empty space was found.
+		keyTable[i] = key;
+		valueTable[i] = value;
+		if (++size >= threshold) resize(keyTable.length << 1);
+	}
+
+	public int putMissing (int key, int value, int defaultValue) {
```
tommyettinger, line 199: I really like being able to set a defaultValue as an optional parameter; this is a feature above and beyond what the JDK does, so it also helps warrant the different name. Could the Object-valued Maps also have a putMissing() with a default parameter (so previously-null values could be distinguished)? It would IMO help with API uniformity.

tommyettinger (approved):
I like this as-is but I'd encourage you to add overloads to the Object-value Maps that match the defaultValue ones in primitive-value Maps! I think that might be useful to distinguish null values already held in the Map, or if there's a need for some other default, like "" for String-valued maps.

NathanSweet:
Primitive value map `put` requires `defaultValue` as a sentinel, since there is no null. `get` and `remove` need the same. It's "return value when there is no mapping".

Object value maps return null without an explicit sentinel. That can't differentiate null from "there was no mapping", so if that matters 1 extra lookup is needed. The question is if that is common enough to warrant overloads. I'd lean toward not having them. Passing 2 values to `put` isn't super clear and also overloading `get` and `remove` seems like a lot for a relatively rare use case where 1 extra lookup is the workaround.

tommyettinger:
Yeah, it would be a pretty rarely-used feature, I imagine, but we thought that about putMissing() too before now. I'm not sure what harm there is to a few more overloads, especially now that Android is more capable about handling large APKs. I'm fine either way, though.

Issue #7812, opened 2026-06-01
NativeOnscreenKeyboard behaves different to default, needs more configurability


SomeTroglodyte:
### Description

Essentially, just plugging in `setOnscreenKeyboard(NativeOnscreenKeyboard())` changes the app behaviour. The "go" soft key no longer "goes", it mimics Tab instead, not as keystroke but as TextField.next(). Previously, you'd get an enter key.

Seems intentional:
https://github.com/libgdx/libgdx/blob/253a0c0b0b40c9b3f9cb1ca72ec7e0ef4d74cb05/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L918-L921

I say unacceptable, as long as the only way to change that is to rewrite that class from scratch (not nice working around missing getters where setters exist). An overridable "onConfirm", or a builder-style overridable that allows modifying configuration just before it's used at the end of `private void openNativeInputField`, or somesuch, would be infinitely better. But I do like "dismuss", sounds comfy.

tommyettinger:
I'm not exactly pleased with this set of changes either; it broke vis-ui and I just sent a PR in to there. The problem with vis-ui is different, but it has to do with how each type of OnscreenKeyboard takes a TextField now, and a VisTextField doesn't extend TextField. Nor does TextraField in TextraTypist extend TextField, which I'll need to deal with at some point... I solved it in vis-ui by just passing `null` to `show(TextField)`, because vis-ui always uses DefaultOnscreenKeyboard, which doesn't even use the TextField parameter.

The behavior here also looks broken in a different way. I wonder what the rationale was for the change...? It was added in https://github.com/libgdx/libgdx/pull/7671 .

Berstanio:
> Essentially, just plugging in setOnscreenKeyboard(NativeOnscreenKeyboard()) changes the app behaviour.

Just to be clear, `NativeOnscreenKeyboard` is not a 1-to-1 replacement of `DefaultOnscreenKeyboard`. If you build your app around specific behaviourisms of `DefaultOnscreenKeyboard`, they are not guarenteed to carry over to `NativeOnscreenKeyboard`.

> I say unacceptable, as long as the only way to change that is to rewrite that class from scratch (not nice working around missing getters where setters exist).

Is your complain that `textField.writeEnters` does not have a getter? If yes, we can fix that. Otherwise, rolling your own `OnscreenKeyboard` seems like the solution for your intent.

[...]

PR #7815, opened 2026-06-04
Make NativeOnscreenKeyboard more configurable and make `DefaultOnscreenKeyboard` respect the OnscreenKeyboardType

Berstanio:
Addresses some concerns raised in https://github.com/libgdx/libgdx/issues/7812 (not all).

It adds missing getters, allows an easier way to configure the `NativeOnscreenKeyboard` and implement their own keyboard focus routing.

It also fixes, that `DefaultOnscreenKeyboard` does not respect the selected `OnscreenKeyboardType` of the TextField.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -1055,11 +1055,6 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 
 				@Override
 				public void writeResults (String text, int selectionStart, int selectionEnd) {
-					if (textField.shouldPreventAutoCorrection()) {
-						text = text.trim();
-						selectionStart = Math.min(selectionStart, text.length());
-						selectionEnd = Math.min(selectionEnd, text.length());
-					}
 					textField.setText(text);
 					if (selectionStart == selectionEnd) {
 						textField.setCursorPosition(selectionEnd);
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
@@ -224,15 +224,39 @@ public class TextField extends Widget implements Disableable, Styleable<TextFiel
 		this.autocompleteOptions = autocompleteOptions;
 	}
 
+	/** A list of possible strings that are valid to autocomplete Currently only has an effect on mobile with
+	 * {@link NativeOnscreenKeyboard} */
+	public String[] getAutocompleteOptions () {
+		return autocompleteOptions;
+	}
+
 	/** Which {@link OnscreenKeyboardType} to use. Mainly used for mobile, will also be referenced for password masking */
-	public void setKeyboardType (Input.OnscreenKeyboardType keyboardType) {
+	public void setKeyboardType (OnscreenKeyboardType keyboardType) {
 		this.keyboardType = keyboardType;
 	}
 
-	/** Whether if auto correction is provided by the system, it should be surpressed Will be considered true, if
+	/** Which {@link OnscreenKeyboardType} to use. Mainly used for mobile, will also be referenced for password masking. `Default`
+	 * will be promoted to `Password` if `isPasswordMode` is true */
+	public OnscreenKeyboardType getResolvedKeyboardType () {
+		// If we are in password mode, assume that the user want to show password keyboard
+		return isPasswordMode() && keyboardType == OnscreenKeyboardType.Default ? OnscreenKeyboardType.Password : keyboardType;
+	}
+
+	/** Whether if auto correction is provided by the system, it should be surpressed. Will be considered true, if
 	 * {@link OnscreenKeyboardType} is `Password` */
 	public void setPreventAutoCorrection (boolean preventAutoCorrection) {
 		this.preventAutoCorrection = preventAutoCorrection;
+	}
+
+	/** Whether if auto correction is provided by the system, it should be suppressed. Will be considered true, if
+	 * {@link TextField#getResolvedKeyboardType()} is `Password` */
+	public boolean shouldPreventAutoCorrection () {
+		return preventAutoCorrection || getResolvedKeyboardType() == OnscreenKeyboardType.Password;
+	}
+
+	/** Whether the TextField does write enters to the text. */
+	public boolean isWriteEnters () {
+		return writeEnters;
 	}
 
 	/** When false, text set by {@link #setText(String)} may contain characters not in the font, a space will be displayed instead.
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java
```diff
@@ -922,7 +946,7 @@ static public interface OnscreenKeyboard {
 	 * @author mzechner */
 	static public class DefaultOnscreenKeyboard implements OnscreenKeyboard {
 		public void show (TextField textField) {
-			Gdx.input.setOnscreenKeyboardVisible(true);
+			Gdx.input.setOnscreenKeyboardVisible(true, textField.getResolvedKeyboardType());
```
tommyettinger, line 949: Oh. Great, this breaks the one way I could avoid a dependency on TextField directly while still using DefaultOnscreenKeyboard. Before, I could pass `null` as the `textField`, and now that throws an NPE. I guess it shouldn't be too hard to implement my own OnscreenKeyboard in vis-ui and TextraTypist, but it can't do this behavior because neither of those extends TextField... It's like this show() change gets worse and worse the more times I look at it.

tommyettinger: It looks like my only way out is to make my own OnscreenKeyboard interface in vis-ui and in TextraTypist, which runs counter to the point of an interface. This leaves a really bad taste in my mouth.

tommyettinger (changes requested):
I don't like this at all. That includes the backward-incompatible change to show() to take a TextField in the first place.

I think this whole experiment needs to be canceled, the changes to show() reverted to 1.14.0's version, and all this thought through really, really thoroughly before we start up again. There is probably a good API possible for on-screen keyboards; the airtight coupling with scene2d.ui.TextField does not seem like a good API.

The fact that there even exist two (and probably more) incompatible re-implementations of TextField, to me, suggest that scene2d.ui's existing classes aren't sufficient for a lot of users. Having the API be extensible should really be a priority, similar to how Action in scene2d is designed to be flexible for end-users. Locking users into TextField and its direct subclasses is just not flexible.

You could pass the next TextField, or some focus resolver that can locate the next focused widget, to the constructor of a Default or Native OnscreenKeyboard, and then show() could just go back to taking a boolean for open or closed. You could have a top-level interface that defines only the behavior that on-screen keyboards need to function, such as being able to get an `Input.OnscreenKeyboardType`, and then other libraries could implement that single interface in their types that need to work with text input. Then, instead of forcing a TextField (or subclass), you could require only that the widget you're showing implements that interface. That could be used within scene2d.ui if other widgets turn out to need text input.

Another option is to have a different, small OnscreenKeyboard implementation for each different OnscreenKeyboardType enum constant, which seems somewhat reasonable to me; they behave differently and show different keys. What isn't reasonable is having three different, incompatible, OnscreenKeyboard interfaces nested in three (or more) different text-field classes, with a VisTextField from vis-ui unable to switch focus to a TextField from scene2d.ui unable to switch to a TextraField from TextraTypist...

Berstanio:
First of all, I would like to point out that the class `OnscreenKeyboard` lives inside `TextField`, next to interfaces like `TextFieldFilter` and `TextFieldListener`. The other interfaces also do have a `TextFIeld` dependencency, I feel like the dependency to `TextField` is natural. If we deliberatly remove the dependency, the `OnscreenKeyboard` interface feels genuinly misplaced. I also don't get, why copying a 2-method interface is a dealbreaker. 
That being said, we can ofc try to make it better.

> You could pass the next TextField, or some focus resolver that can locate the next focused widget, to the constructor of a Default or Native OnscreenKeyboard, and then show() could just go back to taking a boolean for open or closed.

That would mean having to call `setOnscreenKeyboard` on every `TextField` you create, which is very inconvinient in my opinion. The global `DEFAULT_ONSCREEN_KEYBOARD` exists, because OnscreenKeyboards can be mutually exclusive (e.g. NativeIOnscreenKeyboard and DefaultOnscreenKeyboard).
And furthermore, any reasonable implementation of `OnscreenKeyboard` needs to know _something_ about the `TextField`. Let alone be it what keyboard type to show or if we are dealing with a password. So theoretically all `OnscreenKeyboard` would need the same resolver logic.

> You could have a top-level interface that defines only the behavior that on-screen keyboards need to function, such as being able to get an Input.OnscreenKeyboardType, and then other libraries could implement that single interface in their types that need to work with text input.

I would be theoretically fine with this. However, the API surface would be huge and not trivial:
`getResolvedKeyboardType`
`isPasswordMode`
`getMaxLength`
`isWriteEnters`
`shouldPreventAutoCorrection`
`getMessageText`
`getTextFieldFilter` (which already breaks again, because TextFilter is tied to `TextField`. Not sure how to resolve)
`getText`
`getSelectionStart`
`getCursorPosition`
`setText`
`setCursorPosition`
`setSelection`

and some more in the feature, as the API gets extended. Which then either will be API breaking or implemented using default interface methods, which is also easy to miss then.

It also forces all conventions and ideas of `TextField` onto other `TextField` implementation and forces all conventions/ideas of the NativeInputAPI onto all `OnscreenKeyboard` implementations.

> Another option is to have a different, small OnscreenKeyboard implementation for each different OnscreenKeyboardType enum constant, which seems somewhat reasonable to me; they behave differently and show different keys.

There are _a lot_ options to the `NativeOnscreenKeyboard` that would need passing, I think this is unfeasable. It also seems overkill in regards to code duplication to me.

tommyettinger:
> And furthermore, any reasonable implementation of OnscreenKeyboard needs to know something about the TextField. Let alone be it what keyboard type to show or if we are dealing with a password. So theoretically all OnscreenKeyboard would need the same resolver logic.

How much of this could be resolved by passing an OnscreenKeyboardType instead of a TextField? That's easy enough to get from a TextField or anything else that implements this... I'm trying to figure out some kind of better API, I just think the original 1.14.0 and earlier one was cleaner and more versatile for 3rd-party libraries. I guess it isn't enough for the NativeOnscreenKeyboard, but I still don't understand what the Native one solves. Password input, I guess, wasn't handled before?

Berstanio:
> I guess it isn't enough for the NativeOnscreenKeyboard, but I still don't understand what the Native one solves. Password input, I guess, wasn't handled before?

No, it handles password, auto-complete, auto-correct, copy-paste, assembled characters (like korean), gesture navigation, Speech-To-Text, other accessibility features and more things I probably forgot. The "DefaultOnscreenKeyboard" handles none of those. I think the `NativeOnscreenKeyboard` is genuinly the one that people should try to use in production, if they do anything more complex then "User-login". 
So in my opinion, the `NativeOnscreenKeyboard` should rather drive the design then the `DefaultOnscreenKeyboard`.

> How much of this could be resolved by passing an OnscreenKeyboardType instead of a TextField?

Not much imo. Any reasonable mobile implementation needs ways to write back to the TextField whole chunks of text and be able to set the selection/cursor position, and read both too to keep the IME in sync. 
So at minimum that would be:
> getResolvedKeyboardType
getText
getSelectionStart
getCursorPosition
setText
setCursorPosition
setSelection
shouldPreventAutoCorrection (debatable, but I think auto-correct/suggest is really needed on mobile)

tommyettinger (dismissed):
I'm tentatively approving this; I guess there's no hope of proper interop between third-party text-field widgets with scene2d.ui. I'll be creating lots of interfaces, I suppose, that copy the NativeOnscreenKeyboard code...

PR #7819, opened 2026-06-11
fix(TmxMapLoader): resolve per-tile images relative to .tsx, not .tmx

### LINKED ISSUE #7818: addStaticTiles resolves tile image paths relative to .tmx instead of .tsx
### Issue details

When a `.tmx` references an external `.tsx` tileset using "collection of images" (individual `<tile><image>` entries), `addStaticTiles` resolves those image paths relative to the **.tmx file** — but they should be relative to the **.tsx file** where they're declared.

Tiled itself resolves them correctly (relative to the `.tsx`), so maps work fine in the editor but crash on load in libGDX.

### Reproduction

1. Create a TMX + external TSX tileset where `<tile><image source="...">` points to something relative to the TSX dir (e.g. `../gfx/my_tile.png`)
2. Call `new TmxMapLoader().load("maps/mymap.tmx")`
3. Crash: `IllegalArgumentException: texture cannot be null` at `DirectImageResolver.getImage`

### Root cause

`TmxMapLoader.java` — `addStaticTiles` does:

```java
FileHandle tileImage = getRelativeFileHandle(tmxFile, tileImageSource);
```

Should be using the **tsx** file handle instead. The same code in `getTileSetDependencyFileHandle` already does this correctly:

```java
FileHandle image = getRelativeFileHandle(tsxFile, imageSource); // correct
```

Fixing `addStaticTiles` would mean either passing the tsx handle through, or computing it from the existing `source` parameter:

```java
FileHandle tsxFile = source != null ? getRelativeFileHandle(tmxFile, source) : tmxFile;
FileHandle tileImage = getRelativeFileHandle(tsxFile, tileImageSource);
```

### Version & platform

- libGDX 1.14.2
- Affects all platforms (reproduced on Desktop LWJGL3)

### Stacktrace

```
Exception in thread "main" java.lang.IllegalArgumentException: texture cannot be null.
    at com.badlogic.gdx.graphics.g2d.TextureRegion.<init>(TextureRegion.java:37)
    at com.badlogic.gdx.maps.ImageResolver$DirectImageResolver.getImage(ImageResolver.java:41)
    at com.badlogic.gdx.maps.tiled.TmxMapLoader.addStaticTiles(TmxMapLoader.java:216)
    at com.badlogic.gdx.maps.tiled.BaseTmxMapLoader.loadTileSet(BaseTmxMapLoader.java:807)
    at com.badlogic.gdx.maps.tiled.TmxMapLoader.load(TmxMapLoader.java:77)
```

MRZ07:
Fixes #7818

`addStaticTiles` in both `TmxMapLoader` and `TmjMapLoader` used `getRelativeFileHandle(tmxFile, tileImageSource)` for individual `<tile><image>` paths inside external tileset files. Since those paths are declared in the .tsx/.tsj file, they must be resolved relative to the **tileset file's directory**, not the map file's directory.

The tileset-level image path (`resolveTilesetImage` in both `BaseTmxMapLoader` and `BaseTmjMapLoader`) already handles this correctly using the same pattern — this applies it to per-tile images in `addStaticTiles` of both loaders.

### DIFF
--- a/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
@@ -225,7 +225,9 @@ public class TmjMapLoader extends BaseTmjMapLoader<BaseTiledMapLoader.Parameters
 			for (JsonValue tile : tiles) {
 				if (tile.has("image")) {
 					String tileImageSource = tile.getString("image");
-					FileHandle tileImage = getRelativeFileHandle(tmjFile, tileImageSource);
+					String tilesetSource = element.getString("source", null);
+					FileHandle tsxFile = (tilesetSource != null) ? getRelativeFileHandle(tmjFile, tilesetSource) : tmjFile;
+					FileHandle tileImage = getRelativeFileHandle(tsxFile, tileImageSource);
 					TextureRegion texture = imageResolver.getImage(tileImage.path());
 
 					int tileId = firstgid + tile.getInt("id");
--- a/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/TmjMapLoader.java
@@ -222,12 +222,12 @@ public class TmjMapLoader extends BaseTmjMapLoader<BaseTiledMapLoader.Parameters
 				}
 			}
 		} else {
+			String tilesetSource = element.getString("source", null);
+			FileHandle tilesetFile = (tilesetSource != null) ? getRelativeFileHandle(tmjFile, tilesetSource) : tmjFile;
 			for (JsonValue tile : tiles) {
 				if (tile.has("image")) {
 					String tileImageSource = tile.getString("image");
-					String tilesetSource = element.getString("source", null);
-					FileHandle tsxFile = (tilesetSource != null) ? getRelativeFileHandle(tmjFile, tilesetSource) : tmjFile;
-					FileHandle tileImage = getRelativeFileHandle(tsxFile, tileImageSource);
+					FileHandle tileImage = getRelativeFileHandle(tilesetFile, tileImageSource);
 					TextureRegion texture = imageResolver.getImage(tileImage.path());
 
 					int tileId = firstgid + tile.getInt("id");
--- a/gdx/src/com/badlogic/gdx/maps/tiled/TmxMapLoader.java
+++ b/gdx/src/com/badlogic/gdx/maps/tiled/TmxMapLoader.java
@@ -212,7 +212,9 @@ public class TmxMapLoader extends BaseTmxMapLoader<BaseTiledMapLoader.Parameters
 				Element imageElement = tileElement.getChildByName("image");
 				if (imageElement != null) {
 					String tileImageSource = imageElement.getAttribute("source");
-					FileHandle tileImage = getRelativeFileHandle(tmxFile, tileImageSource);
+					String tilesetSource = element.getAttribute("source", null);
+					FileHandle tsxFile = (tilesetSource != null) ? getRelativeFileHandle(tmxFile, tilesetSource) : tmxFile;
+					FileHandle tileImage = getRelativeFileHandle(tsxFile, tileImageSource);
 					TextureRegion texture = imageResolver.getImage(tileImage.path());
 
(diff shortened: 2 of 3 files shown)


BoBIsHere86:
This bug looks like it was introduced  in commit 284898b, when it was merged for version 1.14.1.

This is PR looks good to me. All tests pass, though we don't actually have a specific test which would break for this exact scenario, which is how it got through to begin with. 

Excellent catch. 💯 

I am going to bug @Quillraven just so someone else can have some eyes on this. But I do believe the root cause for the break has been found. At the commit above.

MRZ07:
Thanks! I added some tests on gdx/test/com/badlogic/gdx/maps/tiled/TmxMapLoaderPathResolutionTest.java

BoBIsHere86:
Someone with a bit more knowledge on what's allowed with libGDX tests will need to chime in here.
But normally with Tiled I don't believe they want them as a part of the core gdx test folder.

The Tiled related tests always were placed under the separate "tests/gdx-tests/" section of the repo, so they could be manually tested from the Lwjgl3TestStarter app. Which in that case you would also need to also add an entry into com.badlogic.gdx.tests.utils.GdxTests.java so it becomes an actual part of the gdx-tests.

Sorry if this sounds complicated, but I'm just trying to get ahead of someone else rejecting your change because of that.

Or maybe not? I'll wait to see what someone else says.

SimonIT:
I personally prefer unit tests over the GdxTest because they can run automatically, but where something visual needs to be observed, we don't have an automatic approach, so GdxTest is needed

Quillraven:
> I personally prefer unit tests over the GdxTest because they can run automatically, but where something visual needs to be observed, we don't have an automatic approach, so GdxTest is needed

I totally agree with your unit test comment but afaik all Tiled tests so far are part of GdxTest. I remember specifically one of my PRs which added support for class type properties, which could be run as a unit test but iirc I couldn't find anything for Tiled and added it to GdxTest. If background renders green it passes and if it is red it fails.

But I was always wondering who is manually testing all those GdxTest scenarios before a release. I guess noone? 😅


Regarding the PR itself I try to have a look within this week!

Berstanio:
> I totally agree with your unit test comment but afaik all Tiled tests so far are part of GdxTest. I remember specifically one of my PRs which added support for class type properties, which could be run as a unit test but iirc I couldn't find anything for Tiled and added it to GdxTest. If background renders green it passes and if it is red it fails.

In my opionion those could be just converted to unit tests then. As you guessed yourself, I don't think anyone really runs a GdxTest unless they edit the specific part of the codebase the test covers.

> Someone with a bit more knowledge on what's allowed with libGDX tests will need to chime in here.
But normally with Tiled I don't believe they want them as a part of the core gdx test folder.

As Tiled lives in the core gdx module, imo it's fine to put it into the gdx core unit tests. The difference is imo mostly "does this test need user input and/or observation?" or "is this behaviour platofrm dependent", which determines whether it should land as a unit test or GdxTest.

Quillraven:
PR looks fine to me. One nitpicky thing ... I see the test for XML based Tiled files but we don't have tests for the JSON files, right?  Also, the test creates the XML files via code. If for whatever reason Tiled decides to update the content, then the test won't cover that. Granted, even if it is a file we'd need to save it with the new Tiled version but the advantage is it uses a real file produced by Tiled.

Not sure what the best approach would be here because even if there are breaking changes to the format of files then most likely we'd want to support the new and old format for a specific period.

Quillraven (approved):
see the comment above :)

BoBIsHere86:
> But I was always wondering who is manually testing all those GdxTest scenarios before a release. I guess noone? 😅

Well, for what it's worth. Ever since I started bullying people about the Tiled PRs a few years ago. I have gone in and manually run every single tiled test for every PR I looked at. It's not perfect but I have a good idea of what the maps in the tests "should" look like. So if I ever notice that something looks off or crashes on load, that jumps out as a red flag.

But for a test like this PR above, it's not really a visual test. It either loads or it doesn't so there really isn't a need for a test in the GDXTest scenarios. So a unit test for something like this looks totally acceptable.

MRZ07:
Thanks for the review @Quillraven! The nitpick was spot on, I added 2 JSON (TSJ/TMJ) tests covering the same scenarios as the XML ones:

1. external tileset resolution (../ts/sprites.tsj → ts/img/tile.png) 
2. embedded tileset fallback. Both pass alongside the XML equivalents.

PR #7826, opened 2026-06-30
Handle unavailable primary monitor when centering window

### LINKED ISSUE #7794: `NullPointerException` within `Lwjgl3Application` startup when monitor is detached
### Describe the bug

# Problem: 

`Lwjgl3Application.java` tries to center the window if not fullscreen and not using a pre-specified window size. If the monitor is detached (for example: this happens on MacOS if you lock your screen), it throws a NPE. In reality, it should fail more gracefully than this (perhaps by simply not trying to position the window).

In practice, this can create problems for users that play on a mac. If they launch the game, but immediately lock their screen / close their laptop to walk away, in the gap before the system suspends the application for idle mode, the game can crash with this NPE and it is outside of the developer's hands.

# Expected behavior:

Query the GLFW monitor first to ensure it exists before trying to position the window, and fail more gracefully (like choosing 0,0) in the event the monitor doesn't exist.

# Trace:
```
java.lang.NullPointerException
	at org.lwjgl.system.Checks.check(Checks.java:188)
	at org.lwjgl.glfw.GLFW.nglfwGetMonitorPos(GLFW.java:1432)
	at org.lwjgl.glfw.GLFW.glfwGetMonitorPos(GLFW.java:1455)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3ApplicationConfiguration.toLwjgl3Monitor(Lwjgl3ApplicationConfiguration.java:313)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createGlfwWindow(Lwjgl3Application.java:557)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:451)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:437)
at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:149)
```

### Reproduction

Conditions:
- Use an apple silicon mac (I am not sure if the graphics power management behavior is the same on the older x86 macs)
- Lwjgl3ApplicationConfiguration: Not fullscreen, no window position specified.

To reproduce: 
- launch a Lwjgl3Application and lock the screen immediately (ctrl + command + q), so that it starts the application startup while the screen is locked. This will result in the NullPointerException. A gradle task should be enough of a delay to pull this off, but if not, you can add some sleeps or something before creating the application (to emulate some kind of startup hook that a player may experience).

### version

1.14.0

### Stack trace

```shell
java.lang.NullPointerException
	at org.lwjgl.system.Checks.check(Checks.java:188)
	at org.lwjgl.glfw.GLFW.nglfwGetMonitorPos(GLFW.java:1432)
	at org.lwjgl.glfw.GLFW.glfwGetMonitorPos(GLFW.java:1455)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3ApplicationConfiguration.toLwjgl3Monitor(Lwjgl3ApplicationConfiguration.java:313)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createGlfwWindow(Lwjgl3Application.java:557)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:451)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.createWindow(Lwjgl3Application.java:437)
at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:149)
```

### Affected platforms

- [ ] Android
- [ ] iOS
- [ ] HTML/GWT
- [ ] Windows
- [ ] Linux
- [x] macOS

Kyrixen:
Fixes #7794

According to the GLFW documentation, `glfwGetPrimaryMonitor()` returns `NULL` if no monitors are found or if an error occurs. This change guards against that case:
https://www.glfw.org/docs/latest/group__monitor.html#gac3adb24947eb709e1874028272e5dfc5

This change checks whether the primary monitor handle is `MemoryUtil.NULL` before attempting to center the window, preventing an invalid handle from being passed to `toLwjgl3Monitor()`.

When the primary monitor is unavailable, the window falls back to position `(0, 0)`.

I couldn't reproduce the macOS-specific scenario because I don't have access to macOS hardware, so I'd appreciate verification from someone who can reproduce the issue.

### DIFF
--- a/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
+++ b/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java
@@ -27,7 +27,6 @@ import com.badlogic.gdx.backends.lwjgl3.audio.Lwjgl3Audio;
 import com.badlogic.gdx.backends.lwjgl3.audio.OpenALLwjgl3Audio;
 import com.badlogic.gdx.graphics.glutils.GLVersion;
 
-import com.badlogic.gdx.utils.*;
 import org.lwjgl.glfw.GLFW;
 import org.lwjgl.glfw.GLFWErrorCallback;
 import org.lwjgl.opengl.AMDDebugOutput;
@@ -39,6 +38,7 @@ import org.lwjgl.opengl.GLCapabilities;
 import org.lwjgl.opengl.GLUtil;
 import org.lwjgl.opengl.KHRDebug;
 import org.lwjgl.system.Callback;
+import org.lwjgl.system.MemoryUtil;
 
 import com.badlogic.gdx.Application;
 import com.badlogic.gdx.ApplicationListener;
@@ -57,6 +57,7 @@ import com.badlogic.gdx.utils.Clipboard;
 import com.badlogic.gdx.utils.GdxRuntimeException;
 import com.badlogic.gdx.utils.ObjectMap;
 import com.badlogic.gdx.utils.SharedLibraryLoader;
+import com.badlogic.gdx.utils.Os;
 
 public class Lwjgl3Application implements Lwjgl3ApplicationBase {
 	private final Lwjgl3ApplicationConfiguration config;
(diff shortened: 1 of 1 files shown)


tommyettinger (changes requested):
This looks good, but I'd rather avoid importing MemoryUtil and instead do what the existing code in this file does. It currently just uses 0, which is all MemoryUtil.NULL is, instead of referencing MemoryUtil.NULL. All that would need here is removing the import for MemoryUtil and changing the single place that uses NULL (I think there's just one) to 0.

Kyrixen:
Thank you! I've addressed the requested changes.

tommyettinger (approved):
Looks good to me, ~~though I have a feeling the automatic formatter might not like the one unused import of Os at the end of the imports~~ Never mind, I didn't see that it was a shrunk-down import of the wildcard import. I don't think there's any real problems here, so I'm approving! If the formatter fails, just run `gradlew spotlessApply` and it should get everything ready to merge.

PR #7827, opened 2026-07-01
Fix ActorGestureListener touchUp NPE

### LINKED ISSUE #7660: NullPointerException in ActorGestureListener.longPress (libGDX 1.13.1)

Kyrixen:
Fixes #7660.

`ActorGestureListener.handle()` invokes `GestureDetector.touchUp()`, which may trigger a reentrant callback. During that callback, the shared temporary coordinates (`tmpCoords`) can be modified before `ActorGestureListener.touchUp()` is invoked.

This change preserves the converted touch coordinates before invoking `GestureDetector.touchUp()`, preventing the reentrant callback from affecting the coordinates used by `ActorGestureListener.touchUp()`.

A regression test (`ActorGestureListenerTouchUpTest`) has been added. The test reproduces the issue before the fix and runs successfully after the fix. The test is based on the reproduction shared by @lucas-viva in the discussion on #7660.

### DIFF
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
@@ -119,6 +119,7 @@ public class ActorGestureListener implements EventListener {
 			return true;
 		case touchUp:
 			boolean touchFocusCancel = event.isTouchFocusCancel();
+			if (event.getPointer() <= 1) activeTouches--;
 			if (touchFocusCancel)
 				detector.reset();
 			else {
@@ -131,7 +132,6 @@ public class ActorGestureListener implements EventListener {
 				final float preservedTouchY = tmpCoords.y;
 
 				detector.touchUp(event.getStageX(), event.getStageY(), event.getPointer(), event.getButton());
-				if (event.getPointer() <= 1) activeTouches--;
 				touchUp(event, preservedTouchX, preservedTouchY, event.getPointer(), event.getButton());
 			}
 			if (activeTouches == 0) {
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
@@ -122,9 +122,14 @@ public class ActorGestureListener implements EventListener {
 			else {
 				this.event = event;
 				actor = event.getListenerActor();
-				detector.touchUp(event.getStageX(), event.getStageY(), event.getPointer(), event.getButton());
+
 				actor.stageToLocalCoordinates(tmpCoords.set(event.getStageX(), event.getStageY()));
-				touchUp(event, tmpCoords.x, tmpCoords.y, event.getPointer(), event.getButton());
+				// Preserve the touch coordinates before a possible reentrant callback modifies tmpCoords
+				final float preservedTouchX = tmpCoords.x;
+				final float preservedTouchY = tmpCoords.y;
+
+				detector.touchUp(event.getStageX(), event.getStageY(), event.getPointer(), event.getButton());
+				touchUp(event, preservedTouchX, preservedTouchY, event.getPointer(), event.getButton());
 			}
 			this.event = null;
 			actor = null;
--- a/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
+++ b/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
@@ -35,7 +35,7 @@ public class ActorGestureListener implements EventListener {
 	private final GestureDetector detector;
 	InputEvent event;
 	Actor actor, touchDownTarget;
-	private int activeTouches = 0;
+	private int activeTouches;
 
 	/** @see GestureDetector#GestureDetector(com.badlogic.gdx.input.GestureDetector.GestureListener) */
 	public ActorGestureListener () {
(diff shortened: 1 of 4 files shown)


tommyettinger (dismissed):
This looks solid from as far as I can see; most of the changes were discussed in the linked issue and fixes seemed pretty close to being entered, so thank you for actually making the changes in this PR! The test is a welcome addition, though I haven't tried to run it yet. I'm tentatively approving because it looks like these were just the (tested?) changes from the linked issue. I'll try to run the test with and without the changes to ActorGestureListener to see if I can reproduce the regression.

Kyrixen:
Thank you very much for the review and the approval! I really appreciate it. I'm glad I could help fix this issue.

tommyettinger:
Yep, I ran the test without the changes, and there was an immediate crash, with the changes, no crash at all. Looks good to me!

lucas-viva:
Thank you folks!

I tested the PR for both my tests:

https://github.com/lucas-viva/libgdx/commit/e29b189b739e762561a0aab71acd8675c44a3443 (test 1)

and

https://github.com/lucas-viva/libgdx/commit/b6904b4c81d8e50b45c6d4b315a21a56e87c008a (test 2)

The PR fixes test 2, that is focused on desktop (uses mouse buttons, not pointers).

But test 1 is still failing, that is focused on mobile (uses pointers, not mouse buttons).

Kyrixen:
Thanks for testing! I'll take a look at the mobile case and see what's going on. I'll update the PR if I find the cause.

Kyrixen:
I was able to reproduce the Android case on an emulator. It turned out to be a separate issue from the original one, so I've pushed a follow-up commit that fixes the multiple-pointer crash on Android. I also added a regression test based on the reproduction shared by @lucas-viva. Thanks for helping track this down!

Kyrixen:
The cause of the crash was that `ActorGestureListener` stores `actor` and `event` as shared state. During the Android multiple-pointer sequence, `longPress()` could still run after another pointer had already cleared `actor`, causing the crash.

The fix adds a counter to keep track of active pointers and only clears state after all active pointers have been released.

tommyettinger (dismissed):
Looks good again! I'm not sure what behavior to expect from ActorGestureListenerPointerTest, but it doesn't crash for me on LWJGL3. Neither test has issues with the current changes, but I pretty much can't reasonably test on Android. I'm approving but waiting for @lucas-viva or anyone who is better with Android than me to see if the change helps with bugs they've encountered.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
```diff
@@ -122,13 +124,21 @@ public boolean handle (Event e) {
...
-				detector.touchUp(event.getStageX(), event.getStageY(), event.getPointer(), event.getButton());
+
 				actor.stageToLocalCoordinates(tmpCoords.set(event.getStageX(), event.getStageY()));
-				touchUp(event, tmpCoords.x, tmpCoords.y, event.getPointer(), event.getButton());
+				// Preserve the touch coordinates before a possible reentrant callback modifies tmpCoords
+				final float preservedTouchX = tmpCoords.x;
+				final float preservedTouchY = tmpCoords.y;
+
+				detector.touchUp(event.getStageX(), event.getStageY(), event.getPointer(), event.getButton());
+				if (event.getPointer() <= 1) activeTouches--;
```
obigu: I think we should move this line outside the conditional so that when it's a touch up generated by a touch cancel (`touchFocusCancel` is `true`) it's also executed.

obigu (commented):
Thanks for the PR, I haven't properly tested but except from the special case mentioned in the comment, it looks solid. Pinging @NathanSweet.

gdx/src/com/badlogic/gdx/scenes/scene2d/utils/ActorGestureListener.java
tommyettinger: Good catch.

Kyrixen: Got it! Good catch. I've updated the PR and verified the regression tests still pass on both Linux and Android.

NathanSweet:
Fix is close but need to decrement `activeTouches` after `detector.touchUp()` finishes, otherwise `panStop()` can cancel touch focus reentrantly, clear actor, then `GestureDetector.touchUp()` continues and calls `fling()` causing the same NPE. Added a commit and modified test, think it's ready now.

Kyrixen:
Thank you!

PR #7828, opened 2026-07-05
Use Arrays.fill for Array and Queue.

NathanSweet:
Use Arrays.fill for Array and Queue.

ArrayMap has a loop that sets 2 values, probably fine like that.


tommyettinger (approved):
Nice, small, simple PR; it does what it should. Approved!

PR #7829, opened 2026-07-11
add() with 5-8 parameters for arrays

NathanSweet:
This adds `add()` methods with 5-8 parameters to all arrays. Occasionally I'm annoyed it doesn't go up to 8. Adds a little clutter but is convenient.

FWIW, MeshBuilder has 1-4,6,8.

### DIFF
--- a/gdx/src/com/badlogic/gdx/utils/Array.java
+++ b/gdx/src/com/badlogic/gdx/utils/Array.java
@@ -139,12 +139,62 @@ public class Array<T> implements Iterable<T> {
 
 	public void add (T value1, T value2, T value3, T value4) {
 		T[] items = this.items;
-		if (size + 3 >= items.length) items = resize(Math.max(8, (int)(size * 1.8f))); // 1.75 isn't enough when size=5.
+		if (size + 3 >= items.length) items = resize(Math.max(9, (int)(size * 1.75f)));
 		items[size] = value1;
 		items[size + 1] = value2;
 		items[size + 2] = value3;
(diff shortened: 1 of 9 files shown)


gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -139,14 +139,64 @@ public void add (T value1, T value2, T value3) {
 
 	public void add (T value1, T value2, T value3, T value4) {
 		T[] items = this.items;
-		if (size + 3 >= items.length) items = resize(Math.max(8, (int)(size * 1.8f))); // 1.75 isn't enough when size=5.
+		if (size + 3 >= items.length) items = resize(Math.max(9, (int)(size * 1.75f)));
```
tommyettinger, line 142: Good call on changing max rather than the increase rate. I was a little worried when I saw the comment go away, but this is a good change.

gdx/src/com/badlogic/gdx/utils/CharArray.java
tommyettinger, line 1: The resizeBuffer() here uses a 50% increase instead of a 75% increase, which is fine and is an internal detail, but I just felt it should be noted for some reason in my review. I think if anyone is going to have a problem with that, they should be reading the sources anyway, so docs aren't needed. It's a difference from the other Array types, I guess.

tommyettinger (approved):
Looks good to me. DoubleArray is still not present in the sources, which is a little bit of a head-scratcher, but the PR is fine here. Approved.

PR #7832, opened 2026-07-16
Allow ShaderProgram to use an explicit Application

damn:
ShaderProgram currently relies on global `Gdx.app` / `Gdx.gl20`. I want to pass Application explicitly so I can manage my own application lifecycle without depending on global state.

  Add optional Application-accepting constructors to ShaderProgram. When an Application is supplied it is used to register the managed shader and to resolve the current GL20 via
  `app.getGraphics().getGL20()` (read on each use, so context recreation / backend GL replacement still works).

  The existing constructors delegate through `Gdx.app`, so current code is unchanged.

### DIFF
--- a/gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
+++ b/gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
@@ -148,18 +148,36 @@ public class ShaderProgram implements Disposable {
 	/** reference count **/
 	private int refCount = 0;
 
+	/** application this shader is managed by and resolves GL from **/
+	private final Application app;
+
+	private GL20 gl20 () {
+		return app.getGraphics().getGL20();
+	}
+
 	/** Constructs a new ShaderProgram and immediately compiles it.
 	 *
 	 * @param vertexShader the vertex shader
 	 * @param fragmentShader the fragment shader */
 
 	public ShaderProgram (String vertexShader, String fragmentShader) {
+		this(Gdx.app, vertexShader, fragmentShader);
+	}
+
+	/** Constructs a new ShaderProgram and immediately compiles it.
+	 *
+	 * @param app the application used to resolve the current GL20 instance and manage this shader
+	 * @param vertexShader the vertex shader
+	 * @param fragmentShader the fragment shader */
+	public ShaderProgram (Application app, String vertexShader, String fragmentShader) {
+		if (app == null) throw new IllegalArgumentException("application must not be null");
 		if (vertexShader == null) throw new IllegalArgumentException("vertex shader must not be null");
 		if (fragmentShader == null) throw new IllegalArgumentException("fragment shader must not be null");
 
 		if (prependVertexCode != null && prependVertexCode.length() > 0) vertexShader = prependVertexCode + vertexShader;
 		if (prependFragmentCode != null && prependFragmentCode.length() > 0) fragmentShader = prependFragmentCode + fragmentShader;
 
+		this.app = app;
 		this.vertexShaderSource = vertexShader;
 		this.fragmentShaderSource = fragmentShader;
 		this.matrix = BufferUtils.newFloatBuffer(16);
@@ -168,12 +186,16 @@ public class ShaderProgram implements Disposable {
 		if (isCompiled()) {
 			fetchAttributes();
 			fetchUniforms();
-			addManagedShader(Gdx.app, this);
+			addManagedShader(app, this);
 		}
 	}
 
 	public ShaderProgram (FileHandle vertexShader, FileHandle fragmentShader) {
-		this(vertexShader.readString(), fragmentShader.readString());
+		this(Gdx.app, vertexShader, fragmentShader);
+	}
+
+	public ShaderProgram (Application app, FileHandle vertexShader, FileHandle fragmentShader) {
+		this(app, vertexShader.readString(), fragmentShader.readString());
 	}
 
 	/** Loads and compiles the shaders, creates a new program and links the shaders.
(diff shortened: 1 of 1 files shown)


gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
```diff
@@ -148,18 +148,36 @@ public class ShaderProgram implements Disposable {
 	/** reference count **/
 	private int refCount = 0;
 
+	/** application this shader is managed by and resolves GL from **/
+	private final Application app;
+
+	private GL20 gl20 () {
```
tommyettinger, line 154: I'm wondering if this accessor should be `protected` rather than `private` if a goal for this is extensibility.

gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
```diff
@@ -852,7 +874,12 @@ public static String getManagedStatus () {
 
 	/** @return the number of managed shader programs currently loaded */
 	public static int getNumManagedShaderPrograms () {
-		return shaders.get(Gdx.app).size;
+		return getNumManagedShaderPrograms(Gdx.app);
+	}
+
+	/** @return the number of managed shader programs currently loaded for the given application */
+	public static int getNumManagedShaderPrograms (Application app) {
```
tommyettinger, line 881: It might make sense to provide a way to access a ShaderProgram's instance Application field, if there's this static method that needs an Application, but it's most likely not needed very often.

gdx/src/com/badlogic/gdx/graphics/glutils/ShaderProgram.java
tommyettinger: ...I'm also wondering if the GL20 should be cached, for that matter, since this is called in quite a few places.

crykn:
I think this may be outside the scope of the project. libGDX doesn't really support taking care of the application lifecycle yourself.

damn:
> I think this may be outside the scope of the project. libGDX doesn't really support taking care of the application lifecycle yourself.

  Stepping back from lifecycle: I think the deeper issue is global state.
  ShaderProgram is already tied to an Application in the managed registry, but
  construction and GL access go through Gdx.* instead of that Application.
  So the class isn’t closed over its real dependency — it reaches ambient globals.
  Optional Application constructors aren’t “bring your own lifecycle”; they’re
  making the dependency explicit: ShaderProgram depends on Application (and
  thus graphics/GL through it), not on Gdx.
  Existing constructors can keep using Gdx.app so normal apps don’t change.
  If libGDX considers Gdx the supported coupling for all of core, I’ll close
  this and keep the change in a fork — I mainly wanted to check whether
  making that dependency explicit is in scope.

obigu:
I see no problem with this, I actually prefer dependency injection where possible and because we have the fallback that uses Gdx globals there's no breaking changes. Code complexity is similar and I don't see any performance issue while it improves extensibility for a real use case. Looks good to me.

damn:
I have made some new discoveries

AndroidGraphics handles all managed graphics assets and they are all keyed by 'app'.

Instead they should probably live all in one 'gl-managed-resources' thingy and be keyed by 'graphics' (can just pass `this` in here: https://github.com/libgdx/libgdx/blob/32eb91a76e7d8bef051cf5a6c88153cc6dd88395/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L325

That would be an extension of this PR but it would be interesting as a discussion point. I have also write in discord yesterday but no replies yet.

That could possibly also invert the whole design:
`graphics.newTexture`, `graphics.newShaderProgram`, etc...

`ShaderProgram` should not be aware of who is managing it but only provide `invalidate` and related methods.

NathanSweet:
> the class isn’t closed over its real dependency — it reaches ambient globals. Optional Application constructors aren’t “bring your own lifecycle”; they’re making

Do not post LLM slop here. We have our own clankers if we want that. We could put up a sign but if people don't have common sense they also won't read a sign.

> so I can manage my own application lifecycle

Why?

> the deeper issue is global state

Why?

I've yet to see this is fixing an actual problem. There is no need to make architectural changes for purity reasons. Changing libgdx to not use globals requires changes across the whole library. OP's clanker has [already started that](https://github.com/libgdx/libgdx/pull/7834), plus #7840, #7839.

This is our first taste of LLM generated PRs. I'm surprised it took this long. If we want to use a clanker to clean up our library, we should do it ourselves rather than spend time figuring out if changes from someone else's clanker make sense. We own the usage and maintenance burdens.

I'm inclined to close all of OP's PRs. If there was anything there we need, we would have done it ourselves by now. That we haven't shows its importance to us. In the future we'll need to close LLM PRs without having to type shit like this for each one.

damn:
First of all thanks for your honest reply. I appreciate it.

The big PR is closed now. It was obviously too big, I just got excited about the possibilities of these changes.

> > so I can manage my own application lifecycle
> 
> Why?

Because I prefer to have explicit dependencies. It makes me understand my applications better.

> I've yet to see this is fixing an actual problem. There is no need to make architectural changes for purity reasons. Changing libgdx to not use globals requires changes across the whole library. 

This PR is not breaking any backwards compatibility, it only offers one explicit param (default is still Gdx).

Wait - one point I didnt see yet. You are right - if this change goes through the possibility for allowing explicit parameters could be given everywhere. It would definitely give extra maintainance overhead (more different parameter versions, documentation).

But what about tiny steps like this? For a one class at a time? it is also not required everywhere, maybe a client (me for example) uses 10 classes only.

NathanSweet:
I agree avoiding global state is better in general. On the other hand this is a 16 year old library that has had globals that whole time. Changes come with risk and chopping things up behind the scenes can have costs, both within our codebase and for backends in external projects. We'd need good reason to change it, beyond it being theoretically better.

Users don't generally manage their own application lifecycle and global state is not usually a user concern. It used to come up occasionally on Android where the same process was reused and old statics were still around (or something like that, only PTSD remains). AFAIK it's been years since that's been an issue. Eg Timer.TimerThread juggle app/files references for that reason.

PR #7833, opened 2026-07-19
Make iOSPreferences flush atomic

obigu:
`IOSPreferences` `flush` implementation, compared to Android's i neither atomic nor asynchronous. This can lead to save file corruption issues as well as performance issues due to writing occurring on main thread. This PR fixes ~~both issues.~~ atomicity.

The fix has the problem that it's still using a deprecated `NSDictionary` `write()` method under the hood. 
Current MobiVM is missing the binding for the non deprecated write method and a fix should be under way.
https://developer.apple.com/documentation/foundation/nsdictionary?language=objc#Storing-Dictionaries. In any case, even if the non deprecated method was available now it's not a simple switch as it doesn't guarantee atomicity. The solution might be to convert the `NSDictionary` to `NSData` and use the atomic write method available for that format but it will require more investigation.

So, recommended course of action would be to merge this soon while we wait for MobiVM fix and think about how to make it both non-deprecated and atomic.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,6 @@
 [1.14.3]
 - API Addition: add() with 5-8 parameters for arrays.
-- iOS: Made iOS Preferences flush() implementation atomic and asynchronous (see #7833)
+- iOS: Made iOS Preferences flush() implementation atomic (see #7833)
 
 [1.14.2]
 - [BREAKING CHANGE] Revert InputMultiplexer and set addAll return types that were breaking changes in 1.14.1. #7805
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 [1.14.3]
 - API Addition: add() with 5-8 parameters for arrays.
+- iOS: Made iOS Preferences flush() implementation atomic and asynchronous (see #7833)
 
 [1.14.2]
 - [BREAKING CHANGE] Revert InputMultiplexer and set addAll return types that were breaking changes in 1.14.1. #7805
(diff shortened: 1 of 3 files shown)


Berstanio:
The `DispatchQueue.getGlobalQueue` seems to return a concurrent queue, which I think is a bit dangerous when writing setting: https://developer.apple.com/documentation/dispatch/dispatch_get_global_queue
Preferably we create one with SERIAL just: https://developer.apple.com/documentation/dispatch/dispatch_queue_create
`DispatchQueue.create("com.xxx", DispatchQueueAttr.Serial());`

Also, I'm not entirely sure about the async? Does apple guarentee us, that if we move the app to the bg, that the queue is still actually executed? I think not, so maybe we need something like: https://developer.apple.com/documentation/uikit/uiapplication/beginbackgroundtask(expirationhandler:)
But I haven't found proper docs on this

obigu:
Good points. Looks like background tasks are better at handling interruptions or lifecycle events but haven't found either it guarantees execution.

I think the proper class for this would be `UserDefaults`. The problem with it is it doesn't follow the contract of persisting the changes only on `flush`, it does it automatically when the OS decides, but it does guarantee execution and it's the class iOS has for this use. I don't know if it's reasonable to assume apps that make changes to `Preferences` always do it with the intention to commit/flush the changes at some point in which case we might move to `UserDefaults`.

Berstanio:
> Good points. Looks like background tasks are better at handling interruptions or lifecycle events but haven't found either it guarantees execution.

Just declaring the background tasks seems to give us usually 5 seconds, which should be plenty of time.

> I don't know if it's reasonable to assume apps that make changes to Preferences always do it with the intention to commit/flush the changes at some point in which case we might move to UserDefaults.

It _should_ be fine usually, but if we can make it work with marking the background task + serial queue, probably no need to switch?

obigu:
> It _should_ be fine usually, but if we can make it work with marking the background task + serial queue, probably no need to switch?

I think it's riskier than migrating. If not implementing `flush()` is an acceptable semantics change, using `NSUserDefaults`, at least with my knowledge and the uncertainty of where in the code are users currently calling `flush()` is the superior solution imo.

What I will do is change the scope of this PR to just make the operation atomic.

I'll create a new PR that may be open for further disucssion with the migration.

Berstanio (approved):

PR #7835, opened 2026-07-21
Improvements to Haptics API

obigu:
Relatively recent Haptics support was designed with full backwards compatibility in mind but it has several problems:
1. It's not possible for apps to retrieve the native Vibrator or HapticsEngine to pass custom more complex patterns or haptic effects
2. It's fragile and not easily extensible due to vibration methods being part of `Input` interface.
3. Lacks a common Haptics API interface

With the new API users will need to replace `Gdx.input.vibrate()` methods by `Gdx.input.getHaptics().xxxx()`.

Methods that accepted `fallbak` as a parameter have been removed in favour of a new configuration `AndroidApplicationConfiguration.hapticsFallback`  and `IOSApplicationConfiguration.hapticsFallback`  which by default is `false`. It's also possible to programmatically change it through `Haptics.setFallbackEnabled()`.

It is now possible to get the native haptic engines on Android `(AndroidHaptics)Gdx.input.getHaptics().getVibrator()` and on iOS `(IOSHaptics)Gdx.input.getHaptics().getHapticsEngine()`.


tests/gdx-tests/src/com/badlogic/gdx/tests/VibratorTest.java
```diff
@@ -50,11 +50,18 @@ public void create () {
 		stage.addActor(table);
 
 		final CheckBox fallbackCheckbox = new CheckBox("Fallback", skin);
+		fallbackCheckbox.setChecked(Gdx.input.getHaptics().isFallbackEnabled());
+		fallbackCheckbox.addListener(new ChangeListener() {
+			@Override
+			public void changed (ChangeEvent event, Actor actor) {
+				Gdx.input.getHaptics().setFallbackEnabled(!Gdx.input.getHaptics().isFallbackEnabled());
```
Frosty-J: This works but smells compared to setting it based on checkbox state.

Frosty-J (commented):
I'm not sure about removing `Gdx.input.vibrate()` without so much as a deprecation period. Even if it sucks due to evolutions in device capabilities, it is a simple way to make the device vibrate and will be present in many projects. That said, where are the people who use such things, so they can review this properly?

Perhaps I'm needing my bed, but what if, using the test as an example, instead of `Gdx.input.getHaptics().vibrate(50)`, it was `Gdx.haptics.vibrate(50)`? I understand there may be hesitancy regarding embiggening the base `Gdx` class, given it's a staple, but it might be time.

1. Vibration/haptics very clearly isn't input.
2. `Haptics#isVibratorSupported()` makes `Gdx.input.isPeripheralAvailable(Peripheral.Vibrator)` superfluous (likewise for haptic counterparts).
3. Going beyond the `Gdx.y.z()` pattern gets a bit long to use directly (as in, not storing the result of `getHaptics()` to reference as variable in game project), for my tastes.

I like the flexibility of getting the native engine. Slap on a CHANGES file entry and there won't be much blocking this. I am not formally requesting changes, just making conversation.

obigu:
Thanks, I've updated the test with some of your feedback and added CHANGES.

> I'm not sure about removing `Gdx.input.vibrate()` without so much as a deprecation period. Even if it sucks due to evolutions in device capabilities, it is a simple way to make the device vibrate and will be present in many projects. That said, where are the people who use such things, so they can review this properly?

But `Gdx.input.vibrate()` was already gone, the methods that have been removed are the ones that accepted `fallback` as an argument.

> 
> Perhaps I'm needing my bed, but what if, using the test as an example, instead of `Gdx.input.getHaptics().vibrate(50)`, it was `Gdx.haptics.vibrate(50)`? I understand there may be hesitancy regarding embiggening the base `Gdx` class, given it's a staple, but it might be time.
> 
> 1. Vibration/haptics very clearly isn't input.
> 2. `Haptics#isVibratorSupported()` makes `Gdx.input.isPeripheralAvailable(Peripheral.Vibrator)` superfluous (likewise for haptic counterparts).
> 3. Going beyond the `Gdx.y.z()` pattern gets a bit long to use directly (as in, not storing the result of `getHaptics()` to reference as variable in game project), for my tastes.
> 

It's a bit more verbose but as you say it's simple enough to hold a reference. I don't think we should move the `Haptics` interface to a "top level" interface because it's mobile platform especific. Even if it's not an "input", it's actually an "output" and imo the `Input` interface is enough of a good fit that can hold both. There's probably some room for improvement on how to fit this for vibrating controllers which is something I haven't really investigated.

> I like the flexibility of getting the native engine. Slap on a CHANGES file entry and there won't be much blocking this. I am not formally requesting changes, just making conversation.

Yeah, I think this is the main benefit of this as the flexibility each platforms provides is hard to abstract in a common interface and users may be interested on playing with advanced effects.

Frosty-J:
> But `Gdx.input.vibrate()` was already gone, the methods that have been removed are the ones that accepted fallback as an argument.

Perhaps I have missed something. This line is in SNAPSHOT but removed in this PR, which led me to believe it was not already gone:

https://github.com/libgdx/libgdx/blob/e1d69884015ca19061647505f9509db34df5cc82/gdx/src/com/badlogic/gdx/Input.java#L858

> There's probably some room for improvement on how to fit this for vibrating controllers which is something I haven't really investigated.

Due to having natives I wouldn't even want them combined anyway, but sometimes I wish gdx-controllers was part of core libGDX and shared some of the same APIs. Even though I've never got around to playing with it.

obigu:
> > But `Gdx.input.vibrate()` was already gone, the methods that have been removed are the ones that accepted fallback as an argument.
> 
> Perhaps I have missed something. This line is in SNAPSHOT but removed in this PR, which led me to believe it was not already gone:
> 
> 

Ok, I think I get it, `Gdx.input.vibrate(int milliseconds)` has become `Gdx.input.getHaptics().vibrate(int milliseconds)` and, if I understand correctly, you suggest not to remove this particular method but deprecate it not to break backwards compatibility?

Frosty-J:
Yes. We generally don't have many changes that require changing project code. But there's also the risk of making a bigger mess, like all those `MeshBuilder` methods that have been deprecated for a decade and are still easily encountered due to examples on the wiki and third-party sites like Stack Overflow (and if you blindly use AI then it winds up pulling from these resources). Do what you think is best.

obigu:
> Yes. We generally don't have many changes that require changing project code. But there's also the risk of making a bigger mess, like all those `MeshBuilder` methods that have been deprecated for a decade and are still easily encountered due to examples on the wiki and third-party sites like Stack Overflow (and if you blindly use AI then it winds up pulling from these resources). Do what you think is best.

Ok, I don't think it's worth deprecating in this case. I don't expect the code to be present in third-party libs that could break, I don't expect it's used by a high percentage of projects and the change is simple and easy to understand and doesn't require a large refactor.

PR #7836, opened 2026-07-22
Made AsynchrounousAndroidAudio the default Android Audio backend

obigu:
Default Android Audio implementation has serious performance issues when multiple sounds are played concurrently due to blocking ui thread. `AsynchrounousAndroidAudio` fixes this problem by running on a separate thread. It's been a while since the implementation was complete and should have been battle tested enough for it to become the new default.

Haven't marked it as breaking change as nothing should break but I'm no 100% sure if it should be.


CHANGES
```diff
@@ -1,6 +1,7 @@
 [1.14.3]
 - API Addition: add() with 5-8 parameters for arrays.
 - iOS: Made iOS Preferences flush() implementation atomic (see #7833)
+- Android: Made AsynchrounousAndroidAudio the default Android Audio backend for better performance.
```
Frosty-J: Typo. Should be `AsynchronousAndroidAudio`, not Asynchrounous.

Frosty-J (commented):
This documentation about sound ID not working is outdated, yeah?

https://github.com/libgdx/libgdx/blob/8dea6d0952753e426cf0a6b062467e0fa7c4713d/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AsynchronousAndroidAudio.java#L13-L14

Sounds good to me if it works (I haven't extensively tested either). The synchronous approach is pretty bad, seems to have become a bigger problem over the years for some reason. It does leave `DefaultAndroidAudio` as a misnomer to no longer be the default... not a great name, in hindsight.

obigu:
> This documentation about sound ID not working is outdated, yeah?
> 

Good catch! It works, docs are outdated.

obigu:
> 
>  It does leave `DefaultAndroidAudio` as a misnomer to no longer be the default... not a great name, in hindsight.

Maybe we should Deprecate `DefaultAndroidAudio`.

Berstanio:
I think if we move the defaults the `Sound` docs need adjustments:
> @ return the id of the sound instance if successful, or -1 on failure. */

As this wont be true by default anymore, I think this is worth an adjustment?
And then it may be worth adding a small `public void hasFailed(long soundId)` to recover the lost functionality?

obigu:
I assume you suggest adding the method to `AsynchronousSound`. We can do that but I highly doubt anybody will use it. If a sound fails to play there's really not much you can do.

Berstanio:
> I assume you suggest adding the method to AsynchronousSound.

No, I meant to `Sounds`, where it would be a `return soundId == -1` on most backends and in `AsynchronousSound` a `return getSoundId(soundId) == -1`.
The reason being, it would be beneficial for e.g. modders of games to give them an indication, if their modded sound failed to play.

obigu:
> > I assume you suggest adding the method to AsynchronousSound.
> 
> No, I meant to `Sounds`, where it would be a `return soundId == -1` on most backends and in `AsynchronousSound` a `return getSoundId(soundId) == -1`. The reason being, it would be beneficial for e.g. modders of games to give them an indication, if their modded sound failed to play.

That's a bigger breaking change for 3rd party sound libs if we change the Sound interface. I know that at least `gdx-miniaudio` implements it. Also, afaik, `gdx-miniaudio` it doesn't handle the `-1` scenario atm (should be possible as MiniAudio returns either success or not).

Berstanio:
> That's a bigger breaking change for 3rd party sound libs if we change the Sound interface.

We have default methods now, so we can make it fully non breaking, no?
With the default implementation being `return soundId == -1`.

obigu:
> > That's a bigger breaking change for 3rd party sound libs if we change the Sound interface.
> 
> We have default methods now, so we can make it fully non breaking, no? With the default implementation being `return soundId == -1`.

do we? :D I'm so outdated

Issue #7837, opened 2026-07-22
DefaultAndroidInput: delta reset loop appears incorrect or redundant


cy-elec:
While investigating an issue in another project, I noticed the following code in DefaultAndroidInput:

https://github.com/libgdx/libgdx/blob/8dea6d0952753e426cf0a6b062467e0fa7c4713d/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/DefaultAndroidInput.java#L489-L494

The loop variable i is never used. This seems to indicate either:

- a typo (deltaX[i] / deltaY[i], or equivalent array-wide reset), or
- an unnecessary loop if only pointer 0 is intended to be reset.

I haven't investigated the surrounding input handling deeply enough to determine the intended behavior, but the current code appears either incorrect or redundant.

While searching the project history, I also came across issue #7187, which seems related to this code path.

Once the intended behavior is clarified, I'd be happy to submit a PR with the appropriate fix.

obigu:
Haven't had a deep look. This is the commit where it was added https://github.com/libgdx/libgdx/commit/504d68c027e371fe01cf897125e2fd8c20339b39

tommyettinger:
Looks like a 15-year-old typo, to me. I'd think we'd typically prefer Arrays.fill() these days anyway to zero out a whole array.

PR #7838, opened 2026-07-23
Fix DefaultAndroidInput delta reset

cy-elec:
The existing implementation iterates over all indices but only resets
`deltaX[0]` and `deltaY[0]`.

Based on the discussion in #7837, this appears to be a long-standing
typo. Replace the loop with `Arrays.fill()` to reset all pointer deltas.

I'm opening this as a draft pending clarification of the intended behavior.


obigu:
Thanks, you can remove the Draft state, the logic you apply is the correct expected one and it can be merged. The reason nobody has noticed is because in practice, probably nobody runs this polling method without checking if the pointer `justTouched` which means even if values may be "dirty" nobody is consuming them. From a correctness point of view they need to be set to 0 once consumed though.

obigu (approved):

PR #7839, opened 2026-07-23
Deduplicate ANGLE EGL window creation fallback in Lwjgl3Application

damn:
Fullscreen and windowed paths shared the same Ubuntu/Nvidia ANGLE retry; extract createGlfwWindowHandle so the workaround lives in one place.


Frosty-J:
I've only looked at the diff but I thought this one is good. I'm in favour of deduplicating code in most circumstances.

PR #7840, opened 2026-07-23
Remove obsolete LwjglApplet support.

damn:
Java applets are no longer viable; drop the LWJGL applet entry point and deprecate ApplicationType.Applet while keeping the enum ordinal stable.


tommyettinger (dismissed):
I agree that this PR is useful; I'll probably never be able to run an applet ever again. There's not any point in supporting them when browsers have blocked applets and removed any ability to use them for many years.

Frosty-J:
Honestly, I'd like to see the whole LWJGL(2) backend phased out. It has a maintenance cost. But I believe there's still reasons not to, namely embedding LWJGL3 into AWT doesn't work very well.

As far as removing applet support goes, I don't see much benefit in doing so, nor much harm. The main reason to use applets is for browser support, with Internet Explorer being the only major browser to support them. But it/Oracle Java freaks out if your applet is unsigned, so still not much use for serving games to Grandma unless you're willing to navigate that field. I don't think anyone has actually tested libGDX applets in a long time.

tommyettinger:
You should probably delete that comment in case Hacker News or Reddit thinks you're serious about Internet Explorer being a "major browser" in 2026 AD. Enough people think Java is a joke for game development already; don't contribute to the stereotype.

> Internet Explorer's market share is below 0.2% globally in 2026. This indicates a significant decline as users have largely transitioned to other browsers.

obigu (approved):
Agree it's time we get rid of applets

PR #7841, opened 2026-07-29
Add overloads for btRigidBody methods

Minecraftian14:
Add additional overloads for usual btRigidBody methods like `getLinearVelocity()`, which instead of creating and returning new instances, take one in as a parameter and mutate it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ allprojects {
 		lineEndings 'UNIX'
 
 		java {
-			target 'src/**/*.java', 'test/**/*.java', 'generator/**/*.java'
+			target 'src/**/*.java', 'test/**/*.java', 'generator/**/*.java', 'jni/swig-src/**/*.java'
 			removeUnusedImports()
 			eclipse().configFile new File(rootProject.projectDir.absolutePath, 'eclipse-formatter.xml')
 		}
--- a/extensions/gdx-bullet/jni/swig-src/dynamics/com/badlogic/gdx/physics/bullet/dynamics/DynamicsJNI.java
+++ b/extensions/gdx-bullet/jni/swig-src/dynamics/com/badlogic/gdx/physics/bullet/dynamics/DynamicsJNI.java
@@ -215,7 +215,7 @@ public class DynamicsJNI {
 
 	public final static native Vector3 btRigidBody_getLinearVelocity (long jarg1, btRigidBody jarg1_);
 
-	public final static native Vector3 btRigidBody_getAngularVelocity (long jarg1, btRigidBody jarg1_);
+	public final static native Vector3 btRigidBody_getAngularVelocity__SWIG_0 (long jarg1, btRigidBody jarg1_);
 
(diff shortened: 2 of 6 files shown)


MonstrousSoftware:
This PR adds non-allocating methods to the `btRigidBody` class as alternative for the existing methods that wrap 
the return value in a java object such as a `Vector3` (an allocation). The original methods remain available for backwards compatibility.

The objective is to reduce the amount of object allocations when running an application using these Bullet methods repeatedly to increase efficiency and reduce garbage collection.

These type of helper methods were suggested in the gdx-bullet comments  : [TODO.txt](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-bullet/jni/swig/TODO.txt), making use of the capabilities of SWIG which is the tool used to generate Java and C++ wrappers for JNI

This same technique was already applied in a number of places, e.g. `btCollisionObject#getWorldTransform(btTransform)`,
but not so far for the `btRigidBody` class. There are other classes which could also benefit from such a change (to name some at random from the VehicleTest example: `btRaycastVehicle#getWheelInfo()` and `btWheelInfo#getWorldTransform()`)

The core of the change are 9 method definitions in `btRigidBody.i` (a SWIG input file), the changes under folder `swig-src` are in the Java and C++ code subsequently generated by SWIG.

From a code review, the pull request looks valid and useful and does not break existing applications. 
However, I have not run SWIG myself nor built the `gdx-bullet` update.
This change modifies the C++ wrappers and requires a rebuild of the native binary `gdx-bullet` per platform.

Minecraftian14:
Thanks for the review! Glad to hear the change looks valid and useful.

I did run the swig gen myself, which updated all the related Java and Cpp files in `swig-src`. 
This process is not streamlined with the build process, and hence has to be done manually, and hence the "generated" sources are also committed.
However I'm unsure about rebuilding the native "gdx-bullet" binaries for the supported platforms, do I have to build and test in each platform? 
I only tried the vehicle test on my system, which is windows 64 bit.

One question regarding testing: what would be the recommended way to test these new overloads? So far I am only displaying the magnitudes of getAngularVelocity' both the overloaded side by side.

MonstrousSoftware:
I don't think the PR author has to build the natives for every platform, I think that happens automatically when the PR is accepted and included in a LibGDX snapshot. With regards to testing, I think there is not much to test as there is no new functionality only an additional calling convention.  Just check it returns the same value as the original method.

Minecraftian14:
That's awesome! Just to clear it up, I'm free to write j unit tests? 

Something like

```
class btRigidBodyTest {
   @Test
  void testSwigExtensions() {
    ... Setup code to get a rigid body instance

    rbd.getAngularVelocity(temp);
    assertEquals(rbd.getAngularVelocity(), temp);
  }
}
```

MonstrousSoftware:
There are very few junit tests integrated in the libgdx repo. I don't know if this is due to some policy (maybe to reduce build times?).  Perhaps just do some tests in your personal  environment.

Minecraftian14:
Test on my personal environment...? 
Does that mean I just ensure that the outputs are same and don't have to commit any "testing" code to the repository?

PR #7842, opened 2026-07-30
Add Json#getUsePrototypes().

TCreutzenberg:
A Json.Serializable#write(Json) implementation is handed the writer but cannot read its usePrototypes setting back, so code that flips the flag for a single field has no way to restore what the caller had. The symmetric getter already exists for ignoreUnknownFields.


tommyettinger (approved):
Very reasonable; there's no reason not to have a getter if we have a setter. There's already a CHANGES entry, too! Approved.

PR #7843, opened 2026-07-31
TexturePacker: reject duplicatePadding with padding < 2.

TCreutzenberg:
`duplicatePadding` exists so that a region's edge texels survive filtering. When `paddingX`/`paddingY` is 1, it silently does nothing.

### The problem

`padding` is the whole gap between two adjacent images, not a per-side margin — `MaxRectsPacker` adds it to a rect once. So each image duplicates `padding / 2` texels outward, and the two rings meet exactly in the middle.

At `padding = 1`, `padding / 2` is 0, so every duplication loop in `writeImages()` has an empty body, and `edgePadding` collapses the same way. The atlas still packs, the regions still have their 1 texel gap — and the gap is empty. Bilinear sampling and any seam-clamping shader then read transparent alpha just outside each region, and every sprite edge picks up a fringe.

Nothing warns about this. The setting is on and inert, the failure is invisible at pack time, and it surfaces only as fringing in the finished game.

### Why not just duplicate more

The halving is correct and this PR does not touch it. `plot()` is an unconditional `setRGB`, so duplicating the full padding instead would have neighbours overwrite each other's ring, and a region would sample its neighbour's colour — worse than the current failure. A 1 texel gap cannot hold one duplicated texel for each of its two neighbours, so `padding = 1` can never work with `duplicatePadding`. The fix is to say so rather than to change the arithmetic.

### The change

The check goes in the `TexturePacker(File, Settings)` constructor, directly below the existing `pot` and `multipleOfFour` blocks, which already reject settings the packer cannot honour:

```java
if (settings.duplicatePadding) {
    if (settings.paddingX < 2)
        throw new RuntimeException("If duplicatePadding is true, paddingX must be >= 2: " + settings.paddingX);
    if (settings.paddingY < 2)
        throw new RuntimeException("If duplicatePadding is true, paddingY must be >= 2: " + settings.paddingY);
}
```

It also adds a javadoc line to the `duplicatePadding` field noting that `padding` is the total gap between two neighbouring images and that only `padding / 2` is duplicated per side — the part that cannot be inferred from the field name.

Nothing in the repository is affected: `tests/gdx-tests-android/assets-raw/skin/pack.json` sets `duplicatePadding: false`, and `TiledMapPacker` uses padding 2 with duplication enabled. `CHANGES` has an entry under 1.14.3.

### One thing worth your call

This turns a silent no-op into a hard failure, so a project currently packing with `padding = 1` and `duplicatePadding` on will now fail instead of quietly producing fringed atlases. That is exactly the population that has the bug and no way to notice it, which is the argument for throwing — and it matches the neighbouring `pot` / `multipleOfFour` checks. If you would rather not break those builds, the same information as a `System.err` warning is still a strict improvement over silence, and I am happy to change it.

(@tommyettinger: maybe also interesting for "com.github.tommyettinger:libgdx-texturepacker")

### DIFF
--- a/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
+++ b/extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
@@ -78,6 +78,13 @@ public class TexturePacker {
 				throw new RuntimeException("If mod4 is true, maxHeight must be evenly divisible by 4: " + settings.maxHeight);
 		}
 
+		if (settings.duplicatePadding) {
+			if (settings.paddingX < 2)
+				throw new RuntimeException("If duplicatePadding is true, paddingX must be >= 2: " + settings.paddingX);
+			if (settings.paddingY < 2)
+				throw new RuntimeException("If duplicatePadding is true, paddingY must be >= 2: " + settings.paddingY);
+		}
+
 		if (settings.grid)
 			packer = new GridPacker(settings);
 		else
@@ -900,6 +907,9 @@ public class TexturePacker {
 		public boolean multipleOfFour;
 		public int paddingX = 2, paddingY = 2;
 		public boolean edgePadding = true;
+		/** If true, the edge pixels of each image are duplicated into the padding, so filtering does not sample transparent pixels.
+		 * {@link #paddingX} and {@link #paddingY} are the whole gap between two adjacent images, so each image duplicates only half
+		 * of it on each side, rounded down. Requires padding of 2 or more. */
 		public boolean duplicatePadding = false;
 		public boolean rotation;
 		public int minWidth = 16, minHeight = 16;
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 - API Addition: Json#getUsePrototypes(), so a Json.Serializable can restore the setting it changed.
 - API Removal: Removed LwjglApplet (Java applets are obsolete). ApplicationType.Applet is deprecated.
 - iOS: Made iOS Preferences flush() implementation atomic (see #7833)
+- API Change: TexturePacker now throws if duplicatePadding is true and paddingX or paddingY is less than 2. Padding is the whole gap between two adjacent images, so half of it is duplicated on each side and a padding of 1 duplicated nothing at all, silently.
 
 [1.14.2]
 - [BREAKING CHANGE] Revert InputMultiplexer and set addAll return types that were breaking changes in 1.14.1. #7805


tommyettinger:
I've been pinged! I'll make the change in my standalone repo to spitting a message to System.err, since that should be noticeable when run as a command-line app. I'd prefer logging a message to a hard failure, since there's a number of possible configurations that result in no change but would be good to inform users about. For example, setting the maximum dimensions to a massive value like 100000x100000 shouldn't ever be filled by an atlas, and if it did, it wouldn't be loadable on any GPU today. Using those settings could print a warning describing why it's a bad idea, but it isn't strictly needed; users would quickly notice the atlas page wasn't loadable because of its size if it did get that big.

In this case, duplicatePadding with padding set to 1 isn't easily noticed that it does nothing good, so some kind of message might be good, but I don't like a hard failure as the method of notifying users. I'm not a fan of fringing either, but I agree that existing packing shouldn't change when users depend on it doing about the same thing.

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
```diff
@@ -900,6 +907,9 @@ static public class Settings {
 		public boolean multipleOfFour;
 		public int paddingX = 2, paddingY = 2;
 		public boolean edgePadding = true;
+		/** If true, the edge pixels of each image are duplicated into the padding, so filtering does not sample transparent pixels.
+		 * {@link #paddingX} and {@link #paddingY} are the whole gap between two adjacent images, so each image duplicates only half
+		 * of it on each side, rounded down. Requires padding of 2 or more. */
```
NathanSweet: At least write the comments yourself.

TCreutzenberg: @NathanSweet: I totally understand your point here. But, to be honest, AI wrote this comment in better English and more precise than I could for free, so what is the point in rewriting it worse in my own words?
The comment itself only explains what is already done in line 81-87 so it could be removed. Is that what you are hinting at?
What is your take on log error vs. throw exception?

NathanSweet: Back before the damned computers could think, we only wrote necessary comments. Often none when it just wasn't important enough. Now we can generate paragraphs for every class and member. The information is not usually wrong (though sometimes the LLM will blabber inappropriately about what the old code did, what bug the new code fixes, what we aren't doing, etc) but it adds noise (and I would say maintenance, but clankers again).

When I see generated comments, it's not usually because a human thought that stuff was important. Arguably now that the computers can think, comments about what is relatively obvious in the code are less important than ever.

At any rate, the comment is... fine, I guess, though it doesn't really match the rest of lib. Imagine if we had extensive generated comments for _everything_, would it be better? 🤔

Log vs throw vs do nothing, I'm fine with any of them but:

- Personally I would choose do nothing. The input is valid, the behavior is valid if surprising, the user probably notices when they don't get the result they want. If the problem is an API gotcha, it could warrant a comment.
- I don't generally like logging errors, since I'd rather app code decide what is important enough to log.
- Throwing is aggressive and adds a surprise for existing code that used to work.

tommyettinger: Imagine if we had *any* comments for what expand and fill actually mean in scene2d.ui, so people could actually make sense of Table and Cell... 🤔 I know they don't make sense to me.

Frosty-J: It can remain undocumented in code or copy the description from https://libgdx.com/wiki/tools/texture-packer#settings if it's an issue.

TCreutzenberg:
I will remove the comment as I think it is not necessary here.

Exception vs. logging, my reasoning:

**Logging**
`TexturePacker ` might be used without the Gdx app, so which logger should I use then? I anybody can give me a good proposal, I'd be more than happy.

**Exceptions**
`TexturePacker `already throws exceptions for wrong settings (block starting line 67 and 74). and I'd argue the same here, the setting is wrong/very misleading/silently using 0 instead of 1.

Setting `paddingX/Y` to 0: Expected result: No padding: Fine
Setting `paddingX/Y` to 1: Expected result: Padding: but it will be reduced to 0 silently: No padding. and this is actually really bad because the created artifacts will be more or less subtle. They might even not show up under certain conditions.

I am very confident that a lot of people will either
- not notice there are artifacts but just subconciously notice there is something wrong/off
- or they will not know where these artifacts originate from (-> `paddingX` or `paddingY` is 1 and thus being "calculated to 0)

So why not help the user by throwing an exception, this is obviously wrong input?
If the user really wants those artifacts, he could just set `paddingX/Y` to 0 for exactly the same result (should this be documented?).

Those are my 2 cents.

extensions/gdx-tools/src/com/badlogic/gdx/tools/texturepacker/TexturePacker.java
NathanSweet: Examples of where minimal comments are missing is missing the point. Over commenting is worse than under commenting, especially AI generated over commenting.

PR #7844, opened 2026-08-04
Draw a GlyphLayout with a tint in a single pass

TCreutzenberg:
Tinting cached text currently requires two full passes over the glyph vertices: BitmapFontCache#addText builds them, then #tint walks every layout, run and glyph again to rewrite the color of all four vertices of each glyph. Text that moves and fades every frame, which cannot reuse the cache, pays both passes on every frame. The color of the Batch is no help here, since Batch#draw(Texture, float[], int, int) copies the vertices verbatim.

This adds the tint to the pass that is already there:

  BitmapFont#draw(Batch, GlyphLayout, float, float, Color)
  BitmapFontCache#addText(GlyphLayout, float, float, Color)
  BitmapFontCache#setText(GlyphLayout, float, float, Color)

The tint is multiplied into the color once per color run rather than once per glyph, markup colors keep working, and a null tint leaves the existing behavior untouched.

Since a tint applied while adding is baked into the vertices, the tint of the cached text is no longer known afterwards, so #tint is never skipped for that text.

Adds headless tests comparing the vertices of both paths float for float, for plain and markup text, with integer positions on and off, and for a multi page font.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
 [1.14.3]
+- API Addition: BitmapFont#draw and BitmapFontCache#addText/setText overloads that take a tint for a GlyphLayout. The tint is applied while the glyph vertices are built, avoiding the second pass over the vertices done by BitmapFontCache#tint.
 - API Addition: add() with 5-8 parameters for arrays.
 - API Addition: Json#getUsePrototypes(), so a Json.Serializable can restore the setting it changed.
 - API Removal: Removed LwjglApplet (Java applets are obsolete). ApplicationType.Applet is deprecated.
--- a/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
+++ b/gdx/src/com/badlogic/gdx/graphics/g2d/BitmapFont.java
@@ -39,6 +39,7 @@ import com.badlogic.gdx.utils.Array;
 import com.badlogic.gdx.utils.Disposable;
 import com.badlogic.gdx.utils.FloatArray;
 import com.badlogic.gdx.utils.GdxRuntimeException;
+import com.badlogic.gdx.utils.Null;
 import com.badlogic.gdx.utils.StreamUtils;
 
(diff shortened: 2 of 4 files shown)


tommyettinger (approved):
This looks good! I always like to see headless tests that can verify behavior we couldn't verify before. I don't have any direct evidence that BitmapFont drawing was noticeably slow before in normal usage, but I did have some extreme-case tests that may have backed up some benefit for this in very large GlyphLayouts (like license text). Those don't usually have more than one color used, but the full text could still be tinted, or need to scroll smoothly. I'm glad the markup is being tested in the tests, though I might need to run some tests myself for unusual cases. I'm approving because I have no evidence this does anything wrong, and looks like it does everything right!

NathanSweet:
Colors are already baked into the GlyphLayout, this is tinting them when adding to BitmapFontCache rather than adding them, then tinting in a second loop. Is the difference measurable? If performance at this level mattered (doubt), another option could be to mutate GlyphLayout `colors`.

Comments could be pithier, otherwise it looks OK. It is slightly more efficient in a specific scenario, but is it solving a real problem? Why are we making these changes? If we can't measure it, do we care?

There are many things we _could_ change, but I would not like to point clankers at our codebase and make lots of changes solely because they are theoretically better. I am also not a fan of over commenting, especially in AI style. Over time I expect the number of PRs to grow and many will be of that type. If I spend some of my finite time in this life reviewing PRs, it should be for good reason. There is also some small risk of further work needed for any change. Multiplied by many changes it can grow tiresome. There is no problem here yet, but it will likely increase slowly and we may want to consider being stricter than we were pre-AI.

TCreutzenberg:
@NathanSweet I am totally fine to drop this PR. You are right, it is more to please my inner Monk (anybody knowing the series will understand this) more than it is genuinely needed. Which is not what a framework might really need.

PR #7847, opened 2026-08-10
Make Array cloneable

imthosea:
Makes `Array` implement `Cloneable` and adds `Array.clone` for it.


imthosea:
Nevermind, apparently Cloneable doesn't work on GWT.

gdx/src/com/badlogic/gdx/utils/Array.java
```diff
@@ -652,6 +652,19 @@ public <V> V[] toArray (Class<V> type) {
 		return result;
 	}
 
+	@Override
+	public Array<T> clone() {
+		try {
+			Array<T> array = (Array<T>)super.clone();
```
tommyettinger, line 658: Pretty sure this will fail on GWT due to Object not implementing clone(). See https://github.com/gwtproject/gwt-site/blob/becd13c39f53d45e55233ac29c8dfa7ebcf92dca/src/main/markdown/doc/latest/RefJreEmulation.md , Object doesn't have clone(), and Cloneable is empty.

tommyettinger (commented):
I'm not 100% sure, but I think this will fail tests on GWT. It looks like you already closed it though, so thanks!

PR #7848, opened 2026-08-10
Make Lwjgl3Application not aquire a lock if there's no runnables

imthosea:
Adds a check if `runnables.size > 0` before entering the synchronized block


Tom-Ski:
This breaks concurrency, and can cause runnables to never be executed, It also introduces reference leaks that can retain for lifetime of app potentially.
On desktop, this lock is going to be nanoseconds, if that. Is there a specific problem you are trying to solve?

imthosea:
I thought the overhead of a synchronized block was much higher (I was also going to make the `lifecycleListeners` a `CopyOnWriteArrayList` to avoid the other synchronized in the for loop) though if it's not then this isn't really needed, sorry.

PR #7849, opened 2026-08-13
OpenGL ES 3 support with Google Angle

### LINKED ISSUE #7845: Add OpenGL ES 3 support to LWJGL3 ANGLE backend
### Description

The `gdx-lwjgl3-angle` backend currently only exposes `ANGLE_GLES20`.

When ANGLE support was added in libGDX 1.11.0, GLES 3 was not supported because ANGLE's Metal backend only supported GLES 2 at the time.

ANGLE now lists OpenGL ES 3.0 support over Metal as complete:
https://github.com/google/angle#level-of-opengl-es-support-via-backing-renderers

Would it be possible to add GLES 3 support to the libGDX ANGLE backend, e.g. via an `ANGLE_GLES30` mode exposing `Gdx.gl30`?

This would make ANGLE/Metal a viable alternative on macOS for applications that require GLES 3 features.

Dgzt:
The Google Angle build [was updated](https://github.com/libgdx/gdx-angle-natives/pull/7) and with this update the Google Angle supports GL ES 3+, but libGDX doesn't. 

- **Windows**: The default angle backend is Direct3D 11 what supports ES 3.0, but Vulkan backend support ES 3.2.
- **Linux**: The default angle backend is OpenGL, what supports ES 3.1 and partially ES 3.2. I don't see any reason to use this angle backend, but there is Vulkan on Linux also what supports ES 3.2.
- **MacOS**: The default (and only) backend is Metal what supports ES 3.0.

 [Here](https://github.com/google/angle#level-of-opengl-es-support-via-backing-renderers) you find the compatibility table.

Currently libGDX doesn't support backend changing, only uses default one, I have a plan to implement this change functionality in an another PR, but until that GL ES 3 is enough because the default backends support only GL ES 3.

You can test this changes: `LWJGL3TestStarter --gl30 --angle`


Dgzt:
PR for this issue: https://github.com/libgdx/libgdx/issues/7845

PR #7852, opened 2026-08-22
Preferences: report whether a flush actually persisted

MRZ07:
Adds an additive callback overload to `Preferences.flush(PreferencesSaveCallback)` for asynchronous persistence with a reported result. Existing `flush()` behavior remains unchanged.

## API

```java
preferences.flush(new PreferencesSaveCallback() {
    public void onSuccess () { ... }
    public void onFailure (PreferencesSaveResult result, Throwable cause) { ... }
});
```

The callback is invoked once after persistence completes. Existing custom `Preferences` implementations remain source-compatible through the default interface implementation; platform backends override it to perform asynchronous work.

## Backend behavior

- **Android**: runs `SharedPreferences.Editor.commit()` on a daemon worker thread, preserving the synchronous result without blocking the application thread.
- **LWJGL3 / LWJGL / Headless**: runs the shared NIO write path on a daemon worker thread. `AccessDeniedException` is classified as `ACCESS_DENIED`; other write failures are `IO_ERROR`.
- **iOS (RoboVM and MetalANGLE)**: runs the existing dictionary write on a GCD background queue. RoboVM exposes only a boolean result, so failures are `IO_ERROR`.
- **GWT**: defers the write to the next event-loop turn because JavaScript is single-threaded. Local Storage quota failures are classified as `DISK_FULL` using the browser error identifier `QuotaExceededError`.

`PreferencesSaveResult.from(Throwable)` is deliberately conservative and GWT-safe: `SecurityException` maps to `ACCESS_DENIED`, `IOException` maps to `IO_ERROR`, and other failures map to `UNKNOWN`. It does not parse localized OS error messages. Desktop backends do not claim `DISK_FULL` because Java NIO provides no dedicated ENOSPC exception type.

## Verification

- Full builds passed for `gdx`, Android, GWT, Headless, LWJGL, LWJGL3, RoboVM, and RoboVM MetalANGLE.
- Full `:gdx:test` suite passed.
- `PreferencesSaveResultTest` covers conservative type classification and cause-chain handling.
- MetalANGLE generated sources were regenerated and validated.
- Android, iOS, and browser runtime behavior still need device/browser validation.


obigu:
Even if I agree we lack consistency across backends and potential improvements (on iOS for example), I don't think going synchronous on disk write locking the main thread is the way to go. On mobile Android with `SharedPreferences` and iOS with `UserDefaults` recommend (or even force) async for performance reasons and I'd gess this applies to the rest of platforms.

Ideally persisting should occur in a background thread and we would have a callback that allows users to "block" or display a "Saving...." progress element on their end when ultra critical/rollback necessary write operations are required. For the rest of standard preferences persisting, performance is much more important.

MRZ07:
A verified write must not block the game thread, so I’m keeping the existing no-argument `flush()` and adding an asynchronous `flush(PreferencesSaveCallback)` overload. Android and the JVM backends use a worker thread, RoboVM uses GCD, and GWT defers the write to the next event-loop turn.

Frosty-J:
For async with callback I think a `flush()` overload would be clearer? To use the same terminology for async writes with or without callback.

MRZ07:
Implemented in `4599801`. The callback now lives on `Preferences.flush(...)`, keeping the existing terminology while leaving the no-argument `flush()` unchanged; backend writes run off the game thread where the platform allows it.

MRZ07:
Reworked this in `4599801` based on the feedback here. `flush(PreferencesSaveCallback)` now handles the asynchronous verified write, JVM backends use typed NIO errors instead of localized messages, and GWT handles `QuotaExceededError` separately. The affected builds and tests pass locally.

PR #7854, opened 2026-08-23
Box2D: Fix welding threshold on GWT

Frosty-J:
What is a welding threshold? Dunno, but in C++ it gets squared and in GWT it doesn't. This discrepancy was caused by https://github.com/libgdx/libgdx/commit/06562a4c3646cbc5ef6d1263f6182f017712b546 forgetting to update GWT.

https://github.com/libgdx/libgdx/blob/97f40861878058087be0685ca167ddfde132134b/extensions/gdx-box2d/gdx-box2d/jni/Box2D/Collision/Shapes/b2PolygonShape.cpp#L141

https://github.com/libgdx/libgdx/blob/97f40861878058087be0685ca167ddfde132134b/extensions/gdx-box2d/gdx-box2d-gwt/src/com/badlogic/gdx/physics/box2d/gwt/emu/org/jbox2d/collision/shapes/PolygonShape.java#L120

So this pull request changes it to match that. This caused a very annoying behaviour for me where a 2x2 box appeared in my body on GWT only. I still think the fact this code can lead to `setAsBox(1.0f, 1.0f)` is nasty, but at least it should now behave equally between platforms.

<figure>
Screenshot comparison of a snake body. GWT above, Java below. On GWT, the snake has a big box in its body.
<figcaption><i>The picture I sent at <a href="https://discord.com/channels/348229412858101762/953202082998018068/1449066015701864519">https://discord.com/channels/348229412858101762/953202082998018068/1449066015701864519</a> back when I was being annoyed at this.</i></figcaption>
</figure>


tommyettinger (dismissed):
Looks like a clear bug-fix of an ancient bug to me. Approved.

tommyettinger (approved):
Re-approved.

PR #7857, opened 2026-08-24
Update LWJGL3 to 3.4.3

Dgzt:
LWJGL 3.4.3 was released [yesterday](https://github.com/LWJGL/lwjgl3/releases/tag/3.4.3) and libGDX still uses 3.3.3, so I updated to 3.4.3.

The virus false-positive related problem has been fixed in [3.3.6](https://github.com/LWJGL/lwjgl3/issues/1005) and OpenAL has been updated in [3.4.0](https://github.com/LWJGL/lwjgl3/releases#release-3.4.0) so hopefully it fixes [this](https://github.com/libgdx/libgdx/pull/7583#issuecomment-2628912416) issue.


tommyettinger:
I'm using a constraint to update Liftoff projects to 3.4.3 now. It seems to work fine. That includes with TuningFork, which had faced some issues updating to 3.4.1, but now works with 3.4.3. As before, there's a single point of configuration to drop LWJGL3 to an earlier version if needed. I haven't noticed any issues yet, but I also haven't tested every libGDX 1.14.2 project, of course.

tommyettinger (approved):
Looks good to me, but I'm hoping to hear @Berstanio 's opinion on this LWJGL3 release, since he knows a lot more about LWJGL3 than I do. This is supposed to have some fixes for the Java 25/26 issues in 3.4.2, but I don't know if that was on the JVM side or LWJGL3.

Berstanio (approved):
Seems good to me.

Issue #6844, opened 2022-04-04
[Feature Request] Native QOI Image Format Support


Fios-Dev:
Please ensure you have given all the following requested information in your report.

#### Issue details
QOI is a new lossless image format with comparable compression ratios to PNG but with significantly faster processing speed due to the simplicity of its compression and decompression algorithms. The actual algorithm is fairly simple to implement as well. I'd offer to do it myself, but I wouldn't know where to start in the codebase.

https://qoiformat.org/

#### Reproduction steps/code
N/A

#### Version of libGDX and/or relevant dependencies
All

#### Stacktrace
N/A

#### Please select the affected platforms
- [ x] Android
- [ x] iOS
- [ x] HTML/GWT
- [ x] Windows
- [ x] Linux
- [ x] macOS

Frosty-J:
I'm not particularly well-versed in this, but https://github.com/libgdx/libgdx/blob/master/gdx/jni/gdx2d/stb_image.h seems to be where image decoding takes place.

Personally, I'd rather see support for more mainstream formats, such as WebP and AVIF, which both achieve better compression ratios than PNG and debatably JPEG (I find WebP smoothens photos too much and doesn't really win over MozJPEG, but AVIF is probably objectively better). But they are complex formats that can't make the same 300-line claim QOI does!

I took a look at the sample images, and PNG usually wins in terms of file size. So this mostly boils down to whether anyone's game is spending a long time decoding images. I haven't had any trouble with that, even without exploring multithreading, but my games are small in scope, so we'd really have to hear from someone else.

I assume this would be slower than ZKTX and whatever other texture formats we have, but also smaller in size to get a nice middle ground between that and PNG?  
https://libgdx.com/wiki/graphics/2d/texture-compression

tommyettinger:
[...]

Issue #4858, opened 2017-08-15
Texture packer creates brighter files


Abwuds:
#### Issue details
Textures are brighter when packed with the texture packer.

#### Reproduction steps/code
- Run the MyPacker [example](https://github.com/libgdx/libgdx/wiki/Texture-packer#running-texturepacker) on a png file. With the default configuration of the packer.
- The file created contains a brighter version of the original one.

#### Version of LibGDX and/or relevant dependencies
Gdx 1.9.3

#### The affected platforms
- On MacOS

Scellow:
Have you tested with latest version?

https://libgdx.badlogicgames.com/versions.html

Abwuds:
Yes, I just tried with the 1.9.6 and the result is the error shows up again

piotr-j:
Ive made a simple test for this https://github.com/piotr-j/libgdxplayground/blob/master/core/src/io/piotrjastrzebski/playground/packs/PackColorTest.java
With this image https://github.com/piotr-j/libgdxplayground/blob/master/android/assets/pack_in/colors.png

Packed and source images are identical on windows and mac as far as i can tell.

Perhaps you can make a sample that shows the issue clearly.

hiddicop:
Just had the same issue myself and discovered a solution. My image was encoded as a grayscale mode. Changing it to RGB-mode solved the problem.

kolen:
Seems that it's problem with `AWT`'s `getScaledImage`

```java
package com.test.scalingtest;

import javax.imageio.ImageIO;
import java.awt.*;
import java.awt.image.BufferedImage;
import java.io.File;
import java.io.IOException;

public class ScalingTest {
    public static void main(String[] args) {
        try {
            BufferedImage img = ImageIO.read(new File("grayscale.png"));
            Image image = img.getScaledInstance(100, 100, Image.SCALE_SMOOTH);
            BufferedImage buffered = new BufferedImage(100, 100, BufferedImage.TYPE_BYTE_GRAY);
            buffered.getGraphics().drawImage(image, 0, 0, null);
            ImageIO.write(buffered, "png", new File("grayscale-out.png"));
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

![image](https://user-images.githubusercontent.com/24191/29931112-f5056eec-8e78-11e7-87c1-d1effc0b2308.png)

[Discussion on Stackoverflow](https://stackoverflow.com/a/20084061/123642)

[...]

Issue #4814, opened 2017-07-07
TexturePacker does not preserve transparent background of some PNGs


ttencate:
#### Issue details
I have a set of PNG files which only contain white and transparent pixels. My software seems to have decided to store them as 1-bit grayscale with 1-bit alpha. The problem is that the transparency is not preserved by TexturePacker.

Full disclosure about an affected file:

```
$ identify -verbose repro.png 
Image: repro.png
  Format: PNG (Portable Network Graphics)
  Mime type: image/png
  Class: DirectClass
  Geometry: 9x16+0+0
  Units: Undefined
  Type: Bilevel
  Base type: Grayscale
  Endianess: Undefined
  Colorspace: Gray
  Depth: 8/1-bit
  Channel depth:
    gray: 1-bit
    alpha: 1-bit
  Channel statistics:
    Pixels: 144
    Gray:
      min: 0 (0)
      max: 255 (1)
      mean: 69.0625 (0.270833)
      standard deviation: 113.715 (0.445941)
      kurtosis: -0.964827
      skewness: 1.02065
      entropy: 0.842658
    Alpha:
      min: 0 (0)
      max: 255 (1)
      mean: 69.0625 (0.270833)
      standard deviation: 113.715 (0.445941)
      kurtosis: -0.964827
      skewness: -1.02065
      entropy: 0.842658
  Alpha: graya(0,0)   #00000000
  Colors: 2
  Histogram:
       105: (  0,  0,  0,  0) #00000000 graya(0,0)
        39: (255,255,255,255) #FFFFFFFF graya(255,1)
  Rendering intent: Undefined
  Gamma: 0.45455
  Chromaticity:
    red primary: (0.64,0.33)
    green primary: (0.3,0.6)
    blue primary: (0.15,0.06)
    white point: (0.3127,0.329)
  Background color: graya(255,1)
  Border color: graya(223,1)
  Matte color: graya(189,1)
  Transparent color: graya(0,0)
  Interlace: None
  Intensity: Undefined
  Compose: Over
  Page geometry: 9x16+0+0
  Dispose: Undefined
  Iterations: 0
  Compression: Zip
  Orientation: Undefined
  Properties:
    date:create: 2017-07-07T17:27:58+02:00
    date:modify: 2017-07-07T17:27:14+02:00
    png:bKGD: chunk was found (see Background color, above)
    png:cHRM: chunk was found (see Chromaticity, above)
    png:gAMA: gamma=0.45455 (See Gamma, above)
[...]

PR #7860, opened 2026-08-31
GWT: Add support for KTX, WebP, AVIF, and so on

Frosty-J:
Something that befuddled me for a long time is that WebP wasn't working in GWT. It was not obvious that `DefaultAssetFilter#getType` has to be overriden to enable support for it. Oddly, doing that isn't required to play back other *audio* formats such as FLAC. I use WebP a lot now in web builds - it's supported by all web browsers I care about (including back to iOS 14) and even in lossless mode can be less than a quarter the size of PNG ([this file](https://github.com/user-attachments/assets/023c2733-09a6-432f-9a76-9a1954d55537) is the one that had the biggest saving in my bridge game).

## Assets filter

This started out by expanding the default file extensions. That includes all the unique formats a browser might reasonably support, as well as alternative file extensions. Believe it or not, I did limit myself here - [JPEG's Wikipedia page](https://en.wikipedia.org/wiki/JPEG), and [an Adobe article it cites](https://www.adobe.com/creativecloud/file-types/image/raster/jpeg-file.html#what-is-a-jpeg-file), has `.jif` and `.jfi` in the list of extensions, but nothing on my Windows computer has registered those file extensions so they don't seem to be in use.

I don't think there's inherently a problem with having too many extensions listed here, other than if one happens to collide with something else. I'd rather have redundancy in it than leaving someone with the question of "why doesn't this work?" as I was.

I cannot test HEIF and friends as I don't have an Apple device new enough for that. I'm not sure why you would use them anyway - any HDR information and such will be lost when RGB888-ified. Maybe for some weird personal project where you want to leave your photos in their original format.

<details>
<summary>Safari notes</summary>
<ul>
<li>A minority of users may have Lockdown Mode enabled. This disables TIFF, BMP, JPEG 2000, and PDF image support (<a href="https://www.privsec.nz/releases/apple-lockdown-mode-deep-dive">https://www.privsec.nz/releases/apple-lockdown-mode-deep-dive</a>). As it also disables WebGL (and WebAudio and some other stuff), this is of moot concern to us.</li>
<li>I didn't add JPEG 2000 to the test because it is no longer supported by any web browser (and it has too many file extensions to think about for the asset filter). Safari dropped support with version 18.</li>
<li>The PDF I've used is saved at 64x64pt. Apple uses 72dpi, so it comes out to 64x64px. Elsewhere, 96dpi is the standard, but we don't have to worry about the texture becoming 85x85 because, as far as I can tell, PDF images are just an Apple thing.</li>
<li>Regarding PDF image and TIFF support, there's some interesting history here! PDF seems to have at one point been equivalent to Android's XML, for vector images. <a href="https://en.wikipedia.org/wiki/Comparison_of_web_browsers#cite_note-TIFF-257">Wikipedia says</a> TIFF "was one of the first suggested WWW image formats" but I really have no idea what that's referring to and it could be bogus. While there's no evidence that they will, I wouldn't count on Apple to not drop support for these (not PDFs as a whole, I mean loading them as images instead of documents) like they did with JPEG 2000.</li>
</ul>
</details>

I'm not sure why we determine what's an image based on its file extension anyway. This is not the most reliable way of doing things. Note that only things that can be cast to an `ImageElement` should go in the list - things that are handled differently (such as KTX) than just letting the browser render it as an image must not be in the image extensions.

## libGDX specials

CIM... okay, does anyone actually know what CIM is? It's libGDX's own compressed format for pixmaps, seemingly intended to be faster than compressed (i.e. not compression level 0) PNG for writing images at runtime (I wonder if it could also help with reading? See #6844). It is unsupported in GWT because of no `InflaterInputStream` support. I did take a look at the Browser Compression Streams API, and yeah, async JS with synchronous Java doesn't work, and adding JZlib or something for this wouldn't be worth it.

Likewise, the `.etc1` files produced by `ETC1.encodeImagePKM()` and ZKTX produced by `KTXProcessor` depend on `GZIPInputStream` to be read. So that won't work, but we can still have ETC1-encoded KTX files and gzip them (subject to your web hosting; for itch.io, see https://itch.io/docs/creators/html5#compression - no, I don't know how you'd make this work with libGDX's preloader).

## KTX

KTX1 files are supported in this pull request, but I have not tried anything remotely advanced with them like cube mapping. Something like https://github.com/crashinvaders/gdx-basis-universal looks a lot more appealing, but anyway, it's better than nothing. See in `ImageFormatTest`, how I'm loading ETC1 for mobile users and S3TC for desktop. As we don't have the software decoder fallback on GWT, relying solely on ETC1 isn't possible.

I'm sceptical this will be a problem, but if you ever find an absence of `WEBGL_compressed_texture_etc1` on mobile devices, try `WEBGL_compressed_texture_etc`. That is for ETC2, which is backwards-compatible with ETC1.

## Making the images

The test images are made in Inkscape, then run through GIMP to remove alpha channel and export to more formats than Inkscape supports. The font is 25px black Impact centred on a 64x64 canvas with a white background. Graphic design is my passion!

To make vector images, be sure to convert text to path first, and have a shape in the background to ensure no transparency. You may wish to download [my template](https://github.com/user-attachments/assets/a2da4480-b0e3-4d9c-98b9-8850dfcf879e) for this, or make your own version. In document properties (Ctrl+Shift+D), set units to px and scale to 1.000000 to avoid rounding errors.

To make KTX files, I am using [PVRTexTool](https://developer.imaginationtech.com/downloads/). You can load in a PNG, or any of the same stb_image formats as supported in libGDX for that matter, then Edit -> Encode (Ctrl+E) to change the format. S3TC is labelled as BC1 in the DirectX group. Once you are happy, File -> Save As (Ctrl+Alt+S) `.ktx`. Other tools probably work as well, but beware [KTX Software](https://github.com/KhronosGroup/KTX-Software) seems only to do KTX2 these days.

I included a couple greyscale-encoded images on purpose, because those behaving differently on desktop to web often catches me out (the advice is to avoid them: https://github.com/libgdx/libgdx/issues/2804#issuecomment-89854480, https://github.com/libgdx/libgdx/issues/4814#issuecomment-314303043, https://github.com/libgdx/libgdx/issues/4858#issuecomment-325456017). And one not on purpose: GIMP doesn't support PIC export and I couldn't figure out how to stop ImageMagick from converting it to greyscale (other than by introducing colour into the source image) so I just left it as is.

## The useful and not-so-useful formats

As mentioned, WebP is pretty sweet. I really don't think there's reason not to use it for web games, these days. You can mess around with Gradle to pull in PNG versions of your assets when testing on desktop. I haven't tested, but why not on Android as well like https://github.com/libgdx/libgdx/pull/3980#issuecomment-1498947004? Obviously I'm not going to rope Android in this pull request, but it's something to consider.

AVIF still seems quite slow to encode and isn't supported on iOS versions prior to 16. While it's currently locked behind browser flags for Chrome and Firefox, JPEG XL might prove useful in the future as a middle ground between WebP and AVIF encode/decode time and compression ratio.

For games running locally, e.g. with Electron, the more primitive filetypes may be worth playing with to see if they load any faster. But I don't think there's any real reason to use for example an uncompressed BMP over uncompressed KTX.

Don't get excited about SVG. As with any browser-supported image format, libGDX can load them, but the texture resolution is locked to whatever size the SVG defines itself as. They might save a bit of network traffic (if compressed beforehand like [conform-viewers-01-t.svgz](https://dev.w3.org/SVG/profiles/1.1F2/test/svg/conform-viewers-01-t.svgz) or on-the-fly by web server) compared to rasters, but otherwise I don't think they serve much practical purpose in this state. That would be pretty interesting if TexturePacker supported SVG input and output, just using `transform` to adjust the input SVG data.

## Screenshots

Yawn... that was boring. Open these `<details>` things if you want to see that graphic design I was talking about. Chrome and Firefox yield the same results as each other, at this time. Desktop (and I expect Android and iOS) is still winning in terms of count, but I find (some of) the web's formats comparatively new and exciting!

<details>
<summary>Desktop Java screenshot</summary>
gdx-java
</details>

<details>
<summary>Desktop browser screenshot</summary>
gdx-web
</details>

<details>
<summary>Desktop browser from the future</summary>
gdx-web2
</details>

<details>
<summary>Android browser screenshot</summary>
gdx-android
</details>

<details>
<summary>iPod browser screenshot</summary>
gdx-ipod
</details>


backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/DefaultAssetFilter.java
```diff
@@ -47,19 +49,29 @@ public AssetType getType (String file) {
...
 	}
 
+	// @off
+
+	/** Compares file extensions to be processed as image tags.
+	 * KTX is supported but don't add it here! */
 	private boolean isImage (String extension) {
-		return extension.equals("jpg") || extension.equals("jpeg") || extension.equals("png") || extension.equals("bmp")
-			|| extension.equals("gif");
+		return ObjectSet.with(
```
tommyettinger: I have no idea how often isImage(), isText(), etc. get called at startup, but it's not common for "isFoo()" methods to allocate, and users would generally not expect isImage() to create garbage per-call. It didn't before. I would store each of the sets once as `private static final ObjectSet<String>`, since they don't rely on any runtime state and don't need to be accessed outside this class.

tommyettinger (dismissed):
I like this and I think supporting WebP in particular is valuable. That said, I do think it wouldn't be hard to store the ObjectSet<String>s used to test for file extensions only once, to avoid allocating when it didn't allocate before. GWT is... not the greatest about speed in any case, so not making it do extra work would be nice.

backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/preloader/DefaultAssetFilter.java
Frosty-J: I don't think they do. Asset type is stored in `assets.txt` (the one produced by GWT module) and this will be a drop in the build time bucket. But yeah, that's fair - undoing 6734dc0 would be as you suggest.

tommyettinger: It's private, so it is still in the same scope file-wise (DefaultAssetFilter is all that can see it). I would encourage undoing that commit if it's easy enough.

tommyettinger (approved):
Looks good to me; approved.

PR #7863, opened 2026-09-01
Update iOS backend to MobiVM 2.3.26

obigu:
Updated to MobiVM 2.3.26 which has several important fixes and features (https://github.com/MobiVM/robovm/releases)


tommyettinger:
Should I update GDX-Liftoff to use MobiVM 2.3.26 by default, or will that be determined by a transitive dependency on libGDX's backend? Mostly I need to know if using MobiVM 2.3.26 will break anything in a libGDX 1.14.2 project.

obigu:
> Should I update GDX-Liftoff to use MobiVM 2.3.26 by default, or will that be determined by a transitive dependency on libGDX's backend? Mostly I need to know if using MobiVM 2.3.26 will break anything in a libGDX 1.14.2 project.

Yes, it's safe and recommended to update.

PR #7864, opened 2026-09-10
[Tiled] add list property type support

Quillraven:
This PR adds support for Tiled's `list` attributes. I used the existing `TiledMapPropertiesTest` and added a few more objects. I am not sure if I covered all cases because there are so many combinations now with classes/lists/nested classes/nested lists. Hope some of the Tiled guys like @BoBIsHere86 can have a look at it as well and try out some other cases that I have missed.

----

The solution I went for is that we store list attributes as GDX Array<Object> instances. From what I have seen this is also how Tiled stores it and a Tiled array like [1, 2, "some", "text"] becomes the LibGDX Array counterpart.


Quillraven:
@MRZ07 @damnedpie @tommyettinger notify some people from previous Tiled PRs

gdx/src/com/badlogic/gdx/maps/tiled/BaseTiledMapLoader.java
```diff
@@ -306,6 +311,39 @@ protected void loadJsonClassProperties (String className, MapProperties classPro
...
+				list.add(classProperties);
+			} else if ("object".equals(type)) {
+				// Object references are only known after the whole map is parsed, so fetch them at the end of [loadTiledMap]
+				final int id = Integer.parseInt(item.getString("value"));
+				final int index = list.size;
+				list.add(null);
+				runOnEndOfLoadTiled.add(new Runnable() {
+					@Override
+					public void run () {
+						list.set(index, idToObject.get(id));
```
tommyettinger, line 337: If anything is removed from the middle of the list before this runOnEndOfLoadTiled Runnable runs, then it will not act correctly. I don't have any better ideas here. Instead of adding null and mutating it later, maybe it could add in the Runnable? I don't know how critical it is to keep the size of the list the same.

PR #7865, opened 2026-09-21
add LWJGL3 equivalent for LWJGL2's backgroundFPS

### LINKED ISSUE #6638: `setBackgroundFPS` has no LWJGL3 equivalent, despite `setIdleFPS()` sounding like one

johny7guitar:
Adds background FPS functionality to LWJGL3.

Resolves #6638


tommyettinger (approved):
Looks like a simple and clean improvement to me. I don't know what currently happens to framerate for non-focused windows, but being able to set the behavior is a big improvement over having to guess!

